Issue #493, opened 2023-08-04
Node can't sync with the network


nazar-pc:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

We have discovered with somewhat recent Susbtrate (319a7bdd17bd257357f4db50ab1f9e3bc34fe5b2) nodes sometimes fail to keep up with the network and even after restart they are unable to sync.

We have some heavy blocks on the network and we decreased block response size in our Substrate fork (https://github.com/subspace/substrate/commit/7fceb6c4477f0aa07ddce05b59cdc2cc5e97ce4c), but all blocks are strictly smaller than this lower limit anyway, so that is not an issue.

Here is logs of a node that was previously unable to sync with `-l sync=trace`:
[dariia-node-log.txt](https://github.com/paritytech/substrate/files/12262211/dariia-node-log.txt)

I can see that it tries to download blocks backwards, but stops before reaching correctly identified common ancestor block 9715.

### Steps to reproduce

Requires "broken" state of the node (that I have access to) in order to reliably reproduce the issue.

rahulksnv:
I have a unit test that reproduces the issue. To summarize the problem:
1. Node is at `9718` on a non-canonical chain, comes back online.
2. Other nodes are on the canonical chain, with `9715` being the common ancestor
3. Node ries to sync from the other nodes (in chunks up to 64 at a time).
4. It so happens that one of the peers was requested `[9716 .. 9719]`, but only returned `9719`, possibly because block was too big:
```
New block request for 12D3KooWGpmCSuWK9nCAc3ZiZJ5wZKnk38A4huEunf1NmwTHoG2n, (best:10699, common:9715) BlockRequest { id: 0, fields: HEADER | BODY | JUSTIFICATION, from: Number(9719), direction: Descending, max: Some(4) }
BlockResponse 0 from 12D3KooWGpmCSuWK9nCAc3ZiZJ5wZKnk38A4huEunf1NmwTHoG2n with 1 blocks  (9719)
```
5. At this point the node has downloaded blocks `[9719 .. 10699]`. Since these downloads don't happen on one of the fork targets,  it concludes the chain can be extended, tries importing and fails. This is because even though the block numbers are the same, these are on different chains (i.e) the parent of the downloaded `9719` != local  (non-canonical) `9718`.

[...]

PR #1686, opened 2023-09-24
Changes to handling partial responses

rahulksnv:
Context: https://github.com/paritytech/polkadot-sdk/issues/493#issuecomment-1728917139

Problem: When downloading from a fork, the peer may send a partial response that extends the current chain. Even though the response starts at a block number after the current best number, the parent block from the fork is not yet downloaded/imported. We currently try to import the partial response in this scenario, which causes import to fail as parent is not found in the backend. This causes sync to be aborted/restarted which may not converge.

Fix: when we receive a partial response in this scenario, hold on to the partial blocks and try to download the missing blocks. And import when the full range is downloaded.

### DIFF
--- a/substrate/client/network/sync/src/lib.rs
+++ b/substrate/client/network/sync/src/lib.rs
@@ -428,15 +428,8 @@ pub struct PeerDownloadState<B: BlockT> {
 }
 
 impl<B: BlockT> PeerDownloadState<B> {
-	fn new(
-		common_number: NumberFor<B>,
-		range: Range<NumberFor<B>>,
-	) -> Self {
-		Self {
-			common_number,
-			range,
-			downloaded: Vec::new(),
-		}
+	fn new(common_number: NumberFor<B>, range: Range<NumberFor<B>>) -> Self {
+		Self { common_number, range, downloaded: Vec::new() }
 	}
 
 	/// Handles the new blocks received from the peer.
@@ -447,9 +440,9 @@ impl<B: BlockT> PeerDownloadState<B> {
 	fn handle_blocks(
 		&mut self,
 		who: &PeerId,
-		mut new_blocks:  Vec<BlockData<B>>,
+		mut new_blocks: Vec<BlockData<B>>,
 		request: BlockRequest<B>,
-		collection: &mut BlockCollection<B>
+		collection: &mut BlockCollection<B>,
 	) -> Result<bool, BadPeer> {
 		collection.clear_peer_download(who);
 
@@ -463,14 +456,15 @@ impl<B: BlockT> PeerDownloadState<B> {
 		// Add the new blocks to the downloaded list.
 		let range_len = (self.range.end - self.range.start).saturated_into::<u32>() as usize;
 		if (self.downloaded.len() + new_blocks.len()) > range_len {
-			return Err(BadPeer(*who, rep::BAD_RESPONSE));
+			return Err(BadPeer(*who, rep::BAD_RESPONSE))
 		}
 		let new_blocks_len = new_blocks.len();
 		new_blocks.append(&mut self.downloaded);
 		self.downloaded.append(&mut new_blocks);
 
 		if self.range.start == (self.common_number + One::one()) &&
-			self.downloaded.len() < range_len {
+			self.downloaded.len() < range_len
+		{
 			// The download extends the common ancestor, but the response
 			// was partial. We want to issue further requests to fetch
 			// the incomplete part before importing.
@@ -480,7 +474,7 @@ impl<B: BlockT> PeerDownloadState<B> {
 				common = {}, range = {:?}, new_blocks = {new_blocks_len}, downloaded = {}",
 				self.common_number, self.range, self.downloaded.len()
 			);
-			return Ok(false);
+			return Ok(false)
 		}
 
 		// Done, report the accumulated blocks to the collection.
@@ -497,8 +491,9 @@ impl<B: BlockT> PeerDownloadState<B> {
 		peer_best_number: NumberFor<B>,
 		peer_best_hash: &B::Hash,
 	) -> Option<(Range<NumberFor<B>>, BlockRequest<B>)> {
-		let downloaded_head = if let Some(number) = self.downloaded.first()
-			.and_then(|b| b.header.as_ref()).map(|h| *h.number()) {
+		let downloaded_head = if let Some(number) =
+			self.downloaded.first().and_then(|b| b.header.as_ref()).map(|h| *h.number())
+		{
 			number
 		} else {
 			warn!(
@@ -507,12 +502,9 @@ impl<B: BlockT> PeerDownloadState<B> {
 				common = {}, range = {:?}, downloaded = {}",
 				self.common_number, self.range, self.downloaded.len()
 			);
-			return None;
+			return None
 		};
-		let range = Range {
-			start: self.range.start,
-			end: downloaded_head,
-		};
+		let range = Range { start: self.range.start, end: downloaded_head };
 
 		// The end is not part of the range.
 		let last = range.end.saturating_sub(One::one());
@@ -835,17 +827,17 @@ where
 								// Done with the range, go back to available state.
 								peer.state = PeerSyncState::Available;
 								self.ready_blocks()
-							}
+							},
 							Ok(false) => {
 								// Not done yet, leave it in DownloadingNew state.
 								vec![]
-							}
+							},
 							Err(err) => {
 								peer.state = PeerSyncState::Available;
-								return Err(err);
-							}
+								return Err(err)
+							},
 						}
-					}
+					},
 					PeerSyncState::DownloadingGap(_) => {
 						peer.state = PeerSyncState::Available;
 						if let Some(gap_sync) = &mut self.gap_sync {
@@ -2155,11 +2147,8 @@ where
 					Some((id, ancestry_request::<B>(current)))
 				} else if let PeerSyncState::DownloadingNew(download_state) = &peer.state {
 					// We have partially downloaded range, request the remaining part.
-					let result = download_state.peer_block_request(
-						attrs,
-						peer.best_number,
-						&peer.best_hash
-					);
+					let result =
+						download_state.peer_block_request(attrs, peer.best_number, &peer.best_hash);
 					trace!(
 						target: LOG_TARGET,
 						"Continuing block request for {}, (best:{}, common:{}) {:?}",
@@ -2168,8 +2157,7 @@ where
 						peer.common_number,
 						result,
 					);
-					result
-						.map(|(_, req)| (id, req))
+					result.map(|(_, req)| (id, req))
 				} else if let Some((range, req)) = peer_block_request(
 					&id,
 					peer,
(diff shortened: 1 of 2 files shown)


altonen:
bot fmt

rahulksnv:
> bot fmt

I don't understand how the fmt works with these changes. I try running cargo fmt locally (or with the command from the failing fmt job), but it formats several files not part of the change, or most of code in same file that the PR does not touch. The contributor guidelines also does not talk about this much.

altonen:
@rahulksnv  did you run `cargo +nightly fmt`?

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,122 @@ impl<B: BlockT> PeerSyncState<B> {
...
+		};
+
+		// Add the new blocks to the downloaded list.
+		let range_len = (self.range.end - self.range.start).saturated_into::<u32>() as usize;
+		if (self.downloaded.len() + new_blocks.len()) > range_len {
+			return Err(BadPeer(*who, rep::BAD_RESPONSE))
+		}
+		let new_blocks_len = new_blocks.len();
+		new_blocks.append(&mut self.downloaded);
+		self.downloaded.append(&mut new_blocks);
```
altonen, line 463: What is the purpose of this? Since it's dealing with the block bodies which can be large, it'd be better if we'd tried to handle them efficiently. Looking at this code alone I don't know whether it's making full copies of them or not but this code looks quite dubious in the first place. Why does it first have to append blocks from `self.downloaded` to `new_blocks` and then append them back to `self.downloaded`?

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,122 @@ impl<B: BlockT> PeerSyncState<B> {
...
+
+		if self.range.start == (self.common_number + One::one()) &&
+			self.downloaded.len() < range_len
```
altonen, line 466: Comment for this check please. Otherwise it'll be hard to maintain this code because looking at this alone, I don't know what `self.range.start == (self.common_number + One::one())` implies

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,122 @@ impl<B: BlockT> PeerSyncState<B> {
...
+				common = {}, range = {:?}, new_blocks = {new_blocks_len}, downloaded = {}",
+				self.common_number, self.range, self.downloaded.len()
+			);
+			return Ok(false)
+		}
+
+		// Done, report the accumulated blocks to the collection.
+		let mut downloaded = Vec::new();
+		downloaded.append(&mut self.downloaded);
+		collection.insert(start_block, downloaded, *who);
```
altonen, line 483: Same comment here, why is this `append()` necessary?

nazar-pc:
@altonen this repo should seriously consider using Nightly in `rust-toolchain.toml` and run tests on stable with override in CI. I see every team member on our side is confused with how things are organized right now.

rahulksnv:
> @rahulksnv did you run `cargo +nightly fmt`?

Yes, I also ran the cmd from the failing job, but same behavior

nazar-pc:
@rahulksnv update nightly to latest and `cargo +nightly fmt` will do the job

substrate/client/network/sync/src/lib.rs
rahulksnv: I had comments below, moved it up and expanded it

substrate/client/network/sync/src/lib.rs
rahulksnv: This prepends the new blocks to the existing blocks (updated the comment to reflect it). Prepending is needed because the blocks are returned in the reverse order (i.e) the request sets `request.from = FromBlock::Number(x)`, and `request.direction = Direction::Descending`. 

The `Vec::append()` is a  move only operation, no copies/clones involved. The first append appends/empties `self.downloaded`, second one moves it back

substrate/client/network/sync/src/lib.rs
rahulksnv: This is similar, it moves the completed blocks out and passes it to the `BlockCollection`

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,122 @@ impl<B: BlockT> PeerSyncState<B> {
...
+		{
+			number
+		} else {
+			warn!(
+				target: LOG_TARGET,
+				"Download state: unable to get block number of the downloaded blocks: \
+				common = {}, range = {:?}, downloaded = {}",
+				self.common_number, self.range, self.downloaded.len()
+			);
+			return None
```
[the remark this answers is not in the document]
rahulksnv: This is the existing behavior when blocks are validated: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/network/sync/src/lib.rs#L2796

Blocks are validated before adding to the `self.downloaded` list, so this should never happen (why this is a `warn` message). Initially thought of caching the block number when it is added to the list, but that would be redundant fields, and we need to worry about keeping them in sync

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,122 @@ impl<B: BlockT> PeerSyncState<B> {
...
+		};
+		let range = Range { start: self.range.start, end: downloaded_head };
+
+		// The end is not part of the range.
+		let last = range.end.saturating_sub(One::one());
+		let from = if peer_best_number == last {
+			FromBlock::Hash(*peer_best_hash)
+		} else {
+			FromBlock::Number(last)
+		};
```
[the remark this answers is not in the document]
rahulksnv: This is existing behavior, just retained as is: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/network/sync/src/lib.rs#L2561

dmitry-markin (commented):
Looks very promising, left some comments.

substrate/client/network/sync/src/lib.rs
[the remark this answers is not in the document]
rahulksnv: The `request.from` is the hash of the first block we want to download, which we don't know until the block is downloaded (now I understand why current code sets it hash if `block_number == peer.best_block_number`, in this case we happen to know the corresponding `peer.best_block_hash`)

Unless we want to set it to parent hash of the first block already downloaded. I would prefer to keep this uniform with existing code, don't know why current code does not set it to parent hash.

substrate/client/network/sync/src/lib.rs
```diff
@@ -407,6 +407,133 @@ impl<B: BlockT> PeerSyncState<B> {
...
+			Err(err) => return Err(err),
+		};
+
+		// Prepend the new blocks to the downloaded list.
+		let range_len = (self.range.end - self.range.start).saturated_into::<u32>() as usize;
+		if (self.downloaded.len() + new_blocks.len()) > range_len {
+			return Err(BadPeer(*who, rep::BAD_RESPONSE))
+		}
+		let new_blocks_len = new_blocks.len();
+		new_blocks.append(&mut self.downloaded);
```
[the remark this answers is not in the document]
rahulksnv: Refactored this part to break into smaller functions, added unit tests for the new paths

substrate/client/network/sync/src/lib.rs
rahulksnv: > This check self.range.start == (self.common_number + One::one()) looks like a hackish way to determine if PeerDownloadState is responsible for accumulating blocks, or BlockCollection is responsible for this on the upper level in ChainSync. May be it's possible to incorporate this code into BlockCollection so that not distribute the logic over multiple components?

`BlockCollection` currently does something slightly different: manages the downloaded range (with holes + merging holes) across several peers, which the `PeerDownloadState` manages a range with a single peer. It would be cleaner/less complex to keep this separation. The `PeerDownloadState` is just a staging area before adding the blocks to the collection.

> If for some reasons it's not possible to incorporate this code into BlockCollection, I would introduce a clear separation of the code needed to pass only one chunk to BlockCollection, and the code responsible for holding up the blocks. For example, this can be done by introducing a separate enum variant into PeerSyncState for the latter (DownloadingRange) with comments that it's used specifically for fork downloading in multiple chunks.

This change does the same thing, except it enhances the state maintained by `PeerSyncState::DownloadingNew`, to manage download of new blocks from a single/unified path. I feel that adding separate variant only adds more complexity/code duplication

substrate/client/network/sync/src/lib.rs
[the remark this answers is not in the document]
dmitry-markin: Yes, I meant setting it to the parent hash of the already downloaded first block in the range. Otherwise we can't be sure that we are downloading the right fork and can fail when importing the range.

substrate/client/network/sync/src/lib.rs
dmitry-markin: The check `self.range.start == (self.common_number + One::one())` looks confusing and can be replaced with a single flag `download_full_range`, the value of which we know from the creation of `PeerDownloadState`. Otherwise we don't need `self.common_number` for anything. This is mostly equivalent to having a dedicated enum type for the case when we need to download the full range.

Having said that, IMO updating `BlockCollection` / importing logic to trigger import only when the parent hash corresponds to the already imported block (instead of comparison by number) would be less hacky than introducing a special case for downloading a range starting from `common_number`.

substrate/client/network/sync/src/lib.rs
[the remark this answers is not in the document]
rahulksnv: Changed it to pass parent hash

substrate/client/network/sync/src/lib.rs
rahulksnv: > The check self.range.start == (self.common_number + One::one()) looks confusing and can be replaced with a single flag download_full_range, the value of which we know from the creation of PeerDownloadState. Otherwise we don't need self.common_number for anything. This is mostly equivalent to having a dedicated enum type for the case when we need to download the full range.

Yeah this looks cleaner, changed it to have a separate enum. The current `DownloadingNew` remains unchanged

> Having said that, IMO updating BlockCollection / importing logic to trigger import only when the parent hash corresponds to the already imported block (instead of comparison by number) would be less hacky than introducing a special case for downloading a range starting from common_number.

Agreed. I looked at it, this looks like a much bigger change to use block hash instead of block number (this potentially touches several other paths). I feel this is better done as a separate change

dmitry-markin:
@rahulksnv sorry for not getting back with the review of PR for quite some time. We have been having internal discussions regarding the bug and the PR, and, to be honest, have not decided yet if we should merge this PR or properly fix the underlying issue (import triggered by number instead of parent hash) instead. We are currently looking if it's feasible to properly fix the import.

I'm going to publish the review comments I've already written, but it may be a good idea to hold back and wait until we have a clear understanding of what way of fixing the issue to prefer before addressing them. The review is not complete either, so this is another reason why I'd ask you to wait. We'll try to get back with the decision next week.

dmitry-markin:
CC @arkpar

dmitry-markin:
@rahulksnv We can probably close this PR now as superseded by https://github.com/paritytech/polkadot-sdk/pull/1812.

rahulksnv:
> @rahulksnv We can probably close this PR now as superseded by #1812.

yeah, closing this in favor of https://github.com/paritytech/polkadot-sdk/pull/1812

PR #1687, opened 2023-09-25
contracts: Fix incorrect storage alias in mirgration

kvinwang:
# Description

We are recently trying to upgrade our Substrate version from `polkadot-v0.9.43` to `polkadot-v1.0.0` and we noticed a critical issue: all deployed contracts seem to be experiencing a `CodeNotFound` error. After a thorough investigation, it appears that the root cause of this issue lies in the mismatch between the storage alias of `CodeInfoOf<T>` in the migration and its original definition.

This PR corrects the storage alias to align it with its original definition

I am uncertain about the proper approach for adding tests to this change. Would the team consider taking over this PR to bring it to completion?

### DIFF
--- a/substrate/frame/contracts/src/migration.rs
+++ b/substrate/frame/contracts/src/migration.rs
@@ -337,7 +337,7 @@ impl<T: Config, const TEST_ALL_STEPS: bool> OnRuntimeUpgrade for Migration<T, TE
 
 		log::info!(target: LOG_TARGET, "=== POST UPGRADE CHECKS ===");
 
-		// Ensure that the hashing algorithm is correct for each storage maps.
+		// Ensure that the hashing algorithm is correct for each storage map.
 		if let Some(hash) = crate::CodeInfoOf::<T>::iter_keys().next() {
 			crate::CodeInfoOf::<T>::get(hash).expect("CodeInfo exists for hash; qed");
 		}
--- a/substrate/frame/contracts/src/migration.rs
+++ b/substrate/frame/contracts/src/migration.rs
@@ -328,6 +328,32 @@ impl<T: Config, const TEST_ALL_STEPS: bool> OnRuntimeUpgrade for Migration<T, TE
 		);
 		Ok(Default::default())
 	}
+
+	#[cfg(feature = "try-runtime")]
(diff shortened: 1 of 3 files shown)


pgherveou (changes requested):
This might break everyone else
Looks like you might have deployed a sha in between these 2 commits where the regression was introduced


introduced here: https://github.com/paritytech/substrate/pull/14661
rolled back here https://github.com/paritytech/substrate/pull/14079

Let's discuss to find the best way to fix that

pgherveou (approved):
mm let's work this out

pgherveou:
> Thanks!
> 
> > I am uncertain about the proper approach for adding tests to this change. Would the team consider taking over this PR to bring it to completion?
> 
> This is an interesting question. `try-runtime` testing of the migrations currently can't catch this. We should think on how to improve this.

Added some integrity check in the last commit, to ensure that storage map items are encoded with the proper hashing algorithm

jasl:
@pgherveou could you backport to the `release-polkadot-v1.1.0` branch? so we can avoid monkey patching it

pgherveou:
sure let met push a PR for this

jasl:
> @jasl made a PR here, waiting for the release team to approve and merge it

Thank you!

substrate/frame/contracts/src/migration.rs
```diff
@@ -328,6 +328,32 @@ impl<T: Config, const TEST_ALL_STEPS: bool> OnRuntimeUpgrade for Migration<T, TE
...
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(_state: Vec<u8>) -> Result<(), TryRuntimeError> {
+		if !TEST_ALL_STEPS {
+			return Ok(())
+		}
+
+		log::info!(target: LOG_TARGET, "=== POST UPGRADE CHECKS ===");
+
+		// Ensure that the hashing algorithm is correct for each storage map.
+		if let Some(hash) = crate::CodeInfoOf::<T>::iter_keys().next() {
```
kvinwang, line 341: Would the `iter_keys().next()` returns an incorrect hash or just returns `None` in case of a hash mismatching?

pgherveou: it would return an incorrect hash since keys have been hashed with Twox64Concat during the migration

PR #1642, opened 2023-09-20
update `genesis_config` related docs and tests and error messages

kianenigma:
Follow-up to https://github.com/paritytech/substrate/pull/14306.

I hope this also showcases the important message of: **It is really not that hard to make the examples codes in rust-docs compile, and therefore remain correct. Please embrace this :)** 

It moves the documentation of proc macros to their re-export, such that can link other items in frame-support. This is a patter that we should embrace for all of macro docs, and apply in PRs like https://github.com/paritytech/substrate/pull/13987 as well.

Issue #247, opened 2022-10-18
Switch proc macro docs to be on re-exports so we can have doc links to outer crate items


sam0x17:
Right now the frame proc macros (especially the `pallet::` ones) follow the pattern of documenting proc macros inside their corresponding inner proc macro crates (i.e. `frame_support_procedural`). I standardized this in a number of places in paritytech/substrate#12333 / paritytech/substrate#12334.

A major reason for this is tools like `rust-analyzer` right now do not conform with the `cargo doc` behavior of prepending docs that are attached to re-exported items (i.e. macros, functions, traits, etc) and in fact completely ignore docs on re-exports (see my companion issue in `rust-analyzer`: https://github.com/rust-lang/rust-analyzer/issues/13431).

This is why in paritytech/substrate#12334 I ended up switching all of the `pallet::` macros to have their docs _inside_ `frame_support_procedural` because then `rust-analyzer` will actually pick them up upon mouse-over.

The problem with this is inner crates can only generate doc links to items that are visible from the inner crate. For non-proc-macro crates the workaround is easy -- simply set up a circular reference between the inner crate and the outer crate so the outer crate items are visible/linkable from the inner crate. This doesn't work for proc macro crates, however, because rust restricts proc macro crates so that they are only allowed to publicly export proc macros (no other items are allowed), so even if we `pub use outer_crate::*` to make all the outer crate's items accessible within the inner crate, this will fail to compile since we are trying to export something other than a proc macro in a proc macro crate.

Thus we are at a bit of a stand-still. Ideally `rust-analyzer` would handle docs on re-exports properly the way `cargo doc` does and we would switch to only documenting proc macros where their re-exports are defined in the outer crate.

So this is blocked until https://github.com/rust-lang/rust-analyzer/issues/13431 is addressed in `rust-analyzer` or some other solution (like better support for inter-crate doc linking) is added to rust.

One possible alternative solution would be to link directly to the docs.rs links for the outer items from the inner macro declarations. This has some obvious downsides (like `rust-analyzer` will open a web browser instead of following the links inline), but if this issue sits for a long period of time this might be the way to go.

Note: the one thing I _haven't_ tried is trying to `pub use` the outer crate items in the inner proc macro crate while gating this with a `#[cfg(doc)]` directive. I doubt it will compile but it's the only thing I haven't tried so would be cool if someone could confirm (or if someone knows) that this doesn't work.

## Related Issues
* paritytech/substrate#12333
[...]

PR #1689, opened 2023-09-25
tweak pallet macro (genesis_config etc) to cater for RA users as well.

kianenigma:
Small tweak to https://github.com/paritytech/polkadot-sdk/pull/1642 to incorporate the ideas from https://github.com/paritytech/polkadot-sdk/issues/247.

I think this is the good middle ground, where we have good rust-docs, and the RA users will also have some hope. 

cc @wentelteefje @aaronbassett @sam0x17


sam0x17 (approved):
yes, this is a good compromise

bkchr (approved):
I would prefer we just fix the issue in RA :P

wentelteefje (approved):

liamaharon (approved):

PR #1696, opened 2023-09-25
Contracts: migration v15 acc dec_consumers check

pgherveou:


pgherveou:
@juangirini  when you get back any idea why the consumer refcount could be 0 pre-migration?

Screenshot 2023-09-25 at 14 06 01

agryaznov (approved):

juangirini (changes requested):
Please see https://github.com/paritytech/substrate/pull/14589#discussion_r1285257035, we don't need such a check. If we can't decrease the consumers then there is an error and it is ok to print it. Anyway the flow continues and the rest executes correctly, the only thing that it is done with the conditional is to hide the error.

> @juangirini any idea why the consumer refcount could be 0 pre-migration in some cases?

The error was introduced on [v10](https://github.com/paritytech/substrate/pull/14045/files#diff-a10c326cb0e5ba98e458634fa00b1eb196037d994a356b6a598fb0735d1756e4) when the deposit account was created without incrementing the consumers. As the deposits accounts don't exist anymore it is ok not to fix that v10 bug as is not relevant anymore.

We should close this PR

pgherveou:
Thanks Juan, maybe we should add a comment on the code to explain what you just said. Afk but will update later

PR #1700, opened 2023-09-25
Add query_weight_to_fee

hqwangningbo:
# Description

- What does this PR do?
Add a query_weight_to_fee Rpc method to the transaction-payment pallet.

If I want to send an XCM message, such as WithdrawAsset + BuyExecution + Transact, at this time it is necessary to fill in an estimated fee, through a simple calculation I can get an accurate Weight (3 * unitinstrweight + require_weight_at_most), but how to convert the Weight into fee is a more cumbersome thing, so it needs to be such an rpc to get an accurate XCM execution fee


bkchr:
There is already the runtime api. You don't need a new RPC. Just use `state_call` to call the runtime api.

PR #1703, opened 2023-09-25
contracts: Fix incorrect storage alias in mirgration (#1687)

pgherveou:
Backporting a migration bug from #1687 to 1.1.0

### DIFF
--- a/substrate/frame/contracts/src/migration.rs
+++ b/substrate/frame/contracts/src/migration.rs
@@ -328,6 +328,32 @@ impl<T: Config, const TEST_ALL_STEPS: bool> OnRuntimeUpgrade for Migration<T, TE
 		);
 		Ok(Default::default())
 	}
+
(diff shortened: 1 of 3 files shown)


pgherveou:
@kvinwang are you able to just use v1.2.0 branch that was cut after this bug fix?
If yes let's close this PR

kvinwang:
> @kvinwang are you able to just use v1.2.0 branch that was cut after this bug fix?
> If yes let's close this PR

It's OK to me. However, that would mean `v1.1.0` and `v1.0.0` would be invalid for users running contracts.

pgherveou:
> > @kvinwang are you able to just use v1.2.0 branch that was cut after this bug fix?
> > If yes let's close this PR
> 
> It's OK to me. However, that would mean `v1.1.0` and `v1.0.0` would be invalid for users running contracts.

yes unless you start from this branch and don't need migration

PR #1705, opened 2023-09-25
Paired-key Crypto Scheme

drskalman:
BEEFY needs two cryptographic keys at the same time. Validators should sign BEEFY payload using both ECDSA and BLS key. The network will gossip a payload which contains a valid ECDSA key. The prover nodes aggregate the BLS keys if aggregation fails to verifies the validator which provided a valid ECDSA signature but an invalid BLS signature is subject to slashing.

As such BEEFY session should be initiated with both key. Currently there is no straight forward way of doing so, beside having a session with RuntimeApp corresponding to a crypto scheme contains both keys.

This pull request implement a generic paired_crypto scheme as well as implementing it for (ECDSA, BLS) pair.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3792810

substrate/primitives/core/src/paired_crypto.rs
```diff
@@ -0,0 +1,837 @@
...
+/// ECDSA and BLS-377 paired crypto scheme
+#[cfg(feature = "bls-experimental")]
+pub mod ecdsa_n_bls377 {
+	use crate::crypto::CryptoTypeId;
+	use crate::{bls377, ecdsa};
+
+	/// An identifier used to match public keys against BLS12-377 keys
+	pub const CRYPTO_ID: CryptoTypeId = CryptoTypeId(*b"ecb7");
+
+	const PUBLIC_KEY_LEN: usize = 33 + crate::bls::PUBLIC_KEY_SERIALIZED_SIZE;
```
davxy, line 46: Can we pick these constants to ecdsa module?

drskalman:
I could change the sign algorithm to sign the second signature using the first key. This potentially might give us a 7x speed up on verification time during validators gossip of BEEFY messages:
```
test double::tests::test_bls_verify_signatures_chaum_pedersen  ... bench:   2,967,600 ns/iter (+/- 15,817)                                                                                                                                                    
test double::tests::test_bls_verify_signatures_simple_bls      ... bench:   4,304,517 ns/iter (+/- 6,141,991)                                                                                                                                                 
test double::tests::test_verify_EdDSA_signatures               ... bench:     352,084 ns/iter (+/- 1,802)                                                                                                                                                     
test double::tests::test_verify_Schnorr_on_secp256k1_signatures ... bench:     394,758 ns/iter (+/- 2,064)                                                                                                                                                     
```
I couldn't find an readily implemented ECDSA for Arkworks so I used Schnorr on Secp256k1 instead but I don't expect a huge difference between ECDSA and a single schnorr verification. The result is on par with [this experiment](https://www.dash.org/blog/bls-is-it-really-that-slow/#:~:text=The%20benchmark%20run%20shows%20us,block%20with%201000%20BLS%20signatures.) showin BLS verification being 10 times slower than ECDSA.

substrate/primitives/core/src/paired_crypto.rs
```diff
@@ -0,0 +1,837 @@
...
+
+// }
+
+#[cfg(feature = "full_crypto")]
+impl<LeftPublic: PublicKeyBound, RightPublic: PublicKeyBound, const LEFT_PLUS_RIGHT_LEN: usize>
+	sp_std::hash::Hash for Public<LeftPublic, RightPublic, LEFT_PLUS_RIGHT_LEN>
+{
+	fn hash<H: sp_std::hash::Hasher>(&self, state: &mut H) {
+		self.left.hash(state);
+		self.right.hash(state);
```
davxy, line 156:
```suggestion
		self.inner.hash(state);
```

davxy (commented):
First pass.

In the meantime I'm going to read the interesting article you shared :-)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3887587

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3888101

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3888130

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3888256

drskalman:
@davxy I believe I have applied all your comments and suggestions to the code.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3888697

davxy:
I opened a PR on this branch with some tweaks

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3895144

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3895682

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3896114

davxy (commented):
Painful final point. We need to remove the `unwraps`.

I see two viable options:

Option 1. Remove the constructed couple of public and signature from inside `Public` and `Signature` (leaving only inner...). I don't think there is much overhead to construct the inners on the fly when are required instead of keeping them constructed.
- For what concerns the paired `Signature`, typically it is "constructed -> verified -> dropped". 
So doesn't change a lot if we construct the inner components immediately or at verification time.
- For what concerns the paired `Public`, do we need to check more than one signature with the same key in the typical workflow? Otherwise is the same as the signature.

Option 2. Add some bounds to make the current constructions unfallible.
   For example, in paired `UncheckedFrom` you'd need to add the `UncheckedFrom<[u8, LEFT_LENGTH>` bound to the `LeftPublic`.
   
   WDYT?

drskalman:
> WDYT?

I think currently it gets the public key from the vote messages. Nonetheless, I could imagine that someone would take advantage of the optimization.

It is less relevant with the signatures, but we also for example do other stuff with the signatures, like aggregating them which in BLS, we need them to be in the form of points on the curve rather than the serialized form. But that is much less common than the public key reuse scenario.

So I'll try to make it infallible with constraining first.

drskalman:
Doing this:
```
impl<LeftPublic: PublicT, RightPublic: PublicT, const LEFT_PLUS_RIGHT_LEN: usize, >                                                                           
        UncheckedFrom<[u8; LEFT_PLUS_RIGHT_LEN]> for Public<LeftPublic, RightPublic, LEFT_PLUS_RIGHT_LEN> where LeftPublic : UncheckedFrom<[u8; LeftPublic::LEN]>
```

Results in: 
```
error: generic parameters may not be used in const operations                                                                                                 
   --> substrate/primitives/core/src/paired_crypto.rs:209:138                                                                                                 
    |                                                                                                                                                         
209 | ...e LeftPublic : UncheckedFrom<[u8; LeftPublic::LEN]>                                                                                                  
    |                                      ^^^^^^^^^^^^^^^ cannot perform const operation using `LeftPublic`                                                  
    |                                                                                                                                                         
    = note: type parameters may not be used in const expressions                                                                                              
```
I'm can not introduce a new const generic only for the implementation of UncheckedFrom because Rust complains that Public does not depends on:
```
impl<LeftPublic: PublicT, RightPublic: PublicT, const LEFT_PLUS_RIGHT_LEN: usize, const LEFT_LEN: usize>                                                      
        UncheckedFrom<[u8; LEFT_PLUS_RIGHT_LEN]> for Public<LeftPublic, RightPublic, LEFT_PLUS_RIGHT_LEN> where LeftPublic : UncheckedFrom<[u8; LEFT_LEN]>    
```
Results in:
```
error[E0207]: the const parameter `LEFT_LEN` is not constrained by the impl trait, self type, or predicates                                                   
   --> substrate/primitives/core/src/paired_crypto.rs:208:83                                                                                                  
    |                                                                                                                                                         
208 | impl<LeftPublic: PublicT, RightPublic: PublicT, const LEFT_PLUS_RIGHT_LEN: usize, const LEFT_LEN: usize>                                                
    |                                                                                   ^^^^^^^^^^^^^^^^^^^^^ unconstrained const parameter                   
    |                                                                                                                                                         
    = note: expressions using a const parameter must map each value to a distinct output value                                                                
    = note: proving the result of expressions other than the parameter are unique is not supported                                                            
```
This also doesn't work (not even in nightly):
```
struct foo<const L: usize, const R: usize, > 
{
 inner : [u8; L + R]
}
```
as in 
```
error: generic parameters may not be used in const operations
 --> src/main.rs:3:15
  |
3 |  inner : [u8; L + R]
  |               ^ cannot perform const operation using `L`
  |
  = help: const parameters may only be used as standalone arguments, i.e. `L`
```
This compiles only in nightly:
```
#![feature(generic_const_exprs)]
struct foo<const L: usize, const R: usize, const L_PLUS_R: usize> 
{
    inner : [u8; L_PLUS_R],
}

trait bar {
    
}

impl<const L: usize, const R: usize,> bar for foo<L, R, {L+R}>
{
}
```
with scary warnings about horrible consequences of using `generic_const_exprs` which we do not want to get involved with. Additionally it adds two more generic consts all over the code, making the code much less readable.

Note that in both of your proposed solutions we need to deal with this,  because we need to finally convert `Public::Inner` into `LeftPublic` and `RightPublic` and (their Signature counterparts) either at inception of the object (as it is now) or when we want to sign or verify a signature. While we could return `false` in verification, I assume signing need to be infallible as well.

So checking `Self::LEN = Left::LEN + Right::LEN` during runtime and in the tests, seems to be the only option for now till the compiler matures more to handle const expressions better.

drskalman:
@davxy feel free to remove the `debug_assert!`s if you don't like them.

drskalman:
@davxy I removed the internal object now it is just an array like other cryptos.

davxy (approved):
Looks good 🟢

substrate/primitives/core/src/paired_crypto.rs
```diff
@@ -0,0 +1,658 @@
...
+
+		let left = self.left.derive(path.iter().cloned(), seed.map(|s| s.into()))?;
+		let right = self.right.derive(path.into_iter(), seed.map(|s| s.into()))?;
+
+		let seed = match (left.1, right.1) {
+			(Some(l), Some(r)) if l.as_ref() == r.as_ref() => Some(l.into()),
+			_ => None,
+		};
+
+		Ok((Self { left: left.0, right: right.0 }, seed))
```
davxy, line 405: Mmm... here if the two crypto schemes have different ways to derive the seed we end up with a `None` seed. 
I don't like this a lot. But if is good for the current application I'm not going to block things for this.

Would be nice to at least write this thing explicitly over this method docs.

Lederstrumpf (approved):
tACK - and thanks for removing the seed duplication (a01a814d6764a89a55c2ef7619f25e63f3b35c18).

I recognize this change muddies the `derive` return though (https://github.com/paritytech/polkadot-sdk/pull/1705#discussion_r1354761095).

substrate/primitives/core/src/paired_crypto.rs
Lederstrumpf: @davxy something like https://github.com/paritytech/polkadot-sdk/pull/1705#discussion_r1355039462?
can also integrate directly into [`PairT::derive`](https://github.com/w3f/polkadot-sdk/blob/a069dcd810dc5f7dfea2f1fb5c259e3369961039/substrate/primitives/core/src/crypto.rs#L892-L893) documentation, although it's specific to the implementation here.

drskalman: The only way to be able to return a seed is to make the paired_crypto seed 64 bytes which is not inline with all other crypto schemes (and make it not composable i.e. you can't have a pair of paired schemes).

There is a comment in crypto.rs which says in some crypto schemes it is not possible to provide a direct seed after derivation: 

https://github.com/w3f/polkadot-sdk//blob/skalman-paired-crypto-scheme/substrate/primitives/core/src/crypto.rs#L923

We can also add paired_crypto scheme to that comment perhaps.

davxy:
> @davxy is this good to merge?

Yes

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/6

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/7

Issue #1136, opened 2021-11-23
CI to check each individual pallet compiles with supported feature configurations.


xlc:
There are many pallets (in Substrate & Polkadot) that doesn't compile by itself due to mis configured features.
The runtime compiles fine thanks for feature pollution, but it can cause trouble when working with those pallet directly.

Some related fixes that can be detected by CI:
https://github.com/paritytech/substrate/pull/6995
https://github.com/paritytech/substrate/pull/5833
https://github.com/paritytech/substrate/pull/10324
https://github.com/paritytech/polkadot/pull/4355

xlc:
I am going to leave a message here overtime I encountered a bad feature flag config.

```
error[E0046]: not all trait items implemented, missing: `ensure_successful`
   --> /Users/xiliangchen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pallet-asset-rate-4.0.0/src/lib.rs:229:1
    |
229 | impl<T> ConversionFromAssetBalance<BalanceOf<T>, AssetKindOf<T>, BalanceOf<T>> for Pallet<T>
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `ensure_successful` in implementation
    |
    = help: implement the missing item: `fn ensure_successful(_: <T as pallet::Config>::AssetKind) { todo!() }`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `pallet-asset-rate` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

ggwpez:
Where did you compile that? As part of another repo?  
Master should have proper feature configs now.

xlc:
https://github.com/open-web3-stack/open-runtime-module-library/actions/runs/6777961997/job/18422558667

PR #1537, opened 2023-09-13
CI checks features

AurevoirXavier:
To resolve issue #1136.

This is a cross verification against zepter.

- [cargo-featalign](https://github.com/hack-ink/cargo-featalign): Verifies the proper propagation of all features.
- [zepter](https://github.com/ggwpez/zepter): Checks for accidentally enabled features.

cc @ggwpez

PR #1708, opened 2023-09-26
CI check features

AurevoirXavier:
To resolve issue #1136.

This is a cross verification against zepter.

- [cargo-featalign](https://github.com/hack-ink/cargo-featalign): Verifies the proper propagation of all features.
- [zepter](https://github.com/ggwpez/zepter): Checks for accidentally enabled features.

cc @ggwpez 

---
Switch to a new branch. Original PR #1537.


bkchr (approved):

sergejparity (dismissed):

ggwpez (approved):
Do you know what this `true == true` is doing in the action? Or is this from some variable interpolation for debugging?

Screenshot 2023-10-02 at 14 35 52

I will try to fix the remaining missing features with Zepter since they are currently not detected.

AurevoirXavier:
> Do you know what this `true == true` is doing in the action? Or is this after variable interpolation for some debugging?

This flag indicates that `default-features = std`. Not all projects assume that default features are std.

AurevoirXavier:
I have decided to disable the default std features checker for this repository.

Some crates use `default-features = ["std"]`, while others do not.

For example, in https://github.com/meithecatte/enumflags2/blob/master/Cargo.toml, std is not set as its default feature. However, cargo-featalign will complain if you do not specify `default-features = false` for their crate.

ggwpez:
Okay lets see if we got all... going to rebase

AurevoirXavier:
@ggwpez I think it is ready now.

substrate/client/executor/Cargo.toml
```diff
@@ -58,6 +58,7 @@ harness = false
 default = [ "std" ]
 # This crate does not have `no_std` support, we just require this for tests
 std = [
+	"codec/std",
```
ggwpez, line 61: I actually had the same thing in Zepter until i fixed it, but it is not needed since `std` is in the `default` features for `codec` and we dont disable the default features, see https://github.com/paritytech/parity-scale-codec/blob/c781ce04706f64c9b1fffa7ed5dd5bd14ede76db/Cargo.toml#L40

AurevoirXavier: Didn't notice this before. I will update this later.

AurevoirXavier: I know why, actually cargo-featalign can handle this well.

AurevoirXavier:
I made some improvements today and came across something interesting. I'll update these tomorrow.

AurevoirXavier:
@ggwpez Found 2 problems in proc-macro crates. And it seems the `sc-exector` didn't follow the gernal std rule, so I added it to the ignore list.

.github/workflows/check-features.yml
```diff
@@ -11,8 +11,9 @@ jobs:
       - name: Fetch latest code
         uses: actions/checkout@v4
       - name: Check
-        uses: hack-ink/cargo-featalign-action@v0.1.0
+        uses: hack-ink/cargo-featalign-action@v0.1.1
```
chevdor, line 14:
```suggestion
        uses: hack-ink/cargo-featalign-action@bea88a864d6ca7d0c53c26f1391ce1d431dc7f34 # v0.1.1
```

chevdor (changes requested):

sergejparity (approved):

ggwpez:
Okay one more rebase and then we should be good to merge.

alvicsam (approved):

ggwpez:
Uhg we are waiting for approval from the audit team... the new approval rules are fun 🤡

chevdor (approved):

louismerlin (approved):

PR #1709, opened 2023-09-26
Add `MaxTipAmount` for pallet-tips

AurevoirXavier:
Last week we experienced a governance attack.
Surprisingly, there was no upper limit on the tip amount.

Due to the mechanism of pallet-fragment-election, the council members will be refreshed immediately. Attacker is easy to control the council and give a large tip amount.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3794037

bkchr (approved):

substrate/frame/tips/src/lib.rs
```diff
@@ -336,10 +342,13 @@ pub mod pallet {
 			let tipper = ensure_signed(origin)?;
 			let who = T::Lookup::lookup(who)?;
 			ensure!(T::Tippers::contains(&tipper), BadOrigin);
+
 			let reason_hash = T::Hashing::hash(&reason[..]);
 			ensure!(!Reasons::<T, I>::contains_key(&reason_hash), Error::<T, I>::AlreadyKnown);
-			let hash = T::Hashing::hash_of(&(&reason_hash, &who));
 
+			ensure!(T::MaxTipAmount::get() >= tip_value, Error::<T, I>::MaxTipAmountExceeded);
```
KiChjang, line 349: Feels like this should be done before you calculate the `reason_hash`, otherwise you'd be doing work only to be discarded later.

substrate/frame/tips/src/lib.rs
```diff
@@ -388,6 +397,9 @@ pub mod pallet {
 			ensure!(T::Tippers::contains(&tipper), BadOrigin);
 
 			let mut tip = Tips::<T, I>::get(hash).ok_or(Error::<T, I>::UnknownTip)?;
+
+			ensure!(T::MaxTipAmount::get() >= tip_value, Error::<T, I>::MaxTipAmountExceeded);
```
KiChjang, line 401: Similarly, this should go before fetching the `tip` from storage.

franciscoaguirre (approved):
I agree the `ensure!`s should go as early as possible to avoid extra work

liamaharon (approved):

substrate/frame/tips/src/lib.rs
AurevoirXavier: Yes, I have thought about this before. However, I want to maintain the same business logic as before while making it more intuitive. Moreover, I can place this code before `T::Tippers::contains()` to avoid reading from council storage.

But if you believe it would be better to put this code before the `reason_hash`, I will make that change.

KiChjang: Your comments feels like it was made from a place of fear. What kind of things do you feel you'd break when you make such a change?

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-release-analysis-v1-3-0/4614/1

PR #1714, opened 2023-09-26
Improve FRAME storage docs

wentelteefje:
This is a port (and hopefully a small improvement) of @kianenigma's PR from the old Substrate repo: https://github.com/paritytech/substrate/pull/13987.  Following #1689 I moved the documentation of all macros relevant to this PR from `frame_support_procedural` to `pallet_macros` while including a hint for RA users. 

Question: Again with respect to #1689: Is there a good reason why we should *not* enhance paths with links to our current rustdocs? For example, instead of 
```rust
/// **Rust-Analyzer users**: See the documentation of the Rust item in
/// `frame_support::pallet_macros::storage`.
```
we could write
```rust
/// **Rust-Analyzer users**: See the documentation of the Rust item in
/// [`frame_support::pallet_macros::storage`](https://paritytech.github.io/polkadot-sdk/master/frame_support/pallet_macros/attr.storage.html).
```
This results in a clickable link like this:
image
I don't really expect the links to become outdated any time soon, but I think this would be a great UX improvement over just having paths.

TODOs:
- [ ] Add documentation for `constant_name` macro
- [x] Add proper documentation for different `QueryKinds`, i.e. `OptionQuery`, `ValueQuery`, `ResultQuery`. One example for each. Custom `OnEmpty` should be moved to `QueryKinds` trait doc page.
- [ ] Rework `type_value` docs


substrate/frame/support/src/storage/types/counted_map.rs
```diff
@@ -47,6 +47,33 @@ use sp_std::prelude::*;
...
+///     #[pallet::storage]
+/// 	#[pallet::getter(fn foo)]
+/// 	#[pallet::storage_prefix = "OtherFoo"]
+/// 	#[pallet::unbounded]
+///     pub type Foo<T> = CountedStorageMap<
+/// 		Hasher = Blake2_128Concat,
+/// 		Key = u32,
+/// 		Value = u32,
+/// 		QueryKind = ValueQuery
+/// 	>;
```
kianenigma, line 74: nice, I would experiment with now writing a simple (and mostly hidden) `fn main` that spins up a `TestExternalities`, add a couple of items to this map, then call `count`. 

This might exist in some of the tests in the bottom of this file, ergo you could also use `docify`.

substrate/frame/support/src/storage/types/counted_nmap.rs
```diff
@@ -45,6 +45,36 @@ use sp_std::prelude::*;
...
+///
+/// ```
+/// #[frame_support::pallet]
+/// mod pallet {
+///     # use frame_support::pallet_prelude::*;
+///     # #[pallet::config]
+///     # pub trait Config: frame_system::Config {}
+///     # #[pallet::pallet]
+///     # pub struct Pallet<T>(_);
+/// 	/// A kitchen-sink storage map, with all possible additional attributes.
```
kianenigma, line 62:
```suggestion
/// 	/// A kitchen-sink storage N map, with all possible additional attributes.
/// 	/// It uses 3 storage keys of `u8`, `u16` and `u32`
```

Or something like that.

substrate/frame/support/src/storage/types/double_map.rs
```diff
@@ -31,22 +31,43 @@ use sp_arithmetic::traits::SaturatedConversion;
...
+/// with a value of a specified type stored on-chain.
 ///
-/// Each value is stored at:
-/// ```nocompile
-/// Twox128(Prefix::pallet_prefix())
-/// 		++ Twox128(Prefix::STORAGE_PREFIX)
-/// 		++ Hasher1(encode(key1))
-/// 		++ Hasher2(encode(key2))
-/// ```
+/// A double map with keys `k1` and `k2` can be likened to a [`StorageMap`](frame_support::storage::types::StorageMap) with a key of type `(k1, k2)`.
```
kianenigma, line 37: line wrapping broken? I suggest using https://marketplace.visualstudio.com/items?itemName=stkb.rewrap, although cargo `fmt` should also fix these.

substrate/frame/support/src/storage/types/double_map.rs
```diff
@@ -31,22 +31,43 @@ use sp_arithmetic::traits::SaturatedConversion;
...
+///     #[pallet::storage]
+/// 	#[pallet::getter(fn foo)]
+/// 	#[pallet::storage_prefix = "OtherFoo"]
+/// 	#[pallet::unbounded]
+///     pub type Foo<T> = StorageDoubleMap<
+/// 		Hasher1 = Blake2_128Concat,
+/// 		Key1 = u8,
+/// 		Hasher2 = Twox64Concat,
+/// 		Key2 = u16,
+/// 		Value = u32,
```
kianenigma, line 66: I think similarly, in all of these, we could think further about how we can bring a couple of tests for demonstration in here as well. For example, the partial iteration you mentioned is important, but there is no example for it. Would be cool to have them.

substrate/frame/support/src/storage/types/double_map.rs
```diff
@@ -31,22 +31,43 @@ use sp_arithmetic::traits::SaturatedConversion;
...
-/// 		++ Hasher1(encode(key1))
-/// 		++ Hasher2(encode(key2))
-/// ```
+/// A double map with keys `k1` and `k2` can be likened to a [`StorageMap`](frame_support::storage::types::StorageMap) with a key of type `(k1, k2)`. 
+/// However, a double map offers functions specific to each key, enabling partial iteration 
+/// and deletion based on one key alone.
+///
+/// Additionally, a double map is an alias for [`StorageNMap`](frame_support::storage::types::StorageNMap) using two keys.
+///
+/// For general information regarding the `#[pallet::storage]` attribute, refer to
```
kianenigma, line 43: I love the linking here. Specifics of double-map are here, the general things elsewhere and linked.

substrate/frame/support/src/storage/types/value.rs
```diff
@@ -30,11 +30,29 @@ use sp_arithmetic::traits::SaturatedConversion;
...
+///     # pub trait Config: frame_system::Config {}
+///     # #[pallet::pallet]
+///     # pub struct Pallet<T>(_);
+/// 	/// A kitchen-sink storage value, with all possible additional attributes.
+///     #[pallet::storage]
+/// 	#[pallet::getter(fn foo)]
+/// 	#[pallet::storage_prefix = "OtherFoo"]
+/// 	#[pallet::unbounded]
+///     pub type Foo<T> = StorageValue<_, u32, ValueQuery>;
+/// }
```
kianenigma, line 55: another idea is for you to add a single test/snippet here to each storage type where you re-create the storage key to demonstrate the point. Something like:

```
fn main() {
	let expected_key = twox(pallet_name) ++ twox(Foo)
	let key = Foo::hashed_key();
	assert_eq!(expected., key);
}
```

Sadly would require you to build a runtime, but thanks to all the recent work it is only a few lines of code now :)

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
 	/// }
 	/// ```
 	pub use frame_support_procedural::genesis_build;
+
+	/// The `#[pallet::constant]` attribute can be used to add an associated type trait bounded by `Get`
```
kianenigma, line 2266:
```suggestion
	/// The `#[pallet::constant]` attribute can be used to add an associated type trait bounded by [`Get`]
```

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	///
+	///     #[pallet::config]
+	///     pub trait Config: frame_system::Config { }
+	///
+	///     #[pallet::pallet]
+	///     pub struct Pallet<T>(_);
+	///
+	/// 	#[pallet::call]
+	/// 	impl<T: Config> Pallet<T> {
+	/// 		fn do_stuff(_origin: OriginFor<T>, arg: u32) {
```
kianenigma, line 2315: we can faily easily add something like this: 

```
fn main() {
	// showcase `pallet::Call::do_stuff { .. }`
}
```

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// 	}
+	/// }
+	/// ```
+	///
+	/// TODO: Once we have default pallet config, it would also be easy to create types in the example
+	/// that implement `Config`.
+	pub use frame_support_procedural::call;
+
+	/// Declares a type alias as a storage item. Storage items are pointers to data stored on-chain (the
+	/// *blockchain state*), under a specific key. The exact key is dependent on the type of the
```
kianenigma, line 2327: Yes I want to double down that the key location should be documented *somewhere* in this story.

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	///
+	/// ### Syntax
+	///
+	/// Two general syntaxes are supported, as demonstrated below:
+	///
+	/// 1. Named generics, e.g., `type Foo<T> = StorageValue<Value = u32>`.
+	/// 2. Unnamed generics, e.g., `type Foo<T> = StorageValue<_, u32>`.
+	///
+	/// In both instances, declaring `<T>` is mandatory. While it can optionally
+	/// be written as `<T: Config>`, in the generated code, it is always `<T: Config>`.
```
kianenigma, line 2389: wrapping needs fixing..

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// 	#[pallet::storage]
+	/// 	type Baz<T> = CountedStorageMap<_, Blake2_128Concat, u32, u32>;
+	/// }
+	/// ```
+	/// 
+	/// ## Related Macros
+	/// 
+	/// The following macros can be used in conjunction with the storage macro:
+	///
+	/// * [`macro@getter`]: Creates a custom getter function.
```
kianenigma, line 2373: I think these don't fit great into the story so far. Moving them toward the end might improve, but no sure. Do you think they are well explained in this position, in the overall narrative?

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	///
+	/// Every storage type mentioned above has a generic type called `QueryKind` that determines its "query" type. 
+	/// This refers to the kind of value returned when querying the storage, for instance, through a `::get()` method. 
+	///
+	/// There are three types of queries:
+	///
+	/// 1. [`OptionQuery`](frame_support::storage::types::OptionQuery): The default query type. It returns `Some(V)` if the value is present, or 
+	/// `None` if it isn't, where `V` is the value type.
+	/// 2. [`ValueQuery`](frame_support::storage::types::ValueQuery): Returns the value itself if present; otherwise, it returns 
+	/// `Default::default()`. This behavior can be adjusted with the `OnEmpty` generic parameter, which defaults to `OnEmpty = GetDefault`.
```
kianenigma, line 2425: Would like to see somewhere (probably in the documentation of `ValueQuery`) an example of a storage value that is u32 who's default getter returns 42 if the value is not present.

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// 3. [`ResultQuery`](frame_support::storage::types::ResultQuery): Returns `Result<V, E>`, where `V` is the value type, and `E` is the pallet error type defined by the [`#[pallet::error]`](frame_support::pallet_macros::error) attribute.
+	///
+	/// ### Appending
+	///
+	/// All storage items — such as `StorageValue`, `StorageMap`, and their variants—offer an `::append()` method
```
kianenigma, line 2430: linking? should be easy here

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// 2. [`ValueQuery`](frame_support::storage::types::ValueQuery): Returns the value itself if present; otherwise, it returns 
+	/// `Default::default()`. This behavior can be adjusted with the `OnEmpty` generic parameter, which defaults to `OnEmpty = GetDefault`.
+	/// 3. [`ResultQuery`](frame_support::storage::types::ResultQuery): Returns `Result<V, E>`, where `V` is the value type, and `E` is the pallet error type defined by the [`#[pallet::error]`](frame_support::pallet_macros::error) attribute.
+	///
+	/// ### Appending
+	///
+	/// All storage items — such as `StorageValue`, `StorageMap`, and their variants—offer an `::append()` method 
+	/// optimized for collections. Using this method avoids the inefficiency of decoding and re-encoding 
+	/// entire collections when adding items. For instance, consider the storage declaration 
+	/// `type MyVal<T> = StorageValue<_, Vec<u8>, ValueQuery>`. With `MyVal` storing a large list of bytes,
```
kianenigma, line 2433: probably best to add one good and minimal example for this here.

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// optimized for collections. Using this method avoids the inefficiency of decoding and re-encoding 
+	/// entire collections when adding items. For instance, consider the storage declaration 
+	/// `type MyVal<T> = StorageValue<_, Vec<u8>, ValueQuery>`. With `MyVal` storing a large list of bytes, 
+	/// `::append()` lets you directly add bytes to the end in storage without processing the full list. 
+	/// Depending on the storage type, additional key specifications may be needed.
+	///
+	/// ### Optimized Length Decoding
+	///
+	/// All storage items — such as `StorageValue`, `StorageMap`, and their counterparts — 
+	/// incorporate the `::decode_len()` method. This method allows for efficient retrieval
```
kianenigma, line 2440: same.

kianenigma: (all of these can be docify-ed tests that you bring here, both are fine. whichever you find more convenient). If you add a code snippet here that is replacing a test, you can remove that test and just paste an equivalent of it here.

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// ### Optimized Length Decoding
+	///
+	/// All storage items — such as `StorageValue`, `StorageMap`, and their counterparts — 
+	/// incorporate the `::decode_len()` method. This method allows for efficient retrieval 
+	/// of a collection's length without the necessity of decoding the entire dataset.
+	///
+	/// ### Hashers
+	///
+	/// For all storage types, except `StorageValue`, a set of hashers needs to be specified. The choice
+	/// of hashers is crucial, especially in production chains. The purpose of
```
kianenigma, line 2446: This is one that I would probably tackle in a ref-doc, as it is too detailed. Good for now I would say.

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,221 @@ pub mod pallet_macros {
...
+	/// 2. If the map keys are controllable by the end users, cryptographically secure hashers should
+	/// be used.
+	///
+	/// For more information, look at the types that implement
+	/// [`frame_support::StorageHasher`](frame_support::StorageHasher).
+	///
+	/// Lastly, it's recommended for hashers with "concat" to have reversible hashes. Refer to the
+	/// implementors section of [`hash::ReversibleStorageHasher`](frame_support::hash::ReversibleStorageHasher).
+	///
+	/// ### Prefixes
```
kianenigma, line 2466: ahh, so maybe you don't need to cover the key generation details of each storage type in each type, and tackle them all here?

kianenigma (commented):
Looks mostly good, but I have some improvement suggestions. 

Next, I suggest you to tackle `pallet::Config`. It is probably the last "important" one to tackle among the foundations of pallets. 

I think that would also be a good alignment exercise for us, so I can see how you would tackle one from scratch and give you feedback. Because I had written the foundations of this myself, it was hard for me to measure.

kianenigma:
> Question: Again with respect to https://github.com/paritytech/polkadot-sdk/pull/1689: Is there a good reason why we should not enhance paths with links to our current rustdocs? For example, instead of

Maybe, not against it for now. But I want you to think what would happen if/when we publish to crates.io?

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,234 @@ pub mod pallet_macros {
...
+	///
+	/// ### Syntax
+	///
+	/// Two general syntaxes are supported, as demonstrated below:
+	///
+	/// 1. Named generics, e.g., `type Foo<T> = StorageValue<Value = u32>`.
+	/// 2. Unnamed generics, e.g., `type Foo<T> = StorageValue<_, u32>`.
+	///
+	/// In both instances, declaring `<T>` is mandatory. While it can optionally be written as `<T:
+	/// Config>`, in the generated code, it is always `<T: Config>`.
```
CrackTheCode016, line 2391: Probably good to specify _why_ the trait bound `Config` is important and what this generic means in this context.  While it is obvious to us, it may not be for new readers who lack context.

> in the generated code, it is always `<T: Config>`.

So even without the explicit trait bound, `T` will apply this bound?

Example (forgive the bad formatting) of context:

```suggestion
	/// In both instances, defining the generic parameter `T` (`<T>`) is mandatory.   When compiled, `T` will 
    /// automatically include the trait bound `Config`, which allows the developer to use overarching types from the 
    /// pallet's configuration.   It may also be explicitly declared where `T` may be bounded by `Config` (`<T: 
    /// Config>`).  
```

substrate/frame/support/src/lib.rs
```diff
@@ -2262,6 +2262,234 @@ pub mod pallet_macros {
...
+	/// 
+	/// The following macros can be used in conjunction with the storage macro:
+	///
+	/// * [`macro@getter`]: Creates a custom getter function.
+	/// * [`macro@storage_prefix`]: Overrides the default prefix of the storage item.
+	/// * [`macro@unbounded`]: Declares the storage item as unbounded.
+	///
+	/// ## Common Details to All Storage Types
+	///
+	/// The following details are relevant to all of the aforementioned storage types.
```
CrackTheCode016, line 2381: I think this could be a great section with more elaboration of what the API needs to be consumed.  For example, a high-level explanation of all the factors that using of storage type would fit well here.

If you agree, I can write a more elaborate section.

```suggestion
	/// ## Storage Type Usage
	///
	/// The following details are relevant to all of the aforementioned storage types.  Storage types require the 
    /// following items ... (bullet list of all arguments / generic params here)
```

CrackTheCode016 (commented):
Great work being done here, it is really important to make the usage of these developer-facing APIs super clear.  

I left a few preliminary comments and thoughts, and would be happy to contribute or elaborate more if needed. 

Mostly, I think that documentation for something like this should be: 

- Rife with examples (looks like that's on its way so far)
- Concise bullet point lists when lots of specific APIs are involved 
- Parameters for consuming a particular API be made immediately obvious and available
- Quick overviews for how something works, followed by examples of that thing
- For context specific things, like crypto algos / key formats, those types should be listed as supported as well, then also linked to their respective docs

Again, if any of my feedback is deemed helpful, I can help with it as well :)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3971934

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3972224

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3985931

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3987495

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3989421

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3991823

sam0x17 (approved):
great use of docify!

substrate/frame/support/src/lib.rs
wentelteefje: thanks, I included a section like this :)

substrate/frame/support/src/lib.rs
wentelteefje: I have given an example for StorageValue and StorageMap. However, I'm still thinking about whether I should add them additionally for each type in their respective pages...

substrate/frame/support/src/storage/types/map.rs
```diff
@@ -31,19 +31,35 @@ use sp_arithmetic::traits::SaturatedConversion;
...
+///     # use frame_support::pallet_prelude::*;
+///     # #[pallet::config]
+///     # pub trait Config: frame_system::Config {}
+///     # #[pallet::pallet]
+///     # pub struct Pallet<T>(_);
+/// 	/// A kitchen-sink StorageMap, with all possible additional attributes.
+///     #[pallet::storage]
+/// 	#[pallet::getter(fn foo)]
+/// 	#[pallet::storage_prefix = "OtherFoo"]
+/// 	#[pallet::unbounded]
```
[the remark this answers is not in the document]
kianenigma: I think what each of these should do is exactly something that should be explained elsewhere.

kianenigma (approved):
LGTM, it is not perfect but HUGE step in the right direction, rather merge soon and iterate later.

It is part of https://github.com/paritytech/polkadot-sdk-docs/issues/31

wentelteefje:
> LGTM, it is not perfect but HUGE step in the right direction, rather merge soon and iterate later.

Yes, please go ahead. I still have quite a few ideas, but for these I'll open separate PRs later on. Curious to see how this develops :)

kianenigma:
@wentelteefje feel free to sign up to continuing with this or other macros in https://github.com/paritytech/polkadot-sdk-docs/issues/31

PR #1715, opened 2023-09-26
Add telemetry reports for backed and included candidates

AndreiEres:
# Description

To improve the observability of validators nodes as a part of parachains consensus added telemetry reports for backed and included candidates:
`parachains.candidate_backed`
`parachains.candidate_included`


bkchr:
@AndreiEres what is the status of this? Do we still want this?

PR #1717, opened 2023-09-26
frame-support: `RuntimeDebug\Eq\PartialEq` impls for `Imbalance`

muharem:
Derive `RuntimeDebug\Eq\PartialEq` but do not bound any generics.

This achieved by using their equivalent no bound versions: `EqNoBound\PartialEqNoBound\RuntimeDebugNoBound`.

Deriving  with `Debug`, `Eq`, and `PartialEq` for the `Debt` and `Credit` type aliases of `Imbalance` is not feasible due to the `OnDrop` and `OppositeOnDrop` generic types lacking implementations of the same traits.

This absence posed challenges in testing and any scenarios that demanded the traits implementations for the type.


sam0x17 (approved):
Based on our convo, this makes sense, but curious what others have to say

muharem:
ah a better solution would be to make `RuntimeDebug` derive ignore any `PhantomData`. I will try this

bkchr (approved):

substrate/frame/support/src/traits/tokens/fungible/imbalance.rs
```diff
@@ -141,6 +141,24 @@ impl<B: Balance, OnDrop: HandleImbalanceDrop<B>, OppositeOnDrop: HandleImbalance
...
+#[cfg(any(feature = "std", feature = "force-debug"))]
+impl<B: Balance, OnDrop: HandleImbalanceDrop<B>, OppositeOnDrop: HandleImbalanceDrop<B>>
+	sp_std::fmt::Debug for Imbalance<B, OnDrop, OppositeOnDrop>
+{
+	fn fmt(&self, fmt: &mut sp_std::fmt::Formatter<'_>) -> sp_std::fmt::Result {
+		fmt.debug_struct("Imbalance").field("amount", &self.amount).finish()
+	}
+}
+
+#[cfg(all(not(feature = "std"), not(feature = "force-debug")))]
```
bkchr, line 153: We can probably drop this and just keep the `Debug` implementation all the time

muharem: we can drop the names of `OnDrop` and `OppositeOnDrop` from output, but they also might be helpful to understand neither this is `Credit` or `Dept`

muharem:
@sam0x17 @bkchr 
I also included the impls for `Eq` and `PartialEq`.
Title, and description is updated.
Can you please have a look again?

KiChjang (approved):

sam0x17:
> ah a better solution would be to make `RuntimeDebug` derive ignore any `PhantomData`. I will try this

Ah apologies, this is what I was trying to suggest but I think I wasn't clear enough

substrate/frame/support/src/traits/tokens/fungible/imbalance.rs
bkchr: I just meant to not print `wasm:stripped` in wasm.

substrate/frame/support/src/traits/tokens/fungibles/imbalance.rs
```diff
@@ -158,6 +157,58 @@ impl<
...
+}
+
+impl<
+		A: AssetId,
+		B: Balance,
+		OnDrop: HandleImbalanceDrop<A, B>,
+		OppositeOnDrop: HandleImbalanceDrop<A, B>,
+	> Eq for Imbalance<A, B, OnDrop, OppositeOnDrop>
+{
+}
```
bkchr, line 179: You should be able to just use `EqNoBound` and `PartialEqNoBound` for this?

muharem: Nice, I was not aware of them.

muharem:
now no using equivalent derives, but no bound versions - `EqNoBound\PartialEqNoBound\RuntimeDebugNoBound`.
thanks @bkchr

bkchr (approved):

PR #1719, opened 2023-09-26
Use `Extensions` to register offchain worker custom extensions

### LINKED ISSUE #1671: Offchain worker custom extension not working

### Description of bug

Offchain worker configuration allows passing custom extensions in `node/src/service.rs`(as in substrate-template repo):

https://github.com/paritytech/polkadot-sdk/blob/cfc6cc65cc133f69f472149b2449a673597e8530/substrate/client/offchain/src/lib.rs#L140

which are later registered:

https://github.com/paritytech/polkadot-sdk/blob/cfc6cc65cc133f69f472149b2449a673597e8530/substrate/client/offchain/src/lib.rs#L261

https://github.com/paritytech/polkadot-sdk/blob/cfc6cc65cc133f69f472149b2449a673597e8530/substrate/client/offchain/src/lib.rs#L286

However, since extensions are registered using `TypeId`, they would overwrite each other since they are all of the same type `Box<dyn Extension>`.

https://github.com/paritytech/polkadot-sdk/blob/cfc6cc65cc133f69f472149b2449a673597e8530/substrate/primitives/externalities/src/lib.rs#L336-L338

Moreover, I wouldn't be able to retrieve them later in host function using `self.extension::<MyExtension>()` since the registered TypeId is a generic `Box<dyn Extension>.type_id()`.

Calling `self.extension::<MyExtension>().expect("should be registered")` just gets panic.

### Steps to reproduce

_No response_

skunert:
Closes #1671 

As discovered correctly in that issue, custom externality extensions passed in from the outside as `Box<dyn Extension>` have some limitations. We store them in `Extensions` were the key is a `TypeId`, so multiple extensions are overriding each other.

I think the solution proposed in the issue is reasonable, so I changed the return type of the `custom_extensions` closure to return `Extensions`. They can then be passed to the runtime instance and merged with the existing extensions.


substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -365,6 +365,10 @@ fn generate_runtime_api_base_structures() -> Result<TokenStream> {
 				fn register_extension<E: #crate_::Extension>(&mut self, extension: E) {
 					std::cell::RefCell::borrow_mut(&self.extensions).register(extension);
 				}
+
+				fn register_extensions(&mut self, extension: #crate_::Extensions) {
```
skunert, line 369: The alternative would be to offer some `register_with_type_id` here.

davxy:
Just an idea, but what about adding a method to the `Extension` trait to return an identifier [here](https://github.com/skunert/polkadot-sdk/blob/bce7c2465cfa11cdf51f0160fafecc7ef02e5a75/substrate/primitives/externalities/src/extensions.rs#L38). (i.e. a method like `get_id() -> ExtId` with `ExtId` something  like `const [u8; N]`).

And then use that id to populate the map

https://github.com/skunert/polkadot-sdk/blob/bce7c2465cfa11cdf51f0160fafecc7ef02e5a75/substrate/primitives/externalities/src/extensions.rs#L161-L162

For sure using the `TypeId` is more easy to use and less prone to collision (if not in this case where using the `Box<dyn Extension>`. But if the trait implementors are careful enough creating their unique ids maybe this could be a viable solution

bkchr (changes requested):
We should probably remove the `Any` supertrait from `Extension`. Then we can forward the `type_id` function of `Box<dyn Extension` to the wrapped type. This would ensure that we don't take the `TypeId` of `Box<Extension>`. This should then probably fix this issue without requiring any of the changes in this pr.

substrate/primitives/externalities/src/extensions.rs
```diff
@@ -35,17 +35,24 @@ use sp_std::{
...
 /// As extensions are stored as `Box<Any>`, this trait should give more confidence that the correct
 /// type is registered and requested.
-pub trait Extension: Send + Any {
+pub trait Extension: Send + 'static {
 	/// Return the extension as `&mut dyn Any`.
 	///
 	/// This is a trick to make the trait type castable into an `Any`.
 	fn as_mut_any(&mut self) -> &mut dyn Any;
+
+	/// Get the [`std::any::TypeId`] of this `Extension`.
```
bkchr, line 44:
```suggestion
	/// Get the [`TypeId`] of this `Extension`.
```

bkchr (approved):

michalkucharczyk (approved):

davxy (approved):

PR #1720, opened 2023-09-26
Associated type Hasher for `QueryPreimage`, `StorePreimage` and `Bounded`

### LINKED ISSUE #1701: Hardcoded `type Hash = H256` in Preimages traits
QueryPreimage and StorePreimage traits use a hardcoded Hash type set to H256 https://github.com/paritytech/polkadot-sdk/blob/b18fb35e0dada9a23761f511d502b9a73fac9593/substrate/frame/support/src/traits/preimages.rs#L27 

Pallet Preimage itself is bounding to this trait I think the opposite way as it should be https://github.com/paritytech/polkadot-sdk/blob/b18fb35e0dada9a23761f511d502b9a73fac9593/substrate/frame/preimage/src/lib.rs#L534

As I was working on #1445 I incurred in this issue, and I was forced to do some weird things that include
```rust
let proposal_hash: PreimageHash = 
    Decode::decode(&mut &T::Hashing::hash_of(&proposal).encode()[..])
        .expect("Hash length is always 32; qed");
```
and opposite
```rust
if new_hash != PreimageHash::from_slice(hash.as_ref()) {
    // ...	
}
```

or having to use `BlakeTwo256` explicitly instead of `T::Hashing` in tests/benchmarks.

I think it can be fixed by removing the hardcoded type, and putting a `type Hash` in the traits, to be used like
```rust
impl<T: Config> QueryPreimage for Pallet<T> {
    type Hash = T::Hash;
    // ...
}
``` 
Should I proceed? If so, should I handle it in #1445 or is it better to open a new PR on which it will depend?

@kianenigma

muraca:
I hope it's enough to fix #1701 
the only solution I found to make it happen is to put an associated type to the `Bounded` enum as well.
@liamaharon @kianenigma @bkchr 

Polkadot address: 12poSUQPtcF1HUPQGY3zZu2P8emuW9YnsPduA4XG3oCEfJVp


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3804277

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3804787

substrate/frame/democracy/src/lib.rs
```diff
@@ -1610,7 +1614,7 @@ impl<T: Config> Pallet<T> {
 			// Earliest it can be scheduled for is next block.
 			let when = now.saturating_add(status.delay.max(One::one()));
 			if T::Scheduler::schedule_named(
-				(DEMOCRACY_ID, index).encode_into(),
+				(DEMOCRACY_ID, index).encode_into::<[u8; 32], T::Hashing>(),
```
liamaharon, line 1617:
```suggestion
				(DEMOCRACY_ID, index).encode_into::<_, T::Hashing>(),
```
can be inferred

liamaharon (approved):
This feels like a much more flexible, but also cleaner solution to me! Thanks!

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3811919

kianenigma (approved):
LGTM.

I wonder if there is any good documentation on your choice of `type Hash` and `type Hasher` in frame_system? Most runtimes use `BlakeTwo256` and H256 so this is a noop. Is it even clear where/how those hashers are used? are they meant to be used for everything in the runtime?

muraca:
I think the only documentation in frame_system is:
> /// The hashing system (algorithm) being used in the runtime (e.g. Blake2).
> type Hashing: Hash<Output = Self::Hash> + TypeInfo;

And I always assumed that it is the hasher you're supposed to use as default in the runtime

PR #1727, opened 2023-09-27
Add custom error message for `StorageNoopGuard`

### LINKED ISSUE #375: Custom error message for `StorageNoopGuard`
👉 Good issue for newcomers (please dont pick up when you already have experience).

There should be a constructor like `from_error_message` and/or a setter for a custom error message on the `StorageNoopGuard`.  
Also we should add a `new()` function which does the same as `default()` - seems to be a pattern in Rust.

seadanda:
Expand `StorageNoopGuard` to be able to add extra context through a custom error message. When the guard is triggered it panics with an error message which can be defaulted, set on construction, or set after it has been constructed.

Turn `StorageNoopGuard` into struct with `storage_root` and `error_message` and added `from_error_message` constructor and `set_error_message` setter.

Also added `new()` aliased to `default()`.

Closes #375

### DIFF
--- a/substrate/frame/support/src/storage/storage_noop_guard.rs
+++ b/substrate/frame/support/src/storage/storage_noop_guard.rs
@@ -52,6 +52,11 @@ impl Default for StorageNoopGuard {
 }
 
 impl StorageNoopGuard {
+	/// Alias to default()
+	pub fn new() -> Self {
+		Self::default()
+	}
+
 	/// Creates a new `StorageNoopGuard` with a custom error message
 	pub fn from_error_message(error_message: &'static str) -> Self {
 		Self { storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1), error_message }
@@ -148,6 +153,15 @@ mod tests {
 		TestExternalities::default().execute_with(|| {
 			let mut guard = StorageNoopGuard::default();
 			guard.set_error_message("StorageNoopGuard found unexpected storage changes.");
+			frame_support::storage::unhashed::put(b"key", b"value");
+		});
+	}
+
+	#[test]
+	#[should_panic(expected = "StorageNoopGuard detected wrongful storage changes.")]
+	fn storage_noop_guard_panics_new_alias() {
+		TestExternalities::default().execute_with(|| {
+			let _guard = StorageNoopGuard::new();
 			frame_support::storage::unhashed::put(b"key", b"value");
 		});
 	}
--- a/substrate/frame/support/src/storage/storage_noop_guard.rs
+++ b/substrate/frame/support/src/storage/storage_noop_guard.rs
@@ -37,11 +37,29 @@
 /// });
 /// ```
 #[must_use]
-pub struct StorageNoopGuard(sp_std::vec::Vec<u8>);
+pub struct StorageNoopGuard {
+	storage_root: sp_std::vec::Vec<u8>,
+	error_message: &'static str,
+}
 
 impl Default for StorageNoopGuard {
 	fn default() -> Self {
-		Self(sp_io::storage::root(sp_runtime::StateVersion::V1))
+		Self {
+			storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1),
+			error_message: "StorageNoopGuard detected wrongful storage changes.",
+		}
+	}
+}
+
+impl StorageNoopGuard {
+	/// Creates a new `StorageNoopGuard` with a custom error message
+	pub fn from_error_message(error_message: &'static str) -> Self {
+		Self { storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1), error_message }
+	}
+
+	/// Sets a custom error message for a `StorageNoopGuard`
+	pub fn set_error_message(&mut self, error_message: &'static str) {
+		self.error_message = error_message;
 	}
 }
 
@@ -53,8 +71,9 @@ impl Drop for StorageNoopGuard {
 		}
 		assert_eq!(
 			sp_io::storage::root(sp_runtime::StateVersion::V1),
-			self.0,
-			"StorageNoopGuard detected wrongful storage changes.",
+			self.storage_root,
+			"{}",
+			self.error_message,
 		);
 	}
 }
@@ -109,6 +128,27 @@ mod tests {
 			let _guard = StorageNoopGuard::default();
 			frame_support::storage::unhashed::put(b"key", b"value");
 			panic!("Something else");
+		});
+	}
+
+	#[test]
+	#[should_panic(expected = "StorageNoopGuard found unexpected storage changes.")]
+	fn storage_noop_guard_panics_created_from_error_message() {
+		TestExternalities::default().execute_with(|| {
+			let _guard = StorageNoopGuard::from_error_message(
+				"StorageNoopGuard found unexpected storage changes.",
+			);
+			frame_support::storage::unhashed::put(b"key", b"value");
+		});
+	}
+
+	#[test]
+	#[should_panic(expected = "StorageNoopGuard found unexpected storage changes.")]
+	fn storage_noop_guard_panics_with_set_error_message() {
+		TestExternalities::default().execute_with(|| {
+			let mut guard = StorageNoopGuard::default();
+			guard.set_error_message("StorageNoopGuard found unexpected storage changes.");
+			frame_support::storage::unhashed::put(b"key", b"value");
 		});
 	}
 }


substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,34 @@
...
+	}
+}
+
+impl StorageNoopGuard {
+	/// Alias to default()
+	pub fn new() -> Self {
+		Self::default()
+	}
+
+	/// Creates a new `StorageNoopGuard` with a custom error message
```
liamaharon, line 60:
```suggestion
	/// Creates a new [`StorageNoopGuard`] with a custom error message
```

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,34 @@
...
+	pub fn from_error_message(error_message: &'static str) -> Self {
+		Self { storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1), error_message }
+	}
+
+	/// Sets a custom error message for a `StorageNoopGuard`
```
liamaharon, line 65:
```suggestion
	/// Sets a custom error message for a [`StorageNoopGuard`]
```

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -111,4 +135,34 @@ mod tests {
...
+		TestExternalities::default().execute_with(|| {
+			let mut guard = StorageNoopGuard::default();
+			guard.set_error_message("StorageNoopGuard found unexpected storage changes.");
+			frame_support::storage::unhashed::put(b"key", b"value");
+		});
+	}
+
+	#[test]
+	#[should_panic(expected = "StorageNoopGuard detected wrongful storage changes.")]
+	fn storage_noop_guard_panics_new_alias() {
```
liamaharon, line 162: do you mean something like
```suggestion
	fn storage_noop_guard_panics_with_default_message() {
```

liamaharon (commented):
Looking good, left some comments.

Also, I think we should also add something like

```rust
impl From<&'static str> for StorageNoopGuard {
    fn from(s: &'static str) -> Self {
        StorageNoopGuard {
            storage_root: sp_std::vec::Vec::new(),
            error_message: s,
        }
    }
}
```

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,34 @@
...
+			error_message: "StorageNoopGuard detected wrongful storage changes.",
+		}
+	}
+}
+
+impl StorageNoopGuard {
+	/// Alias to default()
+	pub fn new() -> Self {
+		Self::default()
+	}
```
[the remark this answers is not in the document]
seadanda: This was requested by @ggwpez in the [issue](https://github.com/paritytech/polkadot-sdk/issues/375)

Happy to discuss this part, I just tagged it on to fully satisfy the issue.

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: It's just a test added in to prevent regressions in the behaviour of the new() function if anybody were to make changes. Basically a quick test to check that it equates to default without implementing PartialEq.
A bit contrived I agree.

seadanda:
> Looking good, left some comments.
> 
> Also, I think we should also add something like
> 
> ```rust
> impl From<&'static str> for StorageNoopGuard {
>     fn from(s: &'static str) -> Self {
>         StorageNoopGuard {
>             storage_root: sp_std::vec::Vec::new(),
>             error_message: s,
>         }
>     }
> }
> ```

Great, seems like a nice addition. I'll add it in.

bkchr (approved):

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: Fixed in 9c90e7b679

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: Fixed in 9c90e7b679

substrate/frame/support/src/storage/storage_noop_guard.rs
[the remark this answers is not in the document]
liamaharon: If it's a pattern in Rust, let's do it :) 

I just hadn't seen it before.

liamaharon (approved):
LGTM once `impl From<&'static str> for StorageNoopGuard` is added

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,40 @@
 /// });
 /// ```
 #[must_use]
-pub struct StorageNoopGuard(sp_std::vec::Vec<u8>);
+pub struct StorageNoopGuard {
+	storage_root: sp_std::vec::Vec<u8>,
+	error_message: &'static str,
```
ggwpez, line 42: Why does it need to be static? `String` would allow it to be used with dynamic messages as well.

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,40 @@
...
+		Self {
+			storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1),
+			error_message: "`StorageNoopGuard` detected an attempted storage change.",
+		}
+	}
+}
+
+impl From<&'static str> for StorageNoopGuard {
+	fn from(error_message: &'static str) -> Self {
+		StorageNoopGuard { storage_root: sp_std::vec::Vec::new(), error_message }
```
liamaharon, line 56: did you mean:
```suggestion
		StorageNoopGuard { storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1), error_message }
```

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,40 @@
...
-		Self(sp_io::storage::root(sp_runtime::StateVersion::V1))
+		Self {
+			storage_root: sp_io::storage::root(sp_runtime::StateVersion::V1),
+			error_message: "`StorageNoopGuard` detected an attempted storage change.",
+		}
+	}
+}
+
+impl From<&'static str> for StorageNoopGuard {
+	fn from(error_message: &'static str) -> Self {
```
ggwpez, line 55: Not sure about this, i think the `from_error_message` is enough.  
The `From` is more used in cases where the from type can be *converted*. This is not really a conversion but rather a construction, just to recap how it would look like:  
```rust
let _g: StorageNoopGuard = "Wrong changes".into();
```
is slightly weird to me.

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: I used the static lifetime because of how it is used in `Drop`. happy to switch to String though, I hadn't realised it was just for testing

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: Oops, brainfart. I've removed this in the end though after the comment from @ggwpez which changed how I thought about it.

liamaharon (approved):

substrate/frame/support/src/storage/storage_noop_guard.rs
liamaharon: Yeah you're totally right @ggwpez. Good lesson in thinking about tools for 'constructing' vs 'converting'. Thanks 👌

seadanda: Removed in 7bc5b6157c

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3817297

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3817299

substrate/frame/support/src/storage/storage_noop_guard.rs
```diff
@@ -37,11 +37,34 @@
 /// });
 /// ```
 #[must_use]
-pub struct StorageNoopGuard(sp_std::vec::Vec<u8>);
+pub struct StorageNoopGuard {
+	storage_root: sp_std::vec::Vec<u8>,
+	error_message: String,
```
ggwpez, line 42: Ah damn, looks like this does not work since we dont have strings in the benchmarking no-std context 🤦   
Sorry, then it needs to be `&'static str`, or `&'a str` with the lifetime of the guard.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3817461

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3817460

substrate/frame/support/src/storage/storage_noop_guard.rs
seadanda: haha OK I've reverted that commit and switched to the lifetime of the guard, I now see that since the `Drop` `impl` block has the same lifetime I didn't need static in the first place - I was thinking about it in the wrong way.

Issue #1638, opened 2023-09-19
`AllowTopLevelPaidExecutionFrom` barrier prevents multi currency transfers which also effects `assethub`.


rimbi:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

After [this change](https://github.com/paritytech/polkadot-sdk/commit/26be0727e5aaa5ed569b29bd9a6b043899a5617b), `AllowTopLevelPaidExecutionFrom` prevents multi currency transfers which also effects `assethub`; multi currency transfers are not possible anymore.

Here, also, is [a test case in the orml repo](https://github.com/moonbeam-foundation/open-runtime-module-library/blob/6819d3f28f9005c0a375e359c942490e04e20000/xtokens/src/tests.rs#L393) which failed after upgrading dependencies to `polkadot-v1.1.0` (which includes this change) and  [this is the solution](https://github.com/moonbeam-foundation/open-runtime-module-library/blob/9998244c0f6595c442084de553b9a6a45e18ec1d/xtokens/src/mock/mod.rs#L351) for the problem, which is the exact copy of the old version of the barrier.

librelois:
What was the motivation for limiting the `AllowTopLevelPaidExecutionFrom` barrier, is there a security risk in having multiple assets in the holding register at the BuyExecution step?

xlc:
To my understanding, it is indeed a security risk to allow unbounded asset withdrawal. But limit it to 1 is just too much, especially when the number is non configurable.

bkchr:
@franciscoaguirre @KiChjang please look into this ASAP.

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/polkadot-release-analysis-v1-3-0/4614/1

acatangiu:
This was fixed

PR #1733, opened 2023-09-27
increase MAX_ASSETS_FOR_BUY_EXECUTION

xlc:
Partially addresses #1638

Still need a better solution to allow devs to have better control of this.


polkadot/xcm/xcm-builder/src/barriers.rs
```diff
@@ -52,7 +52,7 @@ impl ShouldExecute for TakeWeightCredit {
 	}
 }
 
-const MAX_ASSETS_FOR_BUY_EXECUTION: usize = 1;
+const MAX_ASSETS_FOR_BUY_EXECUTION: usize = 5;
```
bkontur, line 55: just wonder (maybe I am missing something), if it is good idea to expose and make it runtime-configurable like:
```
pub struct AllowTopLevelPaidExecutionFrom<T, MaxAssetsForBuyExecution>(PhantomData<(T, MaxAssetsForBuyExecution)>);

impl<
    T: Contains<MultiLocation>,
    MaxAssetsForBuyExecution: Get<u8>
> ShouldExecute for AllowTopLevelPaidExecutionFrom<T> {
```

acatangiu (approved):

albertov19:
@KiChjang any comments here? This is needed for Foreign Assets in AssetHub for example. That is, if we want to register GLMR in AssetHub, we need to be able to pay with DOT XCM execution, or I'm missing something?

polkadot/xcm/xcm-builder/src/barriers.rs
kianenigma: Let's already do this if this needs to be 5 only in asset-hub?

xlc: well, the update should really be universal. all the chains that supports multiple assets needs this.

joepetrowski:
This might be needed for backwards compatibility with some programs existing on parachains, but I think the name of the variable makes the name clear: Max assets **for Buy Execution**. It is _not_ max assets in holding.

As the Barrier changes show [where it was added](https://github.com/paritytech/polkadot-sdk/commit/26be0727e5aaa5ed569b29bd9a6b043899a5617b), it's looking for the asset withdrawal before `BuyExecution`. I think the correct way to address the linked issue is to use the following pattern when you want to deposit more than one asset:

```
WithdrawAsset { /* 1 asset for fees */ }
BuyExecution { .. }
WithdrawAsset { /* other assets to use in the program */ }
...
```

polkadot/xcm/xcm-builder/src/barriers.rs
joepetrowski: No, not that support multiple asset. That support _buying execution_ in multiple assets.

Asset Hub does not support this.

joepetrowski: You can see how this works now, using a [Take First Asset Trader](https://github.com/paritytech/polkadot-sdk/blob/bdbe982/cumulus/primitives/utility/src/lib.rs#L120). It (conveniently) allows you to put `MAX_ASSETS` into holding, but only takes the first one for `BuyExecution`. So this is still the same behavior; it allows 1 asset for `BuyExecution`.

Again, I know this PR might solve the short-term problem for existing use cases, but I think the intent of this `const` is clear and that a chain supporting multiple assets does not mean it supports multiple assets simultaneously for weight payment.

KiChjang:
```suggestion
const MAX_ASSETS_FOR_BUY_EXECUTION: usize = 2;
```

franciscoaguirre (approved):

bkchr (approved):

KiChjang (approved):

acatangiu:
`continuous-integration/gitlab-check-runtime-migration-rococo` started failing for all PRs - unrelated to this

bkchr:
> `continuous-integration/gitlab-check-runtime-migration-rococo` started failing for all PRs - unrelated to this

Is this already tracked? Looks like a migration is failing.

PR #1650, opened 2023-09-20
Move requests-responses and polling from `ChainSync` to `SyncingEngine`

dmitry-markin:
Move request-response handling from `ChainSync` to `SyncingEngine` as part of [Sync 2.0](https://github.com/paritytech/polkadot-sdk/issues/534) refactoring aimed at making `ChainSync` a pure state machine.

Resolves https://github.com/paritytech/polkadot-sdk/issues/502.

PR #1735, opened 2023-09-28
Wake the task up when inserting a new response into `PendingResponses`

dmitry-markin:
This is a hot fix for https://github.com/paritytech/polkadot-sdk/pull/1650. Basically, `StreamMap` suffers from the same footgun as `FuturesUnordered`: it must be explicitly polled in order for a newly added stream to be registered in the task. The streams are just pushed into the internal container upon insertion: https://github.com/tokio-rs/tokio/blob/master/tokio-stream/src/stream_map.rs#L445-L453

This PR adds a `Waker` to `PendingResponses` to make sure the task is woken up when a new response is added.

### DIFF
--- a/substrate/client/network/sync/src/pending_responses.rs
+++ b/substrate/client/network/sync/src/pending_responses.rs
@@ -17,7 +17,7 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 //! [`PendingResponses`] is responsible for keeping track of pending responses and
-//! polling them.
+//! polling them. The [`Stream`] implemented by [`PendingResponses`] never terminates.
 
 use futures::{
 	channel::oneshot,
@@ -50,6 +50,7 @@ pub(crate) struct ResponseEvent<B: BlockT> {
 pub(crate) struct PendingResponses<B: BlockT> {
 	/// Pending responses
 	pending_responses: StreamMap<PeerId, BoxStream<'static, (PeerRequest<B>, ResponseResult)>>,
+	waker: Option<Waker>,
 }
 
 impl<B: BlockT> PendingResponses<B> {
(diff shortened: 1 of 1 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3827486

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3827638

altonen (approved):
I think in general it would be good to test all non-trivial networking changes on a live node since the test coverage of `sc-network` has proven to be insufficient

substrate/client/network/sync/src/pending_responses.rs
```diff
@@ -79,6 +80,10 @@ impl<B: BlockT> PendingResponses<B> {
 			);
 			debug_assert!(false);
 		}
+
+		if let Some(waker) = self.waker.take() {
+			waker.wake();
```
bkchr, line 85: This requires that we called `poll` before. Not sure that this is removing any footgun.

bkchr: Looking at the code we are already polling `pending_responses` after we inserted any kind of element, so why is the task not waking up?

dmitry-markin: > This requires that we called `poll` before. Not sure that this is removing any footgun.

Yeah, but we definitely called it at least once.

dmitry-markin: > Looking at the code we are already polling `pending_responses` after we inserted any kind of element, so why is the task not waking up?

There is no problem with waking up right now, but it can shoot us once the polling in `SyncingEngine` is converted into `select!` and the order of inserting / polling `pending_responses` is not strictly defined.

bkchr: But for a `select!`. Once you insert data to the `pending_responses` your match arm will be finished and you start again from the top of the loop, aka calling `select!` again. This will poll all futures, including your `pending_responses` futures. So, not sure why we could miss there to poll it.

PR #1736, opened 2023-09-28
Move import queue from `ChainSync` to `SyncingEngine`

dmitry-markin:
This PR is part of [Sync 2.0](https://github.com/paritytech/polkadot-sdk/issues/534) refactoring aimed at making `ChainSync` a pure state machine.

Resolves https://github.com/paritytech/polkadot-sdk/issues/501.

### DIFF
--- a/substrate/client/network/common/src/sync.rs
+++ b/substrate/client/network/common/src/sync.rs
@@ -316,6 +316,7 @@ pub trait ChainSync<Block: BlockT>: Send {
 	/// Handle a new connected peer.
 	///
 	/// Call this method whenever we connect to a new peer.
+	#[must_use]
 	fn new_peer(
 		&mut self,
 		who: PeerId,
@@ -347,6 +348,7 @@ pub trait ChainSync<Block: BlockT>: Send {
 	///
 	/// If this corresponds to a valid block, this outputs the block that
 	/// must be imported in the import queue.
+	#[must_use]
 	fn on_block_data(
 		&mut self,
 		who: &PeerId,
@@ -357,6 +359,7 @@ pub trait ChainSync<Block: BlockT>: Send {
 	/// Handle a response from the remote to a justification request that we made.
 	///
 	/// `request` must be the original request that triggered `response`.
+	#[must_use]
 	fn on_block_justification(
 		&mut self,
 		who: PeerId,
--- a/substrate/client/network/common/src/sync.rs
+++ b/substrate/client/network/common/src/sync.rs
@@ -116,11 +116,18 @@ impl fmt::Display for BadPeer {
 
 impl std::error::Error for BadPeer {}
 
+/// Action that the parent of [`ChainSync`] should perform if we want to import blocks.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ImportBlocksAction<B: BlockT> {
+	pub origin: BlockOrigin,
+	pub blocks: Vec<IncomingBlock<B>>,
+}
+
 /// Result of [`ChainSync::on_block_data`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum OnBlockData<Block: BlockT> {
 	/// The block should be imported.
-	Import(BlockOrigin, Vec<IncomingBlock<Block>>),
+	Import(ImportBlocksAction<Block>),
 	/// A new block request needs to be made to the given peer.
 	Request(PeerId, BlockRequest<Block>),
 	/// Continue processing events.
@@ -134,7 +141,7 @@ pub enum OnBlockJustification<Block: BlockT> {
 	Nothing,
 	/// The justification should be imported.
 	Import {
-		peer: PeerId,
+		peer_id: PeerId,
 		hash: Block::Hash,
 		number: NumberFor<Block>,
 		justifications: Justifications,
@@ -379,7 +386,8 @@ pub trait ChainSync<Block: BlockT>: Send {
 	/// Call when a peer has disconnected.
 	/// Canceled obsolete block request may result in some blocks being ready for
 	/// import, so this functions checks for such blocks and returns them.
-	fn peer_disconnected(&mut self, who: &PeerId);
+	#[must_use]
+	fn peer_disconnected(&mut self, who: &PeerId) -> Option<ImportBlocksAction<Block>>;
 
 	/// Return some key metrics.
 	fn metrics(&self) -> Metrics;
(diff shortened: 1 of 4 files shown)


substrate/client/network/common/src/sync.rs
```diff
@@ -379,7 +386,8 @@ pub trait ChainSync<Block: BlockT>: Send {
 	/// Call when a peer has disconnected.
 	/// Canceled obsolete block request may result in some blocks being ready for
 	/// import, so this functions checks for such blocks and returns them.
-	fn peer_disconnected(&mut self, who: &PeerId);
+	#[must_use]
```
altonen, line 389: Why only for this function?

substrate/client/network/sync/src/lib.rs
```diff
@@ -200,45 +198,42 @@ impl Default for AllowedRequests {
...
 	best_queued_number: NumberFor<B>,
 	target: NumberFor<B>,
 }
 
-/// An event used to notify [`engine::SyncingEngine`] if we want to perform a block request
-/// or drop an obsolete pending response.
-enum BlockRequestEvent<B: BlockT> {
+/// Action that [`engine::SyncingEngine`] should perform after reporting imported blocks with
+/// [`ChainSync::on_blocks_processed`].
+enum BlockRequestAction<B: BlockT> {
```
altonen, line 209: Ideally all these block/import/response actions would be under one `SyncAction` enum. `SyncingEngine` calls `ChainSync`, `ChainSync` pushes the actions it wants `SyncingEngine` to perform to its internal action queue and after each call to `ChainSync`, `SyncingEngine` would do `while let Some(action) = self.chain_sync.next_action() { match action ... }` until all actions are consumed. We'd get rid of the `OnBlockResponse::Nothing` for free.

Not in this PR though but a future task

substrate/client/network/sync/src/lib.rs
```diff
@@ -1107,11 +1099,8 @@ where
 		});
 
 		let blocks = self.ready_blocks();
-		if let Some(OnBlockData::Import(origin, blocks)) =
-			(!blocks.is_empty()).then(|| self.validate_and_queue_blocks(blocks, false))
-		{
-			self.import_blocks(origin, blocks);
-		}
+
+		(!blocks.is_empty()).then(|| self.validate_and_queue_blocks(blocks, false))
```
altonen, line 1103: `let blocks = self.ready_blocks();` is kind of useless now and you could directly use it instead of creating `blocks`.

substrate/client/network/sync/src/lib.rs
```diff
@@ -1521,44 +1496,53 @@ where
...
-				Ok(OnBlockData::Request(peer, req)) => Some((peer, req)),
-				Ok(OnBlockData::Continue) => None,
+				Ok(OnBlockData::Import(action)) => OnBlockResponse::ImportBlocks(action),
+				Ok(OnBlockData::Request(peer_id, request)) =>
+					OnBlockResponse::SendBlockRequest { peer_id, request },
+				Ok(OnBlockData::Continue) => OnBlockResponse::Nothing,
 				Err(BadPeer(id, repu)) => {
 					self.network_service
 						.disconnect_peer(id, self.block_announce_protocol_name.clone());
 					self.network_service.report_peer(id, repu);
```
altonen, line 1523: Needs a deeper analysis but I'm not sure if `ChainSync` requires a handle to `NetworkService` anymore. Disconnecting & reporting a peer could be another action.

We could probably get rid of the `NetworkServiceProvider` hack if the handle to `NetworkService` was only passed to `SyncingEngine::run()`.

altonen (approved):

substrate/client/network/common/src/sync.rs
dmitry-markin: Done, added for other functions where appropriate.

substrate/client/network/sync/src/lib.rs
dmitry-markin: I kept it because there is a check `!blocks.is_empty()`, and there is no `map()` for `Vec`. Am I missing something?

altonen: What I meant is that why can't the code be:

```rust
(!self.blocks.is_empty()).then(|| self.validate_and_queue_blocks(blocks, false))
```

dmitry-markin: I don't understand. We need to both check that `ready_blocks()` are not empty and also pass `ready_blocks()` to `validate_and_queue_blocks()`. How can we do that without an intermediate variable?

altonen: Sorry I was being a dumbass, I didn't notice the parameter. Nevermind

substrate/client/network/sync/src/lib.rs
dmitry-markin: Makes sense, but may be returning actions from every call to `ChainSync` instead fetching them from `next_action()`? Otherwise we'll need to be extremely careful to call `next_action()` after every other call to `ChainSync`.

dmitry-markin:
@arkpar I've assigned you as a reviewer for this PR because you made some changes to syncing in the past. Let me know if you are not the right person to ask for review of `ChainSync` now.

substrate/client/network/sync/src/lib.rs
altonen: I wouldn't go with returning an action from every function. Some calls to `ChainSync` could result in more than one action so we'd have to return an iterator of actions from every call or make exceptions that some calls return an iterator where others return an `Option<Action>` and then handle those differently on the call site. Doesn't seem too elegant.

If `ChainSync` is only called when there's an I/O-related event, the handling of actions can be a simple loop after `select!`

substrate/client/network/sync/src/lib.rs
arkpar: > Disconnecting & reporting a peer could be another action.

It's probably fine as long as these actions will be handled synchornously. Just want to make sure we won't end up with reputation changes being pushed to an async queue again. We had problems with that in the past.

arkpar (approved):

substrate/client/network/sync/src/lib.rs
dmitry-markin: In fact, right now they are pushed to an async queue: https://github.com/paritytech/polkadot-sdk/blob/0691c91e1558f235081475bfd60ce71035f272a5/substrate/client/network/sync/src/service/network.rs#L83

But luckily after removing `NetworkServiceProvider` we will be able to update reputations synchronously, because `PeerStore` (formerly `PeerSet`) has a sync implementation now: https://github.com/paritytech/polkadot-sdk/blob/0691c91e1558f235081475bfd60ce71035f272a5/substrate/client/network/src/peer_store.rs#L96

PR #1742, opened 2023-09-28
Publish `substrate-cli-test-utils` to crates.io

liamaharon:
Removes `publish = false` from `substrate-cli-test-utils`.

### DIFF
--- a/substrate/test-utils/Cargo.toml
+++ b/substrate/test-utils/Cargo.toml
@@ -7,6 +7,7 @@ license = "Apache-2.0"
 homepage = "https://substrate.io"
 repository.workspace = true
 description = "Substrate test utilities"
+publish = false
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
--- a/substrate/test-utils/Cargo.toml
+++ b/substrate/test-utils/Cargo.toml
@@ -7,7 +7,6 @@ license = "Apache-2.0"
 homepage = "https://substrate.io"
 repository.workspace = true
 description = "Substrate test utilities"
-publish = false
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
--- a/substrate/test-utils/cli/Cargo.toml
+++ b/substrate/test-utils/cli/Cargo.toml
@@ -7,7 +7,6 @@ edition.workspace = true
 license = "Apache-2.0"
 homepage = "https://substrate.io"
 repository.workspace = true
-publish = false
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]


bkchr (approved):

bkchr (changes requested):
Wait. This crate only contains two macros, do we really need to publish this? You can just copy these macros.

Morganamilo:
I don't see any reason to not publish anything at all really. If its useful for us it could be useful for some one else.

liamaharon:
> Wait. This crate only contains two macros, do we really need to publish this? You can just copy these macros.

The root `lib.rs` contains 2 macros, but it also exports the modules in directories like `polkadot-sdk/substrate/test-utils/cli/src/lib.rs` which has much more functionality.

liamaharon:
I've updated this PR to also rename `node-executor` to `polkadot-node-executor` (original name is taken) and set public to true on that crate too.

liamaharon:
> Wait. This crate only contains two macros, do we really need to publish this? You can just copy these macros.

I was actually mistaken here... I meant to publish `substrate-cli-test-utils`, not `substrate-test-utils. Updated the PR.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3838382

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3838510

substrate/bin/node/executor/Cargo.toml
```diff
@@ -1,13 +1,12 @@
 [package]
-name = "node-executor"
+name = "polkadot-node-executor"
```
bkchr, line 2: Don't do this. This crate should not be released!

bkchr (changes requested):
Please revert your changes around releasing the `node-cli` stuff and the renaming etc!

In the test-utils you need to remove the `start_node_inline` function to not require these deps released. Next point would be `start_node` is expecting `substrate-node`. Either make the binary name an input to `start_node` or also remove this function.

substrate/bin/node/executor/Cargo.toml
liamaharon: Why not release it? `try-runtime-cli` currently uses the `ExecutorDispatch` struct. 

Maybe we should just move it back into the monorepo (as a seperate binary) like you suggested earlier..

liamaharon: Maybe we should move it into the monorepo as you suggested to avoid these headaches.. But it still seems that it is reasonable some projects may want to use these exported items.

bkchr: `sc-executor` != `node-executor`. Node executor being the executor declaration for the Substrate node that is nothing that should be released.

bkchr: `sc-executor` is already released.

liamaharon: Ah.. So it turns out `ExecutorDispatch` was being imported from `node-executor`. `ExecutorDispatch` is also available from `sc-executor`, so we can remove `node-executor` as a dependency. I then got `sc-executor` and `node-executor` confused, leading to my confused msg above.

Thank you for flagging this @bkchr. I'll revert my previous commit here and remove `node-executor` as a dependency from `try-runtime-cli.

bkchr:
> Please revert your changes around releasing the `node-cli` stuff and the renaming etc!
> 
> In the test-utils you need to remove the `start_node_inline` function to not require these deps released. Next point would be `start_node` is expecting `substrate-node`. Either make the binary name an input to `start_node` or also remove this function.

@liamaharon this still holds. You need to remove `node-cli` and `node-primitives`.

liamaharon:
@bkchr what is the best practice then for starting a node programatically?  

There are valid use cases for a programatic API for starting nodes.

bkchr:
It makes sense to run a node in a test. I'm not arguing against this. You can also use the current functions, but then you should make the binary name generic, by passing to the respective functions.

bkchr:
@liamaharon what is the status of this?

liamaharon:
> @liamaharon what is the status of this?

I need to investigate the best alternative to using the dep in try-runtime-cli. 

Will close the PR for now.

Issue #877, opened 2022-01-21
Parachain Slot Extension Story


shawntabrizi:
As the first set of parachains are going to end their lease period, we need to make sure there is a well designed story around extending the lease period of existing parachains such that there is no downtime.

There are a bunch of different ideas an approaches to this story, so lets list them all below.

## Extending a Crowdloan with Another Crowdloan

All teams that have won a parachain slot so far has done so via crowdloans.

The intention of crowdloans is to be a one-off bootstrapping mechanism, not a permanent repeatable process.

To run a second crowdloan is also a bit dubious since the main incentive for users to participate in a crowdloan is to get tokens from the parachain. Usually a significant amount of the token supply is reserved for this, like 20 - 30%. In the case of a second crowdloan, will chains just mint another 20% more tokens?

That being said, some teams have declared interest in this, and specifically allowing crowdloan contributors access to their previously contributed funds so they can use it in a second crowdloan, without having any parachain downtime.

The current idea is to release the funds of crowdloan contributors a few weeks ahead of the end of a parachain slot, and thus the users can use those funds for whatever they see fit.

To enable this, we must write some new logic related to when a crowdloan can be dissolved, specifically allowing it to happen even though a parachain will have a slot won by a crowdloan.

### Concerns

None as of yet.

## Allowing a Parachain to Bid on an Overlapping Lease Period

Teams have also been interested in directly bidding for a new slot by winning a full set of lease periods that may overlap a lease period already won by that parachain.

At the moment we have this check which prevents this:

```
// We also make sure that the bid is not for any existing leases the para already has.
ensure!(
	!T::Leaser::already_leased(para, first_slot, last_slot),
	Error::<T>::AlreadyLeasedOut
);
```

### Concerns

The reason we did not allow this was specifically to prevent griefing by a whale parachain from winning auctions from other people. Even in the Slots pallet, we check for this, and bail from giving the parachain any slots at all:

```
// The chain tried to lease the same period twice. This might be a griefing
// attempt.
//
// We bail, not giving any lease and leave it for governance to sort out.
return Err(LeaseError::AlreadyLeased)
```

This is a concern specifically because we allow a parachain to bid with funds already used in a previous bidding process, exactly for preventing parachain downtime. So if we removed this check, a parachain could win slots 1 - 8, then for every auction for the rest of Polkadot, this parachain could bid in that auction with all of those same funds, basically making it so that only one parachain is onbarded onto Polkadot.

[...]

PR #1749, opened 2023-09-29
Allow early refund of slot deposit

Ank4n:
closes https://github.com/paritytech/polkadot-sdk/issues/877
context: https://hackmd.io/4OLfULI6RtaPdE-L_hur4g

## Changes since last approach (Updated on 19th Nov):
Instead of having a config value that determines how early refund can be applied before the slot expiry, now early refund can only be applied anytime during the last LP of the lease. Parachains should have at minimum `MinLeasePeriodsForEarlyRefund` LPs in their current lease to be able to refund early.

## Summary
This change would allow a parachain to refund their slot deposit during their last lease period, given that they leased out minimum of `MinLeasePeriodsForEarlyRefund` in the last auction. For parachains who won the lease with a crowdloan, they could then refund crowdloan funds to contributors, and dissolve the crowdloan. This gives them opportunity to either start a new crowdloan or bid directly with their released funds.

If `MinLeasePeriodsForEarlyRefund` is set to `0`, a slot deposit can never be refunded before it expires.

## Changes
- Config const `MinLeasePeriodsForEarlyRefund`: Determines how many minimum lease periods a leaser in their latest lease to be eligible for early refund. This value should always be greater than 0. If set to 0, `early_lease_refund` is disabled.
- New signed extrinsic `early_lease_refund`: Only para or para manager can call this or can be executed via governance with Lease Admin origin track. It allows a parachain to unreserve their lease deposit during the last lease period of their full lease. This will unreserve the deposit of the slot but would not downgrade the parachain yet. Parachain would be downgraded as it happens today if at the start of the next lease period it does not have any new leases remaining or left alone if a new lease has become available and update the fund reserved to match the new requirement. 
- Storage double map `LeaseInfo`: Keeps track of how much funds is reserved  as well as period count of the latest lease for a leaser.

`early_lease_refund` can be followed by the calls `refund`  and `dissolve` on the crowdloan pallet for the para_id which will remove the crowdloan and allow the parachain to create a new crowdloan for the same para_id.

## TODO
- [x] All existing tests pass in the scenario where `early_lease_refund` is not called. This ensures we do not impact any existing flows.
- [x] Unit + Integration test 
- [x] Allow only Lease Admin or parachain manager to call `early_lease_refund`.
- [x] Migration of LeaseInfo.
- [x] Bench
- [x] Only leasers who have more than one lease can early refund slot deposit.
- [x] Intialise lease count of existing parachains to 4 (we need it to be greater than 2 to enable `early_lease_refund`).
- [x] Allow para manager or para origin to call `early_lease_refund`.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3842261

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -267,11 +312,17 @@ impl<T: Config> Pallet<T> {
...
 					// Then we need to get the new amount that should continue to be held on
 					// deposit for the parachain.
-					let now_held = Self::deposit_held(para, &ended_lease.0);
+					let required_hold = Self::deposit_held(para, &ended_lease.0);
 
 					// If this is less than what we were holding for this leaser's now-ended lease,
 					// then unreserve it.
-					if let Some(rebate) = ended_lease.1.checked_sub(&now_held) {
+					let current_hold = ended_lease.1;
+					// Fixme(ank4n): check reserved balance from state instead of deposit. We might
```
Ank4n, line 320: Fungibles would fix this but probably shouldn't migrate in this PR

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3849007

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Reworking on migrating slots pallet to fungibles [here #1794](https://github.com/paritytech/polkadot-sdk/pull/1794)

kianenigma:
@Ank4n a fellowship release is finally out, and I suspect we will be on faster roll going forward. Let's make sure this is in a ready state to make it to an upcoming release?

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -221,6 +241,46 @@ pub mod pallet {
 			};
 			Ok(())
 		}
+
+		/// Try to refund the lease deposit before the actual end of the lease.
+		///
+		/// This is only allowed if lease is ending within `T:EarliestRefundPeriod` and the
```
kianenigma, line 247:
```suggestion
		/// This is only allowed if lease is ending within [`Config::EarliestRefundPeriod`] and the
```

Ank4n:
> @Ank4n a fellowship release is finally out, and I suspect we will be on faster roll going forward. Let's make sure this is in a ready state to make it to an upcoming release?

Yes, will aim to get into ready state by next couple of weeks. This would need an audit as well. 

Happy that you did a review. I initially thought this would be much less code change but since the Reserved amounts are not named, a new map needs to be added to track the current reserve. The other alternative is to migrate slots pallet to Fungible and then make the change for the `early_lease_refund`. Code wise probably without fungibles is still less changes but may be this is a good chance to migrate to fungibles as well. Let me know if you have any opinion on this?

Ank4n:
> @Ank4n a fellowship release is finally out, and I suspect we will be on faster roll going forward. Let's make sure this is in a ready state to make it to an upcoming release?

Thanks for the review. Yes, will aim to get into ready state in the coming weeks. This would need an audit as well.

I initially thought this would be much less code change but since the Reserved amounts are not named, a new map needs to be added to track the current reserve. The other alternative is to migrate slots pallet to Fungible and then make the change for the `early_lease_refund`. Code wise probably without fungibles is still less changes but may be this is a good chance to migrate to fungibles as well. Let me know if you have any opinion on this?

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -221,6 +241,52 @@ pub mod pallet {
...
+		/// Can only be called by the parachain manager.
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::WeightInfo::early_lease_refund())]
+		pub fn early_lease_refund(origin: OriginFor<T>, para: ParaId) -> DispatchResult {
+			// ensure caller is ForceOrigin or the manager of the parachain.
+			ensure_signed(origin.clone())
+				.map_err(|e| e.into())
+				.and_then(|who| -> DispatchResult {
+					ensure!(Some(who) == T::Registrar::manager_of(para), Error::<T>::NoPermission);
+					Ok(())
```
SBalaguer, line 261: This allows the manager of the Parachain on the Relay Chain to call the function, right? If so, should we instead make it the parachain itself? They could make a call from the parachain via UMP, as it happens today with `swap`.

Ank4n: Yes this will allow parachain manager to call it (or ForceOrigin which can be root or lease admin). Do you want to allow  all three: root, para and para_manager?

Ank4n:
bot bench polkadot-pallet --pallet=slots

Ank4n:
bot bench polkadot-pallet --pallet=slots --runtime westend

Ank4n:
bot bench-all polkadot --runtime=westend

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -1698,3 +1703,136 @@ fn cant_bid_on_existing_lease_periods() {
...
+
+		// User 1 will own para
+		Balances::make_free_balance_be(&account_id(1), 1_000_000_000);
+
+		// Register on-demand parachain
+		let validation_code = test_validation_code(10);
+		assert_ok!(Registrar::reserve(signed(1)));
+		assert_ok!(Registrar::register(
+			signed(1),
+			ParaId::from(2000),
```
liamaharon, line 1724: nit: this magic number is repeated a lot

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -1698,3 +1703,136 @@ fn cant_bid_on_existing_lease_periods() {
...
+		// Withdraw of contributions works
+		assert_eq!(Balances::free_balance(&crowdloan_account), total);
+		for i in 10..20 {
+			assert_ok!(Crowdloan::withdraw(signed(i), account_id(i), ParaId::from(2000)));
+		}
+
+		assert_eq!(Balances::free_balance(&crowdloan_account), 0);
+
+		// reserved balance before dissolve
+		assert_eq!(Balances::reserved_balance(&account_id(1)), (500 + 10 * 2 * 1) + 100);
```
liamaharon, line 1816: nit: magic numbers

polkadot/runtime/common/src/slots/migration.rs
```diff
@@ -16,8 +16,94 @@
...
+	pub type ToV1<T> = frame_support::migrations::VersionedMigration<
+		0,
+		1,
+		v1::MigrateToV1<T>,
+		Pallet<T>,
+		<T as frame_system::Config>::DbWeight,
+	>;
+}
+
+mod v1 {
```
liamaharon, line 41: Cool, happy to see usage of this pattern :)

polkadot/runtime/common/src/slots/migration.rs
```diff
@@ -16,8 +16,94 @@
...
+	use super::*;
+
+	pub struct MigrateToV1<T>(sp_std::marker::PhantomData<T>);
+
+	impl<T: Config> OnRuntimeUpgrade for MigrateToV1<T> {
+		fn on_runtime_upgrade() -> Weight {
+			let mut weight = Weight::zero();
+			for (para, lease_periods) in Leases::<T>::iter() {
+				weight = weight.saturating_add(T::DbWeight::get().reads(1));
```
liamaharon, line 50: could use `saturating_accrue`

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -142,6 +158,11 @@ pub mod pallet {
 		ParaNotOnboarding,
 		/// There was an error with the lease.
 		LeaseError,
+		/// The signing account has no permission to do the operation.
+		NoPermission,
+		/// The signing account has privileges for this operation but the preconditions are not
+		/// met.
+		NotAllowed,
```
liamaharon, line 165: maybe more descriptive: 
```suggestion
		PreconditionNotMet,
```

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -192,11 +213,10 @@ pub mod pallet {
 
 			// Refund any deposits for these leases
 			for (who, deposit) in deposits {
-				let err_amount = T::Currency::unreserve(&who, deposit);
-				debug_assert!(err_amount.is_zero());
```
liamaharon, line 196: Why remove the debug_assert?

polkadot/runtime/common/src/slots/mod.rs
SBalaguer: I'm ok either way! Usually the way I've seen this in the past is that the para_manager is set to do more super-admin like stuff, when a parachain is unlocked (like changing wasm/head), however on normal operations the goal is to use the parachain itself to send the UMP. But happy to learn from other opinions here too.

polkadot/runtime/common/src/integration_tests.rs
Ank4n: Good point I will make it more descriptive

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Yeah this is better 👍

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Its not removed but moved inside the `fn unreserve` as defensive_assert.

Ank4n:
bot bench polkadot-pallet --pallet=runtime_common_slots --runtime rococo

polkadot/runtime/common/src/slots/migration.rs
```diff
@@ -16,8 +16,93 @@
...
+}
+
+mod v1 {
+	use super::*;
+
+	pub struct MigrateToV1<T>(sp_std::marker::PhantomData<T>);
+
+	impl<T: Config> OnRuntimeUpgrade for MigrateToV1<T> {
+		fn on_runtime_upgrade() -> Weight {
+			let mut weight = Weight::zero();
```
kianenigma, line 48: https://paritytech.github.io/polkadot-sdk/master/frame_support/weights/struct.WeightMeter.html

Ank4n:
bot bench-all polkadot --runtime=rococo

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -310,27 +384,100 @@ impl<T: Config> Pallet<T> {
...
-			None => {},
+				None => Zero::zero(),
+			})
+			.max()
+			.unwrap_or_else(Zero::zero)
+	}
+
+	/// Returns true if the current lease period is ending within the next `T::EarliestRefundPeriod`
+	/// blocks.
+	fn lease_period_ending_soon(now: BlockNumberFor<T>) -> Option<bool> {
```
kianenigma, line 411: could just return false if it is `None`?

kianenigma (approved):
Solution seems good to me in general. I suggest some community/research approval on the final parameter that goes into Polkadot runtime. Approving as it is a noop PR if we set the new param to 0.

polkadot/runtime/rococo/src/weights/runtime_common_slots.rs
```diff
@@ -129,4 +129,8 @@ impl<T: frame_system::Config> runtime_common::slots::WeightInfo for WeightInfo<T
 			.saturating_add(T::DbWeight::get().reads(5))
 			.saturating_add(T::DbWeight::get().writes(3))
 	}
+
+	fn early_lease_refund() -> Weight {
+		todo!("run bench")
```
liamaharon, line 134: 👀

liamaharon (approved):

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -310,27 +384,100 @@ impl<T: Config> Pallet<T> {
...
+		let (maybe_next_lease, _) = Self::lease_period_index(soon)?;
+		Some(maybe_next_lease == current_lease.checked_add(&sp_runtime::traits::One::one())?)
+	}
+
+	/// Reserve the amount for the parachain, updating the current reserved amount for this leaser
+	/// and parachain.
+	fn reserve(para: ParaId, leaser: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+		ReservedAmounts::<T>::mutate(para, leaser, |maybe_current| {
+			if let Some(current) = maybe_current {
+				*current = amount.saturating_add(*current)
```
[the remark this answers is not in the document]
Ank4n: That's a really good catch and not okay. Will fix it..

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Don't have strong opinion. I based it similar to `fn lease_period_index`. 

Though it might be slightly better to return `Option`  since it indicates something unexpected happened and we could not really evaluate whether lease period is ending soon? This would return `None` if there is a overflow/underflow (should improve doc comments though). 

(While writing this, I realised it would be much cleaner to return a Result<bool, DispatchError> instead.)

Ank4n:
bot bench polkadot-pallet --pallet= runtime_common::slots --runtime rococo

Ank4n:
bot bench polkadot-pallet --pallet=runtime_common::slots --runtime rococo

polkadot/runtime/rococo/src/weights/runtime_common_slots.rs
Ank4n: Running the bench now :)

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -86,6 +93,10 @@ pub mod pallet {
 		#[pallet::constant]
 		type LeasePeriod: Get<BlockNumberFor<Self>>;
 
+		/// The number of blocks in past upto which early refund is allowed.
+		#[pallet::constant]
+		type EarliestRefundPeriod: Get<BlockNumberFor<Self>>;
```
joepetrowski, line 98: I do find the name a little confusing. When I was first reading the PR description, I assumed this would be number of _lease periods_.
```suggestion
		/// The number of blocks prior to lease expiration, after which a parachain can refund its
		/// bond prematurely. A value of zero means that funds cannot be refunded until the lease
		/// expires.
		#[pallet::constant]
		type EarliestRefundPeriod: Get<BlockNumberFor<Self>>;
```

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -142,6 +158,11 @@ pub mod pallet {
 		ParaNotOnboarding,
 		/// There was an error with the lease.
 		LeaseError,
+		/// The signing account has no permission to do the operation.
+		NoPermission,
+		/// The signing account has privileges for this operation but the preconditions are not
+		/// met.
+		PreconditionNotMet,
```
joepetrowski, line 165: The context in which these errors arise is not meant to be called by signed accounts.
```suggestion
		/// The origin has no permission to do the operation.
		NoPermission,
		/// The origin has privileges for this operation but the preconditions are not met.
		PreconditionNotMet,
```

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -310,27 +384,112 @@ impl<T: Config> Pallet<T> {
...
-			None => {},
+				None => Zero::zero(),
+			})
+			.max()
+			.unwrap_or_else(Zero::zero)
+	}
+
+	/// Returns true if the current lease period is ending within the next `T::EarliestRefundPeriod`
+	/// blocks.
+	fn lease_period_ending_soon(now: BlockNumberFor<T>) -> bool {
```
joepetrowski, line 411: "Soon" is very ambiguous.
```suggestion
	fn within_early_refund_period(now: BlockNumberFor<T>) -> bool {
```

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -310,27 +384,112 @@ impl<T: Config> Pallet<T> {
...
+		T::Currency::reserve(&leaser, amount).map(|_| {
+			ReservedAmounts::<T>::mutate(para, leaser, |maybe_current| {
+				if let Some(current) = maybe_current {
+					*current = amount.saturating_add(*current)
+				} else {
+					*maybe_current = Some(amount)
+				}
+			});
+			()
+		})
```
joepetrowski, line 440: Maybe a dumb Rust question, but won't the `|_|` swallow an error and just pass it into the closure? As in, if `leaser` didn't have enough funds to reserve, then it would still return `Ok(())`?

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -221,6 +241,51 @@ pub mod pallet {
...
+
+			// check if lease is ending soon.
+			ensure!(
+				Self::lease_period_ending_soon(frame_system::Pallet::<T>::block_number()),
+				Error::<T>::PreconditionNotMet
+			);
+
+			// allow this iff parachain has one lease period left.
+			let leases = Leases::<T>::get(para);
+			ensure!(leases.len() == 1, Error::<T>::PreconditionNotMet);
```
joepetrowski, line 277: Add an integrity test to `assert!(T::EarliestRefundPeriod::get() <= T::LeasePeriod::get())`? Otherwise this could be misconfigured to let people refund more than one LP before expiry but they would still run into this error.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -1160,6 +1395,30 @@ mod benchmarking {
...
+			let (para, _) = register_a_parathread::<T>(1);
+			let leaser: T::AccountId = account("leaser", 0, 0);
+			// go to block where we can lease things
+			frame_system::Pallet::<T>::set_block_number(T::LeaseOffset::get() + One::one());
+			T::Currency::make_free_balance_be(&leaser, BalanceOf::<T>::max_value());
+			// lease out a slot
+			Slots::<T>::force_lease(T::ForceOrigin::try_successful_origin().unwrap(), para, leaser.clone(), 10u32.into(), 1u32.into(), 1u32.into())?;
+			// verify deposit is reserved
+			assert_eq!(T::Currency::reserved_balance(&leaser), 10u32.into());
+			// setup lease for period 1
```
joepetrowski, line 1409:
```suggestion
			// set up lease for period 1
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4353905

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4353903

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4353904

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4393690

polkadot/runtime/common/src/slots/mod.rs
Ank4n: `map` only applies on OK value. `map_err` is used for transforming err values.
https://doc.rust-lang.org/std/result/enum.Result.html#method.map

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -221,6 +256,56 @@ pub mod pallet {
...
 		}
+
+		/// Try to refund the lease deposit before the actual end of the lease.
+		///
+		/// This is only allowed if parachain is in its last lease period and they started with more
+		/// than [`Config::MinLeasePeriodsForEarlyRefund`]. This is useful for parachains who want
+		/// to get access to their funds they used in the last lease and rebid using same for the
+		/// next lease.
+		///
+		/// Can only be called by the Lease Admin or the Parachain manager.
```
joepetrowski, line 267: Looks like `Registrar` and `ForceOrigin`?

Ank4n: Registrar is responsible for registration of para_id. We get the manager of para_id from the registrar. 

Good point about Force Origin, I will fix the doc. The ForceOrigin is set to be LeaseAdmin origin in the polkadot and kusama runtimes but from this pallet's POV, it could be any configured origin.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -221,6 +256,56 @@ pub mod pallet {
...
+					T::ForceOrigin::ensure_origin(origin)
+						.map(|_| ())
+						.map_err(|_| Error::<T>::NoPermission.into())
+				})?;
+
+			let min_lease_period_required = T::MinLeasePeriodsForEarlyRefund::get();
+			ensure!(
+				min_lease_period_required > 0u32.into(),
+				// if set to 0, we never allow early refund of slot deposit.
+				Error::<T>::PreconditionNotMet
```
gpestana, line 291: nit but maybe we want to distinguish this error and throw `Error::<T>::EarlyLeaseReturnNotPossible` instead.

gpestana (changes requested):

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Created a new Error `EarlyRefundDisabled` to indicate that the feature is disabled.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -238,22 +323,24 @@ impl<T: Config> Pallet<T> {
...
 				//
 				// `para` is now just an on-demand parachain.
-				//
-				// Unreserve whatever is left.
 				if let Some((who, value)) = &lease_periods[0] {
-					T::Currency::unreserve(&who, *value);
+					let current_held = Self::deposit_held(para, &who);
+					defensive_assert!(current_held <= *value);
+
+					// unreserve whatever is left of the deposit for the ended lease.
```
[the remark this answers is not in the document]
Ank4n: A lease could contain multiple Lease Periods. By ended lease here I meant that the last LP of their lease has expired.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -310,27 +404,100 @@ impl<T: Config> Pallet<T> {
...
+			});
+			()
+		})
+	}
+
+	/// Unreserve the amount for the parachain, updating the current reserved amount for this leaser
+	/// and parachain.
+	fn unreserve(para: ParaId, leaser: &T::AccountId, amount: BalanceOf<T>) {
+		if amount == Zero::zero() {
+			// nothing to unreserve
```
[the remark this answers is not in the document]
Ank4n: In current code, unreserve could be called with zero if there has been an early lease refund. I could push the zero check to the caller but I guess its fine either way.

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -1725,3 +1731,183 @@ fn cant_bid_on_existing_lease_periods() {
...
+
+		// Cant dissolve yet
+		assert_noop!(
+			Crowdloan::dissolve(signed(1), para_id),
+			CrowdloanError::<Test>::NotReadyToDissolve
+		);
+
+		let lease_end_block = 8 * LeasePeriod::get() + LeaseOffset::get();
+		// early refund can be applied from the 1st block in the last lease period.
+		let early_refund_start_block = lease_end_block - LeasePeriod::get() + 1;
```
gpestana, line 1839: Maybe a check on the boundary would be good. ie. 

```rust
run_to_block(lease_end_block - LeasePeriod::get());
assert_noop!(Slots::early_lease_refund(signed(1), para_id));
```

polkadot/runtime/common/src/slots/migration.rs
```diff
@@ -16,8 +16,98 @@
...
+				ensure!(
+					lease_reserve_amount == expected_deposit,
+					"reserved amount not same as required deposit"
+				);
+
+				ensure!(
+					reserved_balance >= lease_reserve_amount,
+					"reserved amount value should be at least the lease reserve amount"
+				);
+
```
gpestana, line 104: excess care but maybe we can check if all leases are migrated with lease period of 4.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -118,8 +144,22 @@ pub mod pallet {
 	pub type Leases<T: Config> =
 		StorageMap<_, Twox64Concat, ParaId, Vec<Option<(T::AccountId, BalanceOf<T>)>>, ValueQuery>;
 
+	/// Information about the currently leased slot of the leaser.
+	///
+	/// Keyed by para_id and leaser, value is a tuple of current_reserved_amount and
+	/// lease_period_count for the latest lease.
```
gpestana, line 150:
```suggestion
	/// Keyed by `ParaId` and leaser, value is a tuple of `current_reserved_amount` and
	/// `lease_period_count` for the latest lease.
```

or maybe just have the whole text without vars.

polkadot/runtime/common/src/slots/mod.rs
```diff
@@ -267,12 +350,19 @@ impl<T: Config> Pallet<T> {
...
+					let current_hold = Self::deposit_held(para, &ended_lease.0);
+
+					// Note: we check reserved balance from state instead of upcoming leases since
+					// we might have unreserved early. Since reserves are not named, we can't check
+					// if this reserve was made by slots pallet or not. An account could probably
+					// cheat by reserving amount for something else so better to track reserved
+					// amount at all times.
+					if let Some(rebate) = current_hold.checked_sub(&required_hold) {
+						Self::unreserve(para, &ended_lease.0, rebate);
 					}
```
gpestana, line 366: nit but how about an `else` arm here with defensive?

gpestana (approved):

polkadot/runtime/common/src/integration_tests.rs
Ank4n: Good point, added!

polkadot/runtime/common/src/slots/migration.rs
Ank4n: Done 👍

polkadot/runtime/common/src/slots/mod.rs
Ank4n: Yeah good point, made them as whole texts.

polkadot/runtime/common/src/slots/mod.rs
Ank4n: It is possible for current hold to be less than required hold when there is an early lease refund. But actually it should never be possible at a start of a lease.
Added a defensive check 👍.

Ank4n:
I guess its too late to be useful anymore? @SBalaguer 

Closing this PR.

Issue #70, opened 2018-10-26
Report Aura Equivocations to runtime


gnunicorn:
Following paritytech/substrate#911, we have an Aura consensus engine embedded in substrate. And while its verifier rejects unverified blocks, it does not issue misbehavior checks on the wire nor chain yet. Unlike with other block production mechanisms, these aren't happening in messages between one another but at the level where the new block would be attempted to be imported.

The main type of misbehavior to check for is equivocating, or signing multiple blocks in the same slot. Other types of misbehavior, like signing a bad block or signing a block out-of-turn have intrinsic economic penalties associated with them, either by skipping out on rewards or losing reliability with peers in the p2p network.

A proof-of-equivocation is just two headers (the body may be valid _or_ invalid), both signed by the same voter, with seal mentioning the same slot. A runtime module should be created for checking equivocations.

This can be tracked by keeping track of the last N ( = 1000) slots we've seen blocks for per author. We add to this after checking the header. If there is ever one in that position already, we will issue a report to the runtime with our proof. See https://github.com/paritytech/substrate/issues/2181 for a rough outline on how to do such reports.

PR #1766, opened 2023-10-01
AURA offence report system

davxy:
(Requires tweaks to cumulus, but first I want to be sure that there are not blockers with the implementation) 

Due to some characteristics of current aura-pallet implementation, the report system depends on some other pallets.

In particular it requires that:

- Runtime implements `pallet-authorship` to get reporter identity when not directly passed to the report system function
    - This is a requirement which is set by other consensus protocols offence report systems provided by Substrate (i.e. babe, grandpa, beefy).
    - not a huge requirement as most of the implementations who may need the offence report system already include this pallet

- Runtime implements `session-pallet` to map the block number to session index.
    - not a huge requirement as most of the implementations who may need the offence report system already include this pallet

- `session-pallet` `Config` must use the `PeriodicSessions` type as its `NextSessionRotation` associated type. This is to reliably map the block number to session number. The session index is simply given by the block number divided by session duration (which in this case is constant).

---

The ideas were mostly picked from BABE, BEEFY and GRANDPA.
So there isn't really anything new beside the trick to compute the session index from the block number.

During the implementation I also spotted a couple of possible improvement areas which I'm going to highlight in the code comments.

---

Closes https://github.com/paritytech/polkadot-sdk/issues/70

### DIFF
--- a/substrate/frame/aura/Cargo.toml
+++ b/substrate/frame/aura/Cargo.toml
@@ -18,15 +18,19 @@ log = { version = "0.4.17", default-features = false }
 scale-info = { version = "2.5.0", default-features = false, features = ["derive"] }
 frame-support = { path = "../support", default-features = false}
 frame-system = { path = "../system", default-features = false}
+pallet-authorship = { path = "../authorship", default-features = false}
 pallet-timestamp = { path = "../timestamp", default-features = false}
 sp-application-crypto = { path = "../../primitives/application-crypto", default-features = false}
 sp-consensus-aura = { path = "../../primitives/consensus/aura", default-features = false}
 sp-runtime = { path = "../../primitives/runtime", default-features = false}
+sp-session = { path = "../../primitives/session", default-features = false}
+sp-staking = { path = "../../primitives/staking", default-features = false, features = ["serde"]}
 sp-std = { path = "../../primitives/std", default-features = false}
 
 [dev-dependencies]
 sp-core = { path = "../../primitives/core", default-features = false}
 sp-io = { path = "../../primitives/io" }
+env_logger = "0.10"
 
 [features]
 default = [ "std" ]
@@ -35,6 +39,7 @@ std = [
 	"frame-support/std",
 	"frame-system/std",
 	"log/std",
+	"pallet-authorship/std",
 	"pallet-timestamp/std",
 	"scale-info/std",
 	"sp-application-crypto/std",
@@ -42,6 +47,8 @@ std = [
 	"sp-core/std",
 	"sp-io/std",
 	"sp-runtime/std",
+	"sp-session/std",
+	"sp-staking/std",
 	"sp-std/std",
 ]
 try-runtime = [
--- a/substrate/frame/aura/Cargo.toml
+++ b/substrate/frame/aura/Cargo.toml
@@ -59,6 +59,7 @@ try-runtime = [
 	"pallet-timestamp/try-runtime",
 	"sp-runtime/try-runtime",
 ]
+# Provide the default equivocation report system.
 default-equivocation-report-system = [
 	"pallet-authorship",
 	"pallet-session",
--- a/substrate/frame/aura/Cargo.toml
+++ b/substrate/frame/aura/Cargo.toml
@@ -33,7 +33,6 @@ sp-core = { path = "../../primitives/core", default-features = false}
 sp-io = { path = "../../primitives/io" }
 pallet-authorship = { path = "../authorship" }
 pallet-session = { path = "../session" }
-env_logger = "0.10"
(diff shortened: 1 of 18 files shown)


substrate/frame/aura/src/default_weights.rs
```diff
@@ -0,0 +1,54 @@
...
+			.saturating_add(Weight::from_parts(110u64 * WEIGHT_REF_TIME_PER_MICROS, 0))
+			// Report offence
+			.saturating_add(Weight::from_parts(110u64 * WEIGHT_REF_TIME_PER_MICROS, 0))
+			.saturating_add(Weight::from_parts(
+				25u64 * WEIGHT_REF_TIME_PER_MICROS * max_nominators_per_validator as u64,
+				0,
+			))
+			.saturating_add(DbWeight::get().reads(14 + 3 * max_nominators_per_validator as u64))
+			.saturating_add(DbWeight::get().writes(10 + 3 * max_nominators_per_validator as u64))
+	}
```
davxy, line 53: Frankly I don't know how these hardcoded were computed.
Are these estimation? Empirical results? Guesses 😃

I mostly picked them from babe/aura/grandpa implementations

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3860792

substrate/frame/aura/src/default_weights.rs
```diff
@@ -0,0 +1,54 @@
...
+//! Default weights for the AURA Pallet.
+//! This file was not auto-generated.
+
+use frame_support::weights::{
+	constants::{RocksDbWeight as DbWeight, WEIGHT_REF_TIME_PER_MICROS, WEIGHT_REF_TIME_PER_NANOS},
+	Weight,
+};
+
+/// Default `WeightInfo` implementation generic over the max number of validator's nominators (`N`).
+pub struct SubstrateWeight<const N: u32>;
```
davxy, line 27: Here I tried to experiment and detach a bit the consensus from the staking.

That is, instead of adding the `MaxNominators` directly in the `Config`, I decided to provide the (const) value as part of the weight type we provide.

IMO this makes sense as:
1. relax the dependency to the staking model which the user may don't want to use.
2. the `MaxNominators` value is only used here, to compute the weight

Maybe I can also provide the `()` implementation where we set the max nominators to `0`.

If the idea is good, I'm going to do the same for the other consensus protocols (in a follow-up PR)

cc @andresilva

substrate/frame/aura/Cargo.toml
```diff
@@ -50,4 +59,9 @@ try-runtime = [
 	"pallet-timestamp/try-runtime",
 	"sp-runtime/try-runtime",
 ]
+# Provide the default equivocation report system.
+default-equivocation-report-system = [
+	"pallet-authorship",
+	"pallet-session",
+]
```
davxy, line 66: These pallets are required only if the user decides to use the offence report system which ships with substrate.

These pallets are not used otherwise.

substrate/frame/aura/src/equivocation.rs
```diff
@@ -0,0 +1,224 @@
...
+			>>::average_session_length();
+
+		let session_index = key_owner_proof.session();
+
+		let idx1 = block_num1.checked_div(&session_len).unwrap_or(Zero::zero());
+		let idx2 = block_num2.checked_div(&session_len).unwrap_or(Zero::zero());
+
+		if BlockNumberFor::<T>::from(session_index as u32) != idx1 || idx1 != idx2 {
+			return Err(Error::<T>::InvalidKeyOwnershipProof.into())
+		}
```
davxy, line 212: this is the trick to map block num to session index

substrate/frame/aura/src/equivocation.rs
```diff
@@ -0,0 +1,224 @@
...
+			.checked_div(&Period::get())
+			.unwrap_or(Zero::zero());
+		let idx2 = block_num2
+			.saturating_sub(Offset::get())
+			.checked_div(&Period::get())
+			.unwrap_or(Zero::zero());
+
+		if BlockNumberFor::<T>::from(session_index as u32) != idx1 || idx1 != idx2 {
+			return Err(Error::<T>::InvalidKeyOwnershipProof.into())
+		}
```
davxy, line 212: This is the trick to map block num to session index.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3861014

davxy:
Stale

PR #1770, opened 2023-10-02
cleanup a few hidden imports in frame-support

kianenigma:
Just making a few hidden imports cleaner and hidden in docs.


substrate/frame/staking/reward-curve/src/lib.rs
```diff
@@ -81,11 +81,12 @@ pub fn build(input: TokenStream) -> TokenStream {
 
 	let imports = match crate_name("sp-runtime") {
 		Ok(FoundCrate::Itself) => quote!(
-			extern crate sp_runtime as _sp_runtime;
+			#[doc(hidden)]
+			pub use crate as _sp_runtime;
```
KiChjang, line 85: This doesn't look correct; originally the external crate is `sp_runtime`, so the code really should look like:

```suggestion
			pub use sp_runtime as _sp_runtime;
```

kianenigma: yes thanks!

substrate/frame/support/src/lib.rs
```diff
@@ -30,6 +30,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 /// Export ourself as `frame_support` to make tests happy.
+#[doc(hidden)]
```
bkchr, line 33: This should not be visible by default any way?

bkchr (approved):

sam0x17 (approved):

Issue #1724, opened 2023-09-27
PVF: improve test ergonomics


mrcnski:
Due to the [retiring of puppet workers](https://github.com/paritytech/polkadot-sdk/issues/583), the PVF tests now depend not on fake workers but on the actual real workers. This is better as we now test the real thing, on the other hand we now have to remember to manually rebuild the actual workers before running tests. There are two issues with this:

1. The error when you forget to build the workers is not clear at all: https://github.com/paritytech/polkadot-sdk/pull/1192#issuecomment-1733377121
2. After https://github.com/paritytech/polkadot-sdk/pull/1495 the workers now depend on polkadot-cli (for `NODE_VERSION`), which itself transitively depends on the PVF system. This means that if you make a change in PVF and need to test it, all of polkadot-cli gets rebuilt when you build the workers.

Couple of ideas:

1. [ ] Maybe the build script of PVF can make sure the workers were built with the current git commit hash?
2. [ ] ~Move `NODE_VERSION` into some shared location. Maybe something like a `.env` file or some equivalent that works with cargo, so an env var gets set which we can use at compile-time.~
    - After some testing, it seems that even if the workers pull in nothing from `pvf`, they always depend on it. Cargo considers them to have this dependency regardless, because they are `bin` targets in the same crate as `polkadot` and share dependencies with it. To fix it, they need to be in their own crate instead of sharing a crate with `polkadot`.

cc @s0me0ne-unkn0wn

PR #1771, opened 2023-10-02
PVF: Add worker check during tests and benches

mrcnski:
Improves test ergonomics. Please see https://github.com/paritytech/polkadot-sdk/issues/1724.

Also:

- Use an env var for the node version to avoid unnecessary dependencies on polkadot-cli.
- Add a benchmark for preparation through the host (I needed this bench and had to fix these issues when writing it.)
- Some random minor fixes.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/1724


.cargo/config.toml
```diff
@@ -31,3 +31,14 @@ rustflags = [
...
+[env]
+# The version of the node.
+#
+# This is the version that is used for versioning this node binary.
+# By default the `minor` version is bumped in every release. `Major` or `patch` releases are only
+# expected in very rare cases.
+#
+# The worker binaries associated to the node binary should ensure that they are using the same
+# version as the main node that started them.
+POLKADOT_NODE_VERSION = "1.1.0"
```
mrcnski, line 44: @bkchr Moved the node version here to avoid depending on `polkadot-cli`. Needed because in PVF test code I added a version check, see `get_and_check_worker_paths`. Maybe the version could be in a helper crate instead, wasn't sure where to put it.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3864647

polkadot/node/core/pvf/src/host.rs
```diff
@@ -130,12 +130,33 @@ impl ValidationHost {
...
+	#[cfg(feature = "test-utils")]
+	pub async fn prune_all_artifacts(
+		&mut self,
+		result_tx: oneshot::Sender<Result<usize, std::io::Error>>,
+	) -> Result<(), String> {
+		self.to_host_tx
+			.send(ToHost::PruneAllArtifacts { result_tx })
+			.await
+			.map_err(|_| "the inner loop hung up".to_string())
+	}
```
mrcnski: Needed this method to clear the artifacts table for the benchmark. Otherwise the preparation would skip work if the artifact already existed. Could have instead deleted the artifact on-disk, and had `precheck_pvf` check if the artifact existed on-disk and not just in the artifacts table. But that seemed wrong, because in production we don't expect to ever retry prechecking.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3864772

s0me0ne-unkn0wn (approved):
Wow, it's an impressive amount of checks for dev-only code! Nothing has caught my eye, looks good!

.cargo/config.toml
bkchr: I would rather opt in for a helper crate. While this also don't being the best idea, it is probably better than the env variable.

But why not just depend on `polkadot-cli` as dev dependency? What is the problem?

polkadot/node/core/pvf/benches/host_prepare_rococo_runtime.rs
```diff
@@ -0,0 +1,134 @@
...
+};
+use polkadot_primitives::ExecutorParams;
+use std::time::Duration;
+use tokio::{runtime::Handle, sync::Mutex};
+
+const TEST_PREPARATION_TIMEOUT: Duration = Duration::from_secs(30);
+
+struct TestHost {
+	#[allow(unused)]
+	cache_dir: tempfile::TempDir,
```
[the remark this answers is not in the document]
mrcnski: I just copy/pasted `TestHost` from `tests/it/main.rs` and didn't want to change it too much. But we can remove this I guess.

polkadot/node/core/pvf/src/host.rs
```diff
@@ -130,12 +130,33 @@ impl ValidationHost {
...
+		pvf: PvfPrepData,
+		result_tx: PrepareResultSender,
+	},
 	ExecutePvf(ExecutePvfInputs),
-	HeadsUp { active_pvfs: Vec<PvfPrepData> },
+	HeadsUp {
+		active_pvfs: Vec<PvfPrepData>,
+	},
+	#[cfg(feature = "test-utils")]
+	PruneAllArtifacts {
```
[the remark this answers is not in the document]
mrcnski: It shouldn't end up in the production binary since it's gated behind `test-utils`. Is that fine? As per my comment above, there's not really great alternatives.

polkadot/node/core/pvf/benches/host_prepare_rococo_runtime.rs
```diff
@@ -0,0 +1,134 @@
...
+	// The host spinning up a worker will be done in criterion's "warmup" stage, and not counted in
+	// the results.
+	group.bench_function("host: prepare Rococo runtime", |b| {
+		b.to_async(&rt).iter(|| async {
+			// `PvfPrepData` is designed to be cheap to clone, so cloning shouldn't affect the
+			// benchmark accuracy
+			let _stats = host.precheck_pvf(&pvf.clone().code(), Default::default()).await.unwrap();
+
+			// Delete the prepared artifact. Otherwise the next iterations will immediately finish.
+			let num_deleted = host.prune_all_artifacts().await.unwrap();
```
bkchr, line 126: If you just use https://docs.rs/criterion/latest/criterion/struct.AsyncBencher.html#method.iter_batched you don't need to add this hack that isn't required in the production code.

Or you just delete the artifacts manually.

.cargo/config.toml
mrcnski: > But why not just depend on polkadot-cli as dev dependency? What is the problem?

@bkchr It can't be a dev-dependency. Since the new utility is exported to integration tests and benches, we can't gate it with `#[cfg(test)]` or it won't get exported. So it's behind a `test-utils` feature flag, which uses the production dependencies (though it should never end up in production).

I just tried making the dep optional and depending on it in the `test-utils` feature, but it was somehow introducing a cyclic dependency. I moved some things around but couldn't break the cycle easily. Probably due to `cargo`'s feature unification of one of the deps somewhere.

polkadot/node/core/pvf/benches/host_prepare_rococo_runtime.rs
mrcnski: Thanks! It's not production code because of the `test-utils` gate, and deleting the artifacts doesn't work as explained above. `iter_batched` looks like a better solution though. 👍

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -47,3 +53,47 @@ pub fn validate_candidate(
...
+		prepare_worker_path.push(PREPARE_BINARY_NAME);
+		let mut execute_worker_path = workers_path.clone();
+		execute_worker_path.push(EXECUTE_BINARY_NAME);
+
+		// Check that the workers are valid.
+		if !prepare_worker_path.is_executable() || !execute_worker_path.is_executable() {
+			panic!("ERROR: Workers do not exist or are not executable. Workers directory: {:?}", workers_path);
+		}
+
+		let node_version = env!("POLKADOT_NODE_VERSION");
```
bkchr, line 78: So this is now the reason the version is put into an env variable?

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -47,3 +53,47 @@ pub fn validate_candidate(
...
+		}
+		let worker_version =
+			get_worker_version(&execute_worker_path).expect("checked for worker existence");
+		if worker_version != node_version {
+			panic!("ERROR: Execute worker version {worker_version} does not match node version {node_version}; worker path: {execute_worker_path:?}");
+		}
+
+		// We don't want to check against the commit hash because we'd have to always rebuild
+		// the calling crate on every commit.
+		eprintln!("WARNING: Workers match the node version, but may have changed in recent commits. Please rebuild them if anything funny happens. Workers path: {workers_path:?}");
```
bkchr, line 92: You already print this here. So, this doesn't really make any sense to compare the version. If you really want to make it more "safe", you can call here `cargo build` to ensure the worker are there.

polkadot/node/core/pvf/src/testing.rs
mrcnski: Yessir.

polkadot/node/core/pvf/src/testing.rs
mrcnski: Well, this warning won't show up unless `-- --nocapture` is passed. I left it because it's better than nothing I guess. I think it's still useful to separately check for worker existence and that they're on the same version.

> If you really want to make it more "safe", you can call here cargo build to ensure the worker are there.

You mean `cargo build` in build.rs, every time the version changes?

bkchr: No, I mean this function does this at "runtime". This is only being used for testing stuff. So, it could ensure that the workers are up to date. It would also make testing easier, as people would not be required to first run `cargo build` after finding out that the workers are not there.

bkchr (approved):
Would still like that `get_and_check_worker_paths` just calls `cargo build` to provide some better UX for tests. However, this isn't a real blocker as the current behavior isn't changed.

Issue #226, opened 2023-01-27
FRAME: Move pallets over to use `fungible` traits


gavofyork:
- Move over all pallets using `Currency` to `fungible` traits:
  - Basic code switch over: pallets should provide a utility for cancelling a `lock`/`reserve` and instating a `freeze`/`hold`.
  - Staking pallet should use `hold` instead of `lock`.
  - Permissionless fee-free dispatchable in all pallets to move over.
  - Mark all `Currency` traits as deprecated.

Eventually, we will also want to deprecate and remove `Balances` dispatchables:
- Substitute all usages of Balances dispatchables with `fungible` API.
- Mark dispatchables as deprecated.
- Remove them (this can only be done once XCM execution from private accounts is safe and enabled).

## Pallet migrations needed

Here's a list of pallets, turning this into a tracking issue of the migration process. 

- [ ] ./frame/**staking**/src/pallet/mod.rs `type Currency: LockableCurrency<`
- [ ] ./frame/**assets**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>;`
- [ ] ./frame/**elections-phragmen**/src/lib.rs `type Currency: LockableCurrency<Self::AccountId, Moment = Self::BlockNumber>`
- [ ] ./frame/**election-provider-multi-phase**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId> + Currency<Self::AccountId>;`
- [ ] ./frame/**lottery**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>;`
- [ ] ./frame/**democracy**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>`
- [ ] ./frame/**treasury**/src/lib.rs `type Currency: Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>;`
- [ ] ./frame/**nicks**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>;`
- [ ] ./frame/**state-trie-migration**/src/lib.rs `type Currency: Currency<Self::AccountId>;`
- [ ] ./frame/**nfts**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>;`
- [ ] ./frame/**uniques**/src/lib.rs `type Currency: ReservableCurrency<Self::AccountId>;`
- [x] /frame/nis/src/lib.rs `type Currency: FunInspect<Self::AccountId, Balance = Self::CurrencyBalance>`
[...]

PR #1772, opened 2023-10-02
migrate pallet-fast-unstake to `fungible::*` trait family

kianenigma:
part of #226. 

Note that many other pallets could possibly use the `Consideration` api if they are purely holding deposits. In the case of this pallet, I decided to tap into the `fungible::*` traits directly because the type of deposit held in this pallet is fixed, and very rudimentary. We don't want to care about if `type Deposit` even changes. The pallet should at any point put `Deposit::get()` on hold, and upon successful operation, should release all of it. 

I think for these pallets the `fungible` based approach is better because it gives you the ability to stop tracking the amount of deposit held. Contrary, in the `Consideration` you need to store the `ticket` so that you can update or release the deposit. 

I might nonetheless try this approach as well. 

The PR contains some logical changes to the pallet and a migration, namely because, as noted above, we no longer need to store the deposit amount.

### DIFF
--- a/substrate/frame/contracts/src/lib.rs
+++ b/substrate/frame/contracts/src/lib.rs
@@ -247,8 +247,7 @@ pub mod pallet {
 		type Randomness: Randomness<Self::Hash, BlockNumberFor<Self>>;
 
 		/// The fungible in which fees are paid and contract balances are held.
-		type Currency: Inspect<Self::AccountId>
-			+ Mutate<Self::AccountId>
+		type Currency: Mutate<Self::AccountId>
 			+ MutateHold<Self::AccountId, Reason = Self::RuntimeHoldReason>;
 
 		/// The overarching event type.
--- a/substrate/frame/fast-unstake/src/lib.rs
+++ b/substrate/frame/fast-unstake/src/lib.rs
@@ -152,7 +152,11 @@ pub mod pallet {
 	use crate::types::*;
 	use frame_support::{
 		pallet_prelude::*,
-		traits::{Defensive, ReservableCurrency, StorageVersion},
+		traits::{
+			fungible::MutateHold as FunMutateHold,
+			tokens::{Fortitude, Precision},
+			Defensive, StorageVersion,
+		},
 	};
 	use frame_system::pallet_prelude::*;
 	use sp_runtime::{traits::Zero, DispatchResult};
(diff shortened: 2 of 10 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3865479

kianenigma:
I was under the impression that this pallet has some bound on the total size of the queue and therefore can be migrated in a one-off manner. But I see now that there is bound, so I need to rework this into a lazy migration.

substrate/frame/contracts/src/lib.rs
```diff
@@ -247,8 +247,7 @@ pub mod pallet {
 		type Randomness: Randomness<Self::Hash, BlockNumberFor<Self>>;
 
 		/// The fungible in which fees are paid and contract balances are held.
-		type Currency: Inspect<Self::AccountId>
-			+ Mutate<Self::AccountId>
+		type Currency: Mutate<Self::AccountId>
```
franciscoaguirre, line 250: Why was this change made? Seems totally independent

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3895814

substrate/frame/contracts/src/lib.rs
PieWol: I think it was a useless line as `Inspect` is a subtrait to `Mutate`

substrate/frame/fast-unstake/src/lib.rs
```diff
@@ -558,26 +576,37 @@ pub mod pallet {
...
-					T::Currency::slash_reserved(&stash, deposit);
-					log!(info, "slashed {:?} by {:?}", stash, deposit);
-					Self::deposit_event(Event::<T>::Slashed { stash, amount: deposit });
+					let _ = T::Currency::burn_all_held(
+						&HoldReason::FastUnstake.into(),
+						&stash,
+						Precision::BestEffort,
+						// This is not a system-level slash, so we politely do our best to slash.
+						Fortitude::Polite,
+					)
```
juangirini, line 606: `slash_reserved` created a negative imbalance, whereas `burn_all_held` just burns it. Don't we still need that negative imbalance?

juangirini: If we need to slash instead of burn, we might need a `slash_all_held`

Issue #1597, opened 2023-09-16
Improve timeout-based tests


s0me0ne-unkn0wn:
Not long ago, a couple of PVF host tests [were gated through the `ci-only-tests` feature](https://github.com/paritytech/polkadot/pull/7390) as they were found to be flaky when run locally. @bkchr was arguing it is a bad idea (and it really is), but reviewers concluded the solution is acceptable for the moment.

Naturally, that was forgotten when building the monorepo. Right now, those tests are never run.

Of course, we could just put the feature back into the CI scripts and forget about it... Again. But the very fact that it happened has proven that the approach is really error-prone.

The tests in question are timeout-based. That is, the test is allowed to finish within some time frame, which proves that the work tested is executed with the expected parallelism (or, vice-versa, without any parallelism where it is undesirable). The concrete examples are [`ensure_parallel_execution`](https://github.com/paritytech/polkadot-sdk/blob/11d1a3955e4bfd536e7716132f5440d7e75e43e5/polkadot/node/core/pvf/tests/it/main.rs#L135) and [`execute_queue_doesnt_stall_with_varying_executor_params`](https://github.com/paritytech/polkadot-sdk/blob/11d1a3955e4bfd536e7716132f5440d7e75e43e5/polkadot/node/core/pvf/tests/it/main.rs#L218), but that's not an exhaustive list; other timeout-based tests also present, we just have never seen them flaky and thus haven't gate them.

Under this issue, I'd like to collect ideas on how to improve them and make them less dependent on host load; ideally, they shouldn't be load-dependent at all.

Please speak out if you have any ideas.

Issue #677, opened 2023-03-28
PVF: Compromised artifact file integrity can lead to disputes


mrcnski:
# ISSUE

## Overview

If we try to execute a corrupt PVF artifact it will likely lead to a dispute. See the comment below for an example of this. 

A simple mitigation is to just put the hash of the file contents in the filename. Artifacts are not that big, on the order of several dozen MB (I think?). If we use something like https://github.com/BLAKE3-team/BLAKE3, which claims 6866 MiB/s throughput, then the time spent hashing would barely register. 

## Original comment

As far as I understand, the main concern about not clearing the artifact cache is:
1) Unsafe functions are used to import the artifact
2) [Their safety guarantees](https://github.com/paritytech/polkadot/blob/2bbef460d24f44aeba9d4014062bb8c631327496/node/core/pvf/src/executor_intf.rs#L199) rely on the fact the node has produces them with `prepare()` and they are not just some random files put there by someone.

Not that we can guarantee nobody will overwrite them during the node run time. I believe the cache pruning was implemented as a measure to mitigate those unwanted outcomes. But I don't have a strong opinion if it makes any sense. If someone wants to screw things up, he will.

_Originally posted by @s0me0ne-unkn0wn in https://github.com/paritytech/polkadot-sdk/issues/685

s0me0ne-unkn0wn:
Not sure I follow. You mean, hash a prepared artifact right after the preparation, include its hash into its name, and check the hash before execution? But how do you know the name of the artifact you're going to open for execution, then? Sounds like a chicken and egg situation.

mrcnski:
> Not sure I follow. You mean, hash a prepared artifact right after the preparation, include its hash into its name, and check the hash before execution? But how do you know the name of the artifact you're going to open for execution, then? Sounds like a chicken and egg situation.

I was thinking on node startup, we just list all the artifacts in the cache path, check consistency with the hash, and load them into the artifacts table? As mentioned above, hashing before each execution would be too much of a performance hit.

s0me0ne-unkn0wn:
Aha, thanks for the elaboration, that does make sense. But that would protect from accidental artifact corruption but not from malicious action. On the other hand, I don't see any easy way to protect against malicious artifacts while still keeping them on disk anyway. (Sign them with the validator key maybe? But that would require some effort)

eagr:
[...]

PR #1774, opened 2023-10-03
Check executor params coherence

### LINKED ISSUE #1704: Stricter checks for `ExecutorParams` soundness
As rightly [noted](https://github.com/paritytech/polkadot-sdk/pull/1192#discussion_r1335599938) by @mrcnski, people make mistakes, and the governance consists of people. Although the `ExecutorParams` restrictions are properly documented, a situation may arise when they are not totally sound. Some probable concerns are:
* Inclusion of multiple occurrences of the same `ExecutorParam` variant, in which case only the first one will take effect;
* Inclusion of a parameter value that overcomes limits imposed by the underlying code (e.g., `PrecheckingMaxMemory` is `u64`, but if its value is over `isize::MAX`, a panic will occur);
* Inclusion of senseless values (e.g., a backing timeout longer than the block time).

`ExecutorParams` are currently set as a part of `HostConfiguration`, and there's already a mechanism to check for the configuration soundness that allows refusing an unsound configuration: [the consistency check](https://github.com/paritytech/polkadot-sdk/blob/f209b31b48586bca81c0ffafa150e1796ee05845/polkadot/runtime/parachains/src/configuration.rs#L362). It should be complemented with appropriate checks for `ExecutorParams` soundness.

eagr:
closes #1704

Really appreciate the help and patience from @s0me0ne-unkn0wn 🙏
I left some unsolved questions in the FIXME comments, would you take a look?


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3866703

s0me0ne-unkn0wn (changes requested):
Thanks for the contribution, it's highly appreciated! Left some comments on how to improve it.

@eskimor @sandreim @slumber we need some brainstorming here. We want to limit values of backing and approval timeouts that could be set via `ExecutorParams` to some reasonable range, that is, to forbid setting a value that breaks things. Technically, for backing timeout, we could limit from 0 to `block_time`, and that's good enough to rule out obvious errors in configuration. Practically, I'd like to have more reasonable values, as setting the backing limit to 0, or, say, 0.01 sec, definitely breaks everything. My first guess is to limit it from 1/12 blocktime to 1/2 blocktime but I may not have a solid enough understanding of underlying logic, especially after async backing has landed. Do you have better ideas?

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +149,138 @@ impl ExecutorParams {
...
+				WasmExtBulkMemory => if enable_bulk_mem {
+					errors.push(DuplicatedParam(param_ident));
+				} else {
+					enable_bulk_mem = true;
+				}
+			}
+
+			// FIXME is it valid if only one is present?
+			if let (Some(lm), Some(nm)) = (logical_max, native_max) {
+				if nm < 128 * lm || nm > 512 * lm {
```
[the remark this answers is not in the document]
eagr: Yeah, I guess this could the policy for the rest of params that if we don't have a good candidate for a bound.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -28,33 +28,52 @@ use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
 use sp_std::{ops::Deref, time::Duration, vec, vec::Vec};
 
+const MEMORY_PAGES_LIMIT: u32 = 65536;
+const SHADOW_STACK_PAGES: u32 = 32;
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Valid point. Okay, we can do it the other way. We can stop accounting for those additional pages in the limit (set the limit to (0, 65536] as it was in the beginning). Instead, we can always add those 32 pages to `*max_pages` [here](https://github.com/paritytech/polkadot-sdk/blob/aad80cce318fd492ce0d9c9951f2d7ec949710f3/polkadot/node/core/pvf/common/src/executor_intf.rs#L110C59-L110C59) (with `.saturating_add()` for the sake of correctness). Thus, the limit set on-chain becomes pure Wasm linear memory limit, implicitly increased by those additional pages. In that case, no constant moving is needed.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +149,138 @@ impl ExecutorParams {
...
+		let mut logical_max: Option<u32> = None;
+		let mut native_max: Option<u32> = None;
+		let mut pvf_mem_max = false;
+		let mut pvf_prep_precheck: Option<u64> = None;
+		let mut pvf_prep_lenient: Option<u64> = None;
+		let mut pvf_exec_backing: Option<u64> = None;
+		let mut pvf_exec_approval: Option<u64> = None;
+		let mut enable_bulk_mem = false;
+
+		for param in &self.0 {
```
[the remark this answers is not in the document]
eagr: I did think about using HashMap coz I also need to cache the values, but couldn't seem to import it easlily. Thanks for the heads up! I'll try BTreeMap.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3871745

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +149,138 @@ impl ExecutorParams {
...
+
+			// FIXME is it valid if only one is present?
+			if let (Some(lm), Some(nm)) = (logical_max, native_max) {
+				if nm < 128 * lm || nm > 512 * lm {
+					errors.push(IncompatibleValues("StackLogicalMax", "StackNativeMax"));
+				}
+			}
+
+			if let (Some(precheck), Some(lenient)) = (pvf_prep_precheck, pvf_prep_lenient) {
+				if precheck >= lenient {
```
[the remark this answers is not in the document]
eagr: Using literals until then.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3873605

polkadot/primitives/src/v6/executor_params.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: For the next couple of weeks, people are mostly OOO, so I don't expect quick responses... Let's just make it, I don't see how someone would object to this change; the timeouts are Polkadot's internal kitchen, and it's quite logical for them to be placed with the primitives.

polkadot/primitives/src/v6/mod.rs
```diff
@@ -1561,7 +1568,7 @@ impl parity_scale_codec::Decode for CompactStatement {
 	) -> Result<Self, parity_scale_codec::Error> {
 		let maybe_magic = <[u8; 4]>::decode(input)?;
 		if maybe_magic != BACKING_STATEMENT_MAGIC {
-			return Err(parity_scale_codec::Error::from("invalid magic string"))
+			return Err(parity_scale_codec::Error::from("invalid magic string"));
```
s0me0ne-unkn0wn, line 1571: Looks like your formatter is adding a bunch of unsolicited corrections... It won't pass formatting check in CI like this. Are you using the latest nightly toolchain and `cargo +nightly fmt`?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3887439

s0me0ne-unkn0wn (commented):
Looks good! Now waiting for other teammates to come back and discuss the exact values of limits.

s0me0ne-unkn0wn (approved):

eagr:
In the mean time, any good next issue I could work on?

s0me0ne-unkn0wn:
Hmm, maybe #677 (somewhat connected with this one), but we'd still need @mrcnski there.

Also, there's a test improvement puzzle #1597. It would be good to have it solved, but it's not urgent and nobody has time to even start thinking about it.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +182,126 @@ impl ExecutorParams {
 		}
 		None
 	}
+
+	/// Check params coherence.
+	pub fn check_consistency(&self) -> Result<(), ExecutorParamError> {
```
mrcnski, line 187: Can we have this logic right in `configuration.rs`? This file seems to be mainly defining a bunch of types with not much actual logic.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +182,126 @@ impl ExecutorParams {
...
+				seen.insert($param, $val as u64);
+			};
+
+			// should check existence before range
+			($param:ident, $val:expr, $out_of_limit:expr $(,)?) => {
+				if seen.contains_key($param) {
+					return Err(DuplicatedParam($param))
+				}
+				if $out_of_limit {
+					return Err(LimitExceeded($param))
```
mrcnski, line 207: Maybe should be renamed to `OutsideLimit`.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -26,35 +26,87 @@ use parity_scale_codec::{Decode, Encode};
...
 /// The different executor parameters for changing the execution environment semantics.
 #[derive(Clone, Debug, Encode, Decode, PartialEq, Eq, TypeInfo, Serialize, Deserialize)]
 pub enum ExecutorParam {
 	/// Maximum number of memory pages (64KiB bytes per page) the executor can allocate.
+	/// A valid value lies within (0, 65536].
 	#[codec(index = 1)]
 	MaxMemoryPages(u32),
-	/// Wasm logical stack size limit (max. number of Wasm values on stack)
+	/// Wasm logical stack size limit (max. number of Wasm values on stack).
+	/// A valid value lies within [1024, 2 * 65536].
```
mrcnski, line 71: Can we reference the constant names instead of having these numbers in the documentation? There's a high risk of the docs and constants going out of sync.

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -130,6 +182,126 @@ impl ExecutorParams {
...
 	}
+
+	/// Check params coherence.
+	pub fn check_consistency(&self) -> Result<(), ExecutorParamError> {
+		use ExecutorParam::*;
+		use ExecutorParamError::*;
+
+		let mut seen = BTreeMap::<&str, u64>::new();
+
+		macro_rules! check {
```
mrcnski, line 193: I'm loving the checks you added, but is there a reason we use a macro instead of a simple fn here?

polkadot/primitives/src/v6/executor_params.rs
```diff
@@ -26,35 +26,87 @@ use parity_scale_codec::{Decode, Encode};
...
+pub const DEFAULT_NATIVE_STACK_MAX: u32 = 256 * 1024 * 1024;
+
+const MEMORY_PAGES_MAX: u32 = 65536;
+const LOGICAL_MAX_LO: u32 = 1024;
+const LOGICAL_MAX_HI: u32 = 2 * 65536;
+const PRECHECK_MEM_MAX_LO: u64 = 256 * 1024 * 1024;
+const PRECHECK_MEM_MAX_HI: u64 = 16 * 1024 * 1024 * 1024;
+
+// Default PVF timeouts. Must never be changed! Use executor environment parameters to adjust them.
+// See also `PvfTimeoutKind` docs.
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Ah yes, it was split into `PvfPrepTimeoutKind` and `PvfExecTimeoutKind` long ago, but docs wasn't updated

```suggestion
// See also `PvfPrepTimeoutKind` and `PvfExecTimeoutKind` docs.
```

polkadot/node/core/pvf/common/src/executor_intf.rs
```diff
@@ -106,8 +109,9 @@ pub fn params_to_wasmtime_semantics(par: &ExecutorParams) -> Result<Semantics, S
 	for p in par.iter() {
 		match p {
 			ExecutorParam::MaxMemoryPages(max_pages) =>
-				sem.heap_alloc_strategy =
-					HeapAllocStrategy::Dynamic { maximum_pages: Some(*max_pages) },
+				sem.heap_alloc_strategy = HeapAllocStrategy::Dynamic {
+					maximum_pages: Some((*max_pages).saturating_add(DEFAULT_HEAP_PAGES_ESTIMATE)),
```
mrcnski, line 113: Nice! But can we just use the value from `sem` instead?

mrcnski (changes requested):
Great stuff, much appreciate your time!

polkadot/primitives/src/v6/executor_params.rs
eagr: Yeah, it would be nice to if we could put the checks in one place. But that may require opening up the internal of `ExecutorParams`, do you think it's worth it? @mrcnski

polkadot/primitives/src/v6/executor_params.rs
eagr: The function implementation might be more complicated: more inputs, branches, and more jumps. Rough idea:

```
fn check_param(seen, name, val, predicate) -> Result
check(seen, name, val, predicate)?
```

Using macro here seems simpler.

polkadot/node/core/pvf/common/src/executor_intf.rs
```diff
@@ -43,7 +46,7 @@ const DEFAULT_HEAP_PAGES_ESTIMATE: u32 = 32;
 const EXTRA_HEAP_PAGES: u32 = 2048;
 
 /// The number of bytes devoted for the stack during wasm execution of a PVF.
-pub const NATIVE_STACK_MAX: u32 = 256 * 1024 * 1024;
+pub const NATIVE_STACK_MAX: u32 = DEFAULT_NATIVE_STACK_MAX;
```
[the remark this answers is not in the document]
eagr: I see it's being used by the executo worker. You mean?

polkadot/node/core/pvf/common/src/executor_intf.rs
s0me0ne-unkn0wn: I think we discussed it [here](https://github.com/paritytech/polkadot-sdk/pull/1774#discussion_r1343770041)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3928083

polkadot/node/core/pvf/common/src/executor_intf.rs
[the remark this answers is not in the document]
mrcnski: I mean why have `DEFAULT_NATIVE_STACK_MAX` and `NATIVE_STACK_MAX` when they are exactly the same? 🤔

polkadot/primitives/src/v6/executor_params.rs
mrcnski: Not sure what you mean exactly, but I *think* it's fine.

polkadot/node/core/pvf/common/src/executor_intf.rs
mrcnski: Oh I misread this code, nevermind. Should it be this instead though?

```suggestion
					maximum_pages: Some((*max_pages).saturating_add(EXTRA_HEAP_PAGES)),
```

And are these values correct? I would have thought the extra pages were 32. @s0me0ne-unkn0wn 

```rs
const DEFAULT_HEAP_PAGES_ESTIMATE: u32 = 32;
const EXTRA_HEAP_PAGES: u32 = 2048;
```

s0me0ne-unkn0wn: No, I believe everything is okay with the current version.

`EXTRA_HEAP_PAGES` is the number of pages of linear memory the instance is allowed to allocate in addition to its initial memory size.

`DEFAULT_HEAP_PAGES_ESTIMATE` is the number of pages the instance will use for its own purposes, they are not part of linear memory.

So, the corresponding executor parameter defines the number of linear memory pages the instance is allowed to allocate, and here, we implicitly add those auxiliary pages to that limit (for example, we can set the limit to 0, and Wasm code not using any linear memory should still be functional, but it's not possible without having some space for shadow stack and other bits).

s0me0ne-unkn0wn (approved):

polkadot/primitives/src/v6/executor_params.rs
eagr: Sorry, I was thinking that we might need to make the internal state public in order to do the checks in another crate, like `pub struct ExecutorParams(pub Vec<ExecutorParam>)`. Then it's possible that other crates may rely on the internal state which is not great.

I just realized that it comes with an iterator :)

But still, for the sake of separation of concerns, I think it wouldn't be a bad idea that the type does its own checks. Like, what if the checks are needed else where (other than `configurations.rs`)?

polkadot/node/core/pvf/common/src/executor_intf.rs
eagr: Would it clarify stuff if we name it `SHADOW_STACK_HEAP_PAGES`? Then current name may be a bit vague.

polkadot/node/core/pvf/common/src/executor_intf.rs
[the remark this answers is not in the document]
eagr: aye aye :)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3942649

polkadot/primitives/src/v6/executor_params.rs
mrcnski: Fair enough. :)

mrcnski (approved):
Thanks! ⭐️

polkadot/node/core/pvf/common/src/executor_intf.rs
s0me0ne-unkn0wn: It's not just shadow stack, according to the comment :thinking: But I doubt if the reasoning behind the calculation provided is correct enough. I didn't dig deeper into this aspect of Wasmtime, but it seems obvious that data segment pages shouldn't be allocated implicitly. Wasm source must specify the initial memory size, and it should always be enough to accommodate the data segment at least. Considering that I'm not the author of the original constant naming, I'm not 100% sure what is better, but probably renaming `DEFAULT_HEAP_PAGES_ESTIMATE` to `INITIAL_HEAP_PAGES_ESTIMATE` would bring more clarity.

eagr:
Thank you guys for spending the time to review! Is this good to go now?

Issue #1765, opened 2023-09-30
Parachains zombienet fails on local machines


alexggh:
The error I'm getting when running zombienet locally is:
```
WARN tokio-runtime-worker parachain::pvf: cannot bind unix socket: Custom { kind: InvalidInput, 
error: "path must be shorter than libc::sockaddr_un.sun_path" } 
debug_id=prepare program_path="/home/alexggh/repos/polkadot-sdk/target/testnet/polkadot-prepare-worker" 
extra_args=["prepare-worker", "--node-impl-version", "1.1.0"] 
worker_dir=WorkerDir { path: "/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO/alice/data/chains/rococo_local_testnet/db/full/pvf-artifacts/worker-dir-prepare-2Q4B1v7qio" } 
socket_path="/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO/alice/data/chains/rococo_local_testnet/db/full/pvf-artifacts/worker-dir-prepare-2Q4B1v7qio/socket"
```

This is because by default zombienet decides to prefix things with `/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO` and that gets us past the socket limitation size.

## Workaround

This can be easily worked around by just specifying explicitly a shorter prefix on zombienet with:
```
zombienet -d /tmp/x/ <your_zombienet_params>
```

This was introduced with https://github.com/paritytech/polkadot-sdk/pull/1373 fyi @mrcnski and it wasn't caught by the CI because it has a shorter prefix. 

Given that we can't really control what prefix paths our users decide to use or already have configured on their validators I think we should try to either prevent this limitation or expose(if it doesn't exist already) some cli parameter or environment variable  for users to be able to manually decide this path.

mrcnski:
Ooh, looks like [the limit is 108 bytes](https://man7.org/linux/man-pages/man7/unix.7.html). This should definitely be fixed, because like you say, users are allowed to use an arbitrary cache path. There may be no need for the socket to be under the cache path though, I think we can revert to using a tmp location like we were before.

PR #1780, opened 2023-10-03
PVF: Add back socket path parameter, use tmp socket path

mrcnski:
## Problem

Running zombienet locally:
```
WARN tokio-runtime-worker parachain::pvf: cannot bind unix socket: Custom { kind: InvalidInput, 
error: "path must be shorter than libc::sockaddr_un.sun_path" } 
debug_id=prepare program_path="/home/alexggh/repos/polkadot-sdk/target/testnet/polkadot-prepare-worker" 
extra_args=["prepare-worker", "--node-impl-version", "1.1.0"] 
worker_dir=WorkerDir { path: "/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO/alice/data/chains/rococo_local_testnet/db/full/pvf-artifacts/worker-dir-prepare-2Q4B1v7qio" } 
socket_path="/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO/alice/data/chains/rococo_local_testnet/db/full/pvf-artifacts/worker-dir-prepare-2Q4B1v7qio/socket"
```

This is because by default zombienet decides to prefix things with `/tmp/zombie-a609ef68a12d8517c8174b62ee3a0899_-484651-6nFd48QbCHfO` and that gets us beyond the socket limitation size.

## Fix

Looks like [the socket path limit is 108 bytes](https://man7.org/linux/man-pages/man7/unix.7.html). This should definitely be fixed, users are allowed to use an arbitrary cache path. There is no need for the socket to be under the cache path though, we can revert to using a tmp location like we were before.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/1765
This was introduced with https://github.com/paritytech/polkadot-sdk/pull/1373

### DIFF
--- a/polkadot/node/core/pvf/common/src/worker/mod.rs
+++ b/polkadot/node/core/pvf/common/src/worker/mod.rs
@@ -18,7 +18,7 @@
 
 pub mod security;
 
-use crate::{worker_dir, SecurityStatus, LOG_TARGET};
+use crate::{SecurityStatus, LOG_TARGET};
 use cpu_time::ProcessTime;
 use futures::never::Never;
 use std::{
@@ -115,6 +115,7 @@ macro_rules! decl_worker_main {
 				},
 			}
 
+			let mut socket_path = None;
 			let mut worker_dir_path = None;
 			let mut node_version = None;
 			let mut can_enable_landlock = false;
@@ -123,6 +124,10 @@ macro_rules! decl_worker_main {
 			let mut i = 2;
 			while i < args.len() {
 				match args[i].as_ref() {
+					"--socket-path" => {
+						socket_path = Some(args[i + 1].as_str());
+						i += 1
+					},
 					"--worker-dir-path" => {
 						worker_dir_path = Some(args[i + 1].as_str());
 						i += 1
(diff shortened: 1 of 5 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3869829

alindima (approved):
LGTM, minus the failing test

s0me0ne-unkn0wn:
@pepoviola what's the rationale behind having that long directory suffixes in the first place?

s0me0ne-unkn0wn:
@mrcnski strictly speaking, the proposed fix fixes the case, but not the entire problem. That weird `sun_path` limit is the reason why it is a good practice to put socket files under `/run`, `/var/run`, or directly under `/tmp`. Is there any reason why shouldn't we follow that practice?

Ah yes, we're running as non-root, and `/run` is not guaranteed to be writeable. But why not just `/tmp` then?

s0me0ne-unkn0wn (approved):
LGTM as a quick fix, but probably better approach should be considered in the future.

pepoviola:
> @pepoviola what's the rationale behind having that long directory suffixes in the first place?

The idea was to have some pattern like `namespace_random` but I think we can shrink to only use the namespace that will be already pretty unique. I will open an issue to track this.
Thanks!

alexggh (approved):

pepoviola (approved):

mrcnski:
> @mrcnski strictly speaking, the proposed fix fixes the case, but not the entire problem. That weird `sun_path` limit is the reason why it is a good practice to put socket files under `/run`, `/var/run`, or directly under `/tmp`. Is there any reason why shouldn't we follow that practice?
> 
> Ah yes, we're running as non-root, and `/run` is not guaranteed to be writeable. But why not just `/tmp` then?

We're using `std::env::temp_dir` (in `with_transient_socket_path` we call `tmppath` which calls `temp_dir`), which should return `/tmp`:

> On Unix, returns the value of the TMPDIR environment variable if it is set, otherwise for non-Android it returns /tmp.

Issue #198, opened 2023-03-23
[FRAME Core] Simple Multi-block Migrations


gavofyork:
Needed for non-trivial migrations on parachains, and helpful for such migration on Relay/solo chains.

This is just a simple strategy and will work in probably around 80% of circumstances. It's good for pallets which are not depended upon by code which fires in `on_initialize`.

Have a migrations pallet (not too dissimilar from [that of Moonbeam](https://github.com/PureStake/moonbeam/blob/4c22f7c010573513076b618c09d2824ece6c51eb/pallets/migrations/src/lib.rs#L123)). This is configured with a tuple of all migrations, which provides introspection and access to each migration which needs to be completed. Migrations look like calls/tasks: they have a possibly-pessimistic weight estimation and can return the actual weight consumed. They are always benchmarked. Migrations should be written to proceed "piecewise" (e.g. one unit at a time, rather than migrating all entries of a storage map all at once).

Each migration defines a type `Cursor` which is `Default + FullCodec + MaxEncodeLen`. The migration function accepts a `Cursor` value and returns an `Option<Cursor>` value which is `None` if the migration is complete. If `Some` then it is up to the migration pallet to ensure it is used in the next call to the migrate function, storing it in state if weight constraints deem it necessary to wait until the next block before resuming.

The migrations pallet should suspend all non-critical subsystems in `on_runtime_upgrade`. This includes XCM and transaction processing. It should only re-enable them when all migrations' cursors are `None`.

One new hook function should be added: `fn poll() -> Weight`. This is guaranteed to execute regularly but not necessarily every block, providing a softer version of `on_initialize`. All Substrate pallets should have their `on_initialize`/`on_finalize` functions checked; if they can be moved or adapted to `poll`, they should be. While `on_initialize`/`on_finalize` should be called within an incomplete migration, `poll` need not (and indeed, *should not*) be.

### Deprecate `on_*` hooks

`on_initialize`/`on_finalize` should be deprecated. Pallets which absolutely require per-block-execution hooks on hard-deadlines (HDH) or mandatory extrinsics (ME) should go through a System config trait item which can give them low-level access at the cost of requiring the System config trait to be explicitly configured to reference them. There should be a regular facility for soft-deadline execution (which will work fine in almost all use-cases of parachain teams) which is implemented using tasks.

### Guarantees

We should mark all storage items so they fall into one of three groups:

1. Can be accessed during a hard-deadline-hook or mandatory extrinsic
[...]

PR #1781, opened 2023-10-03
Multi-Block-Migrations, `poll` hook and new System callbacks

ggwpez:
This MR is the merge of https://github.com/paritytech/substrate/pull/14414 and https://github.com/paritytech/substrate/pull/14275. It implements [RFC#13](https://github.com/polkadot-fellows/RFCs/pull/13), closes https://github.com/paritytech/polkadot-sdk/issues/198.

----- 

This Merge request introduces three major topicals:

1. Multi-Block-Migrations
1. New pallet `poll` hook for periodic service work
1. Replacement hooks for `on_initialize` and `on_finalize` in cases where `poll` cannot be used

and some more general changes to FRAME.  
The changes for each topical span over multiple crates. They are listed in topical order below.

# 1.) Multi-Block-Migrations

Multi-Block-Migrations are facilitated by creating `pallet_migrations` and configuring `System::Config::MultiBlockMigrator` to point to it. Executive picks this up and triggers one step of the migrations pallet per block.  
The chain is in lockdown mode for as long as an MBM is ongoing. Executive does this by polling `MultiBlockMigrator::ongoing` and not allowing any transaction in a block, if true.  

A MBM is defined through trait `SteppedMigration`. A condensed version looks like this:  
```rust
/// A migration that can proceed in multiple steps.
pub trait SteppedMigration {
	type Cursor: FullCodec + MaxEncodedLen;
	type Identifier: FullCodec + MaxEncodedLen;

	fn id() -> Self::Identifier;

	fn max_steps() -> Option<u32>;

	fn step(
		cursor: Option<Self::Cursor>,
		meter: &mut WeightMeter,
	) -> Result<Option<Self::Cursor>, SteppedMigrationError>;
}
```

`pallet_migrations` can be configured with an aggregated tuple of these migrations. It then starts to migrate them one-by-one on the next runtime upgrade.  
Two things are important here:
- 1. Doing another runtime upgrade while MBMs are ongoing is not a good idea and can lead to messed up state.
- 2. **Pallet Migrations MUST BE CONFIGURED IN `System::Config`, otherwise it is not used.**

The pallet supports an `UpgradeStatusHandler` that can be used to notify external logic of upgrade start/finish (for example to pause XCM dispatch).  

Error recovery is very limited in the case that a migration errors or times out (exceeds its `max_steps`). Currently the runtime dev can decide in `FailedMigrationHandler::failed` how to handle this. One follow-up would be to pair this with the `SafeMode` pallet and enact safe mode when an upgrade fails, to allow governance to rescue the chain. This is currently not possible, since governance is not `Mandatory`.  

## Runtime API

- `Core`: `initialize_block` now returns `ExtrinsicInclusionMode` to inform the Block Author whether they can push transactions.

### Integration

Add it to your runtime implementation of `Core` and `BlockBuilder`:
```patch
diff --git a/runtime/src/lib.rs b/runtime/src/lib.rs
@@ impl_runtime_apis! {
	impl sp_block_builder::Core<Block> for Runtime {
-		fn initialize_block(header: &<Block as BlockT>::Header) {
+		fn initialize_block(header: &<Block as BlockT>::Header) -> RuntimeExecutiveMode {
			Executive::initialize_block(header)
		}

		...
	}
```

# 2.) `poll` hook

A new pallet hook is introduced: `poll`. `Poll` is intended to replace mostly all usage of `on_initialize`.  
The reason for this is that any code that can be called from `on_initialize` cannot be migrated through an MBM. Currently there is no way to statically check this;  the implication is to use `on_initialize` as rarely as possible.  
Failing to do so can result in broken storage invariants.

The implementation of the poll hook depends on the `Runtime API` changes that are explained above.

# 3.) Hard-Deadline callbacks

Three new callbacks are introduced and configured on `System::Config`: `PreInherents`, `PostInherents` and `PostTransactions`.  
These hooks are meant as replacement for `on_initialize` and `on_finalize` in cases where the code that runs cannot be moved to `poll`.  
The reason for this is to make the usage of HD-code (hard deadline) more explicit - again to prevent broken invariants by MBMs.  

# 4.) FRAME (general changes)

## `frame_system` pallet

A new memorize storage item `InherentsApplied` is added. It is used by executive to track whether inherents have already been applied. Executive and can then execute the MBMs directly between inherents and transactions.

The `Config` gets five new items:
- `SingleBlockMigrations` this is the new way of configuring migrations that run in a single block. Previously they were defined as last generic argument of `Executive`. This shift is brings all central configuration about migrations closer into view of the developer (migrations that are configured in `Executive` will still work for now but is deprecated). 
- `MultiBlockMigrator` this can be configured to an engine that drives MBMs. One example would be the `pallet_migrations`. Note that this is only the engine; the exact MBMs are injected into the engine.
- `PreInherents` a callback that executes after `on_initialize` but before inherents.
- `PostInherents` a callback that executes after all inherents ran (including MBMs and `poll`).
- `PostTransactions` in symmetry to `PreInherents`, this one is called before `on_finalize` but after all transactions.

A sane default is to set all of these to `()`. Example diff suitable for any chain:  
```patch
@@ impl frame_system::Config for Test {
 	type MaxConsumers = ConstU32<16>;
+	type SingleBlockMigrations = ();
+	type MultiBlockMigrator = ();
+	type PreInherents = ();
+	type PostInherents = ();
+	type PostTransactions = ();
 }
```

An overview of how the block execution now looks like is here. The same graph is also in the rust doc.

<details><summary>Block Execution Flow</summary>
<p>

Screenshot 2023-12-04 at 19 11 29

</p>
</details> 

## Inherent Order

Moved to https://github.com/paritytech/polkadot-sdk/pull/2154

---------------

## TODO

- [ ] Check that `try-runtime` still works
- [ ] Ensure backwards compatibility with old Runtime APIs
- [x] Consume weight correctly
- [x] Cleanup


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3869956

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3874395

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3877338

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3877765

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3879367

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+				return Some(ControlFlow::Continue(cursor))
+			},
+			Err(SteppedMigrationError::InsufficientWeight { required }) => {
+				if is_first || required.any_gt(meter.limit()) {
+					Self::deposit_event(Event::MigrationFailed { index: cursor.index, blocks });
+					Self::upgrade_failed(Some(cursor.index));
+					None
+				} else {
+					// Hope that it gets better in the next block.
+					Some(ControlFlow::Break(cursor))
```
liamaharon, line 543: This seems a bit weird that `InsufficientWeight` could be emitted when `required < meter.limit()`? Does that indicate a buggy migration?

substrate/frame/migrations/src/mock.rs
```diff
@@ -0,0 +1,297 @@
...
+impl Into<u8> for MockedMigrationKind {
+	fn into(self) -> u8 {
+		match self {
+			SucceedAfter => 0,
+			FailAfter => 1,
+			TimeoutAfter => 2,
+			HightWeightAfter(_) => 3,
+		}
+	}
+}
```
liamaharon, line 94: don't we get `Into` for free by implementing `From`?

substrate/frame/migrations/src/mock.rs
```diff
@@ -0,0 +1,297 @@
...
+#[allow(dead_code)]
+pub enum MockedMigrationKind {
+	/// Succeed after its number of steps elapsed.
+	SucceedAfter,
+	/// Fail after its number of steps elapsed.
+	FailAfter,
+	/// Never terminate.
+	TimeoutAfter,
+	/// Cause an [`InsufficientWeight`] error after its number of steps elapsed.
+	HightWeightAfter(Weight),
```
liamaharon, line 69:
```suggestion
	HeightWeightAfter(Weight),
```

substrate/frame/migrations/src/mock.rs
```diff
@@ -0,0 +1,297 @@
...
+pub fn test_closure<R>(f: impl FnOnce() -> R) -> R {
+	let mut ext = new_test_ext();
+	ext.execute_with(f)
+}
+
+pub fn run_to_block(n: u32) {
+	while System::block_number() < n as u64 {
+		if System::block_number() > 1 {
+			Migrations::on_finalize(System::block_number());
+			System::on_finalize(System::block_number());
```
liamaharon, line 219: I thought `on_finalize` is called after migrations every block?

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,369 @@
...
+			Event::MigrationCompleted { index: 0, blocks: 1 },
+			Event::MigrationAdvanced { index: 1, blocks: 0 },
+			Event::MigrationCompleted { index: 1, blocks: 1 },
+			Event::MigrationAdvanced { index: 2, blocks: 0 },
+			Event::MigrationAdvanced { index: 2, blocks: 1 },
+			Event::MigrationCompleted { index: 2, blocks: 2 },
+			Event::UpgradeCompleted,
+		]);
+	});
+}
```
liamaharon, line 91: duplicate tests 
```suggestion
fn simple_works() {
	use Event::*;
	test_closure(|| {
		// Add three migrations, each taking one block longer than the last.
		MigrationsStorage::set(vec![(SucceedAfter, 0), (SucceedAfter, 1), (SucceedAfter, 2)]);

		System::set_block_number(1);
		Migrations::on_runtime_upgrade();
		run_to_block(10);

		// Check that the executed migrations are recorded in `Historical`.
		assert_eq!(
			historic(),
			vec![
				mocked_id(SucceedAfter, 0),
				mocked_id(SucceedAfter, 1),
				mocked_id(SucceedAfter, 2),
			]
		);
		// Check that we got all events.
		assert_events(vec![
			UpgradeStarted { migrations: 3 },
			MigrationCompleted { index: 0, blocks: 1 },
			MigrationAdvanced { index: 1, blocks: 0 },
			MigrationCompleted { index: 1, blocks: 1 },
			MigrationAdvanced { index: 2, blocks: 0 },
			MigrationAdvanced { index: 2, blocks: 1 },
			MigrationCompleted { index: 2, blocks: 2 },
			UpgradeCompleted,
		]);
	});
}
```

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,369 @@
...
+
+#[test]
+fn failing_migration_keep_stuck_the_chain() {
```
liamaharon, line 94: maybe this is better? 
```suggestion
fn failing_migration_sets_cursor_to_stuck() {
```

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,369 @@
...
+		assert_eq!(UpgradesStarted::take(), 1);
+		assert_eq!(UpgradesCompleted::take(), 0);
+		assert_eq!(UpgradesFailed::take(), vec![Some(0)]);
+
+		assert_eq!(Cursor::<T>::get(), Some(MigrationCursor::Stuck), "Must stuck the chain");
+	});
+}
+
+#[test]
+fn failing_migration_force_unstuck_the_chain() {
```
liamaharon, line 124: maybe this is better? 
```suggestion
fn failing_migration_force_unstuck_works() {
```

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,369 @@
...
+		MigrationsStorage::set(vec![(SucceedAfter, 10)]);
+
+		System::set_block_number(1);
+		Migrations::on_runtime_upgrade();
+		run_to_block(3);
+
+		//assert_eq!( // TODO
+		//	historic(),
+		//	vec![mocked_id(SucceedAfter, 0)]
+		//);
```
liamaharon, line 319: 👀

substrate/frame/support/src/migrations.rs
```diff
@@ -344,3 +345,406 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+
+	/// The unique identifier type of this migration.
+	type Identifier: codec::FullCodec + codec::MaxEncodedLen;
+
+	/// The unique identifier of this migration.
+	///
+	/// If two migrations have the same identifier, then they are assumed to be identical.
+	fn id() -> Self::Identifier;
+
+	/// The maximum number of steps that this migration can take at most.
```
liamaharon, line 362:
```suggestion
	/// The maximum number of steps that this migration can take.
```

substrate/frame/support/src/migrations.rs
```diff
@@ -344,3 +345,406 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+	///
+	/// This can be used to enforce progress and prevent migrations to be stuck forever. A migration
```
liamaharon, line 364:
```suggestion
	/// This can be used to enforce progress and prevent migrations becoming stuck forever. A migration
```

substrate/frame/support/src/migrations.rs
```diff
@@ -344,3 +345,406 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+	}
+
+	fn nth_transactional_step(
+		_: u32,
+		cursor: Option<Vec<u8>>,
+		meter: &mut WeightMeter,
+	) -> Option<Result<Option<Vec<u8>>, SteppedMigrationError>> {
+		Some(
+			// FAIL-CI unwrap
+			T::transactional_step(cursor.map(|c| Decode::decode(&mut &c[..]).unwrap()), meter)
```
liamaharon, line 536: is unwrap ok here?

liamaharon (commented):
Really great pallet docs 👌

Just a few small comments

ggwpez:
> > Executive now additionally takes an MultiStepMigrator trait that defaults to ().
> > IT IS PARAMOUNT TO SET THIS TO THE MIGRATIONS PALLET WHEN YOU DEPLOY IT.
> 
> Can we enforce this somehow?

I think so. I will try to add a check into the `integrity_test`, otherwise this is a huge footgun.

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -1948,6 +1948,20 @@ impl pallet_statement::Config for Runtime {
...
 
+parameter_types! {
+	pub MbmServiceWeight: Weight = Perbill::from_percent(80) * RuntimeBlockWeights::get().max_block;
+}
+
+impl pallet_migrations::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Migrations = ();
+	type CursorMaxLen = ConstU32<65_536>;
+	type IdentifierMaxLen = ConstU32<256>;
```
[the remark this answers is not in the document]
ggwpez: Pretty random honestly.  
The only considerations on the upper bound are the PoV since only one of these will be accessed per block. It should therefore be okay to use something like 4MiB.  
Same for the Identifier. For the lower bound, we only assume that the cursor can hold a random key of a map and these map keys are normally not that large. If needed, it can be increased at any time by the runtime dev.

substrate/frame/executive/src/tests.rs
```diff
@@ -0,0 +1,1174 @@
...
+
+		fn offchain_worker(n: BlockNumberFor<T>) {
+			assert_eq!(BlockNumberFor::<T>::from(1u32), n);
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		pub fn some_function(origin: OriginFor<T>) -> DispatchResult {
+			// NOTE: does not make any different.
```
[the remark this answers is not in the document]
ggwpez: I have no clue what this is doing, the code was just moved...

substrate/frame/migrations/Cargo.toml
```diff
@@ -0,0 +1,63 @@
+[package]
+name = "pallet-migrations"
+version = "1.0.0"
+description = "FRAME pallet to execute multi-block migrations."
+authors = ["Parity Technologies <admin@parity.io>"]
+homepage = "https://substrate.io"
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/paritytech/substrate"
```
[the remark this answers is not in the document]
ggwpez: Yep true, it was just copied from the substrate repo.

substrate/frame/support/src/migrations.rs
ggwpez: No its not, so i put the `FAIL-CI`. This makes the CI fail as a reminder, but thanks for commenting 😄

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3892216

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+		///
+		/// A good default needs to selected such that no migration will ever have a cursor with MEL
+		/// above this limit. This is statically checked (FAIL-CI: TODO).
+		#[pallet::constant]
+		type CursorMaxLen: Get<u32>;
+
+		/// The maximal length of an encoded identifier.
+		///
+		/// A good default needs to selected such that no migration will ever have an identifier
+		/// with MEL above this limit. This is statically checked (FAIL-CI: TODO).
```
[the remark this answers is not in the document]
ggwpez: Its checked now.

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+		/// All the multi-block migrations to run.
+		///
+		/// Should only be updated in a runtime-upgrade once all the old migrations have completed.
+		/// (Check that `Cursor` is `None`).
+		type Migrations: SteppedMigrations;
+
+		/// The maximal length of an encoded cursor.
+		///
+		/// A good default needs to selected such that no migration will ever have a cursor with MEL
+		/// above this limit. This is statically checked (FAIL-CI: TODO).
```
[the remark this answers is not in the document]
ggwpez: checked now

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+//!
+//! Migrations are provided to the pallet through the associated type [`Config::Migrations`] of type
+//! `Get<Vec<Box<dyn SteppedMigration<…`. This was done to have the most flexibility when it comes
+//! to iterating and inspecting the migrations. It also simplifies the trait bounds since all
+//! associated types of the trait must be provided by the pallet.  
+//! The actual progress of the pallet is stored in its [`Cursor`] storage item. This can either be
+//! [`MigrationCursor::Active`] or `Stuck`. In the active case it points to the currently active
+//! migration and stores its inner cursor. The inner cursor can then be used by the migration to
+//! store its inner state and advance. Each time when the migration returns `Some(cursor)`, it
+//! signals the pallet that it is not done yet.
```
juangirini, line 69:
```suggestion
//! [`MigrationCursor::Active`] or [`MigrationCursor::Stuck`]. In the active case it points to the
//! currently active migration and stores its inner cursor. The inner cursor can then be used by the
//! migration to store its inner state and advance. Each time when the migration returns
//! `Some(cursor)`, it signals the pallet that it is not done yet.  
```

juangirini: This is just to add the link to `Stuck`

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+//! migrations are skipped without causing an error. Each successfully executed migration is added
+//! to this set (Goal 5).  
+//! This proceeds until no more migrations remain. At that point, the event `UpgradeCompleted` is
+//! emitted (Goal 1).  
+//! The execution of each migration happens by calling [`SteppedMigration::transactional_step`].
+//! This function wraps the inner `step` function into a transactional layer to allow rollback in
+//! the error case (Goal 6).  
+//! Weight limits must be checked by the migration itself. The pallet provides a [`WeightMeter`] for
+//! that purpose. The pallet may return [`SteppedMigrationError::InsufficientWeight`] at any point.
+//! In that scenario, the one of two things will happen: if that migration was exclusively executed
```
juangirini, line 88:
```suggestion
//! In that scenario, one of two things will happen: if that migration was exclusively executed
```

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+//! [`max_steps`][SteppedMigration::max_steps], are error tolerant, check their weight bounds and
+//! have a unique identifier. 2. The runtime upgrade is enacted. `UpgradeStarted` events are
+//! followed by lots of `MigrationAdvanced` events. Finally `UpgradeCompleted` is emitted.  
+//! 3. Cleanup as described in the governance scenario be executed at any time after the migration
+//! completes.
+//!
+//! ### Advice: Failed upgrades
+//!
+//! Failed upgrades cannot recovered from automatically and require governance intervention. Set up
+//! monitoring for `UpgradeFailed` events to be made aware of any failures. The hook
```
juangirini, line 115:
```suggestion
//! Failed upgrades cannot be recovered from automatically and require governance intervention. Set
//! up monitoring for `UpgradeFailed` events to be made aware of any failures. The hook
```

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4135625

substrate/bin/node/runtime/Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
-authors.workspace = true
+authors = ["Parity Technologies <admin@parity.io>"]
 description = "Substrate node kitchensink runtime."
-edition.workspace = true
+edition = "2021"
 build = "build.rs"
 license = "Apache-2.0"
 homepage = "https://substrate.io"
-repository.workspace = true
+repository = "https://github.com/paritytech/substrate/"
```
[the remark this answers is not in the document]
ggwpez: thanks, fixed!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4245902

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4247433

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4247650

substrate/frame/support/src/migrations.rs
ggwpez: fixed

substrate/frame/migrations/src/weights.rs
```diff
@@ -0,0 +1,139 @@
+
```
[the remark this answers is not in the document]
ggwpez: oh yea, need to call the bot.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4286359

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4287298

substrate/client/block-builder/src/lib.rs
```diff
@@ -255,6 +255,15 @@ where
 		})
 	}
 
+	/// Called after inherents but before extrinsics have been applied.
```
gavofyork, line 258:
```suggestion
	/// Called after inherents but before transactions have been applied.
```

inherents are a type of extrinsic. extrinsic simply means "data from the outside".

gavofyork (approved):
Design and docs look good. Haven't done a line-by-line check of the changes. Some invalid usage of extrinsic & transaction terms.

ggwpez:
bot bench substrate-pallet --pallet=pallet_migrations

substrate/frame/executive/src/tests.rs
```diff
@@ -0,0 +1,1297 @@
...
+			.base_block(Weight::from_parts(10, 0))
+			.for_class(DispatchClass::all(), |weights| weights.base_extrinsic = Weight::from_parts(5, 0))
+			.for_class(DispatchClass::non_mandatory(), |weights| weights.max_total = Weight::from_parts(1024, u64::MAX).into())
+			.build_or_panic();
+	pub const DbWeight: RuntimeDbWeight = RuntimeDbWeight {
+		read: 10,
+		write: 100,
+	};
+}
+impl frame_system::Config for Runtime {
```
juangirini, line 304: should `derive_impl` be used here?

substrate/frame/executive/src/tests.rs
```diff
@@ -0,0 +1,1297 @@
...
+	type MaxConsumers = ConstU32<16>;
+	type SingleBlockMigrations = ();
+	type MultiBlockMigrator = ();
+	type PreInherents = ();
+	type PostInherents = ();
+	type PostTransactions = ();
+}
+
+type Balance = u64;
+impl pallet_balances::Config for Runtime {
```
juangirini, line 336: and here?

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+		meter.consumed()
+	}
+
+	/// Try to make progress on the current migration.
+	///
+	/// Returns whether processing should continue or break for this block. The return value means:
+	/// - `None`: The migration process is completely finished.
+	/// - `ControllFlow::Break`: We made *some* progress. Continue in the *next* block with the
+	///   cursor.
+	/// - `ControllFlow::Continue`: We made *no* progress. Continue in the *current* block with the
```
franciscoaguirre, line 612:
```suggestion
	/// - `ControlFlow::Break`: We made *some* progress. Continue in the *next* block with the
	///   cursor.
	/// - `ControlFlow::Continue`: We made *no* progress. Continue in the *current* block with the
```

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,367 @@
...
+			Event::MigrationCompleted { index: 0, took: 1 },
+			Event::MigrationAdvanced { index: 1, took: 0 },
+			Event::MigrationCompleted { index: 1, took: 1 },
+			Event::MigrationAdvanced { index: 2, took: 0 },
+			Event::MigrationAdvanced { index: 2, took: 1 },
+			Event::MigrationCompleted { index: 2, took: 2 },
+			Event::UpgradeCompleted,
+		]);
+	});
+}
```
franciscoaguirre, line 93: Isn't this the same as the `simple_works` test?

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,367 @@
...
+		assert_eq!(UpgradesFailed::take(), vec![Some(0)]);
+
+		assert!(Cursor::<T>::get().is_none(), "Must unstuck the chain");
+	});
+}
+
+/// A migration that reports not getting enough weight errors if it is the first one to run in that
+/// block.
+#[test]
+fn heigh_weight_migration_singular_fails() {
```
franciscoaguirre, line 158:
```suggestion
fn high_weight_migration_singular_fails() {
```

substrate/frame/migrations/src/tests.rs
```diff
@@ -0,0 +1,367 @@
...
+		System::set_block_number(1);
+		Migrations::on_runtime_upgrade();
+		run_to_block(10);
+
+		assert_eq!(historic(), vec![mocked_id(SucceedAfter, 0)]);
+		// Check that we got all events.
+		assert_events::<Event<T>>(vec![
+			Event::UpgradeStarted { migrations: 2 },
+			Event::MigrationCompleted { index: 0, took: 1 },
+			// `blocks=1` means that it was retried once.
```
franciscoaguirre, line 199:
```suggestion
			// `took=1` means that it was retried once.
```

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -25,10 +25,74 @@ use crate::weights::Weight;
...
+#[cfg_attr(all(not(feature = "tuples-96"), not(feature = "tuples-128")), impl_for_tuples(64))]
+#[cfg_attr(all(feature = "tuples-96", not(feature = "tuples-128")), impl_for_tuples(96))]
+#[cfg_attr(feature = "tuples-128", impl_for_tuples(128))]
+impl PostTransactions for Tuple {
+	fn post_transactions() {
+		for_tuples!( #( Tuple::post_transactions(); )* );
+	}
+}
+
+/// Periodically executes logic. Is not guaranteed to run within a specific runtime and should only
```
franciscoaguirre, line 78:
```suggestion
/// Periodically executes logic. Is not guaranteed to run within a specific timeframe and should only
```
?

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -25,10 +25,74 @@ use crate::weights::Weight;
...
+/// be used on logic that has on deadline.
```
franciscoaguirre, line 79:
```suggestion
/// be used on logic that has no deadline.
```

franciscoaguirre (approved):
Awesome work 🚀

juangirini (approved):
Really good stuff

substrate/client/basic-authorship/src/basic_authorship.rs
```diff
@@ -319,11 +319,14 @@ where
...
-		// <https://github.com/paritytech/substrate/pull/14275/>
+		block_builder.last_inherent()?;
 
-		let end_reason =
-			self.apply_extrinsics(&mut block_builder, deadline, block_size_limit).await?;
+		let mode = block_builder.extrinsic_inclusion_mode();
+		let end_reason = match mode {
+			ExtrinsicInclusionMode::AllExtrinsics =>
+				self.apply_extrinsics(&mut block_builder, deadline, block_size_limit).await?,
+			ExtrinsicInclusionMode::OnlyInherents => EndProposingReason::ExtrinsicsForbidden,
```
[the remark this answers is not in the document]
pgherveou: I think this comment explains the terminology well
https://github.com/paritytech/substrate/pull/14414#issuecomment-1602222428

pgherveou:
>  Doing another runtime upgrade while MBMs are ongoing is not a good idea and can lead to messed up state.

Can't we check the active mbm cursor to prevent this?

substrate/frame/support/src/migrations.rs
```diff
@@ -345,3 +348,573 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+	fn nth_max_steps(n: u32) -> Option<Option<u32>>;
+
+	/// Do a [`SteppedMigration::step`] on the `n`th migration.
+	///
+	/// Is guaranteed to return `Some` if `n < Self::len()`.
+	fn nth_step(
+		n: u32,
+		cursor: Option<Vec<u8>>,
+		meter: &mut WeightMeter,
+	) -> Option<Result<Option<Vec<u8>>, SteppedMigrationError>>;
```
pgherveou, line 507: some feedback on our own experience building mbm:

We opted for not taking a meter but having a `fn max_step_weight` that returns the maximum amount of weight a single step can take. This moves the burden of checking whether or not we have enough weight to perform any work to the framework rather than the migration itself

https://github.com/paritytech/polkadot-sdk//blob/6fe7ac4cc8a9004c14873fc477de81b83a1b17dd/substrate/frame/contracts/src/migration.rs#L110-L116

substrate/frame/executive/src/tests.rs
ggwpez: Yea, the code is older than it 🙈

ggwpez:
> > Doing another runtime upgrade while MBMs are ongoing is not a good idea and can lead to messed up state.
> 
> Isn't checking the active mbm cursor sufficient to prevent this?

It cannot really be prevented, since when the runtime code gets overwritten, then the old code is gone. So the governance has to ensure that they dont upgrade when MBMs are still ongoing.  

But yea, maybe we should just reject the `System::set_code` call in this case 🤔

substrate/frame/executive/src/tests.rs
ggwpez: Oh, i did not know that we also have it here, yes nice.

athei:
> But yea, maybe we should just reject the `System::set_code` call in this case 🤔

The danger is that we completely brick ourselves by doing so. What about blocking `System::set_code` as you suggest but adding `System::force_set_code` which still works even if MBM is ongoing?

substrate/frame/support/src/migrations.rs
ggwpez: >We opted for not taking a meter but having a fn max_step_weight that returns the maximum amount of weight a single step can take

I am not sure about this since it makes it less flexible. The migration may decide to make larger steps when there is more weight available to be more efficient.  
Yes the code in the migration has not to do this task, but it also has more freedom of how to use the weight...

ggwpez:
> > But yea, maybe we should just reject the `System::set_code` call in this case 🤔
> 
> The danger is that we completely brick ourselves by doing so. What about blocking `System::set_code` as you suggested but adding `System::force_set_code` which still works even if MBM is ongoing?

Yes, good observation 😅 I think there is already a `set_code_without_checks` that we could use to bypass.  
Sudo should be able to force-stop the Migration process anyway by calling `Migrations::force_set_cursor(None)` to unstuck the chain.

liamaharon (approved):

substrate/frame/support/src/migrations.rs
```diff
@@ -345,3 +348,574 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+	/// migration that exceeds its max steps is treated as failed. `None` means that there is no
+	/// limit.
+	fn max_steps() -> Option<u32> {
+		None
+	}
+
+	/// Try to migrate as much as possible with the given weight.
+	///
+	/// **ANY STORAGE CHANGES MUST BE ROLLED-BACK BY THE CALLER UPON ERROR.** This is necessary
+	/// since the caller cannot return a cursor in the error case. `Self::transactional_step` is
```
bkchr, line 377: Why can it not return a cursor in the error case?

bkchr (commented):
Generally looking good, but still some things needs to be improved.

substrate/frame/system/src/lib.rs
```diff
@@ -448,6 +461,14 @@ pub mod pallet {
 
 		/// The maximum number of consumers allowed on a single account.
 		type MaxConsumers: ConsumerLimits;
+
+		type SingleBlockMigrations: OnRuntimeUpgrade;
+
+		type MultiBlockMigrator: MultiStepMigrator;
```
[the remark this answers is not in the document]
ggwpez: Yea indeed. Currently i wanted to have a grace period where it is possible to configure the migrations on either `Executive` or `System`. But maybe we can do it in one since it is already a breaking change.

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+
+			Ok(())
+		}
+
+		/// Forces the onboarding of the migrations.
+		///
+		/// This process happens automatically on a runtime upgrade. It is in place as an emergency
+		/// measurement. The cursor needs to be `None` for this to suceed.
+		#[pallet::call_index(2)]
+		pub fn force_onboard_mbms(origin: OriginFor<T>) -> DispatchResult {
```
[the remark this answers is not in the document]
ggwpez: For restarting some failed migrations after manual intervention. All these `force_` functions should hopefully never be needed...

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -30,12 +30,14 @@ pub fn expand_outer_inherent(
 ) -> TokenStream {
 	let mut pallet_names = Vec::new();
 	let mut pallet_attrs = Vec::new();
+	let mut pallet_indices = Vec::new();
```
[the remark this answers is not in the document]
ggwpez: Oh yea, it was a left over from this https://github.com/paritytech/polkadot-sdk/pull/2154

substrate/frame/migrations/src/mock.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/// Returns the historic migrations, sorted by their identifier.
+pub fn historic() -> Vec<MockedIdentifier> {
+	let mut historic = Historic::<Test>::iter_keys().collect::<Vec<_>>();
+	historic.sort();
+	historic
+}
+
+// Traits to make using events less insufferable:
```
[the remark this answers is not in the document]
ggwpez: Its just as divider in the file in this case 😄

substrate/test-utils/runtime/src/lib.rs
```diff
@@ -1205,6 +1216,7 @@ mod tests {
 		})
 	}
 
+	#[cfg(not(feature = "disable-genesis-builder"))]
```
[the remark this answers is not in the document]
ggwpez: 🙈

substrate/test-utils/runtime/src/lib.rs
```diff
@@ -26,10 +26,11 @@ pub mod genesismap;
 pub mod substrate_test_pallet;
 
 use codec::{Decode, Encode};
+#[cfg(not(feature = "disable-genesis-builder"))]
```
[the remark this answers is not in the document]
ggwpez: Uhg, maybe the code is older than that... I remember something not compiling.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4416005

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4416039

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4416444

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4416743

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4416709

substrate/bin/node/runtime/src/lib.rs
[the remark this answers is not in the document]
ggwpez: There is now a check in `integrity_test` that should prevent most misconfigurations.

substrate/frame/migrations/src/lib.rs
ggwpez: There is the case that another migration already finished in the current block and ate up ie. 50% of the block.  
Then the second migration realizes that it does not have enough weight and errors, since it needs 75% (as example).

substrate/frame/migrations/src/mock.rs
ggwpez: Generally yes, but in this case these are not complementing. `From<u8>` provides `Into<MockedMigrationKind>`, not `Into<u8>` 😛

substrate/frame/migrations/src/mock.rs
ggwpez: Yea I moved it down now. Normally we do it the other way around (i think for the genesis case) but here it does not matter.

substrate/primitives/block-builder/src/lib.rs
```diff
@@ -48,5 +48,8 @@ sp_api::decl_runtime_apis! {
 
 		/// Check that the inherents are valid. The inherent data will vary from chain to chain.
 		fn check_inherents(block: Block, data: InherentData) -> CheckInherentsResult;
+
+		/// Called after inherents but before transaction application.
+		fn last_inherent();
```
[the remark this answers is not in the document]
ggwpez: It got removed 🙏

substrate/frame/executive/src/lib.rs
```diff
@@ -518,7 +593,8 @@ where
 		}
 	}
 
-	fn initial_checks(block: &Block) {
+	/// Returns the index of the first extrinsic in the block.
```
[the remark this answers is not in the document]
ggwpez: Okay. But its the same thing 😛

substrate/frame/executive/src/lib.rs
```diff
@@ -611,6 +745,13 @@ where
 	/// This doesn't attempt to validate anything regarding the block, but it builds a list of uxt
 	/// hashes.
 	pub fn apply_extrinsic(uxt: Block::Extrinsic) -> ApplyExtrinsicResult {
+		Self::apply_extrinsic_with_mode(uxt, Self::extrinsic_mode())
+	}
+
+	pub fn apply_extrinsic_with_mode(
```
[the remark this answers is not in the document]
ggwpez: Made it private now.

substrate/frame/executive/src/lib.rs
```diff
@@ -628,6 +769,13 @@ where
 
 		// Decode parameters and dispatch
 		let dispatch_info = xt.get_dispatch_info();
+		if dispatch_info.class != DispatchClass::Mandatory &&
+			mode == ExtrinsicInclusionMode::OnlyInherents
+		{
+			// The block builder respects this by using the mode returned by `last_inherent`.
+			panic!("Only Mandatory extrinsics are allowed during Multi-Block-Migrations");
+		}
```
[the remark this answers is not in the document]
ggwpez: Its good for testing, but i can remove it.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4424158

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4424157

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -25,10 +25,74 @@ use crate::weights::Weight;
...
+impl PreInherents for Tuple {
+	fn pre_inherents() {
+		for_tuples!( #( Tuple::pre_inherents(); )* );
+	}
+}
+
+/// Provides a callback to execute logic after the all inherents.
+pub trait PostInherents {
+	/// Called after all inherents were applied.
+	fn post_inherents() {}
```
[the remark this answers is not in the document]
ggwpez: It is supposed to replace `on_finalize`. By having it more explicit in the System config it hopefully makes it harder to accidentally use it without knowing that it is a hard-deadline hook.

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -25,10 +25,74 @@ use crate::weights::Weight;
 use impl_trait_for_tuples::impl_for_tuples;
 use sp_runtime::traits::AtLeast32BitUnsigned;
 use sp_std::prelude::*;
+use sp_weights::WeightMeter;
 
 #[cfg(feature = "try-runtime")]
 use sp_runtime::TryRuntimeError;
 
+/// Provides a callback to execute logic before the all inherents.
```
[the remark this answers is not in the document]
ggwpez: Yea i forgot it. There is a mermaid doc on `Executive::block_flowchart` now:  
mermaid-diagram-2023-11-21-143719

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4428630

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4429072

substrate/frame/support/src/migrations.rs
```diff
@@ -345,3 +348,574 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+	/// provided as convenience for a caller. A cursor of `None` implies that the migration is at
+	/// its end.
+	// TODO: Think about iterator `fuse` requirement.
```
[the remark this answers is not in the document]
ggwpez: No, it wanted to still fix it in this MR.  
I added a comment now that mentions that a migration does not need to be `fuse` and is guaranteed to not be called ever again after returning `None` once.

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -244,12 +244,12 @@ async fn follow_with_runtime() {
 	let event: FollowEvent<String> = get_next_event(&mut sub).await;
 
 	// it is basically json-encoded substrate_test_runtime_client::runtime::VERSION
-	let runtime_str = "{\"specName\":\"test\",\"implName\":\"parity-test\",\"authoringVersion\":0,\
-		\"specVersion\":2,\"implVersion\":2,\"apis\":[[\"0xdf6acb689907609b\",4],\
+	let runtime_str = "{\"specName\":\"test\",\"implName\":\"parity-test\",\"authoringVersion\":1,\
```
ggwpez, line 247: I dont remember bumping `authoringVersion`?

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,718 @@
...
+		mut cursor: ActiveCursorOf<T>,
+		is_first: bool,
+		meter: &mut WeightMeter,
+	) -> Option<ControlFlow<ActiveCursorOf<T>, ActiveCursorOf<T>>> {
+		let Some(id) = T::Migrations::nth_id(cursor.index) else {
+			// No more migration in the tuple - we are done.
+			defensive_assert!(cursor.index == T::Migrations::len(), "Inconsitent MBMs tuple");
+			Self::deposit_event(Event::UpgradeCompleted);
+			Cursor::<T>::kill();
+			T::MigrationStatusHandler::completed();
```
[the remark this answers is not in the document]
ggwpez: Will still fix this. I think we can just include it in a benchmark of `exec_migration`.

substrate/frame/support/src/migrations.rs
ggwpez: But it probably is a good idea to add a `min_weight` to the interface to ensure that it at least can always make *some* progress.

pgherveou: > The migration may decide to make larger steps when there is more weight available to be more efficient

You will need to check the weight regardless no? So why not moving this responsibility to the framework?

In our case the migration framework verify there is enough weight before running a new step and do that in a loop until all the weight is exhausted

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4506873

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4507174

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4558853

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4560406

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4561788

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4594034

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,750 @@
...
+
+		/// The maximum weight to spend each block to execute migrations.
+		type MaxServiceWeight: Get<Weight>;
+
+		/// Weight information for the calls and functions of this pallet.
+		type WeightInfo: WeightInfo;
+
+		// Helper config for setting up benchmarking.
+		//#[cfg(feature = "runtime-benchmarks")]
+		//type BenchmarkSetup: BenchmarkSetup;
```
bkchr, line 327:
```suggestion
```

substrate/frame/migrations/src/lib.rs
```diff
@@ -0,0 +1,750 @@
...
+				Self::deposit_event(Event::MigrationAdvanced { index: cursor.index, took });
+				cursor.inner_cursor = Some(bound_next_cursor);
+
+				if max_steps.map_or(false, |max| took > max.into()) {
+					Self::deposit_event(Event::MigrationFailed { index: cursor.index, took });
+					Self::upgrade_failed(Some(cursor.index));
+					None
+				} else {
+					// A migration cannot progress more than one step per block, we therefore break.
+					Some(ControlFlow::Break(cursor))
```
bkchr, line 693: Why? If there is still space in the block, why not continue?

bkchr (approved):
Great work!

I left some last things. In general it would be nice to remove all the blocks of:
```
	type SingleBlockMigrations = ();
	type MultiBlockMigrator = ();
	type PreInherents = ();
	type PostInherents = ();
	type PostTransactions = ();
```

And use `derive_impl` for them if possible.

I also left some message about introducing some new `ExtriniscPhase`. This makes it better for the outside to find out when a certain kind of event happened etc.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4825626

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4826432

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4826433

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4826972

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827147

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827269

substrate/client/block-builder/src/lib.rs
```diff
@@ -243,23 +244,34 @@ where
...
 			version,
 			estimated_header_size,
 			call_api_at,
+			extrinsic_inclusion_mode,
 		})
 	}
 
+	/// The extrinsic inclusion mode of the runtime for this block.
+	pub fn extrinsic_inclusion_mode(&self) -> ExtrinsicInclusionMode {
+		self.extrinsic_inclusion_mode
```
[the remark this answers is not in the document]
ggwpez: You mean i should add a check in `Executive: apply_extrinsic` and reject non-mandatory extrinsics when there are MBMs ongoing?  
Yea can do.

substrate/frame/executive/src/lib.rs
```diff
@@ -551,53 +636,90 @@ where
...
-			// any initial checks
-			Self::initial_checks(&block);
+			Self::apply_extrinsics(extrinsics.into_iter());
+			<frame_system::Pallet<System>>::note_finished_extrinsics();
 
-			// execute extrinsics
-			let (header, extrinsics) = block.deconstruct();
-			Self::execute_extrinsics_with_book_keeping(extrinsics, *header.number());
+			// In this case there were no transactions to trigger this state transition:
+			if !<frame_system::Pallet<System>>::inherents_applied() {
```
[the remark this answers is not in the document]
ggwpez: Yea it would make more sense. We currently have:  
```rust
pub enum Phase {
	/// Applying an extrinsic.
	ApplyExtrinsic(u32),
	/// Finalizing the block.
	Finalization,
	/// Initializing the block.
	Initialization,
}
```
So i guess we could add an `ApplyInherent(u32)` as first phase after `Initialization`?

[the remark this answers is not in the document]
bkchr: Yes and `AfterInherent`

substrate/client/block-builder/src/lib.rs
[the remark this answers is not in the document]
bkchr: Ignore my comment.

bkchr:
@ggwpez what is this blocked on? And could you also add some default types, so you don't need to touch every file that implements the `system::Config`?

substrate/frame/migrations/src/lib.rs
ggwpez: Yes we could continue here. I will probably change it, there is just the assumption that each migration already does maximal progress when called with a specific weight, so it should be a No-OP to call it again with less weight.

ggwpez:
> @ggwpez what is this blocked on?

Noting, its just waiting for audit.

>And could you also add some default types, so you don't need to touch every file that implements the `system::Config`?

But i did? I removed the clutter in this commit https://github.com/paritytech/polkadot-sdk/pull/1781/commits/642a83bf7cc85122e6370d66a67dc9e5cd55c308

bkchr:
> But i did? I removed the clutter in this commit [642a83b](https://github.com/paritytech/polkadot-sdk/commit/642a83bf7cc85122e6370d66a67dc9e5cd55c308)

Hmm :D Github was showing me some wrong diff :D Ty! 



> Noting, its just waiting for audit

Hmm okay...

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5239386

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5361022

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5380089

acatangiu (approved):

ggwpez:
> I just noticed an error in block production after upgrading to Substrate with this change, while runtime is still the old one:

So the new node cannot import the blocks anymore?

nazar-pc:
> So the new node cannot author blocks anymore?

Block import works fine, it is just block authoring that doesn't.

> Do you have a branch where i can test it?

You can export `dev` chain spec using latest release https://github.com/subspace/subspace/releases/tag/gemini-3h-2024-mar-29
And then run node from `main` branch with `--dev --chain raw-chain-spec.json` and farmer (precompiled from latest release should be fine if you don't want to spend time compiling) with `--dev --max-pieces-in-sector 100 --reward-address SS58_ADDRESS_HERE path=/path/to,size=1G`.

ggwpez:
Could reproduce, thanks.

ggwpez:
Looks like your definition of the `Core` runtime API got out of sync. It reports V5, while exposing the V4 definition.

Your V5:
https://github.com/subspace/polkadot-sdk/blob/d6b500960579d73c43fc4ef550b703acfa61c4c8/substrate/primitives/api/src/lib.rs#L801-L809

What the node expects as V5: 
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/api/src/lib.rs#L806-L817

Note how the amended `initialize_block` function is missing. I dont see any other change, so going to assume that it was a merge conflict?  
Just FYI the spot where it fails is [here](https://github.com/paritytech/polkadot-sdk/blob/eefd5fe4499515da66d088505093f75b5aa22550/substrate/client/block-builder/src/lib.rs#L253-L259).

nazar-pc:
Could be a merge conflict. Annoyingly Parity team squashes everything together and releases are not additive, so when I merged 1.10.1 into 1.8.0-derives fork there were a bunch of merge conflicts from some basic verison changes to other more tricky issues. I'll check it again, thanks!

nazar-pc:
Here is the commit of the fork we actually use and it looks correct: https://github.com/subspace/polkadot-sdk/blob/ac2f3efb476ee3f5ac6bafefb458e9be158adb7f/substrate/primitives/api/src/lib.rs#L804-L819
Not sure how you decided that we use d6b500960579d73c43fc4ef550b703acfa61c4c8

ggwpez:
That `ac2f3efb476ee3f5ac6bafefb458e9be158adb7f` is the master branch - yes. But you said to take the runtime from the latest release tag `gemini-3h-2024-mar-29`, which uses `d6b500960579d73c43fc4ef550b703acfa61c4c8`.  
So the runtime of the release is wrong, but the one of master is right.

nazar-pc:
I think I see the confusion. The runtime is not wrong, the dev runtime generated by `gemini-3h-2024-mar-29` release is similar to what our live chain uses, I suggested to generate chain spec from it intentionally to check compatibility with new client. Upgraded client with `main` branch using `ac2f3efb476ee3f5ac6bafefb458e9be158adb7f` should still be able to work with older runtime that doesn't have new APIs, this is what versioning exists for. But looks like versioning is not checked properly somewhere in block authoring code and it fails to produce a block due to result decoding error.

nazar-pc:
@ggwpez I think I see where the bug is.

This:
```rust
	#[api_version(5)]
	pub trait Core {
		/// Returns the version of the runtime.
		fn version() -> RuntimeVersion;
		/// Execute the given block.
		fn execute_block(block: Block);
		/// Initialize a block with the given header.
		#[changed_in(5)]
		#[renamed("initialise_block", 2)]
		fn initialize_block(header: &<Block as BlockT>::Header);
		/// Initialize a block with the given header and return the runtime executive mode.
		fn initialize_block(header: &<Block as BlockT>::Header) -> ExtrinsicInclusionMode;
	}
```

And this:
```rust
		let extrinsic_inclusion_mode = if core_version >= 5 {
			api.initialize_block(parent_hash, &header)?
		} else {
			#[allow(deprecated)]
			api.initialize_block_before_version_5(parent_hash, &header)?;
			ExtrinsicInclusionMode::AllExtrinsics
		};
```

Are both probably wrong. I checked and old runtime's version was already at version 5, so if you're making breaking changes to API you should have bumped version to 6 and checked `core_version >= 6`.

Otherwise since my old runtime that returns `()` is already at version 5, the whole thing inevitably breaks.

nazar-pc:
Found it, it was caused by this custom patch in our fork of Substrate that bumped the version :disappointed: : https://github.com/substrate/polkadot-sdk/commit/447bbc765020674614e9ac982163f7e11e5b03ea

tmcgroul:
hi there from Subsquid team! we've discovered that this pr introduces event pointing to non-existing extrinsics which is a "crime" in context of data indexing :)
here is a hack that i came up with to handle multi-block migrations - https://github.com/subsquid/squid-sdk/commit/cbae638b2d6aeb94939844aace01cdd06b46bcb0
i’m wondering if there was another way to implement this feature and similar features in the future. would actual extrinsic be an option? smth like `Migrations.progress` or `Migrations.action`. i also believe that a new phase would help to distinguish migration and non-migration events.

ggwpez:
Hey i personally use Subsquid, nice work!  

What is inconsistency that this introduces? You mean an event references an `extrinsicIndex` but there is no extrinsic?

tmcgroul:
> You mean an event references an `extrinsicIndex` but there is no extrinsic?

yes, exactly this

ggwpez:
Originally we wanted to merge this right after: https://github.com/paritytech/polkadot-sdk/pull/3666  
I have to double-check that it fixes what you describe, but  my memory is that it should.

tmcgroul:
got it! thanks for the inside

ggwpez:
There was a bug found and fixed in the original version of this: https://github.com/paritytech/polkadot-sdk/pull/5695

PR #1782, opened 2023-10-03
Migrates pallet-multisig to Consideration

gupnik:
Part of https://github.com/paritytech/polkadot-sdk/issues/226

In order to simply migrate the existing runtimes, a diff can look like the following:
```diff
-	type DepositBase = DepositBase;
-	type DepositFactor = DepositFactor;
+       type Consideration = HoldConsideration<
+		AccountId,
+		Balances,
+		MultisigHoldReason,
+		LinearStoragePrice<DepositBase, DepositFactor, Balance>,
+	>;
```


Polkadot-Forum:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4831465

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4831768

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4837410

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4838425

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4844310

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4844896

substrate/frame/multisig/src/tests.rs
```diff
@@ -692,3 +719,24 @@ fn multisig_handles_no_preimage_after_all_approve() {
...
+
+		insert_old_multisig::<Test>(&multi, &call_hash, 1);
+		assert_eq!(Multisigs::<Test>::iter().count(), 1);
+
+		assert_eq!(Multisig::ensure_updated(&multi, &call_hash, 3), true);
+		assert_eq!(Multisigs::<Test>::iter().count(), 0);
+		assert_eq!(MultisigsFor::<Test>::iter().count(), 1);
+
+		let updated_multisig = MultisigsFor::<Test>::get(&multi, &call_hash).unwrap();
+		assert_eq!(updated_multisig.depositor, 1);
```
liamaharon, line 740: should we check all fields for sanity?

substrate/frame/multisig/src/lib.rs
```diff
@@ -665,6 +691,25 @@ impl<T: Config> Pallet<T> {
...
+	pub(crate) fn ensure_updated(id: &T::AccountId, call_hash: &[u8; 32], threshold: u16) -> bool {
+		#[allow(deprecated)]
+		let r = match Multisigs::<T>::take(id, call_hash) {
+			Some(r) => r,
+			None => return false,
+		};
+		T::Currency::unreserve(&r.depositor, r.deposit);
+		// take consideration
+		let Ok(ticket) =
+			T::Consideration::new(&r.depositor, Footprint::from_parts(1, threshold as usize))
```
liamaharon, line 704: when migrating, shouldn't we take exactly `r.deposit` instead of calculating using the current config? to handle the case where the reserve amounts changed.

substrate/frame/multisig/src/lib.rs
```diff
@@ -665,6 +691,25 @@ impl<T: Config> Pallet<T> {
...
+				.defensive_proof("Unexpected inability to take deposit after unreserved")
+		else {
+			return true
```
liamaharon, line 707: if this returns early, the new multisig is not added to `MultisigsFor`. should we just defensive log but continue anyways to not break the multisig?

substrate/frame/multisig/src/lib.rs
gupnik: As we are un-reserving the previous amount, isn't it a fair assumption to charge the current amount in a migration?

substrate/frame/multisig/src/lib.rs
gupnik: Well, multisig can't exist in the system if we were not able to take the consideration. I can throw an error in this case though.

substrate/frame/multisig/src/lib.rs
liamaharon: I wouldn't say so, because it's not what the user agreed to / understood at the time that they made the multisig. I would expect deposit amount changes to only impact new multisigs, not be applied retroactively

substrate/frame/multisig/src/lib.rs
liamaharon: Wouldn't it be bad ux for a multisig to just vanish?

This shouldn't be an issue if we don't try re-reserving different amounts, though

gupnik: It won't vanish but the migration would fail hinting the user to add funds if needed.

substrate/frame/multisig/src/lib.rs
gupnik: I see your point, but I also don't think that this is a significant issue. @ggwpez @kianenigma Would love to get your thoughts here. I imagine this was the reason we were trying to add `new_from_exact` API in consideration, but this looks to me as a better trade-off.

substrate/frame/multisig/src/lib.rs
liamaharon: I don't see where it fails, maybe I misunderstand but I see it returns true in this conditional if the consideration fails to create

substrate/frame/multisig/src/lib.rs
liamaharon: Yes this is why we were trying to add `new_from_exact`. I guess adding that failed/is still in progress? :P

ggwpez: https://github.com/paritytech/polkadot-sdk/pull/2274 can be used to migrate them over with an exact amount

substrate/frame/multisig/src/lib.rs
```diff
@@ -118,14 +118,34 @@ where
...
+#[derive(Clone, Eq, PartialEq, Encode, Decode, Default, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(MaxApprovals))]
+pub struct Multisig<BlockNumber, Ticket, AccountId, MaxApprovals>
+where
+	MaxApprovals: Get<u32>,
+{
+	/// The extrinsic when the multisig operation was opened.
+	when: Timepoint<BlockNumber>,
+	/// The amount held in reserve of the `depositor`, to be returned once the operation ends.
+	ticket: Ticket,
```
ggwpez, line 131: I think the name `Ticket` was swapped for `Consideration` or so.

substrate/frame/multisig/src/lib.rs
```diff
@@ -176,14 +184,26 @@ pub mod pallet {
...
 		T::AccountId,
 		Blake2_128Concat,
 		[u8; 32],
-		Multisig<BlockNumberFor<T>, BalanceOf<T>, T::AccountId, T::MaxSignatories>,
+		OldMultisig<BlockNumberFor<T>, BalanceOf<T>, T::AccountId, T::MaxSignatories>,
+	>;
+
+	/// The set of open multisig operations.
+	#[pallet::storage]
+	pub type MultisigsFor<T: Config> = StorageDoubleMap<
```
ggwpez, line 200: This is a breaking change and will probably mess with wallets/indexers etc since they look at the old map.  
I dont think there is an easy way out without MBMs (audit stalled for another month), so not sure what to best do.

substrate/frame/multisig/src/lib.rs
```diff
@@ -558,8 +587,8 @@ impl<T: Config> Pallet<T> {
 
 				// Clean up storage before executing call to avoid an possibility of reentrancy
 				// attack.
-				<Multisigs<T>>::remove(&id, call_hash);
-				T::Currency::unreserve(&m.depositor, m.deposit);
+				<MultisigsFor<T>>::remove(&id, call_hash);
+				let _ = m.ticket.drop(&m.depositor);
```
ggwpez, line 591: Does it panic when its not manually dropped?

substrate/frame/multisig/src/lib.rs
```diff
@@ -665,6 +691,25 @@ impl<T: Config> Pallet<T> {
 		signatories.insert(index, who);
 		Ok(signatories)
 	}
+
+	pub(crate) fn ensure_updated(id: &T::AccountId, call_hash: &[u8; 32], threshold: u16) -> bool {
```
ggwpez, line 695: I think we can safely expose this function publicly as extrinsic to manually nudge it to speed up the process.

substrate/frame/multisig/src/lib.rs
liamaharon: > I think we shouldn't change the amounts on the fly, the user only agreed to the amount that they signed and if they are early adopters of Polkadot we should not punish that for no good reason.

I agree

substrate/frame/multisig/src/lib.rs
liamaharon: TBH since this is not the highest priority I don't think too much of a problem to wait for MBMs.. should try to minimise breakage of these apis as much as possible, as you point out. with this change, possibly every multisig UI in the ecosystem would need updating :P

substrate/frame/multisig/src/lib.rs
gupnik: > I think we shouldn't change the amounts on the fly, the user only agreed to the amount that they signed and if they are early adopters of Polkadot we should not punish that for no good reason.

Thanks - but I do see this pattern already being used in preimage pallet [here](https://github.com/paritytech/polkadot-sdk/blob/f80cfc22595bbb39b50d2bd39661ff9eef74cb60/substrate/frame/preimage/src/lib.rs#L260). Won't that be impacted with a similar reasoning?

substrate/frame/multisig/src/lib.rs
gupnik: Yeah, fine to wait for MBM if that's the case.

substrate/frame/multisig/src/lib.rs
gupnik: You mean if someone forgot to call `drop` on the ticket?

substrate/frame/multisig/src/lib.rs
liamaharon: Yes, I would have argued to not have used that pattern in the preimage pallet.

substrate/frame/multisig/src/lib.rs
shawntabrizi: in general i think `let _` should be a no-no across the repo.

If in this situation, you specifically want to drop the result you should call `mem::drop` (AFAIK)

bkchr:
@gupnik what is the status on this?

gupnik:
> @gupnik what is the status on this?

Yes, it's in my todo list but haven't been able to get back to this yet. Will try to prioritise. Thanks for the reminder.

PR #1792, opened 2023-10-04
Switch to ParityDB by default; deprecate RocksDB

koute:
ParityDB seems to be stable and work pretty well, and it's significantly more efficient than RocksDB, so let's finally switch to it by default and get rid of RocksDB.

This PR disables the ability to create new RocksDB databases and makes ParityDB the default database used.

Existing RocksDB databases are still supported and will still work, but will emit a warning about its deprecation.

Here's how the error looks like when starting a Polkadot node with a `--database rocksdb` where there's no existing RocksDB database present:

```
2023-10-04 18:46:02 Parity Polkadot
2023-10-04 18:46:02 ✌️  version 1.1.0-d95ce90cbc3
2023-10-04 18:46:02 ❤️  by Parity Technologies <admin@parity.io>, 2017-2023
2023-10-04 18:46:02 📋 Chain specification: Development
2023-10-04 18:46:02 🏷  Node name: boring-brick-3060
2023-10-04 18:46:02 👤 Role: AUTHORITY
2023-10-04 18:46:02 💾 Database: RocksDb at /tmp/db/chains/rococo_dev/db/full
Error:
   0: Backend error: Creating new RocksDB databases is not supported anymore; please use ParityDB

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

And here's the deprecation warning if a RocksDB database already exists:

```
2023-10-04 21:57:42 Substrate Node
2023-10-04 21:57:42 ✌️  version 3.0.0-dev-0db50256685
2023-10-04 21:57:42 ❤️  by Parity Technologies <admin@parity.io>, 2017-2023
2023-10-04 21:57:42 📋 Chain specification: Development
2023-10-04 21:57:42 🏷  Node name: utter-name-3738
2023-10-04 21:57:42 👤 Role: AUTHORITY
2023-10-04 21:57:42 💾 Database: RocksDb at /tmp/db/chains/dev/db/full
2023-10-04 21:57:45 💀 RocksDB support is deprecated and will be removed on 2024-04-01; please switch to ParityDB
2023-10-04 21:57:45 [0] 💸 generated 1 npos voters, 1 from validators and 0 nominators
2023-10-04 21:57:45 [0] 💸 generated 1 npos targets
```


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3880029

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3880534

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3880668

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3884841

substrate/client/db/src/upgrade.rs
```diff
@@ -204,13 +204,9 @@ mod tests {
...
 
 	fn open_database(db_path: &Path, db_type: DatabaseType) -> sp_blockchain::Result<()> {
-		crate::utils::open_database::<Block>(
-			&DatabaseSource::RocksDb { path: db_path.to_owned(), cache_size: 128 },
-			db_type,
-			true,
-		)
-		.map(|_| ())
-		.map_err(|e| sp_blockchain::Error::Backend(e.to_string()))
+		crate::utils::open_kvdb_rocksdb::<Block>(db_path, db_type, true, 128)
```
bkchr, line 207: Why not use parity db in this test?

polkadot/tests/purge_chain_works.rs
```diff
@@ -35,26 +35,11 @@ async fn purge_chain_rocksdb_works() {
...
-			.arg(tmpdir.path())
-			.arg("--port")
-			.arg("33034")
-			.arg("--no-hardware-benchmarks")
-			.spawn()
-			.unwrap();
+		// Creating new RocksDB databases is not supported anymore, so let's
+		// just use a preexisting snapshot of a minimal database.
+		let blob = std::io::Cursor::new(include_bytes!("rocksdb_test_database.zip"));
+		zip::read::ZipArchive::new(blob).unwrap().extract(&tmpdir).unwrap();
```
bkchr, line 41: Why not just use `open_kvdb_rocksdb`?

cumulus/polkadot-parachain/src/main.rs
```diff
@@ -26,6 +26,9 @@ mod cli;
 mod command;
 mod rpc;
 
-fn main() -> sc_cli::Result<()> {
-	command::run()
+fn main() {
+	if let Err(error) = command::run() {
```
bkchr, line 30: Why? To use the `Display` implementation? Can we not make `Debug` use `Display` internally?

arkpar:
Thank you for taking action ;-)

We're currently working on a major [feature](https://github.com/paritytech/parity-db/issues/199) in paritydb that will change how state data is stored. That might affect stability. I still think we can make the switch now and roll out this new storage method in a fail safe or experimental manner.

> This PR disables the ability to create new RocksDB databases

This seems a bit harsh. It will break existing setups that people may have for testing, automation, etc. 
I suggest deprecating `--db=rocksdb` with a similar warning message first, disabling it in a subsequent release.

bkchr:
> This seems a bit harsh. It will break existing setups that people may have for testing, automation, etc.

I think it is fine. We want to discourage creating new dbs. You can still use `--database rocksdb` to open an old db. As long as your stuff continues to work in production, I don't see any problem in this.

hitchhooker:
took me a month(september) to sync kusama on paritydb with archive mode on 7950x3d + ecc ram + nvme:s in Bangkok, db constantly stopping sync and sync files being. rocksdb syncs in a day or two.  that major feature update is said to fix this, but might be something to consider. 
sync experience was so painful that i really hope rocksdb aint deprecated too soon

gituser:
@koute 
>ParityDB seems to be stable and work pretty well, and it's significantly more efficient than RocksDB

Can you explain more how it is efficient?
It seems ParityDB takes much more time (slower) compared to RocksDB for initial sync, in terms of storage ParityDB is the same as RocksDB (~300 GB with `--pruning=100000`), ParityDB uses more host memory compared to RocksDB.

RocksDB imo much more stable compared to ParityDB and it shouldn't be removed or deprecated at this stage unless all these issues are fixed first.

bkchr:
@hitchhooker @gituser please open issues in the [parity db repo](https://github.com/paritytech/parity-db/). We need to fix them. However, rocksdb has its own downsides and we don't want to continue to maintain it.

koute:
> took me a month(september) to sync kusama on paritydb with archive mode on 7950x3d + ecc ram + nvme:s in Bangkok, db constantly stopping sync and sync files being. rocksdb syncs in a day or two. that major feature update is said to fix this, but might be something to consider.
> sync experience was so painful that i really hope rocksdb aint deprecated too soon

Are you sure this was due to ParityDB and not a generic syncing issue related to networking? (I have seen e.g. syncing stopping myself too in the past, but that was also on RocksDB and was unrelated to the database used.)

> ParityDB uses more host memory compared to RocksDB

Last time I checked the difference was visibly in favor of ParityDB, unless something changed since that time. From time to time I run performance stress tests where I run a testnet composed of *hundreds* of nodes on a *single* machine, and in that test I can run many more nodes when using ParityDB before I hit the memory limit and everything crashes.

---

Anyway, thank you for your feedback. If you see any problems with ParityDB *please* make sure to open an issue! If people don't report the problems they see we can't fix them.

---

@bkchr Perhaps we should go with a little bit softer deprecation here, just in case there are actually still some showstopper problems related to ParityDB?

How about we add a temporary escape hatch, e.g. `--allow-deprecated-create-new-rocksdb-database` parameter (or something like that) which will allow people to still create new RocksDB databases if absolutely necessary? This will still hopefully push people who have `--database rocksdb` parameter hardcoded in their node invocations to switch to ParityDB as that will still produce an error, but it case there *are* actually any showstopper bugs they'll be able to temporarily switch back until we fix the issue.

(So people will have to have `--allow-deprecated-create-new-rocksdb-database --database rocksdb` to force the node to create a new RocksDB database.)

And if there will be no new issues reported we can just nuke this parameter in the next release or two.

I will also do some testing myself syncing my archive node under ParityDB from scratch a couple of times, just in case.

kogeler:
> > This seems a bit harsh. It will break existing setups that people may have for testing, automation, etc.
> 
> I think it is fine. We want to discourage creating new dbs. You can still use `--database rocksdb` to open an old db. As long as your stuff continues to work in production, I don't see any problem in this.

recently we had had some cases when we needed to sync new DBs from scratch. We couldn't do it using ParityDB. We were forced to use binary DB snapshots.

kogeler:
@koute 

I am absolutely sure that the problem is not in the network or in the machines. We use large instances with huge bandwidth with the fastest disks that the large cloud providers can offer. The issue looks like an overloaded disk subsystem. You can check the screenshots of the load charts in [my issue](https://github.com/paritytech/parity-db/issues/215#issue-1811548769).

arkpar:
Archive DBs are indeed slower at the moment on most VPC/VPS machines. This is mostly because there's a lot of small random writes going on. The database hits VPC or hardware IOPS limit. Every inserted trie node results in a random write 
Having lots of RAM for page cache helps with performance here.  
https://github.com/paritytech/parity-db/issues/199 should resolve this issue by eliminating random writes, but it is a substantial change (mostly on the trie/substrate side). I estimate it to be ready in 4-6 weeks.

kogeler:
> Archive DBs are indeed slower at the moment on most VPC/VPS machines. This is mostly because there's a lot of small random writes going on. The database hits VPC or hardware IOPS limit. Every inserted trie node results in a random write Having lots of RAM for page cache helps with performance here. [paritytech/parity-db#199](https://github.com/paritytech/parity-db/issues/199) should resolve this issue by eliminating random writes, but it is a substantial change (mostly on the trie/substrate side). I estimate it to be ready in 4-6 weeks.

[GCP SSD PD](https://cloud.google.com/compute/docs/disks/performance#performance_limits) has performance parameters:
* Read throughput per VM (MB/s) - 1,200
* Write throughput per VM (MB/s) - 1,200
* Read IOPS per VM - 100,000
* Write IOPS per VM - 80,000

Also I used my bare-metal instances with local SSD disks. It doesn't matter. Parity DB can тще sync current Polkadot and Kusama state archive DBs.

I feel that the ParityDB database is not stable enough to disable RocksDB.

kogeler:
another ref:
https://github.com/paritytech/parity-db/issues/212

kogeler:
Also, I created [the feature request](https://github.com/paritytech/parity-db/issues/219} that is related to the ParityDB snapshots.

It is also relevant because we cannot synchronize the database from scratch and cannot quickly load it from a snapshot in a predictable time at the same time. 

It's why we are switching from ParityDB to RocksDB for some of our critical nodes right now.

hitchhooker:
> Archive DBs are indeed slower at the moment on most VPC/VPS machines. This is mostly because there's a lot of small random writes going on. The database hits VPC or hardware IOPS limit. Every inserted trie node results in a random write Having lots of RAM for page cache helps with performance here. [paritytech/parity-db#199](https://github.com/paritytech/parity-db/issues/199) should resolve this issue by eliminating random writes, but it is a substantial change (mostly on the trie/substrate side). I estimate it to be ready in 4-6 weeks.

My setup is designed to reach 250k IOPS read/write per TB at hardware level, but I think ZFS as a filesystem might be the limiting factor for random writes(about 2.5x slower performance than lvm/ext4 for random writes in benchmark).

arkpar:
ZFS uses 128k IO pages by default, which is suboptimal for DB workloads, especially ones involving small random writes. 
https://openzfs.github.io/openzfs-docs/Performance%20and%20Tuning/Workload%20Tuning.html#database-workloads
For parity-db ZFS `recordsize` option should be set to a minimum. E.g. 4k

hitchhooker:
I think main issue for poor performance is just copy on write. Have been running with 16k recordsize for paritydb without really improving benchmarks. Will give 4k a try.

ggwpez:
I started a Kusama archive sync on GCP server `ref-hw-ggwpez` so that anyone from Parity can SSH into from the web-UI.  
It is currently at block 2M and already doing write 17k IOPS. It has 4 local [NVMEs](https://cloud.google.com/compute/docs/disks/local-ssd#nvme) with a combined 680K/360K read/write IOPS.

bkchr:
> How about we add a temporary escape hatch, e.g. `--allow-deprecated-create-new-rocksdb-database` parameter (or something like that) which will allow people to still create new RocksDB databases if absolutely necessary? This will still hopefully push people who have `--database rocksdb` parameter hardcoded in their node invocations to switch to ParityDB as that will still produce an error, but it case there _are_ actually any showstopper bugs they'll be able to temporarily switch back until we fix the issue.
> 
> (So people will have to have `--allow-deprecated-create-new-rocksdb-database --database rocksdb` to force the node to create a new RocksDB database.)

Good idea! But just let's do it by `--database rocksdb-deprecated-whatever`.

hitchhooker:
what I have noticed that zfs is really terrible. 
image
single nvme with ext4

image
2x these nvme in striped zfs with 4k recordsize

so I doubt that actually the issues with syncing archive paritydb databases are related to use of zfs/btrfs instead just lvm/ext4. I guess copy on write slows down random writes, so probably good idea to recommend to avoid usage of  zfs/btrfs for database disks. I think @kogeler had also some alternative filesystem in use.

timorleph:
Hi, is making ParityDB default something you are still working on? If so, when would you expect this to be done?

We are wondering which DB we should use, and knowing what your plans related to this PR are would be helpful.

bkchr:
> We're currently working on a major [feature](https://github.com/paritytech/parity-db/issues/199) in paritydb that will change how state data is stored. That might affect stability. I still think we can make the switch now and roll out this new storage method in a fail safe or experimental manner.

We are still waiting for this.

arkpar:
The parity-db feature is complete and the integration into substrate was pending testing with this PR: https://github.com/paritytech/polkadot-sdk/pull/4176
@MattHalpinParity is taking over the effort to get it merged.

timorleph:
So I assume #4176 is the sole blocker here – do you have an (approximate ofc) expected date of delivery? Asking mostly since the last activity there is from a couple months ago.

rossbulat:
## Concern: Multi-tree Storage Overhead for Validators

I’m concerned about the steady-state storage cost of multi-tree, especially if the long-term goal is to make parity-db the default or eventually retire RocksDB.

From the current parity-db multi-tree implementation, each stored node appears to be packed as:

    stored_size = node_data.len() + num_children * 8 + 1

So multi-tree adds:

    8 bytes per direct child NodeAddress
    + 1 byte child count

That means the direct per-node metadata overhead is:

    overhead_bytes = 8 * child_count + 1
    overhead_percent = ((8 * child_count + 1) / encoded_node_size) * 100

For a full tree, average child count is likely close to `1`, since total edges are roughly `nodes - roots`. That gives a rough average overhead of about `9 bytes per node`.

## Example Direct Overheads

| Encoded node size | Avg children | Extra bytes | Direct overhead |
|---:|---:|---:|---:|
| 64 B | 1 | 9 B | 14.1% |
| 96 B | 1 | 9 B | 9.4% |
| 128 B | 1 | 9 B | 7.0% |
| 192 B | 1 | 9 B | 4.7% |
| 256 B | 1 | 9 B | 3.5% |
| 512 B | 1 | 9 B | 1.8% |
| 128 B | 2 | 17 B | 13.3% |
| 256 B | 2 | 17 B | 6.6% |
| 512 B | 16 | 129 B | 25.2% |

This is only the direct node payload overhead. The bigger issue is that multi-tree columns currently cannot use compression, and the SDK PR disables compression for the multi-tree state column.

So if the current flat parity-db state column benefits from LZ4 compression, the effective increase could be materially larger.

For example, assuming average `1` child per node:

    current ~= encoded_node_size * compression_ratio
    multi_tree ~= encoded_node_size + 9
    increase = ((encoded_node_size + 9) / (encoded_node_size * compression_ratio)) - 1

## Projected Increase With Compression Loss

For 128-byte average encoded nodes:

| Current compression ratio | Approx increase |
|---:|---:|
| 1.0 | 7.0% |
| 0.9 | 18.9% |
| 0.8 | 33.8% |
| 0.7 | 52.9% |

For 256-byte average encoded nodes:

| Current compression ratio | Approx increase |
|---:|---:|
| 1.0 | 3.5% |
| 0.9 | 15.0% |
| 0.8 | 29.4% |
| 0.7 | 47.9% |

These are not full database-size projections. They do not account for hash index savings, refcount table changes, value table fill factor, pruning mode, or root sharing. But they do show that multi-tree has a real storage-cost risk, and that risk is very relevant for validator operators.

For validators, disk cost is not a minor concern. If multi-tree improves sync speed but increases steady-state disk by, say, `15-40%`, many operators may avoid it. That would be a problem if the direction is to make parity-db the single supported database backend.

## Measurements Needed

I think this needs to be quantified before multi-tree is promoted beyond experimental. Suggested required measurements:

1. Average encoded trie node size on real chains.
2. Average child count per stored node.
3. Current LZ4 compression ratio for the state column.
4. Full `STATE` column size before/after.
5. Full database size before/after.
6. Separate numbers for validator/pruned mode and archive mode.
7. Import speed vs final disk-size tradeoff.

Ideally, the proposal should include an explicit storage budget target, for example:

    multi-tree must not increase pruned validator DB size by more than X%

or otherwise explain why the increase is acceptable.

## Potential Mitigations

Potential mitigations worth considering:

1. Compress multi-tree node payloads despite child-address metadata.
2. Encode child addresses more compactly than fixed `u64`.
3. Use varint or tier-local address encoding.
4. Store child-address metadata in a sidecar structure only where it materially improves traversal.
5. Support post-sync compaction/repacking.
6. Benchmark whether multi-tree removes enough hash-index/refcount overhead to offset the new child-address overhead.

I’m supportive of making parity-db faster and more state-aware, but for validators the storage-cost side needs to be treated as a first-class acceptance criterion, not a secondary benchmark.

cheme:
Hey, long time I didn t look at the question, but my first glance opinion on the different points:
I think this needs to be quantified before multi-tree is promoted beyond experimental. Suggested reordering of priority for the 7 action point:


1. Full STATE column size before/after.

  likely also needed for "real chains". Do we mean "archive" by full? (do validator still run archive node)
  for me it is the first step, since 64 bits per children additional size should be compared with size of index (we did on the relay chain long ago but I cannot find back the info), index likely contain an hash for collision (not sure anymore). 

2. Separate numbers for validator/pruned mode and archive mode.

3. Full database size before/after.

4. Import speed vs final disk-size tradeoff.

  Here import speed should be import when synced I think (most issues linked are about initial import).
  Also since import is a bit asynch it is actually access speed that is interesting (read access). I don’t remember
  too well but import (write) was similar or a bit better for one, but main diff was likely read. But these old bench
  where run with first implementation that involves lot of random access (I am not sure if things have been improve
  since my last messages but there were plan to @MattHalpinParity ).

5. Current LZ4 compression ratio for the state column.

  not sure anymore why we disable it for , would need to look at the code again.

6. Average child count per stored node.

  quite interesting but in this question, it only confirm the number from previous result (I don t think it can provide definitive conclusion so I would put this low prio).

7. Average encoded trie node size on real chains.

The real chain is the big question, but should look at actual size first I think.

acatangiu:
Afaik paritydb has issues reported in the wild, and is still an “experimental” feature that was even discussed to be removed since it’s stuck in this stage for a long time.

@bkchr @skunert please correct me if I’m wrong and/or weigh in here.

PR #1793, opened 2023-10-04
Resolving Compile Errors in Core Time Prep Tests

BradleyOlson64:
This leaves one build error I wasn't sure how to handle in builder.rs. 

No guarantees all the tests still pass, but I did my best to swap in the new types without altering any logic.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3880392

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -110,26 +110,41 @@ pub enum SpotTrafficCalculationErr {
...
 
 /// Internal representation of an order after it has been enqueued already.
-#[derive(Encode, Decode, TypeInfo)]
+#[cfg(not(test))]
+#[derive(Encode, Decode, TypeInfo, Debug, PartialEq)]
 pub(super) struct EnqueuedOrder {
 	pub para_id: ParaId,
 }
 
+// Looser encapsulation for tests
```
eskimor, line 141: Can we re-organize tests, so this is not necessary? I see that tests in the scheduler are accessing this, but this seems a bit off from a high level perspective. Scheduler tests should not bother with internal state of the on-demand assigner. It is likely that these checks should be replaces with tests in the on-demand assigner itself.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -491,6 +507,27 @@ where
 		})
 	}
 
+	#[cfg(test)]
+	pub fn add_on_demand_order(
```
eskimor, line 511: Same.

polkadot/runtime/parachains/src/scheduler.rs
```diff
@@ -292,10 +292,8 @@ impl<T: Config> Pallet<T> {
 				.into_iter()
 				.filter(|(freed_index, _)| (freed_index.0 as usize) < c_len)
 				.for_each(|(freed_index, freed_reason)| {
-					match sp_std::mem::replace(
```
eskimor, line 295: What is wrong with sp_std::mem::replace?

polkadot/runtime/parachains/src/scheduler/tests.rs
```diff
@@ -186,7 +189,10 @@ fn claimqueue_ttl_drop_fn_works() {
 		assert!(!claimqueue_contains_para_ids::<Test>(vec![para_id]));
 
 		// Add a claim on core 0 with a ttl in the future (15).
-		let paras_entry = ParasEntry::new(Assignment::new(para_id), now + 5);
+		let paras_entry = ParasEntry::new(
+			UnifiedAssignmentType::<Test>::OnDemand(OnDemandAssignment::new(para_id, core_idx)),
```
eskimor, line 193: We could and likely should use a test config that uses a mock test assignment provider, so we don't need to expose internals of the used assignment provider here.

eskimor (commented):
Good work @BradleyOlson64 ! Left a few comments. I think we should untangle tests a bit. It should not be that we need to expose assigner internals to scheduler tests.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3880791

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
BradleyOlson64: I think we can definitely do this. It seems there are two ways the scheduler tests are breaking encapsulation of the assigner.
1. Calling interface that should be private like `add_on_demand_order` rather than using the more complex public interface, `place_order_keep_alive`. This requires exposing the otherwise private function and the `EnqueuedOrder` parameter type.
2. Constructor for `Assignment`, needed to produce expected `ParasEntry` results. 

For 1, I'd like to use `place_order_keep_alive` rather than `add_on_demand_order` in scheduler tests. If needed, I can create a helper function to reduce the complexity of those repeated calls. That should allow us to keep the related internals private.
For 2, I'd like to just expose the constructor by default rather than just for tests. It doesn't seem that dangerous to me.

polkadot/runtime/parachains/src/scheduler.rs
BradleyOlson64: Maybe some quirk of my local setup, but the sp_std version was producing errors.

polkadot/runtime/parachains/src/scheduler/tests.rs
BradleyOlson64: Not too familiar with how I'd make a mock test assignment provider. If my above suggestion can substitute, that would probably be easier for me to implement.

polkadot/runtime/parachains/src/assigner_parachains.rs
```diff
@@ -44,13 +44,13 @@ pub mod pallet {
...
 
-#[derive(Debug, Encode, Decode, TypeInfo)]
+#[derive(Debug, Encode, Decode, TypeInfo, PartialEq, Clone)]
 pub struct ParachainsAssignment {
 	pub para_id: ParaId,
 }
 
 impl ParachainsAssignment {
-	fn new(para_id: ParaId) -> Self {
+	pub fn new(para_id: ParaId) -> Self {
```
eskimor, line 53: Why does this need to be pub?

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +385,101 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+		#[pallet::pallet]
+		#[pallet::without_storage_info]
+		pub struct Pallet<T>(_);
+
+		#[pallet::config]
+		pub trait Config: frame_system::Config + configuration::Config + paras::Config {}
+
+		/// Creates an empty on demand queue if one isn't present in storage already.
+		#[pallet::type_value]
+		pub(super) fn OnDemandQueueOnEmpty<T: Config>() -> VecDeque<V0Assignment> {
```
eskimor, line 404: Why do we need this in mocking?

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +385,101 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+			VecDeque::new()
+		}
+
+		#[pallet::storage]
+		pub(super) type MockAssignerQueue<T: Config> =
+			StorageValue<_, VecDeque<V0Assignment>, ValueQuery, OnDemandQueueOnEmpty<T>>;
+	}
+
+	impl<T: Config> Pallet<T>{
+		pub fn add_on_demand_order(
```
eskimor, line 414: Why is our mock designer an on-demand assigner?

polkadot/runtime/parachains/src/scheduler/tests.rs
```diff
@@ -724,14 +741,15 @@ fn schedule_schedules_including_just_freed() {
 fn schedule_clears_availability_cores() {
 	let mut config = default_config();
 	config.scheduling_lookahead = 1;
+	config.on_demand_cores = 2;
```
eskimor, line 744: Why do we care about this number here?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3898992

polkadot/runtime/parachains/src/assigner_parachains.rs
BradleyOlson64: In some discarded version of this pr, probably. Though not anymore 👍

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3899632

polkadot/runtime/parachains/src/scheduler.rs
BradleyOlson64: sp_std is compiling fine after `rustup update`

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3899759

polkadot/runtime/parachains/src/mock.rs
BradleyOlson64: Very good point. This function isn't actually the problem on that front though. It was implementing the `AssignmentProvider` api as if `MockAssigner` was just an `OnDemandAssigner`.

I added the legacy auction assigner functionality to it, making it more of a two in one assigner that still supports the testing specific on demand calls, like `add_on_demand_order`.

polkadot/runtime/parachains/src/scheduler/tests.rs
BradleyOlson64: Right, the defaults are fine for this test.

polkadot/runtime/parachains/src/mock.rs
BradleyOlson64: The tests passed without it, so it seems not.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3899891

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3900021

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3900153

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +389,142 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+		}
+
+		fn is_legacy_core(core_idx: &CoreIndex) -> bool {
+			let parachain_cores = <paras::Pallet<Test>>::parachains().len() as u32;
+			(0..parachain_cores).contains(&core_idx.0)
+		}
+	}
+
+	impl<T: Config> AssignmentProvider<BlockNumber> for Pallet<T> {
+		type AssignmentType = TestUnifiedAssignment;
```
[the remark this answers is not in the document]
BradleyOlson64: The main reason is so that `push_back_assignment` knows how to behave. It is a part of the `AssignmentProvider` interface, so I can't change its signature. And it only has the parameter `AssignmentType` to tell it how to behave. Using an `AssignmentType` enum seems elegant enough though perhaps it could be made simpler still. A bunch of tests were failing before I swapped in `TestUnifiedAssignment`.

I'm open to other options if there's something you'd prefer!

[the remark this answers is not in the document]
BradleyOlson64: It occurs to me that checking whether the ParaId in a V0Assignment corresponds to a parachain would likely be sufficient. I'll try that, and get rid of `TestUnifiedAssignment` if so.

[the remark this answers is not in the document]
eskimor: The point is: This interface does not tell anything about the nature of the assignment provider. If assignments would need to be of a certain type, the architecture would be completely broken. The only requirement is that it implements `Assignment`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3911420

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3911797

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3912814

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +388,121 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+				front
+			}
+		}
+
+		// We don't care about core affinity in the test assigner
+		fn report_processed(_assignment: Self::AssignmentType) {}
+
+		fn push_back_assignment(assignment: Self::AssignmentType) {
+			if <paras::Pallet<T>>::is_parachain(assignment.para_id()) {
+			} else {
```
eskimor, line 474: Why are we doing this?

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +388,121 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+					Ok(_) => {},
+					Err(_) => {},
+				}
+			}
+		}
+
+		// These numbers are set via config in individual scheduler tests. Just pass them through
+		// here.
+		fn get_provider_config(core_idx: CoreIndex) -> AssignmentProviderConfig<BlockNumber> {
+			if Self::is_legacy_core(&core_idx) {
```
eskimor, line 486: And this?

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -388,6 +388,121 @@ impl ValidatorSetWithIdentification<AccountId> for MockValidatorSet {
...
+		fn session_core_count() -> u32 {
+			let config = <configuration::Pallet<Test>>::config();
+			(<paras::Pallet<Test>>::parachains().len() as u32)
+				.saturating_add(config.on_demand_cores)
+		}
+
+		// Create a new assignment to pass if core belongs to bulk buyer, else pop
+		// on demand assignment from front of queue.
+		fn pop_assignment_for_core(core_idx: CoreIndex) -> Option<Self::AssignmentType> {
+			if Self::is_legacy_core(&core_idx) {
```
eskimor, line 455: Why are we doing this?

BradleyOlson64: This assigner is fulfilling the role of either a `ParachainsAssigner` or an `OnDemandAssigner` for each call. By calling `is_legacy_core` we determine which way to act using the input `core_idx`. The idea is that for a bulk time core index this assigner will always behave like a `ParachainsAssigner`. The same applies to on-demand core indices.

The follow up question to this would be, "Why do we need the MockAssigner to act like any particular kind of assigner." The answer is that the assigner is how we control the flow of eventual `ParasEntries` in scheduler tests as well as the `AssignmentProviderConfig` telling scheduler tests how to handle them. 

We test the scheduler in scenarios where it needs to behave as if it has a real `OnDemandAssigner`: `on_demand_claims_are_pruned_after_timing_out` and `next_up_on_time_out_reuses_claim_if_nothing_queued`.

We also test it in scenarios where it needs to behave as if it has a `ParachainsAssigner`:
`next_up_on_time_out_is_parachain_always`

BradleyOlson64: In this specific instance, for on-demand core indices, we need to drain an entry from the queue each time so that we can test scenarios where the queue becomes empty.

polkadot/runtime/parachains/src/mock.rs
BradleyOlson64: For assignments corresponding to parachains we act as if we are in a real `ParachainsAssigner`, where there is no order queue to push back onto. 

This keeps bulk parachain assignments that are pushed back from accidentally adding to the on-demand queue.

BradleyOlson64: I'll add more comments describing what's going on here and throughout the mock assigner.

eskimor: But why? You seem to be basically reproducing the real functionality .... why have the mock in the first place then?

BradleyOlson64: Perhaps I misinterpreted this:

“We could and likely should use a test config that uses a mock test assignment provider, so we don't need to expose internals of the used assignment provider here”

I took this to mean, “Let’s construct a mock assignment provider with minimum necessary complexity to do the following.
1. Allow all scheduler tests to compile and pass without re-writing them.
2. Avoid exposing any internals of actual assigners to those tests.

It seems there’s a third criterion I’m not sure can be met without giving up the re-writing part of the first:
3. Avoid replicating any large portion of the existing assigner logic.

It seems like you’re not pleased with the way this mock assigner idea is turning out. Should it be scrapped?

eskimor: No, 3 is not directly the requirement. The requirement is, that if the interna of the assignment provider matter, then the scheduler tests are broken and should be fixed. Any assignment provider internal functionality that is tested there, should be tested in the assignment provider itself.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4051603

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4053971

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4068784

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4074700

polkadot/runtime/parachains/src/scheduler/tests.rs
```diff
@@ -532,190 +414,143 @@ fn schedule_schedules_including_just_freed() {
...
 		now = 3;
 		run_to_block(now, |_| None);
 
 		{
-			let scheduled: BTreeMap<_, _> = Scheduler::scheduled_entries().collect();
+			let scheduled: BTreeMap<_, _> = scheduled_entries().collect();
 
-			// cores 0 and 1 are occupied by lease holding parachains. cores 2 and 3 are occupied by
-			// on-demand parachain claims. core 4 was free.
+			// cores 0 and 1 are occupied by on-demand claims. core 2 was free.
```
eskimor, line 498: I think this comment is outdated. (No on-demand here)

eskimor (approved):
Nice! Thanks @BradleyOlson64 !

polkadot/runtime/parachains/src/scheduler.rs
```diff
@@ -364,34 +364,35 @@ impl<T: Config> Pallet<T> {
...
+						if let Some(dropped) = maybe_dropped {
+							num_dropped += 1;
+							T::AssignmentProvider::report_processed(dropped.assignment);
+						} else {
+							i += 1;
+						}
+					}
 
-						T::AssignmentProvider::report_processed(dropped.assignment);
+					for _ in 0..num_dropped {
```
eskimor, line 387: Note: I refactored/cleaned up this a bit more.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4088846

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4088834

PR #1805, opened 2023-10-06
TryDecodeEntireState check for storage types and pallets

### LINKED ISSUE #241: `DecodeAll` check for `post_migrate` hook
There are a few cases when migrations did not verify that the resulting state of a pallet is decodable with the new runtime.  
This can easily happen when using `StorageMap::translate` since it just ignores un-decodable elements and carries them over into the new runtime.  

We could add a check that tries to decode the whole state after a migration (or batch thereof) to verify that the state is not corrupted.  
Sure there could still be un-decodable data in storage in the form of blobs (eg pallet-preimage) but decoding errors needs to be handled in these cases anyway. This would help but not completly solve with the introduction of `BoundedVec`.  
Migrations still need to truncate any unbounded vectors or use `WeakBoundedVec` instead since otherwise someone could front-run the migration. cc @kianenigma

PieWol:
### This PR is a port of this [PR for substrate](https://github.com/paritytech/substrate/pull/13013) by @kianenigma 

Add infrastructure needed to have a Pallet::decode_entire_state(), which makes sure all "typed" storage items defined in the pallet are decode-able.

This is not enforced in any way at the moment. Teams who wish to integrate/use this in the try-runtime feature flag should add frame_support::storage::migration::EnsureStateDecodes as the LAST ITEM of the runtime's custom migrations, and pass it to frame-executive. This will make it usable in try-runtime on-runtime-upgrade.

This now catches cases like https://github.com/paritytech/polkadot-sdk/pull/1969:  
```pre
ERROR runtime::executive] failed to decode the value at key: Failed to decode value at key: 0x94eadf0156a8ad5156507773d0471e4ab8ebad86f546c7e0b135a4212aace339. Storage info StorageInfo { pallet_name: Ok("ParaScheduler"), storage_name: Ok("AvailabilityCores"), prefix: Err(Utf8Error { valid_up_to: 0, error_len: Some(1) }), max_values: Some(1), max_size: None }. Raw value: Some("0x0c010101010101")
```

... or:
image

Closes #241


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3890900

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4035611

substrate/frame/support/src/traits/try_runtime.rs
```diff
@@ -63,26 +63,47 @@ impl Default for Select {
...
 
+/// A value could not be decoded.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TryDecodeEntireStorageError {
+	/// The key of the undecodable value.
+	key: Vec<u8>,
+	/// The raw value.
+	raw: Option<Vec<u8>>,
+	/// The storage info of the key.
+	info: StorageInfo,
```
ggwpez: I took he liberty to add a custom error type. In the future we can maybe use a custom error instead of `DispatchError`, since it only supports `&'static str`.  
Maybe we can create a wrapper type that can be created from a `DispatchError`.

substrate/frame/support/src/storage/migration.rs
```diff
@@ -385,6 +385,51 @@ pub fn move_prefix(from_prefix: &[u8], to_prefix: &[u8]) {
 	}
 }
 
+/// A phony migration that does nothing, except executing `TryDecodeEntireStorage` on
+/// `post_upgrade`, which implies it is only available if `try-state` feature is used.
+///
+/// This can be used typically in the top level runtime, where `AllPallets` typically comes from
+/// `construct_runtime!`.
+pub struct EnsureStateDecodes<AllPallets>(sp_std::marker::PhantomData<AllPallets>);
```
bkchr, line 393: Why is this a migration and not just part of `try-runtime`? I mean it already requires `try-runtime` and this could just be an integral part of it?

kianenigma: This was discussed in the original PR description as a follow-up idea: https://github.com/paritytech/substrate/pull/13013#issue-1509379972. 

I still think it is good to add this reusable migration here so it can be added to any runtime ad-hoc, and later integrate it into `try-runtime-cli`.

bkchr: Makes no real sense what you are saying. That people can add this adhoc to their runtime, it requires them to update their deps. If instead of adding this as an extra migration, we directly add it to the try runtime logic. People will need to update the deps any way to get the changes.

This can also not really be integrated into try runtime cli, if you are not planing on writing something with subxt that does this.

ggwpez: We can just call the migration from `Executive::try_runtime_upgrade` if we want to bake it into the system permanently.  
It gets passed a `UpgradeCheckSelect` that we could use to select whether it should run.

liamaharon: I'm in favour of this being configurable via `UpgradeCheckSelect`, like the try-state checks. Or, maybe just run it if the user is running `try-state` checks. Since a really obvious storage invarient is that it should be decodable.

bkchr: > when `UpgradeCheckSelect` is configured to run any `try-state` checks. It seems a really obvious storage invariant that all storage should be decodable.

Sounds good!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4061785

substrate/frame/support/src/storage/migration.rs
ggwpez: done

substrate/frame/executive/src/lib.rs
```diff
@@ -353,31 +360,47 @@ where
...
 				checks.pre_and_post(),
 			)?;
+		// Nothing should modify the state after the migrations ran:
+		let _guard = StorageNoopGuard::default();
+
+		// The state must be decodable:
+		if checks.any() {
+			match AllPalletsWithSystem::try_decode_entire_state() {
+				Ok(bytes) => {
+					log::info!(
```
liamaharon, line 383:
```suggestion
					log::debug!(
```

substrate/frame/executive/src/lib.rs
```diff
@@ -353,31 +360,47 @@ where
...
+						target: LOG_TARGET,
+						"decoded the entire state ({bytes} bytes)",
+					);
+				},
+				Err(err) => {
+					log::error!(
+						target: LOG_TARGET,
+						"failed to decode the state: {err}",
```
liamaharon, line 391:
```suggestion
						"`try_decode_entire_state` failed: {err}",
```

substrate/frame/executive/src/lib.rs
```diff
@@ -353,31 +360,47 @@ where
...
+					);
+
+					return Err("failed to decode a value from the storage".into())
```
liamaharon, line 394:
```suggestion
					return Err("`try_decode_entire_state` failed".into())
```

substrate/frame/support/src/traits/storage.rs
```diff
@@ -109,6 +107,19 @@ pub struct StorageInfo {
 	pub max_size: Option<u32>,
 }
 
+// Use strings instead of bytes for the debug output.
```
liamaharon, line 110: Thanks, big win here

substrate/frame/support/src/traits/try_runtime/decode_entire_state.rs
```diff
@@ -0,0 +1,484 @@
...
+
+#[cfg_attr(all(not(feature = "tuples-96"), not(feature = "tuples-128")), impl_for_tuples(64))]
+#[cfg_attr(all(feature = "tuples-96", not(feature = "tuples-128")), impl_for_tuples(96))]
+#[cfg_attr(feature = "tuples-128", impl_for_tuples(128))]
+impl TryDecodeEntireStorage for Tuple {
+	fn try_decode_entire_state() -> Result<usize, TryDecodeEntireStorageError> {
+		let mut len = 0usize;
+		for_tuples!( #( len = len.saturating_add(Tuple::try_decode_entire_state()?); )* );
+		Ok(len)
+	}
```
liamaharon, line 54: Can you please replace this with something more like I've done here: https://github.com/paritytech/polkadot-sdk/pull/2019/files#diff-8d9b13bf2084a5c02e2986c28a4f00e2c2af5ff70f12ba050be4e34bac6fdd4fR147-R167

For better logging and so it doesn't terminate early if one of the pallets fails to decode?

substrate/frame/support/procedural/src/pallet/expand/storage.rs
```diff
@@ -822,12 +822,57 @@ pub fn expand_storages(def: &mut Def) -> proc_macro2::TokenStream {
...
+
+					// NOTE: for now, we have to exclude storage items that are feature gated.
+					let mut decoded = 0usize;
+					#(
+						#frame_support::__private::log::debug!(target: "try-decode-state", "trying to decode storage: \
+						{pallet_name}::{}", stringify!(#storage_names));
+
+						decoded +=
+							<#storage_names as #frame_support::traits::TryDecodeEntireStorage>::try_decode_entire_state()?;
+					)*
```
liamaharon, line 868: If there're multiple undecodable keys, will this terminate early on the first one? My instinct is that we want to log all undecodable keys.

liamaharon: also when this errors, it should log to stderr the undecodable storage names.

liamaharon (commented):
Let's also execute this in `try_execute_block`, unless it's too expensive?

Can you please post a screenshot of the log output when a couple of pallets have some undecodable state?

substrate/frame/support/procedural/src/pallet/expand/storage.rs
ggwpez: Yea, just realizing now that it errors early. Will change.

substrate/frame/support/src/traits/try_runtime/decode_entire_state.rs
ggwpez: Yea I will aggregate the error and then log at the call site :+1:

ggwpez:
>Let's also execute this in try_execute_block, unless it's too expensive?

It takes about 5 secs on Westend, so not too slow. I added it now.

Error print looks like this:
image

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4095465

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4095467

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4095466

liamaharon (approved):
Looks great 🎉

WDYT logging the hex key only in DEBUG? Seems pretty verbose.

substrate/frame/executive/src/lib.rs
```diff
@@ -308,11 +315,15 @@ where
 		let _guard = frame_support::StorageNoopGuard::default();
 		<AllPalletsWithSystem as frame_support::traits::TryState<
 			BlockNumberFor<System>,
-		>>::try_state(*header.number(), select)
+		>>::try_state(*header.number(), select.clone())
 		.map_err(|e| {
 			log::error!(target: LOG_TARGET, "failure: {:?}", e);
 			e
 		})?;
+		if select.any() {
```
kianenigma, line 323: not a strong opinion, but I am a tad bit worried about the execution time of this. 

@liamaharon already brought up somewhere that running `try_state` is already time consuming af. Now, we're attaching something else that would potentially takes tens of minutes (I am guesing -- am I wrong?) to it. 

It is okay for now, but someday we might want to break what "try-runtime checks you run" into a configuration like this: 



```rust
/// What level of depth should each try-state check go into. 0 means be as fast as possible, 255 means be as slow. Possibly represent this with 3 level or so instead of 255. 
type TryStateLevel = u8;

enum TryStateLevelAlternative {
    Fast,
    Medium,
    Verbose
}

/// Which pallets should have their [`try_state`] involed in the context of [`TryStateChecks`]?
enum TryStatePallets {
    /// All pallets. 
    All,
    /// given number of pallets, round-robin based on block number. 
    RoundRobin(u8),
    /// given number of random pallets. Not sure if it is useful.
    Random(u8)
}


/// Final config, shoudl be passed into all `try_*` runtime APIs
struct TryStateChecks {
    decode_all: bool,
    try_state_pallets: TryStatePallets
    try_state_level: TryStateLevel 
}  
```

substrate/frame/support/procedural/src/pallet/expand/event.rs
```diff
@@ -127,11 +127,12 @@ pub fn expand_event(def: &mut Def) -> proc_macro2::TokenStream {
 		let trait_use_gen = &def.trait_use_generics(event.attr_span);
 		let type_impl_gen = &def.type_impl_generics(event.attr_span);
 		let type_use_gen = &def.type_use_generics(event.attr_span);
+		let pallet_ident = &def.pallet_struct.pallet;
 
 		let PalletEventDepositAttr { fn_vis, fn_span, .. } = deposit_event;
 
 		quote::quote_spanned!(*fn_span =>
-			impl<#type_impl_gen> Pallet<#type_use_gen> #completed_where_clause {
+			impl<#type_impl_gen> #pallet_ident<#type_use_gen> #completed_where_clause {
```
kianenigma, line 135: 🙈

substrate/frame/support/procedural/src/pallet/expand/storage.rs
```diff
@@ -822,12 +822,69 @@ pub fn expand_storages(def: &mut Def) -> proc_macro2::TokenStream {
...
+						.expect("Every active pallet has a name in the runtime; qed");
+
+					#frame_support::__private::log::debug!(target: "try-decode-state", "trying to decode pallet: {pallet_name}");
+
+					// NOTE: for now, we have to exclude storage items that are feature gated.
+					let mut errors = #frame_support::__private::sp_std::vec::Vec::new();
+					let mut decoded = 0usize;
+
+					#(
+						#frame_support::__private::log::debug!(target: "try-decode-state", "trying to decode storage: \
```
kianenigma, line 865: all log targets emitted from runtime should start with `runtime` and then be like `runtime::other-stuff::more` if need be. 

@liamaharon this will be quite relevant for monitoring, as we can then filter on all `runtime` log targets, and it will give us anything starting with `runtime`.

kianenigma (approved):
LGTM module https://github.com/paritytech/polkadot-sdk/pull/1805/files#r1374670203

substrate/frame/executive/src/lib.rs
liamaharon: @ggwpez mentioned it takes ~5s on Westend. So not that bad.

ggwpez: On Rococo it was even faster. For now should be fine, but am also in favour of more granular settings.

substrate/frame/support/procedural/src/pallet/expand/event.rs
ggwpez: This was on purpose, you think its wrong?

substrate/bin/node-template/runtime/src/lib.rs
```diff
@@ -313,6 +313,12 @@ pub type SignedExtra = (
 	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 );
 
+/// All migrations of the runtime, aside from the ones declared in the pallets.
+///
+/// This can be a tuple of types, each implementing `OnRuntimeUpgrade`.
+#[allow(unused_parens)]
+type Migrations = ();
```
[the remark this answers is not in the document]
ggwpez: Not sure anymore, but since this is the template it maybe helps to see where migrations are configured.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4224860

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4224862

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4224861

ggwpez (approved):

PR #1807, opened 2023-10-06
Refactor pallet recovery to fungibles

juangirini:
Partial for https://github.com/paritytech/polkadot-sdk/issues/226

The Currency (and `ReservableCurrency`) traits are deprecated after this https://github.com/paritytech/substrate/pull/12951 in favour of the fungible traits.

A few things have changed:

`ReservableCurrency::reserve` becomes `fungible::MutateHold::hold`; `ReservableCurrency::unreserve` becomes `fungible::MutateHold::release` and `ReservableCurrency::repatriate_reserved` becomes `fungible::MutateHold::transfer_on_hold`.
Holds are roughly analogous to reserves, but are now explicitly designed to be infallibly slashed. They do not contribute to the ED but do require a provider reference, removing any possibility of account reference counting from being problematic for a slash. They are also always named, ensuring different holds do not accidentally slash each other's balances.

Also they require an identifier which is configurable and is expected to be an enum aggregated across all pallet instances of the runtime.

In the Pallet Recovery context we only have `HoldReason::RecoveryProcessDeposit` and `HoldReason::ConfigurationDeposit`

```rust
/// The reasons for the pallet recovery placing holds on funds.
#[pallet::composite_enum]
pub enum HoldReason {
	/// The Pallet holds it as the initial Configuration Deposit, when the Recovery
	/// Configuration is created.
	ConfigurationDeposit,
	/// The Pallet holds it as the Deposit for starting a Recovery Process.
	RecoveryProcessDeposit,
}
```
## Upgrade notes
The `pallet_recovery::Config` trait now requires new type `RuntimeHoldReason`:
```rust
impl pallet_recovery::Config for Runtime {
	...
	type RuntimeHoldReason = RuntimeHoldReason;
	...
}
```
Also, `construct_runtime` needs `pallet_recovery::HoldReason`:
```rust
construct_runtime!(
	pub enum Runtime
	{
		...
		Recovery: recovery::{Pallet, Call, Storage, Event<T>, HoldReason},
		...
	}
)
```

### TODO
- [ ] migration


substrate/frame/recovery/src/benchmarking.rs
```diff
@@ -51,10 +50,7 @@ fn generate_friends<T: Config>(num: u32) -> Vec<<T as frame_system::Config>::Acc
 
 	for friend in 0..friends.len() {
 		// Top up accounts of friends
-		T::Currency::make_free_balance_be(
-			&friends.get(friend).unwrap(),
-			BalanceOf::<T>::max_value(),
-		);
+		T::Currency::set_balance(&friends.get(friend).unwrap(), TOP_UP.into());
```
juangirini, line 53: Using `BalanceOf::<T>::max_value()` with `set_balance` would quickly overflow, that's why we use a const way lower than that amount instead

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3893413

substrate/frame/recovery/src/lib.rs
```diff
@@ -235,9 +240,15 @@ pub mod pallet {
 			+ From<frame_system::Call<Self>>;
 
 		/// The currency mechanism.
-		type Currency: ReservableCurrency<Self::AccountId>;
+		#[cfg(feature = "runtime-benchmarks")]
+		type Currency: Mutate<Self::AccountId>
+			+ MutateHold<Self::AccountId, Reason = Self::RuntimeHoldReason>;
```
juangirini, line 245: We need to `set_balance` in benchmarking, so `Mutate` is needed

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3893670

juangirini:
bot bench $ pallet dev pallet_recovery

substrate/frame/recovery/src/benchmarking.rs
gilescope: Another alternative is using max but shifting it down a few but either seems fine to me.

georgepisaltu:
Hey, I was already working on moving `pallet-recovery` to the new `fungible` interface, but the edit on #226 seems to have been removed. Since you opened the PR first, I'll just stick to reviewing yours then.

juangirini:
I did update #226 before starting this, not sure how that got lost in between. I have just updated it again, sorry to hear that you already spent time on it.

juangirini:
bot bench substrate-pallet --pallet=pallet_recovery

substrate/frame/recovery/src/lib.rs
```diff
@@ -187,7 +192,7 @@ type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup
 pub struct ActiveRecovery<BlockNumber, Balance, Friends> {
 	/// The block number when the recovery process started.
 	created: BlockNumber,
-	/// The amount held in reserve of the `depositor`,
+	/// The amount held the `depositor`,
```
franciscoaguirre, line 195:
```suggestion
	/// The amount held by the `depositor`,
```

franciscoaguirre (commented):
You're missing a migration for those runtimes that use this pallet

juangirini:
> You're missing a migration for those runtimes that use this pallet

I guess we can't really migrate until we have the MBM ready. And lazy migrations don't seem to be a great solution as with it 
 we are unable to remove the `Currency` trait, and it adds a lot of noise code that will need to be removed eventually

substrate/frame/recovery/src/lib.rs
ggwpez: It is possible to set custom where bounds for benchmarks, [like this](https://github.com/paritytech/polkadot-sdk/blob/8f6b6322c21c1eff2d075cd2f2af77191652df6f/polkadot/runtime/common/src/paras_registrar/mod.rs#L1507).  
So it should be possible to add a `<T as Config>::Currency: Mutate` there.

substrate/frame/recovery/src/lib.rs
```diff
@@ -637,13 +661,16 @@ pub mod pallet {
 			// Take the active recovery process started by the rescuer for this account.
 			let active_recovery =
 				<ActiveRecoveries<T>>::take(&who, &rescuer).ok_or(Error::<T>::NotStarted)?;
-			// Move the reserved funds from the rescuer to the rescued account.
+			// Move the held funds from the rescuer to the rescued account.
 			// Acts like a slashing mechanism for those who try to maliciously recover accounts.
-			let res = T::Currency::repatriate_reserved(
+			let res = T::Currency::transfer_on_hold(
```
ggwpez, line 666: I am thinking about how to do a lazy migration here. Maybe you can require the `Currency` config trait to be `Currency + MutateHold` so that it requires currency and fungibles.  

Then in this function you can first try to use the new `transfer_on_hold` function, but if that does not work then fallback to the old `repatriate_reserved`. Do you think this could work?  
Otherwise the MBMs are still in review and will need audit, so its a bit out.

ggwpez (commented):
Looks good 👍

substrate/frame/recovery/src/lib.rs
juangirini: I am not a big fan of lazy migrations here, cause that will generate some "garbage" code that will live there forever. It seems more like a patch to me than a real solution. It's been a while already since we wanted the traits to be migrated that I would rather wait for the MBM to be ready and apply a proper solution

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4807116

bkchr:
Stale

PR #1809, opened 2023-10-06
Make `ExecResult` encodable

pmikolajczyk41:
# Description
We derive few useful traits on `ErrorOrigin` and `ExecError`, including `codec::Encode` and `codec::Decode`, so that `ExecResult` is en/decodable as well. This is required for a contract mocking feature (already prepared in drink: https://github.com/Cardinal-Cryptography/drink/pull/61). In more detail: `ExecResult` must be passed from runtime extension, through runtime interface, back to the pallet, which requires that it is serializable to bytes in some form (or implements some rare, auxiliary traits).

**Impact on runtime size**: Since most of these traits is used directly in the pallet now, compiler should be able to throw it out (and thus we bring no new overhead). However, they are very useful in secondary tools like drink or other testing libraries.

# Checklist

  required)
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3893281

pgherveou:
> Impact on runtime size: Since most of these traits is used directly in the pallet now, compiler should be able to throw it out (and thus we bring no new overhead). However, they are very useful in secondary tools like drink or other testing libraries.

Can we verify this just to be safe.

There is also the risk that we use them in production code where it's not optimal. 

Is the option of enabling these for the unsafe_debug (that we removed but could reintroduced) make the tooling much more complicated to use?

pmikolajczyk41:
just to keep a reference: this is a pretty similar PR to this one https://github.com/paritytech/substrate/pull/14723

I run rococo 'benchmark':
```shell
> git checkout codec-result
> cd cumulus/parachains/runtimes/contracts/contracts-rococo/
> cargo build --release
> wc -c ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.compact.wasm 
5400049 ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.compact.wasm
> wc -c ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.wasm 
6018247 ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.wasm

> git checkout 1835c091c42456e8df3ecbf0a94b7b88c395f623
HEAD is now at 1835c091c4 Revive Substrate Crate (#1477)
> cargo build --release
> wc -c ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.compact.wasm
5400064 ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.compact.wasm
> wc -c ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.wasm
6018262 ../../../../../target/release/wbuild/contracts-rococo-runtime/contracts_rococo_runtime.wasm
```
for my understanding they should be equal, but for some reason my PR decreases size... :shrug: anyway, LLVM should remove unused code from the final binary (which actually is a case, the current runtime code doesn't use any of these derivations)

regarding the feature flag, I wouldn't introduce it just for these derivations...

xermicus (approved):
Dead code elimination should definitively pick this up

PR #1813, opened 2023-10-07
Extend `Consideration` trait

NachoPal:
Draft exploring the idea of extending `Consideration` trait with `new_from_exact(who: &AccountId, new: Balance)`

It would be useful for migrations from `Deposits` to `Consideration` tickets where a fixed `Footprint` might not always return the expected deposited `Balance`.

Is it possible to do it in a different way? right now it creates a dependency on `Currency`, which I think is actually supposed to be removed in favor of `Consideration`.

### DIFF
--- a/substrate/frame/support/src/traits/storage.rs
+++ b/substrate/frame/support/src/traits/storage.rs
@@ -208,10 +208,16 @@ where
 /// treated as one*. Don't type to duplicate it, and remember to drop it when you're done with
 /// it.
 #[must_use]
-pub trait Consideration<AccountId>: Member + FullCodec + TypeInfo + MaxEncodedLen {
+pub trait Consideration<AccountId, Balance>: Member + FullCodec + TypeInfo + MaxEncodedLen {
 	/// Create a ticket for the `new` footprint attributable to `who`. This ticket *must* ultimately
 	/// be consumed through `update` or `drop` once the footprint changes or is removed.
 	fn new(who: &AccountId, new: Footprint) -> Result<Self, DispatchError>;
+
+	/// Create a ticket for a `new` balance attributable to `who`. This ticket *must* ultimately
+	/// be consumed through `update` or `drop` once the footprint changes or is removed.
+	/// This is useful when a new ticket needs to be created with a precise balance, instead of deriving
+	/// it from a footprint
+	fn new_from_exact(who: &AccountId, new: Balance) -> Result<Self, DispatchError>;
 
 	/// Optionally consume an old ticket and alter the footprint, enforcing the new cost to `who`
 	/// and returning the new ticket (or an error if there was an issue).
@@ -232,8 +238,11 @@ pub trait Consideration<AccountId>: Member + FullCodec + TypeInfo + MaxEncodedLe
 	}
 }
 
-impl<A> Consideration<A> for () {
+impl<A, B> Consideration<A, B> for () {
 	fn new(_: &A, _: Footprint) -> Result<Self, DispatchError> {
+		Ok(())
+	}
+	fn new_from_exact(_: &A, _: B) -> Result<Self, DispatchError> {
 		Ok(())
 	}
 	fn update(self, _: &A, _: Footprint) -> Result<(), DispatchError> {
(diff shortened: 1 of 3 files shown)


liamaharon:
> Is it possible to do it in a different way? right now it creates a dependancy on Currency, which I think is actually supposed to be removed in favor of Consideration.

Where does it create this dependency? 

This approach seems fine to me.

NachoPal:
> I don't see any dependency on `Currency` being introduced here?

The dependency is introduced when using `BalanceOf<T>`.

See https://github.com/paritytech/polkadot-sdk/pull/1813#issuecomment-1751817387

liamaharon:
The dev only needs to specify the `Balance` type, this type is not bound to `Currency`. 

`Consideration` already assumes the existence of a `Balance`, so this introduces no new dependencies.

NachoPal:
bot fmt

liamaharon (approved):

substrate/frame/support/src/traits/storage.rs
```diff
@@ -208,11 +208,17 @@ where
...
-pub trait Consideration<AccountId>: Member + FullCodec + TypeInfo + MaxEncodedLen {
+pub trait Consideration<AccountId, Balance>: Member + FullCodec + TypeInfo + MaxEncodedLen {
 	/// Create a ticket for the `new` footprint attributable to `who`. This ticket *must* ultimately
 	/// be consumed through `update` or `drop` once the footprint changes or is removed.
 	fn new(who: &AccountId, new: Footprint) -> Result<Self, DispatchError>;
 
+	/// Create a ticket for a `new` balance attributable to `who`. This ticket *must* ultimately
+	/// be consumed through `update` or `drop` once the footprint changes or is removed.
+	/// This is useful when a new ticket needs to be created with a precise balance, instead of
+	/// deriving it from a footprint
```
franciscoaguirre, line 219: The doc comment still mentions a footprint

kianenigma (commented):
High level, one of the main goals of `Consideration` is to abstract away the need for the existence of any "Balance/Currency" sub-system in a runtime, the fact that we need to add a `Balance` generic to the trait now seems like the wrong compromise to me. 

I recall this was solving an important issue with migrations though, so I would not block it. Possibly it can be a temporary hack, in which case you can document the method as "it is temporary".

What alternatives we have?

juangirini:
Which migrations would benefit from this?

NachoPal:
> Which migrations would benefit from this?

Those were the `Deposit` might not have been constant over time (increased by a runtime upgrade).

During the migration, the new `HoldConsideration` will result in a fixed `Balance` (based on a fixed `Footprint` and a constant `Convert` function).

Let's say that during the migration, the new `HoldConsideration` is expected to have `Balance = 10`. What would happen when for a particular account the stored `Deposit = 5`? After `Currency::unreserve(account, 5)`, when trying to create a new ticket to hold `Balance = 10`, it might fail (not enough balance). Currently, `HoldConsideration::new()` returns either `Ok(())` or `Err`. What to do if it is not possible to hold the balance?

> What alternatives we have?

The only alternative I see is to waive the deposit and not generate any ticket.

ggwpez:
After discussing it I think it would be less invasive to create a type solely for migrating from Currency. Something like `HoldConsiderationFromLegacyCurrency` or whatever that has this `new_from_exact` function.  
That way we dont have to modify the normal production version of `Hold / FreezeConsideration`

substrate/frame/support/src/traits/storage.rs
```diff
@@ -208,11 +208,17 @@ where
 /// treated as one*. Don't type to duplicate it, and remember to drop it when you're done with
 /// it.
 #[must_use]
-pub trait Consideration<AccountId>: Member + FullCodec + TypeInfo + MaxEncodedLen {
+pub trait Consideration<AccountId, Balance>: Member + FullCodec + TypeInfo + MaxEncodedLen {
```
gavofyork, line 211: Gah! No!

gavofyork: Considerations need not have anything to do with balances.

NachoPal:
Closing it. I'll give it a try to this alternative: https://github.com/paritytech/polkadot-sdk/pull/1813?notification_referrer_id=NT_kwDOALUd5rM3OTc4MzY4MDE0OjExODY5Njcw#issuecomment-1791081731

PR #1814, opened 2023-10-07
Identity Deposits Relay to Parachain Migration

joepetrowski:
The goal of this PR is to migrate Identity deposits from the Relay Chain to a system parachain.

The problem I want to solve is that `IdentityOf` and `SubsOf` both store an amount that's held in reserve as a storage deposit. When migrating to a parachain, we can take a snapshot of the actual `IdentityInfo` and sub-account mappings, but should migrate (off chain) the `deposit`s to zero, since the chain (and by extension, accounts) won't have any funds at genesis.

The good news is that we expect parachain deposits to be significantly lower (possibly 100x) on the parachain. That is, a deposit of 21 DOT on the Relay Chain would need 0.21 DOT on a parachain. This PR proposes to migrate the deposits in the following way:

1. Introduces a new pallet with two extrinsics: 
    - `reap_identity`: Has a configurable `ReapOrigin`, which would be set to `EnsureSigned` on the Relay Chain (i.e. callable by anyone) and `EnsureRoot` on the parachain (we don't want identities reaped from there).
    - `poke_deposit`: Checks what deposit the pallet holds (at genesis, zero) and attempts to update the amount based on the calculated deposit for storage data.
2. `reap_identity` clears all storage data for a `target` account and unreserves their deposit.
3. A `ReapIdentityHandler` teleports the necessary DOT to the parachain and calls `poke_deposit`. Since the parachain deposit is much lower, and was just unreserved, we know we have enough.

One awkwardness I ran into was that the XCMv3 instruction set does not provide a way for the system to teleport assets without a fee being deducted on reception. Users shouldn't have to pay a fee for the system to migrate their info to a more efficient location. So I wrote my own program and did the `InitiateTeleport` accounting on my own to send a program with `UnpaidExecution`. Have discussed an `InitiateUnpaidTeleport` instruction with @franciscoaguirre . Obviously any chain executing this would have to pass a `Barrier` for free execution.

TODO:

- [x] Confirm People Chain ParaId
- [x] Confirm People Chain deposit rates (determined in https://github.com/paritytech/polkadot-sdk/pull/2281)
- [x] Add pallet to Westend


burdges:
Identity could work with only state proofs, so without messages, but messages should be simpler initially.

substrate/frame/identity/src/lib.rs
```diff
@@ -959,6 +973,100 @@ pub mod pallet {
...
+			T::ReapOrigin::ensure_origin(origin)?;
+			let who = T::Lookup::lookup(target)?;
+
+			// `take` any storage items keyed by `target`
+			let id = <IdentityOf<T>>::take(&who).ok_or(Error::<T>::NotNamed)?;
+			let fields = id.info.additional.len() as u32;
+			let (subs_deposit, sub_ids) = <SubsOf<T>>::take(&who);
+			// check witness data
+			let actual_subs = sub_ids.len() as u32;
+			ensure!(actual_subs <= num_subs, Error::<T>::TooManySubAccounts);
```
georgepisaltu, line 1006: Why do we need to do this check? Is there any scenario where this function would be called with `num_subs` not being equal to `T::MaxSubAccounts`? We can make it account for that with weight refunds if necessary because IIUC in the current impl the caller will pay proportional to their `num_subs` of choice regardless of how many subs there actually are (and I doubt callers want to check each identity to see the exact numbers).

substrate/frame/identity/src/lib.rs
```diff
@@ -959,6 +973,100 @@ pub mod pallet {
...
+					} else {
+						return Err(Error::<T>::NoIdentity.into())
+					}
+				},
+			)?;
+
+			// Subs Deposit
+			let (current_subs_deposit, subs_of) = SubsOf::<T>::take(&target);
+			let new_subs_deposit = Self::subs_deposit(subs_of.len() as u32);
+			let _ = Self::rejig_deposit(&target, current_subs_deposit, new_subs_deposit)?;
```
georgepisaltu, line 1060:
```suggestion
			Self::rejig_deposit(&target, current_subs_deposit, new_subs_deposit)?;
```

substrate/frame/identity/src/lib.rs
```diff
@@ -959,6 +973,100 @@ pub mod pallet {
...
+						reg.deposit = new_id_deposit;
+						return Ok(new_id_deposit)
+					} else {
+						return Err(Error::<T>::NoIdentity.into())
+					}
+				},
+			)?;
+
+			// Subs Deposit
+			let (current_subs_deposit, subs_of) = SubsOf::<T>::take(&target);
```
georgepisaltu, line 1058: Let's use `SubsOf::<T>::try_mutate` instead of `take` and `insert`. The code below just changes the value in the map, but the entry stays.

polkadot/runtime/rococo/src/impls.rs
```diff
@@ -0,0 +1,115 @@
...
+	AccountId: Into<[u8; 32]> + Clone,
+{
+	fn on_reap_identity(who: &AccountId, fields: u32, subs: u32) -> DispatchResult {
+		// calculate deposit needed on parachains
+		// assume `relay_deposit / 100 + ED + buffer`
+		// buffer to cover tx fees on para, maybe ED
+		let para_basic_deposit = 1000 * CENTS / 100;
+		let para_field_deposit = 250 * CENTS / 100;
+		let para_sub_account_deposit = 200 * CENTS / 100;
+		let para_existential_deposit = EXISTENTIAL_DEPOSIT / 10;
```
georgepisaltu, line 42: I'd like to have these moved outside this function to constants and...

polkadot/runtime/rococo/src/impls.rs
```diff
@@ -0,0 +1,115 @@
...
+		let id_deposit =
+			para_basic_deposit.saturating_add(para_field_deposit.saturating_mul(fields as Balance));
+
+		let subs_deposit = para_sub_account_deposit.saturating_mul(subs as Balance);
+
+		let total_to_send = id_deposit
+			.saturating_add(subs_deposit)
+			.saturating_add(para_existential_deposit.saturating_mul(2));
```
georgepisaltu, line 50: ... and have this computation in a separate function which uses those constants
```
impl ToParachainIdentityReaper {
fn total_amount(additional_fields: usize, subs: usize) -> Balance
}
```
to segregate that from the actual XCM stuff here.

georgepisaltu (commented):
First pass, I haven't yet looked into the tests and benches but the approach looks solid to me.

polkadot/runtime/rococo/src/impls.rs
joepetrowski: Yes that's the plan. Once there is a people/identity chain runtime, we can just read the constants right from that.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3923274

substrate/frame/identity/src/lib.rs
georgepisaltu: I favor option 1 over option 2 and the reasons are:
- I don't see a scenario where I'd like to reap an identity but only if the number of subs is less than some arbitrary bound I set. In other words, it looks from the outside like `num_subs` is a protection for the user against excessive weight if they don't want to look into the identity they want to reap, which is not a problem if we do weight refunds.
- It would make sense to have the callers look into the struct off-chain if it was computationally intensive, but it's not the case here since we already have access to the number of subs just by accessing the list. We can do this for the user cheaply, it integrates into the system nicely with the weight refund, and it's a simpler interface IMO.

I'm not going to block on this. Just giving my 2c.

substrate/frame/identity/src/lib.rs
```diff
@@ -959,6 +998,126 @@ pub mod pallet {
...
+		}
+
+		/// Lock the pallet.
+		///
+		/// Origin must be the `LockerOrigin`.
+		#[pallet::call_index(17)]
+		#[pallet::weight(T::WeightInfo::lock_pallet())]
+		pub fn lock_pallet(origin: OriginFor<T>) -> DispatchResult {
+			T::LockerOrigin::ensure_origin(origin)?;
+			Locked::<T>::put(true);
```
georgepisaltu, line 1107: In order to be as explicit as possible with this, we can do `try_mutate` instead of `put` and return an error in case the flag was already `true`. Not a requirement, but more defensive. Same for the unlocking extrinsic.

joepetrowski: Returning an error and putting the same value are both no-ops though. What would it defend against?

georgepisaltu: While the effects in storage would be the same, if you run `lock_pallet` with the flag already set, it's more obvious/alarming to the caller if they get an error that something's not right, because they'd call that only if they thought the pallet wasn't already locked. Functionally it makes no difference, the fees would be charged anyway. I'm fine with whatever you decide here.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4062158

substrate/frame/identity/src/lib.rs
```diff
@@ -922,6 +961,129 @@ pub mod pallet {
...
+			});
+			Ok(())
+		}
+
+		/// Lock the pallet.
+		///
+		/// Origin must be the `LockerOrigin`.
+		#[pallet::call_index(17)]
+		#[pallet::weight(T::WeightInfo::lock_pallet())]
+		pub fn lock_pallet(origin: OriginFor<T>) -> DispatchResult {
```
bkchr, line 1071: To lock the pallet, we should just remove the possibility to call this pallet from the runtime. Not requiring any special logic in this pallet here. This can be done by just removing `Call` in the `Identity` line in `construct_runtime`.

substrate/frame/identity/src/lib.rs
```diff
@@ -922,6 +961,129 @@ pub mod pallet {
...
+		/// Parameters:
+		/// - `target`: The account for which to reap identity state.
+		///
+		/// Origin must be the `ReapOrigin`.
+		#[pallet::call_index(15)]
+		#[pallet::weight(T::WeightInfo::reap_identity(
+			T::MaxRegistrars::get(),
+			T::MaxSubAccounts::get()
+		))]
+		pub fn reap_identity(
```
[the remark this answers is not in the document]
joepetrowski: Because we have to "migrate" the deposits to a new runtime, and this pallet holds funds in reserve. To migrate that directly, it would mean minting funds in the genesis state of the new parachain, and burning funds in a Relay Chain runtime upgrade. IMO that is not very transparent; people should be able to see a transaction that moved their funds from one place to another.

That is the purpose of the `ReapIdentityHandler`: In the case of migrating to a parachain, it will teleport the funds needed for the deposit there. This will make the migration of funds transparent. It will also allow parachain deposit calculation to happen in the handler to ensure that the correct amount is teleported there.

substrate/frame/identity/src/lib.rs
joepetrowski: That was my initial approach but I think that is wrong for two reasons:

1. We still want to be able to call `reap_identity`, and
2. It will result in a lot of downtime because the unlocking action will need to be done via runtime upgrade. I was going to propose that the Fellowship be the `LockerOrigin` such that it could unlock the parachain once all deposits have been migrated. We can then change it to Root.

bkchr: > 2\. such that it could unlock the parachain once all deposits have been migrated

But this will require on the users actually doing this? Sounds like it takes quite some time?

joepetrowski: No, once the parachain launches where we want the deposits to migrate to, the `ReapOrigin` would be set to `EnsureSigned` so that anyone could call this and migrate every account. See the docs on the [Rococo Config](https://github.com/paritytech/polkadot-sdk/pull/1814/files#diff-da220c6c064cc047fef5d3b4a594ae30b8bbf48a29b740742bc9858285c7ba57R623).

Since the `IdentityInfo` and other state will be in the parachain's genesis (with the pallet locked at genesis), we can very quickly migrate all the deposits to the parachain and unlock the pallet there. Then we just remove the pallet from the Relay Chain entirely in the next runtime upgrade.

muharem (commented):
Given these constraints, the current approach seems sound to me. However, I'd like to revisit them and offer an alternative solution for consideration:

- user-friendliness: users shouldn't have to take any action for this migration.
- on-chain accountability for deposit transfers from the relay chain to the parachain.
- minimal downtime for write operations in the identity pallet.

to honor the first constraint, we must handle data migration seamlessly and introduce calls like `reap_identity` with a target account, enabling actions on behalf of the user.

to meet the second constraint, deposits must be migrated via XCM. simple migration without this is not an option.

for the third constraint, we should have the ability to lock and unlock the calls for the identity pallet on the new people chain through a specific extrinsic call. it's important to note that this constraint applies only to the people chain. on the relay chain, we can deactivate all pallet calls with an upgrade dedicated for the migration and completely remove the pallet with a subsequent upgrade.

consider this: what if we think it's acceptable to keep all identity pallet's calls deactivated on the people chain until the next upgrade, or if we can expedite that upgrade?

if so, we could introduce an additional custom pallet tailored for migration (outside of frame). this would allow the identity pallet to remain largely untouched, avoiding any breaking changes.

within the identity pallet, we'd need to unveil two public functions: one to clear identity and refund the deposit and another to calculate and take the deposit. This custom pallet would then implement `reap_identity` and `poke_deposit` using those functions.

@joepetrowski @bkchr

muharem (approved):
Looks good!

georgepisaltu (approved):
LGTM

substrate/frame/identity/src/lib.rs
```diff
@@ -906,4 +927,112 @@ impl<T: Config> Pallet<T> {
...
+	///
+	/// Return type is a tuple of the new Identity and Subs deposits, respectively.
+	///
+	/// NOTE: This function is here temporarily for migration of Identity info from the Polkadot
+	/// Relay Chain into a system parachain. It will be removed after the migration.
+	pub fn poke_deposit(
+		target: &T::AccountId,
+	) -> Result<(BalanceOf<T>, BalanceOf<T>), DispatchError> {
+		// Identity Deposit
+		ensure!(IdentityOf::<T>::contains_key(&target), Error::<T>::NoIdentity);
```
NachoPal, line 977: Already checked inside the `try_mutate`

NachoPal (approved):
LGTM, just some nits

polkadot/runtime/rococo/src/impls.rs
```diff
@@ -0,0 +1,147 @@
...
+			DepositAsset {
+				assets: Wild(AllCounted(1)),
+				beneficiary: Junction::AccountId32 { network: None, id: who.clone().into() }
+					.into_location()
+					.into(),
+			},
+			// Poke the deposit to reserve the appropriate amount on the parachain.
+			Transact {
+				origin_kind: OriginKind::Superuser,
+				require_weight_at_most: Weight::from_parts(2_000_000_000, 16_384),
```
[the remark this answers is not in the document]
joepetrowski: It was a "safe guess", but I will update before merging.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4337801

polkadot/runtime/rococo/src/impls.rs
```diff
@@ -0,0 +1,160 @@
...
+			impls::IdentityMigratorCalls::PokeDeposit,
+			weights::runtime_common_identity_migrator::WeightInfo as MigratorWeights,
+		};
+
+		let total_to_send = Self::calculate_remote_deposit(fields, subs);
+
+		// define asset / destination from relay perspective
+		let roc: MultiAssets =
+			vec![MultiAsset { id: Concrete(Here.into_location()), fun: Fungible(total_to_send) }]
+				.into();
```
bkchr, line 98: Here you also always only call `roc.inner().first().expect()`

bkchr (approved):
LGTM

polkadot/runtime/westend/src/impls.rs
```diff
@@ -0,0 +1,160 @@
...
+			impls::IdentityMigratorCalls::PokeDeposit,
+			weights::runtime_common_identity_migrator::WeightInfo as MigratorWeights,
+		};
+
+		let total_to_send = Self::calculate_remote_deposit(fields, subs);
+
+		// define asset / destination from relay perspective
+		let wnd: MultiAssets =
+			vec![MultiAsset { id: Concrete(Here.into_location()), fun: Fungible(total_to_send) }]
+				.into();
```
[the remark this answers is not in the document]
joepetrowski: Yeah you are right. Before I was doing another approach that took `MultiAssets`, but we don't need it.

polkadot/runtime/common/src/identity_migrator.rs
```diff
@@ -0,0 +1,305 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
```
xlc, line 1: can we stop adding more code to polkadot-runtime-common https://github.com/paritytech/polkadot-sdk/issues/2186

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-release-analysis-v1-4-0/5152/1

PR #1817, opened 2023-10-08
Use defaultConfig for pallet_contracts

pgherveou:


substrate/frame/contracts/src/tests.rs
```diff
@@ -457,28 +436,21 @@ parameter_types! {
 	pub static UnstableInterface: bool = true;
 }
 
+#[derive_impl(crate::config_preludes::TestDefaultConfig as crate::DefaultConfig)]
 impl Config for Test {
```
pgherveou, line 440: Some of the default trait are overridden here as we needs to use non const values to test different scenarios

substrate/frame/contracts/src/lib.rs
```diff
@@ -279,21 +288,25 @@ pub mod pallet {
...
 		/// construct a default cost schedule.
 		type WeightInfo: WeightInfo;
 
 		/// Type that allows the runtime authors to add new host functions for a contract to call.
+		#[pallet::no_default]
 		type ChainExtension: chain_extension::ChainExtension<Self> + Default;
 
 		/// Cost schedule and limits.
 		#[pallet::constant]
+		#[pallet::no_default]
```
pgherveou, line 309: some of these traits that take Self can't have default I suppose...

substrate/frame/contracts/src/lib.rs
```diff
@@ -312,10 +326,12 @@ pub mod pallet {
 		///
 		/// Changing this value for an existing chain might need a storage migration.
 		#[pallet::constant]
+		#[pallet::no_default]
 		type DepositPerByte: Get<BalanceOf<Self>>;
```
ggwpez, line 330: Is a default really not possible here @gupnik ? Thats a bit sad...

gupnik: You an use `pallet::no_default_bounds` here @pgherveou @ggwpez

pgherveou: got it and then in TestDefaultConfig I have to hard code the Balance type defined here?
https://github.com/paritytech/polkadot-sdk//blob/0bf6e1ab48f2a3918a95e6001da618a2ef0ed750/substrate/frame/balances/src/lib.rs#L231
So in a way they are all tightly coupled, which I guess is ok for Test config.

How would that work if we wanted to use that for something else than well defined TestConfig?

gupnik: Once you set `no_default_bounds`, internally it just removes all bounds and creates `type DepositPerByte` in the trait DefaultConfig. 

So, you can pretty much set it to anything as long as it makes sense within your TestDefaultConfig.

pgherveou: is it worthwhile adding dummy value for these fields though if they are going to be overridden anyway?

sam0x17 (approved):

substrate/frame/contracts/src/lib.rs
gupnik: If you want to ensure that users override this value, its absolutely fine to set `no_default`. Is there no sensible default for `DepositPerByte`?

gupnik:
@pgherveou Are we good to proceed here? Any blockers?

pgherveou:
> @pgherveou Are we good to proceed here? Any blockers?

Kind of forgot about it, will bring it up to date :)

pgherveou:
just merge conflicts, will re-go through the different types to pick sensible defaults

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5887491

pgherveou:
cc @gupnik  how does that look now?

substrate/frame/contracts/mock-network/src/parachain/contracts_config.rs
```diff
@@ -15,87 +15,19 @@
...
-	weights::Weight,
-};
-use frame_system::{pallet_prelude::BlockNumberFor, EnsureSigned};
-use pallet_xcm::BalanceOf;
-use sp_runtime::{traits::Convert, Perbill};
-
-pub const fn deposit(items: u32, bytes: u32) -> Balance {
-	items as Balance * 1 * CENTS + (bytes as Balance) * 1 * CENTS
-}
+use crate::parachain::RuntimeHoldReason;
```
pgherveou, line 18: This need to be in scope for the derive_impl macro to work, is that expected @gupnik?

gupnik: Yeah, correct. It assumes that these would be provided by the runtime v2 or construct_runtime macro.

substrate/frame/contracts/src/lib.rs
```diff
@@ -250,9 +250,10 @@ pub mod pallet {
 	#[pallet::storage_version(STORAGE_VERSION)]
 	pub struct Pallet<T>(_);
 
-	#[pallet::config]
+	#[pallet::config(with_default)]
 	pub trait Config: frame_system::Config {
 		/// The time implementation used to supply timestamps to contracts through `seal_now`.
+		#[pallet::no_default]
```
gupnik, line 256: Not possible to provide a default here?

substrate/frame/contracts/src/lib.rs
```diff
@@ -301,21 +310,25 @@ pub mod pallet {
...
 		/// construct a default cost schedule.
 		type WeightInfo: WeightInfo;
 
 		/// Type that allows the runtime authors to add new host functions for a contract to call.
+		#[pallet::no_default_bounds]
 		type ChainExtension: chain_extension::ChainExtension<Self> + Default;
 
 		/// Cost schedule and limits.
 		#[pallet::constant]
+		#[pallet::no_default]
```
gupnik, line 331: Tried with `no_default_bounds`?

substrate/frame/contracts/src/lib.rs
```diff
@@ -326,6 +339,7 @@ pub mod pallet {
 		///
 		/// This setting along with [`MaxCodeLen`](#associatedtype.MaxCodeLen) directly affects
 		/// memory usage of your runtime.
+		#[pallet::no_default]
```
gupnik, line 342: Same?

substrate/frame/contracts/src/tests.rs
```diff
@@ -334,34 +334,24 @@ parameter_types! {
...
+	type Block = Block;
 	type AccountId = AccountId32;
 	type Lookup = IdentityLookup<Self::AccountId>;
-	type Block = Block;
 	type AccountData = pallet_balances::AccountData<u64>;
 }
+
 impl pallet_insecure_randomness_collective_flip::Config for Test {}
+
+#[derive_impl(pallet_balances::config_preludes::TestDefaultConfig)]
```
gupnik, line 345: 👍

gupnik (approved):

substrate/frame/contracts/src/lib.rs
pgherveou: Nop the trait is generic over T
https://github.com/paritytech/polkadot-sdk/blob/pg/contract_default_trait/substrate/frame/contracts/src/schedule.rs?plain=1#L57-L66

substrate/frame/contracts/src/lib.rs
pgherveou: same, the trait is generic over T

substrate/frame/contracts/src/lib.rs
pgherveou: Yeah you would most likely want to override it, if you do anything that use that type in your tests, but I can provide an implementation that is unimplented!

Issue #239, opened 2022-12-02
Tracking issue for adding `try_state` hook to all pallets. 


kianenigma:
This is the tracking issue to add the `try_state` hook to as many pallets as we can. 

This hook was added in https://github.com/paritytech/substrate/pull/10174. The idea is as follows: 

```rust
/// Execute some checks to ensure the internal state of a pallet is consistent.
///
/// Usually, these checks should check all of the invariants that are expected to be held on all of
/// the storage items of your pallet.
///
/// This hook should not alter any storage.
pub trait TryState<BlockNumber> {
	/// Execute the state checks.
	fn try_state(_: BlockNumber) -> Result<(), &'static str>;
}
```

So, the ideas is to focus on invariants that must always be held in a pallet. A native, yet important example of this is that in the balances pallet, iterating over all users should yield the correct total issuance. 

Next, we want all pallets to call their `try_state` after each test. This can easily be achieved with a helper function like: 

https://github.com/paritytech/substrate/blob/dd8a34775aab9b362a5cc7552d5603c2fe812f82/frame/nomination-pools/src/mock.rs#L335

```rust
pub fn build_and_execute(self, test: impl FnOnce() -> ()) {
	self.build().execute_with(|| {
		test();
		// run the try-state of all pallets after each test.  
		<Runtime as TryState>::try_state(_, TryStateSelect::All).unwrap();
	})
}
```

Here is a list of current FRAME pallets, just for the sake of reference. Many of them won't need a `try_state` hook though. 

- [ ] alliance
- [ ] assets
- [ ] atomic-swap
- [ ] aura
- [ ] authority-discovery
- [ ] authorship
- [ ] babe
- [x] bags-list
- [ ] balances
- [ ] beefy
- [ ] beefy-mmr
- [ ] bounties
- [ ] child-bounties
- [x] collective
- [ ] contracts
- [ ] conviction-voting
- [ ] democracy
- [x] election-provider-multi-phase
- [ ] election-provider-support
- [ ] elections-phragmen
- [ ] examples
- [ ] fast-unstake
- [ ] gilt
- [ ] grandpa
- [ ] identity
- [ ] im-online
- [ ] indices
- [ ] lottery
- [ ] membership
- [ ] merkle-mountain-range
- [ ] message-queue https://github.com/paritytech/substrate/issues/13115
- [ ] multisig
- [x] nicks
- [ ] node-authorization
- [x] nomination-pools
- [ ] offences
- [ ] preimage
[...]

PR #1820, opened 2023-10-08
Adding `try_state` hook for `Treasury` pallet

wentelteefje:
This PR adds a `try_state` hook for the `Treasury` pallet.
Part of #239.
The invariants checked are:

1. `ProposalCount` must be $\geq$ number of elements in `Proposals`. The `ProposalCount` storage value is only increased, but never decreased, whereas individual proposals can be removed from the `Proposals` storage map.
2. `T::ProposalBondMinimum` * Number of proposals with non-zero bond $\leq \sum_{p \in \textrm{Proposals}} \textrm{p.bond}$.

Since the introduction of the dispatchable `spend`, a bond is no longer required (bond is set to zero in these cases). However, the old way of proposing using a bond is still available and we can filter for these by looking at proposals with $\textrm{p.bond} \neq 0 $.

3. (Optional) `T::ProposalBondMaximum` * Number of proposals with non-zero bond $\geq \sum_{p \in \textrm{Proposals}} \textrm{p.bond}$.

Analogous to 2., but only relevant when `T::ProposalBondMaximum` is not `None`.

4. Each `ProposalIndex` contained in `Approvals` should exist in `Proposals`. Note, that this
   automatically implies `Approvals.count() <= Proposals.count()`.
5. `SpendCount` must be $\geq$ number of elements in `Spends`. Follows the same reasoning as invariant 1.
6. For each spend entry contained in `Spends` we should have `spend.expire_at > spend.valid_from`.

### DIFF
--- a/substrate/frame/treasury/src/lib.rs
+++ b/substrate/frame/treasury/src/lib.rs
@@ -1026,6 +1026,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			// Must never be less than 0 but better be safe.
 			.saturating_sub(T::Currency::minimum_balance())
 	}
+
+	#[cfg(any(feature = "try-runtime", test))]
+	fn do_try_state() -> Result<(), sp_runtime::TryRuntimeError> {
+		// 1. ProposalCount >= Proposals.len()
+		ensure!(
+			Self::proposal_count() as usize >= <Proposals<T, I>>::iter_keys().count(),
+			"Number of proposals exceeds proposal count!"
+		);
+
+		// 2.
+		let total_amount_bonded = <Proposals<T, I>>::iter().fold(<BalanceOf<T, I>>::zero(), |a, (_index, proposal)| a + proposal.bond);
+		let n: BalanceOf<T, I> = <Proposals<T, I>>::iter_keys().count().saturated_into();
+		print!("{:?}", total_amount_bonded);
+		print!("{:?}", n);
+		ensure!(
+			T::ProposalBondMinimum::get() * n <= total_amount_bonded,
+			"Total amount bonded falls short of required minimum."
+		);
+
+		// 3.
+
+		// 4. Each ProposalIndex contained in Approvals should exist in Proposals
+		Approvals::<T, I>::get().iter().try_for_each(|proposal_index| -> DispatchResult {
+			ensure!(
+				Proposals::<T, I>::contains_key(proposal_index),
+				"Proposal indices in `Approvals` must also be contained in `Proposals`"
+			);
+			Ok(())
+		})?;
+
+		Ok(())
+	}
 }
 
 impl<T: Config<I>, I: 'static> OnUnbalanced<NegativeImbalanceOf<T, I>> for Pallet<T, I> {
--- a/substrate/frame/treasury/src/lib.rs
+++ b/substrate/frame/treasury/src/lib.rs
@@ -82,9 +82,6 @@ pub use benchmarking::ArgumentsFactory;
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 
-#[cfg(any(feature = "try-runtime", test))]
-use sp_runtime::traits::SaturatedConversion;
-
 use sp_runtime::{
 	traits::{AccountIdConversion, CheckedAdd, Saturating, StaticLookup, Zero},
 	Permill, RuntimeDebug,
@@ -1055,6 +1052,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 	/// Note, that this automatically implies Approvals.count() <= Proposals.count().
 	#[cfg(any(feature = "try-runtime", test))]
 	fn try_state_proposals() -> Result<(), sp_runtime::TryRuntimeError> {
+		use sp_runtime::traits::SaturatedConversion;
+		
 		let current_proposal_count = ProposalCount::<T, I>::get();
 		ensure!(
 			current_proposal_count as usize >= Proposals::<T, I>::iter().count(),
--- a/substrate/frame/treasury/src/lib.rs
+++ b/substrate/frame/treasury/src/lib.rs
@@ -462,7 +462,7 @@ pub mod pallet {
 		}
 
 		#[cfg(feature = "try-runtime")]
-		fn try_state(_: T::BlockNumberFor<T>) -> Result<(), sp_runtime::TryRuntimeError> {
+		fn try_state(_: frame_system::pallet_prelude::BlockNumberFor<T>) -> Result<(), sp_runtime::TryRuntimeError> {
 			Self::do_try_state()?;
 			Ok(())
 		}
--- a/substrate/frame/treasury/src/lib.rs
+++ b/substrate/frame/treasury/src/lib.rs
@@ -864,7 +864,7 @@ pub mod pallet {
 				// spend has expired and no further status update is expected.
 				Spends::<T, I>::remove(index);
 				Self::deposit_event(Event::<T, I>::SpendProcessed { index });
-				return Ok(Pays::No.into());
+				return Ok(Pays::No.into())
 			}
 
 			let payment_id = match spend.status {
@@ -881,11 +881,11 @@ pub mod pallet {
 				Status::Success | Status::Unknown => {
 					Spends::<T, I>::remove(index);
 					Self::deposit_event(Event::<T, I>::SpendProcessed { index });
-					return Ok(Pays::No.into());
+					return Ok(Pays::No.into())
 				},
 				Status::InProgress => return Err(Error::<T, I>::Inconclusive.into()),
 			}
-			return Ok(Pays::Yes.into());
+			return Ok(Pays::Yes.into())
 		}
 
 		/// Void previously approved spend.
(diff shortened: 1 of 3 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3902400

substrate/frame/treasury/src/lib.rs
```diff
@@ -855,7 +864,7 @@ pub mod pallet {
 				// spend has expired and no further status update is expected.
 				Spends::<T, I>::remove(index);
 				Self::deposit_event(Event::<T, I>::SpendProcessed { index });
-				return Ok(Pays::No.into())
+				return Ok(Pays::No.into());
```
ggwpez, line 867: Your editor is doing funny things again 😅

substrate/frame/treasury/src/lib.rs
```diff
@@ -1020,6 +1029,72 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	fn do_try_state() -> Result<(), sp_runtime::TryRuntimeError> {
+		ensure!(
+			ProposalCount::<T, I>::get() as usize >= Proposals::<T, I>::iter().count(),
+			"Actual number of proposals exceeds `ProposalCount`."
+		);
+
+		let (total_amount_bonded, non_zero_count) = Proposals::<T, I>::iter().fold(
+			(BalanceOf::<T, I>::zero(), 0usize),
+			|(sum, count), (_index, proposal)| {
+				if proposal.bond != BalanceOf::<T, I>::zero() {
```
ggwpez, line 1055:
```suggestion
				if !proposal.bond.is_zero() {
```

substrate/frame/treasury/src/lib.rs
```diff
@@ -1020,6 +1029,72 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			.try_for_each(|proposal_index| -> DispatchResult {
+				ensure!(
+					Proposals::<T, I>::contains_key(proposal_index),
+					"Proposal indices in `Approvals` must also be contained in `Proposals`."
+				);
+				Ok(())
+			})?;
+
+		ensure!(
+			SpendCount::<T, I>::get() as usize >= Spends::<T, I>::iter().count(),
```
ggwpez, line 1084: I think the value of `SpendCount` is more interesting in the fact, that it should describe the next free index in the `Spends` map.  
So you could check that all indices in the `Spend` map are strictly smaller than this.

substrate/frame/treasury/src/lib.rs
```diff
@@ -1020,6 +1029,72 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	/// 2. [`T::ProposalBondMinimum`] * Number of proposals with non-zero bond <= sum_{p in Proposals} p.bond.
+	/// 3. (Optional) [`T::ProposalBondMaximum`] * Number of proposals with non-zero bond >= sum_{p in Proposals} p.bond.
+	/// 4. Each [`ProposalIndex`] contained in [`Approvals`] should exist in [`Proposals`]. Note, that this
+	///    automatically implies Approvals.count() <= Proposals.count().
+	/// 5. [`SpendCount`] >= [`Spends`].count(). Follows the same reasoning as (1).
+	/// 6. For each spend entry contained in `Spends` we should have spend.expire_at > spend.valid_from.
+	#[cfg(any(feature = "try-runtime", test))]
+	fn do_try_state() -> Result<(), sp_runtime::TryRuntimeError> {
+		ensure!(
+			ProposalCount::<T, I>::get() as usize >= Proposals::<T, I>::iter().count(),
```
ggwpez, line 1048: Same thing that i wrote about the `SpendCount` applies here; akk `Proposal` indices must be smaller than it.

substrate/frame/treasury/src/tests.rs
```diff
@@ -217,16 +217,35 @@ impl Config for Test {
...
+		}
+		.assimilate_storage(&mut t)
+		.unwrap();
+		crate::GenesisConfig::<Test>::default().assimilate_storage(&mut t).unwrap();
+		let mut ext = sp_io::TestExternalities::new(t);
+		ext.execute_with(|| System::set_block_number(1));
+		ext
+	}
+
+	pub fn build_ext_and_execute(self, test: impl FnOnce() -> ()) {
```
ggwpez, line 243: You can also make this as freestanding function, since it seems to always be called on a default `ExtBuilder`.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3909966

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3910120

substrate/frame/treasury/src/lib.rs
wentelteefje: @ggwpez: I have added a check for this one and the equivalent one for `Spend`, but I kept the original check as well, since these are somewhat independent, I think. Depending on how much you trust the code that is.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3913768

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3913769

ggwpez (approved):
Looks very clean, thanks!  
Should probably try this before merging on Polkadot and Kusama states, so we know that it all works.

ggwpez:
bot fmt

wentelteefje:
cargo fmt seems to break markdown lists in comments. I guess I should fix these manually before merging too.

ggwpez:
> cargo fmt seems to break markdown lists in comments. I guess I should fix these manually before merging too.

Yea... lines should wrap at exactly 100 chars - also comments.

wentelteefje:
bot fmt

substrate/frame/treasury/src/tests.rs
```diff
@@ -217,16 +217,35 @@ impl Config for Test {
...
+		self.build().execute_with(|| {
+			test();
+			Treasury::do_try_state().expect("Storage invariants should hold.")
```
liamaharon, line 246: can you also write individual tests to make sure `try-state` fails in each scenario it's supposed to?

liamaharon (approved):
looking good, thanks!

substrate/frame/treasury/src/lib.rs
```diff
@@ -1020,6 +1032,110 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	fn do_try_state() -> Result<(), sp_runtime::TryRuntimeError> {
+		Self::try_state_proposals()?;
+		Self::try_state_spends()?;
+
+		Ok(())
+	}
+
+	/// ### Invariants of proposal storage items
+	///
+	/// 1. `ProposalCount` >= Number of elements in `Proposals`.
```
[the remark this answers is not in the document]
wentelteefje: Not sure if linking is of any use here, since the only place where this would show up in the rust-docs is on the pallet struct page (i think), but there the `try_state` doc string is automatically replaced by this: https://github.com/paritytech/polkadot-sdk/blob/5adcb3e10638c8abcfd532b9163990d584fc5832/substrate/frame/support/src/traits/hooks.rs#L374-L383

substrate/frame/treasury/src/tests.rs
wentelteefje: Are you suggesting to test the `try_state` implementation in the same way as the rest of the pallet? With the current implementation I don't see how that would work. As suggested here: https://github.com/paritytech/polkadot-sdk/issues/239 I'm running `try_state` after each regular test (as seen in the lines quoted by you). For example, if there was a test `fn try_state_invariant_1_works()`, which succeeds whenever invariant 1 is violated, then the following code would fail, because it leaves the storage in an invalid state:

```rust
try_state_invariant_1_works()  # <--- if this goes well...
Treasury::do_try_state().expect("Storage invariants should hold.") # <--- this will panick
```

A possible solution is to restore a valid state at the end of each test, but it seems to me that this  is a lot of overhead.

substrate/frame/treasury/src/lib.rs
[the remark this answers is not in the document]
liamaharon: It still is valuable even with no rust docs because it allow using 'go-to-definition' in the comments

substrate/frame/treasury/src/tests.rs
liamaharon: in that case, I'd prefer removing the 'blindly checking' that try state invariants hold with more deliberate tests.

the blind tests *might* catch a bug but they probably won't and it doesn't really give us any assurances about our code. deliberate tests should make us feel much more comfortable that the code works as expected.

wentelteefje: roger that.

wentelteefje:
I have added tests for all `try_state` scenarios. During this I also noticed that there's no longer a way to check invariants 2 & 3 from above (my initial approach had a mistake). That's because faulty entries with a zero bond cannot be distinguished from valid entries made using the `spend_local` dispatchable. So it was good to think about that some more during test writing :)

liamaharon:
Thanks, tests look great 👌

`Other` must be a String afaik. You might be able to avoid duplication be declaring them as `const`s, but not a big deal for the purposes of `try-state`.

Issue #685, opened 2023-03-23
PVF: Avoid clearing the artifacts cache on restart


mrcnski:
# ISSUE

## Overview

Avoid clearing the artifacts cache on restart when the host did not change

Issue extracted from discussion in https://github.com/paritytech/polkadot/pull/6551#pullrequestreview-1247883039. Let's continue here to unblock that PR.

## Previous discussion

We should imho avoid clearing the artifacts cache when the host did not change.  We could recompile only the parachain blocks when the host version did change, then lazily recompile the parathread ones.  

We should've timings of course, but we'll never stop people building wasm blobs that screw up build times intentionally.

Interpreting kinda works.  We have consensus upon who gets compiled vs interpreted, so interpreted then runs with different approval time parameters.  We could similarly adjust approval parameters to include recompiling parathreads each block.  This makes parathreads more expensive and second class though.  We could've parathreads that "buy" being compiled in advance like parachains. 

We do still have everyone compile the parathread when the PVF initially gets uploaded though, yes?  I'd think this suggests parathreads and parachains should be all be precompiled, which just makes uploading a PVF more expensive.  Implicitly then host upgrades become relatively more expensive, but this makes sense too.

_Originally posted by @burdges in https://github.com/paritytech/polkadot/pull/6551#discussion_r1069707945_
            

---

> We should imho avoid clearing the artifacts cache when the host did not change. We could recompile only the parachain blocks when the host version did change, then lazily recompile the parathread ones.

[...] I actually don't see why we want to clear the artifacts cache. On start-up, we could instead re-populate the `Artifacts` table from the compiled artifacts on disk -- the PVF hash should already be in the filename -- and re-start the 24-hour TTL timers for each artifact. Or we could even use the system's last-modified/accessed metadata for the files (with some sanity checks). Then instead of lazily re-compiling the PVFs, we would lazily delete the ones we end up not needing, which seems a lot more efficient. 🙂

_Originally posted by @mrcnski in https://github.com/paritytech/polkadot/pull/6551#discussion_r1069791385_

---

[...] Do I understand correctly that the proposal is to keep artifacts only if the node is not upgraded? In that case, it might work. But we should always purge the artifacts if the node is upgraded.

_Originally posted by @s0me0ne-unkn0wn in https://github.com/paritytech/polkadot/pull/6551#discussion_r1069934593_

## Related issue

https://github.com/paritytech/polkadot/issues/6941 (PVF preparation in advance)

PR #1828, opened 2023-10-09
Include polkadot version in artifact path

### LINKED ISSUE #695: PVF: include polkadot's version into the artifact path
# ISSUE

## Overview

Compilation artifact filenames should contain the polkadot version. This way if there is a node upgrade, an old version can't override an artifact for the new node.

eagr:
closes #695

Could potentially be helpful to preserving caches when applicable, as discussed in #685 


polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -75,6 +76,8 @@ pub struct ArtifactId {
 
 impl ArtifactId {
 	const PREFIX: &'static str = "wasmtime_";
+	// FIXME or just "v" instead?
+	const NODE_VERSION_PREFIX: &'static str = "polkadot_v";
```
mrcnski, line 80: I'd keep like this. Maybe in the future we can also add the `wasmtime` version. So it will look like `wasmtime_vXXX_polkadot_vYYY_...`

mrcnski (approved):
Thanks so much! If you'd like a tip and you have a Polkadot or Kusama address, you can add it to the PR body.

Format: `{kusama|polkadot|localtest} address: <SS58 Address>`
Doc: https://github.com/paritytech/substrate-tip-bot#pull-request-body
Example: https://github.com/paritytech/polkadot-sdk/pull/1720

Let me know if you want ideas for further contributions!

eagr:
What a pleasant surprise! I'll get an account right away.

Btw, I was trying to add some version check before the execution, but it feels better to put it in the cache purging instead (someday maybe). So any good next issue for me? @mrcnski

mrcnski:
> What a pleasant surprise! I'll get an account right away.

Nice! I'll tip you on https://github.com/paritytech/polkadot-sdk/pull/1774 as it's bigger and more likely for a tip to pass the referendum.

> Btw, I was trying to add some version check before the execution, but it feels better to put it in the cache purging instead (someday maybe). So any good next issue for me? @mrcnski

Cool! Since you're already working on this area, do you want to go ahead and implement artifact persistence? We would need the following:

1. https://github.com/paritytech/polkadot-sdk/issues/685
2. https://github.com/paritytech/polkadot-sdk/issues/677#issuecomment-1691882092 (hash after preparation, but only check it on node restart and not before every execution.)

eagr:
Is there anything left unsolved here? I'd like to start working on those issues. It'd be if this could be merged first as they are closely related.

s0me0ne-unkn0wn (approved):
Looks good, I'd appreciate a review from @bkchr as I remember there were discussions about moving around the node version constant... @mrcnski or did you agree already?

mrcnski:
Yeah, PR looks good but I would like to wait for @bkchr approval as well.

bkchr (approved):
Not the perfect place for the version. However, it is better than using an env variable.

PR #1801, opened 2023-10-05
Refactor `pallet-state-trie-migration` to `fungible::*` traits

bkontur:
## Summary

This PR consolidates `pallet-state-trie-migration` as a part of https://github.com/paritytech/polkadot-sdk/issues/226 / https://github.com/paritytech/polkadot-sdk/issues/171:

`pallet-state-trie-migration`:
- [x] replace `Currency` with `fungible` traits
- [x] run benchmarks
- [x] refactor to `DefaultConfig`

`pallet_nicks`:
- [x]  remove

others:
- [x] remove `as Fn*` or `asFun*` stuff based on discussion [here](https://github.com/paritytech/polkadot-sdk/issues/226#issuecomment-1822861445)

PR #1831, opened 2023-10-09
Improve features dev-ex

ggwpez:
Adds a config file that allows to run `zepter` without any arguments in the workspace to address all issues.  
A secondary workflow for the CI is provided as `zepter run check`. Both the formatting and linting are now in one check for efficiancy.

The latest version also detects some more things that `featalign` was already showing. 

Error message [in the CI](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3916205) now looks like this:  
```pre
...
crate 'test-parachains' (/Users/vados/Documents/work/polkadot-sdk/polkadot/parachain/test-parachains/Cargo.toml)
  feature 'std'
    must propagate to:
      parity-scale-codec
Found 55 issues (run with --fix to fix).
Error: Command 'lint propagate-feature' failed with exit code 1

Polkadot-SDK uses the Zepter CLI to detect abnormalities in the feature configuration.
It looks like one more more checks failed; please check the console output. You can try to automatically address them by running `zepter`.
Otherwise please ask directly in the Merge Request, GitHub Discussions or on Matrix Chat, thank you.

For more information, see:
  - https://github.com/paritytech/polkadot-sdk/issues/1831
  - https://github.com/ggwpez/zepter
```

TODO:
- [x] Check that CI fails correctly


bkontur (approved):

bkontur:
Maybe unrelated question, but should `dev-dependencies` be added  to `std` / `try-runtime` / `runtime-benchmarks` features?

Could this zepter feature detect `[dev-dependencies]` not to be propagated to `std` features?
E.g. here https://github.com/paritytech/polkadot-sdk/pull/1801/files#diff-57e966c3ef58bb810065f964f9d2d0c5ef141de756082abd9b0f877de98f78c3R25-L34
`pallet-balances = { path = "../balances" }`  and it was unnecessary added to `std` as `pallet-balances/std`.

But on the other hand,  `pallet-balances` has to be added to `try-runtime` as  `pallet-balances/try-runtime`.

bkontur:
Maybe unrelated question, but should `dev-dependencies` be added  to `std` / `try-runtime` / `runtime-benchmarks` features?

E.g. here https://github.com/paritytech/polkadot-sdk/pull/1801/files#diff-57e966c3ef58bb810065f964f9d2d0c5ef141de756082abd9b0f877de98f78c3R25-L34
`pallet-balances = { path = "../balances" }`  and it was unnecessary added to `std` as `pallet-balances/std`.
But on the other hand,  `pallet-balances` has to be added to `try-runtime` as  `pallet-balances/try-runtime`, is that ok?

ggwpez:
> Maybe unrelated question, but should `dev-dependencies` be added to `std` / `try-runtime` / `runtime-benchmarks` features?

I am not really sure how to handle dev and build dependencies. Normally we dont disable the default features for `dev` dependencies, so we dont need to add them to `std`. But for runtime-benchmarks and try-runtime it would make sense.

bkchr (approved):

sergejparity (approved):

sam0x17 (approved):

liamaharon (approved):
🎉

oleg-plakida (approved):

substrate/primitives/wasm-interface/Cargo.toml
```diff
@@ -23,5 +23,5 @@ sp-std = { path = "../std", default-features = false}
 
 [features]
 default = [ "std" ]
-std = [ "codec/std", "log", "sp-std/std", "wasmtime" ]
+std = [ "codec/std", "log", "log?/std", "sp-std/std", "wasmtime" ]
```
bkchr, line 26: This makes no sense. log and log?/std can be merged to log/std

bkchr (approved):

ggwpez:
IIUC https://github.com/paritytech/polkadot-sdk/pull/1870 moved some deps upstream. Now the problem is that the introduced dependency [`sp-bls12-381`](https://github.com/paritytech/substrate-curves/blob/133c9471824edcfc0e8702d24f2453e3972d6a8a/bls12_381/Cargo.toml) pulls in `sp-crypto-ec-utils` from the SDK, which in term [wants](https://crates.io/crates/sp-crypto-ec-utils/0.6.0/dependencies) `sp-std v11`.  
Since we did not backport the version bumps, the repo has version `8` and now pulls that stuff in twice - resulting in a build error here in the CI.

bkchr:
Does cargo unify deps which have a path and a version? So, would that be really fixed by merging the version bumps back to master?

ggwpez:
> Does cargo unify deps which have a path and a version? So, would that be really fixed by merging the version bumps back to master?

It does not seem to quite fix it, since we use one GH dependency [here](https://github.com/paritytech/polkadot-sdk/blob/38ef04eb53d43acdf504eea86ee3b1c8e08115e8/substrate/primitives/core/Cargo.toml#L60) which then pulls in duplicate versions again.  
I locally removed that dep and it worked, so hopefully it would be fixed if the bandersnatch also uses all the latest versions and gets published.

juangirini (approved):

PR #1832, opened 2023-10-10
upgrade to macro_magic 0.5.0

sam0x17:
# Description

Upgrades `macro_magic` to ~0.4.3~ 0.5.0, which introduces the ability to have `export_tokens` use the same name as the underlying item for its auto-generated macro name. Ultimately this will allow for better dev ux in our derive_impl feature.


substrate/frame/support/procedural/src/lib.rs
```diff
@@ -82,7 +82,7 @@ fn counter_prefix(prefix: &str) -> String {
 /// and the pallets that are used by the runtime.
 /// `Block` is the block type that is used in the runtime and `NodeBlock` is the block type
 /// that is used in the node. For instance they can differ in the extrinsics type.
-///
+///q
```
gupnik, line 85: Typo?

gupnik (approved):

substrate/frame/support/procedural/src/lib.rs
sam0x17: yes thank you lol

kianenigma (approved):

substrate/frame/support/Cargo.toml
```diff
@@ -30,7 +30,7 @@ sp-weights = { path = "../../primitives/weights", default-features = false}
 sp-debug-derive = { path = "../../primitives/debug-derive", default-features = false}
 sp-metadata-ir = { path = "../../primitives/metadata-ir", default-features = false}
 tt-call = "1.0.8"
-macro_magic = "0.4.2"
+macro_magic = "0.4.3"
```
kianenigma, line 33: I am getting some breaking changes downstream. Wasn't this a breaking change? then why is it only a minor bump?

substrate/frame/support/procedural/src/lib.rs
```diff
@@ -864,7 +864,12 @@ pub fn register_default_impl(attrs: TokenStream, tokens: TokenStream) -> TokenSt
 	let item_impl = syn::parse_macro_input!(tokens as ItemImpl);
 
 	// internally wrap macro_magic's `#[export_tokens]` macro
-	match macro_magic::mm_core::export_tokens_internal(attrs, item_impl.to_token_stream(), true) {
+	match macro_magic::mm_core::export_tokens_internal(
```
kianenigma, line 867: yeah it is evidnetly breaking.

nazar-pc:
macro_magic 0.4.3 should be yanked from crates.io, I can't compile runtimes anymore because runtime compilation doesn't have lock file and this new release essentially breaks all downstram builds for the ecosystem. Please honor semver.

substrate/frame/support/procedural/src/lib.rs
sam0x17: so 0.5.0 is published now going to yank 0.4.3

liamaharon (approved):

sam0x17:
Yeah sorry about that all didn't intend for the core/internal functions to be part of the public API (and therefore semver exempt) but our usage of them in substrate has basically forced them to be part of the public API so here we are lol. 0.4.3 has been yanked and re-published as 0.5.0.

That said, if our versions are locked properly across polkadot-sdk, this shouldn't be an issue, as any inadvertent minor update can be a supply chain attack, so wanted to flag that...

nazar-pc:
I just wiped `target` and other cargo directories and it pulled in 0.4.3 on recompilation, but I'm not sure how to resolve that with the way runtimes are being compiled right now.

sam0x17:
> I just wiped `target` and other cargo directories and it pulled in 0.4.3 on recompilation, but I'm not sure how to resolve that with the way runtimes are being compiled right now.

might have to clear cargo cache

nazar-pc:
> might have to clear cargo cache

I did, that was the problem (while 0.4.3 was still published). This is to the point of lock file being non-existant when WASM version of the runtime is compiled.

sam0x17:
I previously believed that a line like `macro_magic = "0.4.3"` locks the minor version but apparently in rust it should be `macro_magic = "= 0.4.3"`.

This means there are _many_ places in polkadot-sdk where we have third party deps that could easily be subject to supply chain attacks, so opening an issue for that ASAP

sam0x17:
cc @gupnik it's now `0.5.0`

Issue #1842, opened 2023-10-10
Make asset-conversion pallet free from a concept of a native currency


muharem:
The current design of the asset-conversion pallet manages both native and non-native assets by necessitating user-supplied implementations of `fungibles` and `fungible` traits. Moreover, the defined traits and implementations inherently incorporate the concept of a native asset.

This design presents a few drawbacks:
- limits the broader utility of the pallet;
- enforces a dependency on the native currency concept in other generic implementations using the pallet;
- demands the creation of a credit/imbalance type that amalgamates the credits of `fungible` and `fungibles` for `SwapCredit` impl ([here](https://github.com/paritytech/polkadot-sdk/pull/1677));

Proposed Changes:
- mandate only the `fungibles` trait implementation, and eliminate the type parameter with `fungible` bound;
- remove any traits or types that carry the native currency concept, provide an alternative;
- introduce `union_of` types, which supply a `fungibles` implementation from a pair of `fungible` and `fungibles`, as well as for `fungibles` and `fungibles` (will be also useful [here](https://github.com/paritytech/polkadot-sdk/blob/3f5edc52b22f39516c3e1d3bef0023bd6a48daf6/cumulus/parachains/runtimes/assets/common/src/local_and_foreign_assets.rs#L72));

muharem:
I will work on this next week, need this issue to reference in few TODO comments

PR #1677, opened 2023-09-22
pallet-asset-conversion: Swap Credit

muharem:
Introduces a swap implementation that allows the exchange of a credit (aka Negative Imbalance) of one asset for a credit of another asset.

This is particularly useful when a credit swap is required but may not have sufficient value to meet the ED constraint, hence cannot be deposited to temp account before. An example use case is when XCM fees are paid using an asset held in the XCM executor registry and has to be swapped for native currency.

Additional Updates:
- encapsulates the existing `Swap` trait impl within a transactional context, since partial storage mutation is possible when an error occurs;
- supplied `Currency` and `Assets` impls must be implemented over the same `Balance` type, the `AssetBalance` generic type is dropped. This helps to avoid numerous type conversion and overflow cases. If those types are different it should be handled outside of the pallet;
- `Box` asset kind on a pallet level, unbox on a runtime level - here [why](https://substrate.stackexchange.com/questions/10039/boxed-argument-of-a-dispatchable/10103#10103);
- `path` uses `Vec` now, instead of `BoundedVec` since it is never used in PoV;
- removes the `Transfer` event due to it's redundancy with the events emitted by `fungible/s` implementations;
- modifies the `SwapExecuted` event type;

related issue: 
- https://github.com/paritytech/polkadot-sdk/issues/105

related PRs:
- (required for) https://github.com/paritytech/polkadot-sdk/pull/1845
- (caused) https://github.com/paritytech/polkadot-sdk/pull/1717

// DONE make the pallet work only with `fungibles` trait and make it free from the concept of a `native` asset - https://github.com/paritytech/polkadot-sdk/issues/1842

PR #1845, opened 2023-10-11
XCM WeightTrader: Swap Fee Asset for Native Asset

muharem:
Implements an XCM executor `WeightTrader`, facilitating fee payments in any asset that can be exchanged for a native asset.

A few constraints need to be observed:
- `buy_weight` and `refund` operations must be atomic, as another weight trader implementation might be attempted in case of failure.
- swap credit must be utilized since there isn’t an account to which an asset of some class can be deposited with a guarantee to meet the existential deposit requirement. Also, operating with credits enhances the efficiency of the weight trader - https://github.com/paritytech/polkadot-sdk/pull/1677 

related PRs:
- (depends) https://github.com/paritytech/polkadot-sdk/pull/2031
- (depends) https://github.com/paritytech/polkadot-sdk/pull/1677
- (caused) https://github.com/paritytech/polkadot-sdk/pull/1847
- (caused) https://github.com/paritytech/polkadot-sdk/pull/1876

// DONE: impl `OnUnbalanced` for a `fungible/s` credit
// DONE: make the trader free from a concept of a native currency and drop few fallible conversions. related issue - https://github.com/paritytech/polkadot-sdk/issues/1842
// DONE: tests


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3931392

substrate/frame/balances/src/impl_currency.rs
```diff
@@ -100,6 +100,11 @@ mod imbalances {
 			mem::forget(self);
 			(Self(first), Self(second))
 		}
+		fn extract(&mut self, amount: T::Balance) -> Self {
```
muharem, line 103: there is a separate PR for this https://github.com/paritytech/polkadot-sdk/pull/1847

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3931678

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3934181

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4150618

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4150975

cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -70,9 +73,15 @@ parameter_types! {
 	pub PoolAssetsPalletLocation: MultiLocation =
 		PalletInstance(<PoolAssets as PalletInfoAccess>::index() as u8).into();
 	pub CheckingAccount: AccountId = PolkadotXcm::check_account();
+	pub StakingPot: AccountId = CollatorSelection::account_id();
 	pub TreasuryAccount: Option<AccountId> = Some(TREASURY_PALLET_ID.into_account_truncating());
 }
 
+// Foreign chain account alias into local accounts according to a hash of their standard
+// description.
```
franciscoaguirre, line 81:
```suggestion
// The way to derive sovereign accounts for both local and foreign pallets
```

cumulus/primitives/utility/src/lib.rs
```diff
@@ -286,23 +291,238 @@ impl<
...
+						"`total_fee.asset` must be equal to `refund.asset`",
+						(self.total_fee.asset(), refund.asset())
+					);
+				});
+				return None
+			},
+		};
+
+		refund_asset.fun = refund.peek().into().into();
+		drop(refund);
```
franciscoaguirre, line 483: Why do you have to manually call drop?

cumulus/primitives/utility/src/lib.rs
```diff
@@ -286,23 +291,238 @@ impl<
...
+/// ### Parameters:
+/// - `Target`: the asset into which the user's payment will be exchanged using `SwapCredit`.
+/// - `SwapCredit`: mechanism used for the exchange of the user's payment asset into the `Target`.
+/// - `WeightToFee`: weight to the `Target` asset fee calculator.
+/// - `Fungibles`: registry of fungible assets.
+/// - `FungiblesAssetMatcher`: utility for mapping [`MultiAsset`] to `Fungibles::AssetId` and
+///   `Fungibles::Balance`.
+/// - `OnUnbalanced`: handler for the fee payment.
+/// - `AccountId`: the account identifier type.
+pub struct SwapFirstAssetTrader<
```
franciscoaguirre, line 314: Why have this here and not in xcm-builder? Looks like a trader other teams would like to use

franciscoaguirre (approved):

cumulus/primitives/utility/src/lib.rs
muharem: have no strong opinion here, we just have another weight trader in this crate.

cumulus/primitives/utility/src/lib.rs
muharem: I wanna do it explicitly, to emphasize, there is some custom logic on that drop.

cumulus/primitives/utility/src/lib.rs
ggwpez, line 314: Why is it interesting to have it be the first? Could it not be a generic one where the index of the asset is stored in the struct?

cumulus/primitives/utility/src/lib.rs
```diff
@@ -286,23 +291,238 @@ impl<
...
+
+		match self.total_fee.subsume(credit_out) {
+			Err(credit_out) => {
+				// error may occur if `total_fee.asset` differs from `credit_out.asset`, which does
+				// not apply in this context.
+				defensive!(
+					"`total_fee.asset` must be equal to `credit_out.asset`",
+					(self.total_fee.asset(), credit_out.asset())
+				);
+				return Ok(payment)
```
ggwpez, line 423: I guess it should error in this case, but cant since we already did the token swap, or?

cumulus/primitives/utility/src/lib.rs
```diff
@@ -286,23 +291,238 @@ impl<
...
+			target: "xcm::weight",
+			"SwapFirstAssetTrader::refund_weight weight: {:?}, self.total_fee: {:?}",
+			weight,
+			self.total_fee,
+		);
+		if self.total_fee.peek().is_zero() {
+			// noting yet paid to refund.
+			return None
+		}
+		let mut refund_asset = if let Some(asset) = &self.last_fee_asset {
```
ggwpez, line 445: This `last_fee_asset` seems a bit odd to me, since it would not work multiple times. Eg:  
```js
buy()
buy()
refund()
refund() // uses a different asset here
```
is this intended?

cumulus/primitives/utility/src/lib.rs
```diff
@@ -286,23 +291,238 @@ impl<
...
+		let refund = self.total_fee.extract(refund_amount);
+		let refund = match SwapCredit::swap_exact_tokens_for_tokens(
+			vec![Target::get(), refund_swap_asset],
+			refund,
+			None,
+		) {
+			Ok(refund_in_target) => refund_in_target,
+			Err((refund, _)) => {
+				// return an attempted refund back to the `total_fee`.
+				let _ = self.total_fee.subsume(refund).map_err(|refund| {
```
ggwpez, line 470: You could also clone the `total_fee` right before extraction to avoid this defensive path.

cumulus/primitives/utility/src/lib.rs
muharem: yes, but this will also mean that the impl of `SwapCredit` is completely off, because we ask to swap the `swap_asset` for `Target`, but it did return us something different from `Target`.
but now I think I better return an error, it is more likely there will be no other trader in a chain for this case, and this will be within the transactional context. also an error should give more awareness.

cumulus/primitives/utility/src/lib.rs
muharem: yes, I chose a simper solution here. if you paid with asset `a` first, and `b` later. we do the refund with the last asset you used. this should cover most of the cases.
also `refund_weight`, returns only a single `MultiAsset` is succeeded.

cumulus/primitives/utility/src/lib.rs
muharem: since it is a credit, I can not and should not copy it

cumulus/primitives/utility/src/lib.rs
muharem: the implementation we have always passes the payment with one asset. it just make this impl a bit simpler. but I agree, we can ask for an index here.

cumulus/primitives/utility/src/lib.rs
muharem: returning an error now

cumulus/primitives/utility/src/lib.rs
muharem: the implementation of xcm-executor I mean ^.
actually I can only iterate over all btree, since the key is the asset id itself. if sounds good to you, I would keep it as it is now.

cumulus/primitives/utility/src/lib.rs
ggwpez: Okay I see. @franciscoaguirre do you know if this is according to the XCM specs?

cumulus/primitives/utility/src/lib.rs
ggwpez: Yea i dont mind. Just an idea.

ggwpez (approved):
Im not an expert on this, but LGTM.

cumulus/primitives/utility/src/lib.rs
franciscoaguirre: Nothing in the XCM spec says it can't be like that. Since it only returns one asset, we're stuck with doing something like this. It would have to be well documented, that's all.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4775562

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4775775

muharem:
I have just pushed few new commits. They do not change any behaviour to new tooling. Only master merge, same swap trader for rococo as for westend and emulated integration tests.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4870329

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4880435

PR #1850, opened 2023-10-11
PVF worker: bump landlock, update ABI docs

mrcnski:


polkadot/node/core/pvf/common/src/worker/security.rs
```diff
@@ -466,5 +475,38 @@ pub mod landlock {
 
 			assert!(handle.join().is_ok());
 		}
+
+		// Test that checks whether landlock under our ABI version is able to truncate files.
+		#[test]
+		fn restricted_thread_can_truncate_file() {
+			// TODO: This would be nice: <https://github.com/rust-lang/rust/issues/68007>.
+			if !check_is_fully_enabled() {
+				return
```
s0me0ne-unkn0wn, line 484: It's no good the test silently passed in absence of Landlock, but, iiuc, we don't have a better option anyway?

s0me0ne-unkn0wn (approved):
Looks good, thanks for keeping it up to date!

alexggh (approved):

polkadot/node/core/pvf/common/src/worker/security.rs
mrcnski: Yeah, don't believe so. We have a CI job that runs on an old Linux kernel and tests outright failing there is not ideal.

PR #1851, opened 2023-10-11
Small optimisation to `--profile dev` wasm builds

liamaharon:
`wasm-builder` was adjusted to default to building wasm blobs in `release` mode even when cargo is in `debug` because `debug` wasm is too slow.

A side effect of this was `.compact` and `.compact.compressed` getting built when the dev is running build in `debug`, adding ~5s to the build time of every wasm runtime.

I think it's reasonable to assume if the dev is running `debug` build they want to optimise speed and do not care about the size of the wasm binary. Compacting a blob has negligible impact on its actual performance.

In this PR, I adjusted the behavior of the wasm builder so it does not produce `.compact` or `.compact.compressed` wasm when the user is running in `debug`. The builder will continue to produce the bloaty wasm in release mode unless it is overriden with an env var.

As suggested by @koute in review, also refactored the `maybe_compact_wasm_and_copy_blobs` into multiple funuctions, and renamed things to better support RISC-V in the future.

---

There is no `T-runtime` label so @KiChjang told me to put `T1-FRAME` :)


franciscoaguirre (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4055725

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4055889

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -697,69 +754,77 @@ fn build_project(
...
+		Some(WasmBinary(compact_compressed_blob_path))
 	} else {
 		build_helper::warning!(
-			"Writing uncompressed wasm. Exceeded maximum size {}",
+			"Writing uncompressed blob. Exceeded maximum size {}",
 			CODE_BLOB_BOMB_LIMIT,
 		);
-
-		false
+		println!("{}", colorize_info_message("Skipping blob compression"));
```
koute, line 826: Do we need this `println` here, considering just before it a warning is emitted?

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -110,78 +110,111 @@ fn crate_metadata(cargo_manifest: &Path) -> Metadata {
...
+		copy_bloaty_blob(&project, &build_config.blob_build_profile, &bloaty_blob_name);
+
+	// Try to compact and compress the bloaty blob, if the *outer* profile wants it.
+	//
+	// This is because, by default the inner profile will be set to `Release` even when the outer
+	// profile is `Debug`, because the blob built in `Debug` profile is too slow for normal
+	// development activities.
+	let (compact_blob_path, compact_compressed_blob_path) =
+		match build_config.outer_build_profile.wants_compact() {
+			true => {
```
koute, line 154: `match`ing on a `bool` seems strange, why not an `if`? Will reduce one level of indentation.

koute (approved):

substrate/utils/wasm-builder/src/wasm_project.rs
liamaharon: Good point. Will remove

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4056645

Issue #1728, opened 2023-09-27
Rococo beefy error: BadJustification("Duplicate consensus engine ID") on appending justification


PierreBesson:
We are seeing those beefy error logs on all Rococo validators :

```
[...]

PR #1852, opened 2023-10-11
sc-consensus-beefy: improve gossip logic

acatangiu:
- Remove cached messages used for deduplication in `GossipValidator` since they're already deduplicated in upper layer `NetworkGossip`.
- Add cache for "justified rounds" to quickly discard any further (even if potentially different) justifications at the gossip level, once a valid one (for a respective round) is submitted to the worker.
- Add short-circuit in worker `finalize()` method to not attempt to finalize same block multiple times (for example when we get justifications for same block from multiple components like block-import, gossip or on-demand).
- Change a test which had A LOT of latency in syncing blocks for some weird reason and would only run after ~150seconds. It now runs instantly.

Fixes https://github.com/paritytech/polkadot-sdk/issues/1728

### DIFF
--- a/substrate/client/consensus/beefy/src/communication/gossip.rs
+++ b/substrate/client/consensus/beefy/src/communication/gossip.rs
@@ -278,8 +278,7 @@ where
 		self.known_peers.lock().note_vote_for(*sender, round);
 
 		// Verify general usefulness of the message.
-		// We are going to discard old votes right away (without verification)
-		// Also we keep track of already received votes to avoid verifying duplicates.
+		// We are going to discard old votes right away (without verification).
 		{
 			let filter = self.gossip_filter.read();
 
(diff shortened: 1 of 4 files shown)


svyatonik (approved):

substrate/client/consensus/beefy/src/communication/gossip.rs
```diff
@@ -293,10 +290,6 @@ where
 				Consider::Accept => {},
 			}
 
-			if filter.is_known_vote(round, &msg_hash) {
-				return Action::Keep(self.votes_topic, benefit::KNOWN_VOTE_MESSAGE)
-			}
```
serban300, line 298: Nit: Should we also update the comment above ?
```
// Also we keep track of already received votes to avoid verifying duplicates.
```

serban300 (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3953910

Issue #403, opened 2023-08-08
Mint fixed portion of inflation directly into the treasury


gpestana:
It's been proposed and [discussed](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301) a change to the current inflation model where a fixed portion of the inflation is minted directly into the treasury, while the remaining portion is distributed as rewards as per the current mechanism. The main goal is to ensure the sustainability of the treasury in the long term ([more info in this thread](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301/1)).

In addition, it would be helpful to allow governance to change the percentage of inflation minted into the treasury.

### Implementation suggestions

1. [`fn era_payout`](https://github.com/paritytech/substrate/blob/master/frame/staking/src/pallet/impls.rs#L443) calculates the portion of payout that is distributed among stakers (`validator_payout`) and the treasury based on perthing `TreasuryFraction`, in storage.
2. The validator payout stored in `ErasValidatorReward` is `total_payout - total_payout * TreasuryFraction`. From here on, the logic of the stakers' rewards remains the same.
3. The treasury portion for that era is minted at the end of the era into `Treasury::account_id()`  (another option would be to cache treasury rewards and mint into the treasury pot every X eras, but probably not worth the effort/storage).
4. New extrinsic to update `TreasuryFraction` perthing, gated by root.

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/revising-polkadots-ideal-staking-rate-calculation/3897/1

gpestana:
For the record, we'll implement these changes using the parameters pallet (under development) (https://github.com/paritytech/polkadot-sdk/pull/2061) so that the era payout parameters can be changed using selected origins.

PR #1660, opened 2023-09-21
 Implements a % cap on staking rewards from era inflation

gpestana:
This PR implements an (optional) cap of the era inflation that is allocated to staking rewards. The remaining is minted directly into the [`RewardRemainder`](https://github.com/paritytech/polkadot-sdk/blob/fb0fd3e62445eb2dee2b2456a0c8574d1ecdcc73/substrate/frame/staking/src/pallet/mod.rs#L160) account, which is the treasury pot account in Polkadot and Kusama.

The staking pallet now has a percent storage item, `MaxStakersRewards`, which defines the max percentage of the era inflation that should be allocated to staking rewards. The remaining era inflation (i.e. `remaining = max_era_payout - staking_payout.min(staking_payout * MaxStakersRewards))` is minted directly into the treasury. 

The `MaxStakersRewards` can be set by a privileged origin through the `set_staking_configs` extrinsic.

**To finish**
- [x] run benchmarks for westend-runtime

Replaces https://github.com/paritytech/polkadot-sdk/pull/1483
Closes https://github.com/paritytech/polkadot-sdk/issues/403

PR #1859, opened 2023-10-12
Implements maximum validator payout per era

gpestana:
This PR implements an (optional) cap of the era inflation that is allocated to staking rewards. This mechanism can be used to  guarantee a minimum treasury inflow as per [this discussion](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301).

The (percent) maximum validators payout can be set through `SystemCall::set_storage`.

_Note_: The `storage` parameter type in the runtime configs only sets the pre-`set_storage` default value. Once the value is set on-chain, the default will be a noop and not anymore a source of truth.

**Todo** (once merged)
- [ ] update the fellowship runtimes accordingly (`era_payout` impl included)

Related to and possibly replacing https://github.com/paritytech/polkadot-sdk/pull/1660
Closes https://github.com/paritytech/polkadot-sdk/issues/403

### DIFF
--- a/polkadot/runtime/test-runtime/src/lib.rs
+++ b/polkadot/runtime/test-runtime/src/lib.rs
@@ -74,7 +74,7 @@ use sp_runtime::{
 		SaturatedConversion, StaticLookup, Verify,
 	},
 	transaction_validity::{TransactionPriority, TransactionSource, TransactionValidity},
-	ApplyExtrinsicResult, KeyTypeId, Perbill,
+	ApplyExtrinsicResult, KeyTypeId, Perbill, Percent,
 };
 use sp_staking::SessionIndex;
 #[cfg(any(feature = "std", test))]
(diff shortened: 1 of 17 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3946274

polkadot/runtime/test-runtime/src/lib.rs
```diff
@@ -352,7 +353,8 @@ impl pallet_staking::Config for Runtime {
 	type SlashDeferDuration = SlashDeferDuration;
 	type AdminOrigin = frame_system::EnsureNever<()>;
 	type SessionInterface = Self;
-	type EraPayout = pallet_staking::ConvertCurve<RewardCurve>;
+	type MaxValidatorsPayout = MaxValidatorsPayout;
```
kianenigma, line 356:
```suggestion
```

If you were to do this approach, I don't think there is any need to pass `MaxValidatorsPayout` to the staking pallet. Whatever cap is needed, should already be applied to `RewardCurve` or whatever type is implementing `type EraPayout`.

kianenigma (changes requested):
A bit wrong because of https://github.com/paritytech/polkadot-sdk/pull/1859/files#r1382983349, but in any case we probably won't need it.

gpestana:
Closing due to https://github.com/paritytech/polkadot-sdk/issues/403#issuecomment-1794635026

PR #1869, opened 2023-10-13
Refactor `ValidationError`

### LINKED ISSUE #733: Preparation errors should no longer be a `InvalidCandidate` variant
Since we have prechecking enabled, we no longer want to treat preparation error as if the candidate was invalid. Hence [`PrepareError`](https://github.com/paritytech/polkadot/blob/9cf76e37f2783f347b639231a591cb2517991c68/node/core/pvf/src/error.rs#L91) should not be a variant of the `InvalidCandidate` enum, but instead be consolidated with [`ValidationError::InternalError](https://github.com/paritytech/polkadot/blob/9cf76e37f2783f347b639231a591cb2517991c68/node/core/pvf/src/error.rs#L83).

In particular `InternalError` represents spurious errors, while `PrepareError` represents deterministic errors. This differentiation is still useful, so instead of making those two errors one, I would suggest to only move `PrepareError` to the `ValidationError` enum, but reconsider the naming as both `InternalError` and `PrepareError` are prepare errors.

eagr:
closes #733

### DIFF
--- a/polkadot/node/core/candidate-validation/src/lib.rs
+++ b/polkadot/node/core/candidate-validation/src/lib.rs
@@ -630,7 +630,7 @@ async fn validate_candidate_exhaustive(
 	}
 
 	match result {
-		Err(ValidationError::InternalError(e)) => {
+		Err(ValidationError::Internal(e)) => {
 			gum::warn!(
 				target: LOG_TARGET,
 				?para_id,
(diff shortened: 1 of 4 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3954934

PR #1871, opened 2023-10-14
Adding `try_state` hook for Tips pallet 

Doordashcon:
Part of https://github.com/paritytech/polkadot-sdk/issues/239.

Invariant

1. The number of entries in `Tips` should be equal to `Reasons`.
2. If `OpenTip.finders_fee` is true, then `OpenTip.deposit` should be greater than zero.
3. Reasons exists for each Tip[`OpenTip.reason`], implying equal length of storage. 

polkadot address: 12zsKEDVcHpKEWb99iFt3xrTCQQXZMu477nJQsTBBrof5k2h


substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +623,25 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	///
+	/// This should be valid before or after each state transition of this pallet.
+	///
+	/// ## Invariants:
+	///
+	/// If `OpenTip.finders_fee` is true, then OpenTip.deposit should be greater that zero.
+	#[cfg(any(feature = "try-runtime", test))]
+	pub fn do_try_state() -> Result<(), TryRuntimeError> {
+		for tips in Tips::<T, I>::iter_keys() {
+			let open_tip = Tips::<T, I>::get(&tips).unwrap();
```
ggwpez, line 637:
```suggestion
			let open_tip = Tips::<T, I>::get(&tips).expect("All map keys are valid; qed");
```

substrate/frame/tips/src/tests.rs
```diff
@@ -192,7 +192,7 @@ impl Config for Test {
 	type Tippers = TenToFourteen;
 	type TipCountdown = ConstU64<1>;
 	type TipFindersFee = TipFindersFee;
-	type TipReportDepositBase = ConstU64<1>;
+	type TipReportDepositBase = ConstU64<0>;
```
liamaharon, line 195: Please don't change the DepositBase to 0. It's important that it is non-zero so we can ensure it works correctly in the test.

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +623,25 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			Tips::<T, I>::insert(hash, new_tip)
 		}
 	}
+
+	/// Ensure the correctness of the state of this pallet.
+	///
+	/// This should be valid before or after each state transition of this pallet.
```
liamaharon, line 629:
```suggestion
	/// This should be valid before and after each state transition of this pallet.
```

ggwpez (commented):
(We should not invest much effort into this pallet since it is Gov1)

Not sure if this is actually correct, since `DataDepositPerByte` and `TipReportDepositBase` can be configured to zero.  
Maybe also add a check to the  `integrity_test` that `TipReportDepositBase` is not zero since otherwise it does not take a deposit for the entry into the `Reasons` and `Tips` map.

Another invariant would be that the `Reason` for each `Tips` exists.

substrate/frame/tips/src/tests.rs
Doordashcon: The change to zero was needed to test the invariant theory.

Doordashcon:
@ggwpez Thanks for the review
> (We should not invest much effort into this pallet since it is Gov1)

Noted

> Not sure if this is actually correct, since DataDepositPerByte and TipReportDepositBase can be configured to zero.
Maybe also add a check to the integrity_test that TipReportDepositBase is not zero since otherwise it does not take a deposit for the entry into the Reasons and Tips map.

🤝 will include this.

> Another invariant would be that the Reason for each Tips exists.

I previously included this as an Invariant [here](https://github.com/paritytech/substrate/pull/13515/files#diff-6aa2af6f083184cb0859b5e78f5e8fea0b9a8be4ce4d3794cf7683ca7bf4c3d9), seems like it's universal logic in the function calls and also tests for this would always pass.

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +623,25 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	///
+	/// If `OpenTip.finders_fee` is true, then OpenTip.deposit should be greater that zero.
+	#[cfg(any(feature = "try-runtime", test))]
+	pub fn do_try_state() -> Result<(), TryRuntimeError> {
+		for tips in Tips::<T, I>::iter_keys() {
+			let open_tip = Tips::<T, I>::get(&tips).unwrap();
+			if open_tip.finders_fee {
+				ensure!(
+					!open_tip.deposit.is_zero(),
+					TryRuntimeError::Other("finder's fee not present")
```
wentelteefje, line 641: I think these errors should be as descriptive as possible. For example: 
```suggestion
					TryRuntimeError::Other("Tips with `finders_fee` should have non-zero `deposit`.")
```

substrate/frame/tips/src/tests.rs
```diff
@@ -654,3 +654,15 @@ fn report_awesome_and_tip_works_second_instance() {
...
+
+#[test]
+fn try_state_invariant_works() {
+	new_test_ext().execute_with(|| {
+		use frame_support::pallet_prelude::DispatchError::Other;
+		Balances::make_free_balance_be(&Treasury::account_id(), 101);
+
+		assert_ok!(Tips::report_awesome(RuntimeOrigin::signed(0), b"".to_vec(), 3));
+
+		assert_eq!(Tips::do_try_state(), Err(Other("finder's fee not present")))
```
wentelteefje, line 666: w.r.t my comment above you would need to fix it here as well, i.e.
```suggestion
		assert_eq!(Tips::do_try_state(), Err(Other("Tips with `finders_fee` should have non-zero `deposit`.")))
```

substrate/frame/tips/src/tests.rs
liamaharon: We shouldn't test one thing at the expense of something else.

If you want a tip with zero deposit for the purposes of testing this invariant, please insert it manually into storage.

Doordashcon: Thanks for the review, will change accordingly.

substrate/frame/tips/src/tests.rs
Doordashcon: done 🤝

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +630,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			Tips::<T, I>::insert(hash, new_tip)
 		}
 	}
+
+	/// Ensure the correctness of the state of this pallet.
+	///
+	/// This should be valid before and after each state transition of this pallet.
+	///
+	/// ## Invariants:
+	/// 1. Reasons exists for each Tip[`OpenTip.reason`], this implies equal lenth of entries in storage.
```
wentelteefje, line 639: > 1. Reasons exists for each Tip[`OpenTip.reason`], **this implies equal lenth of entries in storage.**

This statement really only implies that $tips.len() \leq reasons.len()$. However, In your code you are actually ensuring the length of both maps is equal `ensure!(reasons.len() == tips.len(), ...)`. So you're really checking for three invariants.

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +630,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	/// 2. If `OpenTip.finders_fee` is true, then OpenTip.deposit should be greater that zero.
```
wentelteefje, line 640:
```suggestion
	/// 2. If `OpenTip.finders_fee` is true, then OpenTip.deposit should be greater than zero.
```

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +630,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	#[cfg(any(feature = "try-runtime", test))]
+	pub fn do_try_state() -> Result<(), TryRuntimeError> {
+		let reasons = Reasons::<T, I>::iter_keys().collect::<Vec<_>>();
+		let tips = Tips::<T, I>::iter_keys().collect::<Vec<_>>();
+
+		ensure!(reasons.len() == tips.len(), TryRuntimeError::Other("Equal length of entries in 'Tips' and 'Reasons` Storage"));
```
wentelteefje, line 646: Maybe try to use only one set of ticks :)
```suggestion
		ensure!(reasons.len() == tips.len(), TryRuntimeError::Other("Equal length of entries in `Tips` and `Reasons` Storage"));
```

substrate/frame/tips/src/lib.rs
Doordashcon: Indeed

They should be of equal length as is in the function calls, an entry to `Tips` means an entry to `Reasons`.

ggwpez (approved):

substrate/frame/tips/src/lib.rs
```diff
@@ -611,4 +630,42 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	/// 2. Reasons exists for each Tip[`OpenTip.reason`].
+	/// 3. If `OpenTip.finders_fee` is true, then OpenTip.deposit should be greater than zero.
+	#[cfg(any(feature = "try-runtime", test))]
+	pub fn do_try_state() -> Result<(), TryRuntimeError> {
+		let reasons = Reasons::<T, I>::iter_keys().collect::<Vec<_>>();
+		let tips = Tips::<T, I>::iter_keys().collect::<Vec<_>>();
+
+		ensure!(
+			reasons.len() == tips.len(),
+			TryRuntimeError::Other("Equal length of entries in `Tips` and `Reasons` Storage")
```
gpestana, line 649:
```suggestion
			TryRuntimeError::Other("Number of `Tips` and of `Reasons` is not the same.")
```

It seems that the error should say exactly what went wrong, not how it would have been correct.

substrate/frame/tips/src/tests.rs
```diff
@@ -655,3 +656,88 @@ fn report_awesome_and_tip_works_second_instance() {
...
+		assert_eq!(Tips::do_try_state(), Err(Other("no reason for this tip")));
+	})
+}
+
+#[test]
+#[should_panic = "`TipReportDepositBase` should not be zero"]
+fn zero_base_deposit_prohibited() {
+	new_test_ext().execute_with(|| {
+		TipReportDepositBase::set(0);
+		Tips::integrity_test();
```
gpestana, line 741: Maybe we could include the integrity tests at the end of the `build_and_execute` so that it is executed for all tests?

gpestana (approved):
Added a couple of comments, but approving. thanks!

substrate/frame/tips/src/tests.rs
Doordashcon: Not sure this would be needed, the integrity test is explicitly for when an instance of `Tips` sets `TipReportDepositBase` to zero. An isolated use of the integrity hook should be sufficient?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4411405

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4412859

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4421285

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4429443

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4665009

Doordashcon:
ready for final review :)

prdoc/pr_1871.prdoc
```diff
@@ -0,0 +1,12 @@
+title: Adding `try-state` hook to tips pallet
+
+doc:
+  - audience: Runtime User
+    description: |
+      Enforces the following invariants;
+      1. The number of entries in Tips should be equal to Reasons.
+      2.  If OpenTip.finders_fee is true, then OpenTip.deposit should be greater than zero.
+      3.  Reasons exists for each Tip[OpenTip.reason], implying equal length of storage.
```
ggwpez, line 9:
```suggestion
      2. If OpenTip.finders_fee is true, then OpenTip.deposit should be greater than zero.
      3. Reasons exists for each Tip[OpenTip.reason], implying equal length of storage.
```

substrate/frame/tips/src/lib.rs
```diff
@@ -465,6 +469,21 @@ pub mod pallet {
 			Ok(())
 		}
 	}
+
+	#[pallet::hooks]
+	impl<T: Config<I>, I: 'static> Hooks<BlockNumberFor<T>> for Pallet<T, I> {
+		fn integrity_test() {
+			assert!(
+				!T::TipReportDepositBase::get().is_zero(),
```
ggwpez, line 477: Please add a comment to `TipReportDepositBase` that mentions this restriction.

substrate/frame/tips/src/tests.rs
gpestana: Sorry, my bad., I misread the line. I meant running the `do_try_state` checks at the end of each test. That's a good way to ensure that the state is valid at the end of tests.

substrate/frame/tips/src/lib.rs
Doordashcon: > Btw i am not sure if we should even restrict it.

Depends on the likelihood a user would not set `DataDepositPerByte` to Zero, which would result in a zero finders fee on `report_awesome` when there is no restriction.

https://github.com/paritytech/polkadot-sdk/blob/c8112e2c6fd9f1f5e7cb767b682fda746fc00520/substrate/frame/tips/src/lib.rs#L263-L265

ggwpez (approved):
LGTM, lets get it merged soon. Its open since quite a while now 😅

Issue #993, opened 2023-08-14
[CI] Add link checker


ggwpez:
Need a link checker for markdown and Rust. Possibly also Adoc. [lychee](https://github.com/lycheeverse/lychee) was used in https://github.com/paritytech/polkadot-sdk/issues/134 and could be tried first.

PR #1875, opened 2023-10-15
[CI] Add link checker

ggwpez:
Adding link checker to the CI (closes https://github.com/paritytech/polkadot-sdk/issues/993). It would be nice to have the docs people own this and extend accordingly.  
Currently all known-bad links are excluded, but we should one-by-one include those as well until all are fixed.  

This check now ensures that 1) no new broken links are introduced into `.rs` files and 2) that no old links break unnoticed.


alvicsam (approved):

sergejparity (approved):

EgorPopelyaev (approved):

.config/lychee.toml
```diff
@@ -0,0 +1,55 @@
...
+	# Rate limited - GitHub likes to throw this.
+	429
+]
+
+exclude_path = [ "./target" ]
+
+
+exclude = [
+  # Place holer
+  "http://visitme/",
```
chevdor, line 26: You already have that one below.
```suggestion
```

.github/workflows/check-links.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0 (22. Sep 2023)
+
+      - name: Lychee link checker
+        uses: lycheeverse/lychee-action@2ac9f030ccdea0033e2510a23a67da2a2da98492 # for v1.8.0 (15. May 2023)
+        with:
+          args: >-
+            --config .config/lychee.toml
+            --no-progress
+            './**/*.rs'
```
chevdor, line 33: Could we add checks for `<root>/prdoc/*.prdoc` as well ?

chevdor (approved):
It would be nice to cover the prdoc files as well but happy to have the current state as a start already.

.github/workflows/check-links.yml
ggwpez: done

PR #1883, opened 2023-10-16
fix: correct the `network` field of sovereign account id evaluation function

realnimish:
This PR fixes the hardcoded `network` param used in the helper functions of `xcm-simulator/example`. While working with the provided example network, I found that the sovereign account id obtained by the helper functions didn't match the actual account ids. The runtime is configured with `RelayNetwork: Kusama` while the helper functions uses hardcoded value `None` for `AccountId32.network` field resulting in incorrect account ids.


franciscoaguirre:
This is only a problem since the account derivation is done via hashing the location, `Account32Hash`, which is the old way of doing it. We should change this to use the `HashDescription` family and this would be solved. The network parameter is meant to be a sanity check, not to change the actual account of the location.

bkchr:
Stale.

PR #1884, opened 2023-10-16
Workspace maintenance

ggwpez:
Changes:
- Add missing crate to the workspace
- Remove versions from local dependency links

Maybe it is finally worth it to add this scrip to the CI to find these things earlier: [check-deps.py](https://github.com/ggwpez/substrate-scripts/blob/master/import-runtime-repos/check-deps.py).  

@paritytech/ci what would be the best location for that check?  
It takes only a second to run, so maybe we can squeeze it into one of the existing checks?  
Otherwise creating a new GH workflow feels a bit wasteful... maybe i can group it with https://github.com/paritytech/polkadot-sdk/pull/1831

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ members = [
 	"cumulus/parachains/integration-tests/emulated/assets/asset-hub-westend",
 	"cumulus/parachains/integration-tests/emulated/bridges/bridge-hub-rococo",
 	"cumulus/parachains/integration-tests/emulated/common",
+	"cumulus/parachains/pallets/collective-content",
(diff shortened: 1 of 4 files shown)


alvicsam:
I think it makes sense to run it in one job with zepter as you suggested. I'd put the script into the `.gitlab` folder.

ggwpez:
> I think it makes sense to run it in one job with zepter as you suggested. I'd put the script into the `.gitlab` folder.

Okay thanks! I will do it as another MR when the stuff here is sorted out.

ordian (approved):

PR #1889, opened 2023-10-16
allow liquidity pools to be deleted.

gilescope:
If an asset is destroyed and then another asset is recreated with the same asset id, then currently it is currently not possible to use the asset-conversion pallet with the recreated asset as the old (non-functioning) pool is in the way.

This PR allows anyone (signed) to clean up the pallet so that the pool can be created. It also allows pools to be destroyed if they are created but no liquidity has yet been placed within them (note that if some liquidity has been placed in the pool then unless one of the assets is destroyed then the pool can not be destroyed).

(There is an AssetCallback api that could allow a runtime to call this base functionality.
That could be done in a separate PR, but the possibility of it being a multi-block destruction could complicate that.)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3973135

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3975257

bkchr:
Stale

PR #1898, opened 2023-10-17
Use prebuilt try-runtime binary in CI

liamaharon:
`cargo install` takes a long time in CI. We want to run it relatively frequently without chewing through so much compute (see https://github.com/paritytech/ci_cd/issues/771) so I added automatic binary releases to the try-runtime-cli repo.

A small added benefit is we can use it in our existing `on-runtime-upgrade` checks, which should cut their execution time by about half.


alvicsam (approved):
Awesome, thank you!

rcny (approved):

rcny:
Note that once caching lands in the monorepo, it will also speed up such CI patterns.

ggwpez (approved):

PR #1904, opened 2023-10-17
Publish `penpal-runtime` crate

NachoPal:
Remove `publish = false` to publish the crate

### DIFF
--- a/cumulus/parachains/runtimes/testing/penpal/Cargo.toml
+++ b/cumulus/parachains/runtimes/testing/penpal/Cargo.toml
@@ -7,7 +7,6 @@ license = "Unlicense"
 homepage = "https://substrate.io"
 repository.workspace = true
 edition.workspace = true
-publish = false
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


ggwpez (approved):

gupnik (approved):

bkchr:
Wait. Why do we publish a test crate?

NachoPal:
> Wait. Why do we publish a test crate?

We need it to be used when `xcm-emulator` integrations tests live outside `polkadot-sdk`.

`penpal-runtime` is used to tests system parachains interactions with 3rd party parachains.

xlc:
why you are testing with test runtime instead of a real runtime? what are you testing?

NachoPal:
> why you are testing with test runtime instead of a real runtime? what are you testing?

Penpal simulates a non-system parachain. Right now it is still a pretty dummy runtime and we just use as recipient of XCM messages. Plan is to improve it to be able to properly handle the received assets.

PR #1908, opened 2023-10-17
upgraded review bot to v2.1.0

Bullrich:
Upgraded to version 2.1.0 which has paritytech/review-bot#94, a change in the logic of the action to overcome some problems with permissions coming from PRs from forks

For this, we needed to divide the actions into two files:
- A first action that triggers on PRs and reviews and uploads the PR number.
- A second action which is triggered under the completion of the first one and runs as the action normally runs (but won't have any problems regarding permissions because it is triggered from the master branch)


Bullrich:
@pavelsupr we can enable environment secrets again with this new implementation.

mordamax (approved):

rzadp (approved):

sergejparity (approved):

rcny (approved):

EgorPopelyaev (approved):

PR #1788, opened 2023-10-04
Fix Asset Hub collator crashing when starting from genesis

georgepisaltu:
Fixes #1566 

# Description

When running `polkadot-parachain --chain asset-hub-*` with no DB, the node would crash because it would try to access Aura API which wasn't available in the genesis `shell` runtime that Asset Hub started with. More details on the issue [here](https://github.com/paritytech/polkadot-sdk/issues/1566#issuecomment-1740649027).

This PR makes the node start with a relay chain driven consensus which doesn't call into `AuraApi::slot_duration`, then after the API becomes available (after the runtime is upgraded to the proper Asset Hub), it uses the Aura consensus.

# Checklist

- [ ] My PR includes a detailed description as outlined in the "Description" section above
- [ ] My PR follows the [labeling requirements](CONTRIBUTING.md#Process) of this project (at minimum one label for `T`
  required)
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)

PR #1911, opened 2023-10-17
Cumulus: Allow aura to use initialized collation request receiver

skunert:
When launching our [small network](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/zombienet/examples/small_network.toml) for testing the node was crashing here shortly after launch:
https://github.com/paritytech/polkadot-sdk/blob/5cdd819ed295645958afd9d937d989978fd0c84e/polkadot/node/collation-generation/src/lib.rs#L140

After changes in #1788  for the asset hub collator we are waiting for blocks of the shell runtime to pass before we initialize aura. However, this means that we attempted to initialize the collation related relay chain subsystems twice, leading to the error.

I modified Aura to let it optionally take an already initialized stream of collation requests.


bkchr (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-release-analysis-v1-3-0/4614/1

PR #1912, opened 2023-10-17
[draft] Move syncing code from `sc-network-common` to `sc-network-sync`

dmitry-markin:
This PR moves syncing-related code from `sc-network-common` to `sc-network-sync`.

Unfortunately, some parts are tightly integrated with networking, so they were left in `sc-network-common` for now:

1. `SyncMode` in `common/src/sync.rs` (used in `NetworkConfiguration`).
2. `BlockAnnouncesHandshake`, `BlockRequest`, `BlockResponse`, etc. in `common/src/sync/message.rs` (used in `src/protocol.rs` and `src/protocol/message.rs`).

More substantial refactoring is needed to decouple syncing and networking, including getting rid of the hardcoded sync protocol.

Also, `ChainSync` trait doesn't make much sense as the only implementation is always accessed directly and actually has some essential methods not covered by the trait. Entire `chain_sync.rs` file should probably be deleted before clearing the draft flag on this PR.

CC @altonen


altonen (approved):
I don't think this should be silent. There are at least two external teams that are working on their own syncing implementations so this will likely affect them. At least the removal of `ChainSync` is something that should be mentioned, with a note that there will be some kind of new `ChainSync` trait in the near future.

substrate/client/network/sync/src/lib.rs
```diff
@@ -206,24 +206,65 @@ struct GapSync<B: BlockT> {
 
 /// Action that [`engine::SyncingEngine`] should perform after reporting imported blocks with
 /// [`ChainSync::on_blocks_processed`].
-enum BlockRequestAction<B: BlockT> {
+pub enum BlockRequestAction<B: BlockT> {
```
bkchr, line 209: Why do we need to make this public?

bkchr (approved):
Looking good. Just a little bit confused why some types are being made public?

bkchr:
> with a note that there will be some kind of new `ChainSync` trait in the near future.

What is the status of completely getting rid off sync in the networking crate? Because ideally there should not exist a chain sync trait at all.

substrate/client/network/sync/src/lib.rs
dmitry-markin: It's consumed by `SyncingEngine`, that owns `ChainSync`.

dmitry-markin:
> > with a note that there will be some kind of new `ChainSync` trait in the near future.
> 
> What is the status of completely getting rid off sync in the networking crate? Because ideally there should not exist a chain sync trait at all.

There won't be `ChainSync` trait in networking for sure. There may be some new traits for syncing in `sc-network-sync` though.

bkchr:
> There may be some new traits for syncing in `sc-network-sync` though.

Do we really need to make sync itself generic? I don't think so? As this is some self contained implementation.

substrate/client/network/sync/src/lib.rs
bkchr: But isn't this in the same crate?

altonen:
> > There may be some new traits for syncing in `sc-network-sync` though.
> 
> Do we really need to make sync itself generic? I don't think so? As this is some self contained implementation.

That is not the current plan at least. We are most likely going to introduce some traits for the strategies that we discussed earlier, simply to make the code easier to maintain and plug additional syncing implementations without major headache.

substrate/client/network/sync/src/lib.rs
dmitry-markin: Technically, `SyncingEngine` is in the same root module as `ChainSync`, because `ChainSync` is defined in `lib.rs`. So `pub` is not needed at all (even as `pub(crate)`). Semantically, these types are public for `SyncingEngine`, so I feel inclined to highlight this with either `pub` or `pub(crate)`.

`pub(crate)` is probably a better option, but may be it's also worth moving `ChainSync` to a separate `chain_sync.rs` file.

@bkchr would it be better to move `ChainSync` and its types to a separate file and use `pub(crate)` for types consumed by `SyncingEngine`?

dmitry-markin: > @bkchr would it be better to move `ChainSync` and its types to a separate file and use pub(crate) for types consumed by SyncingEngine?

Done. PR looks scary now due to moved code, but overall situation with `ChainSync` should have improved.

dmitry-markin: @bkchr @altonen Could you have a look if moving `ChainSync` to a separate file worth it?

bkchr: > `pub(crate)` is probably a better option, but may be it's also worth moving `ChainSync` to a separate `chain_sync.rs` file.

Yeah sorry for these dumb questions :P I should have realized that the module is private to the crate. Pub is fine by me here.

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -0,0 +1,3406 @@
...
+				return Err(BadPeer(*who, rep::BAD_BLOCK))
+			}
+		}
+	}
+
+	Ok(blocks.first().and_then(|b| b.header.as_ref()).map(|h| *h.number()))
+}
+
+#[cfg(test)]
+mod test {
```
bkchr, line 2440: If you are at it, why not also move them into their own file?

bkchr (approved):
Moving to its own file sounds reasonable!

substrate/client/network/sync/src/lib.rs
dmitry-markin: Done

substrate/client/network/sync/src/chain_sync.rs
dmitry-markin: Done

substrate/client/network/sync/src/lib.rs
altonen: I was thinking this reordering could've been done in another PR but whatever. `SyncingEngine` should be moved to `sync/src/lib.rs` but that can be done in another PR.

dmitry-markin: Why `SyncingEngine` should be moved to the root module and not kept in it's own module, may be reexporting it in the root module?

Issue #1576, opened 2023-09-14
PVF: Refactor argument passing


mrcnski:
### Host data

Right now if you want to add some data that is passed from the host startup to a worker, there is a long chain of calls it has to go through. It should be refactored somehow. For example, maybe we could pass around some super-objects to make adding new fields a one-line change. It would make such changes much easier in the future.

For an example, in [this commit](https://github.com/paritytech/polkadot-sdk/pull/1373/commits/8cedd7bb0578f5193b4e7f36fc2250a895d55b22) we added just one new parameter, `cache_path`, and look how noisy the commit is.

### Worker data

On the worker-side, we often pass around the worker kind, pid, and dir together. Perhaps there could be one `WorkerInfo` struct.

### Other?

In general I felt like the PVF code is in dire need of cleaning up. Any other ideas would be highly appreciated.

eagr:
Just skimming through the pvf codebase. Just some random ideas, not sure if they are applicable. :))

**Structural**

Further flatten the structure. Some should be nested, but nested is rarely better in my experience. For instance,

```
src/
  prepare/
    mod.rs
    pool.rs
    queue.rs
    worker_intf.rs
```
could this be  flattened to
```
src/
  prepare_pool.rs
  prepare_queue.rs
  prepare_spawn.rs
```
This is actually more revealing to me, esp. when the pool and queue are both used in host on their own.

Also, the whole thing feels like it could be
```
common/
host/
execute-worker/
prepare-worker
```
which seems be easier to grok.

**Dev UX**

Rather than having inlined test modules, maybe consider moving them out, such that modifying the test code won't cause Cargo to recompile the lib crates. For smaller projects, this probably doesn't matter. But you know how crazy the complation time is. If this makes sense to you, I'm willing to test it out.

mrcnski:
Thanks for the ideas @eagr! Overall I am not sure about them, but maybe I've just grown really accustomed to the way things are right now. I don't think we should change things up without a clear benefit, as it can be quite disruptive, but let me know if I am missing the benefits that you see. After a discussion we can put it to a vote with the other contributors. :)

1. About flattening and adding a prefix, it really seems to be strictly worse. 😅 Structure is helpful so why remove it? Is it difficult navigating the hierarchy? IMO it helps keep things separate when we have prepare-specific and execute-specific code.
2. About having a separate `host` directory, you mean having this kind of structure, and `pvf` itself contains no code?
```
pvf/
  host/
  execute-worker/
[...]

PR #1192, opened 2023-08-27
Tracking/limiting memory allocator

s0me0ne-unkn0wn:
Original PR and discussion: https://github.com/paritytech/polkadot/pull/7206

Closes #718

This PR implements a global memory allocator that wraps a real memory allocator (Jemallocator in this very implementation, but could be any) to track a peak allocation value during the preparation job. The value is then forwarded to the PVF host and made observable in Prometheus metrics.

The allocator also enforces a memory limit if one is set when the allocation tracking is started. When the peak allocation value overflows the limit, an observer-provided failure handler is called. In that handler, the worker sends an out-of-memory error to the host, and then the worker process is exited.

The host treats the out-of-memory condition as a deterministic prepare error. It rips the worker process (which is exiting anyway, but a bit of cleanup is required on the host side) and votes "nah" for the PVF being pre-checked.

Issue #626, opened 2023-05-23
PVF worker version mismatch during development


s0me0ne-unkn0wn:
During the development process, if the code of the `polkadot-node-core-pvf-worker` package is changed, it can lead to a worker version mismatch error on the next run. Probably caused by the `polkadot-node-code-pvf` package not being recompiled and is keeping the old version. Should be fixed as it impedes development.

This does not affect either production or CI builds, as well as any other clean builds.

Some additional details are available here: https://github.com/paritytech/polkadot/pull/7253#issuecomment-1555014422

mrcnski:
Haven't hit a version mismatch yet, nor received more reports about it. :] And [this PR](https://github.com/paritytech/polkadot-sdk/pull/1495) makes the check more lenient:

> now we won't get a node/worker version mismatch between releases (during development). Workers should still get rebuilt on every commit so breakages should be rare, but we just won't enforce that, which is fine for most dev.

Please re-open if there are more reports!

PR #1918, opened 2023-10-17
Preserve artifact cache unless stale

### LINKED ISSUE #685: PVF: Avoid clearing the artifacts cache on restart

eagr:
closes #685 

cc @mrcnski @s0me0ne-unkn0wn


eagr:
Do you think `Artifacts::insert_prepared()` should be made `pub(crate)`?

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -55,18 +55,56 @@
...
+				i += 1;
+			}
+			while i < L {
+				cat[i] = Y[i - XL];
+				i += 1;
+			}
+			cat
+		}
+
+		// SAFETY: safe because x and y are ensured to be valid
```
mrcnski, line 98: I'm not so confident that this is always safe because unicode has a lot of edge cases. Both inputs are unicode, and I'm not 100% sure it's sound to concatenate unicode bytes, even if separately they are valid unicode strings. I asked AI:

> Concatenating the bytes of two Unicode strings is not a safe operation because Unicode strings can contain multi-byte characters. If you concatenate two strings at a byte level, you risk splitting a multi-byte character, which can lead to incorrect and unpredictable results.

Of course, the way it's used right now is fine because we only have ASCII inputs, but there's a chance of this macro being mis-used in the future. I would just go with a non-const solution, e.g. use `format!` inside a `OnceCell` or something. There's not really a need to complicate things with a macro.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,79 @@ pub enum ArtifactState {
...
+	pub(crate) fn new() -> Self {
+		Self { inner: HashMap::new() }
+	}
+
+	#[cfg(test)]
+	pub(crate) fn len(&self) -> usize {
+		self.inner.len()
+	}
+
+	/// Create a table with valid artifacts and prune the invalid ones.
```
mrcnski, line 224: Could use some more detail in this doc. e.g. What does "valid" and "invalid" mean?

mrcnski (changes requested):
Thanks a lot! Add your address to the PR description if you want another tip!

polkadot/node/core/pvf/src/artifacts.rs
eagr: Yeah, the unsafe is unnessary here.

mrcnski:
> Do you think `Artifacts::insert_prepared()` should be made `pub(crate)`?

Why do you think so? AFAICT, `Artifact` is not available outside the crate anyway, so should be fine as-is.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -88,38 +123,34 @@ impl ArtifactId {
...
+		);
+		cache_path.join(file_name)
+	}
+
 	/// Tries to recover the artifact id from the given file name.
-	#[cfg(test)]
-	pub fn from_file_name(file_name: &str) -> Option<Self> {
-		use polkadot_core_primitives::Hash;
-		use std::str::FromStr as _;
+	pub(crate) fn from_file_name(file_name: &str) -> Option<Self> {
```
mrcnski, line 136: nit: Does this have to be `pub` at all? I think it can even be private.

eagr:
> Why do you think so? 

By making it `pub(crate)`, it could be used in `Artifacts`. Right now I'm using an ad hoc version of it there.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -55,18 +55,56 @@
...
 use polkadot_parachain_primitives::primitives::ValidationCodeHash;
 use polkadot_primitives::ExecutorParamsHash;
 use std::{
 	collections::HashMap,
 	path::{Path, PathBuf},
+	str::FromStr as _,
 	time::{Duration, SystemTime},
 };
 
+macro_rules! concat_const {
```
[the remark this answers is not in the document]
eagr: Now that I think of it, I could probably generalize the macro.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,79 @@ pub enum ArtifactState {
...
+	pub async fn new_and_prune(cache_path: &Path) -> Self {
+		let mut artifacts = Self { inner: HashMap::new() };
+		Self::prune_and_insert(cache_path, &mut artifacts).await;
+		artifacts
+	}
+
+	// FIXME eagr: extremely janky, please comment on the appropriate way of setting
+	// * `last_time_needed` set as roughly around startup time
+	// * `prepare_stats` set as Nones, since the metadata was lost
+	async fn prune_and_insert(cache_path: impl AsRef<Path>, artifacts: &mut Artifacts) {
```
mrcnski, line 234: Can this take `&mut self`?

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,79 @@ pub enum ArtifactState {
...
+	/// Create a table with valid artifacts and prune the invalid ones.
+	pub async fn new_and_prune(cache_path: &Path) -> Self {
+		let mut artifacts = Self { inner: HashMap::new() };
+		Self::prune_and_insert(cache_path, &mut artifacts).await;
+		artifacts
+	}
+
+	// FIXME eagr: extremely janky, please comment on the appropriate way of setting
+	// * `last_time_needed` set as roughly around startup time
+	// * `prepare_stats` set as Nones, since the metadata was lost
```
mrcnski, line 233: `last_time_needed` looks good to me.

`prepare_stats` may not even be necessary after the artifact has been prepared and the metrics logged. What happens when this field is removed?

mrcnski:
> > Why do you think so?
> 
> By making it `pub(crate)`, it could be used in `Artifacts`. Right now I'm using an ad hoc version of it there.

Oh, I see what you mean. But I think it can be used in `Artifacts` even if it's private.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,79 @@ pub enum ArtifactState {
...
-	/// Initialize a blank cache at the given path. This will clear everything present at the
-	/// given path, to be populated over time.
-	///
-	/// The recognized artifacts will be filled in the table and unrecognized will be removed.
-	pub async fn new(cache_path: &Path) -> Self {
-		// First delete the entire cache. This includes artifacts and any leftover worker dirs (see
-		// [`WorkerDir`]). Nodes are long-running so this should populate shortly.
-		let _ = tokio::fs::remove_dir_all(cache_path).await;
+	#[cfg(test)]
+	pub(crate) fn new() -> Self {
```
mrcnski, line 215: nit: Maybe the old name `empty` was more descriptive.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,79 @@ pub enum ArtifactState {
...
+				match dir.next_entry().await {
+					Ok(None) => break,
+					Ok(Some(entry)) => {
+						let file_name = entry.file_name();
+						if let Some(file_name) = file_name.to_str() {
+							if is_stale(file_name) {
+								prunes.push(tokio::fs::remove_file(cache_path.join(file_name)));
+							} else if let Some(id) = ArtifactId::from_file_name(file_name) {
+								insert_cache(artifacts, id);
+							}
```
mrcnski, line 265: We should have some logs here for both cases:

1. "stale artifact found for old version, discarding"
2. "valid artifact found on-disk for current version, re-using"

For case (1) it would be really nice to log why the artifact was invalid, e.g. log the stale artifact's version.

polkadot/node/core/pvf/src/artifacts.rs
eagr: I was wrong. Tried a couple of way, couldn't get rid of the unsafe.. AFAIK, `&str` has the same layout as `&[u8]`, so this should be safe in Rust.

> There's not really a need to complicate things with a macro.

I tried a couple of ways. This is by far the simplest from usage perspective, as it is indeed a constant. But if you are really uncomfortable about this, I could replace it with a function. Your call. @mrcnski

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
eagr: > emit some compile-time error instead, that would be ideal

Turns out panic is const. Since const evalutation happens during compilation, so this is practically a compile-time error (as long as we don't export the macro).

```rs
match std::str::from_utf8(&CAT) {
    Ok(s) => s,
    Err(err) => panic!("error converting bytes to str"),
}
```

If unsafe if the main concern here, does this make the macro ok? :)

But again, the unchecked version should be safe in the first place, as `&str` share the same memory layout with those bytes slice.

eagr:
> Oh, I see what you mean. But I think it can be used in `Artifacts` even if it's private.

Yeah, but it's being used in the host mod for testing.

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
eagr: * concationation of two `str`s is guaranteed to be valid utf8
* `s` and `s.as_bytes()` share exactly the same memory layout

So the concatentation of the byte slices is also always valid. If not, that's our fault. ;) It's this kind of guarantees make Rust so lovable.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,85 @@ pub enum ArtifactState {
...
+		let cache_path = cache_path.as_ref();
 		let _ = tokio::fs::create_dir_all(cache_path).await;
 
-		Self { artifacts: HashMap::new() }
-	}
+		if let Ok(mut dir) = tokio::fs::read_dir(cache_path).await {
+			let mut prunes = vec![];
+
+			loop {
+				match dir.next_entry().await {
```
mrcnski, line 246: nit: I think we can just iterate over `dir`, `ReadDir` is an iterator.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,85 @@ pub enum ArtifactState {
...
+			file_name.starts_with(ARTIFACT_PREFIX)
+		}
+
 		// Make sure that the cache path directory and all its parents are created.
+		let cache_path = cache_path.as_ref();
 		let _ = tokio::fs::create_dir_all(cache_path).await;
 
-		Self { artifacts: HashMap::new() }
-	}
+		if let Ok(mut dir) = tokio::fs::read_dir(cache_path).await {
```
mrcnski, line 242: The `Err` case should be handled (I guess all we can do is log a warning).

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,85 @@ pub enum ArtifactState {
...
+			let mut prunes = vec![];
+
+			loop {
+				match dir.next_entry().await {
+					Ok(None) => break,
+					Ok(Some(entry)) => {
+						let file_name = entry.file_name();
+						if let Some(file_name) = file_name.to_str() {
```
mrcnski, line 250: We should probably log a warning on `None`.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,85 @@ pub enum ArtifactState {
...
+							if is_fresh(file_name) {
```
mrcnski, line 251: nit: Maybe it's just me but `is_stale` made more sense to me.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +207,85 @@ pub enum ArtifactState {
...
+								if let Some(id) = ArtifactId::from_file_name(file_name) {
+									self.insert_prepared(id, SystemTime::now(), Default::default());
+									gum::debug!(
+										target: LOG_TARGET,
+										"reusing valid artifact found on disk for current node version v{}",
+										NODE_VERSION,
+									);
```
mrcnski, line 258: I'd include the filename in this log. Otherwise operators will get a bunch of messages on start-up that are all exactly the same. 

All the logs here could also include the cache path.

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
mrcnski: > concationation of two strs is guaranteed to be valid utf8

I didn't know that! Where is it documented? We should document it in the macro.

mrcnski (changes requested):
Almost there! 👍

polkadot/node/core/pvf/src/artifacts.rs
eagr: too bad, tokio::fs::ReadDir is not

mrcnski: Oh. I think it can be turned into a stream, but it's also fine without that boilerplate.

polkadot/node/core/pvf/src/artifacts.rs
eagr: I'll opt for error!() as this could be a bug.

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
eagr: https://github.com/rossmacarthur/constcat/blob/trunk/src/lib.rs

Just found out about this crate that's basically doing the same thing as this macro. They use `from_utf8_unchecked` for the same safety reasoning.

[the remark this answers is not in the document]
mrcnski: Well, just because someone made a crate for it, doesn't make it sound. 😛

[The repo](https://github.com/rossmacarthur/constcat) only has 5 stars, which doesn't inspire confidence. The SAFETY comment there is also missing a step in the reasoning:

        // SAFETY: The original constants were asserted to be &str's
        // so the resultant bytes are valid UTF-8.

I know I sound like a broken record, but where is it documented that directly concatenating `str` bytes results in valid UTF-8? I'm not saying it's unsound, just that we need stronger proofs when introducing `unsafe`.

mrcnski:
I was thinking a bit more about whether this addresses all of the concerns in https://github.com/paritytech/polkadot-sdk/issues/685. Namely:

> 1. Unsafe functions are used to import the artifact
> 2. [Their safety guarantees](https://github.com/paritytech/polkadot/blob/2bbef460d24f44aeba9d4014062bb8c631327496/node/core/pvf/src/executor_intf.rs#L199) rely on the fact the node has produces them with prepare() and they are not just some random files put there by someone.

So, we really need some more integrity checks, to avoid potential disputes (see below).

## More Precise Versioning

One issue right now is that the node version which we use to tag artifacts is not very precise. Namely, it is only bumped on releases, which may not reflect changes in semantics. This can result in the following situation:

1. An operator is running a validator built off the latest release.
2. He switches to running a node built off of the latest commit in `master`. (Maybe because he needed some bugfix that hadn't been released yet.)
3. The node version hasn't changed since the last release, but the wasmtime version might have!
4. If some different semantics is encountered in execution, a dispute can be raised.

I guess you could argue that production validators should only be running official releases, but I don't think we enforce this, and it'd be nice to avoid unnecessary disputes where possible.

So we will need to tag the artifact with either the wasmtime version, or the latest git commit. The former is preferable but the latter might be easier (see `SUBSTRATE_CLI_COMMIT_HASH`).

## File Integrity

I realized that we are still missing the file integrity hashes, which were a pre-requisite for safely re-using pre-built artifacts. See https://github.com/paritytech/polkadot-sdk/issues/677.

## Conclusion

Both these issues need to be fixed. You can either do it in this PR or a separate PR which targets this one. However, to avoid accidentally merging in this PR in its current form I will convert it to a draft.

Please feel free to ask any questions either here or by email. I apologize for the inconvenience, but where consensus is involved it's really important that we get it right!

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
mrcnski: Good to know, thanks 💪

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -96,6 +97,9 @@ macro_rules! concat_const {
             cat
         };
 
+        // FIXME eagr: consider opting for `from_utf8_unchecked()`
+        // SAFETY: The concatenation of two string slices is guaranteed to be valid UTF8,
+        // so are the byte slices as they have the same memory layout.
```
mrcnski, line 102: I would stay with the safe version, so no `unsafe`, since we have a policy around that. (I know we don't always follow it, but we're trying to be better about it going forwards.) It's good to keep the comment but we don't need `SAFETY` anymore:

Also, can we add a doc above `macro_rules! concat_const`?

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -231,55 +235,77 @@ impl Artifacts {
...
 										NODE_VERSION,
 									);
-									continue
 								}
+							} else {
+								gum::warn!(
+									target: LOG_TARGET,
+									"non-Unicode file name found in {:?}",
+									cache_path,
+								);
```
mrcnski, line 283: We can try to log the file path/name also. That would help point users to the offending file.

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: Nice!

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: Thanks, looks good!

eagr:
@mrcnski No problem, man. Correntness is paramount here. Again, really appreciate the help! 🙏

I'll work on the integrity check thing on this branch as they seem to be related. Here is the plan.

* hash the artifact after preparation and attach the hash to file name
    * file name: PREFIX _ CODE HASH _ PARAMS HASH _ CHECKSUM
* check integrity at startup and prune the corrupted ones
* check integrity if candidate has been deemed invalid, but I'm not about what actions to take if an artifact is found corrupted.

I know you're not entirely happy with the ValidationError. How about we do the refactoring before this? @mrcnski

polkadot/node/core/pvf/src/artifacts.rs
s0me0ne-unkn0wn: How about this one? (Trades some weird syntax for the absence of runtime code)

```rust
macro_rules! c_c_combo {
    ( $combined:ident : $($name:ident : $type:ty = $value:expr);+ $(;)? ) => {
        $(
            const $name: $type = $value;
        )+
        const $combined: &'static str = concat!($($value),+);
    }
}

c_c_combo! (PATH:
    BASE: &'static str = "path/to";
    PART: &'static str = "foo";
);
```

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +211,107 @@ pub enum ArtifactState {
...
+						Ok(Some(entry)) => {
+							let file_name = entry.file_name();
+							if let Some(file_name) = file_name.to_str() {
+								let id = ArtifactId::from_file_name(file_name);
+								let file_path = cache_path.join(file_name);
+
+								if is_stale(file_name) || id.is_none() {
+									prunes.push(tokio::fs::remove_file(file_path.clone()));
+									gum::debug!(
+										target: LOG_TARGET,
```
s0me0ne-unkn0wn, line 262: Nit: I have to quibble about 10 levels of indentation, although I'm doing that myself sometimes :slightly_smiling_face: 
The [style guide](https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md) says, "no more than 5 and definitely no more than 8". I'm in favor of relaxing that rule and increasing the allowed line length, but for now, it's there. But you don't have to follow it strictly.

s0me0ne-unkn0wn (changes requested):
Good job! Just fix the docstrings, they need to be kept up to date, the rest are nit suggestions.

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: I agree. Maybe everything in the `Ok(Some(entry)) => {` block can be extracted into a helper. Would make it easier to read. :)

mrcnski:
> I'll work on the integrity check thing on this branch as they seem to be related. Here is the plan.

@eagr That plan looks really good! Great job!

> check integrity if candidate has been deemed invalid, but I'm not about what actions to take if an artifact is found corrupted.

Sounds good, but as you say it might require some more consideration. I'd leave it for a follow-up PR, as we don't need it here for re-using the artifact cache - what do you think?

> I know you're not entirely happy with the ValidationError. How about we do the refactoring before this?

We can do that. So we can get this current PR in (with points 1 and 2 of your plan), then refactor-validation-error, then check-integrity-on-candidate-invalid. :)

polkadot/node/core/pvf/src/artifacts.rs
eagr: This has to be refactored for sure :)

eagr: Gonna punt on the help function. Having a function would complicate the typing a lot (`prunes` is a vec of futures). Compared to that, the indentation wouldn't be so bad.

polkadot/node/core/pvf/src/artifacts.rs
eagr: Nice~ But probably not, the NODE_VERSION is not defined in this mod.

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: For the type you can just do `impl Future` without having to spell out the exact type. :)

I went ahead and refactored it in https://github.com/paritytech/polkadot-sdk/pull/1918/commits/4b1bb0ab6586d611da7201fb0ebb71c129feccb1.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4036058

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: Okay, I get a typing error when I try to compile tests. Now I see what you mean.

TBH, I would keep the helper, but change it so that we call `.await` on `remove_file` immediately, instead of saving the future and calling `join_all` later. I'm sure it is not optimal, but this code should run very rarely and no one would notice it taking 0.1s longer, anyway. :P

eagr: Ok, I'll keep on working with your code. But I'll probably keep the `new` function small, only handling construction as it was before :)

eagr: Deep indentation is often bad because it can be a trait of stuff made unnecessarily harder to reason about. If that's the case, it should be uncluttered for sure, e.g., by returning early, which is often an improvement. The guideline is great for that purpose.

But sometimes, it's just what it takes, like in this case. The logic is simple: register if valid, otherwise prune. The rest is kind of collateral. And even with that, the code isn’t that hard to reason about. We’re not really simplifying stuff this way, just relocating. Idk, is it worth it to make that compromise, for aesthetic reasons?

eagr:
Did a very rough PoC for the checksum. Would love to hear your inputs before proceeding. @mrcnski @s0me0ne-unkn0wn

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4038898

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: Good point. Personally, comparing before and after, I do find it easier to read now. And to me it makes sense to separate out handling of individual entries into a helper function. `insert_and_prune` works on a whole directory, while `insert_or_prune` works on individual entries. I can see it being subjective, but since we have a style guideline, I'd say let's just follow it.

mrcnski: I like how you organize helper functions as subfunctions, we should do that more often!

mrcnski:
@eagr Sorry for the delay. Overall it looks very good! I'm just not sure the checksum should be part of the artifact ID, generally an ID should be immutable. I left some comments on the commit:

https://github.com/paritytech/polkadot-sdk/commit/50b7ccc4b20947188585ce67a716a7de6e79c42a#comments

mrcnski (changes requested):
See https://github.com/paritytech/polkadot-sdk/commit/50b7ccc4b20947188585ce67a716a7de6e79c42a#r130959443 (for some reason my comments on the PoC are not showing up on this PR.)

eagr:
Let’s step back and reconsider this.

Assuming we do need the integrity check, then we should probably include the checksum in the file name. With that assumption, the `ArtifactId::path()` method no longer makes sense.

Rather than storing the checksum somewhere else, only later to pice it together with something else to get the path, we might as well just store the suffix of the path instead. We could have something like `Artifacts::path(id)`.

Then instead of relying on `ArtifactId::path(cache_path)`, maybe we could just pass the artifact path?

mrcnski:
Sounds reasonable to me @eagr. I have no strong opinions here, as long as the code is correct and reasonably clean.

Also, now that we're preserving the artifacts cache, we should make sure that previous test/benchmark runs don't affect subsequent runs, i.e. we always start with `Artifacts::empty`. I believe it's the case already, but just a heads-up.

eagr:
Just for clarification, we are sure that we need the integrity checks right?

mrcnski:
@eagr You mean on startup, or after a candidate was found invalid, or both? Good question. On startup, maybe not essential if we always do it before voting against a candidate. What is the reason not to do it?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4136358

eagr:
Summary of the changes:

* Hash temp artifact (before renaming) to get the checksum, then append checksum to the previous file name, and use the result to rename the artifact.
* `Outcome::Concluded` becomes `{ worker: IdleWorker, result: PrepareResult, path: Option<PathBuf> }`.
* Bounce back the full path of artifact if successfully compiled and store it in the artifact table for later retrieval.
* Compute hash of existing artifact caches at startup. Prune if hashes doen't match.

The tests are broken due to the changes. I'll fix them later if you're ok with the changes.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4136699

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +209,147 @@ pub enum ArtifactState {
...
+			let file_type = entry.file_type().await;
+			let file_name = entry.file_name();
+
+			match file_type {
+				Ok(file_type) =>
+					if !file_type.is_file() {
+						return
+					},
+				Err(err) => {
+					gum::warn!(
```
mrcnski, line 274: nit: There are three warn logs in this fn ("discarding invalid artifact" should be a warn). We could reduce some LoC by returning an error and logging in a single place, at the call site.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -88,37 +128,28 @@ impl ArtifactId {
 		Self::new(pvf.code_hash(), pvf.executor_params().hash())
 	}
 
-	/// Tries to recover the artifact id from the given file name.
-	#[cfg(test)]
-	pub fn from_file_name(file_name: &str) -> Option<Self> {
-		use polkadot_core_primitives::Hash;
-		use std::str::FromStr as _;
-
+	/// Returns the expected path to this artifact given the root of the cache.
```
mrcnski, line 131: Should update the docstring.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -88,37 +128,28 @@ impl ArtifactId {
...
+	pub fn path_prefix(&self, cache_path: &Path) -> PathBuf {
```
mrcnski, line 132: `path_prefix` is not really an accurate name, it's more of a filename prefix. Maybe `path_with_prefix`. But I wonder whether we still need this fn at all.

polkadot/node/core/pvf/src/prepare/worker_intf.rs
```diff
@@ -70,7 +70,7 @@ pub async fn spawn(
 /// If the idle worker token is not returned, it means the worker must be terminated.
 pub enum Outcome {
 	/// The worker has finished the work assigned to it.
-	Concluded { worker: IdleWorker, result: PrepareResult },
+	Concluded { worker: IdleWorker, result: PrepareResult, path: Option<PathBuf> },
```
mrcnski, line 73: It's better to put the path in `PrepareResult`. The path should only exist in the `Ok` case, then it also doesn't have to be an `Option`.

polkadot/node/core/pvf/src/prepare/worker_intf.rs
```diff
@@ -196,14 +196,14 @@ async fn handle_response(
 	result: PrepareResult,
 	worker_pid: u32,
 	tmp_file: PathBuf,
-	artifact_path: PathBuf,
+	artifact_path_partial: PathBuf,
```
mrcnski, line 199: I think your approach of returning the final artifact path in the success case makes sense. Just don't think we need to pass in a partial artifact here. If we instead passed the cache path, then we should have everything necessary to construct the artifact path at the point it's needed. Passing around a partial value seems wrong. For example, elsewhere in the file we are logging this value without indicating that it's a partial value, which can be confusing to readers of the log.

polkadot/node/core/pvf/src/prepare/worker_intf.rs
```diff
@@ -219,6 +219,20 @@ async fn handle_response(
...
+		Err(_) => return Outcome::Unreachable,
+	};
+
+	let file_name_partial = artifact_path_partial
+		.file_name()
+		.expect("the path should never terminate in `..`");
+	let mut file_name = file_name_partial.to_os_string();
+	file_name.push("_0x");
+	file_name.push(checksum.to_hex().as_str());
+	let artifact_path = artifact_path_partial.with_file_name(&file_name);
```
mrcnski, line 234: We should have a helper function for this and some unit tests. In the tests we can call `from_file_name` on the result of the helper function to verify that we get expected values.

mrcnski (changes requested):
Looks good! I mainly have some suggestions to simplify things, overall the approach is good.

> Alternative: I could instead hash the compiled bytes (before writing to the tempfile) to avoid the additional IO, but then the checksum is coupled with how the artifact is constructed and might be prone to change.

I don’t understand the second part of this sentence. But yeah, it’s possible to hash on the worker side and return the checksum in `PrepareResult`, thus saving some IO. It’s not a huge deal since performance in preparation is not critical, but we might as well be more efficient. And I don’t think it would hurt to have the checksum in `PrepareResult`.

> Alternative: put the path in PrepareStats.

I’d for sure put the path in `PrepareResult`, however I’m not sure it quite fits in `PrepareStats`.

> Bounce back the full path of artifact if successfully compiled and store it in the artifact table for later retrieval.

Makes sense! After all, the path is only meaningful if we successfully compiled. I’d also remove this notion of a partial path which is meaningless by itself.

mrcnski:
I realize that our code is not the most well-structured, and has been in need of a refactor for quite some time. So despite all my nitpicks, I am impressed that you've been able to dive in and make really substantial changes!

polkadot/node/core/pvf/src/prepare/worker_intf.rs
eagr: This is the only place left that is not using the full path, coz here we don't have it yet. :)

eagr:
The code is fine. I actually don't have too much trouble navigating around. This is what you get when using actor model I guess, which contributes the majority of the complexity.

> has been in need of a refactor for quite some time

I'd love to help out if there's plan to do it. 😄

mrcnski:
Cool, maybe it's just me being overly perfectionist. 😅

> I'd love to help out if there's plan to do it. 😄

I just have a couple ideas here: https://github.com/paritytech/polkadot-sdk/issues/1576 Would love some help with it, plus any other ideas you may have!

polkadot/node/core/pvf/src/prepare/worker_intf.rs
eagr: Oh, I remember now. Coz `PrepareResult` is propagated from somewhere the path (renamed to) is not available. Now that I think of it, maybe I could just use a placeholder.

eagr:
The only place I could find that the tmp artifact file is written to is in `polkadot-node-core-pvf-prepare-worker`. I just realized that it's gated by the `test-utils` feature, which means it's only used for testing right? But in production code, the tmp file is renamed to the final dst. This part puzzles me 🤔. I'm probably missing something here. Care to explain?

s0me0ne-unkn0wn:
@eagr it's gated in the PVF host package because some code is needed by PVF host tests (that'll be fixed by #1192), but the worker binary itself belongs to the root `polkadot` package, and there it's not gated, it's a direct dependency.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4159473

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -176,32 +209,147 @@ pub enum ArtifactState {
...
+		artifacts
+	}
+
+	async fn insert_and_prune(&mut self, cache_path: &Path) {
+		async fn is_corrupted(path: &Path) -> bool {
+			let file_name = path
+				.file_stem()
+				.expect("path must contain a file name")
+				.to_str()
+				.expect("file name must be valid UTF-8");
```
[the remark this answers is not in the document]
eagr: Yeah, but at the point the function is called, the possibilities are ruled out.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4172295

polkadot/node/core/pvf/src/artifacts.rs
eagr: gonna punt on this, coz only one of them is `Err`

polkadot/node/core/pvf/src/artifacts.rs
eagr: Hmm, not really a filename prefix. This is the prefix of the expected path, without the checksum, which used to be the canonical path of an artifact.

eagr: Idk, `path_with_prefix` conveys that the return is a path, but it is more of a prefix of the canonical path.

eagr:
Tests are green now. Ready for another round of review. :))

polkadot/node/core/pvf/src/artifacts.rs
eagr: Now that I think of it, maybe we should just do a `path(cache_path, checksum)` instead, which is probably more useful.

mrcnski:
Hey @eagr, definitely agree that the errors deserve a good refactor. But, as you can tell I'm struggling to keep up with reviews. I wish you had asked for my feedback first, because this will delay my review. :( We could have probably done this in your dedicated error refactoring PR, but it's too late now I guess. :P Will try to get to this soon.

mrcnski:
It does look good based on a quick skim-through. Thanks for your work. 👍

eagr:
Oh, sorry about that! I'll wait for the feedback next time.

polkadot/node/core/pvf/src/prepare/worker_intf.rs
```diff
@@ -108,25 +109,17 @@ pub async fn start_work(
...
 	let IdleWorker { stream, pid, worker_dir } = worker;
 
-	gum::debug!(
-		target: LOG_TARGET,
-		worker_pid = %pid,
-		?worker_dir,
-		"starting prepare for {}",
-		artifact_path.display(),
-	);
-
```
mrcnski, line 122: Would be nice to keep this. Maybe we can print the `pvf` if that's not too noisy, or print something else to identify the artifact. This log has been useful to me in the past.

polkadot/node/core/pvf/src/prepare/worker_intf.rs
```diff
@@ -157,7 +150,8 @@ pub async fn start_work(
 						prepare_result,
```
mrcnski, line 150: nit: for a bit more clarity:

```suggestion
						prepare_worker_result,
```

polkadot/node/core/pvf/src/host.rs
```diff
@@ -1007,12 +996,26 @@ pub(crate) mod tests {
...
+		fn to_bytes(mut n: u32) -> Vec<u8> {
+			let mut bytes = vec![0; 4];
+			let mut i = 0;
+			while n > 0 {
+				bytes[i] = (n & 0xff) as u8;
+				n >>= 8;
+				i += 1;
+			}
+			bytes
+		}
```
mrcnski, line 1013: nit: Can we instead `.encode()` the discriminator, or something along those lines? I don't feel like reviewing this :P :D

polkadot/node/core/pvf/src/host.rs
```diff
@@ -1007,12 +996,26 @@ pub(crate) mod tests {
 	}
 
 	/// Creates a new PVF which artifact id can be uniquely identified by the given number.
-	fn artifact_id(descriminator: u32) -> ArtifactId {
-		ArtifactId::from_pvf_prep_data(&PvfPrepData::from_discriminator(descriminator))
+	fn artifact_id(discriminator: u32) -> ArtifactId {
```
mrcnski, line 999: Good catch. 👍

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -115,8 +115,8 @@ fn send_response(stream: &mut UnixStream, result: PrepareResult) -> io::Result<(
 ///
 /// 6. If compilation succeeded, write the compiled artifact into a temporary file.
 ///
-/// 7. Send the result of preparation back to the host. If any error occurred in the above steps, we
-///    send that in the `PrepareResult`.
+/// 7. Send the result of preparation back to the host, including the checksum of the artifact. If
+///    any error occurred in the above steps, we send that in the `PrepareWorkerResult`.
```
mrcnski, line 119: Thanks for remembering to update the docstring!

polkadot/node/core/pvf/common/src/error.rs
```diff
@@ -14,13 +14,21 @@
...
+/// preparation if successful.
+pub type PrepareWorkerResult = Result<PrepareWorkerSuccess, PrepareError>;
+
+/// Result of PVF preparation propagated all the way back to the host, with path to the concluded
+/// artifact and stats of the preparation if successful.
+pub type PrepareResult = Result<PrepareSuccess, PrepareError>;
+
+/// Result of prechecking PVF performed by the validation host. Contains stats about the preparation
+/// if successful.
+pub type PrecheckResult = Result<PrepareStats, PrepareError>;
```
mrcnski, line 31: I don't think the stats are used by the caller, can it just be `Result<(), PrepareError>`? (I haven't tried it, but *think* it should work.)

mrcnski (commented):
Niiiice, the refactor is so clean! Sorry that this took so long to review, I didn't have the bandwidth last week. But it looks *really* good. ⭐️

I think the only thing left is [this comment](https://github.com/paritytech/polkadot-sdk/pull/1918#issuecomment-1770294544), specifically "More Precise Versioning"?

mrcnski:
I'll fix the merge conflicts now.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4300493

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4300495

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4300481

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4300494

polkadot/node/core/pvf/common/src/error.rs
eagr: I guess this is used for some response of a cli command right? I was just keeping it as it was before. The code should compile but Is it ok to change the response? Does it affect the user? I'll trust your judgement here. :)

mrcnski: Good questions! And yeah, it was already the case that we returned the stats. Previously the stats were part of the overloaded `PrepareResult`, but now that we're using different results in different contexts, we can reconsider whether the stats are needed here.

So no, removing them here shouldn't affect the user at all. Prechecking is requested by the `pvf-checker` subsystem and the message is handled [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/core/candidate-validation/src/lib.rs#L230). We throw away the specific success value and simply [report the outcome](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/core/candidate-validation/src/lib.rs#L378-L386), whether prechecking succeeded or not. So since the stats are used only within the PVF host itself for metrics purposes, they don't need to be returned higher up. Doing so kind of implies that they are used by `pvf-checker`, when they are not.

polkadot/node/core/pvf/src/host.rs
eagr: Yeah, this is entirely unnecessary. And it doesn't reflect how things really work.

eagr: I'll run the hash function on the pvf code, which is also not necessary for the test to pass, but at least it reflects how things work. People sometimes would try to learn how the code works by reading the tests. It's good for that regard.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4301732

eagr:
For the wasmtime version, some random ideas

1. env variable
2. run a command in a build script to get it at compile time and store it somewhere
3. run a command to get it at runtime lazily then put in a cell or something

1 adds burden to maintenance, could get out of sync without noticing it. 2 seems too much trouble for just this one small thing. I'll try 3 out. Do you have better ideas?

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -16,10 +16,10 @@
 
 //! PVF artifacts (final compiled code blobs).
 //!
-//!	# Lifecycle of an artifact
+//! # Lifecycle of an artifact
```
mrcnski, line 19: Random thought, could we add an `# Artifact integrity` section here? It could explain that artifacts must remain valid to satisfy the `SAFETY` constraints of `execute_artifact`. Therefore we do two things: check file integrity and version compatibility on host startup, and also before voting against a candidate we re-check the file integrity.

mrcnski:
> For the wasmtime version, some random ideas
> 
>     1. env variable
> 
>     2. run a command in a build script to get it at compile time and store it somewhere
> 
>     3. run a command to get it at runtime lazily then put in a cell or something

I'd go with 2. The versions of dependencies may be available to `build.rs` and then we can store it in a compile-time env var accessible with `env!`. If not directly available, then at least through calling `cargo-metadata` or something. It might take some ~Googling~ [Kagiing](https://kagi.com) to figure out how exactly to do this, but I feel that it's possible. If all else fails, we could just parse `Cargo.lock` from `build.rs` and get the version that way.

mrcnski:
(2) could also potentially be useful for us in the future, for worker version checks. I've looked into this very same thing [before](https://github.com/paritytech/polkadot-sdk/issues/626#issuecomment-1691880018). 

For example, if someone downloads the node and workers at a release version, and then switches to master for some bug fixes, he could be running an updated node without updating the workers, but wasmtime may have been bumped since then. Now that I think of it, it'd be good to incorporate this in the version checks.

eagr:
> then we can store it in a compile-time env var accessible with `env!`

Is this even possible? I was thinking something like writing the version to a file and read it at runtime. Or maybe generate a source file which exports the version as a const. That's why I think it's too much trouble.

eagr:
Oh, I see how this works. It's like writing a oneliner shell `FOO=bar` during compilation so that compiler process has access to the value. This is clever. Thanks for the heads up!

mrcnski:
Yep! Here's the doc in case it helps: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env We also have some examples in the repo already, check out some `build.rs` scripts for reference!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4310994

eagr:
I see there's this `build-script-utils` crate. Do you think it should go there instead?

polkadot/node/core/pvf/common/src/pvf.rs
```diff
@@ -115,7 +115,7 @@ impl fmt::Debug for PvfPrepData {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		write!(
 			f,
-			"Pvf {{ code, code_hash: {:?}, executor_params: {:?}, prep_timeout: {:?} }}",
+			"Pvf {{ code_hash: {:?}, executor_params: {:?}, prep_timeout: {:?} }}",
```
mrcnski, line 118: nit: Hmm, maybe this, to make it more obvious that a field is omitted:

```suggestion
			"Pvf {{ code: [...], code_hash: {:?}, executor_params: {:?}, prep_timeout: {:?} }}",
```

polkadot/node/core/pvf/common/build.rs
```diff
@@ -0,0 +1,28 @@
+fn main() {
+	get_wasmtime_version();
+}
+
+pub fn get_wasmtime_version() {
```
mrcnski, line 5: Maybe add a comment that in case there are multiple wasmtime versions being used, is an error. Example:

```
 $ cargo tree --package=aead --depth=0
error: There are multiple `aead` packages in your project, and the specification `aead` is ambiguous.
Please re-run this command with one of the following specifications:
  aead@0.3.2
  aead@0.4.3
  aead@0.5.2
```

I think an error in that case great, we definitely don't want to be mixing wasmtime versions anyway.

polkadot/node/core/pvf/common/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "polkadot-node-core-pvf-common"
 description = "Polkadot crate that contains functionality related to PVFs that is shared by the PVF host and the PVF workers."
 version = "1.0.0"
+build = "build.rs"
```
mrcnski, line 5: nit: I think this is unnecessary

mrcnski (approved):
Looks great! Thanks for all your hard work! @s0me0ne-unkn0wn want to take another look?

> I see there's this `build-script-utils` crate. Do you think it should go there instead?

Yeah, could be useful for the follow-up I mentioned above, to have this in other crates.

eagr:
Kudos to you! I know how demanding code review is, esp. at this level of fidelity. I really don't know how could you keep up with it. Really appreciate it man.

> Yeah, could be useful for the follow-up I mentioned above, to have this in other crates.

Let's do it in this PR, alright? @mrcnski

polkadot/node/core/pvf/src/artifacts.rs
eagr: How about we do this when we have decided what to do if an artifact is found corrupted? I'll make a note of this.

polkadot/node/core/pvf/common/build.rs
mrcnski: I mean a comment would be useful for readers of the code. I was wondering what would happen if there were multiple versions of wasmtime, I didn't know it would be an error until I tried it.

mrcnski:
There be some scary merge conflicts, fixing...

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4335308

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4335309

substrate/utils/build-script-utils/src/version.rs
mrcnski, line 1: cc @bkchr

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -212,53 +365,49 @@ impl Artifacts {
...
-	#[cfg(test)]
-	pub fn insert_prepared(
+	pub(crate) fn insert_prepared(
 		&mut self,
 		artifact_id: ArtifactId,
+		path: PathBuf,
 		last_time_needed: SystemTime,
 		prepare_stats: PrepareStats,
 	) {
 		// See the precondition.
```
s0me0ne-unkn0wn, line 385: The precondition was dropped. Either this comment should be removed as well, or the precondition should be reintroduced (which makes sense, as it's implied by `always!` and it would be good to explain to a code reader why it always holds).

polkadot/xcm/procedural/tests/ui.rs
```diff
@@ -21,7 +21,7 @@
 fn ui() {
 	// Only run the ui tests when `RUN_UI_TESTS` is set.
 	if std::env::var("RUN_UI_TESTS").is_err() {
-		return;
```
s0me0ne-unkn0wn, line 24: None of our business :)

substrate/client/chain-spec/src/chain_spec.rs
```diff
@@ -784,9 +784,7 @@ fn json_eval_value_at_key(
 	path: &mut VecDeque<&str>,
 	fun: &dyn Fn(&json::Value) -> bool,
 ) -> bool {
-	let Some(key) = path.pop_front() else {
-		return false;
-	};
+	let Some(key) = path.pop_front() else { return false };
```
s0me0ne-unkn0wn, line 787: None of our business as well. It may be wrong indeed, or it may be formatter on your side, doesn't matter, we shouldn't introduce totally unrelated changes to a PR, even that small.

s0me0ne-unkn0wn (approved):
Looks very good! :+1: Left some minor comments.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -55,18 +55,64 @@
...
+            Ok(s) => s,
+            Err(_) => panic!("Error converting bytes to str"),
+        }
+    }}
+}
+
+const RUNTIME_PREFIX: &str = "wasmtime_v";
+const NODE_PREFIX: &str = "polkadot_v";
+const ARTIFACT_PREFIX: &str =
+	concat_const!(RUNTIME_PREFIX, RUNTIME_VERSION, "_", NODE_PREFIX, NODE_VERSION);
```
bkchr, line 114:
```suggestion
```

This is over the top. We don't need this macro, nor do we need to concat this at compile time. We are using format anyway to generate the name.

substrate/utils/build-script-utils/src/version.rs
```diff
@@ -59,3 +59,34 @@ fn get_version(impl_commit: &str) -> String {
...
+
+/// Generate `SUBSTRATE_WASMTIME_VERSION`
+pub fn generate_wasmtime_version() {
+	generate_dependency_version("wasmtime", "SUBSTRATE_WASMTIME_VERSION");
+}
+
+fn generate_dependency_version(dep: &str, env_var: &str) {
+	// we only care about the root
+	match std::process::Command::new("cargo")
+		.args(["tree", "--depth=0", "--package", dep])
```
bkchr, line 71:
```suggestion
		.args(["tree", "--depth=0", "--locked", "--offline", "--package", dep])
```

polkadot/node/core/pvf/src/artifacts.rs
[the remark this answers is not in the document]
bkchr: Can you please still handle them gracefully. I mean this is here just by returning true.

polkadot/node/core/pvf/src/artifacts.rs
eagr: You mean we do something like this instead?

```rs
fn artifact_prefix() -> String {
	format!("{}{}_{}{}", RUNTIME_PREFIX, RUNTIME_VERSION, NODE_PREFIX, NODE_VERSION)
}
```

substrate/utils/build-script-utils/src/version.rs
eagr: Turns out that --locked alone would achieve both.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -258,7 +221,13 @@ impl Artifacts {
...
 				},
 			};
-			!file_name.ends_with(checksum.to_hex().as_str())
+
+			if let Some(file_name) = path.file_name {
+				if let Some(file_name) = file_name.to_str() {
+					return !file_name.ends_with(checksum.to_hex().as_str())
+				}
+			}
+			false
```
mrcnski, line 230: Should be `true` right? Could use a test for this case if we don't have one yet.

mrcnski (changes requested):

polkadot/node/core/pvf/src/artifacts.rs
eagr: Oops! That's right. But this line of can only be reached when

* the path end in "/..", or
* the path contain non-Unicode

These are all screened out at the call site, so I don't even know how to write a test that could fail because of this line lol.

mrcnski: I guess `is_corrupt` can be extracted and unit-tested, but whatever, looks good already!

mrcnski (approved):

polkadot/node/core/pvf/src/artifacts.rs
bkchr: Yes

eagr: This is the very first version. Now it still feels odd to compute the same thing over and over.. I understand this is not remotely close to hot and this isn't about performance. When I looked at how much cleaner the code became with the const/static compared to a function call, I felt the macro thing was worth it. A const is just so much easier to use than a function. When a value is static, it probably should be used as one.

eagr: This is the very first version. Now it still feels odd to compute the same thing over and over.. I understand this is not remotely close to hot and this isn't about performance. I understand that the macro thing may feel like too much at the first glance. When I looked at how much cleaner the code became with the const/static compared to a function call, I felt it's worth it. A const is just so much easier to use than a function, as when a value is static, it probably should be used as one. Besides, isn't it nice that drop/dealloc/format calls are wiped off from the assembly? :))

polkadot/node/core/pvf/src/artifacts.rs
eagr: We should probably stick to testing the interfaces, this kind of tests often become maintenance liability (prone to change). 😉

polkadot/node/core/pvf/src/artifacts.rs
mrcnski: If it's a function you can at least put it behind a `OnceCell` or something. Otherwise I did a search and there are some crates that do this: `const_format`, `concat_const`, etc. But it would be sad to pull in a whole dependency for this. I wish std had this... Meh, let's just merge as-is (with the function). I'll press the green button once conflicts are resolved.

eagr: > If it's a function you can at least put it behind a OnceCell or something.

Function, cell, lazy static, considered all that before doing the macro as it results in the simplest interface, a static str. I won't change it back. Let's get this merged. 🚀 I'm just trying to understand why is a function call better here.

mrcnski: > I'm just trying to understand why is a function call better here.

The problem with the macro is that it's a lot of code for readers to understand, just to declare a constant. TBH, all the options are kind of meh, but given that the code is cold, optimizing for maintainability with a plain function seems like the right call.

mrcnski:
Aaaand merged! 🎉 Thanks @eagr, awesome PR! I extracted out a couple follow-ups so they don't get lost.

PR #1920, opened 2023-10-17
Do not force collators to update after enabling async backing

bkchr:
The validators are checking if async backing is enabled by checking the version of the runtime api. If the runtime api is upgraded by a runtime upgrade, the validators start to also enable the async backing logic. However, just because async backing is enabled, it doesn't mean that all collators and parachain runtimes have upgraded. This pull request fixes an issue about advertising collations to the relay chain when it has async backing enabled, but the collator is still using the old networking protocol. The implementation is actually backwards compatible as we can not expect that everyone directly upgrades. However, the collation advertisement logic was requiring V2 networking messages after async backing was enabled, which was wrong. This is now fixed by this pull request.

Closes: https://github.com/paritytech/polkadot-sdk/issues/1923


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3991119

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1072,17 +1081,19 @@ where
...
 		)
 		.map_err(AdvertisementError::Invalid)?;
+
 	if !per_relay_parent.collations.is_seconded_limit_reached(relay_parent_mode) {
 		return Err(AdvertisementError::SecondedLimitReached)
 	}
 
 	if let Some((candidate_hash, parent_head_data_hash)) = prospective_candidate {
-		let is_seconding_allowed = !relay_parent_mode.is_enabled() ||
+		let is_seconding_allowed = !use_non_prospective_chains_mode ||
```
rphmeier, line 1096: This bit doesn't quite make sense to me.

Previously, the candidate backing system was checking the parent head data hash.

We should still do the `can_second` check even if the collator version is old (not rely on them to be honest), but just make sure that the `parent_head_data_hash` is the on-chain included block

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -224,6 +224,7 @@ impl PeerData {
 		&mut self,
 		on_relay_parent: Hash,
 		relay_parent_mode: ProspectiveParachainsMode,
+		use_non_prospective_chain_mode: bool,
```
rphmeier, line 227: This parameter is unnecessary and can be derived from `relay_parent_mode` and the peer's CollationVersion.

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
eskimor: Actually I don't get this at all: We only get here if prospective_candidate is `Some`, which implies that `use_non_prospective_chains_mode` is [false](https://github.com/paritytech/polkadot-sdk/blob/5291601ff4948c9181e4f8edddb00c18fab7ba19/polkadot/node/network/collator-protocol/src/validator_side/mod.rs#L1059). !use_non_prospective_chains_mode will always be `true` here.

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
bkchr: Ahhhh! Good point. I should not program while doing a call :P

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3992856

eskimor (approved):
Looks good to me. I think this should actually work. For @rphmeier comment, I don't think we need to check the parent here for synchronous backing as it is checked in candidate validation anyway. This requires fetching the collation, but adversaries also have other means for letting us fetch a collation that turns out invalid, so checking this beforehand does not help that much.

(Nothing gets worse than it was before.)

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3995998

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-digest-18-oct-2023/4318/1

ordian (approved):

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1077,20 +1084,20 @@ where
 			&state.active_leaves,
 		)
 		.map_err(AdvertisementError::Invalid)?;
+
 	if !per_relay_parent.collations.is_seconded_limit_reached(relay_parent_mode) {
 		return Err(AdvertisementError::SecondedLimitReached)
 	}
 
 	if let Some((candidate_hash, parent_head_data_hash)) = prospective_candidate {
-		let is_seconding_allowed = !relay_parent_mode.is_enabled() ||
```
slumber, line 1085: `!relay_parent_mode.is_enabled() ||` is there in case two nodes communicate with V2 but async backing is still off, meaning we don't need to check for `can_second` -- there's only one slot for candidate per relay parent anyway.

Also, seems like this branch won't be executed at all even with async backing **enabled** if collator chooses to communicate with v1, which opens some room for spamming the node with validation work.

bkchr: https://github.com/paritytech/polkadot-sdk/pull/1920#discussion_r1362441842

> Also, seems like this branch won't be executed at all even with async backing **enabled** if collator chooses to communicate with v1, which opens some room for spamming the node with validation work.

You mean because we would add the collation without further checks?

slumber: > https://github.com/paritytech/polkadot-sdk/pull/1920#discussion_r1362441842

In the previous version of this PR `use_non_prospective_chains_mode` was defined based on what collator told us, in the original code it relies on the runtime request from validators.

> You mean because we would add the collation without further checks?

The idea is to only request collation once the parent parachain block becomes backable. Until then, we don't do any unnecessary work. With this code collator can
1. Choose to use v1 networking
2. Send maximum number of sequentially based collations and validators will start requesting & validating without waiting for the "backable parent" part.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4010046

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4010048

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
rphmeier: yes, in the v1 networking version it should only accept `parent_head_data_hash == current_head_hash`.

bkchr: > yes, in the v1 networking version it should only accept `parent_head_data_hash == current_head_hash`.

Should be done now.

slumber (commented):
I believe it works based on `insert_advertisement` call in `handle_advertisement`.

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
bkchr: Done @slumber

slumber (approved):

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1042,6 +1062,13 @@ where
 		.get(&relay_parent)
 		.map(|s| s.child("advertise-collation"));
 
+	let peer_data = state.peer_data.get_mut(&peer_id).ok_or(AdvertisementError::UnknownPeer)?;
+
+	if peer_data.version == CollationVersion::V1 && !state.active_leaves.contains_key(&relay_parent)
```
eskimor, line 1067: Is this enough? The collator could still try to build an entire prospective parachain, by using the same relay parent. This is actually the most likely case. I don't see why a collator would use an older relay parent on purpose.

slumber: > Is this enough? The collator could still try to build an entire prospective parachain, by using the same relay parent. This is actually the most likely case. I don't see why a collator would use an older relay parent on purpose.

`insert_advertisement` fails for v1 messages if collator attempts to build more than 1 collation.

eskimor: What we need to check is that the para head of the candidate is already included on the relay chain.

slumber: Here we request persisted validation data from chain
https://github.com/paritytech/polkadot-sdk/blob/7ca2a26ddd36c5b44346fa3efea1bfe1cbbf24a7/polkadot/node/network/collator-protocol/src/validator_side/mod.rs#L1828-L1834
And then validate candidate against it
https://github.com/paritytech/polkadot-sdk/blob/7ca2a26ddd36c5b44346fa3efea1bfe1cbbf24a7/polkadot/node/network/collator-protocol/src/validator_side/collation.rs#L147-L149

eskimor: Uh! Missed that part, thanks @slumber ! All good then!

polkadot/node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1783,39 +1807,37 @@ async fn kick_off_seconding<Context>(
...
+					pending_collation.para_id,
+				)
+				.await?,
+			(CollationVersion::V1, _) =>
+				request_persisted_validation_data(
+					ctx.sender(),
+					candidate_receipt.descriptor().relay_parent,
+					candidate_receipt.descriptor().para_id,
+				)
+				.await?,
```
slumber, line 1834: Previously we would've requested pvd from chain even for V2 protocol in case async backing is disabled.
This is correct, because prospective parachains subsystem doesn't do any job otherwise.

With new version this is not the case anymore, and here it implies that V2 -> async backing is enabled, which is wrong assumption.

So it should be
```rust
(CollationVersion::V2, Some(ProspectiveCandidate { parent_head_data_hash, .. })) 
    if relay_parent_mode.is_enabled() => // <-- this
				request_prospective_validation_data(
					ctx.sender(),
					relay_parent,
					parent_head_data_hash,
					pending_collation.para_id,
				)
				.await?,
```

Issue #1725, opened 2023-09-27
FRAME: Syntax for marking calls "feeless"


liamaharon:
Moving an internal FRAME team discussion to a Github issue.

## Motivation

@gavofyork suggested a new syntax which can be used before a call and have it mean that it can skip all normal transaction fee and account existence requirements.

```rust
#[pallet::feeless_if(|origin: &OriginFor<T>, ticket: &Ticket| -> bool {
	let account = ensure_signed(origin.clone())?;
	some_conditions_here(&account, &ticket)
})]
```

@gupnik volunteered to work on it.

## Solution

@gupnik suggested a solution and requested feedback:

1. Pallet Call enum can optionally have a feeless conditional.
2. Construct Runtime creates a `RuntimeSignedExtension` that takes `OnChargeTransaction`.
3. In `post_dispatch`, we check if the current call has a feeless conditional. If so, the fee is never charged (bypassing the requirement the caller has any balance).

A couple of questions

1. How can `RuntimeSignedExtension` know the amount that was charged originally?
2. Is there a way to test the entire flow using a unit test? Any references that I could use?

## Other questions

- @KiChjang suggested rather than an attribute macro this could be implemented in a way more 'rustic'.
- @liamaharon considered rather than the method returning a `bool`, it returning a `Balance` which could be used to modify the fee in more than a single way way.

bkchr:
> * Pallet Call enum can optionally have a feeless conditional.
> 
> * Construct Runtime creates a `RuntimeSignedExtension` that takes `OnChargeTransaction`.
> 
> * In `post_dispatch`, we check if the current call has a feeless conditional. If so and the condition returns true, the fee is refunded.

All this is already supported by returning [`Pays::No`](https://github.com/paritytech/polkadot-sdk/blob/da937d9f6353e36fdc493a63a319c7d6d259483c/substrate/frame/beefy/src/lib.rs#L234).

liamaharon:
> > * Pallet Call enum can optionally have a feeless conditional.
> > * Construct Runtime creates a `RuntimeSignedExtension` that takes `OnChargeTransaction`.
> > * In `post_dispatch`, we check if the current call has a feeless conditional. If so and the condition returns true, the fee is refunded.
> 
> All this is already supported by returning [`Pays::No`](https://github.com/paritytech/polkadot-sdk/blob/da937d9f6353e36fdc493a63a319c7d6d259483c/substrate/frame/beefy/src/lib.rs#L234).

[...]

PR #1926, opened 2023-10-18
Adds syntax for marking calls feeless

gupnik:
Fixes https://github.com/paritytech/polkadot-sdk/issues/1725

This PR adds the following changes:
1. An attribute `pallet::feeless_if` that can be optionally attached to a call like so:
```rust
#[pallet::feeless_if(|_origin: &OriginFor<T>, something: &u32| -> bool {
	*something == 0
})]
pub fn do_something(origin: OriginFor<T>, something: u32) -> DispatchResult {
     ....
}
```
The closure passed accepts references to arguments as specified in the call fn. It returns a boolean that denotes the conditions required for this call to be "feeless".

2. A signed extension `SkipCheckIfFeeless<T: SignedExtension>` that wraps a transaction payment processor such as `pallet_transaction_payment::ChargeTransactionPayment`. It checks for all calls annotated with `pallet::feeless_if` to see if the conditions are met. If so, the wrapped signed extension is not called, essentially making the call feeless. 

In order to use this, you can simply replace your existing signed extension that manages transaction payment like so:
```diff
- pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
+ pallet_skip_feeless_payment::SkipCheckIfFeeless<
+	Runtime,
+	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
+ >,
```

### Todo
- [x] Tests
- [x] Docs
- [x] Prdoc


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3995851

substrate/frame/support/procedural/src/construct_runtime/expand/call.rs
```diff
@@ -207,5 +219,61 @@ pub fn expand_outer_dispatch(
...
+		#[derive(
+			#scrate::RuntimeDebugNoBound,
+			#scrate::CloneNoBound,
+			#scrate::EqNoBound,
+			#scrate::PartialEqNoBound,
+			#scrate::__private::codec::Encode,
+			#scrate::__private::codec::Decode,
+			#scrate::__private::scale_info::TypeInfo,
+		)]
+		pub struct SkipCheckIfFeeless<T: #scrate::sp_runtime::traits::SignedExtension>(T);
```
xlc, line 233: any reason why we need to generate this from macro?

substrate/frame/support/procedural/src/construct_runtime/expand/call.rs
```diff
@@ -207,5 +219,61 @@ pub fn expand_outer_dispatch(
...
+
+			fn pre_dispatch(
+				self,
+				who: &Self::AccountId,
+				call: &Self::Call,
+				info: &#scrate::sp_runtime::traits::DispatchInfoOf<Self::Call>,
+				len: usize,
+			) -> Result<Self::Pre, #scrate::sp_runtime::transaction_validity::TransactionValidityError> {
+				use #scrate::dispatch::CheckIfFeeless;
+				match call.is_feeless(who) {
```
xlc, line 257: why this is not just `if-else`?

gupnik: Yes, will make the change

substrate/frame/support/procedural/src/construct_runtime/expand/call.rs
gupnik: Not particularly, happy to move it out.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4024428

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4024460

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4024596

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4025528

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4026355

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4026630

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4027015

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4070670

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4070638

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4070904

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4071301

KiChjang:
Could we do something about the syntax? I feel something like the following is much easier to follow:

```rust
pub fn do_something(origin: OriginFor<T>, something: u32) -> DispatchResult {
    #[pallet::feeless_if]
    {
        *something == 0
    }
    /* ... */
}
```

kianenigma:
@KiChjang can you elaborate more? what is difficult to follow now, and how your suggestion improves it. 

I think the syntax is good. The more important complication is the existence of this while the signed extension is not there, or the runtime has no fee paying pallet etc, but that is a different concern.

KiChjang:
This is probably a bigger conversation, but I don't feel like attributes should contain complex expressions. That complexity should be kept within the function body. Besides, the visual indicator for having such an expression in an attribute makes it feel like it won't be run on every extrinsic call, when in fact it does. Most of all, though we're in custom syntax territory, placing the the conditional expression inside of the function body feels the most rustic of all -- look at how we can just reuse the parameters from the function, instead of re-declaring them on the attribute.

gupnik:
> The more important complication is the existence of this while the signed extension is not there, or the runtime has no fee paying pallet etc, but that is a different concern.

I am curious to explore this further, potentially as a follow-up. Do you think that we should have some assertions that validate the presence of this extension if any of the calls have been marked with `feeless_if`? Any other suggestions?

substrate/frame/support/procedural/src/lib.rs
```diff
@@ -1142,6 +1142,33 @@ pub fn call_index(_: TokenStream, _: TokenStream) -> TokenStream {
 	pallet_macro_stub()
 }
 
+/// Each dispatchable may also be annotated with the `#[pallet::feeless_if($closure)]` attribute,
+/// which explicitly defines the condition for the dispatchable function to be feeless.
+///
+/// The first argument for the closure must be `account_id: &AccountIdFor<T>`, while the rest
+/// should be the references to arguments of the dispatchable function.
```
ggwpez, line 1149: Please also explain that it only works for signed dispatchables and that the signed extension must be present in the runtime.

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -2158,7 +2167,10 @@ pub type SignedExtra = (
 	frame_system::CheckEra<Runtime>,
 	frame_system::CheckNonce<Runtime>,
 	frame_system::CheckWeight<Runtime>,
-	pallet_asset_conversion_tx_payment::ChargeAssetTxPayment<Runtime>,
+	pallet_skip_feeless_payment::SkipCheckIfFeeless<
+		Runtime,
+		pallet_asset_conversion_tx_payment::ChargeAssetTxPayment<Runtime>,
```
ggwpez, line 2172: What happens when i start using the syntax but forgot to add this? Will it be ignored or error?

sam0x17:
I actually agree with @KiChjang that attributes feel wrong here, and since we're in an outer macro pattern we can really do whatever we want syntax wise. I imagine something like this:

```rust
pub fn do_something(origin: OriginFor<T>, something: u32) -> DispatchResult {
    feeless_if || *something == 0;
}

pub fn do_something_else(origin: OriginFor<T>, something: u32) -> DispatchResult {
    feeless_if {
        let some_local = !something;
        // example of longer code here
        !something && some_local
    }
}
```

I would add the following restrictions however:
1. you can't have more than one `feeless_if` block per function
2. might also restrict it to always being the first line, if it is specified, if such a restriction is needed. I am unclear on whether this is necessary but it very well might be depending on how the expansion of this works

Could also define a doc-only function generated within pallets called `feeless_if` that takes a closure as a stub so we get docs for it via RA

worth exploring, but happy to approve the attribute version if that really is the direction we want to go

substrate/frame/transaction-payment/skip-feeless-payment/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "pallet-skip-feeless-payment"
+version = "4.0.0-dev"
+authors.workspace = true
+edition.workspace = true
+license = "Apache-2.0"
+homepage = "https://substrate.io"
```
ggwpez, line 7:
```suggestion
license.workspace = true
homepage.workspace = true
```

substrate/frame/transaction-payment/skip-feeless-payment/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "pallet-skip-feeless-payment"
+version = "4.0.0-dev"
```
ggwpez, line 3:
```suggestion
version = "1.0.0-dev"
```

substrate/frame/transaction-payment/skip-feeless-payment/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+authors.workspace = true
+edition.workspace = true
+license = "Apache-2.0"
+homepage = "https://substrate.io"
+repository.workspace = true
+description = "pallet to skip feeless payments"
```
ggwpez, line 9: Some better description please 😋

substrate/frame/transaction-payment/skip-feeless-payment/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+readme = "README.md"
```
ggwpez, line 10:
```suggestion
```

substrate/frame/support/procedural/src/pallet/parse/call.rs
```diff
@@ -162,6 +173,31 @@ pub fn check_dispatchable_first_arg_type(ty: &syn::Type) -> syn::Result<()> {
 	Ok(())
 }
 
+/// Check the syntax is `AccountIdFor<T>`
+pub fn check_feeless_first_arg_type(ty: &syn::Type) -> syn::Result<()> {
+	pub struct CheckFeelessFirstArg;
+	impl syn::parse::Parse for CheckFeelessFirstArg {
+		fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+			input.parse::<syn::Token![&]>()?;
+			input.parse::<keyword::AccountIdFor>()?;
```
ggwpez, line 182: Hm do we really need to hard-code this? Maybe someone uses `<T as super::Config>::AccountId`?  
Does it otherwise just throw a normal rust error or what happens?

gavofyork:
Looks fine. I generally favour attribute macros over syntax alterations as they're easier to ignore. It follows the convention set out and much-used in the pallet weight attribute macro, so I don't buy the concern over having logic within the macro.

substrate/bin/node/runtime/src/lib.rs
gupnik: Yeah, it will be completely ignored right now.

substrate/frame/support/procedural/src/pallet/parse/call.rs
gupnik: Given the suggestion to move to `OriginFor` itself, this won't be needed. Will update accordingly.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4199549

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4199805

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4199837

franciscoaguirre:
> > The more important complication is the existence of this while the signed extension is not there, or the runtime has no fee paying pallet etc, but that is a different concern.
> 
> I am curious to explore this further, potentially as a follow-up. Do you think that we should have some assertions that validate the presence of this extension if any of the calls have been marked with `feeless_if`? Any other suggestions?

Having an assertion or a warning sounds good. Anything that would let the dev know what's wrong if fees are charged and they set up the feeless_if correctly, but forgot to add the extension.

Also, that extension sounds important enough, since it doesn't do anything new if you don't use feeless_if, to add to the templates we have. This would make the new syntax work for new projects.

ggwpez (approved):
LGTM!  

Thinking about the problem again that the pallet was forgotten: the only problem would be that the transactions still pay fees. The developer would probably notice that quickly and it is not a horrible problem...

substrate/frame/support/procedural/src/lib.rs
```diff
@@ -1157,6 +1157,36 @@ pub fn call_index(_: TokenStream, _: TokenStream) -> TokenStream {
 	pallet_macro_stub()
 }
 
+/// Each dispatchable may also be annotated with the `#[pallet::feeless_if($closure)]` attribute,
+/// which explicitly defines the condition for the dispatchable function to be feeless.
```
liamaharon, line 1161:
```suggestion
/// Each dispatchable may be annotated with the `#[pallet::feeless_if($closure)]` attribute,
/// which explicitly defines the condition for the dispatchable to be feeless.
```

substrate/frame/support/procedural/src/lib.rs
```diff
@@ -1157,6 +1157,36 @@ pub fn call_index(_: TokenStream, _: TokenStream) -> TokenStream {
...
+/// 		*something == 0
+/// 	})]
+/// pub fn do_something(origin: OriginFor<T>, something: u32) -> DispatchResult {
+///     ....
+/// }
+/// ```
+///
+/// Please note that this only works for signed dispatchables and requires a signed extension
+/// such as `SkipCheckIfFeeless` as defined in `pallet-skip-feeless-payment` to wrap the existing
+/// payment extension. Else, this is completely ignored and the dispatchable is still charged.
```
liamaharon, line 1179: Can we error at compile time instead of ignore?

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_closure_arg1.stderr
```diff
@@ -0,0 +1,5 @@
+error: Invalid pallet::call, feeless_if closure must have same number of arguments as the dispatchable function
+  --> tests/pallet_ui/call_feeless_invalid_closure_arg1.rs:31:24
+   |
+31 |         #[pallet::feeless_if(|| -> bool { true })]
+   |                              ^
```
liamaharon, line 5: Yeah this is awesome 👌

substrate/frame/support/test/tests/pallet_ui/pass/feeless_call.rs
```diff
@@ -0,0 +1,37 @@
+// This file is part of Substrate.
```
franciscoaguirre, line 1: Should/is this be a UI test if it doesn't check stderr?

franciscoaguirre (approved):

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_closure_arg3.stderr
```diff
@@ -0,0 +1,15 @@
+error[E0308]: mismatched types
+  --> tests/pallet_ui/call_feeless_invalid_closure_arg3.rs:32:31
+   |
+31 |         #[pallet::feeless_if(|_: &OriginFor<T>, _s: &u32| -> bool { true })]
+   |                              --------------------------------------------- arguments to this function are incorrect
+32 |         pub fn foo(_: OriginFor<T>, _something: u64) -> DispatchResult { Ok(()) }
+   |                                     ^^^^^^^^^^ expected `&u32`, found `&u64`
+   |
+   = note: expected reference `&u32`
+              found reference `&u64`
```
liamaharon, line 10: This is confusing, not quite what I'd expect.

I'd expect something more like

```suggestion
32 |         pub fn foo(_: OriginFor<T>, _something: u64) -> DispatchResult { Ok(()) }
   |                                     ^^^^^^^^^^ expected `u32`, found `u64`
   |
   = note: expected reference `u32`
              found reference `u64`
```

I wonder if it's feasible? 

Even better if we could get the error on the `feeless_if` args, since the call args are more likely to be the source of truth.

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_type.stderr
```diff
@@ -0,0 +1,5 @@
+error: expected `|`
+  --> tests/pallet_ui/call_feeless_invalid_type.rs:31:24
+   |
+31 |         #[pallet::feeless_if(0)]
+   |                              ^
```
liamaharon, line 5: Not sure why this doesn't say something like 'expected closure' or 'expected ||' instead of just a single '|'

liamaharon (commented):
I'm playing with the LSP and `rustc` support here. Awesome that you have compile errors for some cases, but I think I found a missed one:  

Here's what happens when I pass a `u32` instead of `&u32` for `new_fee` in the `example-pallet-kitchensink` example:

Screenshot 2023-11-10 at 16 04 10

## LSP

LSP shows red squiggle in wrong place and confusingly tells me there's a `&u32` in the call params (when I didn't change it, it's still a `u32`, and gives me bad advice to fix it.

Screenshot 2023-11-10 at 15 58 29

on hover:
Screenshot 2023-11-10 at 16 01 47

## `rustc`

Gives similarly confusing and unactionable advice 

Screenshot 2023-11-10 at 16 05 56

---

also I wonder why I can't just pass `u32` given it implements `Copy`.

gavofyork:
As a later possible improvement to the syntax, we might consider going from this:

```rust
#[pallet::feeless_if(|origin: &OriginFor<T>, secret: u32| -> bool {
	let ticket = hash(secret);
	if Tickets::<T>::contains_key(ticket) { return false }
	true
})]
pub fn do_something_else(origin: OriginFor<T>, secret: u32) -> DispatchResult {
	let who = ensure_signed(origin.clone())?;
	let ticket = hash(secret);
	ensure!(Tickets::<T>::contains_key(ticket), Error::<T>::BadSecret);
	Tickets::<T>::remove(ticket);
	do_something(&who);
	Ok(())
}
```

to this:


```rust
#[pallet::feeless_on_checkpoint]
pub fn do_something_else(origin: OriginFor<T>, secret: u32) -> DispatchResult {
	let (who, ticket) = #[checkpoint_with_refs] {
		let who = ensure_signed(origin.clone())?;
		let ticket = hash(secret);
		ensure!(Tickets::<T>::contains_key(ticket), Error::<T>::BadSecret);
		Ok((who, ticket))
	}?;

	Tickets::<T>::remove(ticket);
	do_something(&who);
	Ok(())
}
```

In this case, we not only avoid the need to duplicate the check code (similar to the other syntaxes proposed), but crucially we also avoid running it twice (once when checking the tx feelessness and again in the call execution) and thus get access to values yielded by these checks.

The clever thing going on here is that whatever the block following `#[checkpoint]` yields is placed in a newly added component to `origin`, `checkpointed_call_data`. The code roughly expands to:

```rust
#[pallet::feeless_checkpoint]
pub fn do_something_else(origin: OriginFor<T>, secret: u32) -> DispatchResult {
	let (who, ticket) = match origin.checkpointed_call_data.try_into() {
		Ok(CheckpointedCallData::DoSomethingElse(d)) => Some(d),
		_ => None
	}.unwrap_or_else(|| {
		let who = ensure_signed(origin.clone())?;
		let ticket = hash(secret);
		ensure!(Tickets::<T>::contains_key(ticket), Error::<T>::BadSecret);
		Ok((who, ticket))
	})?;

	Tickets::<T>::remove(ticket);
	do_something(&who);
	Ok(())
}
```

`try_into()` here basically attempts to turn the `checkpointed_call_data` from an aggregated enum (ala `Call` etc) into the pallet-local type. It matches on this and if there's nothing there (e.g. because the call is not coming via the transaction system), then it runs the checks. If it is, then it skips them and uses the yielded values instead.

`Origin` would need to become aware of not just `Call` but also this new runtime-aggregate `CheckpointedCallData`.

gupnik:
> As a later possible improvement to the syntax, we might consider going from this:

Created an issue for this: https://github.com/paritytech/polkadot-sdk/issues/2278

substrate/frame/support/test/tests/pallet_ui/pass/feeless_call.rs
gupnik: This UI test is actually tested as a success case. The assertion here is that it should compile successfully.

substrate/frame/support/procedural/src/pallet/parse/call.rs
```diff
@@ -323,6 +339,36 @@ impl CallDef {
...
 				let docs = get_doc_literals(&method.attrs);
 
+				if feeless_attrs.len() > 1 {
+					let msg = "Invalid pallet::call, too many feeless_if attributes given";
+					return Err(syn::Error::new(method.sig.span(), msg))
+				}
+				let feeless_check: Option<ExprClosure> =
+					feeless_attrs.pop().map(|attr| match attr {
+						FunctionAttr::FeelessIf(closure) => closure,
+						_ => unreachable!("checked during creation of the let binding"),
```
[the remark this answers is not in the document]
gupnik: Follows the same syntax as earlier. Creation of `feeless_attrs` ensures that it is `FeelessIf`.

substrate/frame/support/procedural/src/lib.rs
gupnik: Not at this time. We will need to rethink about the overall payment flow as discussed by @kianenigma elsewhere. I think Kian already created an issue - will try to find and refer here.

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_closure_arg3.stderr
gupnik: As per the suggested syntax from @gavofyork,`feeless_if` should accept references to the arguments. But I can try to make the error span a bit better.

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_type.stderr
```diff
@@ -0,0 +1,5 @@
+error: expected `|`
```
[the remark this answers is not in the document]
gupnik: Done

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_type.stderr
gupnik: Done

substrate/frame/support/test/tests/pallet_ui/call_feeless_invalid_closure_arg3.stderr
liamaharon: This compile error asking for refs is pointing to the call args, not `feeless_if` args

gupnik: Done

gupnik:
> I'm playing with the LSP and `rustc` support here. Awesome that you have compile errors for some cases, but I think I found a missed one:

These should be fixed now. Thanks!

Polkadot-Forum:

substrate/frame/support/procedural/src/lib.rs
kianenigma: I have not created an issue, and I still think this is a minor, but pretty obvious error here. This is my original commment: https://github.com/paritytech/polkadot-sdk/issues/1725#issuecomment-1741417841 If you see agree high level, please formalize it in an issue.

substrate/frame/transaction-payment/skip-feeless-payment/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// A transaction fee was skipped.
+		FeeSkipped { who: T::AccountId },
```
kianenigma, line 73: Was it intentional that this is a pallet? as in, is this event really needed, to warrant the existence of a pallet? 

If coded a bit differently, we could make it possible for you to use `SkipCheckIfFeeless` without needing the pallet in the runtime. 

The fact that now, fee-related events are scattered into events coming from two different pallets is IMO ever more reason to bake fees into `frame-system` once and for all.

substrate/frame/transaction-payment/skip-feeless-payment/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+	fn pre_dispatch(
+		self,
+		who: &Self::AccountId,
+		call: &Self::Call,
+		info: &DispatchInfoOf<Self::Call>,
+		len: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		if call.is_feeless(&<T as frame_system::Config>::RuntimeOrigin::signed(who.clone())) {
+			Ok((who.clone(), None))
```
kianenigma, line 123: I think it is worth taking a moment for yourself to conclude the following, and document it for external users as well: 

Even if a transaction is feeless, it still consumes weight and they two are totally independent.

substrate/frame/transaction-payment/skip-feeless-payment/src/lib.rs
gupnik: Yeah, the intent was to entirely skip a call to the payment extension (be it the ones provided in this repo or a custom one) if a call is feeless, but still log that as an event. What could be a possible alternative?

> ever more reason to bake fees into frame-system once and for all
Happy to brainstorm on this. Is there an issue where this is described already? Or should I go ahead and create one?

PR #1927, opened 2023-10-18
Enable "cargo run --bin substrate-node"

pgherveou:


bkchr:
So what would `cargo run` do? Thrown an error as there are multiple packages?

pgherveou:
> So what would `cargo run` do? Thrown an error as there are multiple packages?

It should run polkadot as this is the first binary listed in the 'default-members'
maybe this is a bit silly since you can also `cargo run --package node-cli`. I guess I had some muscle memory from the substrate repo that I wanted to restore, but I can live with that as well

bkchr:
> It should run polkadot as this is the first binary listed in the 'default-members'

Did you tried this or are you guessing?

pgherveou:
> > It should run polkadot as this is the first binary listed in the 'default-members'
> 
> Did you tried this or are you guessing?

Yes confirmed.
if you just specify 
```toml
default-members = [ "substrate/bin/node/cli" ]
```
then cargo run will run substrate-node instead

Having both let you use cargo run --bin <binary-name> without specifying -p

Weirdly swapping the two values
```toml
default-members = [ "substrate/bin/node/cli", "polkadot" ] 
```

does not let you run  substrate-node via `cargo run` to run` ¯\_(ツ)_/¯

bkchr:
> does not let you run substrate-node via `cargo run` to run` ¯_(ツ)_/¯

Because this is then probably the default. And the other ones just bring their "binary names" into the global scope to use them via `-p`. Something along these lines.

bkchr (approved):

liamaharon (approved):

franciscoaguirre (approved):
While on the topic, I'd probably request adding `cumulus/polkadot-parachain` so it can be run simply with `--bin polkadot-parachain` without specifying `-p polkadot-parachain-bin`

seadanda (approved):

lexnv (approved):

jsidorenko (approved):

ggwpez:
I think it needs a rebase for the required CI check.

koushiro:
you could use `cargo run --bin substrate-node --workspace` instead of adding new member into `default-members`

PR #1938, opened 2023-10-19
[WIP] Scheduled `try-state` checks and alerting

liamaharon:
Schedule a Github Action every day to run `try-runtime` `try-state` checks on all runtimes.

This is a 0 -> 1 task to get our `try-state` checks deployed *at all*. In the future, it would be better for us to move these checks to dedicated infrastructure that can run them more frequently, perhaps using `follow-chain`. 

## Task list
- [x] `try-state` passes on all runtimes
  - [x] Westend: needs fix https://github.com/paritytech/polkadot-sdk/issues/2020
  - [x] Rococo: checked Oct 26 ✅
  - [x] Asset Hub Westend: checked Oct 26 ✅
  - [x] Bridge Hub Rococo: checked Oct 26 ✅
  - [x] Contracts Rococo: checked Oct 26 ✅
- [x] Better logging: https://github.com/paritytech/polkadot-sdk/pull/2019
- [ ] Use pre-built artefacts from latest release TODO: looking into why workflow isn't triggered on release
- [x] Alerting (POST to a Matrix channel on any failure) https://github.com/paritytech/devops/issues/2917
- [x] ~See if we can optimise build times in the runner~ we'll have caching to resolve this soon
- [x] Figure out `value passed in; value comes out; qed` error from contracts rococo https://github.com/paritytech/polkadot-sdk/actions/runs/6796531510/job/18476812057?pr=1938#step:7:97)
  - Fixed in: https://github.com/paritytech/try-runtime-cli/pull/55


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4012060

.github/workflows/try-state.yml
```diff
@@ -0,0 +1,100 @@
...
+  schedule:
+    - cron: "0 * * * *"
+
+jobs:
+  check_try_state:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
```
alvicsam, line 18:
```suggestion
      - name: Rust Cache
        uses: Swatinem/rust-cache@v2.7.0
        with:
          cache-on-failure: true
```
We could try use this cache action here if the cache is less than 10 GB.

ggwpez: I wonder if we can just use the build artifacts from the latest master commit. Re-building that over and over again is kind of a waste.

liamaharon: > I wonder if we can just use the build artifacts from the latest master commit. Re-building that over and over again is kind of a waste.

Yes this would be ideal. I didn't know we saved the artefacts - will look into using those thanks.

bkchr:
What is the purpose of running it every hour?

liamaharon:
> What is the purpose of running it every hour?

To notify us asap in the case of any storage invariants breaking.

liamaharon:
> Shouldn't this be some continuous job?

> I think follow-chain should check the storage invariants, or?

Ideally, yes we would have dedicated infrastructure that runs `follow-chain` to check these invariants. 

In reality, there're some drawbacks of using `follow-chain`:
- Requires high up-front and ongoing human and capital cost to maintain and monitor infrastructure for every chain, kept up to date with the latest runtimes.
- Couldn't actually run checks 'continuously' anyway. The checks are too expensive to run every block, so would be executed in `round-robin` mode, checking a subset of pallets each block.

Internally, `follow-chain` does the same thing as `execute-block`. The benefit of using `execute-block` directly is:
- Can define the job as a Github Action. 
    - Trivial for Fellowship or parachain teams to replicate these checks in their own repos for their own runtimes (exponentially less initial and on-going cost compared to spinning up infra and CI/CD for `follow-chain`)
    - Trivial for all developers to monitor and maintain the job, rather than requiring work from other (infrastructure) team to maintain, monitor, and manage access to infra 
- With artifact caching we can run it much more frequently than once an hr, so not like it would take that much longer to be notified of issues compared to `follow-chain`.

For these reasons, I suggested to Kian and he agreed to opt to go with this approach for checking storage invariants. My backlog of work is pretty large, so this option is nice as it allows us to get something out the door quickly that works and doesn't (as far as we could tell) have significant drawbacks compared with dedicated infra running `follow-chain` (which does have some significant drawbacks). If we've missed some important detail, something let us know.

.github/workflows/try-state.yml
```diff
@@ -0,0 +1,62 @@
...
+on:
+  push:
+    branches:
+      - liam-try-state-scheduled-checks
+  schedule:
+    # run once per day at midnight
+    - cron: "0 0 * * *"
+
+env:
+  # TODO: Replace this with paritytech/polkadot-sdk 'latest' release URL once 1.3.0 is published
```
ggwpez, line 13: Can do now?

ggwpez (approved):
nice

.github/workflows/try-state.yml
liamaharon: 1.3.0 did not include the Github workflow that builds the runtimes so we need to wait 1 more release for that to happen automatically, but I'm inclined to upload the debug builds there manually so we can use it.

liamaharon:
Now that we're running full try-state checks in the CI I'm questioning how valuable this cron job will be.

Looking into some alternatives that will hopefully provide more value and be more easily accessible to ecosystem chains without them needing to spin up dedicated CI infra.

PR #1946, opened 2023-10-19
PRDoc new schema

chevdor:
## Overview

This PR brings in the new version of prdoc v0.0.6 and allows:
- local schema
- local config
- local template

It also fixes the existing prdoc files to match the new schema.

## todo

- [ ] add a brief doc/tldr to help contributors get started
- [ ] test CI
- [ ] finalize schema

### DIFF
--- a/prdoc/schema_user.json
+++ b/prdoc/schema_user.json
@@ -1,0 +1,228 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema#",
+    "$id": "https://raw.githubusercontent.com/paritytech/prdoc/master/prdoc_schema_user.json",
+    "version": {
+      "major": 1,
+      "minor": 0,
+      "patch": 0,
+      "timestamp": 20230817152351
+    },
+    "title": "Parity PR Documentation Schema",
+    "description": "JSON Schema definition for the Polkadot SDK PR documentation",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+      "title": {
+        "type": "string",
+        "description": "Title for the PR. This is what will show up in the documentation"
+      },
+
+      "authors": {
+        "description": "Optional: You may provide your name, github handle, or email. That can help finding your 'your' prdoc files later on",
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
+      },
+
+      "tags": {
+        "description": "Optional: You may provide some tags here. This is currently not used downstream but can help you find some prdoc files later on",
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
+      },
+
+      "doc": {
+        "type": "array",
+        "description": "Description of the PR. Provide a description for each relevant audience.\nSee the `audience` property for more documentation about audiences",
+        "items": {
+          "$ref": "#/$defs/doc"
+        },
+        "minItems": 1
+      },
+      "crates": {
+        "description": "You have the option to provide input about the semver level of your changes for each crate\n. This is used during the crate publishing to crates.io and to help users understand the impact of the changes introduced in your PR.\nYou can also set it to `no change` if you know there was no significant semver change",
+        "type": "array",
+        "items": {
+          "$ref": "#/$defs/crate"
+        }
+      },
+      "migrations": {
+        "description": "It is important for users to be aware of migrations.\nMake sure to mention any migrations in the appropriate sub-properties",
+        "type": "object",
+        "properties": {
+          "db": {
+            "type": "array",
+            "nullable": false,
+            "description": "List of the Database Migrations or []",
+            "items": {
+              "$ref": "#/$defs/migration_db"
+            },
+            "minItems": 0,
+            "required": [
+              "name",
+              "description"
+            ]
+          },
+          "runtime": {
+            "type": "array",
+            "nullable": false,
+            "description": "List of the Runtime Migrations or []",
+            "minItems": 0,
+            "items": {
+              "$ref": "#/$defs/migration_runtime"
+            },
+            "required": [
+              "db",
+              "runtime"
+            ]
+          }
+
+        }
+      },
+      "host_functions": {
+        "description": "Lits of the host functions involved in this PR.\nThis is used to generate the host functions documentation.",
+        "type": "array",
+        "minItems": 0,
+        "items": {
+          "$ref": "#/$defs/host_function"
+        }
+      }
+    },
+    "required": [
+      "title",
+      "doc",
+      "crates",
+      "migrations",
+      "host_functions"
+    ],
+    "$defs": {
+      "audience": {
+        "oneOf": [
+          {"const": "Node Dev", "title": "Someone working on the node"},
+          {"const": "Runtime Dev", "title": "Someone working on runtimes, frames, pallets"},
+
+          {"const": "Node Operator", "title": "Someone running a node in general, not specific to validators"},
+          {"const": "Validator", "title": "Someone running a Validator node"},
+
+          {"const": "Parachain Builder", "title": "Someone building a parachain"},
+
+          {"const": "App Builder", "title": "Someone building Apps and dApps on the shoulder of the Polkadot network and its parachains"},
+
+          {"const": "Runtime User", "title": "Someone involved or interested in the governance process or noticeable changes to the runtime and/or its dependencies"}
+
+        ]
+      },
+      "crate": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "semver": {
+            "$ref": "#/$defs/semver"
+          },
+          "note": {
+            "type": "string"
+          }
+        }
+      },
+      "semver": {
+        "type": "string",
+        "enum": [
+          "major",
+          "minor",
+          "patch",
+          "no change"
+        ]
+      },
+      "migration_db": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "name",
+          "description"
+        ]
+      },
+      "migration_runtime": {
+        "type": "object",
+        "properties": {
+          "pallet": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "pallet",
+          "description"
+        ]
+      },
+      "doc": {
+        "type": "object",
+        "description": "You have the the option to provide different description of your PR for different audiences.",
+        "additionalProperties": false,
+        "properties": {
+          "audience": {
+            "description": "The selected audience",
+            "$ref": "#/$defs/audience"
+          },
+          "title": {
+            "type": "string",
+            "description": "Optional title override for the PR and for the current audience"
+          },
+          "description": {
+            "description": "Description of the change",
+            "type": "string"
+          },
+          "notes": {
+            "description": "Optional relevant note",
+            "$ref": "#/$defs/array_of_strings"
+          }
+        }
+      },
+      "array_of_strings": {
+        "description": "An array of strings that can be empty",
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
+      },
+      "host_function": {
+        "type": "object",
+        "additionalProperties": false,
+        "description": "List of host functions and their properties",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "enabled": {
+            "type": "boolean"
+          },
+          "description": {
+            "type": "string"
+          },
+          "notes": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "name",
+          "enabled",
+          "description"
+        ]
+      }
+    }
+  }
--- a/prdoc/schema_user.json
+++ b/prdoc/schema_user.json
@@ -7,7 +7,7 @@
       "patch": 0,
       "timestamp": 20230817152351
     },
-    "title": "Parity PR Documentation Schema",
+    "title": "Polkadot SDK PRDoc Schema",
     "description": "JSON Schema definition for the Polkadot SDK PR documentation",
     "type": "object",
     "additionalProperties": false,
--- a/.github/workflows/check-prdoc.yml
+++ b/.github/workflows/check-prdoc.yml
@@ -10,7 +10,6 @@ env:
   REPO: ${{ github.repository }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   GITHUB_PR: ${{ github.event.pull_request.number }}
-  MOUNT: /prdoc
   ENGINE: docker
 
 jobs:
@@ -21,7 +20,7 @@ jobs:
         run: |
           echo "Pulling $IMAGE"
           docker pull $IMAGE
-          docker run --rm $IMAGE --version
+          docker run --rm -v $PWD:/repo $IMAGE --version
 
       - name: Check if PRdoc is required
         id: get-labels
@@ -32,10 +31,11 @@ jobs:
           echo "Labels: ${labels}"
           echo "labels=${labels}" >> "$GITHUB_OUTPUT"
 
-      - name: No PRdoc required
+      # TODO: Here we should check if the prdoc is valid in case the user added one nonetheless
+      - name: Early exit if PR is silent
         if: ${{ contains(steps.get-labels.outputs.labels, 'R0') }}
         run: |
-          echo "PR detected as silent, no PRdoc is required, exiting..."
+          echo "PR detected as silent, no PRDoc is required, exiting..."
           exit 0
 
       - name: Checkout repo
@@ -45,5 +45,5 @@ jobs:
       - name: PRdoc check for PR#${{ github.event.pull_request.number }}
         if: ${{ !contains(steps.get-labels.outputs.labels, 'R0') }}
         run: |
-          echo "Checking for PR#${GITHUB_PR} in $MOUNT"
-          $ENGINE run --rm -v $PWD/prdoc:/doc $IMAGE check -n ${GITHUB_PR} || true
+          echo "Checking for PR#${GITHUB_PR}"
+          $ENGINE run --rm -v $PWD:/repo $IMAGE check -n ${GITHUB_PR} || true
--- a/.prdoc.toml
+++ b/.prdoc.toml
@@ -1,0 +1,7 @@
+# Config file for prdoc, see https://github.com/paritytech/prdoc
+
+version = 1
+schema = "prdoc/schema_user.json"
+output_dir = "/prdoc"
+prdoc_folders = ["prdoc"]
+template = "template.prdoc"
--- a/.prdoc.toml
+++ b/.prdoc.toml
@@ -2,6 +2,6 @@
 
 version = 1
 schema = "prdoc/schema_user.json"
-output_dir = "/prdoc"
+output_dir = "prdoc"
 prdoc_folders = ["prdoc"]
-template = "template.prdoc"
+template = "prdoc/.template.prdoc"
--- a/prdoc/.template.prdoc
+++ b/prdoc/.template.prdoc
@@ -3,10 +3,6 @@
 
 title: ...
 
-authors: []
-
-tags: []
-
 doc:
   - audience: Node Dev
     description: |
--- a/prdoc/pr_1226.prdoc
+++ b/prdoc/pr_1226.prdoc
@@ -1,9 +1,9 @@
 title: Removed deprecated `Balances::transfer` and `Balances::set_balance_deprecated` functions.
 
 doc:
-  - audience: Builder
+  - audience: Runtime User
     description: The Balances pallet's dispatchables `set_balance_deprecated` and `transfer` were deprecated in [paritytech/substrate#12951](https://github.com/paritytech/substrate/pull/12951) and have now been removed.
-    notes: 
+    notes:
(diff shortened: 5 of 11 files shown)


prdoc/schema_user.json
```diff
@@ -0,0 +1,228 @@
...
+          {"const": "Runtime Dev", "title": "Someone working on runtimes, frames, pallets"},
+
+          {"const": "Node Operator", "title": "Someone running a node in general, not specific to validators"},
+          {"const": "Validator", "title": "Someone running a Validator node"},
+
+          {"const": "Parachain Builder", "title": "Someone building a parachain"},
+
+          {"const": "App Builder", "title": "Someone building Apps and dApps on the shoulder of the Polkadot network and its parachains"},
+
+          {"const": "Runtime User", "title": "Someone involved or interested in the governance process or noticeable changes to the runtime and/or its dependencies"}
```
bkchr, line 113: Please update to the audience as discussed in the forum.

prdoc/schema_user.json
```diff
@@ -0,0 +1,228 @@
...
+          "description": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "name",
+          "description"
+        ]
+      },
```
bkchr, line 156: These should be optional parts of the individual audiences.

prdoc/schema_user.json
```diff
@@ -0,0 +1,228 @@
...
+          "notes": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "name",
+          "enabled",
+          "description"
+        ]
+      }
```
bkchr, line 226: Same.

prdoc/schema_user.json
```diff
@@ -0,0 +1,228 @@
...
+        "type": "object",
+        "additionalProperties": false,
+        "description": "List of host functions and their properties",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "enabled": {
+            "type": "boolean"
+          },
```
bkchr, line 213: This depends on the runtime.

bkchr (changes requested):

prdoc/schema_user.json
chevdor: Fixed in https://github.com/paritytech/polkadot-sdk/pull/1946/commits/83982d5d8aa3d956d0264ef2a025c8774d42abf4

prdoc/schema_user.json
chevdor: Same comment than for the `migration_db `.

What could be done however if that makes it easier is to not have `db` and `runtime` embeded under a `migrations` section but leave it flat. It may be even easier this way.

So instead of:
```
migrations:
   db: []
   runtime: []
```

we could have:
```
db_migration: []
runtime_migration: []
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4050926

prdoc/schema_user.json
bkchr: What will you do with this information that there is a db migration?

A proper prdoc would mention this migration in the proper audience any way to tell the user that it is there. If you only want this for some kind of filtering, fine. However then just add it as optional fields to audience and just a bool to indicate that there is either a db migration or a runtime migration.

chevdor: The prdoc just holds the information, it does not know how we use it. The prdoc file does notthing.

During the release, all the prdoc files involved in the release will be fed into one or more templates (tbd).
Those will define what we actually do with the informations.

For the  example with `migration_db`, this will be shown to some audiences.
@kianenigma gave some more examples [here](https://forum.parity.io/t/pr-documentation-prdoc/2065/28?u=wilfried_kopp). For the `migration_db`, the information will be shown to the `Runtime dev` only.

chevdor:
@bkchr is the current schema in an acceptable state to get started ?

bkchr:
```
doc:
  - audience: Node Dev
    description: |
      ...
    db_migration: true
    runtime_migration: true
    host_function: true
```

`db_migration`, `runtime_migration` and `host_function` should be optional. It makes no sense to have them on the "global" level or then somewhere else decide to which audience we want to show these. I also only see these as "information" for further tooling, because I would already expect that the `Runtime Dev` description mentions the migration and how to use it etc. So, these extra fields are just for better machine processing of these information.

chevdor:
> It makes no sense to have them on the "global" level

It does, because those are facts that need to be described, we don't need to know anything about the audiences to know how to descrive those facts.

Having them at a top level makes it more convenient to fill up.

Also if we decide to change WHO sees WHAT later on, having this data separate means we only need to adjust the templates, and not the entire content of the prdocs.

> ... to which audience we want to show these. 

Yes, that's the idea. This will be done in the templates.

> I also only see these as "information" for further tooling, because I would already expect that the Runtime Dev description mentions the migration and how to use it etc. 

Dev don't need to provide the info in multiple places if they provide the info in the dedicated field.
Structurring the document is indeed to optimize the usage of the content with tooling.
I think this will start making more sense as we get the first versions of the release notes based on the pdocs out.

> So, these extra fields are just for better machine processing of these information.

Yes, partly, It allows us to better use the information than a generic chunk of text and will make it easier to document since we can let the templates deal with the formatting and not bother with that in the content.

> `db_migration`, `runtime_migration` and `host_function` optional

I would rather not hold this PR back too long due to that so if you insist, I will make them optional but note that making them optional means we will have no garanty or way to ensure that those did not get "forgotten". It also will make the templates more complex.

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+      "host_functions"
+    ],
+    "$defs": {
+      "audience": {
+        "oneOf": [
+          {"const": "Node Dev",
+          "title": "Those who build around the client side code. Alternative client builders, SMOLDOT, those who consume RPCs. These are people who are oblivious to the runtime changes. They only care about the meta-protocol, not the protocol itself."},
+
+          {"const": "Runtime Dev",
+          "title": "All of those who rely on the runtime. A parachain team that is using a pallet. A DApp that is using a pallet. These are people who care about the protocol (WASM), not he meta-protocol (client)."},
```
EgorPopelyaev, line 107: Small typo he -> the
```suggestion
          "title": "All of those who rely on the runtime. A parachain team that is using a pallet. A DApp that is using a pallet. These are people who care about the protocol (WASM), not the meta-protocol (client)."},
```

EgorPopelyaev (approved):

coderobe (approved):

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+            "description": "Optional title override for the PR and for the current audience"
+          },
+          "description": {
+            "description": "Description of the change",
+            "type": "string"
+          },
+          "notes": {
+            "description": "Optional relevant note",
+            "$ref": "#/$defs/array_of_strings"
+          }
```
bkchr, line 192: What is the difference between notes and description?

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          },
+          "notes": {
+            "type": "string"
+          }
```
bkchr, line 215: Again information that is probably already available in the audience description.

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+          "type": "string"
+        }
+      },
+      "host_function": {
+        "type": "object",
+        "additionalProperties": false,
+        "description": "List of host functions and their descriptions",
+        "properties": {
+          "name": {
+            "type": "string"
```
bkchr, line 208: What would be the name? 

I'm still not convinced that this should be more than a `bool`.

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+      },
+      "migration_db": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          }
```
bkchr, line 148: Duplication...

prdoc/schema_user.json
```diff
@@ -0,0 +1,223 @@
...
+      "doc": {
+        "type": "array",
+        "description": "Description of the PR. Provide a description for each relevant audience.\nSee the `audience` property for more documentation about audiences",
+        "items": {
+          "$ref": "#/$defs/doc"
+        },
+        "minItems": 1
+      },
+      "crates": {
+        "description": "You have the option to provide input about the semver level of your changes for each crate\n. This is used during the crate publishing to crates.io and to help users understand the impact of the changes introduced in your PR.\nYou can also set it to `no change` if you know there was no significant semver change",
```
bkchr, line 45: If you change something in  a crate, what is a "no change" ?

bkchr (changes requested):

bkchr:
> I would rather not hold this PR back too long due to that so if you insist, I will make them optional but note that making them optional means we will have no garanty or way to ensure that those did not get "forgotten"

Document them properly. And yes, please make them optional.

prdoc/pr_1946_prdoc_new_schema.prdoc
```diff
@@ -0,0 +1,23 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://github.com/paritytech/prdoc
+
+title: New PRDoc Schema
+authors: [ chevdor ]
+tags: [ prdoc, release ]
```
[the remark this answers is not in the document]
chevdor: There are good reasons for adding those.

First, both are optional. They can be omited entirely.

Both are documented:
- `authors`: Optional: You may provide your name, github handle, or email. That can help finding your 'your' prdoc files later on
- `tags`: Optional: You may provide some tags here. This is currently not used downstream but can help you find some prdoc files later on

> The author is not required and can be fetched from the pr if required

Indeed, but this is very slow, especially when fetching the information for several PRs. 
This allows users who decided to add the optional infos to find their pr/prdocs quickl, wtihout any GH API, token, etc... 

We can plug some automation later on that auto-fills the content for the Author from GH to make it even easier.
That means the field(s) needs to be "allowed" in the schema.

prdoc/schema_user.json
chevdor: Good point. This is mainly to make prdoc easier to write and to control the rendering better.

The description is a chunk of text. It could end up with a list of notes/notice/heads up in theory but this is not optimum.

The `notes` are optional and managed as an array of strings.
Having `notes` like this allows writing:
```
doc:
  - audience: Node Operator
    description: |
      lorem ipsum
   notes:
      - Make sure to do this
      - Make sure you have that
```

instead of:
```
doc:
  - audience: Node Operator
    description: |
      lorem ipsum
      NOTES:
          - Make sure to do this
          - Make sure you have that
```

That may look like a small difference but it mainly allows controlling the rendering in the templates and not in the content of the prdoc. 

For instance, we may want the notes to be collapsable in the rendered markdown, without having to write those `<details><summary>...</summary></details>` here.

prdoc/schema_user.json
chevdor: A sensible choice would be the name of the host_function to allow readers having a starting point to seach the code/rustdoc. For now, the content would look like:
```
host_functions:
   - name: RuntimeTasks::spawn
     description: Wasm host function for spawning task.
     notes:
        - This should not be used directly. Use `sp_tasks::spawn` instead.
```

prdoc/schema_user.json
chevdor: This is the description of the db migration itself. This is not duplicated.

prdoc/schema_user.json
chevdor: Say you work on the rustdoc of the crate or you do refactoring that does ont affect the interfaces, this will be neither a semver change considered as `major`, `minor` or `patch` (we may consider this as `patch` though, more on that below).

I am not a big fan of the `no change` term to be honest, maybe you have a better idea ?
Alternatively, we could also dump this 4th choice and stick with `major`, `minor` or `patch` and default to `patch`.

prdoc/pr_1946_prdoc_new_schema.prdoc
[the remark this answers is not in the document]
bkchr: We want to have something for now and not want to add stuff for some potential future. Start small and easy. So, please remove.

prdoc/schema_user.json
bkchr: Again, over complicating it. Please keep it simple. So, please remove `Notes`.

TBH, I'm not entirely convinced that  yaml was such a great idea any way.

prdoc/schema_user.json
bkchr: You will mention in the `Node Operator` audience that there is a db migration. Explain it and say that it isn't backwards compatible. So, this here would be duplicate.

prdoc/schema_user.json
bkchr: The docs are a good example. However, releasing them also eagerly is probably a good idea. I mean you could also just omit a crate there to have it released.

prdoc/schema_user.json
bkchr: Host functions will not be added anymore without a RFC. Thus, we should not need to explain it here. More just mention that it was added.

chevdor: Sure, the addition of that field is based on the past but if this barely ever happen again, I can drop the entire property and it will have to be mentioned in the audience.

prdoc/schema_user.json
chevdor: The main consumer of this data is @Morganamilo with `parity-publish `, I am fine with whatever option works best with the downstream tooling and keeps it simple.

@Morganamilo what do you think ?

prdoc/schema_user.json
chevdor: It is a bad idea to add the db migration (to take that as an example) as part of an audience **in the schema** (sure it will in the template), that is not DRY and requires ppl to remember what data goes "into" which audience, making thing way more complex than they should be.

There is no need for an author to even think where a db migration should appear in the end, they only need to mention whether there is one or not.

prdoc/pr_1946_prdoc_new_schema.prdoc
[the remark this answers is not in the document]
chevdor: I removed `authors` and `tags` for now.

prdoc/schema_user.json
```diff
@@ -0,0 +1,224 @@
...
+          "title": "Runtime Dev",
+          "description": "All of those who rely on the runtime. A parachain team that is using a pallet. A DApp that is using a pallet. These are people who care about the protocol (WASM), not the meta-protocol (client)."},
+
+          {"const": "Node Operator",
+          "title": "Node Operator",
+          "description": "Those who don't write any code and only run code."},
+
+          {"const": "Runtime User",
+          "title": "Runtime User",
+          "description": "Token holders who don't run anything but are involved in the network."}
```
bkchr, line 107:
```suggestion
          "description": "Anyone using the runtime. This can be a token holder or a dev writing a front end for a chain."}
```

prdoc/schema_user.json
```diff
@@ -0,0 +1,224 @@
...
+      },
+      "semver": {
+        "type": "string",
+        "enum": [
+          "major",
+          "minor",
+          "patch",
+          "no change"
+        ]
+      },
```
bkchr, line 134: I thought about this again. We probably want to have the versions numbers in the master branch any way. So, why not just bump the version in the crates directly? 

We would require one CI job that ensures that we don't bump a version multiple times, before we do a release again. However, this should be fairly simple.

CC @Morganamilo

docs/prdoc.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Tips
+
+The PRDoc schema is defined in each repo and usually is quite restrictive.
+You cannot simply add a new property to a `PRDoc` file unless the Schema allows it.
+
+There are however a few convenience optional properties that could be useful to authors:
+- `authors`: An array of authors (strings). That can help an author find their PRs easily
+- `tags`: Array of strings. Those are unrelated to the Github Labels and provided for
+  authors to flag their changes as  they wish, in order to find PR and Documentation quicker
```
bkchr, line 76:
```suggestion
```

docs/prdoc.md
```diff
@@ -0,0 +1,76 @@
...
+- `Node Dev`: Those who build around the client side code. Alternative client builders, SMOLDOT, those who consume RPCs.
+   These are people who are oblivious to the runtime changes. They only care about the meta-protocol, not the protocol
+   itself.
+
+- `Runtime Dev`: All of those who rely on the runtime. A parachain team that is using a pallet. A DApp that is using a
+   pallet. These are people who care about the protocol (WASM), not the meta-protocol (client).
+
+- `Node Operator`: Those who don't write any code and only run code.
+
+- `Runtime User`: Token holders who don't run anything but are involved in the network
```
bkchr, line 66:
```suggestion
- `Runtime User`: Anyone using the runtime. This can be a token holder or a dev writing a front end for a chain.
```

bkchr:
We are almost there! Only thing we should probably change is the crate versioning stuff that we should do directly in the `Cargo.toml`.

chevdor:
> Only thing we should probably change is the crate versioning stuff that we should do directly in the `Cargo.toml`.

@Morganamilo what is your take on that part ?
I can adapt/remove the `crates` part, all fine as long as it works for you in the end.
If there is an alternative to the prdoc to manage semver and we can use it right now, we probably should go for the alternative.

prdoc/schema_user.json
Morganamilo: I'm not a fan of this idea as it's less clearly visible and lacks an explicit "no change" option.

bkchr: Why is it less clearly visible? I mean the version is directly bumped in the `Cargo.toml`.

Morganamilo: It's hard to see if a crate was bumped, what kind of bump, if it's been bumped for multiple reasons, what PR the bump was associated with etc. And it lacks an explicit "this crate was changed but not in a way that changes anything".

Thinking from a tooling perspective it will be easier to scan PRdocs. Both for the automated releases and for displaying what changed via some sort of status command.

bkchr: > It's hard to see if a crate was bumped, what kind of bump

This is an easy git diff between the latest release and master. Not really complicated.

> , if it's been bumped for multiple reasons,

Between releases there should be only one bump of the major, or one of minor and one of patch. If there was a bump of patch and then minor, patch is reverted etc.

In generally people are already wanting to know what is the latest crate version in master. This would not be solved by adding this stuff to prdoc.



> Both for the automated releases and for displaying what changed via some sort of status command.

This is a git blame with some more logic.

bkchr: I'm fine if we may mention the "affected" crates, if that makes things easier (while I still think that this could be automated).

bkchr:
> @bkchr is there anything left on your end that prevents merging this PR ?

https://github.com/paritytech/polkadot-sdk/pull/1946#discussion_r1382173292

chevdor:
@bkchr you mentioned:
> I'm fine if we may mention the "affected" crates, if that makes things easier (while I still think that this could be automated).

That means we can leave the `crates` section as we have it today since the intend is indeed to mention the crates.
I agree that tooling to fill up the crates sections would be helpful. For the time being, whether we fill up the `crates` section manually or using a tool does not affect the schema.

bkchr:
@chevdor then please remove the `semver` field and I will approve it.

bkchr (approved):

sergejparity (approved):

the-right-joyce:
wen merge?

docs/CONTRIBUTING.md
```diff
@@ -93,22 +93,12 @@ The reviewers are also responsible to check:
 
 All Pull Requests must contain proper title & description.
 
-Some Pull Requests can be exempt of `prdoc` documentation, those
-must be labelled with
+Some Pull Requests can be exempt of `prdoc` documentation, those must be labelled with
 [`R0-silent`](https://github.com/paritytech/labels/blob/main/ruled_labels/specs_polkadot-sdk.yaml#L89-L91).
 
 Non "silent" PRs must come with documentation in the form of a `.prdoc` file.
```
mrcnski, line 99: Is this being enforced? I haven't been prompted by CI to write a prdoc yet, but have been really excited  for it.

chevdor: The check currently runs **but** the outcome is ignored. So today already you can see in the CI logs whether your PRdoc looks OK or not.
It will be enforced once this PR is merged.

prdoc/schema_user.json
```diff
@@ -0,0 +1,212 @@
...
+          "title": "Node Dev",
+          "description": "Those who build around the client side code. Alternative client builders, SMOLDOT, those who consume RPCs. These are people who are oblivious to the runtime changes. They only care about the meta-protocol, not the protocol itself."},
+
+          {"const": "Runtime Dev",
+          "title": "Runtime Dev",
+          "description": "All of those who rely on the runtime. A parachain team that is using a pallet. A DApp that is using a pallet. These are people who care about the protocol (WASM), not the meta-protocol (client)."},
+
+          {"const": "Node Operator",
+          "title": "Node Operator",
+          "description": "Those who don't write any code and only run code."},
```
mrcnski, line 103: Is it intended that there is no "Validator" option? We would just choose "Node Operator" for that case?

bkchr: Yes. Most of the changes are equally for both groups and normal node operators are also able to just skip the one validator message for every 5th release.

alvicsam (approved):

PR #1948, opened 2023-10-19
`xcm`: Change `TypeInfo::path` to not include `staging`

bkchr:
The `xcm` crate was renamed to `staging-xcm` to be able to publish it to crates.io as someone as squatted `xcm`. The problem with this rename is that the `TypeInfo` includes the crate name which ultimately lands in the metadata. The metadata is consumed by downstream users like `polkadot-js` or people building on top of `polkadot-js`. These people are using the entire `path` to find the type in the type registry. Thus, their code would break as the type path would now be [`staging_xcm`, `VersionedXcm`] instead of [`xcm`, `VersionedXcm`]. This pull request fixes this by renaming the path segment `staging_xcm` to `xcm`.

This requires: https://github.com/paritytech/scale-info/pull/197

### DIFF
--- a/polkadot/xcm/src/tests.rs
+++ b/polkadot/xcm/src/tests.rs
@@ -181,3 +181,12 @@ fn encode_decode_versioned_xcm_v3() {
 	let decoded = VersionedXcm::decode(&mut &encoded[..]).unwrap();
 	assert_eq!(xcm, decoded);
 }
+
+// With the renaming of the crate to `staging-xcm` the naming in the metadata changed aswell and
+// this broke downstream users. This test ensures that the name in the metadata isn't changed.
+#[test]
+fn ensure_type_info_is_correct() {
+	let type_info = VersionedXcm::<()>::type_info();
+
+	assert_eq!(type_info.path.segments, vec!["xcm", "VersionedXcm"]);
+}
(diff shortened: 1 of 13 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4023777

polkadot/xcm/src/tests.rs
```diff
@@ -181,3 +181,12 @@ fn encode_decode_versioned_xcm_v3() {
 	let decoded = VersionedXcm::decode(&mut &encoded[..]).unwrap();
 	assert_eq!(xcm, decoded);
 }
+
+// With the renaming of the crate to `staging-xcm` the naming in the metadata changed aswell and
```
franciscoaguirre, line 185:
```suggestion
// With the renaming of the crate to `staging-xcm` the naming in the metadata changed as well and
```

franciscoaguirre (approved):

KiChjang (approved):
Bit unfortunate that we have to scatter this attribute everywhere where we drive TypeInfo, but I don't see much alternatives that are less intrusive.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4023915

acatangiu (approved):

lexnv (approved):

xlc:
does this means we should avoid using polkadot-sdk until this is merged?

bkchr:
> does this means we should avoid using polkadot-sdk until this is merged?

It will be merged in some minutes and we will backport it.

PR #1685, opened 2023-09-24
change prepare worker to use fork instead of threads

Jpserrat:
# Description
@mrcnski @s0me0ne-unkn0wn This is the implementation for the prepare worker. Once you guys are okay with this implementation I'll do the same for the execute worker as suggested by @mrcnski 

Closes #574

Kusama address: FkB6QEo8VnV3oifugNj5NeVG3Mvq1zFbrUu4P5YwRoe5mQN

PR #1958, opened 2023-10-20
Refactor `ValidationError`

### LINKED ISSUE #660: PVF: Add a new class of "possibly invalid" errors
# ISSUE

## Overview

PVF execution plays a critical role in our dispute process. If we execute a candidate block and find it is invalid, we vote against it. If some validators' votes do not agree, a dispute is initiated.

Now, in the real world we may try to execute a candidate and it fails due to a hardware fault, operator error, some rare bug, etc. In this case, voting against would initiate a dispute, which is not what we want -- it may get the validator slashed!

So, right now we have two main failure states:

1. `InvalidCandidate`, where we always vote against.
1. `InternalError`, where we never vote against, but do retry once (since https://github.com/paritytech/polkadot/pull/7011) to let transient error conditions clear.

And we do actually have a third state: `InvalidCandidate::AmbiguousWorkerDeath`, which happens when the worker process dies for an unknown reason. In this case we retry once, but if it still fails, then we **do** vote against. This would happen if the PVF execution takes up a large amount of memory, causing OOM, in a way that is reproducible.

## Proposal

So, I am proposing that we extend this third category, giving it a catch-all name like `PossiblyInvalid`. We would retry these, and only on continued failure deem the candidate invalid.

Other errors that we could treat this way:

1. **Timeout due to exceeded CPU time.** Currently this always results in `InvalidCandidate`. But, under conditions of high local load we may do very little work, while still counting CPU time, eventually timing out for a valid candidate. This can be retried after a delay, in the hopes that the load died down. We have to be careful though. It may lead to yet more increase to load, on the other hand we may prevent disputes by retrying! 
  a. (We may also want to detect conditions of high load and deal with it somehow.)
  b. NOTE: There may not be enough time for a retry in candidate validation from backing.
1. We can treat **`RuntimeConstruction`** this way. Right now it always results in Invalid, though it should be a local issue. https://github.com/paritytech/polkadot-sdk/issues/661 would be the full fix, but in the meantime it would make sense to retry in this case before voting against.
1. Any others?

Having this separate variant would also just make the code easier to reason about, and easier to add more "retry-then-vote-against" cases in the future.

Also, it would be nice to make these three distinct categories clear in the documentation!

### LINKED ISSUE #661: PVF: Consider treating `RuntimeConstruction` as an internal execution error
# ISSUE

## Overview

[`sc_executor_common::error::Error::RuntimeConstruction`](https://github.com/paritytech/substrate/blob/2e94e25395/client/executor/common/src/error.rs#L88) should theoretically only happen due to some local issue, and not due to the PVF itself. Therefore when it arises during execution we may want to treat it as internal (i.e. not vote against -- possibly raising a dispute -- as we do right now). We just have to ensure that `RuntimeConstruction` is *never* constructed due to an issue with the PVF, and document it as such on the Substrate side.

We may be able to treat some of the other error variants the same way. This would help to **prevent disputes due to local issues**.

## Important note

*Credit to @eskimor for this section.*

In general, for errors not raising a dispute we have to very careful. This is only sound, if we either:

1. Ruled out that error in pre-checking. If something is not checked in pre-checking, even if candidate independent we must raise a dispute.
1. We are 100% confident that it is a hardware/local issue: Like corrupted file, etc.

Reasoning: Otherwise it would be possible to register a PVF where candidates can not be checked, but we don't get a dispute - so nobody gets punished. Second we end up with a finality stall that is not going to resolve!

## Related issue

Somewhat related issue about `WasmError::Other` variant (which gets turned into `Error::RuntimeConstruction`):
https://github.com/paritytech/substrate/issues/13853

### LINKED ISSUE #733: Preparation errors should no longer be a `InvalidCandidate` variant

eagr:
closes #660 
closes #661 
closes #733


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4031812

mrcnski:
Woah, awesome! I totally missed this. Is it ready for review @eagr ?

eagr:
Not ready yet. I plan to continue on this after the artifact pr. :)

mrcnski:
Hey @eagr, quick update from me! Getting https://github.com/paritytech/polkadot-sdk/pull/1685 merged is my highest priority right now, and as part of that PR I need to do a somewhat big error refactor. There hopefully won't be *too* many conflicts with this work, but it's pretty substantial, so consider this a heads-up! I apologize for the delays and any inconvenience, but after that PR goes in I'll be much more free to focus on other things, like reviewing the artifacts PR to unblock this one. :)

eagr:
Isn't having merge conflicts the norm? :) Totally understand. And usually I wouldn't fixate on something. I'll just move on to the next thing, like green threads do, so no worries man.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4230235

eagr:
Gonna redo this based on master as it's probably faster and easier to get right than trying to solve the conflicts. :) @mrcnski

PR #1959, opened 2023-10-20
Add defaults for pallet-xcm

franciscoaguirre:
Somewhat related to #2169.
It's not XcmConfig defaults, but pallet-xcm, which is also useful.


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4432132

kianenigma (commented):
I haven't really seen this, I think it is a great start. I think if you bring it back up to date, it should be almost merge-able. 

@gupnik can you provide context on using `drive_impl` for a trait other than `#[pallet::config] trait Config { }`. I am sure it is possible, but some guidelines or best practices might be useful.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -187,29 +187,33 @@ pub mod pallet {
...
+		/// A lockable currency.
 		type Currency: LockableCurrency<Self::AccountId, Moment = BlockNumberFor<Self>>;
 
 		/// The `MultiAsset` matcher for `Currency`.
+		#[pallet::no_default_bounds]
 		type CurrencyMatcher: MatchesFungible<BalanceOf<Self>>;
 
+		// TODO: Would love to add a default for this, but I need access to `RuntimeOrigin`
+		// from the derive_impl
+		#[pallet::no_default]
```
kianenigma, line 204: What is usually the default that you would want to provide?

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -275,6 +274,89 @@ pub mod pallet {
...
+		use frame_support::{
+			derive_impl, register_default_impl,
+			traits::{Everything, ConstU32, Equals},
+		};
+		use xcm_builder::{
+			IsConcrete,
+			LocationWithAssetFilters,
+		};
+
+		pub struct TestDefaultConfig;
```
kianenigma, line 308: I belive we should be able to make `struct TestDefaultConfig` be `struct TestDefaultConfig<Some, Further, Parameterization>` if you are unsure about any of the things hardcoded in the following `parameter_types!`.

polkadot/xcm/pallet-xcm/src/lib.rs
gupnik: You should be able to use `no_default_bounds` for such cases.

gupnik:
> @gupnik can you provide context on using `drive_impl` for a trait other than `#[pallet::config] trait Config { }`. I am sure it is possible, but some guidelines or best practices might be useful.

Yeah, nothing in `derive_impl` is specific for the pallet config. We already use this for tests. https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/test/tests/derive_impl.rs is one such example.

bkchr:
@franciscoaguirre will you pick this up again?

franciscoaguirre:
I want to but I've been having other priorities

polkadot/xcm/pallet-xcm/src/lib.rs
franciscoaguirre: That would be extremely helpful, however, I'm getting an error when trying to specify these generics in the calling `derive_impl(TestDefaultConfig<Some, Further, Parametrization>)`. Is this supposed to work or it should be added to the macro?

kianenigma: I don't think there is a fundamental blocker for it, but it is quite likely that the syntax is not supported. Please proceed with @gupnik.

PR #1969, opened 2023-10-20
paras-scheduler: Fix migration to V1

bkchr:
The migration was missing to migrate `AvailabilityCores`. If this isn't migrated, all parachains in the availability phase would stall until the next session is started. This pull request fixes this by migrating this data. Besides that it is doing some cosmetics.

### DIFF
--- a/polkadot/runtime/parachains/src/assigner.rs
+++ b/polkadot/runtime/parachains/src/assigner.rs
@@ -53,7 +53,8 @@ impl<T: Config> Pallet<T> {
 	fn is_bulk_core(core_idx: &CoreIndex) -> bool {
 		let parachain_cores =
 			<ParachainAssigner<T> as AssignmentProvider<BlockNumberFor<T>>>::session_core_count();
-		(0..parachain_cores).contains(&core_idx.0)
+
+		core_idx.0 < parachain_cores
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ggwpez (approved):
This should have been caught when we have a check that tries to decode the whole state, or?

ggwpez:
Btw is this really `R0-silent`? I think we should add a prdoc to mention that the migration is broken otherwise.

bkchr:
> This should have been caught when we have a check that tries to decode the whole state, or?

Yes.

bkchr:
> Btw is this really `R0-silent`? I think we should add a prdoc to mention that the migration is broken otherwise.

What other label you want? This is a migration that is only required for the relay chains, aka no downstream chains are required to add this.

liamaharon:
> This should have been caught when we have a check that tries to decode the whole state, or?

Could you please link to where this check happens?

bkchr:
@liamaharon https://github.com/paritytech/polkadot-sdk/pull/1805

eskimor (approved):
Thanks @bkchr !

ggwpez:
Confirmed that #1805 detects this correctly with:  
```pre
failed to decode the value at key: Failed to decode value at key: 0x94eadf0156a8ad5156507773d0471e4ab8ebad86f546c7e0b135a4212aace339.
Storage info StorageInfo { pallet_name: Ok("ParaScheduler"), storage_name: Ok("AvailabilityCores"), prefix: Err(Utf8Error { valid_up_to: 0, error_len: Some(1) }), max_values: Some(1), max_size: None }.
Raw value: Some("0x0c010101010101")

Issue #1970, opened 2023-10-21
Fix consumer handling in pallet balances `update_locks`


bkchr:
When removing all locks of an account there is a bug that leads to reducing the consumer count twice.

First it happens here: https://github.com/paritytech/polkadot-sdk/blob/f0ab54e3e560b661d052ea362a995f3e68acda61/substrate/frame/balances/src/lib.rs#L969

And then the second time here: https://github.com/paritytech/polkadot-sdk/blob/f0ab54e3e560b661d052ea362a995f3e68acda61/substrate/frame/balances/src/lib.rs#L1072

liamaharon:
Self-assigned, I need to spend some time understanding this for writing docs anyway.

xlc:
we need a script to check if this bug have happened on any parachain and generate a list of accounts that needs fix up

xlc:
and this is something should be identified by a fuzzer so I guess we don’t have enough of them

xlc:
I can image it could be hard to write such script that works on all the parachains. That should spark another question: What can we improve to make such task easier in future? e.g. better event system or better indexer or better invariant checker

bkchr:
Writing a script for every chain should be possible. You just need to monitor `pallet_balances::Lock` and the changes to consumers. We would probably only require teams to set the starting block for when the runtime was activated that contained the broken logic.

michalisFr:
@bkchr Does this relate to this issue? https://github.com/paritytech/polkadot-sdk/issues/1404#issuecomment-1767580584

bkchr:
> @bkchr Does this relate to this issue? [#1404 (comment)](https://github.com/paritytech/polkadot-sdk/issues/1404#issuecomment-1767580584)

Could probably be related.

PR #1976, opened 2023-10-23
Stop `Balances` pallet erroneously double incrementing and decrementing consumers

liamaharon:
Closes https://github.com/paritytech/polkadot-sdk/issues/1970

Follow up issue to tackle, once the erroneous double incrementing/decrementing has stopped: https://github.com/paritytech/polkadot-sdk/issues/2037


kianenigma:
> Perhaps we could write a script that scrapes extrinsics for for accounts that had their consumers erroneously incremented and generates a migration to decrement them?

First glance, this will be a race that you're alway behind, as more accounts can get their consumers incremented by mistake in the timespan while we wait for the migration to execute.

bkchr:
We should first deploy the fix and then let the fix come in a later runtime upgrade.

substrate/frame/balances/src/lib.rs
```diff
@@ -1005,26 +1005,9 @@ pub mod pallet {
...
-					// No providers for the locks. This is impossible under normal circumstances
-					// since the funds that are under the lock will themselves be stored in the
-					// account and therefore will need a reference.
-					log::warn!(
-						target: LOG_TARGET,
-						"Warning: Attempt to introduce lock consumer reference, yet no providers. \
-						This is unexpected but should be safe."
-					);
-				}
+			match locks.is_empty() {
```
franciscoaguirre, line 1008: This could just be an `if`

substrate/frame/balances/src/tests/currency_tests.rs
```diff
@@ -144,7 +144,9 @@ fn lock_removal_should_work() {
 		.monied(true)
 		.build_and_execute_with(|| {
 			Balances::set_lock(ID_1, &1, u64::MAX, WithdrawReasons::all());
+			assert_eq!(System::consumers(&1), 1);
 			Balances::remove_lock(ID_1, &1);
+			assert_eq!(System::consumers(&1), 0);
```
franciscoaguirre, line 149: Is it possible for the balances pallet to increase consumers more than 1? We could add more tests in that case

franciscoaguirre (approved):

substrate/frame/balances/src/lib.rs
juangirini: It was an `if` indeed, I do like the readability of `match`

juangirini (approved):

kianenigma (approved):

substrate/frame/balances/src/tests/currency_tests.rs
liamaharon: I don't think it's possible.

ggwpez:
What do do with chains that got fucked by this before the fix applied?  
eg https://substrate.stackexchange.com/questions/11166

liamaharon:
> What to do with chains that got fucked by this before the fix applied? eg https://substrate.stackexchange.com/questions/11166

Need to retroactively fix it https://github.com/paritytech/polkadot-sdk/issues/2037

ggwpez:
Screenshot 2024-03-21 at 18 44 24

I think this fixed the `(F, T)` case but broke the `(T, T)` case.

PR #1985, opened 2023-10-23
Enable parallel key scraping

### LINKED ISSUE #174: remove-externalities: parallel key scraping
## Motivation

The current key scraping implementation is is quite inefficient, fetching 1k keys at a time, waiting for each batch of 1k keys to be returned before asking for the next one 

https://github.com/paritytech/substrate/blob/master/utils/frame/remote-externalities/src/lib.rs#L348-L392

This is not so bad when fetching keys from a local node where latency is not a big factor, but can slow things down very significantly when using a remote node when most of the time spend fetching keys becomes latency rather than the node doing work. 

Unfortunately, most nodes seem to have a hard limit imposed preventing more than 1k keys being fetched in a single request.

## Suggested solution

Rather than fetching all keys for the `prefix` in a single `rpc_get_keys_paged` request here https://github.com/paritytech/substrate/blob/master/utils/frame/remote-externalities/src/lib.rs#L528-L533, split the `prefix` up into N parts and run the `rpc_get_keys_paged` tasks in parallel using async/await. 

e.g. two parallel tasks:
start_keys: 0x000000 end_keys: 0x777777
start_keys: 0x777777 end_keys: 0xffffff

## Bonus (probably for a seperate issue/PR once this one is closed)

Fetch storage values for keys as they're pulled from the node, instead of waiting for all keys to be pulled before fetching values.

## Notes

This is not high priority, but worth doing once try-runtime-cli is in a more complete stable state to improve the experience pulling state from a remote node.

eagr:
closes #174


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4045960

liamaharon:
cargo fmt

liamaharon:
bot fmt

liamaharon (commented):
Thanks for this!

Some comments:

1. Could you please split up the logic of generating batches of key ranges to scrape and fetching them into seperate methods? This way, we can write a test to ensure generating the batches to fetch works as expected.

2. Please write some comments explaining what's happening to explain code that otherwise may not be immediately obvious.

3. Please replace existing serial scraping logic with your new method, so we can run the tests in this file with your new logic and also I'll do some manual testing.

eagr:
Sure, gladly!

eagr:
Got many 503s running the remote tests. I guess this may need more work.

eagr:
Good news is the tests run a lot faster with the changes.

liamaharon (commented):
I had a quick go creating a snapshot using `try-runtime-cli` pactched with this branch and using it, but got a bunch of errors indicating state was not scraped correctly.

Perhaps an easy way to debug could be to create a `remote-test` which scrapes the keys from a chain without much state like `wss://kusama-bridge-hub-rpc.polkadot.io:443` (so it's fast) using both your parallel method and the existing method, and compares the keys scraped using easy method. The keys scraped should be equivalent.

eagr:
I got the totally wrong idea of how a storage key is represented with `StorageKey`. Should've known the problem better before hacking it 🤣

eagr:
I haven't benched this. But this approach seems a lot slower than `rpc_get_keys_paged()` if the block is small (~4000).

liamaharon:
> I haven't benched this. But this approach seems a lot slower than `rpc_get_keys_paged()` if the block is small (~4000).

Yeah, seems something is wrong with it when I choose a parallelisation value >1.

eagr:
I'm not surpised at the result though, after seeing how many storage keys are there. The workload is always divided into 256 calls to `rpc_get_keys_paged()` when parallel>1. It may be slow because this is a dumb idea 🤣. But I couldn't think of a better way to do this, at least not without changing the underlying client methods. Any suggestions?

Out of curiosity, what is the average block size in terms of keys in reality?

eagr:
Is it possible to get the block size (or how many keys it contains) with the block hash before scraping? If so, maybe we could adjust the param dynamically based on the size?

liamaharon:
> I'm not surpised at the result though, after seeing how many storage keys are there. The workload is always divided into 256 calls to `rpc_get_keys_paged()` when parallel>1. It may be slow because this is a dumb idea 🤣. But I couldn't think of a better way to do this, at least not without changing the underlying client methods. Any suggestions?

Why is it always divided into 256 calls when parallel > 1? 

I don't really understand why your implementation must be so complex, I wonder if we can solve this problem with a simpler approach of creating a function which breaks up the keyspace in to N segments:
```rust
use num_bigint::{BigUint, ToBigUint};

/// Given the index of a segment and the total number of segments, compute an approximate
/// starting key for the segment. This assumes a 256-bit hash space.
fn get_starting_key_for_segment(segment_index: usize, total_segments: usize) -> StorageKey {
    // Define the total hash space for a 256-bit number.
    let total_hash_space: BigUint = (1.to_biguint().unwrap() << 256) - 1u32.to_biguint().unwrap();
    
    // Determine the size of each segment.
    let segment_size = total_hash_space / total_segments.to_biguint().unwrap();
    
    // Calculate the starting point of the given segment.
    let segment_start = segment_size * segment_index.to_biguint().unwrap();
    
    // Convert the starting point to a byte vector.
    let segment_start_bytes = segment_start.to_bytes_be();

    // Create a StorageKey with the byte vector.
    StorageKey(segment_start_bytes)
}
```

Then we can use this function to get the start and end of each segment we want to scrape?

WDYT?

eagr:
> I don't really understand why your implementation must be so complex

Coz I had the wrong assumption about how the scraping works, maybe also how data is organized in a block. Thanks for sticking with me here!

Quick question. Given a prefix `[42]`, could there be any matches if the start key is set to `[43, 0, ...]`? If not, maybe we could narrow down the hash space.

liamaharon:
> Coz I had the wrong assumption about how the scraping works, maybe also how data is organized in a block. Thanks for sticking with me here!

Sorry, didn't mean to sound harsh. really do appreciate your work here :) 

> Quick question. Given a prefix [42], could there be any matches if the start key is set to [43, 0, ...]?

No I don't think so, but probably worth checking to make sure.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4250961

eagr:
No worries man. I didn't feel anything harsh there. :))

I rushed out something with the same idea you suggested, slightly twisted. See if you like it? The code is totally rough, so no need to scrutinize now. I'll come back to clean up later. It feels much faster than the previous version.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4252749

eagr:
Cleaned up some unnecessary clones for the new functions. It feels like the right thing to do in general, but it's inconsistent with other functions which don't take references. Is it ok? I don't mind switching it back.

The `can_fetch_parallel` test is kinda ad hoc and `rpc_get_keys_paged()` is not used in lib code now. The new function `rpc_get_keys_in_range()` feels more flexible. I guess we should remove them and replace the test with better alternative?

I'd love to learn more about Substrate. But most docs I could find are for users. Is there some kind of knowledge wiki about the codebase, any pointers? Thanks!

liamaharon:
This is looking really good! 

I used this branch with [`try-runtime-cli`](https://github.com/paritytech/try-runtime-cli) to download Polkadot state on my crappy internet. 

Completely unscientific results shows a ~40% improvement with 4 parallel: 
- `paritytech/master`: 348s
- Parallel (4): 210s
- Parallel (8): 207s
- Parallel (128): 207s (for fun)

> It feels like the right thing to do in general, but it's inconsistent with other functions which don't take references. Is it ok? I don't mind switching it back.

It's fine to leave how you have it. You can improve the other functions if you like, but optional. 

> The can_fetch_parallel test is kinda ad hoc and rpc_get_keys_paged() is not used in lib code now. The new function rpc_get_keys_in_range() feels more flexible. I guess we should remove them and replace the test with better alternative?

Yup sounds great! 

> I'd like to bench the new functions, is there a test server with larger blocks I could use?

You can use Polkadot mainnet state, try `wss://polkadot-try-runtime-node.parity-chains.parity.io:443`

> I'd love to learn more about Substrate. But most docs I could find are for users. Is there some kind of knowledge wiki about the codebase, any pointers? Thanks!

We are working on something very much like that here: https://github.com/paritytech/polkadot-sdk/pull/2102 you can checkout the branch, run `RUSTDOCFLAGS='--html-in-header ./developer-hub/headers/toc.html' SKIP_WASM_BUILD=1 cargo doc --features try-runtime,experimental --workspace --no-deps --open`, then navigate to `{...}/target/doc/developer_hub/index.html`

liamaharon:
> Feels like the keys are not evenly distributed in those segments so some tasks do more lifting than others. Just realized that maybe we shouldn't divide the hash space solely based on the parallel parameter. I have an idea. Let me test it out.

Looking forward to seeing what you come up with :) 

> Nice, this is exactly what I'm looking for. Thanks! :))

If you have any suggestions feel free to comment there or open an issue, your perspective would be very valuable as someone getting started with Substrate.

eagr:
Here's the plan:
* Divide the workload based on the size of the suffix (the part stripped off the given prefix) instead of the parallel param, the bigger the size the more segments. This way, we should have a lower chance of having tasks doing too much work, which could be the bottleneck.
* Limit concurrent downloads to the given param with a semaphore.

Potential problems:
* This approach could be slower when scraping small blocks.

If it's possible to know the block size in advance, this could be addressed by adjusting the number of divided workloads accordingly. Is it?

> If you have any suggestions feel free to comment there or open an issue, your perspective would be very valuable as someone getting started with Substrate.

Sure!

eagr:
Seems ~50% faster than the previous version on my end. :))

liamaharon:
> Divide the workload based on the size of the suffix (the part stripped off the given prefix) instead of the parallel param, the bigger the size the more segments. This way, we should have a lower chance of having tasks doing too much work, which could be the bottleneck.
Limit concurrent downloads to the given param with a semaphore.

Hmm I don't really get this, could you please rephrase? 

> My numbers are much higher probably coz I'm behind vpn. With https://github.com/paritytech/polkadot-sdk/commit/4be72d83cafcaf3555f4b97fef8ccbea752cabe5, it takes ~800s, ~400s with https://github.com/paritytech/polkadot-sdk/commit/82507461a99885b9343c5a5e6a9f4baf494f0831, with parallel(4).

Strange, when I run locally I don't see any difference between performance running parallel(4) between these two commits.

eagr:
> Strange, when I run locally I don't see any difference between performance running parallel(4) between these two commits.

I'm totally not confident about those numbers coz my connection is not reliable (sometimes it took more than an hour to complete). If you don't see a difference, probably means there isn't, and my speculation is wrong or the code is bad.

> Completely unscientific results shows a ~40% improvement with 4 parallel

Theorectically, we should expect ~75% improvement right? But it's way off, so I suspected that maybe one of the tasks is doing much more work than the others maybe because there're more keys in that range.

I thought, what if we divide the whole hash space into smaller chunks? So, instead of 4 chunks on parallel(4), I divide it up to 2048 chunks when given an empty prefix `StorageKey(vec![])`, but the parallel downloads is still limited at 4 by a semaphore. The number of chunks is calculated based on the given prefix, the longer the prefix the lower the number if it makes sense.

Now that I think of it, a good hash function should probably produce hash values that are somewhat evenly distributed, right?

liamaharon:
> I thought, what if we divide the whole hash space into smaller chunks? So, instead of 4 chunks on parallel(4), I divide it up to 2048 chunks when given an empty prefix StorageKey(vec![]), but the parallel downloads is still limited at 4 by a semaphore. The number of chunks is calculated based on the given prefix, the longer the prefix the lower the number if it makes sense.

Ah I think I see what you're saying here, so we try to keep all 4 busy for as long as possible? This sounds sensible. We could try do that in this PR if you like, or get the 40% improvement in now and reach for this new approach in a seperate PR.

 > Now that I think of it, a good hash function should probably produce hash values that are somewhat evenly distributed, right?

It won't be nicely distributed because each pallet has it's own prefix :/ See https://polkadot-blockchain-academy.github.io/pba-book/substrate/storage/page.html#merklized-4

eagr:
> It won't be nicely distributed because each pallet has it's own prefix :/ See https://polkadot-blockchain-academy.github.io/pba-book/substrate/storage/page.html#merklized-4

I was thinking maybe I was chasing ghost. Then I'll give it another shot. Maybe I need to figure out how to do reliable bench first.

liamaharon:
> I was thinking maybe I was chasing ghost.

Wouldn't your approach still be more efficient for a non-evenly distributed tree?

eagr:
I have a very unstable connection, so I can't possibly do a reliable bench with actual network requests.

I've already had a snapshot saved locally. Is there an established way to spin up some lightweight test node with the snapshot, so I could scrape the local node instead? :))

liamaharon:
Hm no very easy way. Let me know which commands to run and I will post results.

eagr:
Could you help running the `can_fetch_parallel` test on your side and let me know the numbers? Sorry about the trouble. 😉 For this approach, I guess performance tuning boils down to finding the appropriate number of chunks we divide the workload into.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4387959

liamaharon:
```
Nov 20 06:45:54.040 ERROR frame_remote_externalities::remote_tests: p4c8: 90.95506214s
Nov 20 06:47:06.326 ERROR frame_remote_externalities::remote_tests: p4c10: 72.286042583s
Nov 20 06:48:16.853 ERROR frame_remote_externalities::remote_tests: p4c16: 70.526270244s
Nov 20 06:49:29.496 ERROR frame_remote_externalities::remote_tests: p4c20: 72.642682352s
Nov 20 06:50:39.128 ERROR frame_remote_externalities::remote_tests: p4c33: 69.632335329s
```

eagr:
Thanks! Are we seeing improvements here? What's the number of the approach on master?

liamaharon:
I added 

```rust
		// original
		let start = Instant::now();
		builder.rpc_get_keys_paged(prefix.clone(), at).await.unwrap();
		log::error!("rpc_get_keys_paged: {:?}", start.elapsed());
```

to measure baseline exiting performance

Screenshot 2023-11-20 at 12 52 13

`p4c8` this time is the same as the others, the earlier run where it higher was probably an anomaly. it seems that adjusting the cutoff has negligible impact on performance.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4399811

eagr:
Looks like the bottleneck is elsewhere. Then let's wrap this up for now, alright?

liamaharon:
> Then let's wrap this up for now, alright?

Sounds good. Tomorrow I'll re-benchmark your parallel implementation at commit 9881284 and if it's similar performance to HEAD shall we revert to 9881284 given the additional complexity in HEAD doesn't seem to provide tangible gains?

eagr:
I'd suggest that we do things the current approach, unless it performs worse.

I suppose how to divide the workload shouldn't be tied to `parallel` in the first place, not a perfect design. Not to mention if the hashes could be distributed unevenly. That could potentially be a bottleneck depending how uneven the distribution becomes.

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -400,41 +402,135 @@ where
...
+				|| self.get_keys_single_page(Some(prefix.clone()), last_key.cloned(), block);
+			let mut page = Retry::spawn(retry_strategy, get_page_closure).await?;
 
-			all_keys.extend(page);
+			// avoid duplicated keys across workloads
+			if let (Some(last), Some(end)) = (page.last(), end_key) {
+				if last >= end {
+					page.retain(|key| key < end);
+				}
+			}
```
liamaharon, line 514: Shouldn't we be able to avoid this through proper selection of start and end keys?

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -644,7 +741,7 @@ where
 			},
 		};
 		bar.finish_with_message("✅ Downloaded key values");
-		print!("\n");
+		println!();
```
liamaharon, line 744: You sure?

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -845,8 +942,9 @@ where
 		let mut keys_and_values = Vec::new();
 		for prefix in &config.hashed_prefixes {
 			let now = std::time::Instant::now();
+			// By observation, 4 parallel tasks almost always perform the best.
 			let additional_key_values =
-				self.rpc_get_pairs_paged(StorageKey(prefix.to_vec()), at, pending_ext).await?;
+				self.rpc_get_pairs(StorageKey(prefix.to_vec()), at, pending_ext, 4).await?;
```
liamaharon, line 947: Let's use the `PARALLEL_REQUESTS` const

substrate/utils/frame/remote-externalities/src/lib.rs
eagr: I don't think so, unless we change how `storage_keys_paged()` works? Right now, we could only tell it to stop with `DEFAULT_KEY_DOWNLOAD_PAGE`, which doesn't prevent it from returning duplicated keys.

substrate/utils/frame/remote-externalities/src/lib.rs
eagr: They work the same. I changed it only to reduce the amount of clippy warnings which were clustering the screen. :))

liamaharon (approved):
Looking good. Excited to get this merged and into `try-runtime-cli` :)

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -400,41 +402,135 @@ where
...
+				prefix.extend(vec![0; scale]);
+				return vec![StorageKey(prefix)]
+			}
+
+			// FIXME: figure out a better algo for dividing workload
+			let chunks = 16;
+			let step = 0x10000 / chunks;
+			let ext = scale - 2;
+
+			(0..chunks)
```
ggwpez, line 429: Instead of dividing them up into a fixed number, we could also have a generator<->consumer pattern, where one thread creates a stream of keys that can then be consumed by a variable number of reader threads.  
Then there is no risk that one threads takes much longer than the others because of large KV pairs in a specific region.  
But anyway, if its faster for now then lets go.

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -1440,4 +1536,26 @@ mod remote_tests {
 			.unwrap()
 			.execute_with(|| {});
 	}
+
+	#[tokio::test]
+	async fn can_fetch_in_parallel() {
+		init_logger();
+
+		let uri = String::from("wss://kusama-bridge-hub-rpc.polkadot.io:443");
```
ggwpez, line 1544: We run this CI check quite often (on each push of every open merge request). I dont want to DoS our own infrastructure with it.  
Would it be possible to spawn a local node for this? Or for now we can maybe feature gate it and not run in the CI?

liamaharon: We run `remote-test`s in CI?

ggwpez: Oh right, that works!

ggwpez (approved):

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -400,41 +402,135 @@ where
...
+			let mut prefix = prefix.as_ref().to_vec();
+			let scale = 32usize.saturating_sub(prefix.len());
+
+			// no need to divide workload
+			if scale < 9 {
+				prefix.extend(vec![0; scale]);
+				return vec![StorageKey(prefix)]
+			}
+
+			// FIXME: figure out a better algo for dividing workload
```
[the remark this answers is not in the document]
eagr: aye :)

substrate/utils/frame/remote-externalities/src/lib.rs
eagr: Sorry, I don't really follow, coz it seems to me there is just one role here. Or do you mean combining the processes of scraping keys and downloading the KVs? I could try it in a following PR. :))

ggwpez: Yea, Producer<->Consumer pattern is a standard approach for this, but i think we should not over-optimize here; its good now, thanks!

eagr: Right now discovering keys and downloading KVs are two separate steps. Given downloading KVs as the end goal, it does seem to me that we could cut more time by starting downloading after we have the first batch of discovered keys. Do you think it makes sense to combine the two steps? @liamaharon

The downside would be it's hard to show downloading progress, as we don't know how many keys are there until the discovery process finishes. Or is it possible to get the total amount of keys somewhere? 🤔

liamaharon: @eagr yes what you described would be great! I don't think there's an api to get the total number of keys unfortunately.

substrate/utils/frame/remote-externalities/src/lib.rs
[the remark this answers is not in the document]
eagr: Opened a follow up issue for this. Guess we could could punt on the comment, as this function will need to change anyway - yielding keys early.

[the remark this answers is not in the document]
liamaharon: Let's link to the followup instead of writing FIXME here

Issue #408, opened 2023-07-08
[NPoS] Staking: Users can delegate their fund to a target which can nominate on their behalf.


Ank4n:
This is first part of enabling funds of nomination pool delegators to be reusable for governance.

A delegator account should be able to delegate their funds to a delegatee account. The delegatee account can nominate validators using the funds of the delegator account. The funds of delegator should be locked in-place in delegator's account. Staking pallet will need to build a delegation ledger that will bookkeep the delegations. As first iteration, we should allow delegations only for direct nominations . 

The slashes should be applied lazily. As long as more funds are staked than pending slashes, staking pallet should be happy. But it should not allow more funds to be withdrawn than pending slashes. Effectively, the balance of a delegated nominator is `delegated_balance - pending_slashes`. This value should always be zero or greater.

Nomination pool will use the above primitive to allow users to delegate their funds to the pool. The pool account would be the direct nominator. The pool is also responsible for managing rewards and slashes if applicable. 

In future, the same primitives can be used to build different implementations/strategies around slashing and rewards for a pool. 

Some more details on the approach
https://hackmd.io/JXvBdTcHTZmwpEZzNst8mA?edit

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/opengov-improvements-on-the-technical-level/6578/3

Ank4n:
complete with https://github.com/paritytech/polkadot-sdk/pull/3905

Issue #454, opened 2022-08-17
[NPoS] Nomination pools: Allow funds to be used for democracy


kianenigma:
Edited:

Nomination pool currently works by 1) moving the funds from delegator to pool account, and 2) stake from pool account.
What we really want though is, 1) delegate the funds from delegator to pool account, and 2) stake from pool account.

Delegate the funds would imply similar pool functions with a very important difference that instead of funds moving from delegator to pool account, the funds will be locked in delegator's account and can be used for participating in governance voting.

In https://github.com/paritytech/polkadot-sdk/issues/408, staking pallet will enable the functionality of delegation of funds to another account (such as pool account) which will enable nomination pools to achieve the `delegate and stake` strategy. By utilising the delegated funds as if they belong to the pool, the pool account gains the ability to stake these funds as a direct nominator.

forgetso:
Thanks for the update @kianenigma

michalisFr:
Since this is already being worked on, this is mostly for the sake of discussion, but let me post an alternative solution.

With the current way that pools work (funds being sent to the pool), pool members could opt in to delegate their funds in the pool to the pool (to keep things simple delegating is binary, either their whole stake or nothing). A new `capital` variable for the pool would need to be introduced that would keep track of the total delegated balance (from the members that have opted in). Then the pool (through its owner account) would be able to vote in referenda with this capital. The conviction should be limited to None, otherwise pool members may not be able to unbond.

Basically this wouldn't be a delegation per se, but an internal mechanism of pools that allows them to have a capital (not necessarily equal to the pool account's balance) with which the pool can vote in referenda.

Is what I'm describing possible? What would be the benefits and drawback compared to the solution you're working on @Ank4n?

Some pros of the solution you work on that I can think of:
1. Funds remain in the user's account -> increased "sense" of security and similar experience to solo staking
2. User can decide for themselves how to vote or delegate their vote
3. Doesn't centralise voting power in the hands of pool operators (as my proposal does). But given that delegation is encouraged, I consider this more of a con than a pro (see my second point below).

Some pros of my proposed solution:
1. Simpler implementation (I think)
2. Users that join a pool some times do it as a simpler solution to staking, even if they can stake solo. These users are not likely to participate in governance. On the other hand, pool operators are likely to vote in referenda and if they carry the votes of their pool's members, this would lead to increased participation. In other words, it would be a simple way for pool members to delegate their votes, even simpler than the current delegation method.

[...]

PR #1986, opened 2023-10-23
[NPoS] Trusted delegation feature in staking

Ank4n:
closes https://github.com/paritytech/polkadot-sdk/issues/408.
pre-requisite for https://github.com/paritytech/polkadot-sdk/issues/454.

This PR is **part 1** of the goal to have nomination pool funds useable for governance. 

Note: This is an early draft to gather feedback on the approach. Few things such as [lazy slash mechanism for delegatee](https://hackmd.io/@ak0n/delegated-staking-slash) account still needs to be implemented and in general needs lot more tests and refactors.

## Summary

Adds a way for an account (`delegator`) to delegate their funds (held in their own accounts) to another ghost account (`delegatee`) that will act as a virtual staker. In first version, these will not be exposed via Call and so the delegation primitives are only exposed via `DelegatedStakingInterface` to trusted runtime code. Nomination Pool will be using these to migrate to a delegated staking. In future, we could expose the delegation functions via extrinsics paving way for external actors such as an offchain entity, a smart contract or a parachain (xcm) to build their own custom implementation of Nomination Pool without keeping custody of user funds.

I call `delegatee` a ghost account as it does not need funds of its own and their staking balance is only made up of its child delegators. Essentially, it is just a key to bond to the ledger and it is convenient to have it as `AccountID` type to minimise refactor and keep ledgers of both direct stakers and delegated stakers in a compatible manner. I tried the approach where a delegatee account can stake with both their free balance and delegated balance but it is much cleaner code if we allow either full delegation or none.

More details about the implementation could be found [here](https://hackmd.io/@ak0n/delegated-stake) (little dated).

This also aims to achieve a permissionless lazy migration of pool account funds to its delegators as described [here](https://hackmd.io/@ak0n/np-delegated-staking-migration).


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4049496

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4049528

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4050244

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4055584

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4055566

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4086345

polkadot/runtime/westend/src/lib.rs
```diff
@@ -296,7 +296,7 @@ impl pallet_balances::Config for Runtime {
 	type RuntimeFreezeReason = RuntimeFreezeReason;
 	type FreezeIdentifier = RuntimeFreezeReason;
 	type MaxFreezes = ConstU32<1>;
-	type MaxHolds = ConstU32<1>;
+	type MaxHolds = ConstU32<3>;
```
kianenigma, line 299: why bump by two?

substrate/frame/staking/src/delegation.rs
```diff
@@ -0,0 +1,226 @@
...
+//! all delegations to it are removed.
+//!
+//! Delegator: Someone who delegates their funds to a delegatee. A delegator can delegate their
+//! funds to one and only one delegatee. They also can not be a nominator or validator.
+//!
+//! Reward payouts are always made to another account set by delegatee. This account is a separate
+//! account from delegatee and rewards cannot be restaked automatically. The reward payouts can then
+//! be distributed to delegators by the delegatee via custom strategies.
+//!
+//! Any slashes to a delegatee (which is equivalent to nominator as long as StakingLedger is
```
kianenigma, line 35: This sounds very much like a pallet that similar to `fast-unstake` can be built externally. Something to consider.

polkadot/runtime/westend/src/lib.rs
Ank4n: Was unintentional 🤦

substrate/frame/staking/src/delegation.rs
Ank4n: I was thinking about this as well to have a `pallet-delegate`. 
- A pallet-delegate that holds funds for an account for a given reason such as for Staking delegation. It maps which account the fund is delegated to.
- Pallet-delegate becomes Currency provider for Staking pallet. When free balance is called for a delegated account, it returns delegated balance else falls back to normal balance.
- When hold funds are called, for delegated funds it just adds some state that it cannot be undelegated until the second lock by the pallet is released.

Would be great to brainstorm this (and also devil's advocate this a bit to not over-engineer it)..

substrate/primitives/staking/src/delegation.rs
```diff
@@ -0,0 +1,108 @@
...
+		value: Self::Balance,
+	) -> DispatchResult;
+
+	/// Update bond whenever there is a new delegate funds that are not staked.
+	fn update_bond(delegatee: &Self::AccountId) -> DispatchResult;
+
+	/// Unbond some funds from a delegatee.
+	///
+	/// Similar to [`StakingInterface::unbond`].
+	fn unbond(delegatee: &Self::AccountId, value: Self::Balance) -> DispatchResult;
```
gpestana, line 60: Would this interface need to define another `fn unbound` or could the logic be abstracted so that it relies on the `StakingInterface::unbound`? In that case, the account ID of the delegator/stash is be the only thing that implicitly has the information of whether the account is a delegatee or a direct nomination stash and that should be handled seamsly, if possible.

I think the overall comment here is that having a trait (`DelegatedStakeInterface`) that depends on another (`StakingInterface`) with the same methods seems a bit wrong to me.

substrate/frame/staking/src/ledger.rs
```diff
@@ -171,7 +172,12 @@ impl<T: Config> StakingLedger<T> {
 			return Err(Error::<T>::NotStash)
 		}
 
-		T::Currency::set_lock(STAKING_ID, &self.stash, self.total, WithdrawReasons::all());
+		if !<Delegatees<T>>::contains_key(&self.stash) {
+			// if the stash is a delegatee, funds are only delegated and already locked/held in the
```
gpestana, line 176: How about we lock the delegatee funds here too? That's one of the goals of having the `StakingLedger` struct wrapper., to consolidate all the ledger and lock logic under the same struct.

gpestana: Or maybe my comment doesn't make much sense in this context. The way that funds in the ledger of delegatees are accounted for is different from other ledgers, since they are accounted based on the locked and delegated stake from delegators. In that case, we should just skit the set lock here. let's discuss about this.

substrate/primitives/staking/src/delegation.rs
Ank4n: Yes I think we could remove this dependency. I started with this thinking DelegatedStakeInterface will be used by NP but now I think we can have a StakingDelegationInterface to add or remove delegation, and then use the normal StakeInterface for NP.

substrate/frame/staking/src/ledger.rs
Ank4n: Yeah I was thinking about how to abstract this. We could have a LockProvider trait and StakingLedger generic over this. The LockProviders would be either currency or staking delegate pallet. I will work on making this cleaner.

kianenigma:
would be good to close this to avoid confusion?

PR #1990, opened 2023-10-23
Always schedule at least one job onto a core

bkchr:
Even if the host configuration is returning `0` for the `lookahead`, we should schedule at least one job on a core if the core exists.


alindima (approved):

sandreim (approved):

eskimor:
Ideally the host configuration should be made sure to never have this set to 0. But that's better than nothing.

PR #1992, opened 2023-10-23
Add docs on encoding extrinsics

jsdw:
This will be merged to `kiz-umbrella-crates`, not `master`, and adds a reference doc page on constructing extrinsics.

I'm just opening this PR to allow for easy feedback; probably good to merge as soon as happy!

### DIFF
--- a/developer-hub/src/reference_docs/extrinsic_encoding.rs
+++ b/developer-hub/src/reference_docs/extrinsic_encoding.rs
@@ -1,0 +1,156 @@
+//! # Constructing and Signing Extrinsics
+//!
+//! Substrate is configurable enough that extrinsics can take any format, in theory. In practice,
+//! runtimes tend to use our [`sp_runtime::generic::UncheckedExtrinsic`] type to represent extrinsics.
+//! In Polkadot, this is configured
+//! [here](https://github.com/polkadot-fellows/runtimes/blob/94b2798b69ba6779764e20a50f056e48db78ebef/relay/polkadot/src/lib.rs#L1478)
+//! at the time of writing.
+//!
+//! What follows is a description of how extrinsics based on this type are encoded into bytes. These
+//! bytes can then be submitted to a chain for inclusion in a block (this is how we make changes to
+//! the state of a chain), and are ultimately stored in the body of a block.
+//!
+//! # Encoding an Extrinsic
+//!
+//! At a high level, all extrinsics are formed from concatenating some details together, ie:
+//!
+//! ```text
+//! extrinsic_bytes = concat(
+//!     compact_encoded_length,
+//!     version_and_maybe_signature,
+//!     call_data
+//! )
+//! ```
+//!
+//! Let's look at how each of these details is constructed.
+//!
+//! ## compact_encoded_length
+//!
+//! This is a SCALE compact encoded integer which is equal to the length, in bytes, of the rest of
+//! the extrinsic details.
+//!
+//! To obtain this value, we must encode and concatenate together the rest of the extrinsic details
+//! first, and then obtain the byte length of these. We can then compact encode that length, and
+//! prepend it to the rest of the details.
+//!
+//! ## version_and_maybe_signature
+//!
+//! If the extrinsic is _unsigned_, then `version_and_maybe_signature` will be just one byte
+//! denoting the _transaction protocol version_, which is 4.
+//!
+//! If the extrinsic is _signed_ (all extrinsics submitted from users must be signed), then
+//! `version_and_maybe_signature` is obtained by concatenating some details together, ie:
+//!
+//! ```text
+//! version_and_maybe_signature = concat(
+//!     version_and_signed,
+//!     from_address,
+//!     signature,
+//!     signed_extensions_extra,
+//! )
+//! ```
+//!
+//! Each of the details to be concatenated together is explained below:
+//!
+//! ### version_and_signed
+//!
+//! This is one byte, equal to `0x84` or `0b1000_0100` (ie an upper 1 bit to denote that it is
+//! signed, and then the transaction version, 4, in the lower bits).
+//!
+//! ### from_address
+//!
+//! This is the SCALE encoded address of the sender of the extrinsic. This address must correspond
+//! to the account ID (public key) whose private key will be used to sign the extrinsic. The exact
+//! type of the address can vary across different chains.
+//!
+//! The address type used on the Polkadot relay chain is [`sp_runtime::MultiAddress<AccountId32>`],
+//! where `AccountId32` is defined [here][`sp_core::crypto::AccountId32`]. When constructing a
+//! signed extrinsic to be submitted to a node, you'll always use the [`sp_runtime::MultiAddress::Id`]
+//! variant.
+//!
+//! ### signature
+//!
+//! This is the SCALE encoded signature. The signature is obtained by signing the _signed payload_
+//! bytes (see below on how this is constructed) using the private key associated with the address,
+//! and a signing algorithm that can vary across different chains.
+//!
+//! The signature type used on the Polkadot relay chain is [`sp_runtime::MultiSignature`]; the
+//! variants there are the types of signature that can be provided.
+//!
+//! ### signed_extensions_extra
+//!
+//! This is the concatenation of the SCALE encoded bytes representing each of the
+//! [_signed extensions_][sp_runtime::traits::SignedExtension] that the chain is configured with,
+//! in the order that they are configured for that chain. Signed extensions are, briefly, a means
+//! for different chains to extend the "basic" extrinsic format with custom data that can be checked
+//! by the runtime.
+//!
+//! When it comes to constructing an extrinsic, each signed extension has two things that we are
+//! interested in here:
+//!
+//! - The actual SCALE encoding of the signed extension type itself; this is what will form our
+//!   `signed_extensions_extra` bytes.
+//! - An `AdditionalSigned` type. This is SCALE encoded into the `signed_extensions_additional`
+//!   data of the _signed payload_ (see below).
+//!
+//! Either (or both) of these can encode to zero bytes.
+//!
+//! Each chain configures the set of signed extensions that it uses in its runtime configuration.
+//! At the time of writing, Polkadot configures them
+//! [here](https://github.com/polkadot-fellows/runtimes/blob/1dc04eb954eadf8aadb5d83990b89662dbb5a074/relay/polkadot/src/lib.rs#L1432C25-L1432C25).
+//! Some of the common signed extensions are defined [here][frame::deps::frame_system].
+//!
+//! Information about exactly which signed extensions are present on a chain and in what order is
+//! also a part of the metadata for the chain. For V15 metadata, it can be
+//! [found here][frame::deps::frame_support::__private::metadata::v15::ExtrinsicMetadata].
+//!
+//! ## call_data
+//!
+//! This data defines exactly which call is made by the extrinsic, and with what arguments, ie:
+//!
+//! ```text
+//! call_data = concat(
+//!     pallet_index,
+//!     call_index,
+//!     call_args
+//! )
+//! ```
+//!
+//! - `pallet_index` is a single byte denoting the index of the pallet that we are calling into.
+//! - `call_index` is a single byte denoting the index of the call that we are making the pallet.
+//! - `call_args` are the SCALE encoded bytes for each of the arguments that the call expects.
+//!
+//! Information about the pallets that exist for a chain (including their indexes), the calls
+//! available in each pallet (including their indexes), and the arguments required for each call
+//! can be found in the metadata for the chain. For V15 metadata, this information
+//! [is here][frame::deps::frame_support::__private::metadata::v15::PalletMetadata].
+//!
+//! # The Signed Payload Format
+//!
+//! All extrinsics submitted to a node from the outside world (also known as _transactions_) need to
+//! be _signed_. The data that needs to be signed for some extrinsic is called the _signed payload_
+//! (also called the _signed payload_), and its shape is described by the following pseudo-code:
+//!
+//! ```text
+//! signed_payload = concat(
+//!     call_data,
+//!     signed_extensions_extra,
+//!     signed_extensions_additional,
+//! )
+//!
+//! if length(signed_payload) > 256 {
+//!     signed_payload = blake2_256(signed_payload)
+//! }
+//! ```
+//!
+//! In other words, we create the signed payload by concatenating the bytes representing
+//! `call_data`, `signed_extensions_extra` and `signed_extensions_additional` together. If this
+//! payload is more than 256 bytes in size, we hash it using a 256bit Blake2 hasher.
+//!
+//! How to construct the `call_data` and `signed_extensions_extra` has already been explained above.
+//! `signed_extensions_additional` is constructed by SCALE encoding the
+//! ["additional signed" data][sp_runtime::traits::SignedExtension::AdditionalSigned] for each signed
+//! extension that the chain is using, in order, and concatenating the resulting bytes together.
+//!
+//! If the resulting bytes have a length greater than 256, then we hash them using a Blake2 256bit
+//! hasher, and that hash is the signed payload.
--- a/developer-hub/src/reference_docs/extrinsic_encoding.rs
+++ b/developer-hub/src/reference_docs/extrinsic_encoding.rs
@@ -46,7 +46,7 @@
 //! ## version_and_maybe_signature
 //!
 //! If the extrinsic is _unsigned_, then `version_and_maybe_signature` will be just one byte
-//! denoting the _transaction protocol version_, which is 4.
+//! denoting the _transaction protocol version_, which is 4 (or `0b0000_0100`).
 //!
 //! If the extrinsic is _signed_ (all extrinsics submitted from users must be signed), then
 //! `version_and_maybe_signature` is obtained by concatenating some details together, ie:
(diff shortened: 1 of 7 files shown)


developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
...
+//! ```
+//!
+//! - `pallet_index` is a single byte denoting the index of the pallet that we are calling into.
+//! - `call_index` is a single byte denoting the index of the call that we are making the pallet.
+//! - `call_args` are the SCALE encoded bytes for each of the arguments that the call expects.
+//!
+//! Information about the pallets that exist for a chain (including their indexes), the calls
+//! available in each pallet (including their indexes), and the arguments required for each call
+//! can be found in the metadata for the chain. For V15 metadata, this information
+//! [is here][frame::deps::frame_support::__private::metadata::v15::PalletMetadata].
```
jsdw, line 126: In acouple of places I wanted to point to `frame-metadata`. Better to add as a direct dependency of this developer-hub crate or perhaps expose more easily via frame-support?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4049319

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
+//! # Constructing and Signing Extrinsics
+//!
+//! Substrate is configurable enough that extrinsics can take any format, in theory. In practice,
+//! runtimes tend to use our [`sp_runtime::generic::UncheckedExtrinsic`] type to represent extrinsics.
+//! In Polkadot, this is configured
+//! [here](https://github.com/polkadot-fellows/runtimes/blob/94b2798b69ba6779764e20a50f056e48db78ebef/relay/polkadot/src/lib.rs#L1478)
+//! at the time of writing.
+//!
+//! What follows is a description of how extrinsics based on this type are encoded into bytes. These
+//! bytes can then be submitted to a chain for inclusion in a block (this is how we make changes to
```
kianenigma, line 10: You can link to the STF ref-doc and say that extrinsics are fed to the STF (as a part of the full block) and thus make changes to the state as a result.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
...
+//! the state of a chain), and are ultimately stored in the body of a block.
+//!
+//! # Encoding an Extrinsic
+//!
+//! At a high level, all extrinsics are formed from concatenating some details together, ie:
+//!
+//! ```text
+//! extrinsic_bytes = concat(
```
kianenigma, line 18: I think this is not arbitrary though, I would show an actual `Encode` impl block here.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
...
+//!     version_and_maybe_signature,
+//!     call_data
+//! )
+//! ```
+//!
+//! Let's look at how each of these details is constructed.
+//!
+//! ## compact_encoded_length
+//!
+//! This is a SCALE compact encoded integer which is equal to the length, in bytes, of the rest of
```
kianenigma, line 29: Would be good to back this up by scale docs saying what is a company encoding etc.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
...
+//!     signature,
+//!     signed_extensions_extra,
+//! )
+//! ```
+//!
+//! Each of the details to be concatenated together is explained below:
+//!
+//! ### version_and_signed
+//!
+//! This is one byte, equal to `0x84` or `0b1000_0100` (ie an upper 1 bit to denote that it is
```
kianenigma, line 57: again, showing this in the `Ecode`/`Decode` impl might be useful.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
...
+//! signed, and then the transaction version, 4, in the lower bits).
+//!
+//! ### from_address
+//!
+//! This is the SCALE encoded address of the sender of the extrinsic. This address must correspond
+//! to the account ID (public key) whose private key will be used to sign the extrinsic. The exact
+//! type of the address can vary across different chains.
```
kianenigma, line 64: Would be good to exactly say that this address is the nth generic of the `UncheckedExtrinsic`.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,156 @@
+//! # Constructing and Signing Extrinsics
+//!
+//! Substrate is configurable enough that extrinsics can take any format, in theory. In practice,
+//! runtimes tend to use our [`sp_runtime::generic::UncheckedExtrinsic`] type to represent extrinsics.
+//! In Polkadot, this is configured
+//! [here](https://github.com/polkadot-fellows/runtimes/blob/94b2798b69ba6779764e20a50f056e48db78ebef/relay/polkadot/src/lib.rs#L1478)
```
kianenigma, line 6: All in all you should give a disclaimer before going forward everything that follows in this page is only valid if you stick to this type. If you craft your own extrinsic type, things might differ. 

But we should also denote that `UncheckedExtrinsic` is generic AF so that hopefully everyone can be happy while using it, so that individual teams rarely need to touch these super sensitive parts of the code.

kianenigma (commented):
Great start, and thanks for contributing! I have some suggestions on how to improve.

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: I didn't do this initially because I felt the the Encode impl is a little over-complicated unless you understand what's going on, ie it's:

```rust
fn encode(&self) -> Vec<u8> {
	let mut tmp = Vec::with_capacity(sp_std::mem::size_of::<Self>());

	// 1 byte version id.
	match self.signature.as_ref() {
		Some(s) => {
			tmp.push(EXTRINSIC_FORMAT_VERSION | 0b1000_0000);
			s.encode_to(&mut tmp);
		},
		None => {
			tmp.push(EXTRINSIC_FORMAT_VERSION & 0b0111_1111);
		},
	}
	self.function.encode_to(&mut tmp);

	let compact_len = codec::Compact::<u32>(tmp.len() as u32);

	// Allocate the output buffer with the correct length
	let mut output = Vec::with_capacity(compact_len.size_hint() + tmp.len());

	compact_len.encode_to(&mut output);
	output.extend(tmp);

	output
}
```

Perhaps it's worth including that snippet of code here, but if so, I'd be tempted to have it in addition to the pseudocode, which I also use to name the parts of each of the sections below. Alternately I could make this pseudocode more rusty!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: This doesn't really have it's own Encode impl; it's just the line `tmp.push(EXTRINSIC_FORMAT_VERSION | 0b1000_0000); s.encode_to(&mut tmp);` lines in the `UncheckedExtrinsic::encode` call

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4060855

jsdw:
I addressed some of your feedback, and namely:

- Added links to the SCALE codec crate in some places (that crate doesn't currently actually describe the SCALE encoding nicely, but I think it should (at which point the links could maybe be improved, too).
- Linked to your STF reference docs, and referenced the generic param of UncheckedExtrinsic responsible for some things (as an "end user" I guess this wasn't relevant to me, but it does feel good to connect the docs to that type more concretely).

I'm not super keen on embedding the actual Encode impl of UncheckedExtrinsic _instead_ of the pseudo-code offhand, because the names I introduce in the pseudocode also are section headers. But I do think it'd be nice to show it after the first pseudocode example to make the rest of the doc more "concrete". For now, I linked to the code used; I couldn't see a way to embed the code via docify offhand, but maybe you have an idea here?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4061249

developer-hub/src/reference_docs/extrinsic_encoding.rs
kianenigma: I kinda agree that the full impl is too long, and leave it up you to decide. 

an impl block can indeed be referenced in docify, see an example here: https://docs.rs/docify/latest/docify/attr.export.html

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//! ```text
+//! extrinsic_bytes = concat(
+//!     compact_encoded_length,
+//!     version_and_maybe_signature,
+//!     call_data
+//! )
+//! ```
+//!
+//! For clarity, the actual implementation
+//! [is here](https://github.com/paritytech/polkadot-sdk/blob/35eb133baab93d3f2f179df216b2cc175d7dcaf2/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L296)
```
kianenigma, line 32: Since we are in the mono-repo, I would 100% avoid using permalinks with commit hash. We can probably create a link to the `impl` header in rust-docs, no?

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//! The signature type used on the Polkadot relay chain is [`sp_runtime::MultiSignature`]; the
+//! variants there are the types of signature that can be provided.
+//!
+//! ### signed_extensions_extra
+//!
+//! This is the concatenation of the [SCALE encoded][frame::deps::codec] bytes representing each of
+//! the [_signed extensions_][sp_runtime::traits::SignedExtension], and are configured by the
+//! fourth generic parameter of [`sp_runtime::generic::UncheckedExtrinsic`]. Signed extensions are,
+//! briefly, a means for different chains to extend the "basic" extrinsic format with custom data
+//! that can be checked by the runtime.
```
kianenigma, line 99: I think it is safe for you to already create a new empty ref-doc page for signed extensions, then link to it here. This topic will 100% need its own ref doc.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//!
+//! When it comes to constructing an extrinsic, each signed extension has two things that we are
+//! interested in here:
+//!
+//! - The actual SCALE encoding of the signed extension type itself; this is what will form our
+//!   `signed_extensions_extra` bytes.
+//! - An `AdditionalSigned` type. This is SCALE encoded into the `signed_extensions_additional`
+//!   data of the _signed payload_ (see below).
```
kianenigma, line 107: I would put an example of this here; A dummy signed extension that has some u32 in it plus some `additional_signed`.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//!
+//! Either (or both) of these can encode to zero bytes.
+//!
+//! Each chain configures the set of signed extensions that it uses in its runtime configuration.
+//! At the time of writing, Polkadot configures them
+//! [here](https://github.com/polkadot-fellows/runtimes/blob/1dc04eb954eadf8aadb5d83990b89662dbb5a074/relay/polkadot/src/lib.rs#L1432C25-L1432C25).
```
kianenigma, line 113: I think it is good that you are using Polkadot as reference, but it would be good to mention that generally you can obtain all of this info either by reading the code in a static way, or from the metadata in a dynamic way. 

btw, do you want to also make a placeholder for a metadata ref-doc and do it? :D

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//! Some of the common signed extensions are defined [here][frame::deps::frame_system].
```
kianenigma, line 114:
```suggestion
//! Some of the common signed extensions are defined [here][frame::deps::frame_system#signed-extensions].
```

developer-hub/src/reference_docs/extrinsic_encoding.rs
kianenigma: I would leave it as-is for now to not distract us, but I expect the `frame` crate to properly expose all relevant types long term to avoid needing to go down these hidden paths.

developer-hub/src/reference_docs/extrinsic_encoding.rs
```diff
@@ -0,0 +1,172 @@
...
+//! `call_data`, `signed_extensions_extra` and `signed_extensions_additional` together. If this
+//! payload is more than 256 bytes in size, we hash it using a 256bit Blake2 hasher.
+//!
+//! How to construct the `call_data` and `signed_extensions_extra` has already been explained above.
+//! `signed_extensions_additional` is constructed by SCALE encoding the
+//! ["additional signed" data][sp_runtime::traits::SignedExtension::AdditionalSigned] for each signed
+//! extension that the chain is using, in order, and concatenating the resulting bytes together.
+//!
+//! If the resulting bytes have a length greater than 256, then we hash them using a Blake2 256bit
+//! hasher, and that hash is the signed payload.
```
kianenigma, line 172: All I am kinda itching for is a more practical example, but you know better, and maybe it is an overkill. How about a snippet that creates a very basic instance of `UncheckedExtrinsic`, then `.encode()` it, then try and cluster the output bytes into different sections that you explained in the doc?

kianenigma (approved):
Some final suggestions, but happy to merge as-is and iterate later.

I kinda suggested you to add a similar one for metadata as well. Similarly, we can be rapid and iterate later. 

btw, try building it now with this command: 

```
RUSTDOCFLAGS="--html-in-header ./developer-hub/headers/toc.html" cargo doc -p developer-hub  --no-deps  --open
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4093128

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4093110

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: I actually tried this I think and it didn't seem to work, but maybe because I was dfoing something else too; I'll try again because this would be ideal!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: Ah yeah; these are sortof hangovers from when it was a gist, but also I'm didn't think you could link in a nice way to impls etc in rust docs (only to definitions of things). Lemme see if there's a nice way!

If that doesn't work, I used permalinks with commit hashes just to guarantee that the link would never "break" (it may fall out of date, which is why I write "at the itme of writing" with these, but at least it won't end up pointing to complete nonsense if the code changes underneath :))

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: I wonder whether metadata needs a ref-doc, or whether it should just be properly documented in the `frame-metadata` crate? Maybe both; might be worth starting with adding good docs to `frame-metadata` and then writing a ref doc if there turns out to be some more general documentation about metadata that would conceptually span several crates?

But yeah, I'm happy to write about this too!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: Yup I think you're right, and I do love examples myself! 

Maybe I can add an example at the end showing how to manually construct the bytes for some extrinsic, and then prove that they are correct (with a nice hidden assert to make sure it stays this way) when using `UncheckedExtrinsic` to do the same!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: Since we'll add a `signed_extensions` section, I'd suggest we leave a proper description+examples of them for that page, and keep the description of them here quite brief just to cover the essentials of encoding? Then in an example below I'll show how to manually encode a signed tx including some signed extensions or something like that :)

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: Worked fine this time; I must have done something else that broke it before!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: For now, I added it!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: I added the doc example instead for now!

developer-hub/src/reference_docs/extrinsic_encoding.rs
jsdw: I added a couple of v basic examples in the `signed_extensions` thing, really just because Iw anted them for an extrinsic example here. They can probably be improved!

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4106309

PR #1993, opened 2023-10-23
Not require all messages to be processed in simulator, for fuzzer

franciscoaguirre:


polkadot/xcm/xcm-simulator/src/lib.rs
```diff
@@ -145,7 +145,7 @@ macro_rules! decl_test_relay_chain {
 								pallet_message_queue::Event::Processed {origin, ..}) => {
 							assert_eq!(origin, &AggregateMessageOrigin::Ump(UmpQueueId::Para(para)));
 						},
-						event => panic!("Unexpected event: {:#?}", event),
+						event => println!("Message not processed: {:#?}", event),
```
ggwpez, line 148: What messages are not being processed?

franciscoaguirre: We allow messages to be constructed that are not really valid
This came up in the fuzzer, that would generate any sort of message and some would fail because of this
I don't know if we should impact the simulator because of this though

PR #1998, opened 2023-10-24
1953 defensive testing extrinsic

### LINKED ISSUE #1953: `defensive!` testing extrinsic
I think we should extend the `RootTesting` pallet with a `trigger_defensive` call that can be used to check that the runtime monitoring for our networks works end-to-end.  
This could then be deployed to Rococo/Westend and called by devops to test it.

adelarja:
# Description

The `trigger_defensive` call has been added to the `root-testing` pallet. The idea is to have this pallet running on `Rococo/Westend` and use it to verify if the runtime monitoring works end-to-end.

To accomplish this, `trigger_defensive` dispatches an event when it is called.

Closes #1953

# Checklist

  required)
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)

You can remove the "Checklist" section once all have been checked. Thank you for your contribution!

✄ -----------------------------------------------------------------------------


substrate/frame/root-testing/src/lib.rs
```diff
@@ -50,5 +59,13 @@ pub mod pallet {
 			ensure_root(origin)?;
 			Ok(())
 		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight(0)]
+		pub fn trigger_defensive(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+			Self::deposit_event(Event::DefensiveTestCall);
```
ggwpez, line 67: Oh, you also have to actually trigger the defensive logic via `frame_support::defensive!("Howdy");`. or something like this.

adelarja: @ggwpez I've added the `frame_support::defensive!` logic. What do you think about the message?

ggwpez:
Good so far, now please also deploy to Rococo/Westend (as hinted in the Issie https://github.com/paritytech/polkadot-sdk/issues/1953).

substrate/frame/root-testing/src/lib.rs
```diff
@@ -50,5 +59,14 @@ pub mod pallet {
 			ensure_root(origin)?;
 			Ok(())
 		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight(0)]
+		pub fn trigger_defensive(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+			frame_support::defensive!("root_testing::trigger_defensive was called.");
```
bkchr, line 67: @ggwpez so you want this to panic? I don't really get what you want to achieve?

xlc: defensive is just log error and debug_assert, which is nop on release build, so this is a bit pointless. if you want to test log error, just log error.

bkchr: Yeah I know that this is using a debug assert, I was assuming @ggwpez wants to build with debug asserts enabled and then use it as overwrite.

ggwpez: The idea here is that we are currently not sure if the monitoring correctly picks up defensive failures. And we really should be notified about them, so this is just as test.  
It could also be logged, but the `defensive!` macro also logs a magic string that we can filter for and alert in the grafana.

polkadot/runtime/rococo/Cargo.toml
```diff
@@ -67,7 +67,7 @@ pallet-scheduler = { path = "../../../substrate/frame/scheduler", default-featur
 pallet-session = { path = "../../../substrate/frame/session", default-features = false }
 pallet-society = { path = "../../../substrate/frame/society", default-features = false, features = ["experimental"] }
 pallet-sudo = { path = "../../../substrate/frame/sudo", default-features = false }
-frame-support = { path = "../../../substrate/frame/support", default-features = false }
+frame-support = { path = "../../../substrate/frame/support", default-features = false, features = ["tuples-96"] }
```
adelarja, line 70: I had to add this feature because the runtime pallets in `Rococo` exceeded 64. After adding this line, the tests started to pass.

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -1504,6 +1508,8 @@ construct_runtime! {
 
 		// Sudo.
 		Sudo: pallet_sudo::{Pallet, Call, Storage, Event<T>, Config<T>} = 255,
+
+		RootTesting: pallet_root_testing::{Pallet, Call, Storage, Event<T>} = 249,
```
adelarja, line 1512: I added random numbers here and in `Westend`'s runtime. Is it ok?

ggwpez: Please put in front of the Sudo, so that the order of the pallets and their indices line up.

adelarja: Do you mean something like this?

```rust
                // Root testing pallet.
		RootTesting: pallet_root_testing::{Pallet, Call, Storage, Event<T>} = 249,

		// Sudo.
		Sudo: pallet_sudo::{Pallet, Call, Storage, Event<T>, Config<T>} = 255,
```

ggwpez (approved):
LGTM!  
cc @PierreBesson I hope you are prepared :laughing:

polkadot/runtime/rococo/src/lib.rs
ggwpez: Yea 254 is already taken, guess its okay.

PierreBesson:
I would suggest to use a new special log level ("log target" in rust) called DEFENSIVE. This will make it easier to identify those log in monitoring systems as the log level can be easily indexed in systems such as Grafana Loki.

ggwpez:
> I would suggest to use a new special log level ("log target" in rust) called DEFENSIVE. This will make it easier to identify those log in monitoring systems as the log level can be easily indexed in systems such as Grafana Loki.

Oh yea, did not think about this. Currently we just use `runtime`, but we could switch it to `runtime::defensive`.  
Could you please also do this @adelarja?

bkchr (approved):

adelarja:
> > I would suggest to use a new special log level ("log target" in rust) called DEFENSIVE. This will make it easier to identify those log in monitoring systems as the log level can be easily indexed in systems such as Grafana Loki.
> 
> Oh yea, did not think about this. Currently we just use `runtime`, but we could switch it to `runtime::defensive`. Could you please also do this @adelarja?

Sure, I can do it. But I'm not entirely sure about what do we need to do.

@PierreBesson do you mean doing something like:

```rust
log::info!(target: "DEFENSIVE", "Example log!");
```

@ggwpez Where are we setting the target to `runtime`?

ggwpez:
> @ggwpez Where are we setting the target to `runtime`?

It needs to be changed in the macro: https://github.com/paritytech/polkadot-sdk/blob/a70617124b4c0c730016a210e2b8b343c829d784/substrate/frame/support/src/traits/misc.rs#L58

adelarja:
> > @ggwpez Where are we setting the target to `runtime`?
> 
> It needs to be changed in the macros to `runtime::defensive` please:
> 
> https://github.com/paritytech/polkadot-sdk/blob/a70617124b4c0c730016a210e2b8b343c829d784/substrate/frame/support/src/traits/misc.rs#L58

Ohh yes I saw that. I thought we needed to implement something more difficult haha. I've made the change.

ggwpez (approved):

ggwpez:
Looks like there is a conflict with master, could you please resolve the conflicts? Then we can merge.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4119530

adelarja:
> Looks like there is a conflict with master, could you please resolve the conflicts? Then we can merge.

Done. The rococo and westend tests are passing.

PierreBesson:
> @PierreBesson do you mean doing something like:
> 
> log::info!(target: "DEFENSIVE", "Example log!");

@adelarja, having logs printed as `runtime::defensive` as @ggwpez will do the job nicely. I see that you have already implemented this.

ggwpez:
Seeing this in the CI:  
```pre
error[E0046]: not all trait items implemented, missing: `RuntimeEvent`
   --> substrate/frame/utility/src/tests.rs:190:1
    |
 190 | impl pallet_root_testing::Config for Test {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `RuntimeEvent` in implementation
    |
    = help: implement the missing item: `type RuntimeEvent = /* Type */;`
```

Mostly all CI checks should be green. Once that is, we should be good to merge.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4141749

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4147574

PR #1999, opened 2023-10-24
Add test to demonstrate the failure scenario

rahulksnv:
The change adds a test to show the failure scenario that caused #1812 to be rolled back (more context: https://github.com/paritytech/polkadot-sdk/issues/493#issuecomment-1772009924)

Summary of the scenario:
1. Node has finished downloading up to block 1000 from the peers, from the canonical chain.
2. Peers are undergoing re-org around this time. One of the peers has switched to a non-canonical chain, announces block 1001 from that chain
3. Node downloads 1001 from the peer, and tries to import which would fail (as we don't have the parent block 1000 from the other chain)


dmitry-markin (commented):
Is it possible to modify the test so that it's testing not the failure scenario but the expected behavior? E.g., if we remove `[should_panic]` and the test passes, that means that we fixed the bug?

rahulksnv:
> Is it possible to modify the test so that it's testing not the failure scenario but the expected behavior? E.g., if we remove `[should_panic]` and the test passes, that means that we fixed the bug?

yes, if the issue is resolved, two changes would be needed: 
1. `[should_panic]`  removed 
2. Minor change to the last part would be needed. The last part should result in no requests being sent out, test should check for that instead

substrate/client/network/sync/src/lib.rs
```diff
@@ -2618,12 +2618,25 @@ mod test {
...
+			build_block_with_fork_data(client, at, None)
+		}
+	}
+
+	/// Build and import a new best block and the fork data.
+	fn build_block_with_fork_data(
+		client: &mut Arc<TestClient>,
+		at: Option<Hash>,
+		fork_data: Option<(Vec<u8>, Vec<u8>)>,
+	) -> Block {
```
dmitry-markin, line 2633: Why is `build_block()` with `fork: true` not enough? There are only two forks in the test at the same block number, so it should be enough to create one of them with `fork: false` and another with `fork: true`?

substrate/client/network/sync/src/lib.rs
```diff
@@ -3364,4 +3377,127 @@ mod test {
...
+			send_block_announce(block.header().clone(), peer, &mut sync);
+			let request = get_block_request(&mut sync, FromBlock::Hash(block.hash()), 1, &peer);
+			let response = create_block_response(vec![block.clone()]);
+			let res = sync.on_block_data(&peer, Some(request), response).unwrap();
+			assert!(matches!(
+				res,
+				OnBlockData::Import(ImportBlocksAction{ origin: _, blocks }) if blocks.len() == 1_usize
+			),);
+			assert!(sync.is_known(&block.header.parent_hash()));
+		}
```
dmitry-markin, line 3501: Could you add a comment to this part stating what changes are needed to convert this test into the desired behavior test? And also reference the issue that needs to be fixed before this is possible (and back-reference the test in the issue)?

dmitry-markin (approved):

substrate/client/network/sync/src/lib.rs
rahulksnv: yeah not needed, removed it

substrate/client/network/sync/src/lib.rs
rahulksnv: Done

rahulksnv:
@dmitry-markin  Addressed comments, please merge if looks good, thanks

substrate/client/network/sync/src/lib.rs
```diff
@@ -3364,4 +3364,129 @@ mod test {
 		pending_responses.remove(&peers[1]);
 		assert_eq!(pending_responses.len(), 0);
 	}
+
+	/// The test demonstrates the issue discussed here:
+	/// https://github.com/paritytech/polkadot-sdk/issues/493#issuecomment-1776648354.
```
dmitry-markin, line 3369: Could you create an issue for this and reference it here instead of the comment? Copy-pasting from the comment would be enough, may be adding some other observations you have already made.

dmitry-markin:
> @dmitry-markin Addressed comments, please merge if looks good, thanks

Looks good, but still need one more approving review to be able to merge.

altonen (approved):

substrate/client/network/sync/src/lib.rs
dmitry-markin, line 3369:
```suggestion
	/// The test demonstrates https://github.com/paritytech/polkadot-sdk/issues/2094.
	/// TODO: convert it into desired behavior test once the issue is fixed (see inline comments).
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4140199

dmitry-markin:
bot help

PR #2006, opened 2023-10-24
Simplify PayOverXcm

franciscoaguirre:
`PayOverXcm` had quite a lot of hacks that arose from how non-generic the salary pallet was.
This PR aims to simplify it to make things easier to reason about, as well as to allow easier customization.

The location to send the message was being codified into the asset, but the system broke if your beneficiary was not in the given destination, hence it's better to couple it in the `beneficiary` argument itself.
Now the beneficiary is a full fledged location, it is split and the last part is used as the account, while the first part is used as the destination to route the XCM program.

## Open question
The best would be to make pallet salary support `MultiLocation`s as ids for the members of the collective.
Would this break anything with the fellowship?
I think not since their accounts should be in AssetHub for salaries to work anyway.
We should then migrate the collective to using this type.

### DIFF
--- a/polkadot/runtime/rococo/src/lib.rs
+++ b/polkadot/runtime/rococo/src/lib.rs
@@ -31,9 +31,7 @@ use primitives::{
 };
 use runtime_common::{
 	assigned_slots, auctions, claims, crowdloan, impl_runtime_weights,
-	impls::{
-		LocatableAssetConverter, ToAuthor, VersionedLocatableAsset, VersionedMultiLocationConverter,
-	},
+	impls::ToAuthor,
 	paras_registrar, paras_sudo_wrapper, prod_or_fast, slots, BlockHashCount, BlockLength,
 	SlowAdjustingFeeUpdate,
 };
@@ -96,7 +94,7 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 use xcm::{
 	latest::{InteriorMultiLocation, Junction, Junction::PalletInstance},
-	VersionedMultiLocation,
+	VersionedMultiLocation, VersionedAssetId,
 };
 use xcm_builder::PayOverXcm;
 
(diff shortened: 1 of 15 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4063529

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4064673

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -428,18 +426,14 @@ impl pallet_treasury::Config for Runtime {
 	type WeightInfo = weights::pallet_treasury::WeightInfo<Runtime>;
 	type SpendFunds = Bounties;
 	type SpendOrigin = TreasurySpender;
-	type AssetKind = VersionedLocatableAsset;
+	type AssetKind = VersionedAssetId;
```
KiChjang, line 429: I don't think this is actually enough; `LocatableAssetId` had both `Location` and an `Asset`, the former is where the latter is located, e.g. I can either have DOT tokens on the relay chain or on the Asset Hub, or any other system parachain, and the `Location` parameter is used to distinguish where the `Asset` is located.

franciscoaguirre: Yes, `AssetId` doesn't hold the same information as `LocatableAssetId`, it's just a simplification because I moved the other piece of information, the location where the asset is held, elsewhere.

The `PayOverXcm` trait always required the account to be in the actual destination chain, given that it did a `TransferAsset` to that account on the destination chain.
Given that, the information of where to send the message should be encoded in the location of the beneficiary itself, which is what I did.

So before we had this:
- beneficiary -> specifies the recipient account on the destination chain, but doesn't specify the destination chain
- asset kind -> specifies the asset and the location where it's held, which is used as the destination chain
This means there's a tight coupling between beneficiary and asset kind because the location to send the message shouldn't live alongside the asset.

Now we have:
- beneficiary -> specifies the destination chain and the recipient account within it
- asset -> just specifies the asset to transfer

I agree you can either have DOT on the relay or asset hub, but the asset is still the same, DOT, you just need to specify where you send the message. Every receiver interprets the asset as they need to to come up with DOT.

bkchr:
Stale

Issue #619, opened 2023-06-02
PVF worker: restrict network access


mrcnski:
# ISSUE

## Overview

We're already working on sandboxing by [blocking all unneeded syscalls](https://github.com/paritytech/polkadot-sdk/issues/882). However, due to the wide scope it will take a while longer. I suggest starting with a much smaller scope and just blocking network access until the above is ready. 

Note that we already have [filesystem restriction](https://github.com/paritytech/polkadot/pull/7303) coming soon. With network access also blocked, we will have decent sandboxing in place until the larger work gets in.

## Proposal

[birdcage](https://github.com/phylum-dev/birdcage/blob/main/src/linux/seccomp.rs) simply blocks socket creation, maybe we can do the same. 

The main risk with this, as with our other `seccomp` work, is missing syscalls on some OS/architecture or new syscalls being introduced. Currently there are only `socket` and `socketpair`, and since we are now targeting Linux x86-64, this should be tractable. For example, we can set up a regular job that downloads the syscall table from the latest kernel version and makes sure no new syscalls containing "socket" were added. However, the risk of a new syscall appearing, and libc libraries using it, in the immediate-term is very low, so I think it shouldn't block this development.

mrcnski:
Still need to enable the restrictions after one release.

PR #1663, opened 2023-09-21
PVF worker: Maintain lists of used syscalls

mrcnski:
- Add list-syscalls.rb for getting the syscalls used by the workers.
  - From https://gist.github.com/koute/166f82bfee5e27324077891008fca6eb.
  - I needed a linter and rubocop was really noisy with its errors, so I had it automatically format the file.
  - I added a `--only-list-syscalls` option to get syscalls without extra noise. Warnings are still printed.
- Add syscall lists for the execute and prepare workers.
  - These will be used to generate the seccomp filters in the future.
- Add tools for building with musl
  - We have to build with musl for best results with list-syscalls.rb.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/1652
Closes https://github.com/paritytech/ci_cd/issues/880 (CI pipeline request)

Issue #882, opened 2022-01-14
PVF host: sandbox/harden worker process


pepyakin:
## Overview

As future work for the PVF validation host we envisioned that the workers should be hardened.

Ideally, if a worker process was compromised then the attacker won't get the rest of the system on a silver plate. To do that, we may want to consider:

1. running workers under a different user
2. jailing/unshare or clone(2) with fresh new namespace
3. **seccomp**

## Summary / Roadmap

https://hackmd.io/@gIXf9c2_SLijWKSGjgDJrg/BkNikRcf2

## Potentially Useful Resources

https://www.redhat.com/sysadmin/container-security-seccomp

mrcnski:
We have landlock, networking restrictions with seccomp, unshare/pivot_root, separate processes, and are in the process of [implementing clone with sandbox args](https://github.com/paritytech/polkadot-sdk/pull/2477). We've decided that this is enough security until the PolkaVM migration.

PR #2009, opened 2023-10-24
PVF worker: Add seccomp restrictions (restrict networking)

mrcnski:
## Overview

We're already working on sandboxing by [blocking all unneeded syscalls](https://github.com/paritytech/polkadot-sdk/issues/882). However, due to the wide scope it will take a while longer. This PR starts with a much smaller scope, only blocking network-related syscalls until the above is ready.

For security we block the following with `seccomp`:

1. creation of new sockets - these are unneeded in PVF jobs, and we can safely block them without affecting consensus.

2. `io_uring` - as discussed [here](https://github.com/paritytech/polkadot/pull/7334#discussion_r1281611134), io_uring allows for networking and needs to be blocked. See below for a discussion on the safety of doing this.

3. `connect`ing to sockets - the above two points are enough for networking and is what birdcage does (or [used](https://github.com/phylum-dev/birdcage/pull/37) to [do](https://github.com/phylum-dev/birdcage/pull/47)) to restrict networking. However, it is possible to [connect to abstract unix sockets](https://lore.kernel.org/landlock/20231023.ahphah4Wii4v@digikod.net/T/#u) to do some kinds of sandbox escapes, so we also block the `connect` syscall.

## Safety of blocking io_uring

(Intentionally left out of implementer's guide because it felt like too much detail.)

`io_uring` is just a way of issuing system calls in an async manner, and there is nothing stopping wasmtime from legitimately using it. Fortunately, at the moment it does not. Generally, not many applications use `io_uring` in production yet, because of the numerous kernel CVEs discovered. It's still under a lot of development. Android outright banned `io_uring` for these reasons.

Considering `io_uring`'s status, and that it very likely would get detected either by our [recently-added static analysis](https://github.com/paritytech/polkadot-sdk/pull/1663) or by testing, I think it is fairly safe to block it.

### Consensus analysis

If execution hits an edge case code path unique to a given machine, it's already taken a non-deterministic branch anyway. After all, we just care that the majority of validators reach the same result and preserve consensus. So worst-case scenario, there's a dispute, and we can always admit fault and refund the wrong validator. On the other hand, if all validators take the code path that results in a seccomp violation, then they would all vote against the current candidate, which is also fine. The violation would get logged (in big scary letters) and hopefully some validator reports it to us.

Actually, a worst-worse-case scenario is that 50% of validators vote against, so that there is no consensus. But so many things would have to go wrong for that to happen:

1. An update to `wasmtime` is introduced that uses io_uring (unlikely as io_uring is mainly for IO-heavy applications)
2. The new syscall is not detected by our static analysis
3. It is never triggered in any of our tests
4. It then gets triggered on some super edge case in production on 50% of validators causing a stall (bad but very unlikely)
5. Or, it triggers on only a few validators causing a dispute (more likely but not as bad?)

Considering how many things would have to go wrong here, we believe it's safe to block `io_uring`.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/619
Original PR in Polkadot repo: https://github.com/paritytech/polkadot/pull/7334
Birdcage io_uring PR: https://github.com/phylum-dev/birdcage/pull/34


eskimor (approved):
Nice!

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
```diff
@@ -0,0 +1,197 @@
...
+//!    stall (bad but very unlikely)
+//!
+//! 5. Or, it triggers on only a few validators causing a dispute (more likely but not as bad)
+//!
+//! Considering how many things would have to go wrong here, we believe it's safe to block
+//! `io_uring`.
+//!
+//! # Action on	caught syscall
+//!
+//!	TODO
```
mrcnski, line 75: Instead of killing the process on violations, we can have seccomp notify the parent, having it kill the process and log what happened. We anyway want logging to ensure that the syscall detection script is really sound, before enabling a seccomp whitelist based on it in production. I've done some experiments and performance does not suffer too much.

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
```diff
@@ -0,0 +1,197 @@
...
+//! - creation of new sockets - these are unneeded in PVF jobs, and we can safely block them without
+//!   affecting consensus.
+//!
+//! - `io_uring` - allows for networking and needs to be blocked. See below for a discussion on the
+//!   safety of doing this.
+//!
+//! - `connect`ing to sockets - the above two points are enough for networking. However, it is
+//!   possible to [connect to abstract unix
+//!   sockets](https://lore.kernel.org/landlock/20231023.ahphah4Wii4v@digikod.net/T/#u) to do some
+//!   kinds of sandbox escapes, so we also block the `connect` syscall.
```
alindima, line 31: you can't `connect` to a socket without first creating it with `socket`. There's no fd to connect on.
We can keep the `connect` in the denylist, just to be safe, but maybe the comment is a bit misleading.

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
alindima: how do you plan to have the parent notified? using ptrace or `SECCOMP_RET_USER_NOTIF`?

If we only want logging, we can just rely on the kernel's [`Audit logging of seccomp actions`](https://man7.org/linux/man-pages/man2/seccomp.2.html)

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
```diff
@@ -0,0 +1,197 @@
...
+	Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use std::{io::ErrorKind, net::TcpListener, thread};
+
+	#[test]
+	fn sandboxed_thread_cannot_use_sockets() {
```
alindima, line 173: shouldn't we only run this test on x86_64?

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
```diff
@@ -0,0 +1,197 @@
...
+		let handle = thread::spawn(|| {
+			// Open a socket, this should succeed before seccomp is applied.
+			TcpListener::bind("127.0.0.1:0").unwrap();
+
+			let status = try_restrict();
+			if !matches!(status, Ok(())) {
+				panic!("Ruleset should be enforced since we checked if seccomp is enabled");
+			}
+
+			// Try to open a socket after seccomp.
```
alindima, line 188: also add an assertion for an allowed syscall, a simple one like `getpid`

polkadot/node/core/pvf/src/host.rs
```diff
@@ -981,6 +988,52 @@ fn check_landlock(
...
+///
+/// We do this check by spawning a new process and trying to sandbox it. To get as close as possible
+/// to running the check in a worker, we try it... in a worker. The expected return status is 0 on
+/// success and -1 on failure.
+async fn check_seccomp(
+	#[cfg_attr(not(target_os = "linux"), allow(unused_variables))]
+	prepare_worker_program_path: &Path,
+) -> bool {
+	cfg_if::cfg_if! {
+		if #[cfg(target_os = "linux")] {
```
alindima, line 1001: we should also only do this for x86.
Maybe we should conditionally compile the entire `seccomp` module to avoid this.

Installing a filter that was compiled for x86 on an arm machine will not fail. It'll install it and kill the process on the first syscall, regardless of what syscall it is or the rest of the filter

polkadot/node/core/pvf/src/host.rs
```diff
@@ -887,17 +893,17 @@ fn pulse_every(interval: std::time::Duration) -> impl futures::Stream<Item = ()>
...
 /// success and -1 on failure.
-fn check_can_unshare_user_namespace_and_change_root(
+async fn check_can_unshare_user_namespace_and_change_root(
 	#[cfg_attr(not(target_os = "linux"), allow(unused_variables))]
 	prepare_worker_program_path: &Path,
 ) -> bool {
 	cfg_if::cfg_if! {
 		if #[cfg(target_os = "linux")] {
-			let output = std::process::Command::new(prepare_worker_program_path)
+			match tokio::process::Command::new(prepare_worker_program_path)
```
alindima, line 902: why did we switch to using tokio here?

alindima (commented):
Looking good

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
mrcnski: ptrace. With `SECCOMP_RET_USER_NOTIF` we'd need to check if it's supported, as you previously pointed out. Maybe not a big deal as it was introduced in 5.0.

I considered audit logging, but AFAICT there are some issues with it. We'd need to manually parse the log which seems hacky (there is a Linux utility for it but not installed by default). Also, it's possible for operators to have disabled seccomp logging. ptrace seemed like the right way to do it, albeit more complex; but I'd really appreciate your input.

polkadot/node/core/pvf/common/Cargo.toml
```diff
@@ -30,6 +29,8 @@ sp-tracing = { path = "../../../../../substrate/primitives/tracing" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 landlock = "0.3.0"
+seccompiler = "0.3.0"
```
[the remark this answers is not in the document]
mrcnski: Yes! I saw that and was going to bump this but forgot.

polkadot/node/core/pvf/common/src/worker/security/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+			const STR: &[u8] = concat!($e, "\0").as_bytes();
+			const STR_VALID: bool = cstr_is_valid(STR);
+			let _ = [(); 0 - (!(STR_VALID) as usize)];
+			#[allow(unused_unsafe)]
+			unsafe {
+				core::ffi::CStr::from_bytes_with_nul_unchecked(STR)
+			}
+		}}
+	}
+
```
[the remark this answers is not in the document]
mrcnski: The macro is more readable, and I believe a bit more performant. `worker_dir_path_c` is a variable and not a literal so we can't use the macro there.

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
mrcnski: The whole module is gated on `x86_64`. As an aside, there should be a compile error on Linux arm64. I'm waiting to see if CI catches it because I have no way of testing that myself.

polkadot/node/core/pvf/src/host.rs
mrcnski: Previously we were running these checks sequentially. So, waiting for each process to spawn and return before running the next check. We can instead do this concurrently for a small performance gain. It's only relevant on node startup, but every 0.1 s regression adds up over time.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4073150

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
mrcnski: Looks like we have no CI job for arm64. 🤔

alindima: > The whole module is gated on x86_64. As an aside, there should be a compile error on Linux arm64. I'm waiting to see if CI catches it because I have no way of testing that myself.

Didn't notice that 👍🏻 

yes, the current code should fail to compile on arm. if you have an Arm macbook you can cross compile for linux like: `cargo build --target aarch64-unknown-linux-gnu` (and `rustup target add ...` beforehand)

polkadot/node/core/pvf/common/src/worker/security/mod.rs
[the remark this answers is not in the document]
alindima: It adds a lot of code IMO, if it can be easily done using `CString` methods.
My suggestion was that we remove this macro altogether (also getting rid of the bit of `unsafe` code).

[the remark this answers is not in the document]
koute: Every use here is of form `cstr!("/").as_ptr()`, so technically we don't actually need `CStr`s, so something like this would suffice too:

```
macro_rules! cstr_ptr {
    ($e:expr) => {
        concat!($e, "\0").as_ptr().cast::<core::ffi::c_char>>()
    };
}
```

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
alindima: > Also, it's possible for operators to have disabled seccomp logging.

well, users can disable ptrace too. But is this a concern? we only have access to the logs of nodes if they're run by us, right? we can configure the system whichever way we see fit.

I don't have a very strong opinion honestly. whichever option works, doesn't impact performance too much and isn't too complicated to implement.
I am guessing the most performant one is audit logging, because it's all happening in the kernel. But it's indeed more complicated to parse the logs and build the right plubming to feed them in loki

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
mrcnski: Thanks, caught the error with your help and https://github.com/messense/homebrew-macos-cross-toolchains and fixed it.

polkadot/node/core/pvf/src/host.rs
mrcnski: As we discussed, for this simple filter we could enable it for aarch64 also:

> using the libc syscall constants will transparently switch to the right numbers when building for ARM. and we just need to supply the right arch to seccompiler when building the filter.

But let's leave this gated on `x86_64` until we have the CI job - added a TODO in the code.

polkadot/node/core/pvf/common/src/worker/security/mod.rs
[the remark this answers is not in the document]
mrcnski: @alindima I also didn't want the extra code at first. However, it makes the rest of the function easier to read and that's the part that benefits from being readable. It makes it easier to e.g. compare side-by-side with the examples in `pivot_root(2)`.

@koute Thanks!

alindima (approved):

polkadot/node/core/pvf/common/src/worker/security/seccomp.rs
mrcnski: I went with parsing the audit logs. Audit events are not guaranteed to be observable, but we don't necessarily need them: the logs we emit are merely informative, so that operators know what happened. If we were to handle seccomp violations differently from regular worker death, then an attacker could perhaps abuse that non-determinism somehow. Using `ptrace` instead would be more reliable and (I think) the performance would be fine, but it seemed more convoluted to implement that strategy.

polkadot/node/core/pvf/tests/it/main.rs
```diff
@@ -153,6 +152,68 @@ async fn terminates_on_timeout() {
 	assert!(duration < TEST_EXECUTION_TIMEOUT * JOB_TIMEOUT_WALL_CLOCK_FACTOR);
 }
 
+// What happens when the prepare worker dies in the middle of a job?
+//
+// To avoid interfering with other running tests or processes, this test is ignored and must be run
```
alindima, line 157: how does it interfere? with the `killall` I would assume.
Could we spawn a process in a new pid namespace to avoid this?

alindima (commented):
Looks reasonable. Another way of getting a log would be to use the Trap return action and install a signal handler for SIGSYS that logs and then kills the process.

But that get's tricky because you can only use async-signal-safe functions in signal handlers (and we'd have to audit all of that code constantly).

mrcnski:
Perhaps we should only log for now, instead of killing the process. Just for one release. We haven't tested with all parachains, after all, and the `socket` call we're blocking is in the syscall lists for the processes.

I say we log, just to be safe. The jobs really shouldn't actually do any networking. (And creating a Unix socket would have already failed with our FS restrictions.)

alindima:
> Perhaps we should only log for now, instead of killing the process. Just for one release. We haven't tested with all parachains, after all, and the socket call we're blocking is in the syscall lists for the processes.
> I say we log, just to be safe. The jobs really shouldn't actually do any networking. (And creating a Unix socket would have already failed with our FS restrictions.)

Sounds good to me. We'll need some script that parses the audit logs from the kernel and notifies us if there's a logged violation

mrcnski:
> Another way of getting a log would be to use the Trap return action and install a signal handler for SIGSYS that logs and then kills the process.
>
> But that get's tricky because you can only use async-signal-safe functions in signal handlers (and we'd have to audit all of that code constantly).

I agree that signal handlers are better avoided where possible. :P And, according to man:

> This value results in the kernel sending a thread-directed SIGSYS signal to the triggering thread.

IIUC, with this scheme, an attacker could overwrite our signal handlers with his own.

> We'll need some script that parses the audit logs from the kernel and notifies us if there's a logged violation

I believe we can just parse the logs after every job, similar to how we do it now. We don't need a script running continuously (if I understood you correctly).

I've confirmed that we have time to log for one release before enabling the protections.

alindima:
> IIUC, with this scheme, an attacker could overwrite our signal handlers with his own.

You could add `sigaction` to the seccomp blacklist. Anyway, it's best to avoid signal handlers if possible, as we both agree

> I believe we can just parse the logs after every job, similar to how we do it now. We don't need a script running continuously (if I understood you correctly).

Sure. We still need some kind of rough notification mechanism if a fault occurred. Or just remember to scan the logs regularly :D

s0me0ne-unkn0wn:
> We'll need some script that parses the audit logs from the kernel

I'm not following the whole discussion so a possibly dumb question... Why do we need to parse audit logs instead of connecting to audit via netlink and observing audit events directly?

mrcnski:
@s0me0ne-unkn0wn I don't know what netlink is. :P Parsing logs is never ideal, but this is the best way I could find since encountering the problem back in April. If your solution is quick to implement I could do it, otherwise I've already implemented and tested the parsing way.

s0me0ne-unkn0wn:
@mrcnski see [man 3 audit_open](https://man7.org/linux/man-pages/man3/audit_open.3.html) and further links. It's the interface to the kernel's audit subsys. I think I sent you a usage example to Matrix PM once.

mrcnski:
Thanks @s0me0ne-unkn0wn, it's a good idea if I can figure out how to do it. Raised a follow-up at https://github.com/paritytech/polkadot-sdk/issues/2080.

polkadot/node/core/pvf/tests/it/main.rs
mrcnski: Rejigged this, see https://github.com/paritytech/polkadot-sdk/pull/2009/commits/54178182b900c66fc7b7da2f5f80f09ea810383f. Slightly hacky, but works, and preferable to having ignored tests in the repo.

polkadot/roadmap/implementers-guide/src/node/utility/pvf-host-and-workers.md
```diff
@@ -126,6 +126,17 @@ with untrusted code does not have unnecessary access to the file-system. This
...
+*Currently this is only supported on Linux.*
+
+### Restricting networking
+
+We also disable networking on PVF threads by disabling certain syscalls, such as
+the creation of sockets. This prevents attackers from either downloading
+payloads or communicating sensitive data from the validator's machine to the
+outside world.
+
+*Currently this is only supported on Linux.*
```
alindima, line 138: this isn't true yet

alindima (approved):

PR #2010, opened 2023-10-24
basic-authorship: Improve time recording and logging

bkchr:


davxy (approved):

substrate/client/basic-authorship/src/basic_authorship.rs
```diff
@@ -539,19 +543,24 @@ where
...
+	/// - `end_reason`: Why did we stop producing the block?
+	/// - `block_took`: How long did it took to produce the actual block?
+	/// - `propose_took`: How long did the entire proposing took?
 	fn print_summary(
 		&self,
 		block: &Block,
 		end_reason: EndProposingReason,
 		block_took: time::Duration,
-		propose_with_took: time::Duration,
+		propose_took: time::Duration,
```
michalkucharczyk, line 556: this could be info!'ed too

michalkucharczyk (approved):

ggwpez (approved):

substrate/client/basic-authorship/src/basic_authorship.rs
bkchr: I mean this is only includes the storage proof extraction, not really something we also need to print. IMO.

PR #2018, opened 2023-10-25
Build workers for testing on demand

### LINKED ISSUE #1957: PVF: consider building workers on-demand in tests
> Would still like that `get_and_check_worker_paths` just calls `cargo build` to provide some better UX for tests. However, this isn't a real blocker as the current behavior isn't changed.

_Originally posted by @bkchr in https://github.com/paritytech/polkadot-sdk/pull/1771#pullrequestreview-1688597620_
            
See also https://github.com/paritytech/polkadot-sdk/pull/1771#discussion_r1352135484.

eagr:
closes #1957

### DIFF
--- a/polkadot/node/core/pvf/src/testing.rs
+++ b/polkadot/node/core/pvf/src/testing.rs
@@ -79,7 +79,7 @@ pub fn get_and_check_worker_paths() -> (PathBuf, PathBuf) {
 			std::process::Command::new("cargo")
 				.args(build_args)
 				.stdout(std::process::Stdio::piped())
-				.current_dir()
+				.current_dir(dir)
 				.status()
 				.expect("Failed to run the build program");
 		}
--- a/polkadot/node/core/pvf/src/testing.rs
+++ b/polkadot/node/core/pvf/src/testing.rs
@@ -64,34 +64,28 @@ pub fn get_and_check_worker_paths() -> (PathBuf, PathBuf) {
 	static WORKER_PATHS: OnceLock<Mutex<(PathBuf, PathBuf)>> = OnceLock::new();
 
 	fn maybe_build_workers(build_prep: bool, build_exec: bool) {
-		let build_args = match (build_prep, build_exec) {
-			(true, true) => "build --bin polkadot-prepare-worker --bin polkadot-execute-worker",
-			(true, false) => "build --bin polkadot-prepare-worker",
-			(false, true) => "build --bin polkadot-execute-worker",
-			(false, false) => "",
-		};
+		let mut build_args = vec!["build"];
+		if build_prep {
+			build_args.push("--bin=polkadot-prepare-worker");
+		}
+		if build_exec {
+			build_args.push("--bin=polkadot-execute-worker");
+		}
 
-		if build_args.len() > 0 {
+		if build_prep || build_exec {
 			eprintln!("Building workers...");
 
-			let mut wd = std::env::current_exe().unwrap();
-			while !wd.ends_with("polkadot-sdk") {
-				wd.pop();
+			let mut dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+			if !dir.ends_with("polkadot-sdk") {
+				dir.pop();
 			}
 
-			let child = match std::process::Command::new("cargo")
-				.args(build_args.split(' '))
+			std::process::Command::new("cargo")
+				.args(build_args)
 				.stdout(std::process::Stdio::piped())
-				.current_dir(wd)
-				.spawn()
-			{
-				Ok(child) => child,
-				Err(err) => panic!("Failed to start cargo build: {}", err),
-			};
-
-			if let Err(err) = child.wait_with_output() {
-				panic!("Failed to build workers: {}", err);
-			}
+				.current_dir(dir)
+				.status()
+				.expect("Failed to run the build program");
 		}
 	}
 
@@ -110,25 +104,25 @@ pub fn get_and_check_worker_paths() -> (PathBuf, PathBuf) {
 		// Check that the workers are valid.
 
 		if !prepare_worker_path.is_executable() {
-			eprintln!("WARNING: Prepare worker does not exist or is not executable. Workers directory: {:?}", workers_path);
+			eprintln!("Prepare worker does not exist or is not executable. Workers directory: {:?}", workers_path);
 			build_prep = true;
 		}
 
 		if !execute_worker_path.is_executable() {
-			eprintln!("WARNING: Execute worker does not exist or is not executable. Workers directory: {:?}", workers_path);
+			eprintln!("Execute worker does not exist or is not executable. Workers directory: {:?}", workers_path);
 			build_exec = true;
 		}
 
 		if let Ok(ver) = get_worker_version(&prepare_worker_path) {
 			if ver != NODE_VERSION {
-				eprintln!("WARNING: Prepare worker version {ver} does not match node version {NODE_VERSION}; worker path: {prepare_worker_path:?}");
+				eprintln!("Prepare worker version {ver} does not match node version {NODE_VERSION}; worker path: {prepare_worker_path:?}");
 				build_prep = true;
 			}
 		}
 
 		if let Ok(ver) = get_worker_version(&execute_worker_path) {
 			if ver != NODE_VERSION {
-				eprintln!("WARNING: Execute worker version {ver} does not match node version {NODE_VERSION}; worker path: {execute_worker_path:?}");
+				eprintln!("Execute worker version {ver} does not match node version {NODE_VERSION}; worker path: {execute_worker_path:?}");
 				build_exec = true;
 			}
 		}
--- a/polkadot/node/core/pvf/src/testing.rs
+++ b/polkadot/node/core/pvf/src/testing.rs
@@ -55,7 +55,7 @@ pub fn validate_candidate(
 	Ok(result)
 }
 
-/// Retrieves the worker paths, checks that they exist and does a version check.
+/// Retrieves the worker paths and builds workers as needed.
 ///
 /// NOTE: This should only be called in dev code (tests, benchmarks) as it relies on the relative
 /// paths of the built workers.
@@ -63,25 +63,24 @@ pub fn get_and_check_worker_paths() -> (PathBuf, PathBuf) {
 	// Only needs to be called once for the current process.
 	static WORKER_PATHS: OnceLock<Mutex<(PathBuf, PathBuf)>> = OnceLock::new();
 
-	fn maybe_build_workers(build_prep: bool, build_exec: bool) {
-		let mut build_args = vec!["build", "--package=polkadot"];
-		if build_prep {
-			build_args.push("--bin=polkadot-prepare-worker");
-		}
-		if build_exec {
-			build_args.push("--bin=polkadot-execute-worker");
-		}
+	fn build_workers() {
+		let build_args = vec![
+			"build",
+			"--package=polkadot",
+			"--bin=polkadot-prepare-worker",
+			"--bin=polkadot-execute-worker",
+		];
+		let exit_status = std::process::Command::new("cargo")
+			// wasm runtime not needed
+			.env("SKIP_WASM_BUILD", "1")
+			.args(build_args)
+			.stdout(std::process::Stdio::piped())
+			.status()
+			.expect("Failed to run the build program");
 
-		if build_prep || build_exec {
-			eprintln!("Building workers...");
-
-			let dir = env!("CARGO_MANIFEST_DIR");
-			std::process::Command::new("cargo")
-				.args(build_args)
-				.stdout(std::process::Stdio::piped())
-				.current_dir(dir)
-				.status()
-				.expect("Failed to run the build program");
+		if !exit_status.success() {
+			eprintln!("Failed to build workers: {}", exit_status.code().unwrap());
+			std::process::exit(1);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,39 @@ pub fn validate_candidate(
...
+		};
+
+		if build_args.len() > 0 {
+			eprintln!("Building workers...");
+
+			let mut wd = std::env::current_exe().unwrap();
+			while !wd.ends_with("polkadot-sdk") {
+				wd.pop();
+			}
+
```
bkchr, line 81:
```suggestion
```

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,39 @@ pub fn validate_candidate(
...
+			let child = match std::process::Command::new("cargo")
+				.args(build_args.split(' '))
+				.stdout(std::process::Stdio::piped())
+				.current_dir(wd)
```
bkchr, line 85:
```suggestion
				.current_dir(env!("CARGO_MANIFEST_DIR"))
```

If someone points the target dir to somewhere outside of the repo.

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,39 @@ pub fn validate_candidate(
...
+				.spawn()
+			{
+				Ok(child) => child,
+				Err(err) => panic!("Failed to start cargo build: {}", err),
+			};
+
+			if let Err(err) = child.wait_with_output() {
+				panic!("Failed to build workers: {}", err);
+			}
```
bkchr, line 94:
```suggestion
				.status()
				.expect("Failed to run the build program");
```

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,39 @@ pub fn validate_candidate(
...
 	// Only needs to be called once for the current process.
 	static WORKER_PATHS: OnceLock<Mutex<(PathBuf, PathBuf)>> = OnceLock::new();
+
+	fn maybe_build_workers(build_prep: bool, build_exec: bool) {
+		let build_args = match (build_prep, build_exec) {
+			(true, true) => "build --bin polkadot-prepare-worker --bin polkadot-execute-worker",
+			(true, false) => "build --bin polkadot-prepare-worker",
+			(false, true) => "build --bin polkadot-execute-worker",
+			(false, false) => "",
+		};
```
bkchr, line 72:
```suggestion
		let mut build_args = vec!["build"];
		
		if build_prep {
		      build_args.push("--bin=polkadot-prepare-worker');
		}
		
		if build_exec {
		      build_args.push("--bin=polkadot-execute-worker');
		}
```

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,39 @@ pub fn validate_candidate(
...
+		if build_args.len() > 0 {
+			eprintln!("Building workers...");
+
+			let mut wd = std::env::current_exe().unwrap();
+			while !wd.ends_with("polkadot-sdk") {
+				wd.pop();
+			}
+
+			let child = match std::process::Command::new("cargo")
+				.args(build_args.split(' '))
```
bkchr, line 83:
```suggestion
				.args(build_args)
```

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -71,22 +104,37 @@ pub fn get_and_check_worker_paths() -> (PathBuf, PathBuf) {
...
 
+		let mut build_prep = false;
+		let mut build_exec = false;
+
 		// Check that the workers are valid.
-		if !prepare_worker_path.is_executable() || !execute_worker_path.is_executable() {
-			panic!("ERROR: Workers do not exist or are not executable. Workers directory: {:?}", workers_path);
+
+		if !prepare_worker_path.is_executable() {
+			eprintln!("WARNING: Prepare worker does not exist or is not executable. Workers directory: {:?}", workers_path);
```
bkchr, line 113: Not sure these warnings have any value, if we build the workers any way.

eagr: Still informative for some cases, let's keep them.

eagr (commented):
Thanks a lot! @bkchr

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,33 @@ pub fn validate_candidate(
...
+			build_args.push("--bin=polkadot-execute-worker");
+		}
+
+		if build_prep || build_exec {
+			eprintln!("Building workers...");
+
+			let mut dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+			if !dir.ends_with("polkadot-sdk") {
+				dir.pop();
+			}
```
bkchr, line 81:
```suggestion
			let dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```

We have a workspace, you can run this command everywhere in the workspace and it works.

bkchr (approved):

polkadot/node/core/pvf/src/testing.rs
eagr: The problem is we have so many of them and what `env!("CARGO_MANIFEST_DIR")` evaluates to depends on where you run `cargo test`, which doesn't necessarily have hose `--bin=*` build targets.

mrcnski (commented):
I haven't had time to try this yet, how is the UX? Can we remove the note in `pvf/tests/README.md` about manually building the workers?

polkadot/node/core/pvf/src/testing.rs
bkchr: I don't get what you are saying? `env!("CARGO_MANIFEST_DIR")` this is evaluated at compile time and thus, always return the same.

eagr: Yeah, it stays the same for a specific crate/package. If you run `cargo test` in any sub-dir in `/polkadot-sdk/polkadot/node/core/pvf`, it would evaluate to `/polkadot-sdk/polkadot/node/core/pvf` because that's where the manifest of this package resides. Then the `cargo build` would break as the manifest doesn't define these build targets.

eagr:
```
error: no bin target named `polkadot-prepare-worker`.
```

You would get something like this if you run `cargo test`.

bkchr: `cargo build -p polkadot --bin=polkadot-prepare-worker`

works.

eagr: Nice to know! This is more straightforward.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4112875

bkchr (approved):

polkadot/node/core/pvf/src/testing.rs
mrcnski: Which cases do you mean? Like if they get rebuilt and are still not executable? Then this could be a panic, including the command to build them manually.

eagr: Oh, I meant it could be a bit frustrating if you know you have the workers in place but a build still happens. It'd be nice to clarify that the worker is stale. Even for people who are aware of that, it doesn't hurt to see those messages.

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,29 @@ pub fn validate_candidate(
...
 	static WORKER_PATHS: OnceLock<Mutex<(PathBuf, PathBuf)>> = OnceLock::new();
+
+	fn maybe_build_workers(build_prep: bool, build_exec: bool) {
+		let mut build_args = vec!["build", "--package=polkadot"];
+		if build_prep {
+			build_args.push("--bin=polkadot-prepare-worker");
+		}
+		if build_exec {
+			build_args.push("--bin=polkadot-execute-worker");
+		}
```
mrcnski, line 73: Any reason not to simplify this and always build both workers, getting rid of the `build_*` arguments? I think it'd be rather rare for one to be up-to-date, but not the other. Plus, if something is already built, `cargo` will skip building it again. 

Maybe we could always try to build the workers. That would also satisfy my request [here](https://github.com/paritytech/polkadot-sdk/pull/2058#discussion_r1376050050) and it's simpler than doing a version check. The version check with the commit hash would only work if you've made a commit since the last build. But if you've changed code without committing, the version check would still pass without a rebuild being requested.

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,29 @@ pub fn validate_candidate(
...
+
+		if build_prep || build_exec {
+			eprintln!("Building workers...");
+
+			let dir = env!("CARGO_MANIFEST_DIR");
+			std::process::Command::new("cargo")
```
mrcnski, line 79: Can we pass `SKIP_WASM_BUILD=1`? I don't think we need the runtimes to be built for the workers.

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -62,6 +62,29 @@ pub fn validate_candidate(
...
+				.args(build_args)
+				.stdout(std::process::Stdio::piped())
+				.current_dir(dir)
+				.status()
+				.expect("Failed to run the build program");
```
mrcnski, line 84: We don't check the exit status of the command to see if was a `success()` or not. 

I think it'd be better to quit on failure. For example, if you look at this output with the terminal scrolled to the bottom, it can be easy to miss the errors:

<details>
<pre>
   Compiling polkadot-node-core-pvf-prepare-worker v1.0.0 (/Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/polkadot/node/core/pvf/prepare-worker)
   Compiling polkadot v1.1.0 (/Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/polkadot)
   Compiling polkadot-service v1.0.0 (/Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/polkadot/node/service)
   Compiling polkadot-cli v1.1.0 (/Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/polkadot/cli)
error: no rules expected the token `env`
  --> polkadot/src/bin/prepare-worker.rs:23:2
   |
23 |     env!("SUBSTRATE_CLI_COMMIT_HASH"),
   |     ^^^ no rules expected this token in macro call
   |
   = note: while trying to match sequence start

error[E0601]: `main` function not found in crate `polkadot_prepare_worker`
  --> polkadot/src/bin/prepare-worker.rs:24:3
   |
24 | );
   |   ^ consider adding a `main` function to `polkadot/src/bin/prepare-worker.rs`

For more information about this error, try `rustc --explain E0601`.
error: could not compile `polkadot` (bin "polkadot-prepare-worker") due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
thread 'main' panicked at polkadot/src/bin/prepare-worker.rs:19:1:
trying to run prepare-worker binary with the exit subcommand
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test worker_common::should_connect ... ok
test prepare_can_run_serially ... ok
Oct 30 16:40:12.350 ERROR runtime: panicked at /Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/polkadot/parachain/test-parachains/adder/src/wasm_validation.rs:36:74:
Executes block: StateMismatch    
test adder::execute_good_block_on_parent ... ok
test adder::execute_bad_block_on_parent ... ok
test adder::execute_can_run_serially ... ok
test adder::execute_good_chain_on_parent ... ok
test adder::stress_spawn ... ok
test worker_common::spawn_immediate_exit ... ok
test worker_common::spawn_timeout ... ok
test terminates_on_timeout ... ok
test deleting_prepared_artifact_does_not_dispute ... ok
test execute_queue_doesnt_stall_if_workers_died ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 21.01s

   Doc-tests polkadot-node-core-pvf

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
</pre>
</details>

polkadot/node/core/pvf/src/testing.rs
mrcnski: Hmm, when I try it I don't see the messages unless `cargo test -- --nocapture` is passed. Even then, there is so much output that they are easy to miss.

If we can figure out a way to *always* print the messages, that would be ideal, and I'd also try to make them stand out:

```
========================
WARNING:

Blah blah workers not available, rebuilding
========================
```

mrcnski (changes requested):
Thank you, really appreciate your contribution! I've played with it and it's already a big improvement. 🎉 I have some even more ideas for making the UX better.

mrcnski:
> I haven't had time to try this yet, how is the UX? Can we remove the note in `pvf/tests/README.md` about manually building the workers?

Yeah, the UX is really good. We can update the note in the readme, saying something like "the workers should get built automatically when testing, but here's the command just FYI".

polkadot/node/core/pvf/src/testing.rs
bkchr: Why clutter test output with this? This doesn't bring any benefit for the running tests.

polkadot/node/core/pvf/src/testing.rs
bkchr: Yeah I was also thinking about the same. However, then we should probably entirely drop any checking of versions or whatever and just build them.

eagr: Let me test it out.

polkadot/node/core/pvf/src/testing.rs
eagr: These are merely informational, nothing to take action for, it's ok that people miss them I guess.

mrcnski: It's possible for the workers to be rebuilt if the code changed but the version didn't change. In that case the warnings wouldn't be displayed. So these warnings, even if you pass `-- --nocapture`, are not even consistent. I'd personally vote for dropping them, but I suppose they don't hurt either.

mrcnski (approved):
Thank you!

eagr:
We gotta get this in. I also got bitten by this just now, changed the worker code but forgot to rebuild. If I didn't know about this, I could be chasing my own tail debugging for a while. 😄

mrcnski:
With pleasure. 👍

PR #1297, opened 2023-08-30
Initialise on-chain `StorageVersion` for pallets added after genesis

liamaharon:
Original PR https://github.com/paritytech/substrate/pull/14641

---

Closes https://github.com/paritytech/polkadot-sdk/issues/109

### Problem
Quoting from the above issue:

> When adding a pallet to chain after genesis we currently don't set the StorageVersion. So, when calling on_chain_storage_version it returns 0 while the pallet is maybe already at storage version 9 when it was added to the chain. This could lead to issues when running migrations.

### Solution

- Create a new trait `BeforeAllRuntimeMigrations` with a single method `fn before_all_runtime_migrations() -> Weight` trait with a noop default implementation
- Modify `Executive` to call `BeforeAllRuntimeMigrations::before_all_runtime_migrations` for all pallets before running any other hooks
- Implement `BeforeAllRuntimeMigrations` in the pallet proc macro to initialize the on-chain version to the current pallet version if the pallet has no storage set (indicating it has been recently added to the runtime and needs to have its version initialised).

### Other changes in this PR

- Abstracted repeated boilerplate to access the `pallet_name` in the pallet expand proc macro.

#### Why create a new hook instead of adding this logic to the pallet `pre_upgrade`?

`Executive` currently runs `COnRuntimeUpgrade` (custom migrations) before `AllPalletsWithSystem` migrations. We need versions to be initialized before the `COnRuntimeUpgrade` migrations are run, because `COnRuntimeUpgrade` migrations may use the on-chain version for critical logic. e.g. `VersionedRuntimeUpgrade` uses it to decide whether or not to execute.

We cannot reorder `COnRuntimeUpgrade` and `AllPalletsWithSystem` so `AllPalletsWithSystem` runs first, because `AllPalletsWithSystem` have some logic in their `post_upgrade` hooks to verify that the on-chain version and current pallet version match. A common use case of `COnRuntimeUpgrade` migrations is to perform a migration which will result in the versions matching, so if they were reordered these `post_upgrade` checks would fail.

#### Why init the on-chain version for pallets without a current storage version?

We must init the on-chain version for pallets even if they don't have a defined storage version so if there is a future version bump, the on-chain version is not automatically set to that new version without a proper migration.

e.g. bad scenario:

1. A pallet with no 'current version' is added to the runtime
2. Later, the pallet is upgraded with the 'current version' getting set to 1 and a migration is added to Executive Migrations to migrate the storage from 0 to 1
    a. Runtime upgrade occurs
[...]

PR #2019, opened 2023-10-25
Improve `try-state` developer experience & fix bug

### LINKED ISSUE #136: try-state: should indicate which pallet is failing
## Description 

`try_state` error handling more or less seems to follow this pattern:
```rust
ensure!(
    invariant_holds,
    "some explanation of the invariant not holding"
)
```

If `!invariant_holds`, the `try_state` call will return a `DispatchError::Other(message)` to `Executive`, who calls it like this: 
```rust
		// run the try-state checks of all pallets, ensuring they don't alter any state.
		let _guard = frame_support::StorageNoopGuard::default();
		<AllPalletsWithSystem as frame_support::traits::TryState<
			BlockNumberFor<System>,
		>>::try_state(*header.number(), select)
		.map_err(|e| {
			frame_support::log::error!(target: LOG_TARGET, "failure: {:?}", e);
			e
		})?;
		drop(_guard);
```

The issue here is that `Executive` does not log which pallet the error emitted from, which is vital information when there are multiple instances of the same pallet in the runtime.

I encountered this issue when investigating a `collective` pallet error in polkadot which has two instances. I needed to add additional logging to `try_state` and replace the `ensure!` with a conditional to learn which instance the error is coming from:

```diff
        fn do_try_state() -> Result<(), TryRuntimeError> {
+               use frame_support::traits::PalletInfo;
+               let pallet_error_log = "try-state invariant failed for pallet ".to_owned() +
+                       T::PalletInfo::name::<Pallet<T, I>>().unwrap_or("unknown_pallet") +
+                       ".";
+
                Self::proposals()
                        .into_iter()
                        .try_for_each(|proposal| -> Result<(), TryRuntimeError> {
@@ -1041,10 +1046,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
                        "The member count is greater than `MaxMembers`."
                );

-               ensure!(
-                       Self::members().windows(2).all(|members| members[0] <= members[1]),
-                       "The members are not sorted by value."
-               );
+               if !Self::members().windows(2).all(|members| members[0] <= members[1]) {
+                       log::error!("{}", &pallet_error_log);
+                       return Err("The members are not sorted by value.".into())
+               };		
```

### Replace `ensure` with conditionals and additional logging

The simplest way to resolve this is likely to simply replace all the `ensure` usage with conditionals, and an extra log containing the pallet name (optionally also index) like I did when investigating the collectives pallet issue.

This solution kind of sucks, making the code verbose. 

### Change try_state to return a new error type

I may be wrong but I'm not sure if it even makes sense for `try_state` to return a `DispatchError`, because it only ever checks storage and never actually dispatches anything?

Something very roughly like
```rust
// If you want P to be something that can be converted into a String or has some relation to String
#[derive(Eq, Clone, Copy, Encode, Debug, PartialEq)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub struct TryStateError<P: PalletInfo>(pub P, pub &'static str);
```

This way, `Executive` will have access to `P` as well as the error message and be able to log which pallet it came from.

liamaharon:
Making some devex improvements as I audit our chains adherence to try-state invariants, in preparation for automated try-state checks and alerting. 

Note to reviewer: while you're here, if you have time would be great to get your eyes on https://github.com/paritytech/polkadot-sdk/pull/1297 also since it touches a similar file and I'd like to avoid merge conflicts :P

## Devex Improvements

- Changes the log level of logs informing the user that try-state checks are being run for a pallet from debug to info
- Improves how errors are communicated
  - Errors are logged when they are encountered, rather than after everything has been executed
  - Exact pallet the error originated from is included with the error log
  - Clearly see all errors and how many there are, rather than only one
  - Closes #136 

### Example of new logs

Screenshot 2023-10-25 at 15 44 44

### Same but with old logs (run with RUST_LOG=debug)

Notice only informed of one of the errors, and it's unclear which pallet it originated

Screenshot 2023-10-25 at 15 39 01
 

## Bug fix

When dry-running migrations and `checks.try_state()` is `true`, only run `try_state` checks after migrations have been executed. Otherwise, `try_state` checks that expect state to be in at a HIGHER storage version than is on-chain could incorrectly fail.


liamaharon:
bot fmt

franciscoaguirre (approved):

substrate/frame/executive/src/lib.rs
```diff
@@ -356,16 +356,6 @@ where
 	pub fn try_runtime_upgrade(
 		checks: frame_try_runtime::UpgradeCheckSelect,
 	) -> Result<Weight, TryRuntimeError> {
-		if checks.try_state() {
```
franciscoaguirre, line 359: You should change the doc comment on this function if you remove this. It currently says you run try-state both before and after migrations.

kianenigma (approved):

gupnik (approved):

substrate/frame/support/procedural/src/pallet/expand/hooks.rs
```diff
@@ -271,12 +258,36 @@ pub fn expand_hooks(def: &mut Def) -> proc_macro2::TokenStream {
...
+					"🩺 Running {:?} try-state checks",
+					pallet_name,
+				);
 				<
 					Self as #frame_support::traits::Hooks<
 						#frame_system::pallet_prelude::BlockNumberFor::<T>
 					>
-				>::try_state(n)
+				>::try_state(n).map_err(|err| {
+					let pallet_name = <
```
ggwpez, line 277: Why is the pallet name here called again?

ggwpez (approved):

Issue #1549, opened 2023-09-13
Add destination info in FeeManager::handle_fee


franciscoaguirre:
In https://github.com/paritytech/polkadot-sdk/pull/1234, `XcmFeesToAccount` was added, which puts the delivery fees in a specified account.

A comment was raised about the benefit of putting the fees on different accounts based on the destination of the transport protocol.
However, this needs a change in the `FeeManager` API.
Right now `FeeManager::handle_fee` doesn't take in the destination, and it should to make this possible.

Something like this:

```rust
fn handle_fee(fee: MultiAssets, context: Option<&XcmContext>, destination: Option<&MultiLocation>);
```

KiChjang:
The original comment is really looking for the ID of the bridge that is going to be used to send XCMs over. Given that's the case, instead of modifying the function signature, we could simply pass in the bridge ID as part of the XcmContext -- I imagine that the ID would be useful in other situations as well.

serban300:
Not sure if possible (didn't try it), but another option that comes to mind would be to modify `handle_fee()` in order to provide the fee reason. Something like this:

```
fn handle_fee(fee: MultiAssets, context: Option<&XcmContext>, reason: FeeReason);
```

And add the destination to the `FeeReason` variants where it makes sense (for example to `FeeReason::TransferReserveAsset`, etc )

Seems a bit more specific then passing the bridge id or destination all the time. And maybe could be useful for other scenarios unrelated to the bridge.

KiChjang:
@serban300 I thought of that too, but in that case, you're just duplicating code for a specific use-case. I would like the solution to be more general than that, hence I suggested putting it into `XcmContext` altogether.

serban300:
@KiChjang Not sure if we would need to duplicate code for this. But you might be right. I didn't think of that.

@franciscoaguirre Could I help with this issue ? I would be happy to tackle it if you don't have time. Since we also need it for the bridges.

serban300:
Published https://github.com/paritytech/polkadot-sdk/pull/2021 for this.

@KiChjang in the end I added the `FeeReason` to `handle_fee()` since we needed it in order to add separate logic only for `ExportMessage` and I think in general it should offer more flexibility regarding fee handling. And there was no need for code duplication. But if you have strong preferences towards adding the network ID in the context, please let me know.

serban300:
Resolved in by https://github.com/paritytech/polkadot-sdk/pull/2021

PR #2021, opened 2023-10-25
Direct XCM `ExportMessage` fees for different bridges to different receiver accounts

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/1549 and https://github.com/paritytech/parity-bridges-common/issues/2554


polkadot/xcm/xcm-builder/src/fee_handling.rs
```diff
@@ -17,42 +17,87 @@
...
 );
+
 impl<
-		XcmConfig: xcm_executor::Config,
-		WaivedLocations: Contains<MultiLocation>,
+		AssetTransactor: TransactAsset,
 		AccountId: Clone + Into<[u8; 32]>,
-		ReceiverAccount: Get<Option<AccountId>>,
-	> FeeManager for XcmFeesToAccount<XcmConfig, WaivedLocations, AccountId, ReceiverAccount>
+		ReceiverAccount: Get<AccountId>,
```
svyatonik, line 51: Do not know if there was some design solution behind having `Get<Option<AccountId>>` in the previous version? Maybe some storage parameter was intended to be used - so fees still may be burnt unless governance has configured chain properly? But in any case - new version looks cleaner to me, just stumbled upon it

svyatonik (approved):
LGTM. During review, I realized that I've been sure that we will be distributing fees between different bridges at the sending chain (AH). But obviously I was wrong and BH is the right place to do that.

polkadot/xcm/xcm-builder/src/fee_handling.rs
serban300: I think `Get<Option<AccountId>>` was needed in order to support the case where we wanted to waive the fees for some locations and burn the ones for the other locations using the `XcmFeeToAccount`.

But now, that we have a separate component for handling fees, we can use `XcmFeeManagerFromComponents` to specify the set of waived locations and the handling logic independently. And we can use `XcmFeeManagerFromComponents<WaivedLocations, ()>` for the scenario above.

vgeddes:
This is also good for Snowbridge. In our case a portion of the collected export fee will be routed to the sovereign account of the parachain from which the computed origin descends. So we will likely implement a custom XcmFeesToAccount that does this routing.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -282,7 +282,26 @@ impl xcm_executor::Config for XcmConfig {
...
+	type FeeManager = XcmFeeManagerFromComponents<
+		WaivedLocations,
+		// Just showcasing that we can direct fees from different bridges to different accounts.
+		// We use the `TreasuryAccount` as the receiver account in any case by design.
+		(
+			XcmExportFeeToAccount<
+				Self::AssetTransactor,
+				RococoGlobalConsensusNetwork,
+				AccountId,
+				TreasuryAccount,
```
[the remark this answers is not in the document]
serban300: Good idea ! Will try to do it.

KiChjang:
I still think the appropriate place for this is in `XcmContext`. There may be situations where it's useful for configurable items to know what the ultimate destination for an XCM program is. The problem is whether we should include intermediate destinations -- maybe a `BoundedVec<MultiLocation, ConstU32<8>>` would do?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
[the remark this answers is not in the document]
serban300: @bkontur 
Done. PTAL ! I added it just for showcasing. Right now it doesn't do anything since we take no fee for `ExportMessage` and I guess it will changed in the future because of https://github.com/paritytech/parity-bridges-common/issues/2585 .

serban300:
@KiChjang 

> I still think the appropriate place for this is in `XcmContext`. There may be situations where it's useful for configurable items to know what the ultimate destination for an XCM program is. 

I experimented with this a bit, but I'm not sure if I understood correctly. I tried to add a `destination` field to the `XcmContext` as `Option<MultiLocation>` and set it whenever executing an instruction that has a destination. Like `TransferReserveAsset`, `ExportMessage`, etc. Is this what you are suggesting ? If so, I don't think this would be the ultimate destination of the XCM program. What about programs that have multiple instructions that call send (not sure, but I guess it's possible) ? For example
```
Xcm([
    ExportMessage( network: A, destination: A-para-1, xcm: ..),
    ClearOrigin,
    ExportMessage( network: B, destination: B-para-1, xcm: ..),
])
```
Each instruction could have a different destination, and some instructions might not have any destination at all (would be executed only locally). So considering that this destination is not "ultimate", do you think it would still make sense to add it ? 

> The problem is whether we should include intermediate destinations -- maybe a `BoundedVec<MultiLocation, ConstU32<8>>` would do?

I'm not sure if this is possible. From what I understand, the intermediate destinations unfold as we execute the sub-parts of the  XCM message on different chains, and the XCM itself doesn't hold a state across multiple chains. Only if we add a separate `intermediate_destinations` argument to each instruction, or an instruction `SetIntermediateDestinations`. Or maybe I'm missing something. Could you add more details on this suggestion please ?

KiChjang:
What does "intermediate destinations unfold as we execute the sub-parts" mean? Why is it not as simple as looking into the `xcm` parameter, and looking for specific arguments in certain instructions, such as the `dest` parameter of a `TransferReserveAsset`/`DepositReserveAsset`/`InitiateTeleport` instruction?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -401,3 +431,89 @@ impl ExportXcm for BridgeHubRococoOrBridgeHubWococoSwitchExporter {
...
+							source_para_account.clone(),
+						);
+
+						let dest_fee = total_fee - source_fee;
+						try_deposit_fee::<AssetTransactor, _>(
+							MultiAsset { id: asset.id, fun: Fungible(dest_fee) }.into(),
+							maybe_context,
+							dest_para_account.clone(),
+						);
+					},
```
acatangiu, line 503: nit: maybe add a comment why we're splitting fees between the two accounts

acatangiu (approved):

serban300:
> What does "intermediate destinations unfold as we execute the sub-parts" mean? Why is it not as simple as looking into the `xcm` parameter in instructions like `InitiateTeleport` or `InitiateReserveWithdraw`, and looking for specific arguments in certain instructions, such as the `dest` parameter of a `TransferReserveAsset`/`DepositReserveAsset`/`InitiateTeleport` instruction?

Sorry, you're right. I don't remember what I was thinking about. Anyway, yes, it would work if we parse the entire message beforehand. But still I think we would have multiple branches, and not an ultimate destination and the route to it.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
serban300: Done

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4149330

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4150230

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -290,7 +301,26 @@ impl xcm_executor::Config for XcmConfig {
 	type SubscriptionService = PolkadotXcm;
 	type PalletInstancesInfo = AllPalletsWithSystem;
 	type MaxAssetsIntoHolding = MaxAssetsIntoHolding;
-	type FeeManager = XcmFeesToAccount<Self, WaivedLocations, AccountId, TreasuryAccount>;
+	type FeeManager = XcmFeeManagerFromComponents<
+		WaivedLocations,
+		(
+			XcmExportFeeToRelayerRewardAccounts<
```
gavofyork, line 307: is this duplicated below? if so, pull out into a separate type to deduplicate.

polkadot/xcm/xcm-executor/src/traits/fee_manager.rs
```diff
@@ -51,9 +41,55 @@ pub enum FeeReason {
...
+				return unconsumed_fee;
+			}
+		)* );
+
+		unconsumed_fee
+	}
+}
+
+/// Handle stuff to do with taking fees in certain XCM instructions.
+pub trait FeeManager {
```
gavofyork, line 81: why is this not `trait FeeManager: HandleFee { ... }`?

polkadot/xcm/xcm-executor/src/traits/fee_manager.rs
```diff
@@ -51,9 +41,55 @@ pub enum FeeReason {
...
+	/// Separate component that handles the fees that are not waived.
+	type HandleFee: HandleFee;
```
gavofyork, line 83: not especially clean to split out a function with a type item.

gavofyork:
AFAICT, altering `FeeHandler` trait to read:

```rust
pub trait FeeManager {
	/// Determine if a fee which would normally payable should be waived.
	fn is_waived(origin: Option<&MultiLocation>, r: FeeReason) -> bool;

	/// Do something with the fee which has been paid. Doing nothing here silently burns the
	/// fees.
	fn handle_fee(fee: MultiAssets, context: Option<&XcmContext>, r: FeeReason);
}
```

should be the only API change which is required here. It's easy enough to write type parameterised utility/adapter structs allowing this trait to be composed from components.

gavofyork (changes requested):
As per my comment.

KiChjang:
> The argument which happens to be named `dest` or `destination` within one or two XCVM instructions is definitely not a property of the XCVM's context.

That's not what I'm arguing about. What I am saying is that this PR shouldn't simply modify `is_waived` to accept an additional `FeeReason`, but to instead add a `destination` field for `XcmContext`, and add it as a parameter to `is_waived`, since I've seen usage patterns where it can be really beneficial to know which destination the XCM is ultimately destined towards, e.g. when calculating the weight in `WeightTrader`.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
serban300: They are very similar, but they aren't duplicated. This one is for the AssetHubRococo -> AssetHubWococo bridge and the one below is for AssetHubWococo -> AssetHubRococo bridge.

polkadot/xcm/xcm-executor/src/traits/fee_manager.rs
serban300: Changed while adressing the high level comment

polkadot/xcm/xcm-executor/src/traits/fee_manager.rs
serban300: Changed while adressing the high level comment

serban300:
> AFAICT, altering `FeeHandler` trait to read:
> 
> ```rust
> pub trait FeeManager {
> 	/// Determine if a fee which would normally payable should be waived.
> 	fn is_waived(origin: Option<&MultiLocation>, r: FeeReason) -> bool;
> 
> 	/// Do something with the fee which has been paid. Doing nothing here silently burns the
> 	/// fees.
> 	fn handle_fee(fee: MultiAssets, context: Option<&XcmContext>, r: FeeReason);
> }
> ```
> should be the only API change which is required here. It's easy enough to write type parameterised utility/adapter structs allowing this trait to be composed from components.

Changed the `FeeManager` trait as you suggested. But we still need the `HandleFee` trait in order to chain fee handling strategies. Only moved this logic to a different place. I hope I understood correctly the suggestion.

gavofyork:
> But we still need the `HandleFee` trait in order to chain fee handling strategies. Only moved this logic to a different place.

Yes, but it's no longer core API but merely a helper trait specific to a utility adapter.

substrate/frame/support/procedural/src/pallet/expand/warnings.rs
```diff
@@ -33,9 +33,7 @@ pub(crate) fn weight_witness_warning(
 	if dev_mode {
 		return
 	}
-	let CallWeightDef::Immediate(w) = &method.weight else {
-		return
-	};
+	let CallWeightDef::Immediate(w) = &method.weight else { return };
```
gavofyork, line 36: Why is this being reformatted?

polkadot/xcm/xcm-executor/src/traits/fee_manager.rs
```diff
@@ -51,9 +41,20 @@ pub enum FeeReason {
...
+/// Handle stuff to do with taking fees in certain XCM instructions.
+pub trait FeeManager {
+	/// Determine if a fee should be waived.
+	fn is_waived(origin: Option<&MultiLocation>, reason: FeeReason) -> bool;
+
+	/// Do something with the fee which has been paid. Doing nothing here silently burns the
+	/// fees.
+	fn handle_fee(fee: MultiAssets, context: Option<&XcmContext>, reason: FeeReason);
+}
+
```
gavofyork, line 53: Avoid changing the declaration order to minimize the diff.

gavofyork (approved):
Modulo the comments.

serban300:
@KiChjang sorry, I merged this PR, maybe the changes make it into the next crates release which should be today from what I understand.

Please let's follow-up on the discussion related to adding the destination in the XcmContext in a separate issue/PR if that's ok.

PR #2022, opened 2023-10-25
Contracts: Add benchmarks to include files

pgherveou:
Project that includes pallet-contracts via crates.io will fail to run 
```bash
cargo check --features=runtime-benchmarks
```

without the currently not included benchmarks files


pgherveou:
cc @xermicus @athei

cmichi (approved):

agryaznov (approved):

PR #2026, opened 2023-10-25
Sort the benchmarks before listing them

ggwpez:


bkchr (approved):

substrate/utils/frame/benchmarking-cli/src/pallet/command.rs
```diff
@@ -748,13 +748,16 @@ impl CliConfiguration for PalletCmd {
...
-	benchmarks_to_run: Vec<(
+	mut benchmarks_to_run: Vec<(
 		Vec<u8>,
 		Vec<u8>,
 		Vec<(BenchmarkParameter, u32, u32)>,
 		Vec<(String, String)>,
 	)>,
 ) {
+	// Sort by pallet and function name.
+	benchmarks_to_run.sort_by(|(pa, sa, _, _), (pb, sb, _, _)| (pa, sa).cmp(&(pb, sb)));
```
bkontur, line 759: What about multiple pallet instances e.g. pallet_assets?
Shouldnt we consider also `BenchmarkList.instance` attribute?

ggwpez: It is currently not possible to benchmark only a single instance of a pallet. And IMHO it does not make much sense.  
But they will be listed multiple times, i will add a dedub to it.

bkontur: ok then,
I know the only one place where `list` command is used: https://github.com/paritytech/command-bot-scripts/blob/main/commands/bench/lib/bench-all-cumulus.sh#L26-L34 but I dont know what is your use case here :)

bkontur (approved):

substrate/utils/frame/benchmarking-cli/src/pallet/command.rs
ggwpez: Yea that also has a `uniq` in the pipe to dedub the instances, but now we already do it here.

PR #2034, opened 2023-10-25
Wasm Builder: Allow explicitly disabling features

JoshOrndorff:
This PR makes the Substrate Wasm Builder slightly more flexible by allowing users to explicitly disable certain features when building their runtimes' wasm binaries.

Before this PR, the wasm builder already allow _en_abling specific features, but not _dis_abling them. It included good defaults in terms of what features should be disabled (std and any default features) that work for most runtimes.

The usecase that I have in mind is a runtime that can be compiled for a sovereign chain or a parachain. The features will look basically like this.

```toml
[features]
default = [ "std" ]
std = [
	"parity-scale-codec/std",
	"sp-core/std",
	# ...
]
parachain = [
	"cumulus-primitives-core",
	"cumulus-pallet-parachain-system",
	# ...
]
parachain-std = [
	"parachain",
	"std",
	"cumulus-primitives-core/std",
	"cumulus-pallet-parachain-system/std",
	# ...
]
```

The trouble comes when I try to build the parachain node. I need the feature `parachain-std` on the client side, but I only need `parachain` in the wasm. Therefore I need to explicitly disable `parachain-std` when building the wasm.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,9 +121,9 @@ dependencies = [
 
 [[package]]
 name = "aes-gcm"
-version = "0.10.3"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+checksum = "209b47e8954a928e1d72e86eca7000ebb6655fe1436d33eefc2201cad027e237"
 dependencies = [
  "aead 0.5.2",
  "aes 0.8.3",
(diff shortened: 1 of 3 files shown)


bkchr:
You don't need `parachain-std`. You can just enable `std` conditionally for these crates, when they are enabled.

```
[features]
default = [ "std" ]
std = [
	"parity-scale-codec/std",
	"sp-core/std",
	"cumulus-primitives-core?/std",
	"cumulus-pallet-parachain-system?/std",
]
parachain = [
	"cumulus-primitives-core",
	"cumulus-pallet-parachain-system",
	# ...
]
```

I think this is a anti pattern to disable features and given that you can fix it the way I presented you above, I will close this pr.

JoshOrndorff:
OMG, I've tried several different ways to do this over the years and today I wanted it so bad I finally broke down and learned how the wasm-builder works.

I can't believe it was as simple as adding the `?`. Thank you for teaching me this!

Glad I read the wasm-builder code anyway.

PR #2038, opened 2023-10-26
feature: add transferrable balance to currency trait to fix balance inconsistency

dnjscksdn98:
# Description

This PR will resolve the balance inconsistency between Substrate's runtime and Frontier's `pallet_evm` that occurred after this prior [PR](https://github.com/paritytech/substrate/pull/12951) was merged. In the mentioned PR, a new concept has been introduced in the balance system such as holds and frozen. However, the update can be disabled by setting some config types to empty tuples in the runtime as below.
```
type FreezeIdentifier = ();
type MaxFreezes = ();
type HoldIdentifier = ();
type MaxHolds = ();
```
Furthermore, the new interface has been moved to the fungible trait and the old interface is left in the currency trait. From this point it affected the issue to `pallet_evm`. Currently, `pallet_evm`'s `Config` trait contains a [Currency](https://github.com/paritytech/frontier/blob/master/frame/evm/src/lib.rs#L144) type, that will mostly be set to `pallet_balances`. The `Currency` type has a trait bound specified as below.
```
type Currency: Currency<Self::AccountId> + Inspect<Self::AccountId>;
```
The problem is that it contains the `Inspect` trait that is from `fungible`, which is the target interface that has been updated to the new balance system. Previously, `Inspect` was added in order to use the `reducible_balance()` method to fetch the current transferrable balance of an account. But now this method has been modified to follow the new balance system, which means that even though the new balance system is disabled for a moment, `pallet_evm` will fetch transferrable balance through the modified logic. As a result, this will lead to balance inconsistency (In the case when an account has locked and reserved balance both).
- locked < reserved
  - evm balance = substrate balance + locked
- locked > reserved
  - evm balance = substrate balance + reserved
- locked == reserved
  - evm balance = substrate balance + locked (or reserved)

So, I added a new method to the `currency` trait named as `transferrable_balance()` that has the same functionality as the old version of `reducible_balance()`. I know that the `currency` trait is deprecated and will be removed further some day, but for a short term until the balance system is fully migrated, I think this will be a good to go for chains that implements Frontier for now. (This issue affects every chain that has implemented Frontier's `pallet_evm` such as Moonbeam.)

And for more information, there is another side effect occurred by this issue. Since the evm balance are shown higher than expected, even though the actual balance is insufficient to pay, evm transactions are still capable to be sent. However, the transaction will be stuck in the mempool infinitely until the balance is filled enough to be payed.

Perhaps if this PR has been merged in any good way, the `pallet_evm` must be updated as below.
- Remove the `Inspect` trait
- Replace [reducible_balance()](https://github.com/paritytech/frontier/blob/master/frame/evm/src/lib.rs#L897) to `transferrable_balance()`
  - This will be good to be rollbacked, when it is fully migrated to the new balance system.


alstjd0921 (approved):

substrate/frame/balances/src/impl_currency.rs
```diff
@@ -487,6 +492,24 @@ where
 		)
 		.unwrap_or_else(|_| SignedImbalance::Positive(Self::PositiveImbalance::zero()))
 	}
+
+	fn transferrable_balance(who: &T::AccountId, preservation: Preservation) -> Self::Balance {
```
ggwpez, line 496: Would it be possible to factor this one function out into a `LegacyCurrencyTransferrableBalance` trait or something?  
Than you can require that in `pallet_evm` like `type Currency: Currency<Balance> + LegacyCurrencyTransferrableBalance<Balance>;` or so.  
It seems like a better solution than editing the existing trait. Do you think it could work?

PR #2040, opened 2023-10-26
add `authorities_len` for aura

yjhmelody:


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4090785

bkchr (approved):

substrate/frame/aura/src/lib.rs
```diff
@@ -230,6 +230,11 @@ impl<T: Config> Pallet<T> {
 		}
 	}
 
+	/// Return current authorities length.
+	pub fn authorities_len() -> usize {
+		Authorities::<T>::decode_len().unwrap_or(0)
```
georgepisaltu, line 235: If `decode_len` returns `None`, it's likely because there are inconsistencies in storage and the value itself won't be able to decode. My first thought was to return an error, but I see the pallet doesn't have one configured and throughout the code `panic` and `assert` are used. So we could do `expect` here or `unwrap_or(T::MaxAuthorities::get())` to be more defensive.

substrate/frame/aura/src/lib.rs
```diff
@@ -363,7 +368,7 @@ impl<T: Config> FindAuthor<u32> for Pallet<T> {
 		for (id, mut data) in digests.into_iter() {
 			if id == AURA_ENGINE_ID {
 				let slot = Slot::decode(&mut data).ok()?;
-				let author_index = *slot % Self::authorities().len() as u64;
+				let author_index = *slot % Self::authorities_len() as u64;
```
georgepisaltu, line 371: With the current changes (possibly even before this PR), this could be a division by 0 here. We could do a check and return no author if the authority count is 0.

substrate/frame/aura/src/lib.rs
bkchr: If you can not decode the authorities, the length would be `0`. I don't think this is wrong.

substrate/frame/aura/src/lib.rs
bkchr: With `0` authorities you don't have anyone that can produce a block and thus, you would never end up here :P

georgepisaltu (approved):

Issue #1915, opened 2023-10-17
`ChainSync` requesting the same block multiple times


altonen:
While running a burn-in for #1631, the logs showed that multiple different peers were banned because they had made the same block request more than once: https://github.com/paritytech/polkadot-sdk/blob/a1171e6e81d5c103e36d40c0a50d2e37157c3dcc/substrate/client/network/sync/src/block_request_handler.rs#L65

This bug is likely in master since the libp2p upgrade doesn't change any syncing-related logic

altonen:
Reported by @dmitry-markin: this has been happening at least on Kusama (likely Polkadot too) already a month ago.

dmitry-markin:
The same warning was happening in Subspace back on 2023-08-09, reported here: https://github.com/paritytech/polkadot-sdk/issues/493#issuecomment-1691854689

nazar-pc:
Could be https://github.com/subspace/subspace/issues/2122 caused by https://github.com/paritytech/polkadot-sdk/pull/1812

dmitry-markin:
> Could be [subspace/subspace#2122](https://github.com/subspace/subspace/issues/2122) caused by #1812

May be https://github.com/paritytech/polkadot-sdk/pull/1812 introduced another issue, but this issue was already happening way before https://github.com/paritytech/polkadot-sdk/pull/1812 was merged.

nazar-pc:
Yes, our conclusion is that there are several issues, but #1812 seems to trigger some other more often that before.

dmitry-markin:
So far, there are at least 4 issues in `ChainSync`:

1. `BlockCollection`: blocks below the common number are requested from peers: [logs](https://gist.github.com/dmitry-markin/d4a9a76468d35efa9883e92b27f71bbd#file-gistfile1-txt-L72). Addressed by https://github.com/paritytech/polkadot-sdk/pull/2045.
[...]

PR #2045, opened 2023-10-26
Do not request blocks below the common number when syncing

dmitry-markin:
This changes `BlockCollection` logic so we don't download block ranges from peers with which we have these ranges already in sync.

Improves situation with https://github.com/paritytech/polkadot-sdk/issues/1915.


substrate/client/network/sync/src/lib.rs
```diff
@@ -2641,104 +2641,104 @@ mod test {
 	/// have) from it. In the meanwhile peer 2 sends us block 4 and 3 and we send another request
 	/// for block 2 to peer 2. Peer 1 answers with block 2 and then peer 2. This will need to
 	/// succeed, as we have requested block 2 from both peers.
-	#[test]
-	fn do_not_report_peer_on_block_response_for_block_request() {
```
dmitry-markin, line 2645: I don't think this test makes sense if the PR is merged, because the logic is changed so we don't request block 2 from peer 2 anymore.

substrate/client/network/sync/src/blocks.rs
```diff
@@ -123,20 +123,36 @@ impl<B: BlockT> BlockCollection<B> {
...
 			loop {
 				let next = downloading_iter.next();
 				break match (prev, next) {
+					// If we are already downloading this range, request it from `max_parallel`
+					// peers (`max_parallel = 5` by default).
+					// Do not request already downloading range from peers with common number above
+					// the range start.
 					(Some((start, &BlockRangeState::Downloading { ref len, downloading })), _)
-						if downloading < max_parallel =>
+						if downloading < max_parallel && *start >= first_different =>
```
altonen, line 137: How much does this help with the situation?

Based on our discussion in Element, I understood the issue to be in the overlapping ranges which for some reason don't get cleared and instead get downloaded again, something I still don't fully understand given what the logs said. It's all because of parallel downloads?

In https://github.com/paritytech/polkadot-sdk/issues/1915#issuecomment-1784175617 you said the peers' common numbers are updated inconsistently (block is queued vs imported) so I'm wondering can we trust these common numbers with any further logic?

Thanks for adding comments for these cases

substrate/client/network/sync/src/blocks.rs
```diff
@@ -365,10 +381,10 @@ mod test {
 		bc.blocks.insert(114305, BlockRangeState::Complete(blocks));
 
 		let peer0 = PeerId::random();
-		assert_eq!(bc.needed_blocks(peer0, 128, 10000, 000, 1, 200), Some(1..100));
-		assert_eq!(bc.needed_blocks(peer0, 128, 10000, 600, 1, 200), None); // too far ahead
+		assert_eq!(bc.needed_blocks(peer0, 128, 10000, 0, 1, 200), Some(1..100));
+		assert_eq!(bc.needed_blocks(peer0, 128, 10000, 0, 1, 200), None); // too far ahead
 		assert_eq!(
-			bc.needed_blocks(peer0, 128, 10000, 600, 1, 200000),
+			bc.needed_blocks(peer0, 128, 10000, 0, 1, 200000),
```
altonen, line 387: Why did these change?

substrate/client/network/sync/src/blocks.rs
```diff
@@ -123,20 +123,36 @@ impl<B: BlockT> BlockCollection<B> {
...
 		let count = (count as u32).into();
 		let (mut range, downloading) = {
+			// Iterate through the ranges in `self.blocks` looking for a range to download
 			let mut downloading_iter = self.blocks.iter().peekable();
 			let mut prev: Option<(&NumberFor<B>, &BlockRangeState<B>)> = None;
 			loop {
 				let next = downloading_iter.next();
 				break match (prev, next) {
+					// If we are already downloading this range, request it from `max_parallel`
+					// peers (`max_parallel = 5` by default).
```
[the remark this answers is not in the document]
dmitry-markin: I don't understand why we need to request same blocks from multiple peers simultaneously in the first place. We'll import only blocks sent by the first responding peer, other responses will be effectively discarded.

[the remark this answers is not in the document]
altonen: [The comment](https://github.com/paritytech/polkadot-sdk/blob/0aeab381380107a6462fea240181f6e2a78f21b8/substrate/client/cli/src/params/network_params.rs#L112-L117) for the CLI flag explains. Slow peers could slow down syncing since each request can take up to 20 seconds. Could be worth checking how much slower sync from genesis to some block X would be if the value was set to 1. Would need a reproducible environment though.

substrate/client/network/sync/src/blocks.rs
dmitry-markin: Without this patch the issue is happening every time versi is scaled up. With this patch applied the only incident (literally two nodes banned) that I observed was after failed import when the sync was restarted.

It was not happening due to overlapping ranges in the first place, but because when some peer had a common number lower than others, we requested old ranges from that peer, but also made requests to other peers below their common numbers (either parallel requests or requests of adjacent block ranges).

I also worry that common numbers are not very trustworthy to rely on. Not sure if "probabilistic" approach to common numbers works good enough to consider this PR safe merging.

substrate/client/network/sync/src/blocks.rs
dmitry-markin: Because the ranges requested in this test were below the common number supplied (600), and therefore not requested with the fix.

substrate/client/network/sync/src/blocks.rs
dmitry-markin: P.S. I also observed the issue with the fix applied when versi was scaled before it was reset to a new chainspec, but it's not reproducible anymore.

altonen: I looked at the syncing code and how it updates common number for peers but I don't think there is anything problematic happening, at least not anything I can see. Common number is updated for all peers that have already gone through ancestry search and the code verifies that the common number is not above peer's best number.

Do you have a concrete concern with the usage of common numbers or can we got ahead and merge this? I understood this has gone through extensive Versi testing already

dmitry-markin: > Do you have a concrete concern with the usage of common numbers or can we got ahead and merge this? I understood this has gone through extensive Versi testing already

Yes, my specific concern is that after we queue blocks for import in `on_block_queued`, we move common numbers of all peers forward to the number of the queued block (see [here](https://github.com/paritytech/polkadot-sdk/blob/8ce16ee6ff190912995dc6c8da85b741eaada395/substrate/client/network/sync/src/lib.rs#L1270-L1271)), not checking if the peers are actually on the same chain as the block imported.

IIRC, there were also some other places where we just put _something_ into the common number (our best block?), not actually checking if we are on the same chain.

dmitry-markin: Here is this another common number update that I don't understand: https://github.com/paritytech/polkadot-sdk/blob/8ce16ee6ff190912995dc6c8da85b741eaada395/substrate/client/network/sync/src/lib.rs#L774

altonen: I'm not convinced it is a problem. We've done an ancestry search for all the peers so they're all in the same chain and the new common block is only set if it's less than the peer's best number. What are the potential issues that can arise from this?

@arkpar do you have any comments?

dmitry-markin: Why can't it be that the peer is on the same chain up to the last common number, but above the common number it's actually on a different fork up to it's best block? In this scenario it's unsafe to fast forward peer's common number between the last common number and peer's best number.

Am I missing something?

altonen: I don't think anything's missing but I don't understand what is going to happen if this is merged as-is. Node attempts to download a range from peer which it doesn't have because the common number was unsafely modified? What happens then? What was happening in Westend sounded far worse because there's no alternative faulty behavior to weigh against it.

dmitry-markin: Possible situation that bothers me — we won't be able to download a block range from a peer if it's common number is believed to be higher than it is. The bug this PR fixes might have facilitated block download in cases where the common number was estimated to be higher than it was in fact. With this PR merged, we will lose opportunity to download block ranges from such peers, and importing the range downloaded above the (incorrectly) estimated common number might fail, if the peer is in fact on a different fork.

The worst that might happen — we are stuck on a fork with common numbers of all peers incorrectly estimated to be higher than they are, downloading block ranges above these common numbers from the real canonical chain, but failing to import them because we lack some blocks in between. Something like https://github.com/paritytech/polkadot-sdk/issues/493.

arkpar: The proposed change looks correct to me. Indeed, there should be no requests for other peers below the common block. Duplicate requests may happen because the same block range is also requested with a fork download mechanism.

> Here is this another common number update after ancestor search that I don't understand: 

Common ancestor search can't be reliable when the node is syncing at full speed. While the ancestry search is happening the node may import thousands of blocks and then detected common block is obsolete and too far behind. 

In general it is not possible to guarantee that block that we consider "common" is up to date. Other peers may reorg at any time wihouth notification and the common number may indeed decrease.

The whole notion of "common number" and "best chain" is really obsolete and should be removed for sync 2.0. There's just the finalized chain (which can be dowloaded in parallel, with gaps) and a buch of unfinalized forks, all threated as equal.

arkpar (approved):

PR #1661, opened 2023-09-21
Make `IdentityInfo` generic in `pallet-identity`

georgepisaltu:
Fixes #179 

# Description

This PR makes the structure containing identity information used in `pallet-identity` generic through the pallet `Config`. Additionally, the old structure is now available in a separate module called `simple` (pending rename) and is compatible with the new interface.

Another change in this PR is that while the `additional` field in `IdentityInfo` stays for backwards compatibility reasons, the associated costs are stil present in the pallet through the `additional` function in the `IdentityInformationProvider` interface. This function is marked as deprecated as it is only a temporary solution to the backwards compatibility problem we had. In short, we could have removed the additional fields in the struct and done a migration, but we chose to wait and do it off-chain through the genesis of the system parachain. After we move the identity pallet to the parachain, additional fields will be migrated into the existing fields and the `additional` key-value store will be removed. Until that happens, this interface will provide the necessary information to properly account for the associated costs.

Additionally, this PR fixes an unrelated issue; the `IdentityField` enum used to represent the fields as bitflags couldn't store more than 8 fields, even though it was marked as `#[repr(u64)]`. This was because of the `derive` implementation of `TypeInfo`, which assumed `u8` semantics. The custom implementation of this trait in https://github.com/paritytech/polkadot-sdk/commit/0105cc0396b7a53d0b290f48b1225847f6d17321 fixes the issue.

PR #2048, opened 2023-10-26
Identity pallet improvements

georgepisaltu:
This PR is a follow up to #1661 

- [x] rename the `simple` module to `legacy`
- [x] fix benchmarks to disregard the number of additional fields
- [x] change the storage deposits to charge per encoded byte of the identity information instance, removing the need for `fn additional(&self) -> usize` in `IdentityInformationProvider`
- [x] ~add an extrinsic to rejig deposits to account for the change above~
- [ ] ~ensure through proper configuration that the new byte-based deposit is always lower than whatever is reserved now~
- [x] remove `IdentityFields` from the `set_fields` extrinsic signature, as per [this discussion](https://github.com/paritytech/polkadot-sdk/pull/1661#discussion_r1371703403)

> ensure through proper configuration that the new byte-based deposit is always lower than whatever is reserved now

Not sure this is needed anymore. If the new deposits are higher than what is currently on chain and users don't have enough funds to reserve what is needed, the extrinisc fails and they're basically grandfathered and frozen until they add more funds and/or make a change to their identity. This behavior seems fine to me. Original idea [here](https://github.com/paritytech/polkadot-sdk/pull/1661#issuecomment-1779606319).

> add an extrinsic to rejig deposits to account for the change above

This was initially implemented but now removed from this PR in favor of the implementation detailed [here](https://github.com/paritytech/polkadot-sdk/pull/2088).


substrate/frame/identity/src/lib.rs
```diff
@@ -97,6 +100,9 @@ type NegativeImbalanceOf<T> = <<T as Config>::Currency as Currency<
 >>::NegativeImbalance;
 type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
 
+/// Maximum size of an identity we can store is 4 MiB.
+const IDENTITY_MAX_SIZE: u32 = 4 * 1024 * 1024;
```
joepetrowski, line 104: This seems very large, I imagine even 1 kB would be enough. Maybe 5 kB?

substrate/frame/identity/src/lib.rs
```diff
@@ -328,19 +334,15 @@ pub mod pallet {
...
 			info: Box<T::IdentityInformation>,
 		) -> DispatchResultWithPostInfo {
 			let sender = ensure_signed(origin)?;
-			#[allow(deprecated)]
-			let extra_fields = info.additional() as u32;
-			ensure!(extra_fields <= T::MaxAdditionalFields::get(), Error::<T>::TooManyFields);
-			let fd = <BalanceOf<T>>::from(extra_fields) * T::FieldDeposit::get();
+			let encoded_byte_size = info.encoded_size() as u32;
+			ensure!(encoded_byte_size <= IDENTITY_MAX_SIZE, Error::<T>::IdentitySize);
+			let byte_deposit = T::ByteDeposit::get() * <BalanceOf<T>>::from(encoded_byte_size);
```
joepetrowski, line 345:
```suggestion
			let byte_deposit = T::ByteDeposit::get().saturating_mul(<BalanceOf<T>>::from(encoded_byte_size));
```

substrate/frame/identity/src/lib.rs
```diff
@@ -357,7 +359,7 @@ pub mod pallet {
 			};
 
 			let old_deposit = id.deposit;
-			id.deposit = T::BasicDeposit::get() + fd;
+			id.deposit = T::BasicDeposit::get() + byte_deposit;
```
joepetrowski, line 362:
```suggestion
			id.deposit = T::BasicDeposit::get().saturating_add(byte_deposit);
```

substrate/frame/identity/src/lib.rs
```diff
@@ -922,6 +900,59 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Update the deposits held by `target` for its identity info.
+		///
+		/// Parameters:
+		/// - `target`: The account for which to update deposits.
+		///
+		/// May be called by any signed origin.
+		#[pallet::call_index(16)]
```
joepetrowski, line 910: I mean either one is fine, but this will get in before #1814 
```suggestion
		#[pallet::call_index(15)]
```

substrate/frame/identity/src/lib.rs
```diff
@@ -922,6 +900,59 @@ pub mod pallet {
...
+			// Identity Deposit
+			ensure!(IdentityOf::<T>::contains_key(&target), Error::<T>::NoIdentity);
+			let new_id_deposit = IdentityOf::<T>::try_mutate(
+				&target,
+				|registration| -> Result<BalanceOf<T>, DispatchError> {
+					let reg = registration.as_mut().ok_or(Error::<T>::NoIdentity)?;
+					// Calculate what deposit should be
+					let encoded_byte_size = reg.info.encoded_size() as u32;
+					let byte_deposit =
+						T::ByteDeposit::get() * <BalanceOf<T>>::from(encoded_byte_size);
```
joepetrowski, line 927:
```suggestion
						T::ByteDeposit::get().saturating_mul(<BalanceOf<T>>::from(encoded_byte_size));
```

substrate/frame/identity/src/lib.rs
georgepisaltu: I just took the constant from the preimage pallet. Here is a more informed view: a "full" `IdentityInfo` without any entries in `additional` has an encoded size of 253 bytes. Each entry in `additional` is then worth at most 66 bytes, depending on the values chosen for the key and value, since they're `Data`. With the current values for `T::MaxAdditionalFields` of 100, that brings us to at most `253 + 66 * 100` or ~6.7 KiB maximum encoded size for any value currently in storage. Considering this, I set it to 7 KiB now and we can lower it later if we wish to do so.

substrate/frame/identity/src/lib.rs
georgepisaltu: Done.

substrate/frame/identity/src/lib.rs
georgepisaltu: Done.

substrate/frame/identity/src/lib.rs
georgepisaltu: Done.

substrate/frame/identity/src/lib.rs
georgepisaltu: Done.

georgepisaltu (commented):
I also added some more checked arithmetic where applicable, just to be consistent.

substrate/frame/identity/src/lib.rs
joepetrowski: Yeah, that sounds good. The preimage pallet is meant to store data such that you can remove it if the system hasn't requested it. Setting an identity clears any judgements, so we don't really expect people to remove their data once they have set it. Of course the byte fee accounts for it, but it seems more flexible to just have a field meant to take a hash that can be hosted elsewhere. Then, people could actually use the preimage pallet if it needed to be on-chain, or some off-chain hosting service if the reader of the Identity Info knew where to look for it.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4111258

georgepisaltu (commented):
> remove IdentityFields from the `set_fields` extrinsic signature, as per https://github.com/paritytech/polkadot-sdk/pull/1661#discussion_r1371703403

The `set_fields` extrinisc now accepts the underlying numerical representation as the input. This is currently `u64` for us and will probably continue to be since there's no reason to change it. This change requires no migration now, since it's the same `u64` in storage as before, but if the underlying numeric type changes when we move away from the `legacy` impls, it will require a migration.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4143061

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4143092

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -1460,7 +1460,7 @@ impl pallet_grandpa::Config for Runtime {
 
 parameter_types! {
 	pub const BasicDeposit: Balance = 10 * DOLLARS;       // 258 bytes on-chain
-	pub const FieldDeposit: Balance = 250 * CENTS;        // 66 bytes on-chain
+	pub const ByteDeposit: Balance = deposit(0, 1);
 	pub const SubAccountDeposit: Balance = 2 * DOLLARS;   // 53 bytes on-chain
```
joepetrowski, line 1464: For Rococo and Westend we can leave them since we'll drop the pallet from the Relay Chains and configure it correctly in genesis of the parachain, but here (being the example runtime), I think we should use `deposit()` for the `BasicDeposit` and `SubAccountDeposit` configs using the smallest encoded size of a set identity.

substrate/frame/identity/src/lib.rs
```diff
@@ -240,6 +243,8 @@ pub mod pallet {
 		InvalidTarget,
 		/// Too many additional fields.
 		TooManyFields,
+		/// Identity encoded size is too high.
```
joepetrowski, line 246:
```suggestion
		/// Identity encoded size is too large.
```

joepetrowski (approved):
Should also get review from @bkchr and/or @ggwpez before merging though

substrate/bin/node/runtime/src/lib.rs
georgepisaltu: To answer the obvious questions first: the smallest identity you can set is an empty one, encoding to 9 bytes. The whole `Registration` `struct` that actually contains the identity would encode to a bit more, but it has stuff we don't charge for right now, such as the deposit held and the judgements.

That being said, I think of this deposit a bit differently. First off, even with the smallest identity, those 9 bytes will be accounted for because we take the encoded size of the whole `struct`, not just the data within it. Right now, the base deposit is way bigger than what it actually needs to be to store the data on chain, and I think that's an intentional barrier of entry into the pallet right now. I don't know if the deposits *need* to be this big because of some other economic reasons. My point is this: the storage cost should already be accounted for by the byte deposit, so the base deposit shouldn't depend on the `struct` size. It should instead be tied to some economic reason and if we can't find one, we can remove it or set it to `0` on our implementations and let downstream users decide if they want to add such a cost.

joepetrowski: I should have checked before commenting, but on Polkadot it's already configured [like I suggested](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/relay/polkadot/src/lib.rs#L665). On Kusama, [not](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/relay/kusama/src/lib.rs#L919).

By using `deposit`, it does actually come out to over 20 DOT:
```python
>>> def deposit(items, bytes):
...     return items * 20 * 10_000_000_000 + bytes * 100 * 100_000
... 
>>> deposit(1, 9) / 10**10
20.009
```

You aren't just paying for the size in bytes, you're paying for having a storage item at all, because more items increases the depth of the trie and therefore increases weights for all transactions since reads, writes, and trie hashing take longer.

For sure other users can configure this with a higher or lower `BasicDeposit` to subsidize or discourage its use, but I think the neutral config is to set it to `(1, min_size(Registration))`.

georgepisaltu: The `BasicDeposit` for identity is now `deposit(1, min_size(Registration) - min_size(IdentityInfo)`, because the bytes in the identity info are already accounted for in the byte deposit. I didn't replicate the change for the subs deposit because if we were to do it, that would be another 2 storage items: one entry in `SuperOf` per sub identity of an account, one entry in `SubsOf` per account which uses the sub-account functionality. On top of that, we would have to also charge per byte and the price for a sub-identity would end up being about twice the price of setting one identity. Right now, the price of a sub-identity is 1/5 of an identity. I'd prefer to keep `SubAccountDeposit` as is, let me know what you think.

substrate/frame/identity/src/legacy.rs
```diff
@@ -24,6 +24,9 @@ use sp_std::prelude::*;
 
 use crate::types::{Data, IdentityFields, IdentityInformationProvider, U64BitFlag};
 
+#[cfg(feature = "runtime-benchmarks")]
+const MAX_ADDITIONAL_FIELDS: usize = 2;
```
ggwpez, line 28: Why can we not use the correct maximum from the runtime?

substrate/frame/identity/src/lib.rs
```diff
@@ -97,6 +102,9 @@ type NegativeImbalanceOf<T> = <<T as Config>::Currency as Currency<
 >>::NegativeImbalance;
 type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
 
+/// Maximum size of an identity we can store is 7 KiB.
+const IDENTITY_MAX_SIZE: u32 = 7 * 1024;
```
ggwpez, line 106: Where does this come from and why is it private?  
is this statically enforced by testing that `Identity::max_encoded_len() <= IDENTITY_MAX_SIZE`?

georgepisaltu: Origin [here](https://github.com/paritytech/polkadot-sdk/pull/2048#discussion_r1373512069). In short, it's the same model as in [preimage](https://github.com/paritytech/polkadot-sdk/blob/9987bbb1d264672d406071fb2e164e4e80117e13/substrate/frame/preimage/src/lib.rs#L94) but adapted to what is currently stored in this pallet.

> why is it private?

I guess it could be public. My reasoning was that users choose what `struct` to use for their `IdentityInformationProvider` and this value of the max size is unreasonably large anyway. If it didn't fail the decoding, then we'll have a hard stop in the pallet regarding the size.

> is this statically enforced by testing that `Identity::max_encoded_len() <= IDENTITY_MAX_SIZE`?

If you mean in integrity test, then not yet but it's a very good idea.

substrate/frame/identity/src/legacy.rs
georgepisaltu: We surely can. I was getting rid of the `MaxAdditionalFields` instances from the configs and I forgot to get rid of this, good catch!

substrate/frame/identity/src/lib.rs
georgepisaltu: Done, PTAL

substrate/frame/identity/src/types.rs
```diff
@@ -241,17 +246,8 @@ pub trait IdentityInformationProvider:
 	/// Check if an identity registered information for some given `fields`.
 	fn has_identity(&self, fields: u64) -> bool;
```
bkchr, line 247:
```suggestion
	/// Check if an identity registered information for some given `fields`.
	fn has_identity(&self, fields: Self::IdentityField) -> bool;
```

`IdentityField` should get a better name.

substrate/frame/identity/src/types.rs
```diff
@@ -32,7 +35,9 @@ use sp_std::{fmt::Debug, iter::once, ops::Add, prelude::*};
 /// An identifier for a single name registrar/identity verification service.
 pub type RegistrarIndex = u32;
 
-pub trait U64BitFlag: BitFlag + RawBitFlags<Numeric = u64> {}
+pub trait U64BitFlag: BitFlag + RawBitFlags<Numeric = Self::NumericRepresentation> {
+	type NumericRepresentation: BitFlagNum + Into<u64> + TryFrom<u64> + Encode + Decode + TypeInfo;
+}
```
bkchr, line 40:
```suggestion
```

We don't need this.

substrate/frame/identity/src/lib.rs
```diff
@@ -669,26 +671,23 @@ pub mod pallet {
 		pub fn set_fields(
 			origin: OriginFor<T>,
 			#[pallet::compact] index: RegistrarIndex,
-			fields: IdentityFields<
-				<T::IdentityInformation as IdentityInformationProvider>::IdentityField,
-			>,
+			fields: <<T::IdentityInformation as IdentityInformationProvider>::IdentityField as U64BitFlag>::NumericRepresentation,
```
bkchr, line 674:
```suggestion
			fields: <T::IdentityInformation as IdentityInformationProvider>::IdentityField,
```

substrate/frame/identity/src/lib.rs
```diff
@@ -669,26 +671,23 @@ pub mod pallet {
...
+			let registrars =
+				<Registrars<T>>::mutate(|registrars| -> Result<usize, DispatchError> {
+					let registrar = registrars
+						.get_mut(index as usize)
+						.and_then(|r| r.as_mut())
+						.filter(|r| r.account == who)
+						.ok_or_else(|| DispatchError::from(Error::<T>::InvalidIndex))?;
+					registrar.fields = IdentityFields(
+						BitFlags::from_bits(fields).map_err(|_| Error::<T>::InvalidFields)?,
+					);
```
bkchr, line 687:
```suggestion
					registrar.fields = fields;
```

bkchr (commented):
Ty! Looking good, but still requires some improvements.

The identity fields stuff is still too much "married" to the old implementation. You are doing there some hops to let it look like some bitfield, while this isn't required. The `IndentityProvider` should just declare what kind of type this is and then it can also do internally the `has_identity` call. However, this will also require that this code: https://github.com/paritytech/polkadot-sdk/blob/9ca267328e0cba753aa528615cc19d1f633ed764/substrate/frame/alliance/src/lib.rs#L1084-L1107

is made generic. TBH, it doesn't make sense that `alliance` by default declares what kind of fields are required. This kind of configuration should be done in the runtime.

bkchr:
I know that this alliance stuff was there already before. However, we should do it correctly now.

substrate/frame/identity/src/lib.rs
```diff
@@ -357,7 +371,7 @@ pub mod pallet {
 			};
 
 			let old_deposit = id.deposit;
-			id.deposit = T::BasicDeposit::get() + fd;
+			id.deposit = T::BasicDeposit::get().saturating_add(byte_deposit);
```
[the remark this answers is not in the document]
joepetrowski: You still have to account for the storage item existing. This pattern is common:

- [Preimage](https://github.com/polkadot-fellows/runtimes/blob/7bd91e8/relay/polkadot/src/lib.rs#L229) (pre-`Consideration`)
- [Multisig](https://github.com/polkadot-fellows/runtimes/blob/7bd91e8/relay/polkadot/src/lib.rs#L917)
- [Proxy](https://github.com/polkadot-fellows/runtimes/blob/7bd91e8/relay/polkadot/src/lib.rs#L935)

substrate/frame/identity/src/lib.rs
```diff
@@ -97,6 +102,9 @@ type NegativeImbalanceOf<T> = <<T as Config>::Currency as Currency<
 >>::NegativeImbalance;
 type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
 
+/// Maximum size of an identity we can store is 7 KiB.
+pub const IDENTITY_MAX_SIZE: u32 = 7 * 1024;
```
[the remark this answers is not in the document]
georgepisaltu: Moved it to config.

substrate/frame/identity/src/types.rs
georgepisaltu: Removed it.

substrate/frame/identity/src/types.rs
georgepisaltu: @joepetrowski suggested `FieldsBitFlags` and I don't have a better idea :sweat_smile:

substrate/frame/alliance/src/lib.rs
```diff
@@ -135,9 +134,14 @@ type NegativeImbalanceOf<T, I> = <<T as Config<I>>::Currency as Currency<
...
 pub trait IdentityVerifier<AccountId> {
+	type FieldsBitFlags;
+
+	/// Returns the relevant identities that an account is required to have set.
+	fn required_identities() -> Self::FieldsBitFlags;
+
 	/// Function that returns whether an account has an identity registered with the identity
 	/// provider.
-	fn has_identity(who: &AccountId, fields: u64) -> bool;
+	fn has_identity(who: &AccountId, fields: Self::FieldsBitFlags) -> bool;
```
georgepisaltu, line 144: These 2 functions could be collapsed into `has_required_identities(who: &AccountId, fields: Self::FieldsBitFlags) -> bool` but I wasn't sure it's better than explicitly declaring the relevant fields in a standalone function as it is now. Let me know what you think.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4194481

substrate/frame/identity/src/types.rs
```diff
@@ -234,24 +231,25 @@ impl<Balance: Encode + Decode + MaxEncodedLen + Copy + Clone + Debug + Eq + Part
 pub trait IdentityInformationProvider:
 	Encode + Decode + MaxEncodedLen + Clone + Debug + Eq + PartialEq + TypeInfo
 {
-	/// Type capable of representing all of the fields present in the identity information as bit
-	/// flags in `u64` format.
-	type IdentityField: Clone + Debug + Eq + PartialEq + TypeInfo + U64BitFlag;
+	/// Type capable of representing all of the fields present in the identity information.
+	type FieldsBitFlags: Encode
```
georgepisaltu, line 235: `BitFlags` had the advantage of validating the input automatically and now we lost that. I was thinking we could add a function in this trait `fn is_valid(fields: Self::FieldsBitFlags) -> bool` to validate input so we don't end up with people setting fields that don't exist.

georgepisaltu:
> I know that this alliance stuff was there already before. However, we should do it correctly now.

In addition to the changes in `pallet-identity` to decouple from `BitFlags`, I also made the requested changes in `pallet-alliance`. @bkchr PTAL

substrate/frame/alliance/src/lib.rs
bkchr: Ha nice! I had the same idea :D And I think we should do it this way. No need to have these two functions!

substrate/frame/alliance/src/lib.rs
```diff
@@ -135,9 +134,14 @@ type NegativeImbalanceOf<T, I> = <<T as Config<I>>::Currency as Currency<
 
 /// Interface required for identity verification.
 pub trait IdentityVerifier<AccountId> {
+	type FieldsBitFlags;
```
bkchr, line 137: Then we can also get rid off this.

substrate/frame/identity/src/lib.rs
```diff
@@ -252,6 +258,8 @@ pub mod pallet {
 		JudgementForDifferentIdentity,
 		/// Error that occurs when there is an issue paying for judgement.
 		JudgementPaymentFailed,
+		/// Invalid identity fields.
+		InvalidFields,
```
bkchr, line 262: Not being used?

substrate/frame/identity/src/types.rs
```diff
@@ -234,24 +231,25 @@ impl<Balance: Encode + Decode + MaxEncodedLen + Copy + Clone + Debug + Eq + Part
...
+		+ Decode
+		+ Clone
+		+ Debug
+		+ Default
+		+ Eq
+		+ PartialEq
+		+ TypeInfo
+		+ MaxEncodedLen;
```
bkchr, line 243:
```suggestion
	/// Type capable of holding information on which fields are set.
	type FieldsInfo: sp_runtime::traits::Member + Default;
```

bkchr: Naming is hard and I still don't like my name. Maybe @joepetrowski has a better idea.

substrate/frame/identity/src/lib.rs
```diff
@@ -280,6 +288,16 @@ pub mod pallet {
 		SubIdentityRevoked { sub: T::AccountId, main: T::AccountId, deposit: BalanceOf<T> },
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn integrity_test() {
+			assert!(
+				T::IdentityInformation::max_encoded_len() <= T::MaxIdentitySize::get() as usize,
```
bkchr, line 295: This should always fail, because the `max_encoded_len` of the `IdentityInforrmation` would be with all fields set at their maximum.

substrate/frame/identity/src/lib.rs
```diff
@@ -129,10 +134,9 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxSubAccounts: Get<u32>;
 
-		/// Maximum number of additional fields that may be stored in an ID. Needed to bound the I/O
-		/// required to access an identity, but can be pretty high.
+		/// The maximum encoded size of an identity information.
 		#[pallet::constant]
-		type MaxAdditionalFields: Get<u32>;
+		type MaxIdentitySize: Get<u32>;
```
bkchr, line 139: Not sure we need this at all? Why have an upper bound here, if this upper bound is already implicitly declared by the `IdentityInformation`.

I don't think we need this.

bkchr: A chain that has 10 different identity fields, should also support that all of them are set.

bkchr (approved):
Thank you! Generally looking really good now! Left some last nitpicks, but overall good work!

substrate/frame/identity/src/lib.rs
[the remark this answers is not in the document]
ggwpez: We should probably be using storage `Consideration`, which abstracts the `num_bytes -> Balance` conversion.

substrate/frame/identity/src/types.rs
bkchr: Ahh just have seen: https://github.com/paritytech/polkadot-sdk/pull/2048#discussion_r1380637556

However, we should not include "BitFlags" in the name, because it could be everything (also may a vector for example).

joepetrowski: > However, we should not include "BitFlags" in the name, because it could be everything (also may a vector for example).

Yeah good point. `FieldsIdentifier`? This one is difficult to name.

substrate/frame/alliance/src/lib.rs
georgepisaltu: Done.

substrate/frame/alliance/src/lib.rs
georgepisaltu: Done.

substrate/frame/identity/src/lib.rs
```diff
@@ -97,6 +99,9 @@ type NegativeImbalanceOf<T> = <<T as Config>::Currency as Currency<
 >>::NegativeImbalance;
 type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
 
+/// Maximum size of an identity we can store is 7 KiB.
+pub const IDENTITY_MAX_SIZE: u32 = 7 * 1024;
```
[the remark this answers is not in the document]
georgepisaltu: Done.

substrate/frame/identity/src/lib.rs
georgepisaltu: Makes sense. It's gone.

substrate/frame/identity/src/lib.rs
georgepisaltu: Removed.

substrate/frame/identity/src/lib.rs
georgepisaltu: Gone now with the other change.

substrate/frame/identity/src/types.rs
georgepisaltu: Renamed it to `FieldsIdentifier`. Maybe we'll come up with a better name in the future :grin:

substrate/frame/identity/src/types.rs
```diff
@@ -234,24 +231,25 @@ impl<Balance: Encode + Decode + MaxEncodedLen + Copy + Clone + Debug + Eq + Part
...
-		0
-	}
+	fn has_identity(&self, fields: Self::FieldsBitFlags) -> bool;
+
+	#[cfg(feature = "runtime-benchmarks")]
+	fn create_identity_info() -> Self;
 
 	#[cfg(feature = "runtime-benchmarks")]
-	fn create_identity_info(num_fields: u32) -> Self;
+	fn all_fields() -> Self::FieldsBitFlags;
```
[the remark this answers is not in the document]
georgepisaltu: Done.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4205213

PR #2058, opened 2023-10-27
PVF: Add test instructions

mrcnski:
I was mentoring someone and figured I could copy/paste this info into a doc. This is especially needed now that we have multiple contributors. :) cc @Jpserrat @eagr


polkadot/node/core/pvf/README.md
```diff
@@ -0,0 +1,56 @@
+# PVF Host
+
+## Running basic tests
+
+The first step is to build the worker binaries:
+
+```sh
+SKIP_WASM_BUILD=1 cargo build --bin polkadot-execute-worker --bin polkadot-prepare-worker
```
mrcnski, line 8: Might not be needed soon after https://github.com/paritytech/polkadot-sdk/pull/2018?

polkadot/node/core/pvf/tests/README.md
mrcnski, line 1: Moved the README out of this subfolder to make it easier to find. Also, the new instructions are not just specific to the integration tests in this subfolder.

polkadot/node/core/pvf/README.md
eagr: Yep, running `cargo test` would do this automatically if the workers aren't already available or stale.

mrcnski: Nice, this just bit me again today where I forgot to rebuild them. I'll try to get to that PR. BTW, how do you tell if they're stale?

eagr: I was referring to those version checks. Thought they're used for that, are they? :)

mrcnski: Yes, you're right. :) What I had in mind is, when testing here we can make the version check more strict. It is not very strict usually (only enforcing the release version of the node), because we don't want to enforce frequent rebuilds for most development, where the PVF workers have not changed. But for these tests specifically, if someone is running them, there is a high likelihood that the workers are being developed and changing. With that in mind, we can additionally take into account the latest git commit hash and rebuild if it's changed. The workers should already have a `--full-version` flag which includes this. Do you think this is reasonable?

mrcnski: This would have saved me some time when I forgot to rebuild them [last week](https://github.com/paritytech/polkadot-sdk/pull/2058#discussion_r1374447111).

eagr: You mean to do a commit hash match instead when given this additinoal flag? Sounds good to me :)

mrcnski: Er, I think so. Just to be clear, in #2018 the new function that rebuilds the workers only runs in PVF tests, and there it makes sense to always include the commit hash in the check. So you can pass this flag to the workers when getting their version for the check.

eagr:
Two things about pvf have been a bit confusing for me.

The `*-intf.rs` file names. I guess the "intf" part is shorthand for "interface"?

The terminology precheck / prepare / pre-validate. I thought they refer to different process, as they are different words. But the more I read the code and docs, the more confusing I got. Sometimes, they seem to refer the same thing, sometimes not. Not sure any more :sweat_smile:.

mrcnski:
Thanks @eagr, that is valuable feedback!

I've previously documented the terms we use. I added some links to the existing docs that are hopefully helpful - let me know.

Yeah, `intf` is short for "interface". These modules are in charge of interfacing with the workers. To me the naming makes sense, but I can see it being confusing. Any suggestions for improving it? 🤔

eagr:
At first, I thought it defines the interfaces between components, that is, traits. Coz it's a common practice in languages with the `interface` feature. But they are not. So not only the name `intf` makes people wonder, even when you guess it right, it may give you a false assumption, which is not great.

I guess you don't want to just use `worker.rs` to avoid giving the assumption that it is a worker implementation, right? Maybe "work", in the sense of "work the worker" or "make it work". For comparison,

* worker-intf.rs
* worker-interface.rs
* worker.rs
* work.rs

s0me0ne-unkn0wn:
> I've previously documented the terms we use. I added some links to the existing docs that are hopefully helpful - let me know.

The doc is a bit ambiguous, though. (Not surprised, our docs are far from perfect).

> **PVF Prechecking:** This is the process of initially checking the PVF when it is first added.

One could read it as "the very first time the PVF appears on chain, that is, during parachain onboarding only" (because the term "first added" is not defined and thus is ambiguous; known relevant processes are onboarding, upgrading, acceptance, rejection, encation, but not "addition"). It is not clear that the process takes place every time a parachain upgrades its STF.

> As there is no prevalidation right now, preparation just consists of compilation.

Technically, that's not true. [There is](https://github.com/paritytech/polkadot-sdk/blob/2d9426f1cc144a0624ea0329ddc7e567bb47d6b2/polkadot/node/core/pvf/prepare-worker/src/executor_intf.rs#L24) a `prevalidate()` function that gets called before preparation and that currently instantiates the `RuntimeBlob` object deserializing raw `&[u8]` code into `parity_wasm::Module` structure. During the deserialization, a lot of checks are made, including section ordering, function/code section consistency, etc. So, there is still a prevalidation phase, although it's not perfect and not complete.

mrcnski:
> I guess you don't want to just use worker.rs to avoid giving the assumption that it is a worker implementation, right? Maybe "work", in the sense of "work the worker". For comparison,

Hmm, good suggestions, but I still see potential for confusion. People may see `worker.rs` or `work.rs` and think that code belongs to the workers. Maybe `worker-spawn` is better?

mrcnski:
> The doc is a bit ambiguous, though. (Not surprised, our docs are far from perfect).

You can always raise an issue or PR to improve them! The docs were even more unclear when I joined, so I tried to improve them based on my notes, but I didn't have a great understanding back then. Thanks for the feedback, I'll make those changes here.

> [There is](https://github.com/paritytech/polkadot-sdk/blob/2d9426f1cc144a0624ea0329ddc7e567bb47d6b2/polkadot/node/core/pvf/prepare-worker/src/executor_intf.rs#L24) a `prevalidate()` function that gets called before preparation and that currently instantiates the `RuntimeBlob` object deserializing raw `&[u8]` code into `parity_wasm::Module` structure.

I see, in that case the comment in `prevalidate` is misleading (it literally says "this function" followed by "do nothing for now"). I'll revise it.

eagr:
> Maybe worker-spawn is better?

If it were up to me, I'd probably just use worker.rs in the sense of related to worker and call it a day. Name is hard. :)

mrcnski:
It shouldn't be `worker.rs` because it's not worker code, but code that interfaces with workers. `worker-interface.rs` seems like the best alternative. There is maybe a potential for confusion with the concept of interfaces/traits, but it's still an improvement.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4395148

polkadot/node/core/pvf/README.md
```diff
@@ -0,0 +1,73 @@
...
+For example:
+
+```sh
+RUST_LOG=parachain::pvf=trace cargo test execute_can_run_serially
+```
+
+For more info on how our logs work, check [the
+docs](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/gum/src/lib.rs).
+
+## Running a test-network with zombienet
```
alexggh, line 34: Nit: Wouldn't this steps be better at the top of this repo ? And if we already have them can we just reference them, since they might change.

alexggh (approved):

s0me0ne-unkn0wn (approved):

polkadot/doc/testing.md
```diff
@@ -8,27 +8,59 @@ The testing strategy for Polkadot is 4-fold:
...
 ### Integration tests
 
-There are two variants of integration tests:
+There are the following variants of integration tests:
 
 #### Subsystem tests (2)
 
 One particular subsystem (subsystem under test) interacts with a mocked overseer that is made to assert incoming and
-outgoing messages of the subsystem under test. This is largely present today, but has some fragmentation in the evolved
-integration test implementation. A `proc-macro`/`macro_rules` would allow for more consistent implementation and
```
mrcnski, line 21: I removed this feature-ask because it should be in an issue, not in user-facing documentation.

polkadot/doc/testing.md
```diff
@@ -8,27 +8,59 @@ The testing strategy for Polkadot is 4-fold:
...
-structure.
+outgoing messages of the subsystem under test. See e.g. the `statement-distribution` tests.
 
 #### Behavior tests (3)
 
 Launching small scale networks, with multiple adversarial nodes without any further tooling required. This should
 include tests around the thresholds in order to evaluate the error handling once certain assumed invariants fail.
 
-For this purpose based on `AllSubsystems` and `proc-macro` `AllSubsystemsGen`.
```
mrcnski, line 29: I removed this stuff because I couldn't find any reference to `AllSubsystems` in the codebase.

mrcnski (commented):
This `testing.md` doc is pretty old and needs some love. I've never heard anyone mention Gurke or simnet, can that be removed? @sandreim

polkadot/doc/testing.md
mrcnski: Since this has been here for like three years, if it was an issue it would have been lost in the monorepo migration. So, I won't create it. :)

polkadot/doc/testing.md
```diff
@@ -8,27 +8,59 @@ The testing strategy for Polkadot is 4-fold:
...
+Currently, we commonly use **zombienet** to run mini test-networks locally on
+your own machine.
```
alindima, line 35: not only on local machines, but CI also

polkadot/doc/testing.md
```diff
@@ -8,27 +8,59 @@ The testing strategy for Polkadot is 4-fold:
...
 
 One particular subsystem (subsystem under test) interacts with a mocked overseer that is made to assert incoming and
-outgoing messages of the subsystem under test. This is largely present today, but has some fragmentation in the evolved
-integration test implementation. A `proc-macro`/`macro_rules` would allow for more consistent implementation and
-structure.
+outgoing messages of the subsystem under test. See e.g. the `statement-distribution` tests.
 
 #### Behavior tests (3)
 
 Launching small scale networks, with multiple adversarial nodes without any further tooling required. This should
```
alindima, line 31: I think this section was referring to some legacy way of testing, because it says without further tooling required. At least we should remove the "without further tooling" because we then say zombienet is the tool we use

polkadot/node/core/pvf/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+To run the PVF-specific zombienet test:
+
+```sh
+RUST_LOG=parachain::pvf=trace zombienet --provider=native spawn zombienet_tests/functional/0001-parachains-pvf.toml
+```
+
+## Testing on Linux
+
+Much of the PVF functionality, especially related to security, is Linux-only. If
```
alindima, line 36: and some of it x86-only

polkadot/node/core/pvf/README.md
```diff
@@ -0,0 +1,45 @@
...
+you touch anything security-related, make sure to test on Linux! If you're on a
+Mac, you can either run a VM or you can hire a VPS and use [EternalTerminal][et]
+to ssh into it. (ET preserves your session across disconnects, and unlike mosh
```
alindima, line 39:
```suggestion
to ssh into it. (ET preserves your session across disconnects, and unlike most shells
```

alindima: didn't know mosh was a tool, I thought it's a typo 😆

alindima (approved):

polkadot/node/core/pvf/README.md
mrcnski: Haha! I have this note here because et is still unknown compared to mosh, but either of these tools are *huge* improvements over ssh.

mrcnski:
I raised a follow-up for full resuscitation of testing.md. (I'm pretty sure the old stuff can just be deleted, but it seemed best to leave it out of scope of this PR).
https://github.com/paritytech/polkadot-sdk/issues/2527

PR #2065, opened 2023-10-27
Use workspace.dependenceis to ensure consistent versions of common crates

jsdw:
# Description

This uses the `[workspace.dependencies]` field to give some consistency over versions of common crates used all over the place and make keeping their versions in sync easier.

For now, I just started with two common crates, `scale-info` and `parity-scale-codec`. If this is something that's wanted, I can add some more common crates to the list, or else others can add to the list.

# Downsides

`parity-scale-codec` is renamed to `codec` in many places. Owing to a quirk of workwspace dependencies, it must be declared in the workspace as:

```toml
codec = { package = "parity-scale-codec", version = "3.6.5" }
```

Owing to how the macro discovers which crate path to use, it currently needs the `package` field to be duplicated as in:

```toml
codec = { package = "parity-scale-codec", workspace = true, ... }
```

This leads to a bunch of warnings. It could be fixed in `parity-scale-codec` with a bit of effort.

The other thing leading to warnings is that eg `default-features = false` should be added to the workspace dependencies, else it is ignored in the use site. Fixing any fallout from that would lead to a little churn perhaps. 

# Unknowns

Does this break the tooling to publish crates automatically, or any other CI workflows?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11265,9 +11265,9 @@ dependencies = [
 
 [[package]]
 name = "parity-scale-codec"
-version = "3.6.4"
+version = "3.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd8e946cc0cc711189c0b0249fb8b599cbeeab9784d83c415719368bb8d4ac64"
+checksum = "0dec8a8073036902368c2cdc0387e85ff9a37054d7e7c98e592145e0c92cd4fb"
 dependencies = [
  "arrayvec 0.7.4",
  "bitvec",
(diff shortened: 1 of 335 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4110719

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4111419

jsdw:
I thought I'd havea go at doing this, because it's generally worked well for us in Subxt (though the number of warnings is small and quite tolerable). I think to do it in `polkadot-sdk` would make consistyfing the versions nicer, but right now there are just too many warnings and things. I might look at fixing `parity-scale-codec` to remove the ones it leads to, but until then I don't think this is worthwhile

jsdw:
Note to anybody reading this; actually `parity-scale-codec` doesn't need the "package = 'parity-scale-codec'" everywhere any more once the next version is released, owing to https://github.com/paritytech/parity-scale-codec/pull/535. Yay!

PR #2066, opened 2023-10-27
Fix `ecdsa_bls` verify in BEEFY primitives

drskalman:
BEEFY ECDSA signatures are on keccak has of the messages. As such we can not simply call

`EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())`

because that invokes ecdsa default verification which perfoms blake2 hash which we don't want.

This bring up the second issue makes: This makes  `sign` and `verify` function in `pair_crypto` useless, at least for BEEFY use case. 
Moreover, there is no obvious clean way to generate the signature given that pair_crypto does not exposes `sign_prehashed`. You could in theory query the keystore for the pair (could you?), invoke `to_raw` and re-generate each sub-pair and sign using each. But that sounds extremely anticlimactic and will be frow upon by auditors . So I appreciate any alternative suggestion.


davxy:
> Moreover, there is no obvious clean way to generate the signature given that pair_crypto does not exposes sign_prehashed. You could in theory query the keystore for the pair (could you?), 

No you can't.

Maybe a solution could be to add some context specific parameters to the sign keystore call? Should be an opaque set of options which can then be used at sign time (like prehashed bool).

Otherwise I don't see any other way other than continue to bloat the Keystore interface with special methods. This also allows to remove the ecdsa "precast sign" method already present.

Don't know... this is just an idea. I need to check better the interface

drskalman:
I have a less intrusive proposal: We introduce a new trait `PrehashedSingerPair` which introduce  `PrehashedSingerPair::sign_prehashed` function. We implement it for `paired_crypto::Pair` only if its elements implement it as well. Then we implement it for `ecdsa` and `bls`, so we have it for `ecdsa_bls377`. Then similar to current `ecdsa_sign_prehashed` function we add `ecdsa_bls377_sign_prehashed` to the keystore.

There is only one cosmetic caveat here: BLS signature needs hash-to-curve (as opposed to hash-to-field) and that makes it much more of an elaborate process. That is why `w3f-bls` library currently does not exposes the hasher choice (because nobody is going to use it as usually there is only one IETF standardized hasher for each curve). Nonetheless, we could just fake `bls::Pair<T>::sign_prehashed` to call `bls:Pair<T>::sign` if nobody gets deeply hurt and we are good to go.

@davxy: let me know what you think?

drskalman:
we can call it `PrehashedSingerPair::maybe_sign_prehashed` it makes people happier :-)

davxy:
I thinked a bit about this.

So, for the keystore we need to follow a route similar to what has been done for ecdsa.
I.e. provide an `ecdsa_bls377_sign_prehashed`  that in the implementation does something like [`ecdsa`](https://github.com/paritytech/polkadot-sdk/blob/ad5163ba93a7e142fd6fb2b9ed975bff9c29c479/substrate/primitives/keystore/src/testing.rs#L215).

For what concerns the way the `sign_prehashed` is provided by `ecdsa-bls12-377` well here we have two routes:
1. define this `sign_prehashed` only for the `Pair` specialized for ecdsa_bls12-377 (i.e. within its module https://github.com/paritytech/polkadot-sdk/blob/ad5163ba93a7e142fd6fb2b9ed975bff9c29c479/substrate/primitives/core/src/paired_crypto.rs#L41). 

2. Be more generic and introduce this trait you are talking about. And implement it for ecdsa, bls and for any paired crypto where the "sub-schemes" implements it.

Solution 1 is more quick and dirty.
Solution 2 is more generic

Going with 1 doesn't preclude to implement 2 later... For me is the same. 

In case you want to go with 2 please add the traits to crypto.rs close to https://github.com/paritytech/polkadot-sdk/blob/ad5163ba93a7e142fd6fb2b9ed975bff9c29c479/substrate/primitives/core/src/crypto.rs#L1116-L1128. Something like:
- `PrehashedSigner`
- `PrehashedVerifier` (yeah introducing the verifier has a more clean design)

drskalman:
> Rust is happy with multiple `impl` blocks.
> 
You are right :-) maybe it is re-implementation after specifying an associated type which doesn't work. 

> The one that is outside ecdsa-bls12-377 module is empty BTW :-D so can be removed
I think you have already removed it. I was pointing to a wrong copy of the code :-)

substrate/primitives/core/src/paired_crypto.rs
```diff
@@ -617,6 +675,16 @@ mod test {
 		assert_eq!(cmp, public);
 	}
 
+	#[test]
+	fn sign_and_verify_with_haser_works() {
```
davxy, line 679:
```suggestion
	fn sign_and_verify_with_hasher_works() {
```
nit

substrate/primitives/consensus/beefy/src/lib.rs
```diff
@@ -257,6 +260,7 @@ pub enum ConsensusLog<AuthorityId: Codec> {
 ///
 /// A vote message is a direct vote created by a BEEFY node on every voting round
 /// and is gossiped to its peers.
+/// TODO: Remove `Signature` generic type, instead get it from `Id::Signature`.
```
davxy, line 263:
```suggestion
// TODO: Remove `Signature` generic type, instead get it from `Id::Signature`.
```
Hide TODOs from docs

substrate/primitives/core/src/paired_crypto.rs
```diff
@@ -71,6 +75,60 @@ pub mod ecdsa_bls377 {
...
+
+	#[cfg(feature = "full_crypto")]
+	impl Pair {
+		/// hashes the `message` with the specified `MsgHasher` and then signs it using ECDSA
+		/// algorithm. It does not affect the behavoir of BLS12-377 component. It generates
+		/// BLS12-377 Signature according to IETF standard and disregard the hasher for the
+		/// BLS12-377 component
+		pub fn sign_with_hasher<MsgHasher: crate::Hasher>(&self, message: &[u8]) -> Signature
+		where
+			<MsgHasher as crate::Hasher>::Out: Into<[u8; 32]>,
```
davxy, line 87: Why a generic `Hasher`? Isn't this always Keccak?

I ask because in the `Keystore` you will have to expose a function which is not generic over the hasher.
(any `Keystore` implementation should be allowed to be a dyn object, thus generic functions are not supported)

drskalman: So I can't do sign_prehashed because BLS standard doesn't support that. 

It is only `Keccak` for now but 1. [this trait](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/consensus/beefy/src/lib.rs#L59) requires to support generic hasher. 2. If you want to bridge to bitcoin probably you might want sha256 for example.

We could hard code the keccak in the keystore right? it is better to postpone non-generic-ness as late as possible, no?

davxy: Yeah sure.

davxy (approved):
🟢  excluding the trivial nitpicks

substrate/primitives/consensus/beefy/src/lib.rs
```diff
@@ -172,12 +172,15 @@ pub mod ecdsa_bls_crypto {
...
-			// `w3f-bls` library uses IETF hashing standard and as such does not exposes
-			// a choice of hash to field function.
-			// We are directly calling into the library to avoid introducing new host call.
-			// and because BeefyAuthorityId::verify is being called in the runtime so we don't have
-
-			EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())
+			// We can not simply call
+			// `EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())`
+			// because that invokes ecdsa default verification which perfoms blake2 hash
+			// which we don't want.
```
Lederstrumpf, line 178: to avoid planting calcified assumptions & since what is wanted may change in future, it would help readers without context to concretely explain why this is undesired (something as simple as "but this is incompatible since we use keccak when signing").

Lederstrumpf (approved):
ditto, 🟢 excluding the trivial nitpicks

substrate/primitives/consensus/beefy/src/lib.rs
drskalman: I assume it was because Keccak ends up being cheaper. I looked around a bit and my understanding is that, only the  compression function of Blake2b is [included](https://github.com/ethereum/EIPs/issues/152) as precompiled and you need to iterate over your data in order to get all of it hashed. [This paper](https://www.scitepress.org/Papers/2021/106066/106066.pdf) claims that blake2b is faster but more expensive: "A user would essentially be charged three times or more for submitting a Blake2 transaction in comparison with a Keccak transaction, despite the effective hash real lower cost in execution. that is does not end up cheap comparing to Keccak."
However, I haven't made this choice nor I'm an expert on Ethereum, I think someone from Snowfork team perhaps is in better position to add the clarification here.

substrate/primitives/consensus/beefy/src/lib.rs
```diff
@@ -172,12 +172,22 @@ pub mod ecdsa_bls_crypto {
...
+			// `EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())`
+			// because that invokes ecdsa default verification which perfoms blake2 hash
+			// which we don't want. This is because ECDSA signatures are meant to be verified
+			// on Ethereum network where Keccak hasher is siginificantly cheaper than Blake2b.
+			// See Figure 3 of [OnSc21](https://www.scitepress.org/Papers/2021/106066/106066.pdf)
+			// for comparison.
+			//
+			// - [OnSc21] Onica, E., & Schifirneţ, C. (2021). Towards efficient hashing in
+			// ethereum smart contracts. Proceedings of the 16th International
+			// Conference on Software Technologies, ()
```
davxy, line 185:
```suggestion
```
We already have the link above to the publication

substrate/primitives/consensus/beefy/src/lib.rs
```diff
@@ -172,12 +172,22 @@ pub mod ecdsa_bls_crypto {
...
-			// and because BeefyAuthorityId::verify is being called in the runtime so we don't have
-
-			EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())
+			// We can not simply call
+			// `EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())`
+			// because that invokes ecdsa default verification which perfoms blake2 hash
+			// which we don't want. This is because ECDSA signatures are meant to be verified
+			// on Ethereum network where Keccak hasher is siginificantly cheaper than Blake2b.
+			// See Figure 3 of [OnSc21](https://www.scitepress.org/Papers/2021/106066/106066.pdf)
+			// for comparison.
```
davxy, line 181:
```suggestion
			// We can not simply call
			// `EcdsaBlsPair::verify(signature.as_inner_ref(), msg, self.as_inner_ref())`
			// because that invokes ECDSA default verification which perfoms Blake2b hash
			// which we don't want. This is because ECDSA signatures are meant to be verified
			// on Ethereum network where Keccak hasher is significantly cheaper than Blake2b.
			// See Figure 3 of [OnSc21](https://www.scitepress.org/Papers/2021/106066/106066.pdf)
			// for comparison.
```

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/8

Issue #2013, opened 2023-10-24
`bandersnatch_vrfs` can cause dependency mess up


ggwpez:
We can currently not enable the `std` feature of the `bandersnatch_vrfs` dependency of `sp-core` [here](https://github.com/paritytech/polkadot-sdk/blob/f3bf5c1acdccf953774ee055a8a0465e1e2a3475/substrate/primitives/core/Cargo.toml#L78).  

If i add it, then it cannot build `sp-runtime-interface` anymore. The last time i tried it was a different build error about a duplicate crate.  
```pre
$ cargo c -p sp-runtime-interface

error: no matching package named `sp-ark-bls12-381` found
location searched: https://github.com/paritytech/ark-substrate
required by package `sp-bls12-381 v0.4.1-beta (https://github.com/paritytech/substrate-curves#60c204f3)`
    ... which satisfies git dependency `sp-bls12-381` of package `bandersnatch_vrfs v0.0.1 (https://github.com/w3f/ring-vrf?rev=4b09416#4b09416f)`
    ... which satisfies git dependency `bandersnatch_vrfs` (locked to 0.0.1) of package `sp-core v21.0.0 (/media/vados/980Pro/work/polkadot-sdk/substrate/primitives/core)`
    ... which satisfies path dependency `sp-core` (locked to 21.0.0) of package `asset-hub-kusama-runtime v0.9.420 (/media/vados/980Pro/work/polkadot-sdk/cumulus/parachains/runtimes/assets/asset-hub-kusama)`
    ... which satisfies path dependency `asset-hub-kusama-runtime` (locked to 0.9.420) of package `integration-tests-common v1.0.0 (/media/vados/980Pro/work/polkadot-sdk/cumulus/parachains/integration-tests/emulated/common)`
    ... which satisfies path dependency `integration-tests-common` (locked to 1.0.0) of package `asset-hub-rococo-integration-tests v1.0.0 (/media/vados/980Pro/work/polkadot-sdk/cumulus/parachains/integration-tests/emulated/assets/asset-hub-rococo)`
```

I think it would help if the `bandersnatch_vrfs` was pushed to crates-io instead of using git. Also AFAIK there is a cyclic dependency between `bandersnatch`  and `sp-core`.  
[...]

PR #2068, opened 2023-10-27
Elliptic curves utilities refactory

davxy:
- Usage the new published [arkworks-extensions](https://github.com/paritytech/arkworks-extensions) crates. 
  Hooks are internally defined to jump into the proper host functions.
- Conditional compilation of each curve (gated by feature with curve name)
- Separation in smaller host functions sets, divided by curve (fits nicely with prev point)

Runtime dev should not care that much about host functions and hooks as these are internally defined and he
can use the library as it'd use Arkworks upstream in his runtimes.

---

Please be aware that this library is mostly designed for internal usage within crates residing in the `polkadot-sdk` workspace, such as Bandersnatch. This is to avoid potential problems, as demonstrated in issues like https://github.com/paritytech/polkadot-sdk/issues/2013.

Devs of pallets outside the workspace can utilize this library or the stand alone crates available on `crates.io` and maintained in [arkworks-substrate](https://github.com/paritytech/arkworks-substrate) repository if they like.
Stand alone crates essentially offer the same functionality as the curve modules found here.


michalkucharczyk (approved):
I did not dive into the logic, implementation details.
Overall it looks good.

I did not notice any test, maybe we should have some, (at least for utils)?

substrate/primitives/crypto/ec-utils/src/lib.rs
```diff
@@ -15,272 +15,39 @@
...
+//! The definitions make use of host functions to offload the non-native
+//! computational environment from the some of the most computationally
+//! expensive operations by internally leveraging the
+//! [arkworks-extensions](https://github.com/paritytech/arkworks-extensions)
+//! library.
+//!
+//! The exported types are organized and named in a way that mirrors the structure
+//! of the types in the original Arkworks library. This design choice aims to make
+//! it easier for users already familiar with the library to understand and utilize
+//! the exported types effectively.
```
michalkucharczyk, line 30: Do I understand correctly that this crate is mainly intended to use in runtime, and it provides some abstraction layer which allows to call the Elliptic curves functionality which is delegated to host?

davxy: Yes correct.

Allows to use stuff in arkworks ZK stuff in the runtime.
Our types are like upstream ones but jumps into host functions when required.

substrate/primitives/crypto/ec-utils/src/lib.rs
```diff
@@ -15,272 +15,39 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Elliptic Curves host functions which may be used to handle some of the *Arkworks*
-//! computationally expensive operations.
+//! Elliptic curves which are mostly compatible with *Arkworks* library
+//! mostly useful in non-native contexts.
```
michalkucharczyk, line 19: maybe?
```suggestion
//! This crate offers elliptic curves functionality which is mostly compatible with *Arkworks* library.
//! It is primarily designed for use in non-native contexts.
```

also wouid be nice to have link to Arkworks.

substrate/primitives/crypto/ec-utils/src/lib.rs
```diff
@@ -15,272 +15,39 @@
...
+//!
+//! The definitions make use of host functions to offload the non-native
+//! computational environment from the some of the most computationally
+//! expensive operations by internally leveraging the
+//! [arkworks-extensions](https://github.com/paritytech/arkworks-extensions)
+//! library.
```
michalkucharczyk, line 25:
```suggestion
//! The implementation make use of host functions to offload the non-native
//! environment from some of the most computationally
//! expensive operations. It internally leverages the
//! [arkworks-extensions](https://github.com/paritytech/arkworks-extensions)
//! library.
```

davxy:
> So these crates here are just a verbatim copy?

@bkchr Not really verbatim (but in practice, at the moment, yes). Here we leverage some tiny `encode/decode` shared functions.

After this is merged [arkworks-substrate](https://github.com/paritytech/arkworks-substrate) will be refactored to just use the hooks and the stuff we are implementing here. So the crates there will act as some kind of convenience facade.

The arkworks-substrate crates are handy if someone wants to include the single crates individually and thus more closely mimic what you'd do when using upstream Arkworks.

skunert (approved):

Issue #994, opened 2023-04-08
Inherit workspace dependency


ggwpez:
Try out what Kilt did and use workspace dependencies to ease the version managment.  
Example [Cargo.toml](https://github.com/KILTprotocol/kilt-node/blob/eeb90d1045dc4157b3713beb8e6c7667154fe166/runtimes/common/Cargo.toml#L41) and [Workspace](https://github.com/KILTprotocol/kilt-node/blob/eeb90d1045dc4157b3713beb8e6c7667154fe166/Cargo.toml#L25).  

quote @bkchr 
>I would say we start with some like log etc. Deps that are shared between most crates and see how it goes. 

Any known downsides with this @weichweich?

ggwpez:
Done

PR #2070, opened 2023-10-27
Lift dependencies to the workspace (Part 1)

ggwpez:
Changes (partial https://github.com/paritytech/polkadot-sdk/issues/994):
- Set log to `0.4.20` everywhere
- Lift `log` to the workspace

Starting with a simpler one after seeing https://github.com/paritytech/polkadot-sdk/pull/2065 from @jsdw.  
This sets the `default-features` to `false` in the root and then overwrites that in each create to its original value. This is necessary since otherwise the `default` features are additive and its impossible to disable them in the crate again once they are enabled in the workspace.

I am using a tool to do this, so its mostly a test to see that it works as expected.


cumulus/client/consensus/common/Cargo.toml
```diff
@@ -10,7 +10,7 @@ async-trait = "0.1.73"
 codec = { package = "parity-scale-codec", version = "3.0.0", features = [ "derive" ] }
 dyn-clone = "1.0.12"
 futures = "0.3.28"
-log = "0.4.20"
+log = { workspace = true, default-features = true }
```
bkchr, line 13:
```suggestion
log = { workspace = true }
```

This and all the other places you have done this "feel weird"

bkchr (commented):
We should not disable the default features in the workspace. IMO this "feels" weird. 

And BTW, the `wasm-builder` currently not supports workspace deps.

Cargo.toml
```diff
@@ -4,6 +4,9 @@ edition = "2021"
 repository = "https://github.com/paritytech/polkadot-sdk.git"
 license = "GPL-3.0-only"
 
+[workspace.dependencies]
+log = { version = "^0.4.20", default-features = false }
```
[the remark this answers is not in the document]
ggwpez: I saw that if the default-features are not disabled in the workspace, then they cannot be disabled lateron in the crates, since features are additive.  
So AFAIK it is best practice to disable them at the workspace and then re-enable at the crate level - if desired. It is explained here https://github.com/rust-lang/cargo/issues/11329

cumulus/client/consensus/common/Cargo.toml
ggwpez: Yea I think the `default-features = true` should be implicit in this case, so we should be able to omit it.

ggwpez:
> And BTW, the `wasm-builder` currently not supports workspace deps.

What does this mean? That we cannot use it for the no-std runtime dependencies?

cumulus/client/consensus/common/Cargo.toml
koute: :+1: 

I think it makes sense to do `default-features = false` in the workspace `Cargo.toml` *on a case-by-case basis* if we actually don't want to always enable all of the features for a given dependency, but this probably shouldn't be done by default (at least not blindly), and if you're adding a `default-features = true` all over the place then it's most likely the wrong thing to do.

koute: And in case of the `log` [it looks like it doesn't even have any default features](https://github.com/rust-lang/log/blob/master/Cargo.toml).

ggwpez: >And in case of the log [it looks like it doesn't even have any default features](https://github.com/rust-lang/log/blob/master/Cargo.toml).

Hm. Then we would need to check on every version update whether some default features were added? I guess that could work since the WASM build should fail when it tries to pull in std.

>and if you're adding a `default-features = true` all over the place then it's most likely the wrong thing to do. 

Yes I agree if this is the case in 100% of the occurences (and under the assumption that this would not change in the future). But for a dependency where we sometimes want default-features and sometimes not, then we need to set it to false in the workspace i think.  

> If a crate needs a specific feature it should explicitly say which one, instead of doing ` default-features = true`  

If this is the goal, then i dont think that it is automatable.

bkchr:
> > And BTW, the `wasm-builder` currently not supports workspace deps.
> 
> What does this mean? That we cannot use it for the no-std runtime dependencies? Then i will continue with other deps after this PoC is done.

Anything crate that is using wasm builder will not work right now with a workspace dependency. We just need to add support for this.

cumulus/client/consensus/common/Cargo.toml
bkchr: > But for a dependency where we sometimes want default-features and sometimes not, then we need to set it to false in the workspace i think.

I don't think so. Cargo unifies features and the same will probably happen here. But for things crates that we use in no_std and somewhere only in std, we should Kot disable the default features in the workspace. IMO.

We should read up on how feature setting works with workspace deps. Especially around default features.

ggwpez: I disabled the default features in the workspace after reading this: https://github.com/rust-lang/cargo/issues/11329  
And the unification does not apply when crates are build on their own, or?  
I can try to find some more resources around the topic, but will probably close the MR then for now.

bkchr: What a fun :heart:

jsdw: I ran into a similar thing when I tried this; you need to specify `default-features = false` in the worksapce in order to be able to use `default-features` on that dependency in the crate's Cargo.toml, else you get a warning like:

```
warning: ..polkadot-sdk/polkadot/xcm/Cargo.toml: `default-features` is ignored for scale-info, since `default-features` was not specified for `workspace.dependencies.scale-info`, this could become a hard error in the future
```

bkchr: Yes, this is what @ggwpez linked above.

ggwpez: So can we go ahead with this:  
- `default-features = false` in the root workspace
- nothing in the crates if the default-features should be true
- `default-features = false` in each crate like before when they should be disabled.

This should make the diff here a bit shorter.

bkchr: If that works, lets do this.

ggwpez:
> Anything crate that is using wasm builder will not work right now with a workspace dependency. We just need to add support for this.

How to do this? I see that all CI checks are green. It should catch that or not @bkchr?

bkchr:
> > Anything crate that is using wasm builder will not work right now with a workspace dependency. We just need to add support for this.
> 
> How to do this? I see that all CI checks are green. It should catch that or not @bkchr?

Thinking again about this, my comment wasn't that correct. As we include the real runtime crate as dependency, it is fetching its deps from its original workspace. So yeah, just continue to merge this pr!

substrate/client/consensus/common/Cargo.toml
```diff
@@ -20,7 +20,7 @@ async-trait = "0.1.74"
 futures = { version = "0.3.21", features = ["thread-pool"] }
 futures-timer = "3.0.1"
 libp2p-identity = { version = "0.1.3", features = ["ed25519", "peerid"] }
-log = "0.4.17"
+log = { workspace = true }
```
bkchr, line 23: In the client crates we should enable `std` by default?

ggwpez: Ah yes, i thought that just writing this would enable them again, but it does not.  
Changed it now.

ggwpez:
@koute i guess you are quite busy, but i settled on this idiomatic pattern now:  
- put `default-features = false` in the workspace
- re-enable them per-crate with `default-features = true` if they were previously enabled and leave them disabled otherwise.  

I think this pattern is applicable to every dependency, so we dont have to think much and just use some tool to replace it.

bkchr (approved):

koute (approved):
Looks mostly good to me, although in this specific case of `log` all of the `default-features` are pretty much useless since the crate doesn't actually have any default features in the first place, so it's just noise that will confuse people.

So I would suggest that we just use a bare `log = { workspace = true }` everywhere.

svyatonik (approved):

ggwpez:
> Looks mostly good to me, although in this specific case of `log` all of the `default-features` are pretty much useless since the crate doesn't actually have any default features in the first place, so it's just noise that will confuse people.
> 
> So I would suggest that we just use a bare `log = { workspace = true }` everywhere.

Yes this could be done for *now*. But if we ever update `log` and it gets default features, then we would need to change it again.  
And at that point we already lost the information in which crates we would like to enable them, since we removed the annotation.

koute:
> Yes this could be done for _now_. But if we ever update `log` and it gets default features, then we would need to change it again.  And at that point we already lost the information in which crates we would like to enable them, since we removed the annotation.

I don't see a problem with this, considering any changes to the default features will have to be semver-compatible anyway.

And the "lost the information in which crates we would like to enable them" is, well, complete nonsense considering we're talking about features which *don't exist yet*. (:

Just because a given crate might enable some kind of a feature by default has nothing to do with whether we want to enable or disable it. And there is also no rule that in the client we should enable all of the features by default and in the runtime disable them. A given feature's value lies in what exactly that feature does, and not in whether it's enabled by default or not.

ggwpez:
> I don't see a problem with this, considering any changes to the default features will have to be semver-compatible anyway.

Oh right. So it should not make a difference anyway until there is a major bump 🙃

>And the "lost the information in which crates we would like to enable them" is, well, complete nonsense considering we're talking about features which don't exist yet. (: How do you know we want to enable a feature if we don't know what that feature is?
Just because a given crate might enable some kind of a feature by default has nothing to do with whether we want to enable or disable it. And there is also no rule that in the client we should enable all of the features by default and in the runtime disable them. A given feature's value lies in what exactly that feature does, and not in whether it's enabled by default or not.

Yea true. I only thought about our convention of treating default-features like `std`.  
For the sake of this MR i think its still fine to do, since it only transforms the existing code into a more readable format, without making assumptions about whether that code is correct or not. Something that i plan to repeat for more dependencies.

ggwpez:
I changed the default logic now to not write `default-features = false` if they are enabled everywhere.  
This resulted in a more readable diff here: https://github.com/paritytech/polkadot-sdk/pull/3366

PR #2071, opened 2023-10-27
Bulk Assigner Compile Fixes

BradleyOlson64:
Many small fixes to remove compile errors in `assigner_bulk`. Also creates test files and stubs out a few tests to be filled in once `assigner_bulk` core functionality is complete.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4112656

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4113569

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4115628

polkadot/runtime/parachains/src/assigner_bulk/mod.rs
```diff
@@ -241,37 +249,43 @@ impl<T: Config> AssignmentProvider<BlockNumberFor<T>> for Pallet<T> {
...
-			}
-
-			match a_type {
-				CoreAssignment::Idle => return None,
-				CoreAssignment::Pool =>
-					return <assigner_on_demand::Pallet<T> as AssignmentProvider<
-						BlockNumberFor<T>,
-					>>::pop_assignment_for_core(core_idx),
-				CoreAssignment::Task(para_id) => return Some(BulkAssignment::Bulk(para_id)),
+			match core_state {
```
eskimor, line 252: We can avoid all of this code being nested by short circuiting the function on `None`.

eskimor: Actually the correct solution is, that this should be a `ValueQuery` not an `OptionQuery`.

eskimor (commented):
Thanks @BradleyOlson64 !

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4134760

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4134886

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -666,12 +666,6 @@ mod enter {
 			// * 3 disputes.
 			assert_eq!(limit_inherent_data.disputes.len(), 2);
 
-			assert_ok!(Pallet::<Test>::enter(
-				frame_system::RawOrigin::None.into(),
-				limit_inherent_data,
-			));
-
-			// TODO [now]: this assertion fails with async backing runtime.
```
eskimor, line 674: Why is it fine to remove this assert?

BradleyOlson64: My understanding is that each `ParachainsInherentData` is intended to be filtered by `create_inherent_inner` and then entered using `enter`. 

But both of these call `process_inherent_data` which makes a bunch of state changes within the inclusion and scheduler pallets regardless of calling context. Those state changes were prematurely draining assignments from the mock assigner, causing the test to fail.

I'm not clear on why or whether we need to call `create_inherent` and `enter` from the same runtime instance. But if that is desirable then `process_inherent_data` needs to be modified and the test assertion can be left as is.

PR #2072, opened 2023-10-28
Change pallet referenda TracksInfo::tracks to return an iterator

olanod:
Returning an iterator in `TracksInfo::tracks()` instead of a static slice allows for more flexible implementations of `TracksInfo` that can use the chain storage without compromising a lot on the performance/memory penalty if we were to return an owned `Vec` instead.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4114554

substrate/frame/referenda/src/types.rs
```diff
@@ -144,15 +148,18 @@ pub trait TracksInfo<Balance, Moment> {
 	/// The origin type from which a track is implied.
 	type RuntimeOrigin;
 
-	/// Return the array of known tracks and their information.
-	fn tracks() -> &'static [(Self::Id, TrackInfo<Balance, Moment>)];
+	/// The iterator returned by `Self::tracks()` to lazily traverse the available tracks
+	type TracksIter: Iterator<Item = (Self::Id, TrackInfo<Balance, Moment>)>;
```
bkchr, line 152: Having something that may returns an iterator or slice would may be better

olanod: Mmm you mean return something like `Either<&'static [(_, _)], Self::TracksIter>`? it feels a bit ugly for users to handle the variants compared to this more natural way. Let me try something like returning an `IntoIterator` there the `Item` can be either an owned or a borrowed value to not lose the benefit of not having to clone/move the `TrackInfo`

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4123790

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4123802

olanod:
*NOTE:* I realized that [return_position_impl_trait_in_trait](https://github.com/rust-lang/rust/pull/115822) was recently stabilized, I think it's worth using it in this feature as it makes usage of the trait much simpler. Please consider updating the rust toolchain :)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4140709

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4140740

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4267563

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4268808

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4269881

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4293199

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4293745

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4293857

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4293888

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4297216

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4297185

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4297328

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4349897

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4350844

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5281974

cumulus/parachains/runtimes/collectives/collectives-westend/src/fellowship/tracks.rs
```diff
@@ -114,211 +116,213 @@ pub struct TracksInfo;
...
 					confirm_period: 30 * MINUTES,
 					min_enactment_period: 5 * MINUTES,
-					min_approval: pallet_referenda::Curve::LinearDecreasing {
+					min_approval: Curve::LinearDecreasing {
 						length: Perbill::from_percent(100),
 						floor: Perbill::from_percent(50),
 						ceil: Perbill::from_percent(100),
 					},
-					min_support: pallet_referenda::Curve::LinearDecreasing {
+					min_support: Curve::LinearDecreasing {
```
ggwpez, line 271: Just as a tip: there are a lot of noisy changes in this MR - it can scare people of to take a look.

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -784,8 +784,8 @@ impl Get<Option<BalancingConfig>> for OffchainRandomBalancing {
 			max => {
 				let seed = sp_io::offchain::random_seed();
 				let random = <u32>::decode(&mut TrailingZeroInput::new(&seed))
-					.expect("input is padded with zeroes; qed") %
-					max.saturating_add(1);
+					.expect("input is padded with zeroes; qed")
+					% max.saturating_add(1);
```
ggwpez, line 788: Please keep the formatting if nothing changed.  
It otherwise just extends the review (and if needed audit) times.

cumulus/parachains/runtimes/collectives/collectives-polkadot/src/ambassador/tracks.rs
```diff
@@ -0,0 +1,286 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The Ambassador Program's referenda voting tracks.
+
+use super::Origin;
+use crate::{Balance, BlockNumber, RuntimeOrigin, DAYS, DOLLARS, HOURS};
+use pallet_referenda::{StaticTracksIter, Track};
```
ggwpez, line 20: I dont find the definition of `StaticTracksIter` anywhere?!

polkadot/runtime/rococo/src/governance/fellowship.rs
```diff
@@ -32,12 +37,13 @@ pub struct TracksInfo;
...
 	type Id = u16;
 	type RuntimeOrigin = <RuntimeOrigin as frame_support::traits::OriginTrait>::PalletsOrigin;
-	fn tracks() -> &'static [(Self::Id, pallet_referenda::TrackInfo<Balance, BlockNumber>)] {
-		static DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 10] = [
-			(
-				0u16,
-				pallet_referenda::TrackInfo {
-					name: "candidates",
+
+	fn tracks() -> impl Iterator<Item = Cow<'static, Track<Self::Id, Balance, BlockNumber>>> {
```
ggwpez, line 41: Could it be `RandomAccessIterator` by any chance? Dont know if that is possible, but could be better in cases where a single one needs to be accessed.

ggwpez (commented):
Plase check the CI output. It shows what is red, currently it does not seem to compile, so i cannot review if it is correct.

cumulus/parachains/runtimes/collectives/collectives-polkadot/src/ambassador/tracks.rs
olanod: Leftover of a previous version, will ✂️

substrate/bin/node/runtime/src/lib.rs
olanod: My bad, format on-save is a bit sneaky. Will revert the auto formatting.

cumulus/parachains/runtimes/collectives/collectives-westend/src/fellowship/tracks.rs
olanod: Will make it less noisy, thanks for having a look :)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5309360

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5309989

polkadot/runtime/rococo/src/governance/fellowship.rs
olanod: To access a particular track there is the `TracksInfo::info(id)` method, it's implemented by default iterating over the tracks to `.find` the right one but it can be overwritten. In our ["dynamic tracks" implementation](https://github.com/virto-network/polkadot-sdk/blob/release-virto-v1.5.0/substrate/frame/referenda-tracks/src/impl_tracks_info.rs#L19) for example we replace it with an access to the particular track in the storage.

ggwpez: Okay 👍  so more performant implementations could make use of that still.

substrate/frame/referenda/src/types.rs
```diff
@@ -136,29 +138,44 @@ pub struct TrackInfo<Balance, Moment> {
...
-		let maybe_index = tracks.binary_search_by_key(&id, |t| t.0).ok()?;
+	/// Return the list of identifiers of the known tracks.
+	fn tracks_ids() -> Vec<Self::Id> {
+		Self::tracks().map(|x| x.id).collect()
+	}
 
-		tracks.get(maybe_index).map(|(_, info)| info)
+	/// Return the track info for track `id`, by default this just looks it up in `Self::tracks()`.
+	fn info(id: Self::Id) -> Option<Cow<'static, TrackInfo<Balance, Moment, N>>> {
+		Self::tracks().into_iter().find(|x| x.id == id).map(|t| match t {
```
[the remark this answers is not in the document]
olanod: It will depend on the underlying iterator, current implementations of the trait are using the slice iterator which I assume is ordered.

substrate/frame/referenda/src/types.rs
```diff
@@ -167,11 +184,24 @@ pub trait TracksInfo<Balance, Moment> {
...
 	{
-		if Self::tracks().windows(2).all(|w| w[0].0 < w[1].0) {
-			Ok(())
-		} else {
-			Err("The tracks that were returned by `tracks` were not sorted by `Id`")
-		}
+		use core::cmp::Ordering;
+		// Adapted from Iterator::is_sorted implementation available in nightly
+		// https://github.com/rust-lang/rust/issues/53485
+		let mut iter = Self::tracks();
```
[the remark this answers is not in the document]
olanod: I still think it's worth the duplication to not create some weird common dependency, let's hope it gets stabilized soon? 🤞🏽

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
ggwpez: It must be ordered since https://github.com/paritytech/polkadot-sdk/pull/3325/

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
ggwpez: Okay will let the next reviewer decide 😄

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
ggwpez: Why does it use `into_iter`? It is already an iterator 🤔

ggwpez (approved):
LGTM

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
olanod: Corrected... Probably a leftover of an initial version returning IntoIterator

muharem (commented):
Looks good. 

Since the motivation for these changes is to provide an option for setting tracks within the storage, and as these changes are breaking changes, it would be nice to have such a setup as an example and proof of concept to ensure we're not missing anything. For example, you could try setting it up for the `kitchensink-runtime` runtime with the `parameters` pallet.

olanod:
@muharem the feature is working, this week we released [pallet-communities](https://github.com/virto-network/virto-node/tree/master/pallets/communities) in Kreivo which allows DAOs to make use of the pallet-referenda infrastructure and set-up/configure their own tracks dynamically with [pallet-referenda-tracks](https://github.com/virto-network/frame-contrib/tree/main/pallets/referenda-tracks) implemented by @pandres95

muharem:
@olanod to have a setup example would be still good

gui1117 (commented):
The goal of the PR is to be able to fetch `tracks` from one storage item, or to have `tracks` in multiple storage item and being able to have a custom `Iterator` implementation which fetch the different chunks dynamically?

I feel like this PR only wants to be able to fetch tracks from one single value in storage, If this is the case we can simply and :
```rust
Cow<'static, [(Self::Id, TrackInfo<Balance, Moment>)]
```
or using the new type:
```rust
Cow<'static, [Track<Balance, Moment, N>]
```

This way we don't loose the binary search and we can process the type more easily.

If there is a usecase for an abstract `Iterator` then it is also ok.

Cargo.lock
```diff
@@ -14804,6 +14804,7 @@ dependencies = [
  "sp-core 28.0.0",
  "sp-io 30.0.0",
  "sp-runtime 31.0.1",
+ "sp-std 14.0.0",
```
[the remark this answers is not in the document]
pandres95: On it. I love this idea better than keep using `sp-std`. Think it's a burden more than a help to have to import an entire library for just a couple of `std`-ish imports that might be perfectly in `alloc`.

substrate/frame/referenda/src/types.rs
```diff
@@ -136,42 +145,74 @@ pub struct TrackInfo<Balance, Moment> {
...
+		Self::tracks().map(|x| x.id)
+	}
 
-		tracks.get(maybe_index).map(|(_, info)| info)
+	/// Return the track info for track `id`, by default this just looks it up in `Self::tracks()`.
+	fn info(id: Self::Id) -> Option<Cow<'static, TrackInfo<Balance, Moment, N>>> {
+		Self::tracks().find(|x| x.id == id).map(|t| match t {
+			Cow::Borrowed(x) => Cow::Borrowed(&x.info),
+			Cow::Owned(x) => Cow::Owned(x.info),
+		})
```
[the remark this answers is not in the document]
pandres95: Yes. This is precisely because to keep the reference in the `'static` lifecycle. 😅 

@olanod can explain it better.

substrate/frame/referenda/src/types.rs
```diff
@@ -111,10 +111,19 @@ pub struct Deposit<AccountId, Balance> {
...
-pub struct TrackInfo<Balance, Moment> {
+pub const DEFAULT_MAX_TRACK_NAME_LEN: usize = 25;
+
+/// Detailed information about the configuration of a referenda track
+#[derive(
+	Clone, Encode, Decode, MaxEncodedLen, TypeInfo, DebugNoBound, PartialEqNoBound, EqNoBound,
+)]
+pub struct TrackInfo<
+	Balance: Debug + Eq,
+	Moment: Debug + Eq,
```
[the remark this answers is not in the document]
pandres95: Done.

substrate/frame/referenda/src/types.rs
```diff
@@ -111,10 +111,19 @@ pub struct Deposit<AccountId, Balance> {
...
+	const N: usize = DEFAULT_MAX_TRACK_NAME_LEN,
+> {
 	/// Name of this track.
-	pub name: &'static str,
+	pub name: [u8; N],
```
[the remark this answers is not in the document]
pandres95: This suggestion, while good, would break the `'static` lifetime required under the specs of this refactor. Also, it's not really breaking encoding at runtime level, since there was no previous storage for this field. If anything, it'd be slightly breaking the display representation of this value (TODO @olanod: include this statement in the PRDoc to inform runtime users).

substrate/frame/referenda/src/types.rs
```diff
@@ -136,42 +145,74 @@ pub struct TrackInfo<Balance, Moment> {
...
+		// Adapted from Iterator::is_sorted implementation available in nightly
+		// https://github.com/rust-lang/rust/issues/53485
+		let mut iter = Self::tracks();
+		let mut last = match iter.next() {
+			Some(ref e) => e.id,
+			None => return Ok(()),
+		};
+		iter.all(|curr| {
+			let curr = curr.as_ref().id;
+			if let Some(Ordering::Greater) | None = last.partial_cmp(&curr) {
```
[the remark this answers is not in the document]
pandres95: Done.

substrate/frame/referenda/src/types.rs
```diff
@@ -136,42 +145,74 @@ pub struct TrackInfo<Balance, Moment> {
...
 }
 
+/// Track groups the information of a voting track with its corresponding identifier
+#[derive(
+	Clone, Encode, Decode, MaxEncodedLen, TypeInfo, DebugNoBound, PartialEqNoBound, EqNoBound,
+)]
+pub struct Track<
+	Id: Debug + Eq,
+	Balance: Debug + Eq,
+	Moment: Debug + Eq,
```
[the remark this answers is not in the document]
pandres95: Done.

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
pandres95: Done.

substrate/frame/referenda/src/tests.rs
```diff
@@ -289,11 +289,12 @@ fn alarm_interval_works() {
 fn decision_time_is_correct() {
 	ExtBuilder::default().build_and_execute(|| {
 		let decision_time = |since: u64| {
+			let track = TestTracksInfo::tracks().into_iter().next().unwrap();
```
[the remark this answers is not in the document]
pandres95: Yes. Done.

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
gui1117: I meant before we had a static vec and we did a binary search. Now we iterate in order to find the track.

As long as the number of track is not too big it is ok.

cumulus/parachains/runtimes/collectives/collectives-westend/src/ambassador/tracks.rs
```diff
@@ -246,9 +249,9 @@ impl pallet_referenda::TracksInfo<Balance, BlockNumber> for TracksInfo {
 						ceil: Perbill::from_percent(50),
 					},
 				},
-			),
+			},
 		];
-		&DATA[..]
+		DATA.iter().map(Cow::Borrowed)
 	}
```
gui1117, line 255: we could also implement `info` method manually to do the binary search.

gui1117 (approved):
In the trait `Polling` there is the implementation:

https://github.com/paritytech/polkadot-sdk/blob/2f3856ac3f3c02c0e74c16e9fe796a98201ca06c/substrate/frame/referenda/src/lib.rs#L769-L771

It will also convert to a vec. But I see it calls `TracksInfo::track_ids` which can be implemented more efficiently than iterating over storage.

Also it seems we only use it for a `try_delegate` to ensure the class exists. So it isn't the best API for our usage.

In general I am ok with the change but I feel like the API is hiding some complexity and if the API gets misused it can be a very inefficient implementation.
As long as the number of track is small it shouldn't matter too much.

Like if the method `Polling::classes` or `TracksInfo::info` is used often, it is not good, they are expensive.

substrate/frame/referenda/src/lib.rs
```diff
@@ -230,10 +238,13 @@ pub mod pallet {
 		/// Information concerning the different referendum tracks.
 		#[pallet::constant]
 		type Tracks: Get<
```
gui1117, line 240: (as a side note: this `Get` doesn't seem really needed, we could have add it in the constant using `pallet::extra_constant` attribute.)

Cargo.lock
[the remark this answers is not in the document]
pandres95: Done.

substrate/frame/referenda/src/lib.rs
pandres95: Done. Great suggestion.

substrate/frame/referenda/src/lib.rs
```diff
@@ -117,27 +117,6 @@ pub mod benchmarking;
 
 pub use frame_support::traits::Get;
 
-#[macro_export]
-macro_rules! impl_tracksinfo_get {
```
gui1117, line 121: Thanks for having done it, to avoid breaking change we could also keep the macro but make it deprecated. The deprecation message can say: tracks info no longer requires get, the macro invocation can be removed.

But keep it removed is also ok to me.

gui1117 (approved):

substrate/frame/referenda/src/lib.rs
pandres95: After a small evaluation, I concluded that removing it is better since we're eliminating some potentially unused code at the almost negligible cost of deleting 5 calls to this macro in the `runtimes` repo (plus one or two for every runtime being built outside the scope of Polkadot).

To me keeping it removed it's a no brainer.

cumulus/parachains/runtimes/collectives/collectives-westend/src/ambassador/tracks.rs
pandres95: We should have created a `static` parameter just to store the `DATA` array outside `TracksInfo::tracks`. If there's a way to achieve something similar without compromising the current performance, could be good in the future.

alindima:
This PR broke polkadot.js apps: https://github.com/polkadot-js/apps/issues/11310
Is this something that needs to be fixed in polkadot.js?

substrate/frame/referenda/src/types.rs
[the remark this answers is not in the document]
gui1117: looking at this PR again we should have warn more strongly in the PRdoc that this type is changing.

pandres95:
Yes. As announced in the PR description, clients should be aware of the new [`Track`](https://github.com/paritytech/polkadot-sdk/blob/c078d2f41cf8ecd28ff5279fcebb22da418a14b9/substrate/frame/referenda/src/types.rs#L149-L152) structure.

While internal encoding is the same, the structure it decodes to changes, so it should be adapted.

kianenigma:
This is a good example of PR where I would ask if it was really necessary to cause a breaking change? and also the PRdoc was not highlighting how this should be treated by the API libraries.

pandres95:
Yes. As announced in the PR description, clients should be aware of the new [`Track`](https://github.com/paritytech/polkadot-sdk/blob/c078d2f41cf8ecd28ff5279fcebb22da418a14b9/substrate/frame/referenda/src/types.rs#L149-L152) structure.

While internal encoding is the same, the structure it decodes to changes, so it should be adapted.

gui1117:
The internal encoding is changed: https://github.com/paritytech/polkadot-sdk/pull/2072#discussion_r1951954478

the field was a `Vec<u8>` it is now `[u8; 25]`, maybe we can use a type with the same encoding?

ggwpez:
> the field was a `Vec<u8>` it is now `[u8; 25]`, maybe we can use a type with the same encoding?

Yea we could do a custom encoding that prepends the length again.

pandres95:
I think the biggest problem here is getting the const, not really storing the content.

A solution might be transforming the tracks iter into a getter that resolves to the previously known structure. I can do that.

gui1117:
> I think the biggest problem here is getting the pallet constant, not really storing the content.
> 
> A solution might be transforming the tracks iter into a getter that resolves to the previously known structure. I can do that.

Yes it will also work.

substrate/frame/referenda/src/lib.rs
```diff
@@ -249,6 +222,14 @@ pub mod pallet {
 		type BlockNumberProvider: BlockNumberProvider;
 	}
 
+	#[pallet::extra_constants]
+	impl<T: Config<I>, I: 'static> Pallet<T, I> {
+		#[pallet::constant_name(Tracks)]
+		fn tracks() -> Vec<Track<TrackIdOf<T, I>, BalanceOf<T, I>, BlockNumberFor<T, I>>> {
+			T::Tracks::tracks().map(|t| t.into_owned()).collect()
+		}
+	}
```
gui1117, line 231: But having this as a constant is incompatible with the trait usage.
In the trait we want to make use of storage to have dynamic tracks. But this being a constant means the track must not change.

I think this PR is inconsistent.
Maybe we should not extend referenda and let other people use a different pallet if they need dynamic tracks.

Or we should make it a view function and remove the constant, which is a bigger breaking change.

olanod:
The initial design of this PR took into account [Gav's requirements](https://github.com/paritytech/polkadot-sdk/issues/3420#issuecomment-1957152156) of not having any kind of performance regression(i.e. keep allowing the const definition of tracks) and not breaking anything, as tracks were not stored it was fine but didn't consider the front-end.
Is the name of the track not encoding properly the only breaking change? We can match the previous `&str` encoding, SCALE encodes string slices like vectors prefixing data with a compact encoding of the length of the bytes, just like the [`const fn str_arr`](https://github.com/paritytech/polkadot-sdk/blob/ececb4cbd68d94d1f73b6ab8e5ad8afed1cc9456/substrate/primitives/runtime/src/lib.rs#L1040) I introduced previous to this PR to help with the easy declaration of track names in the const context we can extend that function to prepend the needed compact length of the bytes.

ggwpez:
I dont see a fixup MR open - going to revert it until it is sorted. Maybe there exists a better approach of the likes as Gui pointed out above.

pandres95:
Hey @ggwpez . The fixup MR will be ready before 5pm (COT) today. I had to figure out the best way possible to make this to be as performant as possible. After some tests, I'm taking a way that's similar to what Gui actually proposed.

substrate/frame/referenda/src/lib.rs
athei: I had to get full sherlock homes to find out whether its okay to delete this macro invocation or whether it had to be replaces. Just trying doesn't help cause you are always at 300+ errors when upgrading your runtime.

gui1117: Yes, you we can reintroduce it an mark it as deprecated.

PR #2077, opened 2023-10-30
pallet-contracts: Do not return an error on already applied upgrades

bkchr:
Migrations in general should just skip the operation, if they were already applied. This is especially useful for CI checks that run in each pr to check that the current set of migrations are valid.

### DIFF
--- a/substrate/frame/contracts/src/migration.rs
+++ b/substrate/frame/contracts/src/migration.rs
@@ -311,11 +311,6 @@ impl<T: Config, const TEST_ALL_STEPS: bool> OnRuntimeUpgrade for Migration<T, TE
 		let storage_version = <Pallet<T>>::on_chain_storage_version();
 		let target_version = <Pallet<T>>::current_storage_version();
 
-		ensure!(
-			storage_version != target_version,
-			"No upgrade: Please remove this migration from your runtime upgrade configuration."
-		);
-
 		log::debug!(
 			target: LOG_TARGET,
 			"Requested migration of {} from {:?}(on-chain storage version) to {:?}(current storage version)",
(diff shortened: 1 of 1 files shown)


pgherveou:
> Migrations in general should just skip the operation, if they were already applied. This is especially useful for CI checks that run in each pr to check that the current set of migrations are valid.

agreed, let me update that to a warn log. I think the is_upgrade_supported might need to be adjusted as well. Double checking..

bkchr:
> agreed, let me update that to a warn log. I think the is_upgrade_supported might need to be adjusted as well. Double checking..

There is a `warn` in `on_runtime_upgrade`.

pgherveou (approved):

agryaznov (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4122601

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4127293

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4127294

pgherveou:
@liamaharon 
I feel that this fix here is a bit more accurate, if a migration step is missing we should panic in try_runtime not just log a warning, it means that the migration will for sure fails on_runtime_upgrade.

the noop case though should just trigger a warning.
Maybe I am missing something, feel free to comment and re-close again

liamaharon:
@pgherveou can that case be hit during normal execution or is it only an err / misconfiguration path? 

What if the migration is left in the runtime, but two additional migrations are made?

PR #2088, opened 2023-10-30
Reduce Impact on Identity Pallet in Migration

joepetrowski:
Addresses https://github.com/paritytech/polkadot-sdk/pull/1814#pullrequestreview-1701818579

- Removes `lock`/`unlock` extrinsics
- Moves `reap_identity` and `poke_deposit` into special migration pallet to avoid adding new one-time calls to the Identity pallet

### DIFF
--- a/polkadot/runtime/common/src/identity_migrator.rs
+++ b/polkadot/runtime/common/src/identity_migrator.rs
@@ -1,0 +1,79 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This pallet is designed to go into a source chain and destination chain to migrate data. The
+//! design motivations are:
+//!
+//! - Call some function on the source chain that executes some migration (clearing state,
+//!   forwarding an XCM program).
+//! - Call some function (probably from an XCM program) on the destination chain.
+//! - Avoid cluttering the source pallet with new dispatchables that are unrelated to its
+//!   functionality and only used for migration.
+//!
+//! After the migration is complete, the pallet may be removed from both chains' runtimes.
+
+pub use pallet::*;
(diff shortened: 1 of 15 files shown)


polkadot/runtime/rococo/src/impls.rs
```diff
@@ -29,14 +29,14 @@ use xcm_executor::traits::TransactAsset;
 /// remote calls.
 #[derive(Encode, Decode)]
 enum PeopleRuntimePallets<AccountId: Encode> {
-	#[codec(index = 50)]
-	Identity(IdentityCalls<AccountId>),
+	#[codec(index = 248)]
```
muharem, line 32: looks like this references the index on the relay chain now.
may be TODO to correct this when we know the index on people chain

substrate/frame/identity/src/benchmarking.rs
```diff
@@ -626,45 +623,16 @@ mod benchmarks {
...
-	}
-
-	#[benchmark]
-	fn unlock_pallet() -> Result<(), BenchmarkError> {
-		let origin =
-			T::LockerOrigin::try_successful_origin().map_err(|_| BenchmarkError::Weightless)?;
-		Identity::<T>::lock_pallet(origin.clone())?;
+		#[block]
+		{
+			let _ = Identity::<T>::poke_deposit(&target);
```
muharem, line 628: may be worth removing new benchs funcs from the pallet. new bench functions introduce new function to the weight trait, which will require all clients to run benchmarks again

muharem (approved):

polkadot/runtime/rococo/src/impls.rs
joepetrowski: Yeah, definitely

bkchr (approved):
Thank you :heart:

polkadot/runtime/common/src/identity_migrator.rs
```diff
@@ -0,0 +1,137 @@
...
+	/// program to another chain. Concretely, a type implementing this trait in the Polkadot
+	/// runtime would teleport enough DOT to the People Chain to cover the Identity deposit there.
+	///
+	/// This could also directly include `Transact { poke_deposit(..), ..}`.
+	///
+	/// Inputs
+	/// - `who`: Whose identity was reaped.
+	/// - `fields`: The number of `additional_fields` they had.
+	/// - `subs`: The number of sub-accounts they had.
+	fn on_reap_identity(who: &AccountId, fields: u32, subs: u32) -> DispatchResult;
```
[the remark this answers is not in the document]
joepetrowski: Yes I will address any downstream changes. My PR is still more a proposal on how to do the migration. George's should definitely be merged first.

substrate/frame/identity/src/benchmarking.rs
joepetrowski: I will address this in the parent PR. Made this one more for indication of this as a better direction.

PR #2095, opened 2023-10-31
xcmp-queue: remove outdated bridging comment

acatangiu:
Removed confusing and outdated `TODO`.

### DIFF
--- a/cumulus/pallets/xcmp-queue/src/bridging.rs
+++ b/cumulus/pallets/xcmp-queue/src/bridging.rs
@@ -63,13 +63,6 @@ impl<SiblingBridgeHubParaId: Get<ParaId>, Runtime: crate::Config>
 			return true
 		}
 
-		// TODO: https://github.com/paritytech/polkadot-sdk/pull/1556 - once this PR is merged, we may
-		// remove the following code.
-		// TODO: the following restriction is arguable, we may live without that, assuming that
-		// There can't be more than some `N` messages queued at the bridge queue (at the source BH)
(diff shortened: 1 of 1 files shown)


svyatonik (approved):

cumulus/pallets/xcmp-queue/src/bridging.rs
```diff
@@ -63,13 +63,6 @@ impl<SiblingBridgeHubParaId: Get<ParaId>, Runtime: crate::Config>
 			return true
 		}
 
-		// TODO: https://github.com/paritytech/polkadot-sdk/pull/1556 - once this PR is merged, we may
-		// remove the following code.
-		// TODO: the following restriction is arguable, we may live without that, assuming that
-		// There can't be more than some `N` messages queued at the bridge queue (at the source BH)
-		// AND before accepting next (or next-after-next) delivery transaction, we'll receive the
-		// suspension signal from the target parachain and stop accepting delivery transactions.
```
acatangiu, line 71: I guess we could keep these lines actually, wdyt @svyatonik ?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4141107

cumulus/pallets/xcmp-queue/src/bridging.rs
svyatonik: We may or may not leave it. Like this "arguable" may also sound confusing to readers. What remains is enough to justify the code imo

bkontur (approved):

PR #2098, opened 2023-10-31
check-each-crate: Do not reference crate to check by name

bkchr:
This pull request changes how `check-each-crate.py` is working. Instead of passing the name of the crate via `-p`, we now jump into the directory of the crate and call there `cargo check`. This should fix issues like https://github.com/paritytech/polkadot-sdk/issues/2013 where a crate is present twice in the `Cargo.lock`.

Besides that it also changes `core/Cargo.toml` to not always pull in bandersnatch.


davxy (approved):
Nice. Ty

ggwpez (approved):

PR #2106, opened 2023-10-31
remove gum dependency on jaeger

### LINKED ISSUE #649: PVF worker: look into removing unneeded dependencies
# ISSUE

## Overview

As part of https://github.com/paritytech/polkadot-sdk/issues/882, we want to block unneeded syscalls in the PVF workers. After https://github.com/paritytech/polkadot-sdk/issues/650 we will have separate worker binaries, whose object files we can analyze for included syscalls. However, right now a lot of unnecessary syscalls are found present in the object files. 

This issue is about removing some big dependencies[^1] and seeing if it has an impact on syscalls or binary size, or if Rust is smart enough that it's already optimizing as much as it can. (Will make sure to try this in combination with enabling LTO.)

[^1]: i.e. we can remove the dependency on `polkadot-node-core-pvf` by moving that into e.g. `host` and common objects into `common`, so we have the following structure: `pvf/host`, `pvf/worker`, `pvf/common`. Then see how that affects binary size.

Jpserrat:
# Description
As [suggested](https://github.com/paritytech/polkadot-sdk/issues/649#issuecomment-1691880952), `tracing-gum` only relies on one function of `jaeger` and the dependency can be removed by copying this function.
Let me know if this is what you guys had in mind.

Closes #649


ordian (approved):

mrcnski:
I had previously used `cargo udeps` to check for unused deps, but it seems like it doesn't really work, because I went and manually checked each dep and many were actually not used. So I just removed them.

It doesn't look like this PR actually cut down on [the number of syscalls](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4164079) in the binaries, which is good. It means that LTO was working well and not including unused code. `tokio` had already been removed previously, which was a [big win](https://github.com/paritytech/polkadot-sdk/pull/2009/commits/cd10afac1ddf2b738b944586cfba217d56fc0949).

s0me0ne-unkn0wn (approved):
A good clean-up! :+1:

PR #2107, opened 2023-10-31
XCM builder pattern

franciscoaguirre:
Added a proc macro to be able to write XCMs using the builder pattern.
This means we go from having to do this:

```rust
let message: Xcm<()> = Xcm(vec![
  WithdrawAsset(assets),
  BuyExecution { fees: asset, weight_limit: Unlimited },
  DepositAsset { assets, beneficiary },
]);
```

to this:

```rust
let message: Xcm<()> = Xcm::builder()
  .withdraw_asset(assets)
  .buy_execution(asset, Unlimited),
  .deposit_asset(assets, beneficiary)
  .build();
```


polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -0,0 +1,93 @@
+use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
+use syn::{parse_macro_input, DeriveInput, Data, Error, Fields};
+use quote::{quote, format_ident};
+
+pub fn derive(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+	let builder_impl = match &input.data {
+		Data::Enum(data_enum) => generate_methods_for_enum(input.ident, data_enum),
+		_ => return Error::new_spanned(&input, "Expected the `Instruction` enum")
```
gupnik, line 10: Would expect UI tests for such errors

polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn generate_methods_for_enum(name: syn::Ident, data_enum: &syn::DataEnum) -> TokenStream2 {
+	let methods = data_enum.variants.iter().map(|variant| {
+		let variant_name = &variant.ident;
+		let method_name_string = to_snake_case(&variant_name.to_string());
+		let method_name = syn::Ident::new(&method_name_string, variant_name.span());
+		match &variant.fields {
+			Fields::Unit => {
+				quote! {
+					pub fn #method_name(mut self) -> Self {
```
KiChjang, line 33: What would be really amazing is if you can pull the docs for the individual XCM instructions to be above this fn item. Should not be hard to do.

polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -0,0 +1,93 @@
...
+            if chars.peek().map_or(false, |next| next.is_lowercase()) {
+                snake.push('_');
+            }
+        }
+
+        snake.extend(ch.to_lowercase());
+    }
+
+    snake
+}
```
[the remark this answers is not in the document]
KiChjang: Aside from `macro_magic`, there's also `Inflection` that you can make use of in order to convert between snake cases, pascal cases and more.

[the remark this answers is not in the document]
franciscoaguirre: I looked into `Inflections` but did not want to add a new dependency 🤷, I can add it or just copy the code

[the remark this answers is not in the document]
franciscoaguirre: I just saw that we already use the inflector crate 😄, I'll make the change then

polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -0,0 +1,95 @@
...
+							Lit::Str(string_literal) => Some(string_literal.value()),
+							_ => None,
+						},
+						_ => None,
+					},
+					_ => None,
+				}
+			} else {
+				None
+			}
```
KiChjang, line 45: match &attr.meta {
    Meta::NameValue(MetaNameValue { value: Expr::Lit(ExprLit { lit: Lit::Str(literal), .. ), .. }) if attr.path().is_ident("doc") => Some(literal.value()),
    _ => None,
}

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4181411

franciscoaguirre:
bot fmt

franciscoaguirre:
bot fmt

ffarall:
Very cool syntax! In fact it would be great if you could do stuff like this:
```rust
let message_builder: XcmBuilder<()> = Xcm::builder()
  .withdraw_asset(assets)
  .buy_execution(asset, Unlimited);

// Some more code

let message: Xcm<()> = message_builder
  .deposit_asset(assets, beneficiary)
  .build();

message.execute();
message.send();
```

polkadot/xcm/procedural/Cargo.toml
```diff
@@ -13,3 +15,4 @@ proc-macro2 = "1.0.56"
 quote = "1.0.28"
 syn = "2.0.38"
 Inflector = "0.11.4"
```
KiChjang, line 17: Is this a dev dependency?

polkadot/xcm/procedural/Cargo.toml
```diff
@@ -13,3 +15,4 @@ proc-macro2 = "1.0.56"
...
+trybuild = { version = "1.0.74", features = ["diff"] }
```
KiChjang, line 18: Is this a dev dependency?

polkadot/xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -649,4 +649,23 @@ mod tests {
...
+
+	#[test]
+	fn builder_pattern_works() {
+		let asset: MultiAsset = (Here, 100u128).into();
+		let beneficiary: MultiLocation = AccountId32 { id: [0u8; 32], network: None }.into();
+		let message: Xcm<()> = Xcm::builder()
+			.withdraw_asset(asset.clone().into())
+			.buy_execution(asset.clone(), Unlimited)
+			.deposit_asset(asset.clone().into(), beneficiary)
+			.build();
```
kaichaosun, line 661: The order of the methods here matters, right?
My intuitive to general builder model is that the order should not matter at all.
I suggest use something like,
```
Xcm::new()
  .withdraw_asset(...)
  .buy_execution(..)
  .deposit_asset(...)
```
If using this chained method syntax, we probably don't need to introduce a new macro? (there are already a lot of macros in the codebase.)

KiChjang: The derive macro acts as a convenient way to create methods that are very similar to the declaration of the variants in the `Instruction` enum.

I also find it quite absurd to suggest that the order shouldn't matter. How could it be different? Every XCM instruction order matters in an XCM program. The argument that you're making with your example is simply whether or not we need to call the `builder` method before chaining instruction methods, however this has been a [Rust conventional pattern to create a `builder` method](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html). Is there a good reason why we shouldn't do this, other than the mistaken assumption that it would remove the derive macro?

franciscoaguirre: I think what he means is in the traditional builder pattern, the order of methods does not matter, and aims to make this clear by using `new()` instead of the usual `builder()`.
I have nothing against doing this change. It's a few characters less.

franciscoaguirre: What is true is we do need the macro, although we could change its name from `Builder` to something else, like `VariantsToMethods` but I'd argue it's not that off from the builder pattern, just the order thing, which we would easily deal with via documentation/communication

polkadot/xcm/procedural/Cargo.toml
franciscoaguirre: Indeed, I moved it

polkadot/xcm/xcm-simulator/example/src/lib.rs
KiChjang: > the order of methods does not matter

How does this make _any_ sense? Would the order of XCM instructions not matter? Why is this even being suggested as something viable?

Polkadot-Forum:

https://forum.polkadot.network/t/xcm-user-and-developer-experience-improvements/4511/1

polkadot/xcm/xcm-simulator/example/src/lib.rs
KiChjang: > in the traditional builder pattern, the order of methods does not matter

Yet we are using the builder pattern to build XCM, where the instruction order matters. If @kaichaosun means what you described, then it would seem it is a suggestion coming from a lack of understanding of what XCM instructions and programs are.

kaichaosun: > But then the `build` at the end feels kind of weird to me.

If the chained method makes sense, we don't need `build` any more.

> The derive macro acts as a convenient way to create methods that are very similar to the declaration of the variants in the `Instruction` enum.

Yeah it's convenient for writers, but it's not convenient for the readers/users, and this interface is meant to be used by runtime devs, using macro here seems take the wrong balance. Generally I'm not fan of using macros if the logic is business heavy, especially in user interface.

BTW, not sure if it makes sense, but I'd like to see something like this in the future with xcm,
```
Xcm.new()
   .withdraw(...),
   .fuel(...)
   .deposit(...)
```
Which feels just like using `map`, `filter` on iterator. There is `iter()` and `collect()` to convert between iterator and vector, you probably know better identifiers what should be used in Xcm if such abstraction is needed when multi Xcm exists. Before that, `new` seems enough? Also, for better UX, it may not be necessarily 1 to 1 map between such convenient method and the xcm instruction variant.

Note, I'm not a heavy XCM user, my suggestion is based on the my general programing experience, and what potentially can provide a better dev experience.

KiChjang: I'm sorry, but you're not making much sense. Users do not dig deep into the macro code to look at the function body of e.g. `withdraw_asset`, the only API they care about is just the function. Hence readability of the macro code is not important for DevEx.

KiChjang: > Also, for better UX, it may not be necessarily 1 to 1 map between such convenient method and the xcm instruction variant.

You are basically saying that we should build yet another abstraction layer on top of XCM, so that users can use XCM without actually knowing about XCM. This is a non-goal; the goal of XCM isn't to make it easier for end-users to craft XCM programs, but rather facilitate developers (especially parachain developers) to innovate upon cross-chain interactions. What you're describing is also no different from a macro, from the perspective that they're both trying to hide complexity, when the goal here _isn't_ that, but simply to make XCM nicer for developers to work with.

franciscoaguirre: I think using different types, so `XcmBuilder` accessed via `Xcm::builder()` that then has `.build()` to go back to `Xcm` is a good thing.
It lets us add some nice limitations when building the message and then making sure that when calling `build()` it's valid, as we deem it.

KiChjang (approved):

gupnik (approved):

Polkadot-Forum:

PR #2112, opened 2023-11-01
remove transitional code wrt executor parameters

### LINKED ISSUE #714: Remove transitional code wrt executor parameters
After the changes introduced by paritytech/substrate#6161 are released and are on-chain for no less than seven sessions, [the transitional code](https://github.com/paritytech/polkadot/blob/9a17c4e9bf44bc431e48cdb145ebedc7d91c0aa7/runtime/parachains/src/runtime_api_impl/vstaging.rs#L33) intended to work around paritytech/polkadot-sdk#64 should be removed for the sake of purity and determinism.

Jpserrat:
I've replaced the code with the comment suggestion, @s0me0ne-unkn0wn.

Closes #714


s0me0ne-unkn0wn:
Well done, it's definitely a legit change, as the executor params have been on Polkadot for a long time already... I'm just asking myself if, for the sake of paranoia, it needs a burn-in prior to merging. I'll ask fellows if we have the capacity for that.

mrcnski (approved):

bkchr (approved):
I don't see any reason for a burn in. In general I have to say that the runtime api returns an `Option` was not a very good idea here. As we actually expect that this always returns something...

alexggh (approved):

bkchr (changes requested):
Actually this implementation is broken for chains that start from genesis. We should keep the current code. We should just remove the comment. As I said above, this runtime api function should never have returned an option..

s0me0ne-unkn0wn:
> Actually this implementation is broken for chains that start from genesis.

Why? TBH, I didn't test that, but currently, we should bootstrap session zero executor parameters from `HostConfiguration::executor_params`. I believe the host configuration should be there either way.

s0me0ne-unkn0wn:
> As I said above, this runtime api function should never have returned an option.

The executor parameters are side-data for `SessionInfo`, and `session_info()` returns an `Option`. What would you want it returned if requested for a non-existing session index?

bkchr:
> > 
> 
> Why? TBH, I didn't test that, but currently, we should bootstrap session zero executor parameters from `HostConfiguration::executor_params`. I believe the host configuration should be there either way.

There is no code that does this. At least I couldn't find it?

s0me0ne-unkn0wn:
> There is no code that does this. At least I couldn't find it?

This one: https://github.com/paritytech/polkadot-sdk/blob/2726d5af65d041adbcd3e96986a233e7c8fa07a8/polkadot/runtime/parachains/src/session_info.rs#L196

`initializer_on_new_session()` is run for the session zero, right?

bkchr:
> `initializer_on_new_session()` is run for the session zero, right?

Ahh yeah, that is actually executed on genesis. Then sorry!

bkchr:
> > 
> 
> The executor parameters are side-data for `SessionInfo`, and `session_info()` returns an `Option`. What would you want it returned if requested for a non-existing session index?

Maybe a Result, but that wouldn't change that much probably. I mean we could then have at least covered the case where the session was already pruned. (but you could only have guessed what the parameters were back then)

bkchr (approved):
Approving based on the discussion above.

PR #2114, opened 2023-11-01
Bandersnatch dependency update

davxy:
Closes https://github.com/paritytech/polkadot-sdk/issues/2013

Previous version was using old names for crates hosted @ https://github.com/paritytech/arkworks-substrate


Cargo.lock
```diff
@@ -17478,16 +17570,34 @@ dependencies = [
 name = "sp-std"
 version = "8.0.0"
 
+[[package]]
+name = "sp-std"
+version = "8.0.0"
+source = "git+https://github.com/paritytech/polkadot-sdk#9ca267328e0cba753aa528615cc19d1f633ed764"
```
ggwpez, line 17576: So we have two versions of `sp-std` now?  
It seems that the pallets use the external version now...

davxy: There are two versions of the substrate crates which are pulled in by `bandersnatch_vrf` crate.

This is due to the fact that we are referencing crates by path while the external crate is referencing the git repo.
These two may be different.

Honestly I don't see another solution to this as (while improbable) external crates can legitimately depend on one of the Substrate crates.

Another approach could be to directly embed the `bandersnatch_vrfs` crate code in Polkadot repo. 
But this seems a bit wrong as:
1. we can't prevent some external crate to depend on something we expose in our crates
2. `bandersnatch_vrfs` crate is something which is not built for Substrate but should be useful in general

If we embed its code we add some maintenance burden to keep it aligned with the upstream.

I'm open to other approaches and suggestions

michalkucharczyk (approved):

bkchr (approved):

skunert (approved):

PR #1246, opened 2023-08-29
Use `Message Queue` as DMP and XCMP dispatch queue

ggwpez:
(imported from https://github.com/paritytech/cumulus/pull/2157)

## Changes

This MR refactores the XCMP, Parachains System and DMP pallets to use the [MessageQueue](https://github.com/paritytech/substrate/pull/12485) for delayed execution of incoming messages. The DMP pallet is entirely replaced by the MQ and thereby removed. This allows for PoV-bounded execution and resolves a number of issues that stem from the current work-around.

All System Parachains adopt this change.  
The most important changes are in `primitives/core/src/lib.rs`, `parachains/common/src/process_xcm_message.rs`, `pallets/parachain-system/src/lib.rs`, `pallets/xcmp-queue/src/lib.rs` and the runtime configs.

### DMP Queue Pallet

The pallet got removed and its logic refactored into parachain-system. Overweight message management can be done directly  through the MQ pallet.

The undeployment of the pallet commences through the `on_idle` hook step-by step and moves all the remaining messages into the `DmpSink`.

```rust
impl cumulus_pallet_dmp_queue::Config for Runtime {
	type DmpSink = MessageQueue;
	...
```

The pallet can be removed after the lazy migration is done and event `Completed` is emitted.

### XCMP Queue pallet

Removed all dispatch queue functionality. Incoming XCMP messages are now either: Immediately handled if they are Signals, enqueued into the MQ pallet otherwise.

New config items for the XCMP queue pallet:
```rust
/// The actual queue implementation that retains the messages for later processing.
type XcmpQueue: EnqueueMessage<ParaId>;

/// How a XCM over HRMP from a sibling parachain should be processed.
type XcmpProcessor: ProcessMessage<Origin = ParaId>;

/// The maximal number of suspended XCMP channels at the same time.
#[pallet::constant]
type MaxInboundSuspended: Get<u32>;
```

How to configure those:

```rust
// Use the MessageQueue pallet to store messages for later processing. The `TransformOrigin` is needed since
// the MQ pallet itself operators on `AggregateMessageOrigin` but we want to enqueue `ParaId`s.
type XcmpQueue = TransformOrigin<MessageQueue, AggregateMessageOrigin, ParaId, ParaIdToSibling>;

// Process XCMP messages from siblings. This is type-safe to only accept `ParaId`s. They will be dispatched
// with origin `Junction::Sibling(…)`.
type XcmpProcessor = ProcessFromSibling<
	ProcessXcmMessage<
[...]

PR #2117, opened 2023-11-01
RFC: Reorganise AggregateMessageOrigin to support Snowbridge

vgeddes:
The `AggregateMessageOrigin` enum now 2 levels of hierarchy, with two top-level variants, `Xcm` and `Other`, to be used by (dmp-queue, hrmp-queue) and snowbridge respectively.

This PR only changes the `bridge-hub-rococo` runtime as a demonstration of the refactor. It also demonstrates how we intend to modify BridgeHub so that our bridge can also use MessageQueue.

```rust
/// The aggregate origin of an inbound message.
#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
pub enum AggregateMessageOrigin {
	/// The message came from the DMP or HRMP subsystem
	Xcm(MessageOrigin),
	/// The message came from some other pallet
	Other(MessageOrigin),
}

/// The origin of an inbound message.
#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
pub enum MessageOrigin {
	/// The message came from the para-chain itself.
	Here,
	/// The message came from the relay-chain.
	///
	/// This is used by the DMP queue.
	Parent,
	/// The message came from a sibling para-chain.
	///
	/// This is used by the HRMP queue.
	Sibling(ParaId),
}
```

The other change is to register the `MessageQueue` pallet last in the runtime definition for `bridge-hub-rococo`. This is because we need our bridge `on_initialize` handler to run before the MessageQueue `on_initialize` handler.

### DIFF
--- a/cumulus/primitives/core/src/lib.rs
+++ b/cumulus/primitives/core/src/lib.rs
@@ -120,9 +120,9 @@ impl From<u32> for AggregateMessageOrigin {
 	fn from(x: u32) -> Self {
 		use MessageOrigin::*;
 		match x {
-			0 => Self::Xcmp(Here),
-			1 => Self::Xcmp(Parent),
-			p => Self::Xcmp(Sibling(ParaId::from(p))),
+			0 => Self::Xcm(Here),
+			1 => Self::Xcm(Parent),
+			p => Self::Xcm(Sibling(ParaId::from(p))),
 		}
 	}
 }
--- a/cumulus/primitives/core/src/lib.rs
+++ b/cumulus/primitives/core/src/lib.rs
@@ -78,9 +78,18 @@ impl From<MessageSendError> for &'static str {
 	}
 }
 
-/// The origin of an inbound message.
+/// The aggregate origin of an inbound message.
 #[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
 pub enum AggregateMessageOrigin {
+	/// The message came from the DMP or HRMP subsystem
+	Xcm(MessageOrigin),
+	/// The message came from some other pallet
+	Other(MessageOrigin),
+}
+
+/// The origin of an inbound message.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
+pub enum MessageOrigin {
 	/// The message came from the para-chain itself.
 	Here,
 	/// The message came from the relay-chain.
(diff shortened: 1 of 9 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4164297

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4165881

cumulus/primitives/core/src/lib.rs
```diff
@@ -78,9 +78,18 @@ impl From<MessageSendError> for &'static str {
 	}
 }
 
-/// The origin of an inbound message.
+/// The aggregate origin of an inbound message.
 #[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
 pub enum AggregateMessageOrigin {
+	/// The message came from the DMP or HRMP subsystem
```
ggwpez, line 84: I dont think we can just change this enum, since its use by the MQ pallet as queue ID.  
That would mean that we need a lazy migration across all parachain which is a huge pain.  
It would be better to just add a variant to it like `Abridged` or `Other` and keeping the others compatible.

ggwpez: Unless this change goes in directly after https://github.com/paritytech/polkadot-sdk/pull/1246 in the same release, but its difficult.

joepetrowski: Also this doesn't make sense, `XCM` is not an origin or transport protocol. I can't tell if you are indicating the _format_ of the message or its origin with this type? If the latter, why not just add a new variant to the `AggregateMessageOrigin` enum, like `ExternalConsensus(Junction)`, where the `Junction` is expected to be of the variant `GlobalConsensus(NetworkId)`?

vgeddes: > I dont think we can just change this enum, since its use by the MQ pallet as queue ID.

Good point, didn't think of that. I'll will just add another variant to the enum then.

Here's a new design, inline with the above suggested changes.

Since the `NetworkId` and `Junction` types are not stable across XCM versions (and because they are used at higher protocol layers), I have not used them here.

```rust
/// The origin of an inbound message.
#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
pub enum AggregateMessageOrigin {
	/// The message came from the para-chain itself.
	Here,
	/// The message came from the relay-chain.
	///
	/// This is used by the DMP queue.
	Parent,
	/// The message came from a sibling para-chain.
	///
	/// This is used by the HRMP queue.
	Sibling(ParaId),
	/// External consensus
	ExternalConsensus {
		origin: ExternalConsensusOrigin,
		dest: ExternalConsensusDestination
	}
}

/// External consensus destinations
#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
pub enum ExternalConsensusDestination {
	// This network is reachable via Snowbridge on BridgeHub
	Ethereum { chain_id: u64 }
}

/// The origin of a message destined for an external consensus system
#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
pub enum ExternalConsensusOrigin {
	Here,
	Sibling(ParaId)
}
```

acatangiu:
@vgeddes this was automatically closed when `oty-message-queue` branch was deleted, to reopen rebase over `master`

vgeddes:
> @vgeddes this was automatically closed when `oty-message-queue` branch was deleted, to reopen rebase over `master`

Oopsie! ok, I've opened this PR: https://github.com/paritytech/polkadot-sdk/pull/2146

PR #2118, opened 2023-11-01
[FRAME] Short-circuit fungible self transfer

ggwpez:
Changes:
- Change the fungible(s) logic to treat a self-transfer as No-OP (as long as all pre-checks pass).

Note that the self-transfer case will not emit an event since no state was changed.

### DIFF
--- a/bridges/modules/relayers/src/stake_adapter.rs
+++ b/bridges/modules/relayers/src/stake_adapter.rs
@@ -36,7 +36,7 @@ impl<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 	StakeAndSlash<AccountId, BlockNumber, Currency::Balance>
 	for StakeAndSlashNamed<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 where
-	AccountId: Codec + Debug + PartialEq,
+	AccountId: Codec + Debug,
 	Currency: NamedReservableCurrency<AccountId>,
 	ReserveId: Get<Currency::ReserveIdentifier>,
 	Stake: Get<Currency::Balance>,
--- a/bridges/modules/relayers/src/stake_adapter.rs
+++ b/bridges/modules/relayers/src/stake_adapter.rs
@@ -36,7 +36,7 @@ impl<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 	StakeAndSlash<AccountId, BlockNumber, Currency::Balance>
 	for StakeAndSlashNamed<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 where
-	AccountId: Codec + Debug,
+	AccountId: Codec + Debug + PartialEq,
 	Currency: NamedReservableCurrency<AccountId>,
 	ReserveId: Get<Currency::ReserveIdentifier>,
 	Stake: Get<Currency::Balance>,
--- a/bridges/primitives/relayers/src/lib.rs
+++ b/bridges/primitives/relayers/src/lib.rs
@@ -115,7 +115,7 @@ where
 impl<T, Relayer> PaymentProcedure<Relayer, T::Balance> for PayRewardFromAccount<T, Relayer>
 where
 	T: frame_support::traits::fungible::Mutate<Relayer>,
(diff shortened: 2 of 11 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4164388

franciscoaguirre (approved):
I'd add in the documentation of the `transfer` function that a self transfer is a noop

juangirini (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4166149

substrate/frame/support/src/traits/tokens/fungibles/regular.rs
```diff
@@ -363,6 +369,10 @@ pub trait Mutate<AccountId>: Inspect<AccountId> + Unbalanced<AccountId> {
 		let _extra = Self::can_withdraw(asset.clone(), source, amount)
 			.into_result(preservation != Expendable)?;
 		Self::can_deposit(asset.clone(), dest, amount, Extant).into_result()?;
+		if source == dest {
```
KiChjang, line 372: Probably not important, but since the trait bound is `PartialEq` and not `Eq`, we are permitting a weird edge case where `source == dest`, but `source != source` and/or `dest != dest`. In practice, I don't think an account ID would ever _not_ be equal to itself, but just making a note here to ensure that we believe `PartialEq` is enough.

ggwpez: Yea, might be saner to use `Eq` then.

substrate/frame/support/src/traits/tokens/fungible/item_of.rs
```diff
@@ -256,7 +256,7 @@ impl<
 impl<
 		F: fungibles::MutateHold<AccountId>,
 		A: Get<<F as fungibles::Inspect<AccountId>>::AssetId>,
-		AccountId,
+		AccountId: PartialEq,
```
seadanda, line 259: likewise

bridges/primitives/relayers/src/lib.rs
```diff
@@ -104,7 +104,7 @@ pub struct PayRewardFromAccount<T, Relayer>(PhantomData<(T, Relayer)>);
 
 impl<T, Relayer> PayRewardFromAccount<T, Relayer>
 where
-	Relayer: Decode + Encode,
+	Relayer: Decode + Encode + PartialEq,
```
seadanda, line 107: likewise

bridges/modules/relayers/src/stake_adapter.rs
```diff
@@ -36,7 +36,7 @@ impl<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 	StakeAndSlash<AccountId, BlockNumber, Currency::Balance>
 	for StakeAndSlashNamed<AccountId, BlockNumber, Currency, ReserveId, Stake, Lease>
 where
-	AccountId: Codec + Debug,
+	AccountId: Codec + Debug + PartialEq,
```
seadanda, line 39: Why PartialEq instead of Eq here? Isn't reflexive-ness exactly what the intention is here for self comparison?

seadanda (commented):
Looks good, just one query (in three places) about the use of `PartialEq` over `Eq`

substrate/frame/support/src/traits/tokens/fungibles/regular.rs
seadanda: Ah I've just seen this, I think a few are missing, but that also explains why it's only in a small number of places.

bridges/modules/relayers/src/stake_adapter.rs
ggwpez: Yea they were not needed. Thanks for the comments!

seadanda (approved):

bkchr (approved):

Issue #708, opened 2023-02-22
candidate-validation subsystem makes bounded channels pointless


eskimor:
candidate-validation unconditionally [spawns tasks](https://github.com/paritytech/polkadot/blob/d9c8e6c1c7faa27b189220dee943f45e742c20bd/node/core/candidate-validation/src/lib.rs#L140) on each new message.

This is bad for at least 3 reasons:

1. It defeats the point of the bounded message channel, which is to back-pressure on the sending subsystem if necessary.
2. Related to 1, this unconditional spawning is essentially unbounded - hence we might run into situations where we run out of memory.
4. Due to the unboundedness also latency is unbounded.
3. It weakens our understanding of the overall system. The bounded channel size to candidate-validation is always 0, although it might already be severely loaded.

Superficially everything seems to work as is, but some checking on these unbounded queuing (there is more than one, if I remember correctly) is in order. It might be causing hidden issues.

How it should work: Indeed spawn a couple of workers in parallel, but track how many are active - once too many, start blocking on incoming messages.

sandreim:
implemented by https://github.com/paritytech/polkadot-sdk/pull/2125

Issue #599, opened 2023-07-14
Move ZSTD PVF decompression to a blocking task


alindima:
I noticed that pvf validation tasks, like `validate_from_exhaustive ` and `validate_from_chainstate` have a consistent duration of about 4-5 ms for `poll` (on versi).
This means that we're spending quite a lot of time doing blocking CPU work before yielding back to the runtime, which hogs an executor thread needlessly.
Based on the code, I think the vast majority of time is spent decompressing the PVF code and POV data, before the code dispatches the request to the PVF executor.

To mitigate this, we could move the decompression part to the `pvf-validation-host` blocking task. As I'm not very familiar with that part of the code, would it be worth it?

PR #2125, opened 2023-11-01
Make candidate validation bounded again

s0me0ne-unkn0wn:
This PR aims to channel the backpressure of the PVF host's preparation and execution queues to the candidate validation subsystem consumers.

Related: #708


alindima:
Maybe I don't have the full picture, but how does this PR introduce boundedness to the number of spawned tasks? AFAICT, the number is still not bounded. The tasks are just added to a FuturesUnordered collection. Maybe we want to check the size before pushing another one?

s0me0ne-unkn0wn:
It's totally possible that I'm doing something wrong as I'm not a Rust async code expert yet (but trying to become one :slightly_smiling_face:)

My idea was like this: when `validate_from_exhaustive()` (or another function communicating to the PVF host) is ready to submit a job to the PVF host, it `send`s a `ToHost` message to the PVF host through an mpsc channel, which is bounded. If the PVF host's message queue is full, that `send()` will block creating (or channeling) the backpressure, and the `await` on the `validate_from_exhaustive()` future wouldn't resolve until there's free buffer space in the PVF host's queue, channeling that backpressure further up. Does it work like that, or I'm getting the concept totally wrong?

s0me0ne-unkn0wn:
Ah, you probably mean that the number of added tasks is not bounded by itself, so no real backpressure is channeled up from the candidate validation subsys... Yes, looks like a problem I've overseen. I'll think about how to fix it (and ideas are appreciated).

alindima:
> Ah, you probably mean that the number of added tasks is not bounded by itself, so no real backpressure is channeled up from the candidate validation subsys... Yes, looks like a problem I've overseen. I'll think about how to fix it (and ideas are appreciated).

Yes, that's what I was referring to. I guess we need to add check the size of the FuturesUnordered collection before pushing another element there

s0me0ne-unkn0wn:
Well, it looked easy at first, but after a lot of research, reading, and chatting with people, I gave birth to this monster. I'd definitely appreciate any help in improving it.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -152,105 +153,109 @@ async fn run<Context>(
...
 		}
+	});
+
+	read_stream
+		// NB: Cloning `sender` inside `async` block of `for_each_concurrent` renders the whole
+		// thing not `Send`, so we `zip` the message stream with the stream of `sender` clones here.
+		.zip(stream::repeat(sender))
+		// FIXME: The backlog size here is the same as in PVF host queues. It should either use a
+		// common constant, or another appropriate value should be chosen
+		.for_each_concurrent(30, |message_and_sender| {
```
alexggh, line 179: Another thing is that `for_each_concurrent` is not equivalent with calling `ctx.spawn` several times, because 
>the closure are run concurrently (but not in parallel– this combinator does not introduce any threads).

ctx.spawn gives us the opportunity if this being possibly run in parallel, not sure if it really matters, it might not.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -152,105 +153,109 @@ async fn run<Context>(
...
-							let _ = response_sender.send(precheck_result);
-						}
-					};
-
-					ctx.spawn("candidate-validation-pre-check", bg.boxed())?;
-				},
+	let mut res = Ok(());
+	let sender = ctx.sender().to_owned();
+
+	let read_stream = stream::poll_fn(|c| loop {
```
alexggh, line 159: Another possible way for implementing this would be to use a async_semaphore.
```
let s = Semaphore::new(MAX_CONCURENT_REQUESTS))
```

s.acquire() before all the spawn and you move the resulting guard in the future.

alexggh (approved):
Looks good to me, left you some other ideas if you want to experiment with them.

polkadot/node/core/candidate-validation/src/lib.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: We can, but I'm not totally sure it makes sense. If the candidate validation messages were just retransmitted to the PVF host's queue, it would be fine. But some invariants (like `ValidateFromChainState`) require candidate validation subsys to do additional work involving runtime requests on which it `await`s. So, if 30 `ValidateFromChainState` messages arrive, they are processed concurrently, and the PVF host queue is empty. If some `ValidateFromExhaustive` arrives at the moment, it is waiting until a free slot appears. But it could indeed fall through into the queue as it doesn't require any additional processing. That's why I think the limit here could be higher than the queue limit.
On the other hand, 30 concurrent `ValidateFromChainState`s is an unlikely situation. The real message stream is mixed `Validate*` requests with a handful of pre-checking requests. So I don't know, I just cannot decide which value is appropriate here.

s0me0ne-unkn0wn: Speculatively, it has to be okay. What it actually does under the hood is to send a payload to the PVF host and then wait for the answer. All the actual processing is done by the PVF host within its own threading model. But I'm not 100% sure I'm not missing something not obvious here.

alindima: hmm, it also does code decompression, which can take a couple of ms: https://github.com/paritytech/polkadot-sdk/issues/599.
So that means that one future will block all other from making progress, since they are not being spawned on new tasls

s0me0ne-unkn0wn: Hmm, fair enough. And now we have a dilemma. Go back to the `FuturesUnordered` and add a semaphore, or move the decompression to the validation host (something we want to do anyway) and leave this PR as-is.

alexggh: Your call :D.

s0me0ne-unkn0wn: Benchmarked the decompression with the Moonbeam runtime (a large one), it takes 4.33 msec, not the end of the world, but still significant.

As far as I understand the problem, moving the decompression to the PVF host side (which requires some effort by itself) is only meaningful if we run the PVF host on the blocking pool? And if we ever switch to non-blocking (which we want to do), we'll still have the same problem with some executor threads being blocked in decompression?

If that's true, it's better not to move it but spawn it as a blocking task on the subsystem side. I benchmarked that just in case; it doesn't introduce any overhead and is a future-proof solution.

alindima: sounds good to me. or we can just use regular spawn like before and see if https://github.com/paritytech/polkadot-sdk/issues/599 is worth taking a look at separately.

maybe it's too much of a micro-optimisation that doesn't make any difference

mrcnski: > And if we ever switch to non-blocking (which we want to do), we'll still have the same problem with some executor threads being blocked in decompression?

> If that's true, it's better not to move it but spawn it as a blocking task on the subsystem side.

Sounds good to me too. I do think the PVF host should be non-blocking as it does no blocking work itself. It waits on the child processes but that shouldn't block.

s0me0ne-unkn0wn:
Okay, so here's take three: semaphores.

I didn't do anything with decompression yet (as it's a micro-optimization anyway and may be left for a follow-up). The problem is, IIUC, we cannot spawn tasks from other tasks. The subsystem context exposing `spawn()` and `spawn_blocking()` is not `clone`able, it cannot be moved into a task as the subsystem main loop needs it, and it cannot be borrowed in a task as `spawn()` takes `&mut self`. In C, I'd just pass a handler to `.spawn_blocking()` to the task and yolo ftw. Here, I don't see a way.

I could've spawned my own thread pool from a task to do parallel decompressions, but it sounds like firing a cannon at sparrows. I can also spawn a separate task on the blocking pool that will do decompressions, but in that case, it will do it FIFO; that is, tasks requesting decompression will not be blocked but will wait until all the decompression requests submitted by other tasks are processed.

All in all, it seems to be a more complex problem than it sounds, or I'm just not enlightened enough yet to invent a simple solution.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -153,106 +154,112 @@ async fn run<Context>(
...
-							let precheck_result = precheck_pvf(
-								&mut sender,
-								validation_host,
-								relay_parent,
-								validation_code_hash,
-							)
-							.await;
-
-							let _ = response_sender.send(precheck_result);
+		futures::select! {
```
alexggh, line 164: Why did we gave up on spawn I imagined we can do something like this, what am I missing ?
```
--- a/polkadot/node/core/candidate-validation/src/lib.rs
+++ b/polkadot/node/core/candidate-validation/src/lib.rs
@@ -152,6 +152,8 @@ async fn run<Context>(
        )
        .await;
        ctx.spawn_blocking("pvf-validation-host", task.boxed())?;
+       use async_semaphore::Semaphore;
+       let semaphore = Arc::new(Semaphore::new(10));
 
        loop {
                match ctx.recv().await? {
@@ -171,9 +173,12 @@ async fn run<Context>(
                                                let mut sender = ctx.sender().clone();
                                                let metrics = metrics.clone();
                                                let validation_host = validation_host.clone();
+                                               let guard = semaphore.acquire_arc().await;
 
                                                async move {
                                                        let _timer = metrics.time_validate_from_chain_state();
+                                                       let _guard = guard;
+
                                                        let res = validate_from_chain_state(
                                                                &mut sender,
                                                                validation_host,
@@ -205,9 +210,12 @@ async fn run<Context>(
                                        let bg = {
                                                let metrics = metrics.clone();
                                                let validation_host = validation_host.clone();
+                                               let guard = semaphore.acquire_arc().await;
 
                                                async move {
                                                        let _timer = metrics.time_validate_from_exhaustive();
+                                                       let _guard = guard;
+
                                                        let res = validate_candidate_exhaustive(
                                                                validation_host,
                                                                validation_data,
@@ -236,8 +244,11 @@ async fn run<Context>(
                                        let bg = {
                                                let mut sender = ctx.sender().clone();
                                                let validation_host = validation_host.clone();
+                                               let guard = semaphore.acquire_arc().await;
 
                                                async move {
+                                                       let _guard = guard;
+
```

alexggh: Never mind :D I was dumb we need the select if don't want to block signals processing while the queue is full.

alexggh (approved):

polkadot/node/core/candidate-validation/src/lib.rs
alindima: > I was dumb we need the select if don't want to block signals processing while the queue is full.

AFAICT, in its current form, this PR will still not process signals while the queue is full. The `select` only helps with the other tasks being processed concurrently with the overseer messages (if we were using spawn, we wouldn't need the select)

s0me0ne-unkn0wn: True, and as far as I understand, that's exactly what we want. When the queue is full, signal processing stops, thus channeling the backpressure upstream.

alindima: If we'd like to not block signal processing when the queue is full (which is really only for Conclude) we should add a separate task that handles `CandidateValidationMessage`s (and the main task sends work to it via a channel).

Not having this would prevent the subsystem from shutting down if the queue is full I guess. Not sure if it's a real problem with a queue size of 10.. idk

alindima: side note: maybe we should also have some tests that simulate a full queue, if that's feasible

s0me0ne-unkn0wn: > we should add a separate task that handles `CandidateValidationMessage`s (and the main task sends work to it via a channel).

Hmm, that's interesting, really interesting. CC @eskimor @rphmeier 

Let's consider a separate task that's processing only candidate validation messages. The main subsystem loop sends the messages to that task via a channel. There are two possibilities:
1) If the queue is full, that task ceases to receive new messages. In that case, the main loop `await`s on `send` to that task, and nothing changes in practice; the main loop is still blocked and doesn't process incoming signals.
2) The task swallows all the messages but only processes N validation jobs in parallel. Sounds good, but no backpressure then! That is, that design works exactly as it works right now, just with a limited number of tasks (which is probably better than now), but it doesn't channel any backpressure upstream.

Trying to solve that, I went to see how it works in other subsystems. [Backing main loop `await`s on `handle_communication`](https://github.com/paritytech/polkadot-sdk/blob/6b7be115fd523057d01cea46e9680a62a0b08d97/polkadot/node/core/backing/src/lib.rs#L356), so it doesn't process overseer signals during that handling either. But what happens next is even more interesting. I would expect that at some point, it sends a candidate validation message and `await`s the answer, thus creating backpressure on the backing subsystem. But in reality... Kaboom, [it's spawning a non-blocking task to handle validation](https://github.com/paritytech/polkadot-sdk/blob/6b7be115fd523057d01cea46e9680a62a0b08d97/polkadot/node/core/backing/src/lib.rs#L1713)! And the number of those tasks is unbounded as well, so the backing subsys doesn't even try to create that backpressure I'm trying to channel upstream in this PR!

I haven't had a look at other subsystems yet, but unless I'm missing something, it's not the candidate validation subsystem that makes bounded channels useless. The backpressure thing seems broken at many points.

eskimor (approved):

polkadot/node/core/candidate-validation/src/lib.rs
eskimor: Not processing overseer signal sounds worse than it is. Signal processing is prioritized and there is only a very limited amount. Also the signal channel has size 50 (if I remember correctly). If this ever gets full, you have a bug.

If the task you are blocking on is not blocking for something crazy like 30 seconds, but will always finish at least in the single digit second range, then there should be no problem ever.

Let's assume we get really a lot of signals (10 per slot). This means to get the signal channel full (to actually back pressure), even in this scenario we have 5*6s = 30s time. If your task is faster than this, we will be processing all that queued up signals before processing anything else. Therefore the next time we wait, the signal queue is empty and we have another 30s (pessimistically).

Therefore in practice, assuming signals themselves are not inducing work, we only really back pressure on messages, not signals.

That being said, we could be smarter in back pressure with candidate validation. E.g. the backing subsystem, when it notices a full channel to candidate validation - it could acknowledge that candidate validation is loaded and just drop the message.

s0me0ne-unkn0wn: > If the task you are blocking on is not blocking for something crazy like 30 seconds

It does :sweat_smile: 
Pre-checking requests go the same way, and they have 60-sec hard timeout by default.
So it's still better to make them async. Not merging this one yet, will make some improvements.

sandreim: I think we just need to rewrite in way that we only allow a bounded amount of tasks to be running at a certain time. Also, the semaphore is not the way I would do it. I propose we implement backpressure differently, by returing something like Err(BusyPleaseRetry). This allows us to still process signals,but we'll have to handle this backpressure in backing/approval voting/disputes.

sandreim: Alternatively, we could downgrade `candidate-validation` to be a library rather than subsystem, and then yes, we could use this semaphore approach to limit how many run in parallel and block at caller.

eskimor: This is essentially the same as I am proposing, but with more complexity. Letting the channels go full, does exactly that: "Err(BussyPleaseRetry)" ... you can either retry/drop or block, if you really need it.

What would be the problem adding a call that only receives signals? If I remember correctly the normal recv is a select over signals and messages - should be super trivial to add this.

sandreim: Yeah, that is easy to add

eskimor: @s0me0ne-unkn0wn you wanna take a stab? Apart from finding the code in that macro soup, the actual addition should be trivial.

s0me0ne-unkn0wn: @eskimor no problem, I just need to understand your idea better. Do you mean you want to introduce methods like `ctx.recv_signal()` and `ctx.recv_non_signal()` to be able to process them in parallel? Because if we have normal `ctx.recv()` along with `ctx.recv_signal()`, it looks somewhat redundant, we'll get the same signals in two places.

Also, are there any points where sequential processing of signal and non-signal messages can make a difference? For example, if some message relies on the leaf activation signal being already processed by the subsystem?

mrcnski: Maybe a dumb question but how exactly do signals and messages differ? Signals only notify of some change in state (i.e. shutdown, leaf update, etc)? Do they only originate from the overseer or from somewhere else as well?

eskimor: Exactly signals originate from the overseer and are prioritized.

The suggestion is to have an additional call recv_signal(), which only receives signals, but no messages ever. Indeed, we could make it three functions in total:

- recv ... the existing call, receiving everything.
- recv_signal ... only receiving signals.
- recv_msg .... only receiving messages.

Then you would either use recv or the recv_signal/recv_msg combination, giving you more control

Prioritizing signals even more than what is done by default should not do any harm. The overseer already ensures that you are never receiving messages that have been sent after a leaf activation, before you have seen that signal as well.

mrcnski: > Exactly signals originate from the overseer and are prioritized.

Is it possible to document this @s0me0ne-unkn0wn? Also, the docs for `OverseerSignal` are not very helpful: "Signals sent by an overseer to a subsystem".

drahnr: Signals are sent as a form of broadcast to all subsystems by integration code, here it's (from memory) from the substrate block import pipeline (with a few intermediate steps).

sandreim: I think the better solution is to change orchestra to allow implementation of subsystems that don't really care about signals, maybe except `Conclude`. What I mean is that we either have a per subsystem configurable `SIGNAL_TIMEOUT`, or annotate these subsystems such that `broadcast_signal` only forwards them `Conclude`. Then we'd have to be careful to ensure nodes get stuck at shutdown because of this.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4818226

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -130,6 +132,105 @@ impl<Context> CandidateValidationSubsystem {
...
+			candidate_receipt,
+			pov,
+			executor_params,
+			exec_kind,
+			response_sender,
+			..
+		} => {
+			// let mut sender = ctx.sender().clone();
+			// let metrics = metrics.clone();
+			// let validation_host = validation_host.clone();
```
rphmeier, line 155: some commented out code remains

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4822980

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827629

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4835830

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -81,6 +82,8 @@ const PVF_APPROVAL_EXECUTION_RETRY_DELAY: Duration = Duration::from_secs(3);
 #[cfg(test)]
 const PVF_APPROVAL_EXECUTION_RETRY_DELAY: Duration = Duration::from_millis(200);
 
+const TASK_LIMIT: usize = 30;
```
sandreim, line 85: A comment describing this arbitrary value is useful.

sandreim (approved):
Currently `message_channel_capacity` is 2048, if we really want to backpressure here, we'd have to make the channel size much smaller.

s0me0ne-unkn0wn:
@sandreim thanks for the review! I'm not going to merge before the new `orchestra` crate is released anyway. I've temporarily switched everything to the orchestra's `master` branch in this PR to see how CI reacts.

s0me0ne-unkn0wn:
And I wonder if we should push further to achieve the real backpressure.

@rphmeier what's the point of spawning background tasks for the validation in the backing subsystem? If we wanted a real backpressure, we'd need to make them foreground so that backing could be blocked if the candidate validation queue is overflowed. Leaf activations could still be processed separately in that case thanks to the new separate signal processing mechanism, but other messages like `CanSecond` would also be blocked, thus backpressuring on the upstream subsystems. 

That all sounds like a correct backpressure chain to me, but I'm not sure if some time-critical work the backing subsystem does should be prioritized over the backpressure correctness.

s0me0ne-unkn0wn:
I gave it a local zombienet burn-in, 10 hours, 20 validators, 6 parachains. Seems to be stable.

PR #2135, opened 2023-11-02
Adding gitspiegel-trigger workflow

mutantcornholio:
GitHub has a setting that requires manual click for executing GHA on the branch, for the first-time contributors: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks.

After this PR, gitspiegel will respect that setting. So, for PRs from first-time contributors, gitspiegel won't do mirroring until the button in PR is clicked.    More info: https://github.com/paritytech/gitspiegel/issues/169

### DIFF
--- a/.github/workflows/gitspiegel-trigger.yml
+++ b/.github/workflows/gitspiegel-trigger.yml
@@ -1,0 +1,20 @@
+name: gitspiegel sync
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - unlocked
+      - ready_for_review
+      - reopened
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


alvicsam (approved):

mutantcornholio:
**UPD:** The first attept to use a workflow to protect GitLab CI from untrusted contributors failed, because GitHub doesn't pass secrets to workflows for PRs that originate from forks. 
 
This uses a different approach: instead of triggerring gitspiegel API directly from the workflow, we're just spawning an empty workflow with a specific path, and gitspiegel listens for `workflow_run` event to start mirroring.  

The idea is the same: for the first-time contributors, running workflows would require manual aciton and that would block mirroring. But this time, we don't need any secrets to make it work.

sergejparity (approved):

PR #2139, opened 2023-11-02
[testnet] Allow governance to control fees for Rococo <> Westend bridge

svyatonik:
Right now governance could only control byte-fee component of Rococo <> Westend message fees (paid at Asset Hubs). This PR changes it a bit:
1) governance now allowed to control both fee components - byte fee and base fee;
2) base fee now includes cost of "default" delivery and confirmation transactions, in addition to `ExportMessage` instruction cost.


bridges/primitives/chain-bridge-hub-rococo/src/lib.rs
```diff
@@ -82,3 +82,18 @@ pub const WITH_BRIDGE_ROCOCO_TO_WESTEND_MESSAGES_PALLET_INDEX: u8 = 51;
 
 decl_bridge_finality_runtime_apis!(bridge_hub_rococo);
 decl_bridge_messages_runtime_apis!(bridge_hub_rococo);
+
```
svyatonik, line 85: I've moved all fee-related constants from `asset-hub` primitives to `bridge-hub` primitives because:
1) now asset hubs are using different different crates to compute total fee (because delivery tx happens at the other chain);
2) tests for those values reside in bridge hubs

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs
```diff
@@ -966,7 +986,7 @@ pub mod bridging {
 					// base delivery fee to local `BridgeHub`
 					Some((
 						XcmBridgeHubRouterFeeAssetId::get(),
-						bp_asset_hub_wococo::BridgeHubWococoBaseFeeInWocs::get(),
+						XcmBridgeHubRouterBaseFee::get(),
```
svyatonik, line 989: Technically incorrect (`bp_asset_hub_wococo` -> `bp_asset_hub_rococo`), but Wococo has been dropped, so this code will die soon

bridges/bin/runtime-common/src/messages_generation.rs
```diff
@@ -104,6 +105,38 @@ where
 	(root, storage_proof)
 }
 
+/// Prepare storage proof of given messages delivery.
```
svyatonik, line 108: This PR is the first PR, where this code is changed => no review for bridges changes yet. But I've just moved code around

serban300 (approved):

acatangiu (approved):

PR #2144, opened 2023-11-03
substrate: sysinfo: Expose failed hardware requirements

alexggh:
The check_hardware functions does not give us too much information as to what is failing, so let's return the list of failed metrics, so that callers can print it.

This would make debugging easier, rather than try to guess which dimension is actually failing.


bkchr (approved):
nice :+1:

ordian (approved):

tdimitrov (approved):

PR #2146, opened 2023-11-03
Configure MessageQueue for compatibility with Snowbridge

vgeddes:
Changes:
* Add `Export` variant to AggregateMessageOrigin enum, to identify queues owned by the[ snowbridge-outbound-queue](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/outbound-queue/src/lib.rs) pallet.
* Change registration of MessageQueue as the last pallet in BridgeHub runtimes, to ensure correct `on_initialize` order. The [on_initialize](https://github.com/Snowfork/snowbridge/blob/ab6d7a00e6888aacba8032cb39c524801f714391/parachain/pallets/outbound-queue/src/lib.rs#L265) handler of `snowbridge-outbound-queue` needs to run first.

Checklist:
- [ ] Ensure tests and linters pass

### DIFF
--- a/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
+++ b/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
@@ -468,11 +468,13 @@ construct_runtime!(
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin, Config<T>} = 31,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
-		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
+
+		// MessageQueue. Comes last so that its on_initialize handler runs last
+		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 100,
 	}
 );
 
--- a/cumulus/primitives/core/src/lib.rs
+++ b/cumulus/primitives/core/src/lib.rs
@@ -102,7 +102,7 @@ pub enum MessageOrigin {
 	/// The message came from the relay-chain.
 	Parent,
 	/// The message came from a sibling para-chain.
-	Sibling(ParaId)
+	Sibling(ParaId),
 }
 
 impl From<AggregateMessageOrigin> for xcm::v3::MultiLocation {
@@ -111,18 +111,10 @@ impl From<AggregateMessageOrigin> for xcm::v3::MultiLocation {
 		match origin {
 			Here => MultiLocation::here(),
 			Parent => MultiLocation::parent(),
-			Sibling(id) => {
-				MultiLocation::new(1, Junction::Parachain(id.into()))
-			},
-			Export(MessageOrigin::Here) => {
-				MultiLocation::here()
-			},
-			Export(MessageOrigin::Parent) => {
-				MultiLocation::parent()
-			},
-			Export(MessageOrigin::Sibling(id)) => {
-				MultiLocation::new(1, Junction::Parachain(id.into()))
-			}
+			Sibling(id) => MultiLocation::new(1, Junction::Parachain(id.into())),
+			Export(MessageOrigin::Here) => MultiLocation::here(),
+			Export(MessageOrigin::Parent) => MultiLocation::parent(),
+			Export(MessageOrigin::Sibling(id)) => MultiLocation::new(1, Junction::Parachain(id.into())),
 		}
 	}
 }
--- a/cumulus/primitives/core/src/lib.rs
+++ b/cumulus/primitives/core/src/lib.rs
@@ -91,15 +91,38 @@ pub enum AggregateMessageOrigin {
 	///
 	/// This is used by the HRMP queue.
 	Sibling(ParaId),
+	/// This is used by the snowbridge-outbound-queue pallet
+	Export(MessageOrigin),
+}
+/// The origin of a message
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
+pub enum MessageOrigin {
+	/// The message came from the para-chain itself.
+	Here,
+	/// The message came from the relay-chain.
+	Parent,
+	/// The message came from a sibling para-chain.
+	Sibling(ParaId)
 }
 
 impl From<AggregateMessageOrigin> for xcm::v3::MultiLocation {
 	fn from(origin: AggregateMessageOrigin) -> Self {
+		use AggregateMessageOrigin::*;
 		match origin {
-			AggregateMessageOrigin::Here => MultiLocation::here(),
-			AggregateMessageOrigin::Parent => MultiLocation::parent(),
-			AggregateMessageOrigin::Sibling(id) =>
-				MultiLocation::new(1, Junction::Parachain(id.into())),
+			Here => MultiLocation::here(),
+			Parent => MultiLocation::parent(),
+			Sibling(id) => {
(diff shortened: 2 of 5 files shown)


cumulus/primitives/core/src/lib.rs
```diff
@@ -91,15 +91,30 @@ pub enum AggregateMessageOrigin {
 	///
 	/// This is used by the HRMP queue.
 	Sibling(ParaId),
+	/// This is used by the snowbridge-outbound-queue pallet
+	Export(MessageOrigin),
+}
+/// The origin of a message
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
+pub enum MessageOrigin {
```
ggwpez, line 99: What is with the naming here? Just `MessageOrigin` is not very expressive.

cumulus/primitives/core/src/lib.rs
```diff
@@ -91,15 +91,30 @@ pub enum AggregateMessageOrigin {
...
+	/// The message came from the para-chain itself.
+	Here,
+	/// The message came from the relay-chain.
+	Parent,
+	/// The message came from a sibling para-chain.
+	Sibling(ParaId),
```
ggwpez, line 105: Why are these variants the same as in the `AggregateMessageOrigin`?

cumulus/primitives/core/src/lib.rs
```diff
@@ -91,15 +91,30 @@ pub enum AggregateMessageOrigin {
 	///
 	/// This is used by the HRMP queue.
 	Sibling(ParaId),
+	/// This is used by the snowbridge-outbound-queue pallet
+	Export(MessageOrigin),
```
ggwpez, line 95: I hope we can make this more general somehow. But just so that i understand, this is the outbound queue for bridged messages?  
Will there also be an inbound?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -468,11 +468,13 @@ construct_runtime!(
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin, Config<T>} = 31,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
-		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
+
+		// MessageQueue. Comes last so that its on_initialize handler runs last
```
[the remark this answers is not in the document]
vgeddes: Its necessary as we need to install a pallet [OutboundQueue](https://github.com/Snowfork/snowbridge/tree/main/parachain/pallets/outbound-queue/src), documented [here](https://github.com/Snowfork/snowbridge/blob/c9196e196f650e1362dfdf7e911dc9b292cf0859/parachain/pallets/outbound-queue/src/lib.rs#L3), with an [on_initialize](https://github.com/Snowfork/snowbridge/blob/c9196e196f650e1362dfdf7e911dc9b292cf0859/parachain/pallets/outbound-queue/src/lib.rs#L267) handler that must run before any messages are consumed from `MessageQueue`.

This `on_initialize` handler kills storage (obsolete processed messages) from the previous block.

cumulus/primitives/core/src/lib.rs
vgeddes: Our Ethereum bridge acts as an adaptor over XCM, essentially allowing parachains to send XCM messages to Ethereum.

On BridgeHub, to process these XCM messages, we need to track their origin, with the granularity shown in  `MessageQueue`.

Messages from different Polkadot origins are treated differently, for example in this implementation of QueuePausedQuery: https://github.com/Snowfork/snowbridge/blob/c9196e196f650e1362dfdf7e911dc9b292cf0859/parachain/pallets/outbound-queue/src/queue_paused_query_impl.rs#L17

cumulus/primitives/core/src/lib.rs
vgeddes: > I hope we can make this more general somehow.

I thought about this, came to the conclusion that a generalized solution wouldn't really help, as we can't predict the requirements (or even a need) for the generalized solution. So we could end designing something that will never be used by any projects other than Snowbridge.

But just so that i understand, this is the outbound queue for bridged messages?

Yes. Essentially it takes inbound XCM messages, processes them, and puts them back on the `MessageQueue` for eventual final processing and commitment. Full documentation [here](https://github.com/Snowfork/snowbridge/blob/c9196e196f650e1362dfdf7e911dc9b292cf0859/parachain/pallets/outbound-queue/src/lib.rs#L5). 

> Will there also be an inbound?

Nope. 

> I am not sure if we actually need to change this in the SDK, or if there could be an enum type in the runtime that has this `AggregatedMessageOrigin` as one variant and then a bunch of other custom variants for your snowbridge pallets.

I kinda originally tried something like that in the other old draft PR. Though as you said, it would increase the complexity of runtime upgrades.
```rust
enum AggregateMessageOrigin {
   Xcmp(XcmpOrigin)
   Snowbridge(SnowbridgeOrigin)
}
```

ggwpez: Can I see the snowbridge MR to the Polkadot-SDK please? To me it seems like you change this enum but we will never use it in the SDK.  
In that case you can just change it in your snowbridge runtime.

vgeddes: Just to clarify, Snowbridge is a set of pallets which will be installed on these system runtimes:
* `bridge-hub-rococo`
* `bridge-hub-kusama`
* `bridge-hub-polkadot`

So within those runtimes, the XCMP and Snowbridge subsystems will need to agree on a common `AggregateMessageOrigin`, unless I'm missing something:

For example here we install the pallets in the `bridge-hub-rococo` runtime: https://github.com/Snowfork/polkadot-sdk/blob/71197605d792abca283c01c5873bfe8e1dba83c8/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L681-L684

Here we configure the `MessageQueue` to use our `OutboundQueue` as a message processer: https://github.com/Snowfork/polkadot-sdk/blob/71197605d792abca283c01c5873bfe8e1dba83c8/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L493-L495

And here's our interim `AggregateMessageOrigin`: 
https://github.com/Snowfork/snowbridge/blob/c9196e196f650e1362dfdf7e911dc9b292cf0859/parachain/primitives/core/src/outbound.rs#L364


Is hard to give you a clean MR right now, as the base branch for our working branch is a bit of frankenstein at the moment.

cumulus/primitives/core/src/lib.rs
```diff
@@ -91,15 +91,30 @@ pub enum AggregateMessageOrigin {
...
+	Parent,
+	/// The message came from a sibling para-chain.
+	Sibling(ParaId),
 }
 
 impl From<AggregateMessageOrigin> for xcm::v3::MultiLocation {
 	fn from(origin: AggregateMessageOrigin) -> Self {
+		use AggregateMessageOrigin::*;
 		match origin {
-			AggregateMessageOrigin::Here => MultiLocation::here(),
```
vgeddes, line 99: If we design the `AggregateMessageOrigin` to look like this:

```rust
enum AggregateMessageOrigin {
   Xcmp(XcmpOrigin)
   Snowbridge(SnowbridgeOrigin)
}
```

Then this change would be less hacky, as we would only need to implement `From<XcmpOrigin> for Multilocation`.

cumulus/primitives/core/src/lib.rs
ggwpez: But then there should maybe a uniform `AggregateMessageOrigin` across all `bridge-hub` runtimes. So it should be in a `bridge-hub-common` crate or something. It does not make sense to force this onto parachain that will never use it.

vgeddes: That makes very good sense, narrowing the change down to the BridgeHub level. Let me try that.

vgeddes:
Closing off. Will open a new PR with the suggested changes.

PR #2154, opened 2023-11-04
[FRAME] Enforce inherent order

ggwpez:
Currently the inherent order is not enforced by the runtime. This means that the block builder can push the inherents in any order. A mechanism is introduced to prevent this. Note that this change is not needed to make this MR complete; it is just added as bonus. This completely locks down the inclusion and order of inherents and gives the BB no room to fudge any of the inherents anymore (not in order nor existence). 

Changes:
- Move executive tests to own file
- Trait `ProvideInherent` gets a new function `fn inherent_order() -> Option<InherentOrder>` that can set an explicit order of an inherent. `None` indicates that the pallet index should be used.
- Executive enforces this order by requiring trait `EnsureInherentsAreOrdered` and calling `ensure_inherents_are_ordered` immediately after `ensure_inherents_are_first`.


substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -204,13 +207,59 @@ pub fn expand_outer_inherent(
...
 		}
 
+		impl #scrate::traits::EnsureInherentsAreOrdered<#block> for #runtime {
+			fn ensure_inherents_are_ordered(block: &#block, num_inherents: usize) -> Result<(), #scrate::__private::sp_inherents::InherentOrderError> {
+				use #scrate::inherent::ProvideInherent;
+				use #scrate::traits::{IsSubType, ExtrinsicCall};
+				use #scrate::sp_runtime::traits::Block as _;
+				use #scrate::__private::sp_inherents::{InherentOrder, InherentOrderError};
+
+				// The last order that we encountered. Must increase monotonic.
```
liamaharon, line 217:
```suggestion
				// The last order that we encountered. Must increase monotonically.
```

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -204,13 +207,59 @@ pub fn expand_outer_inherent(
...
+						return Err(InherentOrderError::InherentWithoutPallet);
+					};
+
+					if let Some(last) = last {
+						if last >= current {
+							return Err(InherentOrderError::OutOfOrder(last, current));
+						}
+					}
+
+					last = Some(current);
```
liamaharon, line 249: Maybe some comments here

substrate/frame/support/src/inherent.rs
```diff
@@ -81,6 +82,13 @@ pub trait ProvideInherent {
 		Ok(())
 	}
 
+	/// The order of the inherent relative to all other inherents.
+	///
+	/// A return value of `None` indicates that the pallet index should be used.
+	fn inherent_order() -> Option<InherentOrder> {
+		None
+	}
```
liamaharon, line 90: Maybe 'priority' instead of 'order' is better description for the api.

So, inherents are 'ordered' by 'priority'. 

```suggestion
	fn inherent_priority() -> Option<InherentPriority> {
		None
	}
```

substrate/primitives/inherents/src/lib.rs
```diff
@@ -367,6 +369,53 @@ impl PartialEq for CheckInherentsResult {
 	}
 }
 
+/// The order of an inherent.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Encode, Decode, scale_info::TypeInfo)]
+pub enum InherentOrder {
+	/// The first inherent.
+	First,
```
liamaharon, line 376: isn't index 0 always first? why do we need a special case for that?

substrate/frame/support/src/inherent.rs
```diff
@@ -81,6 +82,13 @@ pub trait ProvideInherent {
 		Ok(())
 	}
 
+	/// The order of the inherent relative to all other inherents.
+	///
+	/// A return value of `None` indicates that the pallet index should be used.
+	fn inherent_order() -> Option<InherentOrder> {
+		None
```
librelois, line 89: I think that defining the order at pallet level is not good design. It's up to the runtime to define the order of inherents, because for the same pallet set, different runtimes will have different requirements.

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -204,13 +207,59 @@ pub fn expand_outer_inherent(
...
+
+				for (i, xt) in block.extrinsics().iter().take(num_inherents).enumerate() {
+					let call = <#unchecked_extrinsic as ExtrinsicCall>::call(xt);
+
+					let mut current: Option<InherentOrder> = None;
+
+					#(
+						if let Some(call) = IsSubType::<_>::is_sub_type(call) {
+							if #pallet_names::is_inherent(&call) {
+								let order = #pallet_names::inherent_order().unwrap_or(InherentOrder::Index(#pallet_indices as u32));
```
librelois, line 228: Inherent order shouldn't depend on pallet indexes, because pallet indexes can't be modified for a chain that's already been in production for a long time (this breaks metadata and SCALE encoding). I suggest instead defaulting to the order of declaration in the construct_runtime macro, and allowing the runtime to define a different explicit order.

PR #2155, opened 2023-11-05
Do not panic if the `fdlimit` call to increase the file descriptor limit fails

nazar-pc:
# Description

Sometimes changing file descriptor limits is not allowed, but there is no need to crash the node if/when this happens. Since `fdlimit`'s author decided to use panics instead of returning `Result`, we need to catch it.

# Checklist

  required)
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)


koute:
Looks like this is actually... [our crate](https://github.com/paritytech/fdlimit), in which case the proper fix is to just fix that crate instead of adding a `catch_panic` here.

nazar-pc:
Would you prefer `Option<Result<u64, E>>` or just `Result<Outcome, E>` there as an API?

bkchr:
> just `Result<Outcome, E>` there as an API?

Looks better.

nazar-pc:
Opened https://github.com/paritytech/fdlimit/pull/10, please check if it compiles on macOS, I'm on Linux myself

nazar-pc:
Updated with new `fdlimit` version, lock file was apparently outdated, `rand` got updated there by cargo

bkchr (approved):

koute (approved):

PR #2156, opened 2023-11-05
Tracking allocator: mark `Spinlock::unlock()` as unsafe and provide a safety contract

s0me0ne-unkn0wn:
As [it was discussed](https://github.com/paritytech/polkadot-sdk/pull/1192#discussion_r1374320227) in #1192, explicitly unlocking the allocator's spinlock is not indeed a safe operation. This PR declares it `unsafe` and provides a safety contract.

### DIFF
--- a/polkadot/node/tracking-allocator/src/lib.rs
+++ b/polkadot/node/tracking-allocator/src/lib.rs
@@ -73,7 +73,8 @@ impl<T> Spinlock<T> {
 	}
 
 	// SAFETY: It should be only called from the guard's destructor. Calling it explicitly while
-	// the guard is alive is undefined behavior.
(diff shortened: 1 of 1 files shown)


alexggh (approved):

koute (approved):

polkadot/node/tracking-allocator/src/lib.rs
```diff
@@ -72,8 +72,10 @@ impl<T> Spinlock<T> {
 		}
 	}
 
+	// SAFETY: It should be only called from the guard's destructor. Calling it explicitly while
+	// the guard is alive is undefined behavior.
```
mrcnski, line 76: Would be nice to explain why it's UB. After all, it wasn't obvious until we saw koute's example which segfaults.

mrcnski (approved):

PR #2167, opened 2023-11-06
add pallet nomination-pools versioned migration to kitchensink

brunopgalvao:
The versioned migrations are already there in pallet nomination-pools:
https://github.com/paritytech/polkadot-sdk/blob/f6ee4781f633f0f89598f7b230595afe401da8dc/substrate/frame/nomination-pools/src/migration.rs#L27-L48

Just updating the kitchensink runtime to point to them.

This is also nice because it points the dev to an example of how to use `VersionedMigration`.


liamaharon (approved):
Good idea. We should show best practice where ever we can.

While we're here, can we rename that `versioned_migrations` module to just `versioned`? It's already clear in a `migration` module so no requirement to repeat that info.

brunopgalvao:
> While we're here, can we rename that `versioned_migrations` module to just `versioned`? It's already clear in a `migration` module so no requirement to repeat that info.

Updated accordingly.

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -2196,7 +2196,7 @@ pub type Executive = frame_executive::Executive<
 // All migrations executed on runtime upgrade as a nested tuple of types implementing
 // `OnRuntimeUpgrade`.
```
ggwpez, line 2197: Maybe also put a comment that these are just examples and do not need to run directly after the genesis block?

ggwpez (approved):

substrate/bin/node/runtime/src/lib.rs
brunopgalvao: Done.

PR #2168, opened 2023-11-06
`serde_json`: bumped to 1.0.108

michalkucharczyk:
This PR updates the version of `serde_json` to `1.0.108` throughout the codebase.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1470,8 +1470,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "93f2635620bf0b9d4576eb7bb9a38a55df78bd1205d26fa994b25911a69f212f"
 dependencies = [
  "bitcoin_hashes",
- "rand 0.7.3",
- "rand_core 0.5.1",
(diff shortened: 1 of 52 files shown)


michalkucharczyk:
Maybe we should just use `1.0` version?

ggwpez (approved):
>Maybe we should just use 1.0 version everywhere?

Maybe we should be using workspace dependencies https://github.com/paritytech/polkadot-sdk/pull/2070, but i dont think there was an agreement yet.

bkchr (approved):

bkchr:
> Maybe we should just use `1.0` version everywhere?

You should also include the patch version in the `Cargo.toml` to ensure that downstream users of your crate are using the same version. So downstream includes some fix or some new function as well.

michalkucharczyk:
> > Maybe we should just use `1.0` version everywhere?
> 
> You should also include the patch version in the `Cargo.toml` to ensure that downstream users of your crate are using the same version. So downstream includes some fix or some new function as well.

Yeah, good point. Haven't thought of this.

PR #1644, opened 2023-09-20
Add availability-recovery from systematic chunks

alindima:
**Don't look at the commit history, it's confusing, as this branch is based on another branch that was merged**

Fixes #598 
Also implements [RFC #47](https://github.com/polkadot-fellows/RFCs/pull/47)

## Description

- Availability-recovery now first attempts to request the systematic chunks for large POVs (which are the first ~n/3 chunks, which can recover the full data without doing the costly reed-solomon decoding process). This has a fallback of recovering from all chunks, if for some reason the process fails. Additionally, backers are also used as a backup for requesting the systematic chunks if the assigned validator is not offering the chunk (each backer is only used for one systematic chunk, to not overload them).
    - Quite obviously, recovering from systematic chunks is much faster than recovering from regular chunks (4000% faster as measured on my apple M2 Pro).
- Introduces a `ValidatorIndex` -> `ChunkIndex` mapping which is different for every core, in order to avoid only querying the first n/3 validators over and over again in the same session. The mapping is the one described in RFC 47.
- The mapping is feature-gated by the [NodeFeatures runtime API](https://github.com/paritytech/polkadot-sdk/pull/2177) so that it can only be enabled via a governance call once a sufficient majority of validators have upgraded their client. If the feature is not enabled, the mapping will be the identity mapping and backwards-compatibility will be preserved.
- Adds a new chunk request protocol version (v2), which adds the ChunkIndex to the response. This may or may not be checked against the expected chunk index. For av-distribution and systematic recovery, this will be checked, but for regular recovery, no. This is backwards compatible. First, a v2 request is attempted. If that fails during protocol negotiation, v1 is used.
- Systematic recovery is only attempted during approval-voting, where we have easy access to the core_index. For disputes and collator pov_recovery, regular chunk requests are used, just as before.

## Performance results

Some results from subsystem-bench:

with regular chunk recovery: CPU usage per block 39.82s
with recovery from backers: CPU usage per block 16.03s
with systematic recovery: CPU usage per block 19.07s

End-to-end results here: https://github.com/paritytech/polkadot-sdk/issues/598#issuecomment-1792007099

#### TODO:

- [x] [RFC #47](https://github.com/polkadot-fellows/RFCs/pull/47)
- [x] merge https://github.com/paritytech/polkadot-sdk/pull/2177 and rebase on top of those changes
- [x] merge https://github.com/paritytech/polkadot-sdk/pull/2771 and rebase 
- [x] add tests
[...]

Issue #628, opened 2023-05-22
approval-voting: enable v2 assignments


sandreim:
Once https://github.com/paritytech/polkadot/pull/6782 is merged both v1/v2 will be supported. The plan is to enable v2 assignment generation once enough validators have upgraded but only for Kusama and the other testnets. Once we rule out any issues we should also enable on Polkadot.

alexggh:
Assignment v2 are enabled on kusama, next will be Polkadot.

alexggh:
Enabled on polkadot with: https://polkadot.subsquare.io/referenda/888.

PR #2177, opened 2023-11-06
add NodeFeatures field to HostConfiguration and runtime API

alindima:
Adds a `NodeFeatures` bitfield value to the runtime `HostConfiguration`, with the purpose of coordinating the enabling of node-side features, such as:  https://github.com/paritytech/polkadot-sdk/issues/628 and https://github.com/paritytech/polkadot-sdk/issues/598.
These are features that require all validators enable them at the same time, assuming all/most nodes have upgraded their node versions.

This PR doesn't add any feature yet. These are coming in future PRs.

Also adds a runtime API for querying the state of the client features and an extrinsic for toggling a feature by its index in the bitfield.

Note: originally part of: https://github.com/paritytech/polkadot-sdk/pull/1644, but posted as standalone to be reused by other PRs until the initial PR is merged


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4229449

alexggh:
>Open question: Note that this PR does NOT add a way of enabling any feature. Since this is a bitfield, I believe that an interface that would enable opengov/sudo to supply the entire u64 as input would be error-prone and a leaky abstraction. IMO it's best to add setters extrinsics on a per-feature basis. There are differing views: https://github.com/paritytech/polkadot-sdk/pull/1644#discussion_r1368508085

Thank you @alindima for opening this, let me describe what I was thinking about.

My idea was that this should become is a generic way for us to enable different behaviours of the node(features) via a runtime API call, the way I would imagine the flow once we have such an API would be:

1. I go in the node codebase and pick up  an unused flag(bit).
2. Code the behaviour in the node based on that flag.
3. Release the node.
4. Once we are ready to enable the feature someone creates a referendum that sets the flag(bit).
5. Optional: I would leave the door open for disabling features as well via a referendum.

I think there are benefits of having such a flow implemented end-to-end, like:
1. No need for every person to go and add a new api for each feature flag.
2. No need to have to coordonate node and runtime releases for things that don't impact each other.

Of course, there are downsides as well, the most dangerous that I think you already spelled out is wrongly enabling/disabling the wrong bit/flag, but I'm hopping that gets seriously mitigated by the fact that it is a referendum after all and it should have several pairs of eyes looking at it or maybe we can find a way to make it user friendly.

alindima:
> My idea was that this should become is a generic way for us to enable different behaviours of the node(features) via a runtime API call, the way I would imagine the flow once we have such an API would be:

Just a small clarification: it's not a runtime API call, but an extrinsic. The runtime API for querying the client features will stay the same regardless of how many of them we add.

I see your point and agree that it would be nice to have this end-to-end on the client side, without a runtime upgrade. 👍🏻 

It would be great if we could somehow couple the no-runtime-upgrade path with the nice, hard-to-misuse interface. From the looks of it, we have to pick one :(
Let's also see what others think

polkadot/node/core/runtime-api/src/cache.rs
```diff
@@ -446,6 +448,14 @@ impl RequestResultCache {
 		self.minimum_backing_votes.insert(session_index, minimum_backing_votes);
 	}
 
+	pub(crate) fn client_features(&mut self, key: &Hash) -> Option<vstaging::ClientFeatures> {
+		self.client_features.get(key).copied()
+	}
+
+	pub(crate) fn cache_client_features(&mut self, key: Hash, features: vstaging::ClientFeatures) {
```
eskimor, line 455: I don't think we need those to be able to change on a per block basis. We should make ClientFeatures session buffered.

polkadot/runtime/parachains/src/configuration.rs
```diff
@@ -261,6 +261,8 @@ pub struct HostConfiguration<BlockNumber> {
 	/// The minimum number of valid backing statements required to consider a parachain candidate
 	/// backable.
 	pub minimum_backing_votes: u32,
+	/// Client features enablement.
+	pub client_features: ClientFeatures,
```
eskimor, line 265: In fact it is part of the `HostConfiguration` which is already session buffered. So caching by SessionIndex should work already.

bkchr:
> I see your point and agree that it would be nice to have this end-to-end on the client side, without a runtime upgrade. 👍🏻

I'm also not entirely on board that we need this? This will then require random guessing if enough validators have upgraded to set this flag. Why isn't this done by bumping the networking protocol version and then let the validators discover on their own if they can use the new feature? Like all the nodes they would need to fetch the nodes of are supporting the new networking protocol version, so let's go. This will also make the separation more clean.

alexggh:
> > I see your point and agree that it would be nice to have this end-to-end on the client side, without a runtime upgrade. 👍🏻
> 
> I'm also not entirely on board that we need this? This will then require random guessing if enough validators have upgraded to set this flag. Why isn't this done by bumping the networking protocol version and then let the validators discover on their own if they can use the new feature? Like all the nodes they would need to fetch the nodes of are supporting the new networking protocol version, so let's go. This will also make the separation more clean.

Wouldn't this approach works just for situations where you can translate between the newer protocol and the older one, if you got cases where you can't do that, you need to decide somehow that from now on we are going to use the new logic format through a referendum to enable a feature flag sounded like a way to random guess if enough validators have upgraded. 

I guess you are suggesting to  actually do it dynamically from the node, and have the node decide to use the new logic if they see that 90%(or any number) of the nodes have the new protocol version.

bkchr:
> if you got cases where you can't do that

I mean we already had these cases and then we have bumped the `ParachainHost` version. Like for enabling async backing et all. And the node will need to support the old and new the new protocol any way until you pull the switch, so it could may also run both in parallel?

alexggh:
> > if you got cases where you can't do that
> 
> I mean we already had these cases and then we have bumped the `ParachainHost` version. Like for enabling async backing et all. 

Yes, definitely we can do it with bumping the runtime version, I was hopping that we can build a specific explicit mechanism for enabling feature, so that we don't need to bump the runtime version. It works, but I think it overloads with meaning the runtime upgrades, we don't need any extra logic from the runtime, we just use it to signal that nodes should enable some new behaviour.

> And the node will need to support the old and new the new protocol any way until you pull the switch, so it could may also run both in parallel?

At least for this case https://github.com/paritytech/polkadot-sdk/issues/628 they can't run in parallel, once we enable the creation of the new messages, older nodes won't be able the deserialise and check them. Also, newer nodes can't translate the message to the old message because of the way signatures are generated and checked.

Anyways, it was just an idea of streamlining things even more than what this PR proposes, you guys convinced me it is a bad idea :D.

alindima:
> I mean we already had these cases and then we have bumped the ParachainHost version. Like for enabling async backing et all. And the node will need to support the old and new the new protocol any way until you pull the switch, so it could may also run both in parallel?

It's also the case for https://github.com/paritytech/polkadot-sdk/issues/598 that it can't run with both the feature being disabled on some nodes and enabled on others at the same time. It has to be an atomic switch. The node-side code is able handle both cases, but not both at the same time.

The systematic availability recovery feature is not tied to a networking protocol upgrade neccessarily and also does not have some configuration params that need to be queried from the runtime (unlike async backing).

Bumping the runtime API version like we did for async backing works, and this PR mimics just that, but in a way that can be (at least partially) reused by other features in the future.

bkchr:
But do we really need a secondary versioning scheme? Or isn't the runtime api version good enough for this?

polkadot/node/core/runtime-api/src/cache.rs
alindima: done. please have a look

alindima:
> > But do we really need a secondary versioning scheme? Or isn't the runtime api version good enough for this?
> 
> If by versioning scheme you mean the fact that we have an additional bitfield of features stored in the runtime, it'll help with:
> 
> * being able to disable/reenable the feature after it went live (if a reason/bug comes up). Using only the runtime API version would require a runtime upgrade when disabling a feature. Having a value stored in the `HostConfiguration` would just require a referendum/sudo call.
> * having a way of using the runtime API for querying the status of a feature even if the feature does not need a specific runtime API to be available in order to function. This is not the case for async backing because it already requires some params to be stored in the runtime. But for features like systematic recovery, there isn't such a parameter. The alternative would be to have a runtime API like: `is_feature_X_enabled` which just returns a bool. Conceptually, this is what this PR enables, while also adding some genericitiy so that we won't have a bunch of `is_feature_X_enabled` runtime APIs in the future.

@bkchr, taking this into account, does the PR seem reasonable?

bkchr:
I'm not totally sold, however I don't want to block this. 

However, I would request the following changes:
- Please name it `NodeFeatures`.
- Please make `NodeFeatures` a `BitVec`. This enables us to add as many features as we like :P

alindima:
> Please name it NodeFeatures.

will do

> Please make NodeFeatures a BitVec. This enables us to add as many features as we like :P

It currently uses `bitflags`, which should do the same thing, with the addition of having named bits. Still, I'll switch to `BitVec` as you suggest, because it's more widespread in our codebase.

bkchr:
> It currently uses `bitflags`, which should do the same thing, with the addition of having named bits. Still, I'll switch to `BitVec` as you suggest, because it's more widespread in our codebase.

I would just make the type returned by the runtime a `BitVec`. Internally using a `bitflag` for having named bits makes sense or we just create some  statics for this. Best to check out what the easiest is with the `BitVec` api.

sandreim (commented):
Looks good to me, but I think it is missing that safe interface for flipping individual bits. If we'll rely on the introduction of per client feature setterst then we'd do one extra runtime upgrade for what could have been just a single configuration set call.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4272962

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4272963

alindima:
bot help

alindima:
bot bench polkadot-pallet --runtime=rococo --pallet=configuration

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4281875

alexggh (approved):

alindima:
bot bench polkadot-pallet --runtime=rococo --pallet=configuration
bot bench polkadot-pallet --runtime=westend --pallet=configuration

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4282823

alindima:
bot bench polkadot-pallet --runtime=rococo --pallet=configuration
bot bench polkadot-pallet --runtime=westend --pallet=configuration

alindima:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::configuration
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::configuration

alindima:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::configuration

alindima:
I added a `set_node_feature(index, bool)` extrinsic for enabling/disabling a feature and the benchmark for it.
PR is ready now

polkadot/runtime/parachains/src/configuration.rs
```diff
@@ -261,6 +261,8 @@ pub struct HostConfiguration<BlockNumber> {
 	/// The minimum number of valid backing statements required to consider a parachain candidate
 	/// backable.
 	pub minimum_backing_votes: u32,
+	/// Node features enablement.
+	pub node_features: NodeFeatures,
```
sandreim, line 265: I think a better name is `required_client_capabilities`.

polkadot/runtime/parachains/src/configuration.rs
```diff
@@ -1195,6 +1203,23 @@ pub mod pallet {
 				config.minimum_backing_votes = new;
 			})
 		}
+		/// Set/Unset a node feature.
+		#[pallet::call_index(53)]
+		#[pallet::weight((
+			T::WeightInfo::set_node_feature(),
+			DispatchClass::Operational
+		))]
+		pub fn set_node_feature(origin: OriginFor<T>, index: u8, value: bool) -> DispatchResult {
```
sandreim, line 1212:
```suggestion
		pub fn set_required_client_capability(origin: OriginFor<T>, index: u8, value: bool) -> DispatchResult {
```

polkadot/node/subsystem-util/src/runtime/mod.rs
```diff
@@ -507,3 +507,32 @@ pub async fn request_min_backing_votes(
...
+	sender: &mut impl overseer::SubsystemSender<RuntimeApiMessage>,
+) -> Result<Option<NodeFeatures>> {
+	let res = recv_runtime(
+		request_from_runtime(parent, sender, |tx| {
+			RuntimeApiRequest::NodeFeatures(session_index, tx)
+		})
+		.await,
+	)
+	.await;
+
```
sandreim, line 526: On session changes it would be a good idea to emit an warning to upgrade the node if not all bits make sense(at least a required feature we don't have)

sandreim (approved):
👍🏼 

I think it needs one final fix on the extrisinc/api/config naming as suggested.

polkadot/node/subsystem-util/src/runtime/mod.rs
alindima: I don't think that's doable if we want to be able to add new features without a runtime upgrade.
The runtime does not have any idea of what the feature bits mean. They only make sense on the node-side.

If we want to emit warnings for new features that are added, they have to come via a runtime upgrade (which would defeat the purpose of what I've done with this PR so far in order to have seamless feature addition/enabling). I think it's best to leave it as it is

polkadot/runtime/parachains/src/configuration.rs
alindima: I don't think this name is strictly better. I don't have a strong opinion but we've just decided on `node_features`. WDYT @bkchr @alexggh

polkadot/node/subsystem-util/src/runtime/mod.rs
sandreim: >The runtime does not have any idea of what the feature bits mean. They only make sense on the node-side.
> 

Yes exactly, those are bits that only the client understands. This is just to inform the operator, that currently a client requirement is not fulfilled and a node upgrade is required.

> If we want to emit warnings for new features that are added, they have to come via a runtime upgrade (which would defeat the purpose of what I've done with this PR so far in order to have seamless feature addition/enabling). I think it's best to leave it as it is

No runtime upgrade is needed as the client will see a new `1` bit  when the feature gets enabled. Even if it doesn't know what it means it can still print the warning saying to upgrade the node.

alindima: makes sense, I didn't fully get this from the beginning.
I can make the `RuntimeApiSubsystem` track the last session and when seeing a new session check the highest bit set returned from the `node_features` runtime API. If that's higher than some `LAST_NODE_FEATURE` constant on the node side, issue a warning.

As spoken on element, I'll add this on the next PR that uses this runtime API, as it'd be redundant right now.
Good suggestion 👍🏻

alindima:
bot clean

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4330326

PR #2178, opened 2023-11-06
Add the Header::Hash type param to the metadata

serban300:
We need this field to be in the metadata for the generated runtimes that we use for the bridges.

More details:
https://github.com/paritytech/parity-bridges-common/issues/2669
https://github.com/paritytech/subxt/issues/1247

### DIFF
--- a/substrate/primitives/runtime/src/generic/header.rs
+++ b/substrate/primitives/runtime/src/generic/header.rs
@@ -34,7 +34,6 @@ use sp_std::fmt::Debug;
 
 /// Abstraction over a block header for a substrate chain.
 #[derive(Encode, Decode, PartialEq, Eq, Clone, sp_core::RuntimeDebug, TypeInfo)]
-#[scale_info(skip_type_params(Hash))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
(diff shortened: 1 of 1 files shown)


svyatonik (approved):
:+1: for us it is fine, but I wonder if it may break something else, so I'd wait for approval from those who are working on that code

bkchr:
It gets even weirder if you look into the [`Config`](https://docs.rs/subxt/latest/subxt/config/trait.Config.html) that you need to provide to subxt. It already includes the `Hasher` and the `Header`, both of the types we are speaking here about. 

CC @jsdw

serban300:
> It gets even weirder if you look into the [`Config`](https://docs.rs/subxt/latest/subxt/config/trait.Config.html) that you need to provide to subxt. It already includes the `Hasher` and the `Header`, both of the types we are speaking here about.
> 
> CC @jsdw

I am not very familiar with this, but I guess that config needs to be provided only if someone uses subxt for submitting the extrinsics. We only use `subxt-codegen` to generate the runtime api and we have our own code for building and submitting the extrinsics. Also I guess the hasher and the header in that config are the ones of the current chain. For our extrinsics it's the header and hash of the bridged chains, so it gets a bit more complicated.

But for someone that submits extrinsics using subxt, I think that config makes sense

bkchr:
> We only use `subxt-codegen` to generate the runtime api and we have our own code for building and submitting the extrinsics

I mean you are using the metadata of the target chain or? I mean I also have no real idea how `subxt` works, but @jsdw should be able to help here :P

serban300:
> I mean you are using the metadata of the target chain or?

Yes, we're only using the metadata of the target chain to get the definition of the Runtime Call. And with that we can build and submit the calls that we need from the relayers.

serban300:
Closing this PR. I will think of alternatives.

jsdw:
I left a response in the subxt issue for this (https://github.com/paritytech/subxt/issues/1247).

PR #2179, opened 2023-11-06
minor: overseer availability-distribution message declaration update

alindima:
availability-distribution subsystem is not sending availability-recovery messages. Update the overseer declaration to reflect this


alexggh (approved):

sandreim (approved):

mrcnski (approved):
Why was it here in the first place? 🤔

alindima:
> Why was it here in the first place? 🤔

Just human error it seems. Checked out the repo on [the PR that added that line](https://github.com/paritytech/polkadot/pull/5314/files) and av-distribution was not sending this message back then either

PR #2180, opened 2023-11-06
Unify `ChainSync` actions under one enum

dmitry-markin:
All `ChainSync` actions that `SyncingEngine` should perform are unified under one `ChainSyncAction`. Processing of these actions put into a single place after `select!` in `SyncingEngine::run` instead of multiple places where calling `ChainSync` methods.


substrate/client/network/sync/src/engine.rs
```diff
@@ -713,11 +713,36 @@ where
 			self.is_major_syncing
 				.store(self.chain_sync.status().state.is_major_syncing(), Ordering::Relaxed);
 
+			// Perform actions requested by `ChainSync` in during `select!`.
+			self.perform_chain_sync_actions();
+
 			// Send outbound requests on `ChanSync`'s behalf.
 			self.send_chain_sync_requests();
```
altonen, line 720: Is it necessary to have both of these functions? Conceptually block requests should be sent when some action is performed for `ChainSync` (peer connected, import queue drained) which should result in further actions.

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -1346,7 +1337,7 @@ where
 	/// Restart the sync process. This will reset all pending block requests and return an iterator
 	/// of new block requests to make to peers. Peers that were downloading finality data (i.e.
 	/// their state was `DownloadingJustification`) are unaffected and will stay in the same state.
-	fn restart(&mut self) -> impl Iterator<Item = Result<BlockRequestAction<B>, BadPeer>> + '_ {
+	fn restart(&mut self) -> impl Iterator<Item = ChainSyncAction<B>> + '_ {
```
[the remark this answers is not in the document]
dmitry-markin: The only place where `restart()` is called is in [`on_blocks_processed()`](https://github.com/paritytech/polkadot-sdk/blob/37a488a192f19381c176fb91a7c37f4b3b311ff6/substrate/client/network/sync/src/chain_sync.rs#L1928), and we push everything returned by `restart()` to `self.actions` anyway. So, it's not an issue if there are already some actions there, the new actions will be added in the correct order.

substrate/client/network/sync/src/engine.rs
dmitry-markin: May be it's possible to refactor the code in a way so new requests are generated only in response to some event, but right now the are generated "on their own" by going over the peers in `ChainSync` and checking if we can perform a request to each of them.

dmitry-markin: @altonen do you think we can merge this PR? It's hard to say if it's possible to get rid of a separate stage of generating requests in `ChainSync`, but it should probably go as a separate change anyway.

altonen: Sorry, I didn't mean it would be done in this PR either way but something to look into for sure. Might be slightly nastier than it looks on the surface because `peer_connected()/block_imported()` would likely have to call `send_requests()` or something like that. But please take a look at some point if we could simplify this further.

substrate/client/network/sync/src/engine.rs
```diff
@@ -713,11 +713,36 @@ where
 			self.is_major_syncing
 				.store(self.chain_sync.status().state.is_major_syncing(), Ordering::Relaxed);
 
+			// Perform actions requested by `ChainSync` in during `select!`.
```
altonen, line 716:
```suggestion
			// Perform actions requested by `ChainSync` during `select!`.
```

substrate/client/network/sync/src/engine.rs
```diff
@@ -713,11 +713,36 @@ where
...
 			// Send outbound requests on `ChanSync`'s behalf.
 			self.send_chain_sync_requests();
 		}
 	}
 
+	fn perform_chain_sync_actions(&mut self) {
+		self.chain_sync.take_actions().for_each(|action| match action {
+			ChainSyncAction::SendBlockRequest { peer_id, request } => {
+				// drop obsolete pending response first
+				self.pending_responses.remove(&peer_id);
```
altonen, line 728: This could use a comment because it's a little confusing if you're reading this for the first time. Maybe worth adding a trace log as well in case it actually dropped a stale response?

altonen: Actually I think each action could have a trace log. I'm sure at some point we'll be glad it's logging something

substrate/client/network/sync/src/engine.rs
```diff
@@ -713,11 +713,36 @@ where
 			self.is_major_syncing
 				.store(self.chain_sync.status().state.is_major_syncing(), Ordering::Relaxed);
 
+			// Perform actions requested by `ChainSync` in during `select!`.
+			self.perform_chain_sync_actions();
```
altonen, line 717:
```suggestion
			self.process_chain_sync_actions();
```

substrate/client/network/sync/src/engine.rs
```diff
@@ -713,11 +713,36 @@ where
...
+
 			// Send outbound requests on `ChanSync`'s behalf.
 			self.send_chain_sync_requests();
 		}
 	}
 
+	fn perform_chain_sync_actions(&mut self) {
```
altonen, line 724:
```suggestion
	fn process_chain_sync_actions(&mut self) {
```

altonen (approved):

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -1212,7 +1200,11 @@ where
 
 		let blocks = self.ready_blocks();
 
-		(!blocks.is_empty()).then(|| self.validate_and_queue_blocks(blocks, false))
+		if !blocks.is_empty() {
+			let ImportBlocksAction { origin, blocks } =
+				self.validate_and_queue_blocks(blocks, false);
+			self.actions.push(ChainSyncAction::ImportBlocks { origin, blocks })
```
bkchr, line 1206: Why is this not happening inside of `validate_and_queue_blocks`?

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -2085,7 +2073,14 @@ where
 		}
 
 		self.allowed_requests.set_all();
-		Box::new(output.into_iter())
+
+		self.actions.append(&mut output);
```
bkchr, line 2077: Why don't we directly push to `self.actions`?

bkchr (approved):

substrate/client/network/sync/src/chain_sync.rs
dmitry-markin: We need to do `self.actions.extend(self.restart())` in multiple places in this function, but we can't do that due to borrowing issues because `restart()` returns an iterator. So we accumulate all the actions until the end of the function in a separate `Vec`.

bkchr: Why isn't `restart` directly pushing to `self.actions` as well?

substrate/client/network/sync/src/chain_sync.rs
dmitry-markin: Fixed — now all actions are pushed at the place they are generated. Except `BadPeer` errors — those are sometime deferred until the function returns to simplify the error handling.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4275508

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4275793

substrate/client/network/sync/src/chain_sync.rs
dmitry-markin: Yeah, makes sense. Fixed.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4275943

substrate/client/network/sync/src/chain_sync.rs
dmitry-markin: Well, actually there is one function left returning actions — `new_peer_inner`.  But we need to track in `restart` that it is not requesting any actions to cancel obsolete requests, so it's hard to make it not return actions.

dmitry-markin: Still need to fix tests :see_no_evil:

dmitry-markin: Now unit tests are not that unit anymore...

dmitry-markin: The tests are updated, but they became uglier because of the need to track actions pushed into `self.actions` instead of just checking return values.
@bkchr could you have a look again?

dmitry-markin: I'm going to merge this to move forward, but we can revisit these changes later if you know better ways of testing structs with an output queue.

PR #2181, opened 2023-11-06
pallet-grandpa: Remove `GRANDPA_AUTHORITIES_KEY`

bkchr:
Remove the `GRANDPA_AUTHORITIES_KEY` key and its usage. Apparently this was used in the early days to communicate the grandpa authorities to the node. However, we have now a runtime api that does this for us. So, this pull request is moving from the custom managed storage item to a FRAME managed storage item.

This pr also includes a migration for doing the switch on a running chain.


substrate/frame/grandpa/src/lib.rs
```diff
@@ -428,12 +432,12 @@ pub enum StoredState<N> {
 impl<T: Config> Pallet<T> {
 	/// Get the current set of authorities, along with their respective weights.
 	pub fn grandpa_authorities() -> AuthorityList {
-		storage::unhashed::get_or_default::<VersionedAuthorityList>(GRANDPA_AUTHORITIES_KEY).into()
+		Authorities::<T>::get().to_vec()
```
davxy, line 435:
```suggestion
		Authorities::<T>::get().into_inner()
```

substrate/frame/grandpa/src/lib.rs
```diff
@@ -522,10 +526,14 @@ impl<T: Config> Pallet<T> {
...
 	// Perform module initialization, abstracted so that it can be called either through genesis
 	// config builder or through `on_genesis_session`.
-	fn initialize(authorities: &AuthorityList) {
+	fn initialize(authorities: AuthorityList) {
 		if !authorities.is_empty() {
 			assert!(Self::grandpa_authorities().is_empty(), "Authorities are already initialized!");
-			Self::set_grandpa_authorities(authorities);
+			Self::set_grandpa_authorities(
+				&BoundedAuthorityList::<T::MaxAuthorities>::try_from(authorities).expect(
+					"Granpa: `Config::MaxAuthorities` is smaller than the number of genesis authorities!",
```
davxy, line 534:
```suggestion
					"Grandpa: `Config::MaxAuthorities` is smaller than the number of genesis authorities!",
```

davxy (approved):

substrate/primitives/consensus/grandpa/src/lib.rs
```diff
@@ -469,55 +462,6 @@ pub const PENDING_CHANGE_CALL: &str = "grandpa_pending_change";
 /// WASM function call to get current GRANDPA authorities.
 pub const AUTHORITIES_CALL: &str = "grandpa_authorities";
```
andresilva, line 463:
```suggestion
```

Unrelated to this PR but these can be removed (they're unused). The one above as well but github doesn't allow me to comment on it.

substrate/frame/grandpa/src/lib.rs
```diff
@@ -428,12 +432,12 @@ pub enum StoredState<N> {
...
 	/// Get the current set of authorities, along with their respective weights.
 	pub fn grandpa_authorities() -> AuthorityList {
-		storage::unhashed::get_or_default::<VersionedAuthorityList>(GRANDPA_AUTHORITIES_KEY).into()
+		Authorities::<T>::get().to_vec()
 	}
 
 	/// Set the current set of authorities, along with their respective weights.
-	fn set_grandpa_authorities(authorities: &AuthorityList) {
-		storage::unhashed::put(GRANDPA_AUTHORITIES_KEY, &VersionedAuthorityList::from(authorities));
+	pub(crate) fn set_grandpa_authorities(authorities: &BoundedAuthorityList<T::MaxAuthorities>) {
```
andresilva, line 439: Maybe we could remove this method now.

substrate/frame/grandpa/src/migrations/v5.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct MigrateImpl<T>(PhantomData<T>);
+
+impl<T: crate::Config> OnRuntimeUpgrade for MigrateImpl<T> {
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
+		use codec::Encode;
+
+		let authority_list_len = load_authority_list().len() as u32;
+
+		if authority_list_len > T::MaxAuthorities::get() {
```
andresilva, line 48: Should this fail as well if `authority_list_len` is 0?

substrate/frame/grandpa/src/migrations/v5.rs
```diff
@@ -0,0 +1,92 @@
...
+		crate::Pallet::<T>::set_grandpa_authorities(
+			&BoundedAuthorityList::<T::MaxAuthorities>::force_from(
+				load_authority_list(),
+				Some("Grandpa: `Config::MaxAuthorities` is smaller than the actual number of authorities.")
+			)
+		);
+
+		storage::unhashed::kill(GRANDPA_AUTHORITIES_KEY);
+
+		T::DbWeight::get().reads_writes(1, 1)
```
andresilva, line 84: Should be two writes (because of kill)?

andresilva (approved):

h4x3rotab:
It looks like we are using this storage key to get the updates of the authoriy set for the light client. We need to not only get the changed authority set, but also with the storage proof. How to get the proof if it's moved to a function?

bkchr:
Just call the GrandpaApi::authorities runtime api.

h4x3rotab:
> Just call the GrandpaApi::authorities runtime api. 

Correct me if I'm wrong. That can be used to fetch the authorities, but not producing any storage proof for the light client to update the authority set.

bkchr:
You can do an execution proof. You execute the function and record the state accesses. Then you re-execute this on the light client. Substrate supports this.

h4x3rotab:
@bkchr Does that mean my client client must add the capability to run wasm code in order to validate the storage access is correct? In the past, we only need to make sure the storage access can be proven. After this change, it seems I will need to build a much complex client.

Usually a light client is used when we cannot afford complex computation. But if the wasm execution is needed, I can imagine this would be super hard to build the light client running on very resource limited scenarios. One example is the light client compiled inside a smart contract (e.g. in Cosmwasm).

bkchr:
An on chain light client would just follow set changes? And set changes are part of the header, to make it quite easy for light clients to fetch it.

Nevertheless, you can still calculate the storage key and do a storage read proof. This is almost the same as before, just that the key is may not the same in every runtime.

h4x3rotab:
> An on chain light client would just follow set changes? And set changes are part of the header, to make it quite easy for light clients to fetch it.
That's super helpful. Where can I find the doc / code related to the set change info?

bkchr:
https://github.com/paritytech/polkadot-sdk/blob/e5dc8ed8f5a137298a8c4f9ead509ea13682d7e6/substrate/primitives/consensus/grandpa/src/lib.rs#L149-L162 not sure if there is that much more info.

andresilva:
https://github.com/paritytech/polkadot-sdk/tree/master/bridges/modules/grandpa this might also be useful for reference.

Polkadot-Forum:

https://forum.polkadot.network/t/runtime-release-schedule-and-coordination-on-breaking-changes/5431/1

h4x3rotab:
When will be this change released on Kusama or Polkadot?

bkchr:
Should be part of the next update

h4x3rotab:
> Should be part of the next update 

Got it. Should I go to the Polkadot-fellows repo to submit an issue to discuss the release schedule?

I sincerely hope there can be more time for the community to prepare for this change. For our blockchain, if this get released on Kusama, most of our miners will brick immediately. We need around 1 month to communicate with the community to coordinate such large scale update smoothly.

bkchr:
It will take around 3 weeks until this hits kusama, is that enough time?

h4x3rotab:
> It will take around 3 weeks until this hits kusama, is that enough time?

It's tricky. Would be nice if we can have at least 4 weeks. I would highly appreciate it if we can rollback this specific change and leave it for the next release.

We have just tested everything locally this week, but this change would require a "hard fork" level client upgrade to the community. All the miners running the old clients would stop working. Usually this level of coordination would take a quarter, but we can try hard to push it as soon as possible. Even if we start to push it tomorrow, a lot of miners will sync their client from scratch, which will take 1-2 weeks. Including the communication time, I think at least 4 weeks is necessary.

bkchr:
Why do nodes need to resync? Can they not just switch over or better, you write your client in a way to handle both keys?

But yeah, I will do a revert. Please ensure that you are ready for the next upgrade.

h4x3rotab:
> But yeah, I will do a revert. Please ensure that you are ready for the next upgrade.

Thanks you so much! Btw, could you suggest the best way to track the release schedule? I read the runtime release analysis on the forum, but it doesn't mention the actual release schedule. We will coordinate better the next time.

> Why do nodes need to resync? Can they not just switch over or better, you write your client in a way to handle both keys?

Technically they don't need. And yes, the new client is compatible with the both grandpa keys. The difficulty lies in handling the TEE key properly between the old and upgraded TEE worker (the offchain TEE worker, not the substrate node). This process is error prune, as it involves key handover between the old and the new worker. Any misconfiguration may result in data loss, and therefore they need to resync the node.

We don't upgrade the worker client frequently as it has been running very stably for one year. Usually we leave 1-2 quarters for the community to upgrade the worker binary. A lot of miners just close their old worker and resync another to avoid facing the technical complexity. A quarter long upgrade window is usually sufficient, but 3 weeks are too short.

bkchr:
Production runtime releases are happening here: https://github.com/polkadot-fellows/runtimes

They are not that connected to the releases happening from this repo. The release schedule for the production runtimes is also not yet that fixed :see_no_evil:. Not sure what would be the way to track these releases right now, besides following the repo.

nanometerzhu:
> https://github.com/paritytech/polkadot-sdk/tree/master/bridges/modules/grandpa this might also be useful for reference.

I have some concerns about the authority set changes part:

> when validators set changes, the GRANDPA gadget adds next set to the header. So light client doesn't need to verify storage proofs when this happens - it only needs to look at the header and see if it changes the set. Once set is changed, all following justifications are generated by the new set. Header that is changing the set is called "mandatory" in the pallet. As the name says, the light client need to import all such headers to be able to operate properly.

Basically this is what warp sync do. Every time the authority set changes, justification and the old authority set are used to verify the hash, this part is fine. But the problem is, after passing the verify of hash, the code also accepts the digests in the header: https://github.com/paritytech/polkadot-sdk/blob/930c1519281ac7ec54dba2224f5cdea1d3959b5a/substrate/client/consensus/grandpa/src/warp_proof.rs#L216-L223

Is it possible, for example, a hacker use a right hash and justification, but fake the digests in headers so the light client accepts bad new authority set?

andresilva:
If you change the digests the header hash would change.

nanometerzhu:
> If you change the digests the header hash would change.

It explains, thanks a lot!

Issue #2186, opened 2023-11-07
refactor polkadot-runtime-common


xlc:
It is like an unloved child received all the garbage gift from people don't know where to put their stuff.

Yet I am forced to inherent all those via transitive dependency.

cumulus-pallet-xcmp-queue -> polkadot-runtime-common -> pallet-asset-rate and https://github.com/paritytech/polkadot-sdk/issues/1136#issuecomment-1797063298

xlc:
why `trait PriceForMessageDelivery` lives in polkadot-runtime-common?

https://github.com/paritytech/polkadot-sdk/blob/6f1b33befef5c164ccca53595d8f03dda203fae4/polkadot/runtime/common/src/xcm_sender.rs#L34

xlc:
polkadot-runtime-common received another gift https://github.com/paritytech/polkadot-sdk/pull/1814

asiniscalchi:
I'm totally behind the idea of taking the refactoring of `polkadot-runtime-common` one step at a time. Breaking it down into manageable pieces not only makes it simpler to oversee and evaluate each change, but it also allows us to adapt our strategy based on real-time feedback and unforeseen challenges. 

In line with this, could you give us an example of a particular component or functionality within `polkadot-runtime-common` that you believe should be relocated?

xlc:
We need to figure out what is `polkadot-runtime-common` exactly first. Maybe someone can write a single sentence description? And everything that doesn't fit within that description should be relocated.

If you really want an answer from me, I will just say everything should be relocated.

But we can start with pallets. Every crate can only contain a single pallet. Let's enforce this policy.

asiniscalchi:
Taking `pallet_para_sudo_wrapper` as an example, if we were to extract it, where do you think it should be relocated? This could serve as our starting point. By tackling this step by step, the `common` crate will gradually slim down, perhaps even finding its true identity through this process.

Relocating the pallets from the `common` crate can be tackled in parallel, making it a straightforward task for the community. This approach will allow us to efficiently distribute the work and make significant progress quickly.

I can draft a couple of user stories (issues) to kickstart this if you're interested.

bkontur:
> We need to figure out what is `polkadot-runtime-common` exactly first. Maybe someone can write a single sentence description? 

https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/common/Cargo.toml#L4
```
name = "polkadot-runtime-common"
[...]

PR #2187, opened 2023-11-07
mark pallet-asset-rate optional in polkadot-runtime-common

xlc:
Part of #2186

The only usage of pallet-asset-rate is guarded by `runtime-benchmarks` feature. I don't want ORML to be forced to include this pallet in deps for no good reason.


acatangiu (approved):
LGTM

kianenigma (approved):

bkchr (approved):

PR #2202, opened 2023-11-07
Disable incoming light-client connections for minimal relay node

### LINKED ISSUE #1896: High network usage when using --relay-chain-rpc-url

### Description of bug

After upgrading to moonbeam 0.33 (based on v0.9.43), we started to see a massive increase of network usage.
When using the `--relay-chain-rpc-url` flag, the node is consuming a massive amount of network traffic, not related to p2p nor rpc traffic. We also started to see a much higher CPU usage.

On the left side of the graph we are using the embedded relay, and on the right side we are attached to en external relay.
The traffic does not show up on the opposing relay rpc side.

image

Config with `relay-cahin-rpc-url`
```
      --base-path=/data/moonbeam
      --rpc-port=9944
      --prometheus-port=9615
      --prometheus-external
      --name=node_name
      --listen-addr=/ip4/0.0.0.0/tcp/30333
      --chain=moonbeam
      --database=rocksdb
      --relay-chain-rpc-url=ws://nlg3-polkadot-rpc-int:9945
      --execution=wasm
      --wasm-runtime-overrides=/chain-network/tracing-substitutes
      --pruning=archive
      --db-cache=1024
      --trie-cache-size=1073741824
      --rpc-cors=all
      --unsafe-rpc-external
      --rpc-methods=safe
      --ethapi=debug,txpool,trace
      --runtime-cache-size=32
      --eth-log-block-cache=2000000000
      --eth-statuses-cache=2000000000
      --ethapi-max-permits=7
      --detailed-log-output
      --no-hardware-benchmarks
      --rpc-max-connections=1000
      --max-runtime-instances=32
      --
      --chain=polkadot
```

Config without `relay-cahin-rpc-url`
```
      --base-path=/data/moonbeam
      --rpc-port=9944
      --prometheus-port=9615
      --prometheus-external
      --name=node_name
      --listen-addr=/ip4/0.0.0.0/tcp/30333
      --chain=moonbeam
      --database=rocksdb
      --execution=wasm
      --wasm-runtime-overrides=/chain-network/tracing-substitutes
      --pruning=archive
      --db-cache=1024
      --trie-cache-size=1073741824
      --rpc-cors=all
      --unsafe-rpc-external
      --rpc-methods=safe
      --ethapi=debug,txpool,trace
      --runtime-cache-size=32
      --eth-log-block-cache=2000000000
      --eth-statuses-cache=2000000000
      --ethapi-max-permits=7
      --detailed-log-output
      --no-hardware-benchmarks
      --rpc-max-connections=1000
      --max-runtime-instances=32
      --
      --chain=polkadot
      --base-path=/data/polkadot
      --rpc-port=9945
      --prometheus-port=9616
      --prometheus-external
      --name=node_name-relay
      --listen-addr=/ip4/0.0.0.0/tcp/30334
      --database=rocksdb
      --pool-limit=0
      --db-cache=1024
      --execution=wasm
      --state-pruning=2000
      --blocks-pruning=2000
      --rpc-cors=all
      --rpc-methods=safe
      --unsafe-rpc-external
      --in-peers-light=0
      --in-peers=5
      --out-peers=25
```

### Steps to reproduce

_No response_

skunert:
When running with `--relay-chain-rpc-url` we received multiple reports of high traffic that disappears when `--in-peers-light 0` is set. Indeed it does not make much sense for light clients to connect to the minimal node since it is not running the block announce protocol and the request/response protocol for light clients.

This is intended to alleviate the traffic issues for now.

closes #1896
probably related https://github.com/paritytech/cumulus/issues/2563


cumulus/client/relay-chain-minimal-node/src/network.rs
```diff
@@ -114,6 +119,18 @@ pub(crate) fn build_collator_network(
 	Ok((network_service, network_starter, Box::new(SyncOracle {})))
 }
 
+fn adjust_network_config_light_in_peers(config: &mut NetworkConfiguration) {
+	let light_client_in_peers = (config.default_peers_set.in_peers +
+		config.default_peers_set.out_peers)
+		.saturating_sub(config.default_peers_set_num_full);
```
dmitry-markin, line 125: Even though it looks ugly, it seems there is no other way to estimate `in_peers_light` using current implementation of `NetworkConfiguration`...

dmitry-markin (approved):

cumulus/client/relay-chain-minimal-node/src/network.rs
skunert: Yes, this should probably become a proper field in the config some day.

davxy: There is something I don't understand, maybe due to some limited domain knowledge in the networking subsystem. How reducing the `in_peers` number just trims the light clients connections?
I mean, aren't the full clients connections affected as well?

skunert: This is a bit tricky, when users specify `--in-peers-light` we actually add them to the normal `in_peers` in config construction: https://github.com/paritytech/polkadot-sdk/blob/b44de38dde1a6d76131ba2c4123a500aa9e0b70e/substrate/client/cli/src/params/network_params.rs#L235-L245.

And here is the logic that checks for the light client slots actually occupied:
https://github.com/paritytech/polkadot-sdk/blob/b44de38dde1a6d76131ba2c4123a500aa9e0b70e/substrate/client/network/sync/src/engine.rs#L1035-L1041

The whole concept of light client slots is kind of implicit currently. But in the end you can still calculate the light client slots with  by calculating the difference as in the code.

altonen (approved):

cumulus/client/relay-chain-minimal-node/src/network.rs
dmitry-markin: I was going to comment, but @skunert beat me. Could just add that it's the limitation of light peer slots bookkeeping, we calculate the slot count in `SyncingEngine` from `in_peers`, `out_peers`, `full_peers`:
https://github.com/paritytech/polkadot-sdk/blob/4caa3d8d8e6f90b2689cd2666c94044394712abf/substrate/client/network/sync/src/engine.rs#L422-L426
So it's enough to update `config.default_peers_set.in_peers` for light peer slots to be correctly calculated later.

davxy: clear Ty

davxy (approved):

lexnv (approved):

PR #2212, opened 2023-11-07
add missing syscalls for workers

tugytur:
# Description

Since the binary split additional syscalls are getting blocked in relation to the workers.

With the hardened systemd file it shows the following warning:

```
Cannot fully enable landlock, a Linux kernel security feature. Running validation of malicious PVF code has a higher risk of compromising this machine. Consider upgrading the kernel version for maximum security. status=Ok(NotEnforced) abi=1
```

For it to work we need to allow additionally:
- mount
- umount2
- pivot_root

and set `RestrictNamespaces=false`

Added new line `SystemCallFilter=pivot_root` because otherwise it would get blocked by ~\@\privileged


s0me0ne-unkn0wn:
This one could partially resolve #2728, it's a shame it got lost. Let's merge it.

s0me0ne-unkn0wn (approved):

s0me0ne-unkn0wn:
@lovelaced could you please have a look?

lovelaced (approved):
Looks sane to me, I would definitely prefer users do this rather than blanket-disable all systemd protections as was done in the other issue.

bkchr:
bot merge

bkchr (approved):

PR #2214, opened 2023-11-07
validate-block: Fix `TrieCache` implementation

bkchr:
The trie cache implementation was ignoring the `storage_root` when setting up the value cache. The problem with this is that the value cache works using `storage_keys` and these keys are not unique across different tries. A block can actually have different tries (main trie and multiple child tries). This pull request fixes the issue by not ignoring the `storage_root` and returning an unique `value_cache` per `storage_root`. It also adds a test for the seen bug and improves documentation that this doesn't happen again.

### DIFF
--- a/cumulus/pallets/parachain-system/src/validate_block/trie_cache.rs
+++ b/cumulus/pallets/parachain-system/src/validate_block/trie_cache.rs
@@ -29,7 +29,7 @@ use trie_db::{node::NodeOwned, Hasher};
 /// have already been loaded and decoded from the storage proof.
 pub(crate) struct TrieCache<'a, H: Hasher> {
 	node_cache: RefMut<'a, BTreeMap<H::Out, NodeOwned<H::Out>>>,
-	value_cache: Option<RefMut<'a, BTreeMap<Box<[u8]>, trie_db::CachedValue<H::Out>>>>,
+	value_cache: Option<RefMut<'a, &'a mut BTreeMap<Box<[u8]>, trie_db::CachedValue<H::Out>>>>,
 }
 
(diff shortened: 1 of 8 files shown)


ggwpez:
Can you author blocks on `rococo-contracts` with this fix?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4252032

cumulus/pallets/parachain-system/src/validate_block/trie_cache.rs
```diff
@@ -29,7 +29,7 @@ use trie_db::{node::NodeOwned, Hasher};
 /// have already been loaded and decoded from the storage proof.
 pub(crate) struct TrieCache<'a, H: Hasher> {
 	node_cache: RefMut<'a, BTreeMap<H::Out, NodeOwned<H::Out>>>,
-	value_cache: Option<RefMut<'a, BTreeMap<Box<[u8]>, trie_db::CachedValue<H::Out>>>>,
+	value_cache: Option<RefMut<'a, &'a mut BTreeMap<Box<[u8]>, trie_db::CachedValue<H::Out>>>>,
```
skunert, line 32:
```suggestion
	value_cache: Option<RefMut<'a, BTreeMap<Box<[u8]>, trie_db::CachedValue<H::Out>>>>,
```
This needs to go for it to compile.

skunert (approved):
Very good find!

cumulus/pallets/parachain-system/src/validate_block/trie_cache.rs
bkchr: It compiled locally for me 🙃

michalkucharczyk (approved):

Issue #2163, opened 2023-11-05
PVF worker: switch on seccomp networking restrictions


mrcnski:
We've merged https://github.com/paritytech/polkadot-sdk/pull/2009/ which logs seccomp violations for networking, but does not switch on full seccomp restrictions (voting against blocks on violations).

We should monitor [releases](https://github.com/paritytech/polkadot-sdk/releases) and once https://github.com/paritytech/polkadot-sdk/pull/2009/ has been released, we can switch this on for the next release.

PR #2221, opened 2023-11-08
PVF worker: switch on seccomp networking restrictions

mrcnski:
We've merged #2009 which logs seccomp violations for networking, but does not switch on full seccomp restrictions (voting against blocks on violations).

We should monitor releases and once #2009 has been released, we can switch this on for the next release. That way validators have some time to run with the logging enabled to make sure nothing unexpected happens.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/2163


polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -243,12 +243,6 @@ pub fn worker_entrypoint(
 					),
 				};
 
-				gum::trace!(
-					target: LOG_TARGET,
-					%worker_pid,
-					"worker: sending response to host: {:?}",
-					response
-				);
```
mrcnski, line 251: This was producing a bit too much noise in the log output.

polkadot/node/core/pvf/src/execute/worker_intf.rs
```diff
@@ -176,19 +176,6 @@ pub async fn start_work(
 						return Outcome::IoErr
 					},
 					Ok(response) => {
-						// Check if any syscall violations occurred during the job. For now this is
-						// only informative, as we are not enforcing the seccomp policy yet.
-						for syscall in security::check_seccomp_violations_for_worker(audit_log_file, pid).await {
-							gum::error!(
```
mrcnski, line 182: We still do this check elsewhere in this file, but only if the worker died.

s0me0ne-unkn0wn (approved):

eskimor (approved):

alindima (approved):

alindima:
I guess this should be Draft or have "Do not merge" in the title so that it's not merged until we've created a release including #2009 first

mrcnski:
@alindima good point. I was going to raise this with someone: I don't know how the release schedule works now. I've noticed my PRs not getting included in the next release, but rather the 2nd or 3rd one after. It seems quite opaque, but if we want this in by EOY, then perhaps it should be merged now. I'll try to get some clarity on this. cc @the-right-joyce

mrcnski:
Quick note. I found out that the seccomp audit logging would only work if someone is running as root, which is unlikely for production validators. 

We wanted this PR merged in this release (same one as secure-validator-mode). So, not much time to change our logging strategy, release, wait, etc. But, maybe we can skip that, since it is very unexpected for a legitimate PVF to try doing networking.

Is it possible to execute production blocks with these changes as a sanity-check, using something like kuddelmuddel? @ordian

Otherwise, I'll merge this, but in preparation for the general syscall blocking, we will need to solve the logging issue. I think ptrace is really the only option. I'm a bit worried that if ptrace and the OOM handler ever made contact, the explosion would destroy the universe. But we can have seccomp generate the ptrace events, they should never happen in the OOM handler.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4395596

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4395597

mrcnski:
There were some issues caused by me not updating seccomp detection after the worker/job separation was implemented. As stated above, the audit log is unlikely to be available, but of course I fixed the issues anyway.

I've tested with pvf-checker and confirmed that enabling these restrictions does not cause any issues for preparation of existing PVFs. I also tested a POV with kuddelmuddel, but we only had one available. It would be good to test more if possible.

mrcnski:
I tested some more POVs with kuddelmuddel against this branch without any issues. I also tested an invalid candidate (by flipping a bit in one of the POVs), as well as manually tested that bad syscalls do trigger the filter.

s0me0ne-unkn0wn (approved):
LGTM :+1:

PR #2223, opened 2023-11-08
Add prospective-parachain subsystem to minimal-relay-node + QoL improvements

skunert:
This PR contains some fixes and cleanups for parachain nodes:

1. When using async backing, node no longer complains about being unable to reach the prospective-parachain subsystem.
2. Parachain warp sync now informs users that the finalized para block has been retrieved.
```
2023-11-08 13:24:42 [Parachain] 🎉 Received finalized parachain header #5747719 (0xa0aa…674b) from the relay chain.
```
3. When a user supplied an invalid `--relay-chain-rpc-url`, we were crashing with a very verbose message. Removed the `expect` and improved the error message.
```
2023-11-08 13:57:56 [Parachain] No valid RPC url found. Stopping RPC worker.
2023-11-08 13:57:56 [Parachain] Essential task `relay-chain-rpc-worker` failed. Shutting down service.
Error: Service(Application(WorkerCommunicationError("RPC worker channel closed. This can hint and connectivity issues with the supplied RPC endpoints. Message: oneshot canceled")))
```


cumulus/client/service/src/lib.rs
```diff
@@ -560,11 +560,15 @@ where
...
+			let finalized_header = B::Header::decode(&mut &validation_data.parent_head.0[..])
 				.map_err(|e| format!("Failed to decode parachain head: {e}"))?;
 
-			log::debug!(target: LOG_TARGET_SYNC, "Target block reached {:?}", target_block);
-			let _ = sender.send(target_block);
+			log::info!(
+				"🎉 Received finalized parachain header #{} ({}) from the relay chain.",
+				finalized_header.number(),
+				finalized_header.hash()
+			);
```
bkchr, line 570: The syncing logs already inform us about the progress?

skunert: Yeah it is not an absolute necessity, you already get the occasional `⏩ Waiting for pending target block (7 peers), best: #0 (0x4823…771a), finalized #0 (0x4823…771a)`. However this here is tells you which block is the new target and is immediately printed when we retrieve the finalized one. I am also against spamming logs, but in my opinion this additional info does not hurt.

bkchr (approved):

cumulus/client/service/src/lib.rs
bkchr: But then maybe change the text to `Received target parachain header ..` or something like this.

davxy (approved):

cumulus/client/service/src/lib.rs
```diff
@@ -560,11 +560,15 @@ where
...
 				.ok_or("Could not find parachain head in relay chain")?;
 
-			let target_block = B::Header::decode(&mut &validation_data.parent_head.0[..])
+			let finalized_header = B::Header::decode(&mut &validation_data.parent_head.0[..])
 				.map_err(|e| format!("Failed to decode parachain head: {e}"))?;
 
-			log::debug!(target: LOG_TARGET_SYNC, "Target block reached {:?}", target_block);
-			let _ = sender.send(target_block);
+			log::info!(
+				"🎉 Received finalized parachain header #{} ({}) from the relay chain.",
```
skunert, line 567:
```suggestion
				"🎉 Received target parachain header #{} ({}) from the relay chain.",
```

PR #2228, opened 2023-11-08
sc-state-db: Keep track of `LAST_PRUNED` after warp syncing

bkchr:
When warp syncing we import the target block with all its state. However, we didn't store the `LAST_PRUNED` block which would then lead to `pruning` to forget about the imported block after a restart of the node. We just set `LAST_PRUNED` to the parent block of the warp sync target block to fix this issue.

### DIFF
--- a/substrate/client/state-db/src/pruning.rs
+++ b/substrate/client/state-db/src/pruning.rs
@@ -406,12 +406,24 @@ impl<BlockHash: Hash, Key: Hash, D: MetaDb> RefWindow<BlockHash, Key, D> {
 		commit: &mut CommitSet<Key>,
 	) -> Result<(), Error<D::Error>> {
 		if self.base == 0 && self.is_empty() && number > 0 {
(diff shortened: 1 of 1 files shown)


koute (approved):

skunert (approved):

arkpar:
What is the issue exactly? Pruning a warped state is no-op anyway. When a normal block is imported we generate a set of inserted trie nodes and a set of removed trie nodes for the state transition. Pruning a block means actually applying deletions in the database. Warp sync downlads the whole state and applies it to the empty trie. No nodes are removed from the trie, so pruning would do nothing.

arkpar (approved):

bkchr:
> This fix looks correct nvertheless, but I'm curious how does the issue manifest.

I warp synced a stalled parachain. When restarting the node it was complaining that the state of the target block, which was the last finalized + best block, was not available anymore. Which was not true.

PR #2229, opened 2023-11-08
Rococo: Build two versions of the wasm binary

bkchr:
One for local networks with `fast-runtime` feature activated (1 minute sessions) and one without the feature activated that will be the default that runs with 1 hour long sessions.

### DIFF
--- a/polkadot/runtime/rococo/constants/src/lib.rs
+++ b/polkadot/runtime/rococo/constants/src/lib.rs
@@ -37,14 +37,15 @@ pub mod currency {
 
 /// Time and blocks.
 pub mod time {
+	use runtime_common::prod_or_fast;
+
 	use primitives::{BlockNumber, Moment};
 	pub const MILLISECS_PER_BLOCK: Moment = 6000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
 
 	frame_support::parameter_types! {
-		pub storage EpochDurationInBlocks: BlockNumber = option_env!("ROCOCO_EPOCH_DURATION")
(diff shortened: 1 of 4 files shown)


polkadot/runtime/rococo/constants/src/lib.rs
```diff
@@ -37,14 +37,15 @@ pub mod currency {
...
 	use primitives::{BlockNumber, Moment};
 	pub const MILLISECS_PER_BLOCK: Moment = 6000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
 
 	frame_support::parameter_types! {
-		pub storage EpochDurationInBlocks: BlockNumber = option_env!("ROCOCO_EPOCH_DURATION")
-			.map(|s| s.parse().expect("`ROCOCO_EPOCH_DURATION` is not a valid `BlockNumber`"))
-			.unwrap_or(1 * MINUTES);
+		pub EpochDurationInBlocks: BlockNumber =
+			prod_or_fast!(1 * HOURS, 1 * MINUTES, "ROCOCO_EPOCH_DURATION");
```
ggwpez, line 48: Should it not rather check for feature `fast-runtime` here?

bkchr: https://github.com/paritytech/polkadot-sdk/blob/50390950d87fb745ec5ff46be698cac93113efb7/polkadot/runtime/common/src/lib.rs#L263-L279

acatangiu (approved):

pepoviola (approved):

Issue #312, opened 2021-06-28
Vision: Execute a hook after inherent but before transactions


xlc:
We have a lot of per block logics that we want to execute them in the beginning of the block. However we don't really want to use `on_initialize` because it is executed before inherent are been processed. This means data from inherent are not available, such as latest timestamp and relaychain block number. Having those logic in `on_finalize` is also not idea due to risk of exceed block time limit.

PR #2230, opened 2023-11-08
BridgeHub Runtimes: Change registration order of `MessageQueue` pallet

vgeddes:
This PR changes the registration order of the `MessageQueue` pallet so that it is registered last.

This is necessary so that the [on_initialize](https://github.com/Snowfork/snowbridge/blob/df8d5da82e517a65fb0858a4f2ead533290336b5/parachain/pallets/outbound-queue/src/lib.rs#L267) hooks for Snowbridge can run before `MessageQueue` delivers messages using its own `on_initialize`.

Generally, I think this is preferable regardless of Snowbridge's particular requirements. Other pallets may want to do housekeeping before MessageQueue starts delivering messages.

I'm hoping this PR, if accepted, can be included in the same release as https://github.com/paritytech/polkadot-sdk/pull/1246. As otherwise, changing the order of pallet registration is an ABI-breaking change.

### DIFF
--- a/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
+++ b/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
@@ -468,11 +468,14 @@ construct_runtime!(
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin, Config<T>} = 31,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
-		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
+
+		// Message Queue. Importantly, is registered last so that messages are processed
+		// after the `on_initialize` hooks of bridging pallets (snowbridge in particular).
+		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 99,
 	}
 );
 
--- a/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
+++ b/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
@@ -468,14 +468,11 @@ construct_runtime!(
 		PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin, Config<T>} = 31,
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
+		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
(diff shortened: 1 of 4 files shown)


cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -468,11 +468,14 @@ construct_runtime!(
...
 		CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
-		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
+
+		// Message Queue. Importantly, is registered last so that messages are processed
+		// after the `on_initialize` hooks of bridging pallets (snowbridge in particular).
```
ggwpez, line 477: I think just mentioning the other pallets is enough, since there is no bridging pallet (yet).

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -468,11 +468,14 @@ construct_runtime!(
...
+		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 99,
```
ggwpez, line 478:
```suggestion
		MessageQueue: pallet_message_queue::{Pallet, Call, Storage, Event<T>} = 250,
```
Then we still have room at the end and dont need to move the MQ backwards again.

ggwpez (commented):
Just to be clear; it is not an error it we dont change it, or? It would just need one more block to do things?

ggwpez:
>I'm hoping this PR, if accepted, can be included in the same release as https://github.com/paritytech/polkadot-sdk/pull/1246. As otherwise, changing the order of pallet registration is an ABI-breaking change.

I think it should be fine.

bkontur (approved):
@vgeddes 
please, revert changes from bridge-hub-kusama/bridge-hub-polkadot, not needed there, because kusama/polkadot-related runtimes will be removed soon from `polkadot-sdk` repo, we will fix this just in `polkadot-fellows/runtimes` repo for bridge-hub-kusama/bridge-hub-polkadot.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
vgeddes: Fixed in https://github.com/paritytech/polkadot-sdk/pull/2230/commits/91265507fe7e790fac20eea2bbd952d671c332f7.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/lib.rs
vgeddes: Good point! Fixed in https://github.com/paritytech/polkadot-sdk/pull/2230/commits/91265507fe7e790fac20eea2bbd952d671c332f7.

ggwpez (approved):

vgeddes:
> Just to be clear; it is not an error it we dont change it, or? It would just need one more block to do things?

It would result in errors if we don't change the order. Specifically messages outbound to Ethereum would be discarded.

xlc:
We really need https://github.com/paritytech/polkadot-sdk/issues/312
I am sure not every current and future parachain teams will be aware of this and someone is going to waste a lot of time on this exact same issue again.

ggwpez:
> We really need #312 I am sure not every current and future parachain teams will be aware of this and someone is going to waste a lot of time on this exact same issue again.

Yep. https://github.com/paritytech/polkadot-sdk/pull/1781 and https://github.com/polkadot-fellows/RFCs/pull/13. Implementation is about completed.  
Such a hook will be available as `System::PostInherents`.

Issue #1444, opened 2023-09-07
Secure validator mode


mrcnski:
## Overview

**Summary:** Running `--validator` on a platform other than Linux x86-64 requires the `--insecure-validator-i-know-what-i-do flag`.

Due to https://github.com/paritytech/polkadot-sdk/issues/882 becoming high-priority for parathreads, we are now forced to provide a secure validator mode only for Linux x86-64 (to start).

We will still support other platforms with an `--insecure-validator-i-know-what-i-do flag`. (Naming follows [interpreted-i-know-what-i-do](https://github.com/paritytech/substrate/blob//client/cli/src/arg_enums.rs?rgh-link-date=2023-04-13T13%3A54%3A01Z#L58).)

## Todo

Preparation:
- [x] Announce this change to validators ahead of time in a release notes https://github.com/w3f/polkadot-wiki/issues/4881
- [x] Document secure-mode. https://github.com/w3f/polkadot-wiki/pull/5346
- [x] Make unavailable security features print a warning that says this will soon be required to run securely, with a link to the announcement. https://github.com/paritytech/polkadot-sdk/pull/2244

Code:
- [x] Make all candidate-validation security features a requirement to run in secure-mode.
  - [x] Exception: if one FS feature is missing and the other is present, that's fine.
- [x] Have a flag to run in "insecure mode"
- [x] When running in secure-mode, link to https://wiki.polkadot.network/docs/maintain-guides-secure-validator on startup.

Followup:
- [ ] Add automatically flag to zombienet

## Related

Previous draft PR: https://github.com/paritytech/polkadot/pull/7073
Proposed announcement: https://github.com/w3f/polkadot-wiki/issues/4881

pepoviola:
Just a note to me. We should add the logic to enable the flag automatically in zombienet so users can still run from the current configs.

mrcnski:
Currently the "can-unshare-and-change-root" capability is required for secure-mode. But we just had a user for whom it failed: https://github.com/paritytech/polkadot-sdk/pull/2304 *However*, he was on a Linux version (5.15) which should support Landlock. So I think that *either* the unshare capability *or* the Landlock capability should be present, since either provides filesystem security. This way the user can still run securely without a CLI flag. :)

mrcnski:
Hi @hitchhooker! I *think* you should be fine:

- The landlock error is most likely because the Linux version you are using doesn't support it. It is optional if the unshare-and-change-root capability is available.
[...]

PR #2244, opened 2023-11-09
PVF host: Make unavailable security features print a warning

mrcnski:
The warning should say that this will soon be required to run securely, with a link to the announcement.

See https://github.com/paritytech/polkadot-sdk/issues/1444

## Release plan

We plan to get this in the next release (1.4). The merge window for that ends Nov 15.

In the following release (1.5) we will make this warning a hard error. I will raise a PR and merge it sometime starting Nov 16. In the same release (1.5) we will switch on networking restrictions; see https://github.com/paritytech/polkadot-sdk/issues/2163.

[Release calendar](https://www.notion.so/paritytechnologies/Release-Calendar-cc269daa514f4dc996702e50689d1bca)

## Visual Changes

For an example of the warnings and link to [more information](https://wiki.polkadot.network/docs/maintain-guides-secure-validator#secure-validator-mode), see below.

Screenshot 2023-11-10 at 10 55 25

### DIFF
--- a/polkadot/node/core/pvf/src/security.rs
+++ b/polkadot/node/core/pvf/src/security.rs
@@ -29,9 +29,6 @@ const SECURE_MODE_ANNOUNCEMENT: &'static str =
 /// Run checks for supported security features.
 pub async fn check_security_status(config: &Config) -> SecurityStatus {
 	let Config { prepare_worker_program_path, .. } = config;
-	// TODO: Get this from `Config` once Secure Validator Mode is implemented.
-	// Right now we set it to `true`, but only display a warning if conditions are not met.
-	let secure_validator_mode = true;
 
 	// TODO: add check that syslog is available and that seccomp violations are logged?
 	let (landlock, seccomp, change_root) = join!(
@@ -50,7 +47,7 @@ pub async fn check_security_status(config: &Config) -> SecurityStatus {
 		.into_iter()
 		.filter_map(|result| result.err())
 		.collect();
-	let err_occurred = print_secure_mode_message(secure_validator_mode, errs);
+	let err_occurred = print_secure_mode_message(errs);
 	if err_occurred {
 		gum::error!(
 			target: LOG_TARGET,
@@ -102,7 +99,7 @@ impl fmt::Display for SecureModeError {
 /// # Returns
 ///
 /// `true` if an error was printed, `false` otherwise.
-fn print_secure_mode_message(secure_validator_mode: bool, errs: Vec<SecureModeError>) -> bool {
+fn print_secure_mode_message(errs: Vec<SecureModeError>) -> bool {
 	// Trying to run securely and some mandatory errors occurred.
 	const SECURE_MODE_ERROR: &'static str = "🚨 Your system cannot securely run a validator. \
 		 \nRunning validation of malicious PVF code has a higher risk of compromising this machine.";
@@ -115,8 +112,7 @@ fn print_secure_mode_message(secure_validator_mode: bool, errs: Vec<SecureModeEr
 		return false
 	}
 
-	let errs_allowed =
-		!secure_validator_mode || errs.iter().all(|err| err.is_allowed_in_secure_mode());
+	let errs_allowed = errs.iter().all(|err| err.is_allowed_in_secure_mode());
 	let errs_string: String = errs.iter().map(|err| format!("\n  - {}", err)).collect();
 
 	if errs_allowed {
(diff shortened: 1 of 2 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4267156

bkchr (changes requested):
Can we please collect the missing features and print the warning once? And it should also not link to some issue in the wiki. Can we not integrate it into the wiki?

polkadot/node/core/pvf/src/security.rs
```diff
@@ -14,155 +14,247 @@
...
+use std::{fmt, path::Path};
 use tokio::{
 	fs::{File, OpenOptions},
 	io::{AsyncReadExt, AsyncSeekExt, SeekFrom},
 };
 
-/// Check if we can sandbox the root and emit a warning if not.
+const SECURE_MODE_ANNOUNCEMENT: &'static str =
+	"In the next release this will be a hard error by default.
+     \nMore information: https://wiki.polkadot.network/docs/maintain-guides-secure-validator#secure-validator-mode";
```
alindima, line 27: I would also add a note about the new flag that will enable validators to run "unsecurely". They should know that this is will be an option

polkadot/node/core/pvf/src/security.rs
```diff
@@ -14,155 +14,247 @@
...
+	CannotEnableSeccomp(String),
+	CannotUnshareUserNamespaceAndChangeRoot(String),
+}
+
+impl SecureModeError {
+	/// Whether this error is allowed with Secure Validator Mode enabled.
+	fn is_allowed_in_secure_mode(&self) -> bool {
+		use SecureModeError::*;
+		match self {
+			CannotEnableLandlock(_) => true,
```
eskimor, line 77: Wasn't landlock the feature with the lowest requirements on Kernel versions? Why is this optional?

eskimor (approved):

polkadot/node/core/pvf/src/security.rs
mrcnski: The wiki link does mention it. I would rather validators read the full information, instead of being given a flag here and applying it without thinking. When we have the error it will spell out the exact flag here, but maybe it shouldn't...

polkadot/node/core/pvf/src/security.rs
mrcnski: Actually, it's the only one that requires a relatively recent kernel version. Many validators are not on 5.13 yet so we decided not to burden them with this requirement, and we already have FS sandboxing a different way, with pivot_root.

polkadot/node/core/pvf/src/security.rs
bkchr: The error should spell out the entire flag. However, it is fine to wait until this is a hard error.

polkadot/node/core/pvf/src/security.rs
```diff
@@ -14,155 +14,256 @@
...
+impl fmt::Display for SecureModeError {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		use SecureModeError::*;
+		let display = match self {
+			CannotEnableLandlock(err) => format!("Cannot enable landlock, a Linux 5.13+ kernel security feature: {}", err),
+			CannotEnableSeccomp(err) => format!("Cannot enable seccomp, a Linux-specific kernel security feature: {}", err),
+			CannotUnshareUserNamespaceAndChangeRoot(err) => format!("Cannot unshare user namespace and change root, which are Linux-specific kernel security features: {}", err),
+		};
+
+		write!(f, "{}", display)
```
bkchr, line 93:
```suggestion
		match self {
			CannotEnableLandlock(err) => write!(f, "Cannot enable landlock, a Linux 5.13+ kernel security feature: {err}"),
			CannotEnableSeccomp(err) => write!(f, "Cannot enable seccomp, a Linux-specific kernel security feature: {err}"),
			CannotUnshareUserNamespaceAndChangeRoot(err) => write!(f, "Cannot unshare user namespace and change root, which are Linux-specific kernel security features: {err}"),
		}
```

bkchr (approved):

polkadot/node/core/pvf/src/security.rs
mrcnski: Alright cool. Thanks for the review!

sandreim (approved):

PR #2248, opened 2023-11-09
Add a deprecation section to the Contributing notes

juangirini:
A brief explanation of the Deprecation Checklist is added to the Contributing notes with a link to it


docs/CONTRIBUTING.md
```diff
@@ -158,3 +158,12 @@ will run the tests for the specified rust version and specified image
 If you're member of **paritytech** org - you can use command-bot to run various of common commands in CI:
 
 Start with comment in PR: `bot help` to see the list of available commands.
+
+
+## Deprecating code
+
+Polkadot SDK does not stick to [SemVer](https://semver.org/), thus breaking changes can happen between "minor version"
+updates.
```
davxy, line 166: We recently had a chat about this with @bkchr 

At rigor, we currently stick to SemVer, since we:
- always bump major versions on release of crates with version >= 1.0.0
- we always bump minor versions on release of crates with version < 1.0.0

Before v1.0.0 you can break stuff and just bump minor version https://semver.org/#spec-item-4

The only thing to improve IMO is that we bump versions even if we don't increment the `Cargo.toml` file of the crate. So probably it would be nice to add to the guidelines to actually bump major/minor versions in `Cargo.toml` inline with the `SemVer` rules

docs/DEPRECATION_CHECKLIST.md
```diff
@@ -64,6 +62,13 @@ with the Release team.
 The deprecated code finally gets removed.
 Don’t forget to [announce this accordingly](#announce-the-deprecation-and-removal).
 
+## Update the crate's version
+
+Inline with [SemVer's guidelines](https://semver.org/#spec-item-4), update the major/minor version in `Cargo.toml`
+depending on the following:
+- bump the major version when the crate's version >= 1.0.0
+- bump the minor version when the crate's version < 1.0.0
```
muharem, line 70: should this be first aligned with the current release process? it's probably part of the pipeline now, and should it be even done by a developer?

davxy: Indeed. The ideal (IMO) is to have the dev mark the PR as a breaking change. 
The release eng then will bump the version in Cargo.toml and publish the crate on crates.io (possibly keeping it aligned with the version in `Cargo.toml` (now is not 🙄 )

muharem (approved):

seadanda (approved):

michalkucharczyk (approved):

PR #2044, opened 2023-10-26
Support for `keyring` in runtimes

michalkucharczyk:
This functionality is required for #1984.

This PR enables [`sp-keyring`](https://github.com/paritytech/polkadot-sdk/blob/21d36b7b4229c4d5225944f197918cde23fda4ea/substrate/primitives/keyring/src/sr25519.rs#L31-L40) in `no-std` environments, allowing to generate the public key (e.g. `AccountKeyring::Alice.public().to_ss58check()`), which can be later used in the any of built-in [_runtime-genesis-config_ variant](https://github.com/paritytech/polkadot-sdk/blob/21d36b7b4229c4d5225944f197918cde23fda4ea/polkadot/node/service/src/chain_spec.rs#L1066-L1073).

The proposal is as follows:
- expose [`core::Pair` trait](https://github.com/paritytech/polkadot-sdk/blob/d6f15306282e3de848a09c9aa9cba6f95a7811f0/substrate/primitives/core/src/crypto.rs#L832) in `no-std`,
- `full_crypto` feature enables `sign` method,
- `std` feature enables  `generate_with_phrase` and `generate` methods (randomness is required),
- All other functionality, currently gated by `full_crypto` will be available unconditionally (`no-std`):
-- `from_string`
-- `from_string_with_seed`
-- `from seed`
-- `from_seed_slice`
-- `from_phrase`
-- `derive`
-- `verify`

---

Depends on https://github.com/rust-bitcoin/rust-bip39/pull/57

PR #2250, opened 2023-11-09
crypto: `lazy_static` removed, light parser for address URI added

michalkucharczyk:
The `lazy_static` package does not work well in `no-std`: it requires `spin_no_std` feature, which also will propagate into `std` if enabled. This is not what we want.

This PR provides simple address uri parser which allows to get rid of _regex_ which was used to parse the address uri, what in turns allows to remove lazy_static.

Three regular expressions (`SS58_REGEX`,`SECRET_PHRASE_REGEX`,`JUNCTION_REGEX`) were replaced with the parser which unifies all of them.

The new parser does not support Unicode, it is ASCII only.

Related to: #2044


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4270720

koute:
NB: It might be worth noting in the PR description that this is *technically* a breaking change as the new parsing code is ASCII only, unlike the old regexps which were Unicode aware. But AFAIK that was entirely unintentional and I don't think we care, so it'd just be an unnecessary complication.

michalkucharczyk:
> NB: It might be worth noting in the PR description that this is _technically_ a breaking change as the new parsing code is ASCII only, unlike the old regexps which were Unicode aware. But AFAIK that was entirely unintentional and I don't think we care, so it'd just be an unnecessary complication.

good point, I forgot we talked about that. Added.

substrate/primitives/core/src/crypto/address_uri.rs
davxy, line 1: IMO this should be just in `core/src` folder. There is no need for this new `crypto` sub folder because:
1. this is not crypto or a crypto wrapper (crypto as cryptography)
2. for cryptographic primitives there is already `primitives/crypto` folder. As I proposed, at some point there will be a PR to move cryptographic primitives wrappers there (sr255, ecdsa, bandersnatch, ...)

michalkucharczyk: https://github.com/paritytech/polkadot-sdk/pull/2250/commits/b6dbf0d47599aed92f42ba5c548842d955455d77

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+		});
+
+		let mut pass = None;
+		let mut paths = Vec::new();
+		while !input.is_empty() {
+			input = if let Some(mut maybe_pass) = input.strip_prefix("///") {
+				pass = match Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n') {
+					Some(pass) => Some(pass),
+					None => Some(""),
+				};
```
bkchr, line 53:
```suggestion
				pass = Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n');
```

`pass` is an `Option`?

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+				maybe_hard
+			} else if let Some(mut maybe_soft) = input.strip_prefix("/") {
+				let Some(path) = Self::extract_prefix(&mut maybe_soft, &|ch: char| ch != '/')
+				else {
+					return None;
+				};
+				paths.push(path);
+				maybe_soft
+			} else {
+				return None;
```
bkchr, line 73: An error would be better instead of just returning `None`. Especially for the user to know what is wrong.

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+			input = if let Some(mut maybe_pass) = input.strip_prefix("///") {
+				pass = match Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n') {
+					Some(pass) => Some(pass),
+					None => Some(""),
+				};
+				maybe_pass
+			} else if let Some(mut maybe_hard) = input.strip_prefix("//") {
+				let Some(mut path) = Self::extract_prefix(&mut maybe_hard, &|ch: char| ch != '/')
+				else {
+					return None;
```
bkchr, line 58: Same

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+				};
+				assert!(path.len() > 0);
+				// hard path shall contain leading '/', so take it from input.
+				path = &input[1..path.len() + 2];
+				paths.push(path);
+				maybe_hard
+			} else if let Some(mut maybe_soft) = input.strip_prefix("/") {
+				let Some(path) = Self::extract_prefix(&mut maybe_soft, &|ch: char| ch != '/')
+				else {
+					return None;
```
bkchr, line 68: Same

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+	pub ss58: Option<&'a str>,
+	pub paths: Vec<&'a str>,
+	pub pass: Option<&'a str>,
+}
+
+impl<'a> AddressUri<'a> {
+	fn extract_prefix(input: &mut &'a str, is_allowed: &dyn Fn(char) -> bool) -> Option<&'a str> {
+		let output = input.trim_start_matches(is_allowed);
+		let prefix_len = input.len() - output.len();
+		let prefix = if prefix_len > 0 { Some(&input[..prefix_len]) } else { None };
```
bkchr, line 32: This also means that we are expecting that there is at least a new line at the end of the URI, that sounds wrong.

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+	pub fn parse(mut input: &'a str) -> Option<Self> {
+		let ss58 = Self::extract_prefix(&mut input, &|ch: char| {
+			ch.is_ascii_digit() || ch.is_ascii_alphabetic() || ch == ' '
+		});
+
+		let mut pass = None;
+		let mut paths = Vec::new();
+		while !input.is_empty() {
+			input = if let Some(mut maybe_pass) = input.strip_prefix("///") {
+				pass = match Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n') {
```
bkchr, line 50: What is with just the input ending?

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+// limitations under the License.
+
+//! Little util for parsing an address URI. Replaces regular expressions.
+
+#[derive(Debug, PartialEq)]
+/// A container for results of parsing the address uri string.
+pub struct AddressUri<'a> {
+	pub ss58: Option<&'a str>,
+	pub paths: Vec<&'a str>,
+	pub pass: Option<&'a str>,
```
bkchr, line 25: Some docs please.

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Little util for parsing an address URI. Replaces regular expressions.
+
+#[derive(Debug, PartialEq)]
+/// A container for results of parsing the address uri string.
```
bkchr, line 21:
```suggestion
/// A container for results of parsing the address uri string.
#[derive(Debug, PartialEq)]
```

substrate/primitives/core/src/crypto/address_uri.rs
michalkucharczyk: I added it in crypto, because it is util used only in crypto module.

substrate/primitives/core/fuzz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+lazy_static = "1.4.0"
+libfuzzer-sys = "0.4"
+regex = "1.10.2"
+
+[dependencies.sp-core]
+path = ".."
```
bkchr, line 15:
```suggestion
sp-core = { path = ".." }
```

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+	/// Intended to be equivalent of:
+	/// Regex::new(r"^(?P<phrase>[a-zA-Z0-9 ]+)?(?P<path>(//?[^/]+)*)(///(?P<password>.*))?$")
+	pub fn parse(mut input: &'a str) -> Option<Self> {
+		let ss58 = Self::extract_prefix(&mut input, &|ch: char| {
+			ch.is_ascii_digit() || ch.is_ascii_alphabetic() || ch == ' '
+		});
+
+		let mut pass = None;
+		let mut paths = Vec::new();
+		while !input.is_empty() {
```
bkchr, line 48: We don't enforce the "order". Aka first ss58, then paths and then the password.

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,258 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Little util for parsing an address URI. Replaces regular expressions.
+
+#[derive(Debug, PartialEq)]
+/// A container for results of parsing the address uri string.
+pub struct AddressUri<'a> {
+	pub ss58: Option<&'a str>,
```
bkchr, line 23: As this could also be the `phrase`, naming it `ss58` is not totally correct.

bkchr (commented):
Very nice work! Quite easy to read and to understand! (Reads easier than parsing the regex :D)

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: Sorry did not get your comment. This is has nothing to do with new line.
If provided `input` does not start with  `is_allowed` characters, then extracted prefix should be `None`. 
In this case the `output` will be equal to `input` and thus `prefix_len` will be 0.

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: Yes. But regex will return `Some("")` if intput ends with  `"///"` (e.g. `hello world///`), and there is no password.
I added this case to all seamless testing agains regex.

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: the `.` in regex matches everything expect the newline.
So the password is any character till end of input that is different from `\n`.

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: Sorry, can you explain?
The original `regex` was expecting the atoms in order: `phrase|paths|pass`, and the same order is implemented here.

substrate/primitives/core/src/address_uri.rs
bkchr: Sorry, I didn't checked how `trim_start_matches` works!

substrate/primitives/core/src/address_uri.rs
bkchr: Yeah, me not knowing how `trim_start_matches` works :see_no_evil:

substrate/primitives/core/src/address_uri.rs
bkchr: Not my best day :P

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: 047eb835f1

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: 047eb835f1

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: 047eb835f1

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: 047eb835f1

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: 047eb835f1

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4304417

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,302 @@
...
+
+		let mut pass = None;
+		let mut paths = Vec::new();
+		while !input.is_empty() {
+			input = if let Some(mut maybe_pass) = input.strip_prefix("///") {
+				pass = match Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n') {
+					Some(pass) => Some(pass),
+					None => Some(""),
+				};
+				maybe_pass
```
koute, line 83: You could add another helper to simplify this and other branches a little bit, e.g. maybe something like this:

```rust
fn strip_prefix(input: &mut &str, prefix: &str) -> bool {
    if let Some(new_input) = input.strip_prefix(prefix) {
        *input = new_input;
        true
    } else {
        false
    }
}
```

And then this snippet could be:

```suggestion
			if strip_prefix(&mut input, "///") {
				pass = Some(Self::extract_prefix(&mut input, &|ch: char| ch != '\n').unwrap_or(""));
```

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl<'a> AddressUri<'a> {
+	fn extract_prefix(input: &mut &'a str, is_allowed: &dyn Fn(char) -> bool) -> Option<&'a str> {
+		let output = input.trim_start_matches(is_allowed);
+		let prefix_len = input.len() - output.len();
+		let prefix = if prefix_len > 0 { Some(&input[..prefix_len]) } else { None };
+		*input = output;
+		prefix
+	}
```
koute, line 67: Nit: personally I'd put this outside of the `impl`, since it doesn't have to be here, and it's going to be one less level of indentation. But up to you.

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,302 @@
...
+				pass = match Self::extract_prefix(&mut maybe_pass, &|ch: char| ch != '\n') {
+					Some(pass) => Some(pass),
+					None => Some(""),
+				};
+				maybe_pass
+			} else if let Some(mut maybe_hard) = input.strip_prefix("//") {
+				let Some(mut path) = Self::extract_prefix(&mut maybe_hard, &|ch: char| ch != '/')
+				else {
+					return Err(Error::InvalidCharacterInHardPath);
+				};
```
koute, line 88:
```suggestion
				let mut path = Self::extract_prefix(&mut maybe_hard, &|ch: char| ch != '/').ok_or(Error::InvalidCharacterInHardPath)?;
```

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,302 @@
...
+		check("////////", Ok(AddressUri { phrase: None, pass: Some("/////"), paths: vec![] }))
+	}
+
+	#[test]
+	fn test02() {
+		check(
+			"sdasd///asda",
+			Ok(AddressUri { phrase: Some("sdasd"), pass: Some("asda"), paths: vec![] }),
+		);
+		//
```
koute, line 171:
```suggestion
```

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: I like the idea, code looks simpler. The first `if branch` does not fit here however, since we need original input. I'll try to figure out something.

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: [d556091](https://github.com/paritytech/polkadot-sdk/pull/2250/commits/d556091f268b26cbf015e9f606ede17a22f0a668)

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: done: 77ad7f1428

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,305 @@
...
+					Some(pass) => Some(pass),
+					None => Some(""),
+				};
+			} else if strip_prefix(&mut input, "//") {
+				let mut path = extract_prefix(&mut input, &|ch: char| ch != '/')
+					.ok_or(Error::InvalidCharacterInHardPath)?;
+				assert!(path.len() > 0);
+				// hard path shall contain leading '/', so take it from unstripped input.
+				path = &unstripped_input[1..path.len() + 2];
+				paths.push(path);
```
koute, line 99: The `path` here doesn't have to be mutable. (:

```suggestion
				let path = extract_prefix(&mut input, &|ch: char| ch != '/')
					.ok_or(Error::InvalidCharacterInHardPath)?;
				assert!(path.len() > 0);
				// hard path shall contain leading '/', so take it from unstripped input.
				paths.push(&unstripped_input[1..path.len() + 2]);
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4308198

ggwpez:
bot update-ui

substrate/primitives/core/fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+lazy_static = "1.4.0"
+libfuzzer-sys = "0.4"
+regex = "1.10.2"
+
+sp-core = { path = ".." }
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
ggwpez, line 18: Wait, its not part of the workspace? Will it have its own `Cargo.lock` then?

michalkucharczyk: @ggwpez: Yes.

Do you think it is a problem? I included fuzzer as a convenience tool intended to be executed locally.
What would be your advice here?

koute (approved):

substrate/primitives/core/src/address_uri.rs
```diff
@@ -0,0 +1,303 @@
...
+				paths.push(
+					extract_prefix(&mut input, &|ch: char| ch != '/')
+						.ok_or(Error::InvalidCharacterInSoftPath)?,
+				);
+			} else {
+				return if pass.is_some() {
+					Err(Error::InvalidCharacterInPass)
+				} else {
+					Err(Error::InvalidCharacterInPhrase)
+				};
```
bkchr, line 109: Would be nice if the error could directly include the string/character that is wrong. I mean an even much better would be something like

```
Error at parsing: Hello World?/Whats Up
                                                    ^
                                                    Invalid character
```

https://crates.io/crates/miette (but this crate probably being an overkill)

However, all of that would be the cherry on top. Not really required.

bkchr (approved):

sam0x17:
Just a note that `once_cell` is the canonical replacement for `lazy_static` these days, especially now that much of it has been brought into `std` (though you still need the crate for `Lazy`).

bkchr:
We know.

substrate/primitives/core/src/address_uri.rs
michalkucharczyk: done in: 9fa5ff8

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4359184

substrate/primitives/core/src/address_uri.rs
```diff
@@ -288,16 +362,61 @@ mod tests {
...
+	#[test]
+	fn test_invalid_char_info_4() {
+		let expected = "123\\n567\n    ^";
+		let f = format!("{}", InvalidCharacterInfo::new("123\n567", 4));
+		assert_eq!(expected, f);
+	}
+
+	#[test]
+	fn test_invalid_char_info_5() {
+		let expected = "123\\n567\n       ^";
```
bkchr: Nice! :heart:

PR #2251, opened 2023-11-09
xcm-emulator: add Rococo<>Westend bridge and add tests for assets transfers over the bridge

acatangiu:
- switch from Rococo<>Wococo to Rococo<>Westend bridge
- add bidirectional simple tests
- remove Wococo chains from xcm-emulator
- added tests for assets transfers over Rococo<>Westend bridge 

fixes https://github.com/paritytech/parity-bridges-common/issues/2405


cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/lib.rs
```diff
@@ -21,7 +21,7 @@ pub use xcm::{
 	prelude::{AccountId32 as AccountId32Junction, *},
 	v3::{
 		Error,
-		NetworkId::{Rococo as RococoId, Wococo as WococoId},
+		NetworkId::{Rococo as RococoId, Westend as WestendId},
```
NachoPal, line 24:
```suggestion
		NetworkId::Westend as WestendId,
```

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/lib.rs
```diff
@@ -19,7 +19,10 @@ pub use frame_support::assert_ok;
 // Polkadot
 pub use xcm::{
 	prelude::{AccountId32 as AccountId32Junction, *},
-	v3::{Error, NetworkId::Rococo as RococoId},
+	v3::{
+		Error,
+		NetworkId::{Rococo as RococoId, Westend as WestendId},
```
NachoPal:
```suggestion
		NetworkId::Rococo as RococoId,
```

NachoPal (changes requested):

NachoPal (dismissed):

NachoPal:
Looks good. Just `rococo-wococo-system` network should be removed also.

NachoPal:
@acatangiu I would like to keep the `tests` folder structure split by runtimes (`tests/assets/bridge-hub-rococo` & `tests/assets/bridge-hub-westend`)

Is there a good reason for having to merge both in one?

acatangiu:
> @acatangiu I would like to keep the `tests` folder structure split by runtimes (`tests/assets/bridge-hub-rococo` & `tests/assets/bridge-hub-westend`)
> 
> Is there a good reason for having to merge both in one?

IMO all resources (other dirs) make perfect sense to be structured by runtimes, but for `tests` the lines blur since really useful integration tests will be complex scenarios spanning commands across multiple chains. `bridges` tests are a good example here, where I plan to add tests like:
- send XCM from [Relay, SP, Penpal] on one side to [Relay, SP, Penpal] on other side (cover all combos of src/dest)
  - verify in said tests both TX and RX queues of all involved chains
- bidirectional asset transfers between AHs on different sides of the bridge

Where would above tests live? under `tests/bridges/bridge-hub-rococo` or under `tests/bridges/bridge-hub-westend`? do we do a mix? put some tests in one or another based on direction (what about bidirectional tests)? do we just duplicate tests (not cool, they'd be actual identical copies taking time to build&run)?

So instead I went the route of adding `tests/bridges/rococo-westend` which holds all of these Rococo<>Westend specific tests. I am open to alternatives if you have a better idea 😄

acatangiu:
> Looks good. Just `rococo-wococo-system` network should be removed also.

Yup, removed. No traces of Wococo left under `cumulus/parachains/integration-tests`

NachoPal:
> put some tests in one or another based on direction

Yes, that is the idea.

- Rococo -> Westend in `tests/assets/bridge-hub-rococo`
- Westend -> Rococo in `tests/assets/bridge-hub-westend`

> do we just duplicate tests

In the end that's what you are still doing, duplicating tests but all under a single crate. I guess that you'd like to tests both directions, so duplication seems difficult to avoid.

> what about bidirectional tests

Not sure if there would be a need of bidirectional tests (instead of being 2 properly initialised tests of oposite direction), but if there were, I guess that oposite direction would be needed to be tested as well, wouldn't be?

- Rococo -> Westend -> Rococo  in `tests/assets/bridge-hub-rococo`
- Westend -> Rococo -> Westend in `tests/assets/bridge-hub-westend`

If for some reason `AssetHubRococo` and `AssetHubWestend` diverge in the future with different features, it will be more clear to have feature/runtime specific tests under its own folder.

acatangiu:
> If for some reason AssetHubRococo and AssetHubWestend diverge in the future with different features, it will be more clear to have feature/runtime specific tests under its own folder.

xcm-emulator bridge tests should first of all cover usecases as they are deployed to real AHR+BHR and BHW+AHW. So it's actually good to catch any features divergence in `rococo-westend-bridge` integration tests, rather than further down the line.

>> put some tests in one or another based on direction

> Yes, that is the idea.

This seems forced because test is covering both BHs regardless of direction.
By splitting them you cover AHR + BHR TX  in `bridge-hub-rococo-integ-tests` and AHR + BHR RX (still rococo side) in `bridge-hub-westend-integ-tests`  crate and viceversa; which is counter-intuitive to me.

But I'm not gonna split hairs, if you think they're better sitting in BH-specific crates, I'll move them back. Functionally it should not make any difference.

bkontur:
> > But I'm not gonna split hairs, if you think they're better sitting in BH-specific crates, I'll move them back. Functionally it should not make any difference.
> 
> I don't want to block this PR for that reason either. Ultimately, I expect you guys to be the primary contributors to the bridge-related tests, so in principle whatever works best for you should be ok. However, my concern remains about the `BridgeHubRococo/Westend` specific tests possibly getting lost within an unrelated shared `rococo-westend` folder, which would also disrupt the organized structure of the tests folder split by runtimes.
> 
> An example is the `teleport` test between `BridgeHub` and the rest of the System Parachains under the same Global Consensus, which has nothing to do with the `rococo-westend` bridge. This would be also counter-intuitive to me.
> 
> Maybe a second opinion here? @bkontur

> `rococo-westend`


not really, make sense to have tests which touch both `rococo-westend` there, and test who touches just one consensus under BridgeHubConsensus - like `teleport`, more tests we have then better

acatangiu:
@NachoPal @bkontur reverted to original dir structure and added tests for assets transfers over the bridge - please review

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/asset_transfers.rs
```diff
@@ -0,0 +1,155 @@
...
+					..
+				}) => { asset_id: *asset_id == wnd_at_asset_hub_rococo, },
+			]
+		);
+	});
+	wnd_at_asset_hub_rococo
+}
+
+#[test]
+fn send_wnds_from_asset_hub_westend_to_asset_hub_rococo() {
```
bkontur, line 50: @acatangiu @NachoPal 
actually, this test looked better under `rococo-westend` directory, like I tried to say [here](https://github.com/paritytech/polkadot-sdk/pull/2251#issuecomment-1808097043) and Adrian was right [here](https://github.com/paritytech/polkadot-sdk/pull/2251#issuecomment-1803948590) with `tests/bridges/rococo-westend`
BH does not need to know about all use cases and combination.

acatangiu: too late, I asked and didn't get clear preference other than Nacho's so went with that :) this is fine too - functionally it's the same thing

acatangiu: I'll maybe refactor again at the end on reviews on PR if there is consensus to do so - otherwise will stay as is

bkontur (approved):

acatangiu:
@bkontur since your review, I've added tests to cover reverse direction as well - ROCs and WNDs can be transferred back and forth between AHR and AHW - please take another look

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/asset_transfers.rs
```diff
@@ -15,27 +15,7 @@
...
-	);
-
-	let sender_balance_before =
-		<AssetHubRococo as Chain>::account_data_of(AssetHubRococoSender::get()).free;
-	let receiver_rocs_before = AssetHubWestend::execute_with(|| {
-		type Assets = <AssetHubWestend as AssetHubWestendPallet>::ForeignAssets;
-		<Assets as Inspect<_>>::balance(roc_at_asset_hub_westend, &AssetHubWestendReceiver::get())
-	});
-
+fn send_asset_from_asset_hub_rococo_to_asset_hub_westend(id: MultiLocation, amount: u128) {
```
bkontur, line 18: @acatangiu I dont know if this is possible to deduplicate easily:
```
fn send_asset_from_asset_hub_rococo_to_asset_hub_westend
fn send_asset_from_asset_hub_westend_to_asset_hub_rococo
```
to the one generic
```
fn send_asset_from_asset_hub_to_asset_hub
```

they are doing exactly the same thing AH1 -> BH1 -> BH2 -> AH2 and AH2 -> BH2 -> BH1 -> AH1,
we could possibly reuse for BHKusama/BHPolkadot,
maybe with some macro helper? But sure, also no that big issue to have copies.

but very cool that we have bi-directional tests now 👍 👍 👍

bkontur (approved):

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/asset_transfers.rs
```diff
@@ -0,0 +1,208 @@
...
+		assert_ok!(
+			<AssetHubRococo as AssetHubRococoPallet>::PolkadotXcm::limited_reserve_transfer_assets(
+				signed_origin,
+				bx!(destination.into()),
+				bx!(beneficiary.into()),
+				bx!(assets.into()),
+				fee_asset_item,
+				WeightLimit::Unlimited,
+			)
+		);
```
svyatonik, line 47: Maybe also check that we have withdrawn assets from origin account and deposited it to the local dest account? I guess there'll be some events on `AssetHubRococo` for that?

svyatonik: I see there are checks like `assert!(sender_rocs_before > sender_rocs_after);` in the caller functions - so maybe instead of events also check that `reserve_account_rocs_before < reserve_account_rocs_after` if checking events is hard here?

svyatonik (approved):
Looks super cool

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/7

PR #2253, opened 2023-11-09
Different XCM builders, default one requires fee payment

franciscoaguirre:
Adding on top of the new builder pattern for creating XCM programs, I'm adding some more APIs:

```rust
let paying_fees: Xcm<()> = Xcm::builder() // Only allow paying for fees
  .withdraw_asset() // First instruction has to load the holding register
  .buy_execution() // Second instruction has to be `buy_execution`
  .build();

let paying_fees_invalid: Xcm<()> = Xcm::builder()
  .withdraw_asset()
  .build(); // Invalid, need to pay for fees

let not_paying_fees: Xcm<()> = Xcm::builder_unpaid()
  .unpaid_execution() // Needed
  .withdraw_asset()
  .deposit_asset()
  .build();

let explicitly_without_fees: Xcm<()> = Xcm::builder_unsafe() // You can do anything
  .withdraw_asset()
  .deposit_asset()
  .build();
```


Polkadot-Forum:

https://forum.polkadot.network/t/xcm-user-and-developer-experience-improvements/4511/13

polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -17,51 +17,53 @@
...
+	let builder_unpaid_impl = generate_builder_unpaid_impl(&input.ident, data_enum)?;
 	let output = quote! {
-		pub struct XcmBuilder<Call>(Vec<Instruction<Call>>);
+		/// Type used to build XCM programs that require fee payment
+		pub struct XcmBuilder<Call>(pub(crate) Vec<Instruction<Call>>);
+		/// Type used to build XCM programs that require explicitly stating no fees need to be paid
+		pub struct XcmBuilderUnpaid<Call>(pub(crate) Vec<Instruction<Call>>);
+		/// Type used to build arbitrary XCM, without restrictions.
+		/// Should only be used when you know what you're doing, for experimenting or for learning/teaching purposes.
+		pub struct XcmBuilderRaw<Call>(pub(crate) Vec<Instruction<Call>>);
```
gupnik, line 42: These types should probably be generated conditionally (under a feature flag maybe ?) as they are not supposed to be used in production.

Instead of creating multiple types, I would create the same one with different functionality based on an attribute in the macro.

For example, `dev_mode` in pallet construction allows you to omit certain things, that would complain as soon as you remove it for production. Can we do something similar here?

polkadot/xcm/procedural/src/builder_pattern.rs
```diff
@@ -103,13 +105,168 @@ fn generate_methods_for_enum(name: syn::Ident, data_enum: &syn::DataEnum) -> Tok
...
 			pub fn build(self) -> Xcm<Call> {
 				Xcm(self.0)
 			}
 		}
 	};
 	output
 }
+
+/// All instructions that load the holding register
+const LOAD_HOLDING_INSTRUCTIONS: &[&str] = &[
```
gupnik, line 126: Is it not possible to auto-generate this? Otherwise, this will be another thing we will need to keep track of.

polkadot/xcm/procedural/src/builder_pattern.rs
franciscoaguirre: I was thinking they were supposed to be used in production. Sure, we could just use the raw one but the paid and unpaid are really good for making sure we don't forget to pay or explicitly show we are not paying. I don't see a reason not to allow these for everyone using the xcm crate

polkadot/xcm/procedural/src/builder_pattern.rs
franciscoaguirre: I couldn't find a way, since these are instructions that we know behave in a certain way. I agree it is another thing to keep in mind when adding or removing instructions and that's a bummer.

Actually, one thing we could do is add an attribute on top of those enum variants, something like `#[loads_holding]`. Then, we wouldn't need to maintain this list separately.

polkadot/xcm/procedural/src/builder_pattern.rs
gupnik: But there's a comment there stating `Should only be used when you know what you're doing, for experimenting or for learning/teaching purposes.` ?

Bullrich (approved):

polkadot/xcm/procedural/src/builder_pattern.rs
franciscoaguirre: Oh yeah, we could probably feature flag only the raw one, you're right. That'd be good. However, my other builders piggyback on that one, so probably do it private and feature flag a public version?

KiChjang: You can also make use of typestates, which would obviate the need to have different structs that look almost the same (albeit you still need some unit structs).

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4349229

franciscoaguirre:
bot fmt

polkadot/xcm/procedural/tests/ui.rs
```diff
@@ -28,5 +28,5 @@ fn ui() {
 	std::env::set_var("SKIP_WASM_BUILD", "1");
 
 	let t = trybuild::TestCases::new();
-	t.compile_fail("tests/ui/*.rs");
+	t.compile_fail("tests/ui/*/*.rs");
```
KiChjang, line 31: Does this make it only test the files 1 directory deeper, or does it also test the files under the `tests/ui` directory?

franciscoaguirre: It only worked for files 1 directory deeper, fixed it by using **

KiChjang (approved):

PR #2254, opened 2023-11-09
Give up liveness slashing

s0me0ne-unkn0wn:
In accordance with #1964, we give up liveness slashing. The first step is to always return zero offense threshold. This will be followed up by removing the `im-online` pallet entirely.

### DIFF
--- a/substrate/frame/im-online/src/lib.rs
+++ b/substrate/frame/im-online/src/lib.rs
@@ -855,15 +855,8 @@ impl<Offender: Clone> Offence<Offender> for UnresponsivenessOffence<Offender> {
 		DisableStrategy::Never
 	}
 
-	fn slash_fraction(&self, offenders: u32) -> Perbill {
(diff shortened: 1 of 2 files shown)


ordian (approved):

eskimor (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4274022

bkchr:
I don't get why we need this pr. Why we not just remove the pallet from runtime.

juangirini:
> I don't get why we need this pr. Why are we not just remove the pallet from the runtime?

I will adhere to this. Do we really need to change this pallet functionality for a particular use case that can be achieved by just not using it?

bkchr:
Yeah. I also don't like this and I really don't see any reason. I'm going to close this for the other pr that just removes the stuff.

PR #2262, opened 2023-11-10
Bulk Assigner Test Implementations

BradleyOlson64:
Implemented bulk assigner tests covering all invariants thought of thus far.


polkadot/runtime/parachains/src/assigner_bulk/tests.rs
```diff
@@ -461,12 +534,21 @@ fn assignment_proportions_in_core_state_work() {
 				Some(BulkAssignment::Bulk(task_1.into()))
 			);
 			// Pos should have incremented, as assignment had remaining < step
-			assert_eq!(Workload::<Test>::get(core_idx).pos, 1u16);
+			assert_eq!(
+				CoreDescriptors::<Test>::get(core_idx)
```
eskimor, line 538: One more pop would be nice, to see that it advances to task 2. You are checking pos already, but still would be nice :-)

Issue #2108, opened 2023-11-01
Improved TryState and TryOnRuntimeUpgrade runtime api parameters


liamaharon:
Follow up to https://github.com/paritytech/substrate/pull/13382 and some suggestions by @kianenigma and @ggwpez.

## TryState

I suggest an overhaul of `TryStateSelect` into a struct `TryStateConfig`, looking something like:

```rust
#[derive(codec::Encode, codec::Decode, Clone, scale_info::TypeInfo, Debug)]
pub struct TryStateConfig {
	/// Select which pallets to run the try-state checks on.
	pub pallet_select: PalletSelect,
	/// Filter the checks based on time complexity.
	pub speed_select: SpeedSelect,
	/// Whether to also attempt to decode the entire pallet storage.
	pub decode_entire_storage: bool,
}

/// Choose pallets to execute try-runtime checks on.
#[derive(codec::Encode, codec::Decode, Clone, scale_info::TypeInfo)]
pub enum PalletSelect {
	/// None of them.
	None,
	/// All of them.
	All,
	/// Run a fixed number of them in a round robin manner.
	RoundRobin(u32),
	/// Run only pallets who's name matches the given list.
	///
	/// Pallet names are obtained from [`super::PalletInfoAccess`].
	Only(Vec<Vec<u8>>),
}

/// Filter try-runtime depending on their speed of execution.
#[derive(codec::Encode, codec::Decode, Clone, scale_info::TypeInfo, Debug)]
pub enum SpeedSelect {
	/// Only run checks deemed 'fast'. There is no strict definition of what 'fast' means, but as
	/// a rule of thumb, if the check takes longer than 60 seconds to run on commonly used networks
	/// it probably should not be considered 'fast'.
	Fast,
	/// Run all checks, including those which may take hours to execute.
	All,
}
```

- `speed_select` will be available in pallet try-state definitions, allowing slow-running checks to be filtered out by pallet developers. 

- Using a struct allowed easily extending the api to support new checks like https://github.com/paritytech/polkadot-sdk/pull/1805

- This is a breaking change, so callers of these runtime apis (e.g try-runtime-cli) should optimistically try the new config format and on error fallback onto the old schema, warning users to upgrade their runtime apis.

## TryOnRuntimeUpgrade

I suggest replacing the `UpgradeCheckSelect` enum with a struct like

```rust
#[derive(codec::Encode, codec::Decode, Clone, Debug, scale_info::TypeInfo)]
[...]

PR #2263, opened 2023-11-10
Move `AllPalletsWithSystem::decode_entire_state` to own runtime API

liamaharon:
`AllPalletsWithSystem::decode_entire_state` is currently called in `Executive::try_runtime_upgrade` and `Executive::try_execute_block` if any checks are selected.

This is very inflexible and cripples the usefulness of existing runtime APIs if there are any issues with decoding the entire state, as we are currently seeing with Asset Hub Westend CI.

We need to make running these checks optional. 

Rather than add complexity to the existing runtime APIs to make the checks optional, in the spirit of @xlc's suggestions here https://github.com/paritytech/polkadot-sdk/issues/2108#issuecomment-1788277595, I've opted to create a new runtime api for these checks. This pushes the complexity of deciding whether and when to run the checks out of the runtime and into the caller (such as `try-runtime-cli`).

This approach seems cleaner to me. If in agreement, in the future we may wish to deprecate the existing `try-runtime` runtime APIs and replace them with simpler, minimal configuration runtime APIs that perform atomic pieces of work.

### DIFF
--- a/substrate/bin/node/runtime/src/lib.rs
+++ b/substrate/bin/node/runtime/src/lib.rs
@@ -2747,6 +2747,12 @@ impl_runtime_apis! {
 			// have a backtrace here.
 			Executive::try_execute_block(block, state_root_check, signature_check, select).unwrap()
 		}
+
+		fn decode_entire_state() {
+			// NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+			// have a backtrace here.
+			Executive::try_decode_entire_state().unwrap()
+		}
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -322,10 +322,6 @@ where
 			log::error!(target: LOG_TARGET, "failure: {:?}", e);
 			e
 		})?;
-		if select.any() {
-			let res = AllPalletsWithSystem::try_decode_entire_state();
-			Self::log_decode_result(res)?;
-		}
 		drop(_guard);
 
 		// do some of the checks that would normally happen in `final_checks`, but perhaps skip
@@ -362,6 +358,12 @@ where
 		Ok(frame_system::Pallet::<System>::block_weight().total())
 	}
 
+	pub fn try_decode_entire_state() -> Result<(), TryRuntimeError> {
+		let _ = frame_support::StorageNoopGuard::default();
+		let res = AllPalletsWithSystem::try_decode_entire_state();
+		Self::log_decode_result(res)
+	}
+
 	/// Execute all `OnRuntimeUpgrade` of this runtime.
 	///
 	/// The `checks` param determines whether to execute `pre/post_upgrade` and `try_state` hooks.
@@ -374,12 +376,6 @@ where
 			)?;
 		// Nothing should modify the state after the migrations ran:
 		let _guard = StorageNoopGuard::default();
-
-		// The state must be decodable:
-		if checks.any() {
-			let res = AllPalletsWithSystem::try_decode_entire_state();
-			Self::log_decode_result(res)?;
-		}
 
 		// Check all storage invariants:
 		if checks.try_state() {
--- a/cumulus/parachain-template/runtime/src/lib.rs
+++ b/cumulus/parachain-template/runtime/src/lib.rs
@@ -722,6 +722,13 @@ impl_runtime_apis! {
 			// have a backtrace here.
 			Executive::try_execute_block(block, state_root_check, signature_check, select).unwrap()
 		}
+
+		fn decode_entire_state() {
+			// NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+			// have a backtrace here.
+			Executive::try_decode_entire_state().unwrap()
+		}
+
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
--- a/cumulus/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs
+++ b/cumulus/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs
@@ -1188,6 +1188,12 @@ impl_runtime_apis! {
 			// have a backtrace here.
 			Executive::try_execute_block(block, state_root_check, signature_check, select).unwrap()
 		}
+
+		fn decode_entire_state() {
+			// NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+			// have a backtrace here.
+			Executive::try_decode_entire_state().unwrap()
+		}
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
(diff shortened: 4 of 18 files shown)


substrate/bin/node/runtime/src/lib.rs
```diff
@@ -2747,6 +2747,12 @@ impl_runtime_apis! {
 			// have a backtrace here.
 			Executive::try_execute_block(block, state_root_check, signature_check, select).unwrap()
 		}
+
+		fn decode_entire_state() {
+			// NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+			// have a backtrace here.
+			Executive::try_decode_entire_state().unwrap()
```
bkchr, line 2754: Why not just panic inside `try_decode_entire_state` instead of copying this comment everywhere? :see_no_evil:

substrate/frame/executive/src/lib.rs
```diff
@@ -375,12 +377,6 @@ where
 		// Nothing should modify the state after the migrations ran:
 		let _guard = StorageNoopGuard::default();
 
-		// The state must be decodable:
-		if checks.any() {
-			let res = AllPalletsWithSystem::try_decode_entire_state();
-			Self::log_decode_result(res)?;
-		}
```
bkchr, line 382: Why do we have the `checks` parameter? I mean we could also just add a new variant to check? So no new runtime api function.

liamaharon: `UpgradeCheckSelect` is currently an enum:

```rust
pub enum UpgradeCheckSelect {
	None,
	All,
	PreAndPost,
	TryState,
}
```

So if we want to make `try_decode_entire_state` configurable by that arg, I think we'd want to update it to a struct something like this:

```rust
struct UpgradeCheckSelect {
	pre_and_post: bool,
	try_state: Select,
	decode_entire_state: bool	
}
```

Otherwise we'd need to add an enum variant for every combination of checks the developer wants to run. Let me refactor to do that..

I there there still is value to de-coupling the try-state and decode-entire-state checks from on-runtime-upgrade and execute-block methods, but that requires a bigger change and can leave it for a future PR. Let me work on the struct refactor first.

liamaharon: Let me answer this question in 2 parts.

1. Suggest how we could add it to checks parameter, but why I don't think this is the best solution
2. Suggest an alternative

## 1. How we could add it to `checks` 

`UpgradeCheckSelect` is currently an enum:

```rust
pub enum UpgradeCheckSelect {
	None,
	All,
	PreAndPost,
	TryState,
}
```

To keep the enum, we'd need to add a new variant for every combination of checks the user may want to run. Which is obviously not scalable or a good solution.

Instead, I think we'd want to change it to a struct something like this:

```rust
struct UpgradeCheckSelect {
	pre_and_post: bool,
	try_state: Select,
	decode_entire_state: bool	
}
```

However, I don't think this is optimal for 2 reasons: 

1. It's a breaking change every time we want to add a new `try-runtime` task
2. It's inflexible in the sense that we can only run checks in combination with an `on_runtime_upgrade` check or an `execute_block` check. What if we just want to decode entire state? or run try-state checks?
3. We need to clutter the configuration to both `try_on_runtime_upgrade` and `try_execute_block` with the check options

## 2. Proposed alternative 

1. Add a new `enum` describing all the atomic `try-runtime` related tasks, something like
```rust
pub enum TryRuntimeTask {
	OnRuntimeUpgrade(pre_and_post: bool),
	ExecuteBlock(block: Block, state_root_check: bool, signature_check: bool),
	TryState(TryStateSelect),
	TryDecodeEntireState
}
```

4. Remove all `try-state` and `try-decode-entire-state` logic from `try_on_runtime_upgrade` and `try_execute_block` entirely, instead create dedicated methods  for running those checks `try_state` and `try_decode_entire_state`.

2. Create a new `Executive` method and runtime API that accepts a vec of `TryRuntimeTasks`, and executes them in order

```rust
// runtime api
fn execute_try_runtime_tasks(tasks: Vec<TryRuntimeTask>) -> (Weight, Weight) {
	Executive::execute_try_runtime_tasks(tasks)
}

// executive method
pub fn execute_try_runtime_tasks(Vec<TryRuntimeTasks>) -> (Weight, Weight) {
	let agg_weight = Weight::zero();
	for task in tasks {
		match task {
			TryRuntimeTask::OnRuntimeUpgrade(pre_and_post) => {
				let weight = Self::try_on_runtime_upgrade(pre_and_post);
				agg_weight.saturating_acc(weight);
			}
			TryRuntimeTask::ExecuteBlock(block, state_root_check, signature_check) => {
				let weight = Self::try_execute_block(block, state_root_check, signature_check, select).unwrap();
				agg_weight.saturating_acc(weight);
			}
			TryRuntimeTask::TryState(try_state_select) {
				Self::try_state(try_state_select);
			}
			TryRuntimeTask::TryDecodeEntireState {
				Self::try_decode_entire_state();
			}
		}
	}
	(weight, BlockWeights::get().max_block)
}
```

This allows `try_on_runtime_upgrade` and `try_execute_block` to not be concerned about *what* checks to run or *when* (before or after) to run them or *how* (configuration) to run them. 

The developer instead, with ultimate flexibility, just specifies a Vec like `vec![OnRuntimeUpgrade(OnRuntimeUpgradeConfig), TryState(TryStateConfig), TryDecodeEntireState]` describing what they want to run in what order.

Once this is implemented, we can add a deprecation notice to the current try-runtime runtime APIs and eventually remove them.

bkchr: Okay ty for the writeup. 

> It's a breaking change every time we want to add a new try-runtime task

Your second proposal has the same issues ;) I don't think in general that this is such a problem, because these are development apis that don't need to stay stable or we need to support them over X years.

> It's inflexible in the sense that we can only run checks in combination with an on_runtime_upgrade check or an execute_block check. What if we just want to decode entire state? or run try-state checks?

Not sure why it also requires the `on_runtime_upgrade` check. However, it will always require that we run the runtimes upgrades. You always need to run the upgrades as the new code you are running could be incompatible to the old runtime that created the state. This actually shows a "flaw" in your current pr here, because you don't run the upgrades before.

liamaharon: > Your second proposal has the same issues ;) I don't think in general that this is such a problem, because these are development apis that don't need to stay stable or we need to support them over X years.

If it's an enum then I think we can add new variants without it being a breaking change to the caller (e.g. try-runtime-cli)? Since `try-runtime-cli` will just construct the Vec with enums it knows about. 

You're right though not a huge deal, and we likely won't change it frequently so not a huge deal.. 

Do you feel I should proceed with the struct approach (described in 1.) then?

bkchr: > If it's an enum then I think we can add new variants without it being a breaking change to the caller (e.g. try-runtime-cli)? Since `try-runtime-cli` will just construct the Vec with enums it knows about.

Yeah that is a good point. 

Maybe we could also just introduce a bitflag, assuming we don't need to pass some data. However, if we need to pass some data, I would probably use your enum approach given your reasoning above.

PR #2271, opened 2023-11-10
XCM: Silently drop v3 SetTopic when translating into v2.

gavofyork:
Should not be merged unless we decide this is needed.

Note this will break indexers which rely on `SetTopic` being honoured. They should ignore chains considered by others as being on v2.


bkchr:
I assume it is not required anymore.

PR #2275, opened 2023-11-10
Add `s` utility function to frame support

olanod:
A utility function I consider quite useful to declare string literals that are backed by an array.

### DIFF
--- a/substrate/primitives/runtime/src/lib.rs
+++ b/substrate/primitives/runtime/src/lib.rs
@@ -954,6 +954,23 @@ pub fn print(print: impl traits::Printable) {
 	print.print();
 }
 
+/// Utility function to declare string literals backed by an array of length N.
+/// The input can be shorter than N, in that case the end of the array is padded with zeros.
+///
+/// [`str_array`] is useful when converting strings that end up in the storage as fixed size arrays
+/// or in const contexts where static data types have strings that could also end up in the storage.
+pub const fn str_array<const N: usize>(s: &str) -> [u8; N] {
+	debug_assert!(s.len() <= N, "String literal doesn't fit in array");
+	let mut i = 0;
+	let mut arr = [0; N];
+	let s = s.as_bytes();
+	while i < s.len() {
+		arr[i] = s[i];
+		i += 1;
+	}
+	arr
+}
+
 /// Describes on what should happen with a storage transaction.
 pub enum TransactionOutcome<R> {
 	/// Commit the transaction.
--- a/substrate/primitives/runtime/src/lib.rs
+++ b/substrate/primitives/runtime/src/lib.rs
@@ -955,6 +955,7 @@ pub fn print(print: impl traits::Printable) {
 }
 
 /// Utility function to declare string literals backed by an array of length N.
+///
 /// The input can be shorter than N, in that case the end of the array is padded with zeros.
 ///
 /// [`str_array`] is useful when converting strings that end up in the storage as fixed size arrays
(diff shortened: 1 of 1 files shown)


ggwpez:
>A utility function I consider quite useful to declare string literals that are backed by an array.

It looks useful, but why does it have to live in the Polkadot-SDK?

olanod:
> It looks useful, but why does it have to live in the Polkadot-SDK or `frame-support`? It does not seem FRAME specific to me in any way.

@ggwpez we often use arrays to represent string-like data that is hard-coded in the runtime or that might end up in the storage, with `s` is more ergonomic to declare a `s("string literal")`. My current use case is for https://github.com/paritytech/polkadot-sdk/pull/2072

substrate/frame/support/src/lib.rs
```diff
@@ -117,12 +117,27 @@ pub const LOG_TARGET: &str = "runtime::frame-support";
 #[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
 pub enum Never {}
 
+/// Utility function to declare string literals backed by an array of length N.
+/// The input can be shorter than its underlaying array in which case the end of the array is padded
+/// with zeros. It can be used in const contexts.
+pub const fn s<const N: usize>(s: &str) -> [u8; N] {
```
bkchr, line 123: Maybe we put this into `sp-runtime` as @ggwpez said. General I'm not happy with this position either. The naming should also be improved `s` is not really expressive.

olanod: No problem, changing to `str_array`, I use it often in embedded projects to the point `s` becomes a "well-known identifier"

substrate/primitives/runtime/src/lib.rs
```diff
@@ -954,6 +954,23 @@ pub fn print(print: impl traits::Printable) {
 	print.print();
 }
 
+/// Utility function to declare string literals backed by an array of length N.
+/// The input can be shorter than N, in that case the end of the array is padded with zeros.
```
davxy, line 958:
```suggestion
///
/// The input can be shorter than N, in that case the end of the array is padded with zeros.
```
Nit to stick with the docs guidelines

substrate/frame/support/src/lib.rs
bkchr: Yeah I also have seen this before, but I like it to be more expressive. Ty!

bkchr (approved):

substrate/primitives/runtime/src/lib.rs
```diff
@@ -954,6 +954,24 @@ pub fn print(print: impl traits::Printable) {
 	print.print();
 }
 
+/// Utility function to declare string literals backed by an array of length N.
+///
+/// The input can be shorter than N, in that case the end of the array is padded with zeros.
+///
+/// [`str_array`] is useful when converting strings that end up in the storage as fixed size arrays
+/// or in const contexts where static data types have strings that could also end up in the storage.
```
ggwpez, line 962:
```suggestion
/// or in const contexts where static data types have strings that could also end up in the storage.
///
/// # Example
///
/// ```rust
/// # use sp_runtime::str_array;
/// const MY_STR: [u8; 6] = str_array("data");
/// assert_eq!(MY_STR, *b"data\0\0");
/// ```
```

Otherwise we dont know if it works 😄

ggwpez (approved):
Please update title, since it will be in the changelog.

PR #2277, opened 2023-11-11
Bulk assigner tests update

BradleyOlson64:
Addressed comments from https://github.com/paritytech/polkadot-sdk/pull/2262

Also added tests to cover the new cases:
- Test that assignments are served correctly. E.g. two equal assignments will be served as ABABAB
- Have a test that checks that core is shared fairly, even in case of `ratio` not being divisible by `step`


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4295643

eskimor (approved):
Nice! Thank you Bradley!

Issue #2241, opened 2023-11-09
Asset Hub Westend has undecodable `Balances::Hold` and `Nfts::ItemMetadataOf` state


liamaharon:
https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4265399

liamaharon:
Likely culprit for `Nfts::ItemMetadataOf` is storage schema change wo migration: https://github.com/paritytech/substrate/pull/13124/files#diff-97b1c740f13b52f2aff6c03ba154b7f8a0313c2c579fc4095e35726b834d3814L266

Likely culprit for `Balances::Hold` is a hold being there with a Reason that is no longer valid.

Will continue investigating next week.

liamaharon:
`NftItemMetadataOf` collection 3 is indeed in an incorrect format due to the above change made without a migration.

Here is the state of collection 3, item 1, queried at the block it was created (so we can decode it), compared to the state of collection 4, item 1, queried at the current block.

<img width="1508" alt="Screenshot 2023-11-14 at 12 43 52" src="https://github.com/paritytech/polkadot-sdk/assets/16665596/6735b47c-e06c-4c40-b02b-e72061128cd0">

liamaharon:
This `Hold` has a `RuntimeHoldReason` `__Unused0`, which is not currently a `RuntimeHoldReason` making it undecodable.

<img width="1413" alt="Screenshot 2023-11-14 at 12 53 26" src="https://github.com/paritytech/polkadot-sdk/assets/16665596/67567939-3f83-418b-a448-a8637a4f4ff3">

PR #2283, opened 2023-11-13
Disable `try_decode_entire_state` try-runtime checks

liamaharon:
These checks are currently breaking CI. 

See this PR for more info: https://github.com/paritytech/polkadot-sdk/pull/2263 

The checks will be reenabled in the above PR in a more configurable and scalable way. 

The reason I am disabling them in a seperate PR is because the above PR may take some time to merge, and we need to fix the CI asap.

### DIFF
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -322,10 +322,6 @@ where
 			log::error!(target: LOG_TARGET, "failure: {:?}", e);
 			e
 		})?;
-		if select.any() {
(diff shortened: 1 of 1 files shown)


bkchr:
Maybe we should just look into fixing the chains to make the CI check pass? :D

liamaharon:
> Maybe we should just look into fixing the chains to make the CI check pass? :D

Yup I am here: https://github.com/paritytech/polkadot-sdk/issues/2241

liamaharon:
It will be fixed once Westend is unstuck.

PR #2295, opened 2023-11-13
chainHead: Support multiple hashes for `chainHead_unpin` method

lexnv:
This PR adds support for multiple hashes being passed to the `chainHeda_unpin` parameters.

The `hash` parameter is renamed to `hash_or_hashes` per https://github.com/paritytech/json-rpc-interface-spec/pull/111.

While at it, a new integration test is added to check the unpinning of multiple hashes. The API is checked against a hash or a vector of hashes.

cc @paritytech/subxt-team

### DIFF
--- a/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
+++ b/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
@@ -435,12 +435,14 @@ where
 		follow_subscription: String,
 		hash_or_hashes: ListOrValue<Block::Hash>,
 	) -> RpcResult<()> {
-		let hashes = match hash_or_hashes {
-			ListOrValue::Value(hash) => vec![hash],
-			ListOrValue::List(hashes) => hashes,
+		let result = match hash_or_hashes {
+			ListOrValue::Value(hash) =>
+				self.subscriptions.unpin_blocks(&follow_subscription, [hash]),
+			ListOrValue::List(hashes) =>
+				self.subscriptions.unpin_blocks(&follow_subscription, hashes),
 		};
 
-		match self.subscriptions.unpin_blocks(&follow_subscription, hashes) {
+		match result {
 			Ok(()) => Ok(()),
 			Err(SubscriptionManagementError::SubscriptionAbsent) => {
 				// Invalid invalid subscription ID.
--- a/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
+++ b/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
@@ -21,6 +21,7 @@
 use super::{
 	chain_head_storage::ChainHeadStorage,
 	event::{MethodResponseStarted, OperationBodyDone, OperationCallDone},
+	HashOrHashes,
 };
 use crate::{
 	chain_head::{
@@ -432,7 +433,7 @@ where
 	fn chain_head_unstable_unpin(
 		&self,
 		follow_subscription: String,
-		hash: Block::Hash,
+		hash: HashOrHashes<Block::Hash>,
 	) -> RpcResult<()> {
 		match self.subscriptions.unpin_block(&follow_subscription, hash) {
 			Ok(()) => Ok(()),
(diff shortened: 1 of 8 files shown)


substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -432,9 +433,14 @@ where
 	fn chain_head_unstable_unpin(
 		&self,
 		follow_subscription: String,
-		hash: Block::Hash,
+		hash_or_hashes: ListOrValue<Block::Hash>,
 	) -> RpcResult<()> {
-		match self.subscriptions.unpin_block(&follow_subscription, hash) {
+		let hashes = match hash_or_hashes {
+			ListOrValue::Value(hash) => vec![hash],
```
niklasad1, line 439: perhaps you could use the iterator impl instead and just do std::iter::once(hash) instead of creating a vec here

substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
```diff
@@ -750,22 +750,46 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionsInner<Block, BE> {
...
-	pub fn unpin_block(
+	pub fn unpin_blocks(
 		&mut self,
 		sub_id: &str,
-		hash: Block::Hash,
+		hashes: Vec<Block::Hash>,
 	) -> Result<(), SubscriptionManagementError> {
-		let Some(sub) = self.subs.get_mut(sub_id) else {
+		// Check if the subscription ID is valid or not.
+		if self.subs.get(sub_id).is_none() {
```
niklasad1, line 759: hmm, how expensive is it to clone the SubscriptionState here? 
It's a bit annoying to do that many HashMap looks here

lexnv: To work around this, since we cannot use `RefCell` because `ChainHead` needs to be `Send`, I've removed the mappings from the subscription first, which requires only one `HashMap::get`.

Then, I've removed the global mappings of the hashes. The downside if that we need to do `O(3*n)` instead of ideal `O(2*n)`, but given the hashes provided in unpin are bounded to the total of the subscription I would say this is pretty good in terms of performance

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -1645,6 +1651,160 @@ async fn follow_with_unpin() {
...
+	let _res: () = api
+		.call("chainHead_unstable_unpin", rpc_params![&sub_id, &block_1_hash])
+		.await
+		.unwrap();
+
+	// One block hash is invalid. Block 1 is already unpinned.
+	let err = api
+		.call::<_, serde_json::Value>(
+			"chainHead_unstable_unpin",
+			rpc_params![&sub_id, vec![&block_1_hash, &block_2_hash, &block_3_hash]],
```
niklasad1, line 1770: :+1:

niklasad1 (approved):
nice, LGTM

substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
niklasad1: Ok, yeah realized that now but as you said the number pinned blocks are bounded and should fast enough I reckon as well.

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
lexnv: That makes total sense, have made the lower levels of `unpin_blocks` take `impl IntoIterator< ...>`. 

To remove the allocation, I've called into `self.subscriptions.unpin_blocks` with different iterators. Since, on one hand, we have `std::iter::Once` for a single value, and on the other, we have `vec`. Using some type-erasure would still make an allocation :D

tadeohepperle (approved):
Nice!

PR #2299, opened 2023-11-13
Add CI to claim crates

Morganamilo:


.github/workflows/claim-crates.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  claim-crates:
+    strategy:
+      matrix:
+        os: ["ubuntu-latest"]
```
coderobe, line 12: why use a matrix for this?

Morganamilo: Other workflows do so just staying consistent.

coderobe: hmm i don't think it makes sense here since this will never need to be ran on more than one

Morganamilo: Sure I'll remove it then. And from the other publish CI

alvicsam (approved):

.github/workflows/claim-crates.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8 # v2.7.1
+        with:
+          cache-on-failure: true
+
+      - name: install parity-publish
+        run: cargo install parity-publish
```
altaua, line 23: can we pin the crate version?

Morganamilo: Any reason? It's a crate controlled by us so we push updates as needed. But I can pin it if that's wanted

Morganamilo: Pinned now.

coderobe (approved):

PR #2304, opened 2023-11-13
PVF: fix detection of unshare-and-change-root security capability

mrcnski:
## Overview 

One of our security features isn't working on a user's machine. Right now this feature should be optional, but the bug is that the node detects that the feature works, so the worker tries to apply it.

The problem is that the detection and application are happening in different directories. Detection tries `/tmp` whereas application tries the database path.

User reported back:

> there are different permissions for the /tmp/ folder (root ownership) vs the
.local folder (my user account)

## Error

For discoverability:

```
ERROR parachain::pvf-common: Could not change root to be the worker cache path: mount MS_BIND: Permission denied (os error 13) worker_kind=prepare
```

### DIFF
--- a/polkadot/node/core/pvf/src/host.rs
+++ b/polkadot/node/core/pvf/src/host.rs
@@ -35,6 +35,7 @@ use polkadot_node_core_pvf_common::{
 	error::{PrepareError, PrepareResult},
 	pvf::PvfPrepData,
 };
+use polkadot_node_subsystem::SubsystemResult;
 use polkadot_parachain_primitives::primitives::ValidationResult;
 use std::{
 	collections::HashMap,
@@ -203,11 +204,14 @@ impl Config {
 /// The future should not return normally but if it does then that indicates an unrecoverable error.
 /// In that case all pending requests will be canceled, dropping the result senders and new ones
 /// will be rejected.
-pub async fn start(config: Config, metrics: Metrics) -> (ValidationHost, impl Future<Output = ()>) {
+pub async fn start(
+	config: Config,
+	metrics: Metrics,
+) -> SubsystemResult<(ValidationHost, impl Future<Output = ()>)> {
 	gum::debug!(target: LOG_TARGET, ?config, "starting PVF validation host");
 
 	// Run checks for supported security features once per host startup. Warn here if not enabled.
-	let security_status = security::check_security_status(&config).await;
+	let security_status = security::check_security_status(&config).await?;
 
 	let (to_host_tx, to_host_rx) = mpsc::channel(10);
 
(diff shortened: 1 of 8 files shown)


polkadot/node/core/pvf/src/host.rs
```diff
@@ -203,11 +204,14 @@ impl Config {
 /// The future should not return normally but if it does then that indicates an unrecoverable error.
 /// In that case all pending requests will be canceled, dropping the result senders and new ones
 /// will be rejected.
-pub async fn start(config: Config, metrics: Metrics) -> (ValidationHost, impl Future<Output = ()>) {
+pub async fn start(
+	config: Config,
+	metrics: Metrics,
+) -> SubsystemResult<(ValidationHost, impl Future<Output = ()>)> {
```
mrcnski, line 210: Already was going to add `-> SubsystemResult` here in another PR soon.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4315856

s0me0ne-unkn0wn (approved):
Looks good!

More of a general thought than a comment on this very PR: the `security` module has a lot of code gated with `#[cfg(target_os = "linux")]`. Indeed, all the security features we use are only available on Linux. Thus, we assume that only the validator running on Linux is secure. Why not gate the whole module with `#[cfg(target_os = "linux")]` then? On other platforms, there could be a single big fat warning like "you're not running Linux so you're not safe", and you wouldn't have to gate literally everything with like `#[cfg_attr(not(target_os = "linux"), allow(unused_variables))]`, etc.

mrcnski:
@s0me0ne-unkn0wn Yeah, we could do a bit of a refactor there. I think there's only one security function that works on Mac (the env vars one). I think my original idea was to leave things general but it's unlikely that Mac will adopt any of the other features soon...

polkadot/node/core/pvf/src/host.rs
```diff
@@ -203,11 +204,14 @@ impl Config {
...
 	gum::debug!(target: LOG_TARGET, ?config, "starting PVF validation host");
 
 	// Run checks for supported security features once per host startup. Warn here if not enabled.
-	let security_status = security::check_security_status(&config).await;
+	let security_status = security::check_security_status(&config).await?;
```
eskimor, line 214: Wait a second. This looks worse, with regards to the issue. If we can't check, we can now no longer do any validation, despite being not yet strict in enforcing security? Am I missing something?

mrcnski: Right now the only possible error should never happen (https://github.com/paritytech/polkadot-sdk/blob/2e2a75ff81/polkadot/node/core/pvf/src/worker_intf.rs#L189-L196). But I could have done this better. Thanks!

eskimor (approved):

mrcnski:
@s0me0ne-unkn0wn raised https://github.com/paritytech/polkadot-sdk/issues/2321 as a follow-up.

PR #2317, opened 2023-11-14
Unify `ChainSync` actions under one enum (follow-up)

dmitry-markin:
Get rid of public `ChainSync::..._requests()`  functions and return all requests as actions.


substrate/client/network/sync/src/service/mod.rs
```diff
@@ -18,6 +18,6 @@
 
 //! `ChainSync`-related service code
 
-pub mod chain_sync;
+pub mod engine;
```
altonen, line 21: Why is `SyncingEngine` moved to service? I don't think this makes sense. I imagined `service` to be for files that pertain to services provide by syncing, such as `NetworkServiceProvider` and `SyncingService`. If there's now tighter coupling between `SyncingService` and `SyncingEngine` and you really want to move files, I think it'd make more sense to have `engine` directory but even that's a push IMO. I thought it was fine as it was before.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4326739

substrate/client/network/sync/src/engine.rs
```diff
@@ -741,7 +738,25 @@ where
...
+				);
+			},
+			ChainSyncAction::SendWarpProofRequest { peer_id, request } => {
+				self.send_warp_proof_request(peer_id, request.clone());
+
+				trace!(
+					target: LOG_TARGET,
+					"Processed `ChainSyncAction::SendWarpProofRequest` to {}, request: {:?}.",
+					peer_id,
+					request,
```
altonen, line 758:
```suggestion
					"Processed `ChainSyncAction::SendWarpProofRequest` to {peer_id}, request: {request:?}.",
```

altonen (commented):
LGTM otherwise but I don't think `service` is the right place for `SyncingEngine`, I thought directory root was a fine place for it

substrate/client/network/sync/src/service/mod.rs
dmitry-markin: It's just the `SyncingService` in the `service` directory. `SyncingEngine` itself is left in the directory root.

substrate/client/network/sync/src/engine.rs
dmitry-markin: We don't fit here into 100 character per line. It's the only reason why this is not inlined.

altonen (approved):

lexnv (approved):

skunert (approved):

substrate/client/network/sync/src/service/engine.rs
skunert, line 1: Hmm maybe `syncing_service.rs` would be better?

PR #2332, opened 2023-11-15
Added new checks to add_assignment

BradleyOlson64:
We now check for every requirement of `add_assignment` described in [RFC 5](https://github.com/polkadot-fellows/RFCs/blob/gav-corejam/text/0005-coretime-interface.md). These are:

- Not more than 100 assignments in a schedule
- Assignment parts add up to 57600 (neither overscheduled nor underscheduled)
- Starting blocks must be greater than or equal to current block + 10 when schedules are added
- Assignments within an added set must be unique
- Assignments within an added set must be sorted

### DIFF
--- a/polkadot/runtime/parachains/src/assigner_bulk/mod.rs
+++ b/polkadot/runtime/parachains/src/assigner_bulk/mod.rs
@@ -223,13 +223,22 @@ pub mod pallet {
 	#[pallet::error]
 	pub enum Error<T> {
 		AssignmentsEmpty,
+		TooManyAssignments,
 		/// Assignments together exceeded 57600.
 		OverScheduled,
+		/// Assignments together less than 57600
+		UnderScheduled,
 		/// assign_core is only allowed to append new assignments at the end of already existing
 		/// ones.
 		DisallowedInsert,
 		/// Tried to insert a schedule for the same core and block number as an existing schedule
 		DuplicateInsert,
+		/// Tried to add an unsorted set of assignments 
+		AssignmentsNotSorted,
+		/// Two or more of the same assignment contained in assignment set
+		AssignmentsNotUnique,
+		/// Assignments must be added with at least 10 blocks of lead time
+		InsufficientLeadTime,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4338097

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4338234

polkadot/runtime/parachains/src/assigner_bulk/mod.rs
```diff
@@ -418,18 +429,37 @@ impl<T: Config> Pallet<T> {
 		assignments: Vec<(CoreAssignment, PartsOf57600)>,
 		end_hint: Option<BlockNumberFor<T>>,
 	) -> Result<(), DispatchError> {
-		// There should be at least one assignment
+		// TODO: Add this assert once the calls `request_core_count` and `notify_core_count`
+		// have been established. assert!(core < legacy_core_count + core_count);
```
eskimor, line 433: Thank you! We have to subtract legacy cores from the core number before passing it on to the bulk assignment provider. Fixed. With this, you don't have to think about legacy_core_count here at all.

polkadot/runtime/parachains/src/assigner_bulk/mod.rs
```diff
@@ -418,18 +429,37 @@ impl<T: Config> Pallet<T> {
...
+
+		// Begin should be no less than the Relay-chain block number on arrival of
+		// the message plus 10
+		let now = <frame_system::Pallet<T>>::block_number();
+		ensure!(begin >= now + BlockNumberFor::<T>::from(REQUIRED_LEAD_TIME), Error::<T>::InsufficientLeadTime);
```
eskimor, line 438: I don't think we should fail on this. We should just handle it best-effort. The RFC only states that 10 blocks should suffice for expected behavior. In practice less will do as well. Even past blocks would work, it is just that assignments would no longer be fully served.

In other words this is something the coretime chain should check, so it lives up to its side of expectations. The most important reason for the 10 block limit is certainly that processing of messages might be delayed. Therefore this check might fail, even if the coretime chain did nothing wrong: Sends 10 blocks in advance, we have a queue of messages and only process that message two blocks later -> Check fails, although there is absolutely no reason why this would not be valid.

polkadot/runtime/parachains/src/assigner_bulk/tests.rs
```diff
@@ -230,20 +230,31 @@ fn assign_core_enforces_well_formed_schedule() {
...
+			(CoreAssignment::Idle, PartsOf57600::from(19200u16)),
 		];
 
 		// Attempting to assign_core with bad assignments
 		assert_noop!(
 			BulkAssigner::assign_core(
 				core_idx,
 				BlockNumberFor::<Test>::from(11u32),
-				bad_assignment_count,
+				empty_assignments,
```
eskimor, line 257: Comment above seems outdated.

polkadot/runtime/parachains/src/assigner_bulk/mod.rs
```diff
@@ -223,13 +223,22 @@ pub mod pallet {
...
 		/// ones.
 		DisallowedInsert,
 		/// Tried to insert a schedule for the same core and block number as an existing schedule
 		DuplicateInsert,
+		/// Tried to add an unsorted set of assignments
+		AssignmentsNotSorted,
+		/// Two or more of the same assignment contained in assignment set
+		AssignmentsNotUnique,
+		/// Assignments must be added with at least 10 blocks of lead time
+		InsufficientLeadTime,
```
eskimor, line 241: That error seems redundant now.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4351120

polkadot/runtime/parachains/src/assigner_bulk/mod.rs
BradleyOlson64: Nice catch!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4352087

PR #2338, opened 2023-11-15
[Cumulus] Add `GeneralKey` variant to `AggregateMessageOrigin`

vgeddes:
Adds `GeneralKey` variant to `AggregateMessageOrigin`.

This change is needed for the system BridgeHub runtimes, so that Snowbridge can also make use of the `MessageQueue` pallet. See previous discussion at https://github.com/paritytech/polkadot-sdk/pull/2146#discussion_r1382568956.

We have reworked Snowbridge to utilize the more generic approach introduced by this PR.
This avoids the need for BridgeHub runtimes to define their own `AggregateMessageOrigin` and associated helper types.

 Other parachain teams may also benefit from this.

### DIFF
--- a/cumulus/primitives/core/src/lib.rs
+++ b/cumulus/primitives/core/src/lib.rs
@@ -91,20 +91,39 @@ pub enum AggregateMessageOrigin {
 	///
 	/// This is used by the HRMP queue.
 	Sibling(ParaId),
+	/// The message came from an abstract origin that can be identified by
+	/// a 32-byte ID.
+	///
+	/// Due to its versatility, this can be used by subsystems other than
+	/// the DMP and HRMP queue.
+	///
+	/// If multiple subsystems want to enqueue messages using this variant,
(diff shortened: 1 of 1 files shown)


ggwpez:
>This avoids the need for BridgeHub runtimes to define their own AggregateMessageOrigin and associated helper types.

But that is cleaner? If we only ever need those variants in the bridgehub runtimes, then only the bridge hub runtimes should expose them.  
Otherwise all other parachain like asset-hub will now have the case in their code that a message from `General` comes in, although they never use it.

vgeddes:
Those are good points - I thought the generalized approach would have been the lightest touch approach. Though as you say, its not very type safe or introspectable.

I appreciate the time you've spent reviewing this and previous drafts. Thanks Oliver.

Closing this off. Instead, will submit a PR providing the specialized `AggregateMessageOrigin` together with all other Snowbridge stuff we intend to put into the BridgeHub runtimes.

Issue #2319, opened 2023-11-14
MessageQueue pallet does not process message


claravanstaden:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

When using the MessageQueue pallet to enqueue a message, and the `process_message` implementation of the enqueued message enqueues another message of a different origin, the other message is never processed. 

To illustrate:
- [enqueue message](https://github.com/paritytech/polkadot-sdk/blob/39cc95740aa03f53a0e8bfb8537f0ddb5fa3e19e/substrate/frame/message-queue/src/lib.rs#L1476) (origin: `Here` message: `first message`)
- when ☝🏻 is processed using trait [ProcessMessage](https://github.com/paritytech/polkadot-sdk/blob/39cc95740aa03f53a0e8bfb8537f0ddb5fa3e19e/substrate/frame/support/src/traits/messages.rs#L52) (method `processed_message`), enqueue another message (origin: `There` message: `another message`)
- `another message` never gets processed

This happens specifically when the other message is added to the `neighbour.prev` pointer in the message queue `ready_ring_unknit` method (when the first message is linked to the other message). 

### Steps to reproduce

Here are tests that replicate the bug:

https://github.com/paritytech/polkadot-sdk/compare/master...claravanstaden:master

acatangiu:
cc @ggwpez

vgeddes:
Some further background information - Clara discovered the bug while updating [Snowbridge](https://github.com/Snowfork/snowbridge) to use a bleeding-edge version of the SDK. This version includes https://github.com/paritytech/polkadot-sdk/pull/1246, which changes the XCMP subsystem to use `MessageQueue` as a message buffer.

This results in the behavior Clara described in general terms. In our case specifically, when `MessageQueue` services a message with a `Sibling(ParaId)` origin, and this message contains an `ExportMessage(dest=Ethereum)` XCM instruction, then our handler for `ExportMessage` ends up enqueuing another message with a different origin.

gavofyork:
would be good if you could add a test to demonstrate the issue. @vgeddes @claravanstaden

claravanstaden:
> would be good if you could add a test to demonstrate the issue. @vgeddes @claravanstaden

@gavofyork I added a test over here: https://github.com/paritytech/polkadot-sdk/compare/master...claravanstaden:master

ggwpez:
[...]

PR #2356, opened 2023-11-15
[FRAME] Make MQ pallet re-entrancy safe

ggwpez:
Closes https://github.com/paritytech/polkadot-sdk/issues/2319

Changes:
- Ensure that only `enqueue_message(s)` is callable from within the message processor. This prevents messed up storage that can currently happen when the pallet is called into recursively.
- Use `H256` instead of `[u8; 32]` for clearer API.

## Details

The re-entracy check is done with the `environmental` crate by adding a `with_service_mutex(f)` function that runs the closure exclusively. This works since the MQ pallet is not instantiable.


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4351938

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4351936

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4352220

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4352351

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4352488

substrate/frame/message-queue/src/lib.rs
```diff
@@ -460,6 +477,10 @@ pub mod pallet {
 
 		/// Processor for a message.
 		///
+		/// Storage changes are not rolled back on error.
```
ggwpez, line 480: Gav mentioned that maybe @kianenigma or @shawntabrizi could help to answer this, but we wondered whether or not this message processor should roll-back storage changes on error.  

The MQ pallet should be generic enough to do any kind of work. On one hand that would mean to have an explicit `TransactionalProcessor` wrapper type to not restrict the user, but on the other hand throughout FRAME we often assume this error rollback mechanism. WDYT?

substrate/frame/support/src/traits/messages.rs
```diff
@@ -119,8 +121,10 @@ impl<OverweightAddr> ServiceQueues for NoopServiceQueues<OverweightAddr> {
 /// The resource footprint of a queue.
 #[derive(Default, Copy, Clone, Eq, PartialEq, RuntimeDebug)]
 pub struct QueueFootprint {
-	/// The number of pages in the queue (including overweight pages).
+	/// The number of pages in the queue (excluding overweight pages).
```
ggwpez, line 124:
```suggestion
	/// The number of pages in the queue (including overweight pages).
```

substrate/frame/message-queue/src/lib.rs
shawntabrizi: I do not know when that specific line was written, but it seems right that everything in the runtime assumes storage rollback when it errors.

I believe it is impossible to create a system where you have both kinds of logic, since you need to make some base level assumption about what will happen when writing code. Especially where that code can co-mingle with code where there may be different assumptions.

substrate/frame/message-queue/src/integration_test.rs
```diff
@@ -147,6 +147,74 @@ fn stress_test_enqueue_and_service() {
 	});
 }
 
+/// Very similar to `stress_test_enqueue_and_service`, but enqueues messages while processing them.
+#[test]
+#[ignore] // Only run in the CI.
```
liamaharon, line 152: Why?

substrate/frame/message-queue/src/lib.rs
```diff
@@ -516,18 +537,51 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Message discarded due to an error in the `MessageProcessor` (usually a format error).
-		ProcessingFailed { id: [u8; 32], origin: MessageOriginOf<T>, error: ProcessMessageError },
+		ProcessingFailed {
+			/// The `blake2_256` hash of the message.
+			id: [u8; 32],
```
liamaharon, line 542: Is it better to use the `H256` type here?

substrate/frame/message-queue/src/lib.rs
```diff
@@ -516,18 +537,51 @@ pub mod pallet {
...
+			///
+			/// This error is pretty opaque. More fine-grained errors need to be emitted as events
+			/// by the `MessageProcessor`.
+			error: ProcessMessageError,
+		},
 		/// Message is processed.
-		Processed { id: [u8; 32], origin: MessageOriginOf<T>, weight_used: Weight, success: bool },
+		Processed {
+			/// The `blake2_256` hash of the message.
+			id: [u8; 32],
```
liamaharon, line 554: ditto

substrate/frame/message-queue/src/integration_test.rs
ggwpez: Its too slow 😆

substrate/frame/message-queue/src/lib.rs
ggwpez: I guess we could change it, but its a breaking change and so far this MR has not been...

substrate/frame/message-queue/src/lib.rs
ggwpez: Yea... I will make a follow-up issue to change that.

substrate/frame/message-queue/src/tests.rs
```diff
@@ -1662,3 +1662,174 @@ fn integrity_test_checks_service_weight() {
...
+		use MessageOrigin::*;
+		MessageQueue::enqueue_message(msg("callback=0"), Here);
+
+		// while servicing queue Here, "anothermessage" of origin There is enqueued in
+		// "firstmessage"'s process_message
+		assert_eq!(MessageQueue::service_queues(1.into_weight()), 1.into_weight());
+		assert_eq!(MessagesProcessed::take(), vec![(b"callback=0".to_vec(), Here)]);
+
+		assert_eq!(MessageQueue::service_queues(1.into_weight()), 1.into_weight());
+		// expects anothermessage to be processed, but it isn't procssed at all
```
franciscoaguirre, line 1683: But it is processed, right?

franciscoaguirre (approved):

gupnik (approved):

liamaharon (approved):

substrate/frame/message-queue/src/tests.rs
ggwpez: Yea. It is the regression test from the original issue; i just copy&pasted it. Now since it is fixed it does not fail anymore and the comment is wrong. Will fix 👍

xlc:
this is not the first time we had re-entrancy bug. I still don't know if we have an official best practice guide but if we do, it should have a section about re-entrancy and the pattern to workaround it. i.e. explain `environmental` and `with_service_mutex`

PR #2359, opened 2023-11-16
New publish instruction

franciscoaguirre:
This instruction is meant to be used for the relay chain pub sub mechanism.
Parachains can publish some key value pair to the relay chain by sending a message with this instruction via UMP.
Afterwards, other parachains can subscribe to the value changes.

### DIFF
--- a/polkadot/xcm/src/v4/mod.rs
+++ b/polkadot/xcm/src/v4/mod.rs
@@ -1016,6 +1016,9 @@ pub enum Instruction<Call> {
 	///
 	/// Errors: If the given origin is `Some` and not equal to the current Origin register.
 	UnpaidExecution { weight_limit: WeightLimit, check_origin: Option<Location> },
+
+	/// A directive to indicate some arbitrary key value pair wants to be published somewhere.
+	Publish { data: Vec<Vec<u8>, Vec<u8>> },
 }
 
 impl<Call> Xcm<Call> {


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4356967

polkadot/xcm/src/v4/mod.rs
```diff
@@ -1016,6 +1016,9 @@ pub enum Instruction<Call> {
 	///
 	/// Errors: If the given origin is `Some` and not equal to the current Origin register.
 	UnpaidExecution { weight_limit: WeightLimit, check_origin: Option<Location> },
+
+	/// A directive to indicate some arbitrary key value pair wants to be published somewhere.
+	Publish { data: Vec<Vec<u8>, Vec<u8>> },
```
ggwpez, line 1021: I dont know how XCM normally handles this or if it makes sense, but maybe we can use a `BoundedVec` to avoid the case that the receiver chain cannot decode it?

franciscoaguirre: I opened an RFC for this to discuss all these things: https://github.com/paritytech/xcm-format/pull/48
It's probably safer for it to be a boundedvec

Issue #2365, opened 2023-11-16
Improve crate versioning


bkchr:
We are currently bumping the major/minor version every time we are doing a release. This is a quite "heavy" operation and also means that crates are never compatible with each other even if they didn't changed. Another issue is that we don't have the versions in the master branch and thus, it is quite hard to for example patch a branch to some external repo that is using the crate releases. The proper way to solve this is that we start versioning the `Cargo.toml` files manually in the pull requests. Then on each release we only pick the crates which version changed and release them using this version. To make this job more easier, we need the following improvements to the repository:

- Introduce a CI check that helps in identifying SemVer breaking changes. Luckily there exists already a [github action](https://github.com/obi1kenobi/cargo-semver-checks-action) for this. While the tool is not perfect, it may helps to detect certain kind of SemVer breaking changes. The tool should be configured to always run against the latest release branch.
- We should move all the internal crates to `workspace.dependencies` and then use them from there in all our crates. This should reduce the merge conflicts to a single file when it comes to version changes.
- Introduce another CI check that ensures that we don't bump the any of the versions multiple times in between two releases. For example there are two pull requests that would bump the `patch` version of one crate, the CI job should reject the second bump. This is not that important, but it would be nice to prevent these double bumps most of the time if possible.

bkchr:
@ggwpez this is probably now a duplicate of one of the issues you created? If yes, please close.

PR #2366, opened 2023-11-16
Write some docs about the release process

bkchr:
This is only a first step and not the end. Later it is probably required that we come up with a more detailed document. However, this should be a good start for an overview.

The text is explaining the "goal" and not what we currently have. The things about the required CI tooling changes and some other small improvements are tracked here: https://github.com/paritytech/polkadot-sdk/issues/2365

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -34,7 +34,7 @@ No PR should be merged until all reviews' comments are addressed. It is expected
 there is no PR merged that is not ready. It is only allowed to merge such a PR if 
 the code is disabled, not used or in any other way clearly tagged as experimental. 
 It is expected that master is always releasable and that all PRs got tested well 
-enough that they likelyhood of critical bugs is quite small.
+enough that the likelihood of critical bugs is quite small.
 
 ### Labels
 
@@ -138,11 +138,11 @@ Please label issues with the following labels:
 ## Release Process
 
 We are aiming for a **two week** release process of the `polkadot-sdk` repository. The output of a release
-are the Parity `polkadot` node implementation, the runtimes for the Westend & Rococo and new versions 
+are the Parity `polkadot` node implementation, the runtimes for the Westend & Rococo networks and new versions 
 of the crates. Given the cadence of two weeks, there is no need to halt a release process for any kind 
 of PR (no BRAKES for the release train). There is only to be made an exception for high security issues 
 for code that is already running in production. However, these critical bug fixes may warrant an out of 
-band release any way, but this should be decided based on the severity on a case by case basis.
+band release anyway, but this should be decided based on the severity on a case by case basis.
 
 ### Versioning
 
@@ -193,7 +193,7 @@ version bump doesn't ripple and only requires a `minor`/`patch` bump.
 
 ### Backports
 
-Backports should most of the time not required. We should only backport critical 
+Backports should most of the time not be required. We should only backport critical 
 bug fixes and then release the fixed crates. There should be no need to backport 
 anything from a release branch. Bumping the `NODE_VERSION` and `spec_version` of 
 the test networks should be done before the release process is started on master.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -30,7 +30,11 @@ If a change does not alter any logic (e.g. comments, dependencies, docs), then i
 `A1-insubstantial` and merged faster.
 If it is an urgent fix with no large change to logic, then it may be merged after a non-author
 contributor has reviewed it well and approved the review once CI is complete.
-No PR should be merged until all reviews' comments are addressed.
+No PR should be merged until all reviews' comments are addressed. It is expected that 
+there is no PR merged that is not ready. It is only allowed to merge such a PR if 
+the code is disabled, not used or in any other way clearly tagged as experimental. 
+It is expected that master is always releasable and that all PRs got tested well 
+enough that they likelyhood of critical bugs is quite small.
 
 ### Labels
 
@@ -50,8 +54,6 @@ The set of labels and their description can be found [here](https://paritytech.g
    `T13-documentation`. The docs team will get in touch.
 5. If your PR changes files in these paths:
 
-   `polkadot` : `^runtime/polkadot`
-   `polkadot` : `^runtime/kusama`
    `polkadot` : `^primitives/src/`
    `polkadot` : `^runtime/common`
    `substrate` : `^frame/`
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
 2. `D*`  issue difficulty, suggesting the level of complexity this issue has. AT MOST ONE ALLOWED.
 3. `T*`  Issue topic. MULTIPLE ALLOWED.
 
-## Releases
-
-Declaring formal releases remains the prerogative of the project maintainer(s).
+## Release Process
+
+We are aiming for a **two week** release process of the `polkadot-sdk` repository. The output of a release
```
mrcnski, line 140: Is it worth mentioning/explaining the merge window? I.e. just because a PR got merged before a release, doesn't mean it made it into the release.

docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
...
+bump. If possible, it should be prevented, but we also don't want to carry
+"dead code" with us for too long. As long as there are clear instructions on how 
+to integrate the breaking changes it should be fine to break things.
+
+So, the general approach is that developers are required to bump the versions in 
+all crates that they are changing. If there is a `major` (or `minor` on `<1.0.0`)
+crate version bump, it is important to also bump any crate that depends on this
+bumped version crate and re-export it. However, if the re-export is done in a
+`__private` module (that makes clear that it is internal api) the `major`/`minor`
+version bump doesn't ripple and only requires a `minor`/`patch` bump.
```
mrcnski, line 192: I'm confused by this paragraph. Can you please explain the re-exporting? What is the `__private` module (is this just some example name?) Why do we need to bump a crate version if only the internal API changed, is it just because our tooling requires it?

mrcnski (commented):
Mostly very clear!

docs/CONTRIBUTING.md
```diff
@@ -30,7 +30,11 @@ If a change does not alter any logic (e.g. comments, dependencies, docs), then i
 `A1-insubstantial` and merged faster.
 If it is an urgent fix with no large change to logic, then it may be merged after a non-author
 contributor has reviewed it well and approved the review once CI is complete.
-No PR should be merged until all reviews' comments are addressed.
+No PR should be merged until all reviews' comments are addressed. It is expected that 
+there is no PR merged that is not ready. It is only allowed to merge such a PR if 
+the code is disabled, not used or in any other way clearly tagged as experimental. 
+It is expected that master is always releasable and that all PRs got tested well 
+enough that they likelyhood of critical bugs is quite small.
```
joepetrowski, line 37:
```suggestion
enough that the likelihood of critical bugs is quite small.
```

docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
 2. `D*`  issue difficulty, suggesting the level of complexity this issue has. AT MOST ONE ALLOWED.
 3. `T*`  Issue topic. MULTIPLE ALLOWED.
 
-## Releases
-
-Declaring formal releases remains the prerogative of the project maintainer(s).
+## Release Process
+
+We are aiming for a **two week** release process of the `polkadot-sdk` repository. The output of a release
+are the Parity `polkadot` node implementation, the runtimes for the Westend & Rococo and new versions
```
joepetrowski, line 141: What about `polkadot-parachain`?
```suggestion
are the Parity `polkadot` node implementation, the runtimes for the Westend & Rococo networks and new versions 
```

docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
...
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in 
+`polkadot/node/primitives/src/lib.rs`.
+
+#### Westend & Rococo
+
+For the test networks we only increment the `spec_version`. The spec version is also following
+the node version. So, `10020` is for example the node release `1.2.0`. This versioning has no
+further meaning and is only done to map from an on chain `spec_version` easily to the 
+release in this repository.
```
joepetrowski, line 168: We have already not followed this. Rococo and Westend Relays are on `103000` (with `1.3.0` release) and the system paras are on `1_003_000` (using the Fellowship "style"). Can we just use the Fellowship style for consistency, `M_mmm_ppp`?

docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
...
+So, the general approach is that developers are required to bump the versions in 
+all crates that they are changing. If there is a `major` (or `minor` on `<1.0.0`)
+crate version bump, it is important to also bump any crate that depends on this
+bumped version crate and re-export it. However, if the re-export is done in a
+`__private` module (that makes clear that it is internal api) the `major`/`minor`
+version bump doesn't ripple and only requires a `minor`/`patch` bump.
+
+### Backports
+
+Backports should most of the time not required. We should only backport critical
```
joepetrowski, line 196:
```suggestion
Backports should most of the time not be required. We should only backport critical 
```

docs/CONTRIBUTING.md
bkchr: This is a public re-export for macros. This is a known pattern to make people on the outside aware that they should not use this module. However, from a pure SemVer perspective, this would be a breaking change.

docs/CONTRIBUTING.md
bkchr: I don't care. We can do this.

docs/CONTRIBUTING.md
```diff
@@ -133,9 +135,76 @@ Please label issues with the following labels:
...
+Following SemVer isn't that easy, but there exists [a guide](https://doc.rust-lang.org/cargo/reference/semver.html) 
+in the Rust documentation that explains the small details on when to bump what. The bumping 
+should be done in between the releases in the PRs. However, it isn't required to 
+bump e.g. the `patch` version multiple times in between two releases. There
+exists a CI job that checks that the versions are not bumped multiple times to help 
+the developer. Another CI job is also checking for SemVer breaking changes. It is using
+[`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks). While 
+the tool isn't perfect, it should help to remind the developer of checking the SemVer 
+compatibility of its changes. In general there is not any guarantee to downstream 
+that there isn't a breaking in between and thus a `major` (or `minor` on `<1.0.0`)
```
KiChjang, line 182:
```suggestion
that there isn't a breaking change in between and thus a `major` (or `minor` on `<1.0.0`) 
```

docs/CONTRIBUTING.md
mrcnski: Oh I've never seen that pattern. I grepped and the `__private` looks like it's only used in substrate. But there is nothing substrate-specific about this doc or this "Crates" section, so it should be clarified. Also, I still don't understand this:

> the `major`/`minor` version bump doesn't ripple and only requires a `minor`/`patch` bump.

But, this paragraph should just be rewritten, and ideally should explain the "why". For sure it is very hard to understand, especially for new contributors.

bkchr:
Closing for now. Will come back :P

PR #1189, opened 2023-08-26
[NPoS] Paging reward payouts in order to  scale rewardable nominators 

Ank4n:
helps https://github.com/paritytech/polkadot-sdk/issues/439.
closes https://github.com/paritytech/polkadot-sdk/issues/473.

PR link in the older substrate repository: https://github.com/paritytech/substrate/pull/13498.

# Context
Rewards payout is processed today in a single block and limited to `MaxNominatorRewardedPerValidator`. This number is currently 512 on both Kusama and Polkadot.

This PR tries to scale the nominators payout to an unlimited count in a multi-block fashion. Exposures are stored in pages, with each page capped to a certain number (`MaxExposurePageSize`). Starting out, this number would be the same as `MaxNominatorRewardedPerValidator`, but eventually, this number can be lowered through new runtime upgrades to limit the rewardeable nominators per dispatched call instruction.

The changes in the PR are backward compatible.

## How payouts would work like after this change
Staking exposes two calls, 1) the existing `payout_stakers` and 2) `payout_stakers_by_page`.

### payout_stakers
This remains backward compatible with no signature change. If for a given era a validator has multiple pages, they can call `payout_stakers` multiple times. The pages are executed in an ascending sequence and the runtime takes care of preventing double claims.

### payout_stakers_by_page
Very similar to `payout_stakers` but also accepts an extra param `page_index`. An account can choose to payout rewards only for an explicitly passed `page_index`.

**Lets look at an example scenario**
Given an active validator on Kusama had 1100 nominators, `MaxExposurePageSize` set to 512 for Era e. In order to pay out rewards to all nominators, the caller would need to call `payout_stakers` 3 times.

- `payout_stakers(origin, stash, e)` => will pay the first 512 nominators.
- `payout_stakers(origin, stash, e)` => will pay the second set of 512 nominators.
- `payout_stakers(origin, stash, e)` => will pay the last set of 76 nominators.
...
- `payout_stakers(origin, stash, e)` => calling it the 4th time would return an error `InvalidPage`.

The above calls can also be replaced by `payout_stakers_by_page` and passing a `page_index` explicitly.

## Commission note
Validator commission is paid out in chunks across all the pages where each commission chunk is proportional to the total stake of the current page. This implies higher the total stake of a page, higher will be the commission. If all the pages of a validator's single era are paid out, the sum of commission paid to the validator across all pages should be equal to what the commission would have been if we had a non-paged exposure.

### Migration Note
[...]

PR #2369, opened 2023-11-16
[NPoS] Check if staker is exposed in paged exposure storage entries

Ank4n:
Addresses a bug caused by https://github.com/paritytech/polkadot-sdk/pull/1189.  The changes are still not released yet, so would like to push the fix soon so it can go together with the release of the above PR.

`fast_unstake` checks if a staker is exposed in an era. However, this fn is still returning whether the staker is exposed based on the old storage item. This PR fixes that by looking in both old and new exposure storages.

Also adds some integrity tests for paged exposures.


rossbulat (approved):

gpestana (approved):

gpestana:
Good spot, approved! Maybe we could add a test case in the `fast-unstake` pallet that would have picked the issue?

georgepisaltu (approved):

Issue #669, opened 2023-04-07
Charge core time chains fees for upgrading their runtime code


eskimor:
parachains which have won slot auctions are constrained by time on how often they can upgrade their runtime code. We need to charge parathreads for this, as the total number of parathreads is not limited by anything. This could lead to a constant flow of very large code upgrades.

PR: https://github.com/paritytech/polkadot-sdk/pull/2372

eskimor:
In fact there is no reason to special case on-demand. We should also charge bulk parachains for upgrades. Idea would be to leave legacy auction parachains as is, but charge all other chains a fee for upgrades.

Reasoning: With coretime, even "bulk" parachains are not really limited in magnitude.

bkchr:
Or we could just increase the time between upgrades, but offer a way for paying to shorten this time to the current default timeout between upgrades.

eskimor:
@antonva @Szegoo While I can see us launching on-demand on Kusama without fees for upgrading. The issue with the low deposit is a deal breaker. Those two in combination makes it a no-go to launch without a fix, even for Kusama.

eskimor:
@Szegoo thanks for looking into this. Whenever you have something, you can request a review from @antonva and me.

eskimor:
Partially [handled](https://github.com/paritytech/polkadot-sdk/pull/3020)

bkchr:
https://github.com/paritytech/polkadot-sdk/pull/2372#discussion_r1461873933 more discussion happened here.

PR #2372, opened 2023-11-16
On-demand & Coretime parachain upgrade fees

Szegoo:
## Summary

When a parachain is registered, the parachain manager is responsible for the cost of storing the validation code on-chain. After successful registration in the current system, parachains are allowed to perform code upgrades without being required to pay for the differences in validation code size. This means that someone could theoretically register a parachain with minimal code and then, at a later point, freely register validation code that is significantly larger than the original code.

This is obviously a problem that needs to be fixed, as being imprecise about the deposit requirements for storing data on-chain can lead to issues in the system.

Additionally, this PR introduces a 'code upgrade fee', that is charged whenever an upgrade is scheduled. This is used to discourage parachains from spamming code upgrades.

## Paying for para upgrades

All on-demand and Coretime parachains will be charged for performing validation code upgrades. Upon registration, the billing account responsible for covering the deposit and paying upgrade fees is set to `None`. 
Before scheduling code upgrades the billing account has to be explicitly set by calling the `set_parachain_billing_account_to_self` extrinsic which will allow setting the billing account to either the sovereign account or the manager.

Whenever the billing account of a parachain changes, the new account will have the entire required deposit reserved, and the previous account will have its associated deposit unreserved.

## Rebate

If the new required deposit for storing the validation code changes during a code upgrade, the billing account will receive a refund once the upgrade is successfully performed

## System & Lease holding parachains

Since we want to avoid causing a breaking change, all current lease-holding parachains that are registered will be eligible to continue performing code upgrades without any costs.

Since this PR allows Root to perform upgrades for free, it also means all system chains do not require any code upgrade costs.

## Additional benefits

These additional deposit requirements will enable us to reduce the deposit needed for reserving a ParaId. Without this change, if we significantly lower it, there's a risk of making it cheap to exploit the network by registering a bunch of paras with small validation codes and then upgrading them for free to a much larger size.

Should Close: #669

TODOs:

- [x] Add events
- [x] Improve docs
- [x] Merge `BillingInfo` into `ParaInfo`
- [x] Add test to ensure that the billing account has to be explicitly set for lease holding paras
- [x] Clean up the code
- [x] Generate weights
- [x] Update PR description to explain the setting of the billing account in more detail


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4372581

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4383533

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4384205

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4390624

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4390612

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4391980

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4392257

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4392409

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -642,6 +683,73 @@ impl<T: Config> Pallet<T> {
...
+				WithdrawReasons::FEE,
+				ExistenceRequirement::KeepAlive,
+			)?;
+		}
+
+		if current_deposit > new_deposit {
+			// In case the existing deposit exceeds the required amount due to validation code
+			// reduction, the excess deposit will be returned to the caller.
+
+			// The reason why the deposit is not instantly refunded is because schedule a code
```
eskimor, line 734: Typo.

eskimor (commented):
Looks sensible on a quick fist pass. :rocket:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4406409

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4407955

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4409573

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4412422

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4413259

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4415231

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4417762

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4418170

Szegoo:
@eskimor @antonva  The PR should be ready for a review now :)

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4578641

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4579460

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4579461

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4579459

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -205,6 +229,19 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub type Deposits<T: Config> =
+		StorageMap<_, Twox64Concat, ParaId, DepositInfo<T::AccountId, BalanceOf<T>>>;
+
+	/// Stores all the `ParaId`s of parachains that purchased a slot using the legacy auctions
+	/// model.
+	///
+	/// Each of these parachains is eligible to perform a single code upgrade without upgrade fees.
+	#[pallet::storage]
+	pub type LegacyParas<T> = StorageMap<_, Twox64Concat, ParaId, ()>;
```
eskimor, line 243: I don't think this cuts it. How is allowing one free update helping?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -404,16 +441,45 @@ pub mod pallet {
...
-			Self::ensure_root_para_or_owner(origin, para)?;
-			runtime_parachains::schedule_code_upgrade::<T>(para, new_code, SetGoAhead::No)?;
-			Ok(())
+			Self::ensure_root_para_or_owner(origin.clone(), para)?;
+
+			let fee_payer = if ensure_root(origin.clone()).is_ok() {
+				// Root doesn't pay. This, also means that system parachains do not pay for upgrade
+				// fees.
+				None
+			} else if LegacyParas::<T>::get(para).is_some() {
```
eskimor, line 463: A better way would be to just make [this](https://github.com/paritytech/polkadot-sdk/blob/18257373b313fcfad4f7a2ab2cb02cb8bfc5b8f1/polkadot/runtime/parachains/src/paras/mod.rs#L686) accessible and use that.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -404,16 +441,45 @@ pub mod pallet {
...
+				None
+			} else if let Ok(caller) = ensure_signed(origin) {
+				let para_info = Paras::<T>::get(para).ok_or(Error::<T>::NotRegistered)?;
+				ensure!(!para_info.is_locked(), Error::<T>::ParaLocked);
+				ensure!(para_info.manager == caller, Error::<T>::NotOwner);
+
+				Some(caller)
+			} else {
+				let location: MultiLocation = (Parent, Parachain(para.into())).into();
+				let sovereign_account = T::SovereignAccountOf::convert_location(&location).unwrap();
```
eskimor, line 478: I think we should make it so that in order to do any upgrades you have to manually configure an account paying for fees and providing the deposit(changes). This way, if an parachain has not set this up, updates will simply fail.

Otherwise, teams not reading documentation might be surprised where all their funds went ...

For now, I think the best option is to force users/projects to manually configure the paying/depositing account in order for upgrades to succeed, e.g. we should offer an extrinsic: "set_upgrade_fee_and_deposit_paying_account" and upgrades will only succeed if this extrinsic has been called before and the set account is sufficiently funded (obviously). 

We could make this a one-shot configuration as part of the PVF registration, but I think it has to be explicit. On the other hand (not part of this PR), we should also have crystal clear documentation (or even better tooling), so users can determine what their upgrade will cost.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -404,16 +441,45 @@ pub mod pallet {
...
+
+			let fee_payer = if ensure_root(origin.clone()).is_ok() {
+				// Root doesn't pay. This, also means that system parachains do not pay for upgrade
+				// fees.
+				None
+			} else if LegacyParas::<T>::get(para).is_some() {
+				// Each legacy para is permitted to perform one upgrade for free.
+				//
+				// This is introduced to  avoid causing a breaking change to the system once para
+				// upgrade fees are required.
```
eskimor, line 467: By only allowing one upgrade we are not fulfilling that requirement. What if a legacy para does two updates?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +862,23 @@ impl<T: Config> OnNewHead for Pallet<T> {
 	}
 }
 
+impl<T: Config> OnCodeUpgrade for Pallet<T> {
+	fn on_code_upgrade(id: ParaId) -> Weight {
+		let mut deposit_info =
+			Deposits::<T>::get(id).expect("The depositor must be known for each parachain; qed");
```
eskimor, line 868: That's not a valid proof. Proofs have to be locally reasonable to be worth anything.

eskimor (commented):
Great work! Looks mostly good. Handling of system chains should work as described ... although, do we really want to give all system chains sudo access ever? @joepetrowski ?

I just have a couple of concerns:

- Handling of legacy chains is not sufficient, but it should be easy to fix, by just making the storage accessible that holds all legacy chains.
- I believe we should be more careful with charging accounts. Before you charge someone, you should have some form of contract ensuring that this is ok.
- `expect` should have locally reasonable proofs.

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -579,6 +601,493 @@ fn basic_end_to_end_works() {
...
+		// Should remain unchanged:
+		assert_eq!(
+			Balances::reserved_balance(&account_id(1)),
+			ParaDeposit::get() + (total_bytes_stored * DataDepositPerByte::get())
+		);
+		assert_eq!(Balances::total_balance(&account_id(1)), free_balance);
+		assert_eq!(Balances::total_balance(&sovereign_account), free_balance);
+
+		// However, legacy chains are only allowed to perform a free upgrade once.
+		// Doing a new upgrade will require fees:
```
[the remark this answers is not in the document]
Szegoo: My idea was to allow this free upgrade so that any ongoing upgrades initiated before incurring extra costs would proceed. Initially, my thought process was that, afterward, parachains should be made aware of the change and adjust before proceeding with additional code upgrades. However, I now realize that this requirement for parachains to prepare is too rigorous.

I will implement the solution you described below since it is much more forgiving to parachains and less complex.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4623560

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4624063

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4624968

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -431,6 +483,46 @@ pub mod pallet {
...
+		/// Can be called by Root, the parachain, or the parachain manager if the parachain is
+		/// unlocked.
+		#[pallet::call_index(9)]
+		#[pallet::weight(<T as Config>::WeightInfo::set_parachain_stash())]
+		pub fn set_parachain_stash(
+			origin: OriginFor<T>,
+			para: ParaId,
+			new_stash: T::AccountId,
+		) -> DispatchResult {
+			Self::ensure_root_para_or_owner(origin, para)?;
```
eskimor, line 505: We are going to charge that passed in `new_stash` account. Hence we have to verify that `new_stash` is controlled by the caller. 

We could as a start introduce `set_parachain_billable_stash_to_self` which sets the parachain stash to the origin _iff_ the origin is either the para manager or the parachain itself.

If we wanted more flexibility, we could have instead of `set_parachain_billable_stash_to_self`, two calls:

1. `allow_account_for_parachain_stash(new_stash)` ... where origin needs to be the manager or the parachain itself and merely states, that it would be ok for the passed account `new_stash` to become the stash account.
2. `accept_becoming_billable_stash_for(para_id`). This call will fail if `allow_account_for_parachain_stash` has not been called  for the origin before and the origin is not the manager nor the parachain itself.

`accept_becoming_billable_stash_for` would be identical to `set_parachain_billable_stash_to_self` if the origin is the manager or the parachain itself (no prior `allow_account_for_parachain_stash`) required in that case.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -431,6 +483,46 @@ pub mod pallet {
...
+			// the required amount from the new one.
+
+			<T as Config>::Currency::reserve(&new_stash, info.deposit)?;
+
+			// After reserving the required funds from the new stash we can now safely
+			// refund the old account.
+			<T as Config>::Currency::unreserve(&stash_info.stash, info.deposit);
+
+			stash_info.stash = new_stash;
+			ParaStashes::<T>::insert(para, stash_info);
```
eskimor, line 522: How about using `mutate`?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -614,8 +706,10 @@ impl<T: Config> Pallet<T> {
 			<T as Config>::Currency::unreserve(&who, rebate);
 		};
 		let info = ParaInfo { manager: who.clone(), deposit, locked: None };
+		let stash_info = StashInfo { stash: who.clone(), pending_refund: None };
```
eskimor, line 709: It seems sensible to merge this with `ParaInfo`? And I would suggest making the stash account an `Option` - it has to be set explicitly. We really need to have some form of "contract" ensuring that it is ok to charge this account. An extrinsic, with a descriptive name, that makes it crystal clear that this account will get charged.

polkadot/runtime/common/src/paras_registrar/mod.rs
eskimor: On top of that the stash account should be able to revoke itself, in case it no longer wants to be billed for this parachain.

polkadot/runtime/common/src/paras_registrar/mod.rs
eskimor: The upgrade will then fail, if such an account has not been set before. I wonder whether just returning an error suffices here ... maybe we should trigger an event for better visibility of what went wrong? :thinking:

eskimor (commented):
While the stash account can now be configured, one still does not need to manually commit to getting charged. I believe it is prudent to ensure the to be charged account does consent.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4630246

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4630651

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Seems like I got really carried away with rewriting the code to the point that I missed such an obvious detail 🙈

I think having `set_parachain_billable_stash_to_self` and `force_set_parachain_billing_account` should be sufficient. 
However, If we want to give more flexibility, we should have a replacing functionality instead of revoking since there must always be an account that has the required reserved.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4630669

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -1060,8 +1061,15 @@ impl parachains_slashing::Config for Runtime {
 	type BenchmarkingConfig = parachains_slashing::BenchConfig<200>;
 }
 
+pub type LocationToAccountId<AccountId> = (
```
franciscoaguirre, line 1064: Rococo already has a way of turning locations into accounts, using a different one will definitely cause issues.
You should import it from the `xcm_config` module right next to `lib`.

```rust
use xcm_config::LocationConverter;
```

polkadot/runtime/westend/src/lib.rs
```diff
@@ -1276,16 +1277,25 @@ impl parachains_slashing::Config for Runtime {
 
 parameter_types! {
 	pub const ParaDeposit: Balance = 2000 * CENTS;
+	pub const UpgradeFee: Balance = 50 * CENTS;
 	pub const RegistrarDataDepositPerByte: Balance = deposit(0, 1);
 }
 
+pub type LocationToAccountId<AccountId> = (
```
franciscoaguirre, line 1284: Same comment.

```rust
use xcm_config::LocationConverter;
```

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -211,21 +215,32 @@ impl paras::Config for Test {
...
 }
 
 parameter_types! {
 	pub const ParaDeposit: Balance = 500;
 	pub const DataDepositPerByte: Balance = 1;
+	pub const UpgradeFee: Balance = 2;
+	pub const RelayNetwork: NetworkId = NetworkId::Kusama;
 }
 
+pub type LocationToAccountId = (
```
franciscoaguirre, line 229: We should be using the newest things in mocks if we can.

This is the new location converter that does the same as previous ones but in an easier to standardize and more extensible way:

```rust
pub type LocationToAccountId = HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>;
```

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -579,6 +602,513 @@ fn basic_end_to_end_works() {
...
+
+		// CASE 1: Attempting to set the billing account for the parachain to the parachain
+		// sovereign account with an insufficient balance to cover the required deposit should
+		// result in a failure.
+
+		let location: MultiLocation = (Parent, Parachain(para_id.into())).into();
+		let sovereign_account =
+			<Test as paras_registrar::Config>::SovereignAccountOf::convert_location(&location)
+				.unwrap();
+		let para_origin: runtime_parachains::Origin = u32::from(LOWEST_PUBLIC_ID).into();
```
franciscoaguirre, line 947: What's the difference between using `LOWEST_PUBLIC_ID` and just `para_id` itself?

polkadot/runtime/common/src/integration_tests.rs
```diff
@@ -579,6 +602,513 @@ fn basic_end_to_end_works() {
...
+		// Force a new head to enact the code upgrade.
+		assert_ok!(Paras::force_note_new_head(
+			RuntimeOrigin::root(),
+			para_id,
+			genesis_head.clone()
+		));
+		assert_eq!(Paras::current_code(&para_id), Some(code_1.clone()));
+
+		// The reserved deposit should cover for the size difference of the new validation code.
+		let total_bytes_stored = code_size as u32 + head_size as u32;
```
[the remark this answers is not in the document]
Szegoo: Yeah, the deposit is reserved for the genesis head.

polkadot/runtime/common/src/integration_tests.rs
Szegoo: There is no difference, I will update it to `para_id` to be persistent throughout the tests.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4632847

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4636558

franciscoaguirre (approved):
Looks good to me on the XCM side

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4644176

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -431,6 +492,64 @@ pub mod pallet {
...
+		pub fn set_parachain_billing_account_to_self(
+			origin: OriginFor<T>,
+			para: ParaId,
+		) -> DispatchResult {
+			Self::ensure_root_para_or_owner(origin.clone(), para)?;
+
+			let new_billing_account = if let Ok(caller) = ensure_signed(origin.clone()) {
+				let para_info = Paras::<T>::get(para).ok_or(Error::<T>::NotRegistered)?;
+				ensure!(!para_info.is_locked(), Error::<T>::ParaLocked);
+				ensure!(para_info.manager == caller, Error::<T>::NotOwner);
```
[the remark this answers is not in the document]
Szegoo: I understand your point; however, I am still unsure whether this is necessary, mainly because the `T::DataDepositPerByte` probably isn't something that is going to change frequently. Calculating the associated cost of an upgrade depends on only one parameter: the byte size of the new validation code, and two constant values (`T::DataDepositPerByte` and `T::UpgradeFee`) so it should be pretty hard to miscalculate.

However, a much simpler solution to this problem would be to add a new optional parameter to the `schedule_code_upgrade` extrinsic that specifies the maximum cost of an upgrade that the parachain is willing to pay. Before performing the upgrade, the extrinsic will just check upfront whether the required cost is under the specified maximum.

mrcnski:
@Szegoo Can we send a medium tip to the Polkadot address in your profile?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -43,16 +43,39 @@ use sp_runtime::{
...
 	locked: Option<bool>,
+	/// The billing account of a parachain. This account is responsible for holding the required
+	/// deposit for this registered parachain.
+	///
+	/// This account will be responsible for covering all associated costs related to performing
+	/// parachain validation code upgrades.
+	///
+	/// When a parachain is newly registered, this will be set to the parachain manager. However,
+	/// at a later point, this can be updated by calling the
+	/// `set_parachain_billing_account_to_self` extrinsic.
```
eskimor, line 69: This no longer seems to be true.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -43,16 +43,39 @@ use sp_runtime::{
...
+	///
+	/// None means that the billing account hasn't been explicitly set. In such a case, if the
+	/// parachain intends to schedule a code upgrade and is not a lease-holding parachain, either
```
eskimor, line 72: mildly confusing. Mentioning that "update will fail (if this is not explicit set)" would make it more clear.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -43,16 +43,39 @@ use sp_runtime::{
...
+	/// the parachain manager or the parachain itself has to explicitly set itself as the billing
+	/// account by calling the `set_parachain_billing_account_to_self` extrinsic.
+	billing_account: Option<Account>,
+	/// In case there is a pending refund for the current billing account, this stores information
+	/// about the amount that will be refunded upon a successful code upgrade.
+	pending_refund: Option<Balance>,
```
eskimor, line 78: Maybe being more explicit here, both in name and docs would make sense: "pending_deposit_refund"?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -578,7 +718,13 @@ impl<T: Config> Pallet<T> {
 
 		let deposit = deposit_override.unwrap_or_else(T::ParaDeposit::get);
 		<T as Config>::Currency::reserve(&who, deposit)?;
-		let info = ParaInfo { manager: who.clone(), deposit, locked: None };
+		let info = ParaInfo {
+			manager: who.clone(),
+			deposit,
+			locked: None,
+			billing_account: Some(who.clone()),
```
eskimor, line 725: Shouldn't this be `None`?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -613,7 +759,13 @@ impl<T: Config> Pallet<T> {
 		} else if let Some(rebate) = deposited.checked_sub(&deposit) {
 			<T as Config>::Currency::unreserve(&who, rebate);
 		};
-		let info = ParaInfo { manager: who.clone(), deposit, locked: None };
+		let info = ParaInfo {
+			manager: who.clone(),
+			deposit,
+			locked: None,
+			billing_account: Some(who.clone()),
```
eskimor, line 766: None?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -642,6 +794,79 @@ impl<T: Config> Pallet<T> {
...
+			// If we returned the deposit here, a possible attack scenario would be to register
+			// the validation code of a parachain and then schedule a code upgrade to set the
+			// code to an empty blob. In such a case, the pre-checking process would fail, so
+			// the old code would remain on-chain even though there is no deposit to cover it.
+
+			info.pending_refund = Some(current_deposit.saturating_sub(new_deposit));
+		}
+
+		Paras::<T>::insert(para, info);
+		runtime_parachains::schedule_code_upgrade::<T>(para, new_code, SetGoAhead::No)?;
```
eskimor, line 864: I just realized that none of this works here. The most common way to do a runtime upgrade is by the parachain itself, which calls [this function](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L885) directly. 

Hence even with your PR, parachains could still upgrade for free and also their deposits won't be adjusted. On top of that the check for root origin to determine special rights for system parachains will also not work this way.

Sorry, did not spot this either initially. I just realized this, because I wanted to check how the origin is even set if the upgrade is initiated by the parachain itself .... result: It is not.

polkadot/runtime/common/src/paras_registrar/mod.rs
eskimor: Maybe I am missing something, but setting this to `Some` immediately defeats the whole point of making this an `Option` - no?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -1451,6 +1745,60 @@ mod tests {
...
+			// Legacy lease holding parachains won't have their billing account set. So if a
+			// parachain no longer has a parachain slot, the billing account must be explicitly
+			// set before initiating code upgrades.
+
+			// Downgrade the lease holding parachain to a parathread.
+			assert_ok!(Registrar::make_parathread(para_id));
+			run_to_session(START_SESSION_INDEX + 6);
+			assert!(Registrar::is_parathread(para_id));
+
+			// To simulate this we will have to manually set the billing account to `None`.
```
eskimor, line 1784: Why? How would this happen in the real world?

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: It is true. All the parachains that are registered after this change is part of the runtime will have the billing account set to the para manager.
However, all the current parachains will have this set to `None`.

Szegoo: There is a migration that sets these to `None` for all paras.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: All the current parachains will have the billing account set to `None`.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Agreed, I made this more clear.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Referring to the updated explanation in the code: 
```
/// The billing account of a parachain. This account is responsible for holding the required
/// deposit for this registered parachain.
///
/// This account will be responsible for covering all associated costs related to performing
/// parachain validation code upgrades.
///
/// When a parachain is newly registered, this will be set to the parachain manager. However,
/// at a later point, this can be updated by calling the
/// `set_parachain_billing_account_to_self` extrinsic.
///
/// None indicates that the billing account hasn't been set, which is the case for all legacy
/// parachains. Attempting to schedule a parachain upgrade will fail if the billing account is
/// set to None, except in the case where the parachain is not a lease-holding parachain.
```

Szegoo: Also, I added an explanation about this in the PR description to make this part clear.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4668086

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4668651

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4671842

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4671841

polkadot/runtime/common/src/paras_registrar/migration.rs
```diff
@@ -56,14 +58,24 @@ impl<T: Config, UnlockParaIds: Contains<ParaId>> OnRuntimeUpgrade
 		let new_count = Paras::<T>::iter_values().count() as u32;
 
 		ensure!(old_count == new_count, "Paras count should not change");
-		Ok(())
+
+		Paras::<T>::iter_keys().try_for_each(|para_id| -> Result<(), _> {
+			let info = Paras::<T>::get(para_id).unwrap();
+			ensure!(
+				info.billing_account.is_none(),
+				"The billing account must be set to the para manager"
```
eskimor, line 66: It must be `None`, but it also must be set to the para manager?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -43,16 +43,38 @@ use sp_runtime::{
...
+	/// This account will be responsible for covering all associated costs related to performing
+	/// parachain validation code upgrades.
+	///
+	/// When a parachain is newly registered, this will be set to the parachain manager. However,
+	/// at a later point, this can be updated by calling the
+	/// `set_parachain_billing_account_to_self` extrinsic.
+	///
+	/// None indicates that the billing account hasn't been set, which is the case for all legacy
+	/// parachains. Attempting to schedule a parachain upgrade will fail if the billing account is
+	/// set to None, except in the case where the parachain is not a lease-holding parachain.
```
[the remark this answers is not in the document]
Szegoo: This seems to be a typo from my end.
```suggestion
	/// set to None, except in the case where the parachain is a lease-holding parachain.
```

polkadot/runtime/common/src/paras_registrar/migration.rs
Szegoo: Yeah, this must be `None`, my bad.
```suggestion
				"The billing account must be set to the None"
```

Szegoo:
> For new coretime parachains, we want this to be None by default, until explicitly set. If not set, any upgrade attempt will simply fail, until the "user" agreed via an extrinisic to get charged for it.

The billing account is intentionally set to the parachain manager during registration. In the registration process, the parachain manager is the one whose deposit is reserved, essentially making it the 'billed' account. 
I don't really see the issue with setting the parachain manager as the initial billing account since they are the one who registered the parachain and covered the associated costs. Also, this design makes it convenient to determine who will receive the rebate upon a successful code upgrade.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Ok, I understand now what you are referring to. I will fix this.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4676848

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4676850

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4676849

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +906,138 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+	) -> Result<Weight, Weight> {
+		let Some(head) = paras::Pallet::<T>::para_head(para) else {
+			return Err(T::DbWeight::get().reads(1))
+		};
+
+		let Some(mut info) = Paras::<T>::get(para) else { return Err(T::DbWeight::get().reads(2)) };
+		let lease_holding = Self::is_parachain(para);
+
+		// The billing account doesn't need to be set for lease holding parachains.
+		if info.billing_account.is_none() && !lease_holding {
```
eskimor, line 935: This does not seem to cut it. We don't need a billing account on any free update. I think the code would be easier to reason about, if we removed this check but add a (false, None) match below to handle this.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +906,138 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+
+				// Update the deposit to the new appropriate amount.
+				info.deposit = new_deposit;
+			},
+			_ => {
+				// No upgrade costs are required.
+			},
+		}
+
+		if current_deposit > new_deposit {
```
eskimor, line 989: I think we should also only do refunds for non-free upgrades. (Move this into the match above.)

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +906,138 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+	fn on_code_upgrade(id: ParaId) -> Weight {
+		let maybe_info = Paras::<T>::get(id);
+		if maybe_info.is_none() {
+			return T::DbWeight::get().reads(1)
+		}
+
+		let mut info = maybe_info
+			.expect("Ensured above that the deposit info is stored for the parachain; qed");
+
+		if let Some(rebate) = info.pending_deposit_refund {
```
eskimor, line 1022: Seems like we are only refunding if the billing_account was set anyways, but then why bother setting `pending_deposit_refund` in that case?

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -882,13 +882,31 @@ impl<T: Config> Pallet<T> {
...
-				&config,
-				SetGoAhead::Yes,
-			));
+				new_code.clone(),
+				UpgradeRequirements::EnforceRequirements,
+			) {
+				Ok(consumed_weight) => {
+					weight.saturating_accrue(consumed_weight);
+
+					weight.saturating_accrue(<paras::Pallet<T>>::schedule_code_upgrade(
```
eskimor, line 893: Nit: Would be nice to have this less nested.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +906,138 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+
+		let current_deposit = info.deposit;
+
+		let free_upgrade = requirements == UpgradeRequirements::SkipRequirements ||
+			para.is_system() ||
+			lease_holding;
+
+		match (free_upgrade, info.billing_account.clone()) {
+			(false, Some(billing_account)) => {
+				if let Err(_) = <T as Config>::Currency::withdraw(
```
eskimor, line 955: Nit: Would be nice to have the business logic less deeply nested.

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -882,13 +882,31 @@ impl<T: Config> Pallet<T> {
...
+						receipt.descriptor.para_id,
```
eskimor, line 894: On a different note: Would be good to check (if you haven't done so), whether just not doing the upgrade is sound. We can definitely not do it, as this also happens if pre-checking fails, but do we need to report anything back to the parachain? I would guess not, but I am not 100% sure.

eskimor (commented):
Nice! Almost good to merge!

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: I rearranged the code a bit, should be much easier to follow now.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: I don't think we should be keeping a deposit reserved that no longer serves a purpose.

Szegoo: That won't work. If the upgrade is initiated by Root, the deposit that is no longer required should be unreserved and returned to the billing account.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: https://github.com/paritytech/polkadot-sdk/pull/2372#discussion_r1426235943

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Should be better now :)

polkadot/runtime/parachains/src/inclusion/mod.rs
Szegoo: > whether just not doing the upgrade is sound.

I am a bit confused, why wouldn't we do the upgrade if it passed`pre_code_upgrade` successfully?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -946,44 +937,49 @@ impl<T: Config> PreCodeUpgrade for Pallet<T> {
...
-						CodeUpgradeScheduleError::FailedToReserveDeposit,
-					));
-					// This is an overestimate of the used weight, but it's better to be safe than
-					// sorry.
-					return Err(<T as Config>::WeightInfo::pre_code_upgrade())
-				}
+			let billing_account = info
+				.billing_account
+				.clone()
+				.expect("Ensured above that the billing account is set; qed");
```
eskimor, line 957: The previous pattern match was good. Checking with `is_none` and then expecting is unnecessarily brittle.

eskimor: E.g. 

```rust
let Some(billing_account) = info.billing_account else {
  Self::deposit_event(Event::<T>::CodeUpgradeScheduleFailed(
	CodeUpgradeScheduleError::BillingAccountNotSet,
	
  ));
  // An overestimate of the used weight, but it's better to be safe than sorry.
  return Err(<T as Config>::WeightInfo::pre_code_upgrade())
}

```

polkadot/runtime/parachains/src/inclusion/mod.rs
eskimor: Because it failed the pre-checking phase by validators. (They had problems compiling the artifact)

Szegoo: I am still uncertain about what you are referring to here. The line you pointed to above schedules a code upgrade. From here we cannot determine whether the outcome of the scheduled upgrade will be successful or not(i.e. whether it passes pre-checking or not).

Szegoo:
@eskimor I do not have the perms to do so, but the weights have to be regenerated before merging.

eskimor (approved):

eskimor:
Great work @Szegoo !

eskimor:
Ok. Someone still needs to write PRdocs.

eskimor:
bot bench --help

eskimor:
bot help

eskimor:
bot bench-all polkadot

Szegoo:
> Ok. Someone still needs to write PRdocs.

Added the PRdocs.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4799334

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4799749

Szegoo:
@eskimor The zombienet para upgrade tests are failing because zombienet initializes a para using the `parasSudoWrapper` pallet instead of the registrar: [zombienet code](https://github.com/paritytech/zombienet/blob/9a0fb44fc1e398ca1a29209cbc9fce535cf4af70/javascript/packages/orchestrator/src/jsapi-helpers/index.ts#L56). Because of this the billing account cannot be set for the parachain since from the perspective of the registrar the para is not registered.

eskimor:
> @eskimor The zombienet para upgrade tests are failing because zombienet initializes a para using the `parasSudoWrapper` pallet instead of the registrar: [zombienet code](https://github.com/paritytech/zombienet/blob/9a0fb44fc1e398ca1a29209cbc9fce535cf4af70/javascript/packages/orchestrator/src/jsapi-helpers/index.ts#L56). Because of this the billing account cannot be set for the parachain since from the perspective of the registrar the para is not registered.

Any ideas how to solve this?

Szegoo:
> Any ideas how to solve this?

Long-term, we should have Zombienet register parachains the way they are in production (using the registrar). 
However, for the purposes of these tests, we can simply use another relay chain runtime that doesn't require upgrade fees (e.g., Versi).

Szegoo:
@eskimor Given that Westend doesn't have any parachains(other than system chains), I think we are fine with not having upgrade fees there. This means that we can use Westend in these Zombienet tests to resolve the issue.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4810123

pepoviola:
> > Any ideas how to solve this?
> 
> Long-term, we should have Zombienet register parachains the way they are in production (using the registrar). However, for the purposes of these tests, we can simply use another relay chain runtime that doesn't require upgrade fees.

Hi @Szegoo, I can add a new `helper` that use the `registrar` with a new syntax in the `DSL` to use one or another. wdyt?
Thx!

eskimor:
> Hi @Szegoo, I can add a new helper that use the registrar with a new syntax in the DSL to use one or another. wdyt?
Thx!

Either that, or change the existing one or you could make the registrar just pass any upgrade fee free, if it has no data about the chain. "Chain has come to live, without registrar - no fees charged" Does that make sense?

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -431,6 +507,76 @@ pub mod pallet {
...
+
+				caller
+			} else if ensure_root(origin).is_ok() {
+				// Root should use `force_set_parachain_billing_account` since it doesn't make sense
+				// for the root to set itself as the billing account.
+				return Ok(())
+			} else {
+				let location: MultiLocation = (Parent, Parachain(para.into())).into();
+				let sovereign_account = T::SovereignAccountOf::convert_location(&location).unwrap();
+				sovereign_account
```
bkchr, line 545:
```suggestion
				para.into_account_truncating()
```

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -142,6 +187,17 @@ pub mod pallet {
...
+		/// The fee required to perform a validation code upgrade for a parachain.
+		///
+		/// This is used to discourage spamming parachain upgrades.
+		#[pallet::constant]
+		type UpgradeFee: Get<BalanceOf<Self>>;
+
+		/// Type used to get the sovereign account of a parachain.  
+		///
+		/// This is used to enable reserving or refunding deposit from parachains.
+		type SovereignAccountOf: ConvertLocation<Self::AccountId>;
```
bkchr, line 199:
```suggestion
```

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -431,6 +507,76 @@ pub mod pallet {
...
+		///
+		/// ## Deposits/Fees
+		/// For this call to be successful, the caller account must have a sufficient balance
+		/// to cover the current deposit required for this parachain.
+		///
+		/// ## Events
+		/// The `BillingAccountSet` event is emitted in case of success.
+		#[pallet::call_index(9)]
+		#[pallet::weight(<T as Config>::WeightInfo::set_parachain_billing_account_to_self())]
+		pub fn set_parachain_billing_account_to_self(
```
bkchr, line 526: Maybe being able to "whitelist" some kind account by the para for being fees would also be nice, but not really a requirement for now.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -675,6 +855,46 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Sets the billing account of a parachain.
+	///
+	/// The newly set account will be responsible for covering all associated costs related to
+	/// performing parachain validation code upgrades.
+	fn set_parachain_billing_account(
+		para: ParaId,
+		new_billing_account: T::AccountId,
+	) -> DispatchResult {
+		let mut info = Paras::<T>::get(para).map_or(Err(Error::<T>::NotRegistered), Ok)?;
```
bkchr, line 867:
```suggestion
		let mut info = Paras::<T>::get(para).ok_or_else(|| Error::<T>::NotRegistered)?;
```

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -675,6 +855,46 @@ impl<T: Config> Pallet<T> {
...
+
+		<T as Config>::Currency::reserve(&new_billing_account, info.deposit)?;
+
+		// After reserving the required funds from the new billing account we can now safely
+		// refund the current deposit holder.
+		let current_deposit_holder = info.billing_account.clone().unwrap_or(info.manager.clone());
+		<T as Config>::Currency::unreserve(&current_deposit_holder, info.deposit);
+
+		info.billing_account = Some(new_billing_account.clone());
+		Paras::<T>::insert(para, info.clone());
```
bkchr, line 881:
```suggestion
		Paras::<T>::insert(para, info);
```

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -506,6 +512,51 @@ impl OnNewHead for Tuple {
 	}
 }
 
+pub trait OnCodeUpgrade {
+	/// A function to execute some custom logic once the pre-checking is successfully completed.
+	///
+	/// This is currently used by the registrar pallet to perform refunds upon validation code
+	/// size reduction.
+	fn on_code_upgrade(id: ParaId) -> Weight;
```
bkchr, line 520:
```suggestion
/// A callback for when a parachain successfully upgraded its code.
pub trait OnCodeUpgraded {
	/// A function to execute some custom logic once the pre-checking is successfully completed.
	fn on_code_upgraded(id: ParaId) -> Weight;
```

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -506,6 +512,51 @@ impl OnNewHead for Tuple {
...
+	fn on_code_upgrade(_id: ParaId) -> Weight {
+		Weight::zero()
+	}
+}
+
+pub trait PreCodeUpgrade {
+	/// A function that performs custom logic to before sceduling a code upgrade.
+	///
+	/// This is currently utilized by the registrar pallet to ensure that the necessary validation
+	/// code upgrade costs are covered.
```
bkchr, line 535:
```suggestion
```

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -506,6 +512,51 @@ impl OnNewHead for Tuple {
...
+	///
+	/// `requirements` signals whether to enforce the pre code upgrade requirements.
+	///
+	/// As a result, it indicates either the success or failure of executing the pre code upgrade
+	/// scheduling logic. In both cases, it returns the consumed weight.
+	fn pre_code_upgrade(
+		id: ParaId,
+		new_code: ValidationCode,
+		requirements: UpgradeRequirements,
```
bkchr, line 544:
```suggestion
	fn can_pay_for_code_upgrade(
		id: ParaId,
		new_code: &ValidationCode,
```

Making it a little bit more explicit and then we can remove the `requirements` stuff. If we don't care about the payment (when done by root), we can just skip the check.

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -618,6 +669,13 @@ pub mod pallet {
 		/// Runtime hook for when a parachain head is updated.
 		type OnNewHead: OnNewHead;
 
+		/// A type that performs custom logic to determine whether a code upgrade is allowed to be
+		/// performed.
+		type PreCodeUpgrade: PreCodeUpgrade;
+
+		/// Type that executes some custom logic upon a successful code upgrade.
+		type OnCodeUpgrade: OnCodeUpgrade;
```
bkchr, line 677: Maybe we should just merge both and also the traits.

The trait could maybe just be `CodeUpgradeHandler`. 

(the comment is not a strict requirement)

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1124,6 +1119,41 @@ impl<T: Config> Pallet<T> {
...
+		) {
+			Ok(consumed_weight) =>
+				consumed_weight.saturating_add(<paras::Pallet<T>>::schedule_code_upgrade(
+					para_id,
+					new_code,
+					now,
+					&config,
+					SetGoAhead::Yes,
+				)),
+			Err(consumed_weight) => {
```
bkchr, line 1147: We need to call the following code in `paras`:
```
	UpgradeGoAheadSignal::<T>::insert(&para_id, UpgradeGoAhead::Abort);
```

There should also be some test for this.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +912,125 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+				// An overestimate of the used weight, but it's better to be safe than sorry.
+				return Err(<T as Config>::WeightInfo::pre_code_upgrade())
+			};
+
+			if let Err(_) = <T as Config>::Currency::withdraw(
+				&billing_account,
+				T::UpgradeFee::get(),
+				WithdrawReasons::FEE,
+				ExistenceRequirement::KeepAlive,
+			) {
```
bkchr, line 961:
```suggestion
			if <T as Config>::Currency::withdraw(
				&billing_account,
				T::UpgradeFee::get(),
				WithdrawReasons::FEE,
				ExistenceRequirement::KeepAlive,
			).is_err() {
```

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +912,125 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+				Self::deposit_event(Event::<T>::CodeUpgradeScheduleFailed(
+					CodeUpgradeScheduleError::FailedToPayUpgradeFee,
+				));
+				// An overestimate of the used weight, but it's better to be safe than sorry.
+				return Err(<T as Config>::WeightInfo::pre_code_upgrade())
+			}
+
+			let additional_deposit = new_deposit.saturating_sub(current_deposit);
+			if let Err(_) = <T as Config>::Currency::reserve(&billing_account, additional_deposit) {
```
bkchr, line 970:
```suggestion
			if <T as Config>::Currency::reserve(&billing_account, additional_deposit).is_err() {
```

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +912,125 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+		let Some(head) = paras::Pallet::<T>::para_head(para) else {
+			return Err(T::DbWeight::get().reads(1))
+		};
+
+		let Some(mut info) = Paras::<T>::get(para) else { return Err(T::DbWeight::get().reads(2)) };
+
+		let new_deposit = Self::required_para_deposit(head.0.len(), new_code.0.len());
+		let current_deposit = info.deposit;
+
+		let lease_holding = Self::is_parachain(para);
```
bkchr, line 942: We should leave some comment on why this is correct right now. In the future this will maybe change.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +912,125 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+				return Err(<T as Config>::WeightInfo::pre_code_upgrade())
+			}
+
+			let additional_deposit = new_deposit.saturating_sub(current_deposit);
+			if let Err(_) = <T as Config>::Currency::reserve(&billing_account, additional_deposit) {
+				Self::deposit_event(Event::<T>::CodeUpgradeScheduleFailed(
+					CodeUpgradeScheduleError::FailedToReserveDeposit,
+				));
+				// An overestimate of the used weight, but it's better to be safe than sorry.
+				return Err(<T as Config>::WeightInfo::pre_code_upgrade())
```
bkchr, line 975: Do we on purpose not refund the `withdrawn` fee above? If not, we could run both in a storage transaction and revert both if one failed.

polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -692,6 +912,125 @@ impl<T: Config> OnNewHead for Pallet<T> {
...
+		if !free_upgrade {
+			let Some(billing_account) = info.billing_account.clone() else {
+				Self::deposit_event(Event::<T>::CodeUpgradeScheduleFailed(
+					CodeUpgradeScheduleError::BillingAccountNotSet,
+				));
+				// An overestimate of the used weight, but it's better to be safe than sorry.
+				return Err(<T as Config>::WeightInfo::pre_code_upgrade())
+			};
+
+			if let Err(_) = <T as Config>::Currency::withdraw(
```
bkchr, line 956: Burning the fee here is what we want to do?

bkchr (commented):
Main logic looks good to me. I found one bug where we don't tell the parachain of the failed upgrade. The other things are mainly cosmetic stuff.

One thing I thought about is devex of this. People wanting to try runtime upgrades will be surprised. Maybe we should at least change zombienet to give each parachain sovereign some funds.

polkadot/zombienet_tests/smoke/0002-parachains-upgrade-smoke-test.toml
```diff
@@ -4,7 +4,7 @@ bootnode = true
 
 [relaychain]
 default_image = "{{ZOMBIENET_INTEGRATION_TEST_IMAGE}}"
-chain = "rococo-local"
+chain = "westend-local"
```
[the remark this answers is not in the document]
bkchr: Ahh just have seen that you already have discussed this.

bkchr:
> Long-term, we should have Zombienet register parachains the way they are in production (using the registrar).
> However, for the purposes of these tests, we can simply use another relay chain runtime that doesn't require upgrade fees.

For now you can just write a custom `js` script that is called by `zombienet` that calls `force_register` in `registrar`. In the future we don't need any Zombienet things for this, we should start to clean up this mess with para id and paras being in different pallets etc. In general with coretime the entire parachain/parathread distinction is not correct anymore and needs to be removed.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4958532

polkadot/runtime/parachains/src/paras/mod.rs
Szegoo: But this isn't just performing checks, it is also updating the state regarding deposits of the parachain in the registrar. If we were to skip this when called by root the state wouldn't get properly updated.

bkchr: Hmm. I mean makes sense, but also in general you would just get back with the next runtime upgrade. (the deposit)

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -879,15 +879,10 @@ impl<T: Config> Pallet<T> {
 		// initial weight is config read.
 		let mut weight = T::DbWeight::get().reads_writes(1, 0);
 		if let Some(new_code) = commitments.new_validation_code {
-			// Block number of candidate's inclusion.
-			let now = <frame_system::Pallet<T>>::block_number();
-
-			weight.saturating_add(<paras::Pallet<T>>::schedule_code_upgrade(
+			weight.saturating_accrue(Self::try_schedule_code_upgrade(
```
eskimor, line 882: I just realized that this is too late. We need to drop the candidate in backing already, but in a way that should be safe. Mainly there exists a mechanism for this already: [UpgradeRestrictionSignal](https://github.com/paritytech/polkadot-sdk/blob/69434d9a32af6103d7cbde9798cf594027fb620e/polkadot/runtime/parachains/src/paras/mod.rs#L800). If an upgrade is not allowed for whatever reason this signal should be set. This is read by the parachain and it won't even try (which is mandatory otherwise it would be bricked, because we would then filter the candidate [here](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1306), called [here](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1250) from [verify_backed_candidate](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1194).

Reason being: If we only drop the upgrade in inclusion we did not actually prevent the spam vector, as most of the work has already been done. Most importantly a relay chain block was already filled with megabytes of data, affecting other candidates.

Also we should update the guide as well. In particular I forgot again why we need fees at all:

1. Deposit is large for large runtimes.
2. Frequency of updates is limited by cool down.
3. Block validation is already being paid for.

I think the major reason is that upgrading a runtime is rather costly, not only in terms of block utilization (size), but also because of pre-checking. This essentially breaks sharding (all validators do it). So it makes sense to be a bit cautious here. The more important part of this PR is surely that we adjust the deposit.

polkadot/runtime/parachains/src/inclusion/mod.rs
eskimor: Thanks for catching this!

polkadot/runtime/common/src/paras_registrar/mod.rs
eskimor: This is specifically to handle legacy lease holding chains, once they are gone, this code can be removed.

polkadot/runtime/parachains/src/inclusion/mod.rs
eskimor: In order to not delay this further, let's make this a follow up.

bkchr: > I just realized that this is too late. We need to drop the candidate in backing already, but in a way that should be safe. Mainly there exists a mechanism for this already: [UpgradeRestrictionSignal](https://github.com/paritytech/polkadot-sdk/blob/69434d9a32af6103d7cbde9798cf594027fb620e/polkadot/runtime/parachains/src/paras/mod.rs#L800). If an upgrade is not allowed for whatever reason this signal should be set. This is read by the parachain and it won't even try (which is mandatory otherwise it would be bricked, if we only had the filtering of candidates [here](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1306), called [here](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1250) from [verify_backed_candidate](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1194)).

This is not really true, `UpgradeRestrictionSignal` is for telling the parachain that it is currently not allowed to do an upgrade. However, we currently have no way to signal that we want to do an upgrade to the relay chain without sending the full new wasm. You are right that the spam vector isn't closed with the current implementation and we should fix it. I mean we could maybe abuse `UpgradeRestrictionSignal`, but then we should entirely drop the dynamic reserve based on the size of the wasm blob. We should instead charge a constant fee for this (probably just based on the maximum blob size). Then we could use the `UpgradeRestrictionSignal` to communicate if the billing account has the required `UpgradeFee`. But yeah, still not a super solution. Someone could still announce an upgrade and move the funds in between backing and availability.

The real proper solution is really to rewrite the current signaling to the relay chain to first only announce the size plus hash of the blob. Then we should reserve the amount on the billing account and signal that it is okay to send the code.

polkadot/runtime/common/src/paras_registrar/mod.rs
Szegoo: Made this transactional, should make more sense now.

polkadot/runtime/parachains/src/inclusion/mod.rs
eskimor: Fixed deposit would be indeed a lot simpler, I see two downsides with it:

1. People are not incentivized to save space.
2. If we ever increase the limit those earlier parachains can now increase their size for free.

both are not dramatic and also if we would ever find those to be a problem we can still write code to adjust it on upgrades.

I hate to say this, but maybe we should just adjust the deposit to fixed maximum for now and get back to this later, e.g. with me actually writing a solid design for this first (proper signaling, actual concerns, potential solutions, actually picked solution with reason) and ideally also then also clean up the code while we are at it.

Szegoo: Probably a dumb question, but what if we restricted upgrades to only happen from the registrar pallet through the extrinsic? This way the spam vector wouldn't be possible.

Szegoo: Can't a parachain make an XCM call to the relay chain to schedule a code upgrade?

eskimor: How would that make a difference? Even if we bumped XCM message size limitations to account for that, we would have the exact same situation, just wrapped differently.

Szegoo: Yeah, I had an idea, but realized it wouldn't work.

What if we put the parachain whose upgrade schedule failed into cooldown? @eskimor as you mentioned the block validation is already being paid for, so from the perspective of the rc it shouldn't really matter if the upgrade failed or succeeded, and with the cooldown we prevent spamming.
Also, we can probably remove the `UpgradeFee`.

bkchr: Reminds me of my idea [here](https://github.com/paritytech/polkadot-sdk/issues/669#issuecomment-1763325926) :P

bkchr: But yeah, if we do this. Then we don't need to change any of the signaling logic etc.

Szegoo: Great, wanted to get this confirmed :). I will implement this tomorrow then.

bkchr: For sure we should not use any kind of dynamic fee based on wasm size for the deposit. This fee should be based on the maximum size of the wasm file.

Szegoo: We have the ‘UpgradeFee’ and the deposit. I am not sure I understand what problems would the dynamic deposit cause if we put the para into cooldown regardless whether the scheduling succeeded or failed.

bkchr: I mean the logic in this pr would not be required anymore? 

But we should require the max at registration, the point being that people otherwise register a very small blob and then upgrade to something much bigger.

Szegoo: I was thinking of adding this to this PR. It would actually be a minimal change(inside the inclusion module put the para into cooldown if calling  `try_schedule_code_upgrade` failed).

This way we would have an overall robust system with:
- dynamic deposits to incentivize people to save space. This also overall makes more sense since you only pay for the space you use.
- cooldown to prevent spamming
- `UpgradeFee` is implemented to cover the cost of pre-checking. One might argue that this is not necessary since we have a cooldown, but I believe it makes sense for the following reason: The cooldown only limits the frequency at which upgrades can be scheduled. Without the `UpgradeFee`, it would cost the same to schedule code upgrades at the interval of the cooldown duration as it would to schedule upgrades less frequently. With the `UpgradeFee`, a parachain that schedules upgrades more frequently will incur higher costs.

bkchr: > dynamic deposits to incentivize people to save space. This also overall makes more sense since you only pay for the space you use.

We need to tackle this entire topic any way in a different way. As you remember Gav and Shawn weren't that happy about your state rent model for the validation code. So, we will need to come up with something better there, which will change whatever we would do here any way. I don't see a reason to make this more complicated right now.

> UpgradeFee to cover the cost of pre-checking.

The entire reason why this was is too prevent spamming. The spamming would be prevented by the timeout. So, I don't see any reason to add a second "barrier". On the other side this makes the "user" (aka parachains) experience worse. They will need to have  some billing account and whatever. If possible, I would prevent adding fees for upgrades. We would also need to update tutorials etc. This is quite a lot of work for no real benefit IMO. 

Maybe let's see what @eskimor thinks.

Szegoo: Yeah, I don't think I presented the state rent model well on the call. The focus seemed to be on the removal of the validation code for parachains that did not pay the rent. The main argument against it was that it would 'break' parachains. However, if we consider that to be breaking couldn't we say the same thing for the Coretime model? If a parachain doesn't purchase Coretime it can't produce blocks. I think as long as the `RentDuration` is set to a sensible value like 4 weeks there shouldn't be problems.

Shawn described an alternative solution that would involve the addition of new social constructs rather than introducing a new model. While this could be effective, I believe an automated process might be preferable to avoid the extra social overhead for a problem that could be handled by code automatically.

Overall I think it would be good if the rent model RFC could get some more reviews.

eskimor: > Maybe let's see what @eskimor thinks.

I had similar thoughts: Is the complexity really worth it? 

@szegoo has a very good point about incentivization: If the upgrade costs something, then we incentivize teams to not "needlessly" upgrade, which is very sensible as there is a significant cost for the network. But I like @bkchr idea of having a very long cooldown period (which is enforced also on failed upgrades) and then you could optionally pay to reduce the cooldown. This would get us all the benefits, without risking irritation about having to set up some account and stuff. People would only need to look into it once they are unhappy with the cooldown time.

So the simplest way forward would be:

# Phase 1:

1. Make the deposit always account for the maximum code size.
2. Increase cooldown period .... actually this is also something we would need to communicate ... but likely less irritating than suddenly having to setup an account.
3. Make cooldown to also be enforced on failed updates.

# Phase 2:

Then we can have a second (not time critical) phase:

This phase is then merely an optimization, as opposed to a necessity. Hence we should also write an RFC here (design document), highlighting exactly all the requirements and options considered.


4. Find good solution for deposit size: Having it dependent on actual size is a good thing in general as it just aligns incentives with cost, leading to better utilization of resources.
5. Implement/adjust what we already have in this PR to enable being able to pay for a reduced cooldown.
6. See downsides: For rent based model/reduced deposit fees, upgrade fees might be mandatory.

# Downsides

If we don't charge for upgrades (usually), then this also increases pressure on deposits not getting too low. As deposits then serve two purposes:

1. Limit state used (obviously).
2. Limit number of upgrades that can happen (together with cooldown)

Phase 1: Should be fast to implement ... if not, maybe we can rely on the large deposit for now and move (2) and (3)  to phase 2 as well. Just changing the deposit calculation (making it bigger) should not even require a security audit.

eskimor: For the rent based model. Indeed there was quite some confusing on pruning, I don't think Gav nor Shawn understood what was actually meant here. Pruning was specifically designed to not be able to break anything - at least not anymore than an ending lease would for example.

The real question whether the rent based model is feasible is, whether we are for security mostly relying on the opportunity costs of the provided deposit or the actually locked capital. In a very first analysis of mine, it is unfortunately in part also the amount of locked capital: Basically if you fill up the storage, managing to degrade the service enough, this would have an impact on the price of DOT, so your deposited DOT would decrease in value -> Disincentivizing the attack.

eskimor: For moving forward in launching coretime on Kusama, @Szegoo would you be interested into implementing  phase 1? Phase 1 MVP, would entail only changing the deposit amount to the maximum PVF size.

bkchr: > 3\. Make cooldown to also be enforced on failed updates.

This is already the case, at least I checked this last week. [We set the cooldown before we kick off the pvf check](https://github.com/paritytech/polkadot-sdk/blob/c83ccda4a32c8b54a57accb1a874337ec1e368cf/polkadot/runtime/parachains/src/paras/mod.rs#L1931-L1943).

Szegoo: >  @Szegoo would you be interested into implementing phase 1? 

Yes, I will open a PR for that.

Issue #2285, opened 2023-11-13
BEEFY broken on Rococo after GRANDPA revert


serban300:
![image](https://github.com/paritytech/polkadot-sdk/assets/18124062/33a80399-288d-4eb0-8a82-8dee936cd7d3)

serban300:
This is caused by some issues with `rococo-rpc-node-1`

It emits a lot of logs that go like this

`🥩 ran out of peers to request justif #7829924 from` logs

And also the `beefy_on_demand_justification_peer_refused` is very high lately

![image](https://github.com/paritytech/polkadot-sdk/assets/18124062/155f1da5-7387-41f6-bae8-a0bc114a2c99)

So probably the peers refuse to answer

serban300:
Investigating on Adrian's test node which has the same issue as `rococo-rpc-node-1`

node identity: `12D3KooWHXsxWZxiBkNd1sFKSkyKTCxAPz6uoBQyGgvXZLKBdd3a`

On the node there are a lot of logs like:
```
2023-11-14T10:01:20+02:00	2023-11-14 08:01:20.116 DEBUG tokio-runtime-worker beefy::sync: 🥩 requesting justif #7829924 from peer PeerId("12D3KooWDMqQL2m3ExdVpBXBmxBVF1R2HY9baF15NPVFNtyurpaq")    
2023-11-14T10:01:20+02:00	2023-11-14 08:01:20.116 DEBUG tokio-runtime-worker beefy::sync: 🥩 for on demand justification #7829924, peer PeerId("12D3KooWG8MczUuMUjXm4S1Y7eZeJuWhNZEQ7aHvAAkK2VuYsSYG") error: Refused    
2023-11-14T10:01:20+02:00	2023-11-14 08:01:20.115 DEBUG tokio-runtime-worker beefy: 🥩 vote target - mandatory block: #7829924    
2023-11-14T10:01:20+02:00	2023-11-14 08:01:20.115 DEBUG tokio-runtime-worker beefy::sync: 🥩 requesting justif #7829924 from peer PeerId("12D3KooWG8MczUuMUjXm4S1Y7eZeJuWhNZEQ7aHvAAkK2VuYsSYG")    
[...]

PR #2378, opened 2023-11-17
Beefy: small fixes

serban300:
Related to #2285

- save the state of the BEEFY gadget after processing a finality proof. We need this in order to avoid skipping blocks.
- avoid reprocessing the old state when not necessary

### DIFF
--- a/substrate/client/consensus/beefy/src/worker.rs
+++ b/substrate/client/consensus/beefy/src/worker.rs
@@ -456,6 +456,7 @@ where
 			.filter(|genesis| *genesis == self.persisted_state.pallet_genesis)
 			.ok_or(Error::ConsensusReset)?;
 
+		let mut new_session_added = false;
 		if *header.number() > self.best_grandpa_block() {
 			// update best GRANDPA finalized block we have seen
 			self.persisted_state.set_best_grandpa(header.clone());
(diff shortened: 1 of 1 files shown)


substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -478,6 +478,9 @@ where
 				}
 			}
 
+			crate::aux_schema::write_voter_state(&*self.backend, &self.persisted_state)
+				.map_err(|e| Error::Backend(e.to_string()))?;
+
```
acatangiu, line 483: for efficiency purposes I would move this to `fn init_session_at()`, because that's what we really care to persist - new detected sessions that are mandatory to vote on.

That way we avoid writing full persisted state to aux-db on every GRANDPA finality - usually this new state change leads to voting which already persists data.

TLDR: your code is the fully correct thing to do, but I'm suggesting we can relax a bit the data backup here and only do it for mandatory blocks finality (new sessions) instead of any block finality.

acatangiu (approved):
looks good! thanks for the fix!

substrate/client/consensus/beefy/src/worker.rs
serban300: Makes sense. Added https://github.com/paritytech/polkadot-sdk/pull/2378/commits/51bac2f63e395a79704cc5b1c9e8f6f8a61bb414 . PTAL.

I added a flag for this and only saving when the flag is true. If we save inside `init_session_at()`, in theory we might save multiple times per notification. Even though I don't think it should happen in practice.

I'm also ok with saving inside `init_session_at()`  if you think it's better.

svyatonik (approved):

substrate/client/consensus/beefy/src/worker.rs
acatangiu: > in theory we might save multiple times per notification

no, that should never happen, first block in each new session **has to be** GRANDPA finalized same as it has to be BEEFY finalized (mandatory block) - aka we will never have a single GRANDPA finality notification covering more than 1 session

(later on, we should really add some fn on `PersistedState` like `fn writer() -> Writer(&mut PersistedState)` that allows mutating `PersistedState` but also automatically persists it to AUX DB on `drop()`)

serban300: > (later on, we should really add some fn on `PersistedState` like `fn writer() -> Writer(&mut PersistedState)` that allows mutating `PersistedState` but also automatically persists it to AUX DB on `drop()`)

Agree. Would be really good to have such a mechanism.

PR #2073, opened 2023-10-28
Improve Client CLI help readability

davxy:
Currently the CLI `-h/--help` commad output is almost unreadable as (for some commands) it:
- doesn't provide a short brief of what the command does.
- doesn't separate the options description in smaller paragraphs.
- doesn't use a smart wrap strategy for lines longer than the number of columns in the terminal.

Follow some pics taken with a 100 cols wide term

### Before

20231028-174531-grim

### After

20231028-175041-grim

### Before

20231028-175257-grim

### After

20231028-175155-grim

PR #2381, opened 2023-11-17
Make collator RPC mode non-experimental

skunert:
The `--relay-chain-rpc-urls` CLI flag has been available for a while now. We have collators with this running and parachain teams are also using it. It should be fine now to remove the experimental status.

### DIFF
--- a/cumulus/client/cli/src/lib.rs
+++ b/cumulus/client/cli/src/lib.rs
@@ -296,7 +296,14 @@ pub struct RunCmd {
 	#[arg(long, conflicts_with = "validator")]
 	pub collator: bool,
 
-	/// EXPERIMENTAL: Specify an URL to a relay chain full node to communicate with.
+	/// Creates a node that uses fewer resources by retrieving relay chain data from a remote node.
+	/// The provided URLs should point to RPC endpoints of the relay chain.
+	///
+	/// This node connects to the remote nodes following the order they were specified in. If the
+	/// connection to one fails, it attempts to connect to the next in the list.
+	///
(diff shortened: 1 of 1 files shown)


dmitry-markin (approved):

michalkucharczyk (approved):

cumulus/client/cli/src/lib.rs
```diff
@@ -296,7 +296,14 @@ pub struct RunCmd {
 	#[arg(long, conflicts_with = "validator")]
 	pub collator: bool,
 
-	/// EXPERIMENTAL: Specify an URL to a relay chain full node to communicate with.
+	/// Creates a node that uses fewer resources by retrieving relay chain data from a remote node.
+	/// The provided URLs should point to RPC endpoints of the relay chain.
```
davxy, line 300:
```suggestion
	/// Creates a node that uses fewer resources by retrieving relay chain data from a remote node.
	///
	/// The provided URLs should point to RPC endpoints of the relay chain.
```

Not long ago I've made [this](https://github.com/paritytech/polkadot-sdk/pull/2073) PR where I make the CLI output readable.

With the `-h` option the "short" help is shown.

It is best if this short output is a "one liner".

davxy (approved):

cumulus/client/cli/src/lib.rs
davxy: Maybe the fact that "uses fewer resources" can be part of the description below (i.e. not part of the short one liner).

bkchr (approved):

PR #2383, opened 2023-11-17
Updates to create chain-spec for coretime-rococo-local

0xmovses:
I needed to create a chain-spec .json file for `coretime-rococo-local`, this would enable me to run the new coretime e2e tests using the https://github.com/paritytech/parachains-integration-tests tool.

Some logic was missing in order for the script `create_coretime_rococo_spec.sh` to run. This PR adds the chain spec plus the missing logic. 

Please Note. When running the script it produces the expected json, but also outputs an error `Error: Service(Other("Error parsing spec file: expected value at line 110 column 5"`, though I think this a misnomer. 

If you'd like to reproduce: 
```bash
bash cumulus/scripts/create_coretime_rococo_spec.sh ./target/release/wbuild/coretime-rococo-runtime/coretime_rococo_runtime.compact.compressed.wasm 1004
```


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4377598

PR #2387, opened 2023-11-17
`keyring`: remove `lazy_static` public hash maps

michalkucharczyk:
The `lazy_static` package does not work well in `no-std`: it requires `spin_no_std` feature, which also will propagate into `std` if enabled. This is not what we want.

This PR provides removes public/private key hashmaps and replaces them with simple static byte arrays.

### DIFF
--- a/substrate/primitives/keyring/src/ed25519.rs
+++ b/substrate/primitives/keyring/src/ed25519.rs
@@ -3,28 +3,27 @@
 // Copyright (C) Parity Technologies (UK) Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed under the Apache License, Version 2.0 (the => "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
 // 	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// distributed under the License is distributed on an => "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
 //! Support code for the runtime. A set of test accounts.
 
-use lazy_static::lazy_static;
 pub use sp_core::ed25519;
 use sp_core::{
 	ed25519::{Pair, Public, Signature},
 	ByteArray, Pair as PairT, H256,
 };
 use sp_runtime::AccountId32;
-use std::{collections::HashMap, ops::Deref};
+use std::ops::Deref;
 
 /// Set of test accounts.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, strum::Display, strum::EnumIter)]
@@ -44,6 +43,8 @@ pub enum Keyring {
 	One,
 	Two,
 }
+
+use hex_literal::hex;
 
 impl Keyring {
 	pub fn from_public(who: &Public) -> Option<Keyring> {
@@ -99,6 +100,39 @@ impl Keyring {
 	pub fn to_seed(self) -> String {
 		format!("//{}", self)
 	}
+
+	fn public_bytes(&self) -> &'static [u8; 32] {
+		match *self {
+			Keyring::Alice =>
+				&hex!("88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee"),
+			Keyring::Bob =>
+				&hex!("d17c2d7823ebf260fd138f2d7e27d114c0145d968b5ff5006125f2414fadae69"),
+			Keyring::Charlie =>
+				&hex!("439660b36c6c03afafca027b910b4fecf99801834c62a5e6006f27d978de234f"),
+			Keyring::Dave =>
+				&hex!("5e639b43e0052c47447dac87d6fd2b6ec50bdd4d0f614e4299c665249bbd09d9"),
+			Keyring::Eve =>
+				&hex!("1dfe3e22cc0d45c70779c1095f7489a8ef3cf52d62fbd8c2fa38c9f1723502b5"),
+			Keyring::Ferdie =>
+				&hex!("568cb4a574c6d178feb39c27dfc8b3f789e5f5423e19c71633c748b9acf086b5"),
+			Keyring::AliceStash =>
+				&hex!("451781cd0c5504504f69ceec484cc66e4c22a2b6a9d20fb1a426d91ad074a2a8"),
+			Keyring::BobStash =>
+				&hex!("292684abbb28def63807c5f6e84e9e8689769eb37b1ab130d79dbfbf1b9a0d44"),
+			Keyring::CharlieStash =>
+				&hex!("dd6a6118b6c11c9c9e5a4f34ed3d545e2c74190f90365c60c230fa82e9423bb9"),
+			Keyring::DaveStash =>
+				&hex!("1d0432d75331ab299065bee79cdb1bdc2497c597a3087b4d955c67e3c000c1e2"),
+			Keyring::EveStash =>
+				&hex!("c833bdd2e1a7a18acc1c11f8596e2e697bb9b42d6b6051e474091a1d43a294d7"),
+			Keyring::FerdieStash =>
+				&hex!("199d749dbf4b8135cb1f3c8fd697a390fc0679881a8a110c1d06375b3b62cd09"),
+			Keyring::One =>
+				&hex!("16f97016bbea8f7b45ae6757b49efc1080accc175d8f018f9ba719b60b0815e4"),
+			Keyring::Two =>
+				&hex!("5079bcd20fd97d7d2f752c4607012600b401950260a91821f73e692071c82bf5"),
+		}
+	}
 }
 
 impl From<Keyring> for &'static str {
@@ -128,16 +162,9 @@ impl From<Keyring> for sp_runtime::MultiSigner {
 	}
 }
 
-lazy_static! {
-	static ref PRIVATE_KEYS: HashMap<Keyring, Pair> =
-		Keyring::iter().map(|i| (i, i.pair())).collect();
-	static ref PUBLIC_KEYS: HashMap<Keyring, Public> =
-		PRIVATE_KEYS.iter().map(|(&name, pair)| (name, pair.public())).collect();
-}
-
 impl From<Keyring> for Public {
 	fn from(k: Keyring) -> Self {
-		*(*PUBLIC_KEYS).get(&k).unwrap()
+		Public::from_raw(*k.public_bytes())
 	}
 }
 
@@ -155,38 +182,32 @@ impl From<Keyring> for Pair {
 
 impl From<Keyring> for [u8; 32] {
 	fn from(k: Keyring) -> Self {
-		*(*PUBLIC_KEYS).get(&k).unwrap().as_array_ref()
+		*k.public_bytes()
 	}
 }
 
 impl From<Keyring> for H256 {
 	fn from(k: Keyring) -> Self {
-		(*PUBLIC_KEYS).get(&k).unwrap().as_array_ref().into()
+		k.public_bytes().into()
 	}
 }
 
 impl From<Keyring> for &'static [u8; 32] {
 	fn from(k: Keyring) -> Self {
-		(*PUBLIC_KEYS).get(&k).unwrap().as_array_ref()
+		k.public_bytes()
 	}
 }
 
 impl AsRef<[u8; 32]> for Keyring {
 	fn as_ref(&self) -> &[u8; 32] {
-		(*PUBLIC_KEYS).get(self).unwrap().as_array_ref()
-	}
-}
-
-impl AsRef<Public> for Keyring {
-	fn as_ref(&self) -> &Public {
-		(*PUBLIC_KEYS).get(self).unwrap()
+		self.public_bytes()
 	}
 }
 
 impl Deref for Keyring {
 	type Target = [u8; 32];
 	fn deref(&self) -> &[u8; 32] {
-		(*PUBLIC_KEYS).get(self).unwrap().as_array_ref()
+		self.public_bytes()
 	}
 }
 
@@ -212,5 +233,17 @@ mod tests {
 			b"I am Alice!",
 			&Keyring::Bob.public(),
 		));
+	}
+
+	#[test]
+	fn verify_static_public_keys() {
+		assert!(Keyring::iter().all(|k| { k.pair().public().as_ref() == *k.public_bytes() }));
+		// little helper to print out public keys hex string
+		// use array_bytes::Hex;
+		// Keyring::iter().map(|i| (i, i.pair())).for_each(|(name, pair)| {
+		// 	let public = pair.public();
+		// 	let bytes: &[u8; 32] = public.as_ref();
+		// 	println!("Keyring::{}: {:?}", name, bytes.hex(""));
+		// });
 	}
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17696,6 +17696,8 @@ dependencies = [
 name = "sp-keyring"
 version = "24.0.0"
 dependencies = [
+ "array-bytes 6.1.0",
+ "hex-literal",
  "lazy_static",
  "sp-core",
  "sp-runtime",
--- a/substrate/primitives/keyring/Cargo.toml
+++ b/substrate/primitives/keyring/Cargo.toml
@@ -14,10 +14,14 @@ readme = "README.md"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
+hex-literal = "0.4.1"
 lazy_static = "1.4.0"
 strum = { version = "0.24.1", features = ["derive"], default-features = false }
 sp-core = { path = "../core" }
 sp-runtime = { path = "../runtime" }
+
+[dev-dependencies]
+array-bytes = { version = "6.1" }
 
 [features]
 # This feature adds Bandersnatch crypto primitives.


substrate/primitives/keyring/src/ed25519.rs
```diff
@@ -45,6 +44,8 @@ pub enum Keyring {
 	Two,
 }
 
+use hex_literal::hex;
```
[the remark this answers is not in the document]
michalkucharczyk: > `hex_literal` crate crate has been removed by substrate in favor of using `array_bytes`.
> 
> But I also understand that here you require an `&'static [u8; 32]`.

Yes, exactly this is a reason.

> Here the `public_bytes()` and `public()` are a bit redundant. Maybe we can do this without having to introduce the `public_bytes()` function but:
> 
>     * gate behind `full_crypto` the `pair`, `sign` and `public` (the one which uses `pair()`) functions

So I think the whole point of this `lazy_static(hashmaps)` was to avoid calling `pair()` which is probably at some cost. I don't know the reason/origin of this optimization, but I am in favor of keeping it. Also because we need to provide alternative implementation for deployments with no `full_crypto` available. Keeping single code for keyring public keys for all feature sets makes sense for me.

> In this case you'll need to remove stuff like:
> 
> ```
> impl From<Keyring> for &'static [u8; 32] {
> ```
> 
> Which I'm not really sure is super relevant

Yeah, we can remove `hex_literal`, but we will need to remove some conversions to refs. Maybe they are not needed at all. Actually I had a version with `array_bytes::hex2bytes_unchecked`.

I'll re-iterate.

michalkucharczyk:
We could also go w/o `hex-literal`, with raw version:
<details>
```
      fn public_bytes(&self) -> &'static [u8; 32] {
          match *self {
~             Keyring::Alice => &[
~                 0x88, 0xdc, 0x34, 0x17, 0xd5, 0x05, 0x8e, 0xc4, 0xb4, 0x50, 0x3e, 0x0c, 0x12, 0xea,
~                 0x1a, 0x0a, 0x89, 0xbe, 0x20, 0x0f, 0xe9, 0x89, 0x22, 0x42, 0x3d, 0x43, 0x34, 0x01,
~                 0x4f, 0xa6, 0xb0, 0xee,
~             ],
~             Keyring::Bob => &[
~                 0xd1, 0x7c, 0x2d, 0x78, 0x23, 0xeb, 0xf2, 0x60, 0xfd, 0x13, 0x8f, 0x2d, 0x7e, 0x27,
~                 0xd1, 0x14, 0xc0, 0x14, 0x5d, 0x96, 0x8b, 0x5f, 0xf5, 0x00, 0x61, 0x25, 0xf2, 0x41,
~                 0x4f, 0xad, 0xae, 0x69,
~             ],
~             Keyring::Charlie => &[
~                 0x43, 0x96, 0x60, 0xb3, 0x6c, 0x6c, 0x03, 0xaf, 0xaf, 0xca, 0x02, 0x7b, 0x91, 0x0b,
~                 0x4f, 0xec, 0xf9, 0x98, 0x01, 0x83, 0x4c, 0x62, 0xa5, 0xe6, 0x00, 0x6f, 0x27, 0xd9,
~                 0x78, 0xde, 0x23, 0x4f,
~             ],
```
</details>

But it does not look good.

substrate/primitives/keyring/src/ed25519.rs
```diff
@@ -3,28 +3,27 @@
 // Copyright (C) Parity Technologies (UK) Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed under the Apache License, Version 2.0 (the => "License");
```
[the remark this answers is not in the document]
michalkucharczyk: 599045c

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4385273

paritytech-cicd-pr:
**Job name**: build-test-collators
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4403306

paritytech-cicd-pr:
**Job name**: build-test-collators
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4403306

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4403292

davxy (commented):
The code obviously works. There is not much to say about it.

The only thing that I still don't like (but maybe is subjective) is that there are two obvious ways to do the same thing.

That is, getting a public key using:
- `Keyring::pair().public()`
- `From<Keyring> or [u8; 32]`

In practice this new approach is the same as the one used in the previous iteration only that the code that previously was in a member function has not being moved in `From<Keyring>` for [u8; 32]`.

Again, **IMO** would be way better to have one way or another.

- One which is used for `std` or `full_crypto`  or whatever you want to call the feature in this crate (which also includes the Pair` and capability to sign).
- Another where `fn public` internally uses this new approach of pre-constructed arrays.

If at some point you want to include this stuff in `no_std` you may want to remove `fn pair()` anyway as to sign with a Schnorr key you need randomness.

michalkucharczyk:
The consequence of your proposal will be calling the `pair().public()` everytime (w/o a `PUBLICK_KESY` cache) which was just removed in `full_crypto` and `std`.

Having the public key in form of raw bytes prevents this call (and is close to cached approach).

michalkucharczyk:
> If at some point you want to include this stuff in `no_std` you may want to remove `fn pair()` anyway as to sign with a Schnorr key you need randomness.

I would make `pair()` accessible (`from_string` will work in `no-std`), `sign` will be gated.

michalkucharczyk:
ok, I think I got it, you simply want this function to use predefined raw bytes, right?
https://github.com/paritytech/polkadot-sdk/blob/88ae185a3ec253db98efd2b06afddb73e755815d/substrate/primitives/keyring/src/sr25519.rs#L94-L96

davxy:
> IUC: the consequence of your proposal will be calling the `pair().public()` everytime (w/o a `PUBLICK_KESY` cache which was just removed) in `full_crypto` and `std`.
> 
> Having the public key in form of raw bytes prevents this call (and is closer to cached approach).

Caching is pointless here. Alice, Bob and co keyring is used for tests, templates and a like. Never in production

davxy:
> ok, I think I got it, you simply want this function to use predefined raw bytes, right?

The ideal is:
- if I can compute from pair() => compute from pair
- if I can't => use precomputed

As I said above, caching is not really important for these keys

michalkucharczyk:
> > IUC: the consequence of your proposal will be calling the `pair().public()` everytime (w/o a `PUBLICK_KESY` cache which was just removed) in `full_crypto` and `std`.
> > Having the public key in form of raw bytes prevents this call (and is closer to cached approach).
> 
> Caching is pointless here. Alice, Bob and co keyring is used for tests, templates and a like. Never in production

This will be done in chainspec generation (for all testnets, zombienets, etc...), which will be done in runtime. Why not keep this optimization?

michalkucharczyk:
Is the cost of calling `pair()` actually negligible?

davxy:
> > > IUC: the consequence of your proposal will be calling the `pair().public()` everytime (w/o a `PUBLICK_KESY` cache which was just removed) in `full_crypto` and `std`.
> > > Having the public key in form of raw bytes prevents this call (and is closer to cached approach).
> > 
> > 
> > Caching is pointless here. Alice, Bob and co keyring is used for tests, templates and a like. Never in production
> 
> This will be done in chainspec generation (for all testnets, zombienets, etc...), which will be done in runtime. Why not keep this optimization?

1. I'm not saying that is negligible. But for sure we do a bazilion of hashes and computations around which probably makes it negligible (I haven't done any benchmarks BTW)
2. Haven't you already removed the cache with this PR? (i.e. `PUBLIC_KEYS` and `PRIVATE_KEYS`) 

My suggestion was such that in the runtime you end up using the `fn public()` version which doesn't use `pair()`.  I.e. the one which uses the pre-computed stuff. There is not real value exposing `fn pair()` in the runtime as you can't use `fn sign()` for Schnorr and ECDSA (which in general requires randomness)

michalkucharczyk:
Support for `no-std` (e.g removal of sign method) is added in:
https://github.com/paritytech/polkadot-sdk/pull/2044

michalkucharczyk:
Yeah, sorry my oversight, Obviously `fn public()` method should have been using the raw bytes. Fixed in 25a0a89.

davxy:
> Support for `no-std` (e.g removal of sign method) is added in: #2044

Ok I remember. But I also was thinking that with the introduction of statically computed `Public` having the `Pair` trait for `no_std` was not required. 

What is the usage of `Pair` in `no_std` at this point? Wasn't all that work done because you were requiring to compute `Public`? Maybe offchain workers who doesn't want to use the keystore via HF?

michalkucharczyk:
We can gate `pair()` method also. I will add it in #2044 (as it introduces support of Keyring in runtimes).

Probably there is no use case for this in runtime. On the other hand we will have this functionality for free, and if not user linked should strip it off.

substrate/primitives/keyring/src/lib.rs
```diff
@@ -40,3 +40,99 @@ pub mod test {
 	/// The keyring for use with accounts when using the test runtime.
 	pub use super::ed25519::Keyring as AccountKeyring;
 }
+
+/// Util that generates const array from (static) string literal
+pub(crate) const fn hex2array<const N: usize>(hex: &str) -> [u8; N] {
```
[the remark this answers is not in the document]
michalkucharczyk: Yeah, I know, I will move it to sp_core.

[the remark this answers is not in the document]
michalkucharczyk: 8d94b6b

michalkucharczyk:
bot fmt

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,115 @@
...
+
+	#[test]
+	fn t16() {
+		const T16: [u8; 16] = hex2array("000102030405060708090a0b0c0d0e0f");
+
+		assert_eq!(T16, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+	}
+
+	#[test]
+	fn t33() {
```
davxy, line 103: Some considerations about the tests:
- maybe a single "smoke" test  like t33 is sufficient here. But is up to you...
- this point I'd add a couple of tests to exercise the panic conditions.
- i'd add the `#panic` section in the docs of `hex2array`

michalkucharczyk: done in: bcaacc6

However, I don't know how to write `#[should_panic]` test for const context (checking that we panic during compilation).

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,149 @@
...
+	#[test]
+	#[should_panic]
+	fn t_panic_incorrect_length() {
+		let f = |n: u8| -> Option<[u8; 2]> {
+			match n {
+				0 => Some(hex2array("454")),
+				_ => None,
+			}
+		};
+		assert_eq!(f(1), Some([0, 0]));
```
michalkucharczyk, line 135:
```suggestion
		assert_eq!(f(0), Some([0, 0]));
```

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,149 @@
...
+	#[test]
+	#[should_panic]
+	fn t_panic_incorrect_character() {
+		let f = |n: u8| -> Option<[u8; 2]> {
+			match n {
+				0 => Some(hex2array("45ag")),
+				_ => None,
+			}
+		};
+		assert_eq!(f(1), Some([0, 0]));
```
michalkucharczyk, line 147:
```suggestion
		assert_eq!(f(0), Some([0, 0]));
```

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -15,8 +15,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Provides const function allowing converting hex string to `u8` array at compile time if used in
-//! proper context.
+//! Provides a const function for converting a hex string to a `u8` array at compile time, when used
+//! in the proper context. Valid characters are `[0-9a-fA-F]`, and the hex string should not start
+//! with the `0x` prefix.
+//!
+//! # Panics
```
davxy, line 22: The panics section should be in the function doc

michalkucharczyk: done: ab13181

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -112,4 +122,28 @@ mod testh2b {
 			]
 		);
 	}
+
+	#[test]
+	#[should_panic]
```
davxy, line 129:
```suggestion
	#[should_panic = "hex string length is not valid"]
```
(same below)

Also why not just:

```
	#[test]
	#[should_panic = "hex string length is not valid"]
	fn t_panic_incorrect_length() {
		let _ = hex2array::<2>("454");
	}
```

michalkucharczyk: yeah, thanks, I over-complicated that.

michalkucharczyk: f01fe2f

davxy:
As said. Code is good and works.

But before my 🟢 light I also want to have other opinions about https://github.com/paritytech/polkadot-sdk/pull/2387#pullrequestreview-1746752834

I.e. I still have the opinion of having two versions of `public` is better. One gated by some of `full_crypto` (or other name...) which also exposes the `pair` function. This would use `pair` to construct `public`. And another (fallback) `public` which instead uses the hardcoded keys.

If I'm alone :-) well... I'm not going to refuse this as it is a solution which works.
I just like to hear some other opinions.

michalkucharczyk:
ok, let's wait for more opinions.

michalkucharczyk:
@davxy one more question: what is the reason to have two implementations providing same result?

davxy:
> @davxy one more question: what is the reason to have two implementations providing same result?

That is the whole point of my discussion. There is not reason :-)

With this PR we have two ways of getting the bytes of public:
- 1 via the `Keyring::public()` and then using `AsRef<[u8; 32]>` implementation of Public
-2 via the `impl From<Keyring> for [u8; 32] {`
```

The first and the second are using two different approached in providing the same output.
-1 is using pair().public()
-2 is using the static pre computed keys.

---

My whole point is to not provide this two things for the same outcome. But only one, which is conditionally compiled.
I.e. conditionally compile one version of  `Keyring::public`  which
- if `full_crypto` => use `pair().public()`
- else => use precomputed

**Comment by michalkucharczyk** [comment]
> With this PR we have two ways of getting the bytes of public:
> 
> * 1 via the `Keyring::public()` and then using `AsRef<[u8; 32]>` implementation of Public
>   -2 via the `impl From<Keyring> for [u8; 32] {`
> 
> The first and the second are using two different approaches to provide the same output. -1 is using `pair().public()` -2 is using the static pre computed keys.

I may misunderstood your comment, but it is not true:
https://github.com/paritytech/polkadot-sdk/blob/f01fe2fab56aa06018ca21ccac2dda7079487eee/substrate/primitives/keyring/src/sr25519.rs#L94-L96
https://github.com/paritytech/polkadot-sdk/blob/f01fe2fab56aa06018ca21ccac2dda7079487eee/substrate/primitives/keyring/src/bandersnatch.rs#L74-L76
https://github.com/paritytech/polkadot-sdk/blob/f01fe2fab56aa06018ca21ccac2dda7079487eee/substrate/primitives/keyring/src/ed25519.rs#L94-L96

None of `fn public` is internally using `pair()`.

**Comment by davxy** [comment]
> None of `fn public` is internally using `pair()`.

This is changed since the last time I looked into it (was using `pair()` once). My bad.

Mmm. so `public()` now always uses the hardcoded array.

Ok. May be good then

**Review by davxy — APPROVED** [review]
Sorry for bikeshedding on this

**Comment by michalkucharczyk** [comment]
> Sorry for bikeshedding on this

It was different at the beggining, I spotted my mistake here: https://github.com/paritytech/polkadot-sdk/pull/2387#issuecomment-1824771171

File: substrate/primitives/keyring/src/ed25519.rs [hunk 9]
```diff
@@ -153,40 +145,51 @@ impl From<Keyring> for Pair {
...
-impl From<Keyring> for [u8; 32] {
-	fn from(k: Keyring) -> Self {
-		*(*PUBLIC_KEYS).get(&k).unwrap().as_array_ref()
-	}
+macro_rules! hex2arr {
+	($input:expr) => {{
+		const PUBLIC_BYTES: [u8; 32] = hex2arr($input);
+		PUBLIC_BYTES
+	}};
 }
```
koute, line 153: You could make this macro generic over size:

```suggestion
macro_rules! hex2arr {
	($input:expr) => {{
		const PUBLIC_BYTES: [u8; $input.len()] = hex2arr($input);
		PUBLIC_BYTES
	}};
}
```

...and then you could export it from `const_hex2array.rs`, and hide the `hex2array` function (can't make it non-public because of the macro, but can be made `#[doc(hidden)]`) so that no one accidentally calls it in a non-const context.

michalkucharczyk: done in: 96c3574

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,162 @@
...
+
+//! Provides a const function for converting a hex string to a `u8` array at compile time, when used
+//! in the proper context.
+
+/// Provides a const array from given string literal.
+///
+/// Valid characters are `[0-9a-fA-F]`, and the hex string should not start
+/// with the `0x` prefix.
+#[macro_export]
+macro_rules! hex2arr {
```
koute, line 26: Maybe:

```suggestion
macro_rules! hex2array {
```

...for consistency with the module name/function name?

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,162 @@
...
+/// # Panics
+///
+/// The function will panic at compile time when used in a const context if:
+/// - The given hex string has an invalid length.
+/// - It contains invalid characters.
+///
+/// The function will panic at runtime when used in a non-const context if the above conditions are
+/// met.
+#[doc(hidden)]
+pub const fn hex2array<const N: usize>(hex: &str) -> [u8; N] {
```
koute, line 47:
```suggestion
pub const fn private_hex2array<const N: usize>(hex: &str) -> [u8; N] {
```

... or something like this, so that people don't accidentally call it.

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -18,7 +18,19 @@
...
-/// Util that generates array from (static) string literal.
+/// Provides a const array from given string literal.
+///
+/// Valid characters are `[0-9a-fA-F]`, and the hex string should not start
+/// with the `0x` prefix.
+#[macro_export]
+macro_rules! hex2arr {
+	($input:expr) => {{
+		const PUBLIC_BYTES: [u8; $input.len() >> 1] = $crate::const_hex2array::hex2array($input);
+		PUBLIC_BYTES
```
davxy:
```suggestion
		const BYTES: [u8; $input.len() >> 1] = $crate::const_hex2array::hex2array($input);
		BYTES
```
Nit: I'd rename it to just `BYTES` as this is a generic macro which can be used for not just public keys

michalkucharczyk: done in: 0f7cccc

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,162 @@
...
+//! in the proper context.
+
+/// Provides a const array from given string literal.
+///
+/// Valid characters are `[0-9a-fA-F]`, and the hex string should not start
+/// with the `0x` prefix.
+#[macro_export]
+macro_rules! hex2array {
+	($input:expr) => {{
+		const BYTES: [u8; $input.len() >> 1] = $crate::const_hex2array::private_hex2array($input);
```
koute, line 28: Not that it practically matters, but why a shift instead of a division? (:

substrate/primitives/core/src/const_hex2array.rs
```diff
@@ -0,0 +1,162 @@
...
+///
+/// The function will panic at runtime when used in a non-const context if the above conditions are
+/// met.
+#[doc(hidden)]
+pub const fn private_hex2array<const N: usize>(hex: &str) -> [u8; N] {
+	const fn c2b(c: u8) -> u8 {
+		match c as char {
+			'0'..='9' => c - 48,
+			'a'..='f' => c - 87,
+			'A'..='F' => c - 55,
```
koute, line 52: Slightly more readable version without the magic constants?

```suggestion
			'0'..='9' => c - b'0',
			'a'..='f' => c - (b'a' - 10),
			'A'..='F' => c - (b'A' - 10),
```

substrate/primitives/core/src/const_hex2array.rs
michalkucharczyk, line 28:
```suggestion
		const BYTES: [u8; $input.len() / 2] = $crate::const_hex2array::private_hex2array($input);
```

michalkucharczyk:
bot update-ui

PR #2393, opened 2023-11-18
[FRAME] Use 'ready' pages in XCMP suspend logic

ggwpez:
Changes:
- `QueueFootprint` gets a new field; `ready_pages` that contains the non-overweight and not yet processed pages.
- `XCMP` queue pallet is change to use the `ready_pages` instead of `pages` to calculate the channel suspension thresholds.

This should give the XCMP queue pallet a more correct view of when to suspend channels.


substrate/frame/message-queue/src/mock.rs
```diff
@@ -367,6 +367,6 @@ pub fn num_overweight_enqueued_events() -> u32 {
 		.count() as u32
 }
 
-pub fn fp(pages: u32, count: u64, size: u64) -> QueueFootprint {
-	QueueFootprint { storage: Footprint { count, size }, pages }
+pub fn fp(ready_pages: u32, pages: u32, count: u64, size: u64) -> QueueFootprint {
```
seadanda, line 370:
```suggestion
pub fn fp(pages: u32, ready_pages: u32, count: u64, size: u64) -> QueueFootprint {
```
Small one, but maintaining the order of `pages` before `ready_pages` might be better here for people looking in the tests for examples of how they both change in certain cases. With a swapped order people might naively confuse the two when they just see the `fp` function call.

seadanda (approved):

gpestana (approved):
LGTM, I suggest adding a simple try state assertion that checks if the number of ready pages is larger than the number of pages and any other invariants that could be useful.

franciscoaguirre (approved):

PR #2394, opened 2023-11-18
Add router to split XCM fee and assets

franciscoaguirre:
This router is meant to be used for being able to send more assets before BuyExecution than the limit specified in the AllowTopLevelPaidExecutionFrom barrier. It works by splitting those assets into the fee first and the other assets afterward.

TODO:
- [ ] More tests

### DIFF
--- a/polkadot/xcm/xcm-builder/src/routing.rs
+++ b/polkadot/xcm/xcm-builder/src/routing.rs
@@ -104,3 +104,139 @@ impl<Inner: SendXcm, TopicSource: SourceTopic> SendXcm for WithTopicSource<Inner
 		Ok(unique_id)
 	}
 }
+
+fn split_message(message: &mut Option<Xcm<()>>) {
+    if let Some(xcm) = message {
+        let instructions = xcm.inner_mut();
+        let mut initial_fund = false;
+        let mut clear_origin_instructions = 0;
+
+        for item in instructions.iter().enumerate() {
+            match item {
+                (0, WithdrawAsset(assets, ..)) if assets.len() > 1 => {
+                    initial_fund = true;
+                }
+                (n, ClearOrigin) if n > 0 && n <= 4 => {
+                    clear_origin_instructions += 1;
+                }
+                (n, BuyExecution { fees, .. }) if n > 0 && n <= 5 && initial_fund => {
+                    if let Some(WithdrawAsset(assets)) = instructions.first() {
+                        let fee_asset = MultiAssets::from(
+                            assets
+                                .inner()
+                                .iter()
+                                .filter(|asset| asset.id == fees.id)
+                                .cloned()
+                                .collect::<Vec<_>>(),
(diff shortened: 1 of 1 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4390952

acatangiu (commented):
I find this approach of mutating in-flight XCMs to "fix" original XCMs unnatural and obfuscating.. Long term this will just add even more issues and corner-cases.
It adds (to) problems like fees/weights, debugability, control over exact XCM being sent and its intended effects, and general obfuscation of what is happening under the hood.

Instead of this approach, I am very much in favor of biting the bullet and fixing the original miss of these corner-cases in the pallets building XCMs and in the `xcm-executor`.

With a clear limit of max 1 asset allowed for buying execution, it then becomes clear to XCM builders that XCMs sent to be executed on remote chains they need to do explicit:
```
Xcm(vec![
	ReceiveTeleportedAsset(fee)/ReserveAssetDeposited(fee)/WithdrawAsset(fee)/ClaimAsset(fee),
	BuyExecution(fee),
	<rest of the program here>,
])
```

This also means we need to do breaking changes to `TransferReserveAsset`, `DepositReserveAsset`, `InitiateReserveWithdraw`, `InitiateTeleport` to build fully correct onward XCMs in the first place.
Ideas for this ^ :
 - take extra `Option<fees>` parameter and make the same split your current router does and add `BuyExecution` before adding `ClearOrigin`
 - instead of single `xcm` they take `with_origin_xcm` and `without_origin_xcm` specifying which custom instructions to run before `ClearOrigin` and which to run after `ClearOrigin`

KiChjang:
@acatangiu What you're proposing seems to require an XCM version upgrade. I think I agree with the general approach of solving the issue from the roots, however we'd also want to have a solution while we prepare a long-term fix in the next version.

acatangiu:
> @acatangiu What you're proposing seems to require an XCM version upgrade. I think I agree with the general approach of solving the issue from the roots, however we'd also want to have a solution while we prepare a long-term fix in the next version.

Yes, it'd require a new XCM version.

> however we'd also want to have a solution while we prepare a long-term fix in the next version

I'm not against this, but I'm wondering if current `MAX_ASSETS_FOR_BUY_EXECUTION = 2` is not enough of a "temporary solution" until proper handling of it in a new XCM version.

I do see how some router solution like this would be useful/necessary during transition to the new XCM version (aka the window between when we change `MAX_ASSETS_FOR_BUY_EXECUTION` back to `1` and chains actually upgrade to new version - unless we can also make `MAX_ASSETS_FOR_BUY_EXECUTION` version specific..)

franciscoaguirre:
Closing since `MAX_ASSETS_FOR_BUY_EXECUTION = 2` was a good temporary solution

PR #2406, opened 2023-11-20
Refactor `ValidationError`

eagr:
closes https://github.com/paritytech/polkadot-sdk/issues/733
helps https://github.com/paritytech/polkadot-sdk/issues/660
helps https://github.com/paritytech/polkadot-sdk/issues/661

Summary:

`ValidationError::InternalError` ->
`ValidationError::Internal`

`ValidationError::InvalidCandidate` ->
`ValidationError::Invalid`

`InvalidCandidate::PrepareError` ->
`ValidationError::Preparation`

`ValidationError::PossiblyInvalid(PossiblyInvalidError)`

`InvalidCandidate::AmbiguousWorkerDeath`->
`PossiblyInvalidError::AmbiguousWorkerDeath`

`InvalidCandidate::AmbiguousJobDeath`->
`PossiblyInvalidError::AmbiguousJobDeath`

`InvalidCandidate::JobError`->
`PossiblyInvalidError::JobError`


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4398359

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -640,22 +640,22 @@ async fn validate_candidate_exhaustive(
...
-		Err(ValidationError::InvalidCandidate(WasmInvalidCandidate::AmbiguousJobDeath(err))) =>
+		Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::AmbiguousJobDeath(err))) =>
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(format!(
 				"ambiguous job death: {err}"
 			)))),
-		Err(ValidationError::InvalidCandidate(WasmInvalidCandidate::PrepareError(e))) => {
+		Err(ValidationError::Preparation(e)) => {
 			// In principle if preparation of the `WASM` fails, the current candidate can not be the
 			// reason for that. So we can't say whether it is invalid or not. In addition, with
 			// pre-checking enabled only valid runtimes should ever get enacted, so we can be
```
mrcnski, line 661: Maybe we can move this comment to the `ValidationError::Preparation` variant? It's nice to keep docs like these centralized, it makes them easier to find and it also explains why this is its own variant.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -756,14 +756,14 @@ trait ValidationBackend {
...
 			match validation_result {
-				Err(ValidationError::InvalidCandidate(
-					WasmInvalidCandidate::AmbiguousWorkerDeath |
-					WasmInvalidCandidate::AmbiguousJobDeath(_),
+				Err(ValidationError::PossiblyInvalid(
+					PossiblyInvalidError::AmbiguousWorkerDeath |
+					PossiblyInvalidError::AmbiguousJobDeath(_),
 				)) if num_death_retries_left > 0 => num_death_retries_left -= 1,
-				Err(ValidationError::InvalidCandidate(WasmInvalidCandidate::JobError(_)))
+				Err(ValidationError::Invalid(WasmInvalidCandidate::JobError(_)))
```
mrcnski, line 763: Since we retry in this case, it belongs under `PossiblyInvalid`.

polkadot/node/core/pvf/common/src/executor_intf.rs
```diff
@@ -140,21 +140,20 @@ pub unsafe fn create_runtime_from_artifact_bytes(
...
 }
 
-pub fn params_to_wasmtime_semantics(par: &ExecutorParams) -> Result<Semantics, String> {
+pub fn params_to_wasmtime_semantics(par: &ExecutorParams) -> Semantics {
 	let mut sem = DEFAULT_CONFIG.semantics.clone();
 	let mut stack_limit = if let Some(stack_limit) = sem.deterministic_stack_limit.clone() {
 		stack_limit
 	} else {
-		return Err("No default stack limit set".to_owned())
+		panic!("No default stack limit set");
```
mrcnski, line 156: Just to double-check here, this is following up on a discussion with @s0me0ne-unkn0wn right? Can you remind me what the reasoning was?

polkadot/node/core/pvf/src/error.rs
```diff
@@ -19,31 +19,44 @@ use polkadot_node_core_pvf_common::error::{InternalValidationError, PrepareError
 /// A error raised during validation of the candidate.
 #[derive(Debug, Clone)]
 pub enum ValidationError {
-	/// The error was raised because the candidate is invalid.
-	///
-	/// Whenever we are unsure if the error was due to the candidate or not, we must vote invalid.
-	InvalidCandidate(InvalidCandidate),
-	/// Some internal error occurred.
-	InternalError(InternalValidationError),
+	/// Deterministic preparation issue.
```
mrcnski, line 22: Maybe can be reworded? While these *should* be deterministic, in practice they should have been caught by prechecking so are more likely internal.

mrcnski:
From the issue:

> Timeout due to exceeded CPU time.
> [...]
> b. NOTE: There may not be enough time for a retry in candidate validation from backing.

I think it makes sense to retry this case, but only for approval. Maybe if `WorkerResponse::JobTimedOut` happened we can convert it to  `PossiblyPossiblyInvalid`. (😛😛😛) cc @Jpserrat since you worked on retries.

mrcnski (changes requested):
Awwww yeah. Mostly good, just missed one variant I think!

polkadot/node/core/candidate-validation/src/lib.rs
eagr: Oops, nice catch!

polkadot/node/core/pvf/common/src/executor_intf.rs
eagr: Yep. Coz that would be triggered when the default config is messed up. Deliberately or not, it's a sign of suicidal behavior, which should never happen. That's what you told me lol.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4405739

polkadot/node/core/pvf/common/src/executor_intf.rs
mrcnski: I see we have a comment on `DEFAULT_CONFIG` saying never to change it. So I think instead of a `panic!` we can make this an `expect` instead, with a [proof](https://github.com/paritytech/polkadot-sdk/blob/master/docs/STYLE_GUIDE.md#style). Or, since that doc says to avoid panics where possible, we can make some const like `DEFAULT_STACK_LIMIT` and refer to it and have `DEFAULT_CONFIG` also refer to it. Though that might require `clone` which I'm not sure is possible in a const context. Maybe `DeterministicStackLimit` can be `Copy` though.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -760,7 +755,7 @@ trait ValidationBackend {
 					PossiblyInvalidError::AmbiguousWorkerDeath |
 					PossiblyInvalidError::AmbiguousJobDeath(_),
 				)) if num_death_retries_left > 0 => num_death_retries_left -= 1,
-				Err(ValidationError::Invalid(WasmInvalidCandidate::JobError(_)))
+				Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::JobError(_)))
 					if num_job_error_retries_left > 0 =>
 					num_job_error_retries_left -= 1,
 				Err(ValidationError::Internal(_)) if num_internal_retries_left > 0 =>
```
mrcnski, line 767: Sorry to be pedantic, again but maybe instead of `_` below we can enumerate all the possible variants, since there shouldn't be many left: just `Prepare` and `Invalid`. This makes it more explicit, and also ensures that we never miss handling `PossiblyInvalid`. Actually, this comment isn't even pedantic but genuinely useful.

mrcnski (approved):
Yes!

polkadot/node/core/candidate-validation/src/lib.rs
eagr: That's actually a nice catch!

polkadot/node/core/pvf/common/src/executor_intf.rs
eagr: > So I think instead of a panic! we can make this an expect instead, with a [proof](https://github.com/paritytech/polkadot-sdk/blob/master/docs/STYLE_GUIDE.md#style).

But that's just a subtler way of doing what we do right now.

```rs
let mut stack_limit = sem.deterministic_stack_limit.clone()
	.expect("Default stack limit should always be available");

// louder, noisier, but maybe this is exactly what we want?
let mut stack_limit = if let Some(stack_limit) = sem.deterministic_stack_limit.clone() {
	stack_limit
} else {
	panic!("No default stack limit set");
};
```

The only way this could panick is when someone changes the code, which is not supposed happen ever I guess?

mrcnski: I don't see why we'd reimplement `expect` here TBH. 

> The only way this could panick is when someone changes the code, which is not supposed happen ever I guess?

Exactly, we can write a proof stating this, and that there is a comment saying not to change the code.

mrcnski:
> From the issue:
> 
> > Timeout due to exceeded CPU time.
> > [...]
> > b. NOTE: There may not be enough time for a retry in candidate validation from backing.
> 
> I think it makes sense to retry this case, but only for approval. Maybe if `WorkerResponse::JobTimedOut` happened we can convert it to `PossiblyPossiblyInvalid`. (😛😛😛) cc @Jpserrat since you worked on retries.

@eagr Are you planning to do this here? I think JP is busy with a different task.

polkadot/node/core/candidate-validation/src/lib.rs
mrcnski: Hmm, you're right. I'd still refactor it to move the ifs out of the guards. It's worth enumerating all the variants here.

polkadot/node/core/pvf/common/src/executor_intf.rs
eagr: As I said, `panic!` is louder and stands out in the code. But I didn't write the code. That's just my guess. :))

eagr:
I thought you're asking JP to do it in a different PR. Sure, I will. But could we do it in a different PR?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4411700

polkadot/node/core/candidate-validation/src/lib.rs
eagr: By moving the ifs into the branch, every branch would become `if ... else break`. Probably not worth doing. Another it-ain't-broke-don't-fix-it example lol.

mrcnski: I know, but it'd be really nice to enumerate all the variants here. Some extra verbosity seems worth it to maintain correctness.

polkadot/node/core/pvf/common/src/executor_intf.rs
```diff
@@ -140,22 +140,20 @@ pub unsafe fn create_runtime_from_artifact_bytes(
...
+pub fn params_to_wasmtime_semantics(par: &ExecutorParams) -> Semantics {
 	let mut sem = DEFAULT_CONFIG.semantics.clone();
-	let mut stack_limit = if let Some(stack_limit) = sem.deterministic_stack_limit.clone() {
-		stack_limit
-	} else {
-		return Err("No default stack limit set".to_owned())
-	};
+	let mut stack_limit = sem
+		.deterministic_stack_limit
+		.expect("Default stack limit should always be available")
```
mrcnski, line 155: Proofs should look like this:

```suggestion
		.expect("There is a comment to not change the default stack limit; it should always be available; qed")
```

Now, I'm not sure if we always follow this rule. 😅 Ping me if you see any incomplete proofs elsewhere.

polkadot/node/core/pvf/common/src/executor_intf.rs
mrcnski: I don't think it needs to be loud here; we can have a proof, and there is already a loud comment on the config. Sorry if we gave a different impression, I honestly don't remember the original conversation.

polkadot/node/core/candidate-validation/src/lib.rs
eagr: Are you really sure about this?

```rs
match validation_result {
	Err(ValidationError::PossiblyInvalid(
		PossiblyInvalidError::AmbiguousWorkerDeath |
		PossiblyInvalidError::AmbiguousJobDeath(_),
	)) =>
		if num_death_retries_left > 0 {
			num_death_retries_left -= 1;
		} else {
			break;
		},

	Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::JobError(_))) =>
		if num_job_error_retries_left > 0 {
			num_job_error_retries_left -= 1;
		} else {
			break;
		},

	Err(ValidationError::Internal(_)) =>
		if num_internal_retries_left > 0 {
			num_internal_retries_left -= 1;
		} else {
			break;
		},

	Ok(_) | Err(ValidationError::Invalid(_) | ValidationError::Preparation(_)) => break,
}
```

mrcnski: Hmm, I don't think it's *too* bad. I'm worried that someone in the future may add a `PossiblyInvalid` case and forget to update this spot, because the `_` swallows it.

Maybe a small helper macro would be less verbose and more clear, e.g.

```rs
	Err(ValidationError::Internal(_)) => retry_or_break!(num_internal_retries_left),
```

Or would it just obscure this code more? cc @s0me0ne-unkn0wn for a second opinion.

polkadot/node/core/pvf/common/src/executor_intf.rs
mrcnski: https://en.wikipedia.org/wiki/Q.E.D.

polkadot/node/core/candidate-validation/src/lib.rs
eagr: I get that from the beginning. That's why I think it's a very good catch! If you still think it's worth it after seeing the code, alright then~

s0me0ne-unkn0wn (approved):
Trying to stop myself from proposing to rename `PossiblyInvalid` into `SchrödingerCandidate` :grin: 
Ideally, from the point of view of determinism, we shouldn't have anything that is "possibly valid" or "possibly invalid", every possibility should deterministically resolve into either "valid" or "invalid". But the world's not perfect :man_shrugging:

mrcnski:
@s0me0ne-unkn0wn Yeah, and at least spelling out explicitly how we handle the non-determinism, in one place, by having these variants, is very useful.

> Trying to stop myself from proposing to rename PossiblyInvalid into SchrödingerCandidate 😁

I'm honestly not against this, we could use more creative names in the code. 😛

Issue #171, opened 2023-05-26
[FRAME Core] tracking issue for full unleash of `DefaultConfig` for FRAME


kianenigma:
https://github.com/paritytech/substrate/pull/13454 will bring in the basic functionality needed to start using this feature. See this example to learn more. 

This will mark the first version of this work complete, but a lot has to be done in order for it to be fully complete.

## Improve Error Handling, Bugs

* [ ] Improve error handling. There seems to be an error in the `parse/config.rs` file that allows `#[pallet::whatever]` to be parsed and ignored in `#[pallet::config]`. This has to be fixed.
* [x] Properly hide `__export_tokens_tt_test_default_config` and similar from docs. 
* [ ] Handle pallet instancing. 

## Full usage in Substrate Pallets 

* [ ] migrate the docs for `derive_impl` to use real code. As all of this functionality can work on normal traits, and as per paritytech/substrate#14115, we should strive to make the docs of `derive_impl` be real rust code. We might even want to remove `./frame/example/default-config` in favor this. 

* Provide testing defaults for common pallets
  * [x] System: already provided
  * [x] Balances https://github.com/paritytech/substrate/pull/14453
  * [x] Assets https://github.com/paritytech/polkadot-sdk/pull/3637

* Then, we should start migrating existing pallets and their test setup to use the unified testing prelude of system, balances and so on. Here's the tracking issue: https://github.com/paritytech/polkadot-sdk/issues/3237

## Full Usage in `node-template-runtime` and `parachain-template-runtime`

- [x] We should provide runtime templates that utilize this, evangelising its usage. This requires us to carefully consider what is a reasonable default for a node template in order to use. We should provide possibly two new preludes: 

## Full usage in `westend` -> `kusama` -> `polkadot` runtimes. 

- [x] Once we have are more and more confident in both the core functionality, and our skills to define sensible defaults, we can slowly start migrating perhaps some of the configs define in Polkadot to use this feature. 

For security reasons we should be super caution here. Adding `#[derive_impl]` will open the door for upstream changes manipulating the configurations used in the Polkadot runtime. Perhaps we should only ever do this once we have more tooling that could notify us of unexpected upstream changes to `config_preludes`. 

I would personally find it reasonable to never use this feature in production runtimes for this reason. 

## V2

At the moment, the system that is introduced is has some limitations. Most notably: 

[...]

PR #2409, opened 2023-11-20
Moves all test runtimes to use `derive_impl`

gupnik:
Step in https://github.com/paritytech/polkadot-sdk/issues/171

This PR adds `derive_impl` on all `frame_system` config impls for mock runtimes. The overridden configs are maintained as of now to ensure minimal changes.


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4398985

liamaharon (approved):

bkchr (commented):
I read that you wanted to keep the changes minimal. However, should we not just remove all the "overwrites" which are just the default value anyway? Especially for testing and that people can follow this example in their own code base.

gupnik:
> However, should we not just remove all the "overwrites" which are just the default value anyway?

Yes, that's why I am keeping the tracking issue open to remove these overwrites for each pallet. However, I wanted to get this PR merged, to ensure that any new changes (like `RuntimeTask`) can be continued without changes everywhere. I think removing all of these in a single PR would make it too cumbersome.

ggwpez:
Can we please merge this soon? I just added 5 default config items to ~140 runtimes in one of my MRs...  
Lets do the follow up of removing default in another MR then.

franciscoaguirre (approved):

acatangiu (approved):

ordian (approved):

cumulus/pallets/xcmp-queue/src/mock.rs
```diff
@@ -55,6 +55,7 @@ parameter_types! {
 
 type AccountId = u64;
 
+#[derive_impl(frame_system::config_preludes::TestDefaultConfig as frame_system::DefaultConfig)]
```
bkchr, line 58: We should really work on this syntax and make it better.

bkchr (approved):

cumulus/pallets/xcmp-queue/src/mock.rs
gupnik: Yes, will add this to the tracking issue

PR #2413, opened 2023-11-20
Update documentation for SafeMode and TxPause Pallets

### LINKED ISSUE #274: Transaction pause pallet
A pallet that allow governance origin to pause a certain calls.

It can be used as an emergency mechanism to pause an exploitable transaction type.
It can be used to disable a new feature and enable it at a later date.
It can be used to pause certain transaction type during a multi block migration.

Examples:

Acala: https://github.com/AcalaNetwork/Acala/blob/master/modules/transaction-pause/src/lib.rs
Bifrost: https://github.com/bifrost-finance/bifrost/blob/develop/pallets/call-switchgear/src/lib.rs
Manta: https://github.com/Manta-Network/Manta/tree/manta/pallets/tx-pause
Moonbeam: https://github.com/PureStake/moonbeam/blob/master/pallets/maintenance-mode/src/lib.rs
Parallel: https://github.com/parallel-finance/parallel/blob/master/pallets/emergency-shutdown/src/lib.rs

Related: https://github.com/paritytech/polkadot-sdk/issues/302

### LINKED ISSUE #302: Safe mode
Underway in https://github.com/paritytech/substrate/pull/12092

Introduce a system flag `SafeMode: Option<T::BlockNumber>`.

Also introduce a second `BaseFilter` called `SafeBaseFilter` into runtimes which restricts everything except calls to the governance pallets.

The actual filter is selected between `BaseFilter` and `SafeBaseFilter` depending on whether `SafeMode` is `Some` or not.

In the block initialization phase, then if `SafeMode` is `Some` and the inner block number is less than or equal to the current block number, then it is replaced with `None`.

A weak governance origin (e.g. one councillor) is able to upgrade `SafeMode` from `None` to be `Some` with a modest number of blocks (e.g. 10 minutes) higher than the current block number.

A strong governance origin (e.g. the entire council) is able to downgrade `SafeMode` to `None` or upgrade `SafeMode` from `None` to be `Some` with a substantial number of blocks higher than the current block number (e.g. 3 hours).

Anyone may place a signficant amount (e.g. 1,000 KSM) of their funds into reserve in order to upgrade `SafeMode` from `None` to be `Some` with a modest number of blocks (e.g. 10 minutes) higher than the current block number.

In that case, the reserved funds may be unreserved only after 24 hours and may be slashed or unreserved prior by a strong governance origin.

wilwade:
# Description

Documentation for the TxPause and SafeMode pallets.

Based on reading and the notes from the following related PRs:
- https://github.com/paritytech/substrate/pull/12092

I believe this also completes the checklist to be able to close the related PRs:
- Close #302 
- Close #274


substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//! Trigger for stopping all extrinsics outside of a specific whitelist.
+//!
+//! ## Pallet API
+//!
+//! See the [`pallet`] module for more information about the interfaces this pallet exposes,
+//! including its configuration trait, dispatchables, storage items, events and errors.
+//!
+//! ## Overview
+//!
+//! The safe-mode pallet provides a big STOP button to to put the chain in safe mode and thereby
```
joepetrowski, line 29: Redundant

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//! only permitting a certain subset of operations.
+//! The pallet provides a `WhitelistedCalls` which contains all calls that can be executed in
```
joepetrowski, line 31: A `WhitelistedCalls` what?

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//! while in safe mode.
+//!
+//! ### Primary Features:
```
joepetrowski, line 34:
```suggestion
//! ### Primary Features

```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//! - Entering safe mode can be permissionless with deposit, origin based, or some combination of
```
joepetrowski, line 35: Inconsistent comma usage with L25. (I prefer this way, with the serial comma, but it should at least be consistent.)

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//!   those
+//! - Separate origin configuration for entering and exiting safe mode
+//! - Safe mode must always be configured to exit at a specific block number
+//! - Safe mode exit block number may be extended by additional calls
```
joepetrowski, line 39:
```suggestion
//!   those.
//! - Separate origin configuration for entering and exiting safe mode.
//! - Safe mode must always be configured to exit at a specific block number.
//! - Safe mode exit block number may be extended by additional calls.
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//!
+//! ### Example
+//!
+//! Configuration of call filters:
+//!
+//! ```ignore
+//! impl frame_system::Config for Runtime {
+//!   …
+//!   type BaseCallFilter = InsideBoth<DefaultFilter, SafeMode>;
+//!   …
```
joepetrowski, line 49:
```suggestion
//!   // …
//!   type BaseCallFilter = InsideBoth<DefaultFilter, SafeMode>;
//!   // …
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,47 @@
...
+//! }
+//! ```
+//!
+//! ## Low Level / Implementation Details
+//!
+//! ### Use Cost
+//! A storage value (`EnteredUntil`) is used to store the block safe mode will be exited on.
+//! Using the call filter will require a db read of that storage on each extrinsic.
```
joepetrowski, line 57:
```suggestion
//!
//! A storage value (`EnteredUntil`) is used to store the block at which safe mode will end.
//! Using the call filter will require a DB read of that storage on each extrinsic.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! # Transaction Pause
+//!
+//! Allows dynamic chain state based pausing and unpausing of specific extrinsics via call filters.
```
joepetrowski, line 20:
```suggestion
//! Allows dynamic, chain-state-based pausing and unpausing of specific extrinsics via call filters.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
...
+//!
+//! ## Pallet API
+//!
+//! See the [`pallet`] module for more information about the interfaces this pallet exposes,
+//! including its configuration trait, dispatchables, storage items, events and errors.
+//!
+//! ## Overview
+//!
+//! Think of it as a dynamic call filter that can be controlled with extrinsics.
```
joepetrowski, line 29: Too conversational for reference docs.

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
...
+//!
+//! Pausing an extrinsic means that the extrinsic CANNOT be called again until it is are unpaused.
```
joepetrowski, line 31:
```suggestion
//! Pausing an extrinsic means that the extrinsic CANNOT be called again until it is unpaused.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
...
+//! The exception is calls that use `dispatch_bypass_filter`, typically only with the root origin.
+//!
+//! ### Primary Features:
+//! - Calls that should never be paused can be added to a whitelist.
+//! - Separate origin configuration for pausing and pausing
+//! - Pausing is triggered using the string representation of the call
+//! - Pauses can target a single extrinsic or an entire pallet
```
joepetrowski, line 38:
```suggestion
//! ### Primary Features
//!
//! - Calls that should never be paused can be added to a whitelist.
//! - Separate origin configuration for pausing and pausing.
//! - Pausing is triggered using the string representation of the call.
//! - Pauses can target a single extrinsic or an entire pallet.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
...
+//!
+//! ### Example
+//!
+//! Configuration of call filters:
+//!
+//! ```ignore
+//! impl frame_system::Config for Runtime {
+//!   …
+//!   type BaseCallFilter = InsideBoth<DefaultFilter, TxPause>;
+//!   …
```
joepetrowski, line 48:
```suggestion
//!   // …
//!   type BaseCallFilter = InsideBoth<DefaultFilter, TxPause>;
//!   // …
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,46 @@
...
+//! }
+//! ```
+//!
+//! ## Low Level / Implementation Details
+//!
+//! ### Use Cost
+//! A storage map (`PausedCalls`) is used to store currently paused calls.
+//! Using the call filter will require a db read of that storage on each extrinsic.
```
joepetrowski, line 56:
```suggestion
//! ### Use Cost
//!
//! A storage map (`PausedCalls`) is used to store currently paused calls.
//! Using the call filter will require a DB read of that storage on each extrinsic.
```

substrate/frame/safe-mode/src/lib.rs
wilwade: Updated this section some to try to make it more useful.

substrate/frame/tx-pause/src/lib.rs
wilwade: Yeah. I had lifted it from the PR. Updated to read more clearly.

wilwade:
@joepetrowski Updated with your suggestions. Thanks for the review!

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,48 @@
...
+//! ## Pallet API
+//!
+//! See the [`pallet`] module for more information about the interfaces this pallet exposes,
+//! including its configuration trait, dispatchables, storage items, events, and errors.
+//!
+//! ## Overview
+//!
+//! Safe mode is entered via two paths (deposit or forced) until a set block number.
+//! The mode is exited when the block number is reached or a call to one of the exit extrinsics is
+//! made. A `WhitelistedCalls` configuration contains all calls that can be executed in while in
```
ggwpez, line 31:
```suggestion
//! made. A `WhitelistedCalls` configuration contains all calls that can be executed while in
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,48 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! # Safe Mode
```
ggwpez, line 18: Can you please put a warning on top that it is not yet audited? Something like:  

```
# WARNING

NOT YET AUDITED. DO NOT USE IN PRODUCTION.
```

Also for the tx-pause pallet.

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,48 @@
...
+//!   type BaseCallFilter = InsideBoth<DefaultFilter, SafeMode>;
+//!   // …
+//! }
+//! ```
+//!
+//! ## Low Level / Implementation Details
+//!
+//! ### Use Cost
+//! A storage value (`EnteredUntil`) is used to store the block safe mode will be exited on.
+//! Using the call filter will require a db read of that storage on each extrinsic.
```
ggwpez, line 58: Yes and please note that the DB read will only hit the DB on the first time, after that it will be on the overlay and read quickly.

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,48 @@
...
+//!
+//! ### Primary Features
+//!
+//! - Entering safe mode can be permissionless with deposit, origin based, or some combination of
+//!   those.
+//! - Separate origin configuration for entering and exiting safe mode.
+//! - Safe mode must always be configured to exit at a specific block number.
+//! - Safe mode exit block number may be extended by additional calls.
+//!
+//! ### Example
```
ggwpez, line 42: I think there could be more example, especially using `docify` to show how its used.  
You can probably just embed one of the basic tests for this.

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,48 @@
...
+//! Pausing an extrinsic means that the extrinsic CANNOT be called again until it is unpaused.
+//! The exception is calls that use `dispatch_bypass_filter`, typically only with the root origin.
+//!
+//! ### Primary Features
+//!
+//! - Calls that should never be paused can be added to a whitelist.
+//! - Separate origin configuration for pausing and pausing.
+//! - Pausing is triggered using the string representation of the call.
+//! - Pauses can target a single extrinsic or an entire pallet.
+//!
```
ggwpez, line 40: One niche thing here, not sure if it is worth to mention, but it is possible to pause transactions of not-yet deployed pallets. This could be useful to selectively open up functionality of a pallet after its initial deployment.

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,48 @@
...
+//! ### Example
+//!
+//! Configuration of call filters:
```
ggwpez, line 43: Please also another simple example of how to use it, possibly by `docify` embedding a test 🙏

ggwpez (commented):
Thanks for the work!

substrate/frame/tx-pause/src/lib.rs
wilwade: Tried setting this up. Does this look useful/correct to you?
image

substrate/frame/safe-mode/src/lib.rs
wilwade: Tried setting this up. Does this look useful/correct to you?
image

substrate/frame/safe-mode/src/lib.rs
ggwpez: Yea, a `StorageMap` is cached per key. So if the same Call gets sent twice, then it is cached (but not in general).  
On the `SafeMode` it is easier since it uses a `StorageValue` that gets cached from the first access on.

substrate/frame/tx-pause/src/lib.rs
ggwpez: Yea this looks good. Maybe some more comments in the tests, but i think it works for a first version 👍

ggwpez (approved):

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,63 @@
...
+//! ## Pallet API
+//!
+//! See the [`pallet`] module for more information about the interfaces this pallet exposes,
+//! including its configuration trait, dispatchables, storage items, events, and errors.
+//!
+//! ## Overview
+//!
+//! Safe mode is entered via two paths (deposit or forced) until a set block number.
+//! The mode is exited when the block number is reached or a call to one of the exit extrinsics is
+//! made. A `WhitelistedCalls` configuration contains all calls that can be executed while in
```
franciscoaguirre, line 35:
```suggestion
//! made. A `WhitelistedCalls` configuration item contains all calls that can be executed while in
```

franciscoaguirre (approved):
Glad to get more docify-powered pallet docs 🚀

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,63 @@
...
+//! safe mode.
```
joepetrowski, line 36: Line length.
```suggestion
//! Safe mode is entered via two paths (deposit or forced) until a set block number. The mode is
//! exited when the block number is reached or a call to one of the exit extrinsics is made. A
//! `WhitelistedCalls` configuration contains all calls that can be executed while in safe mode.
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,63 @@
...
+//!
+//! ### Primary Features
+//!
+//! - Entering safe mode can be permissionless with deposit, origin based, or some combination of
+//!   those.
+//! - Separate origin configuration for entering and exiting safe mode.
+//! - Safe mode must always be configured to exit at a specific block number.
+//! - Safe mode exit block number may be extended by additional calls.
```
joepetrowski, line 44: Structural consistency. Also I don't see how you can have a combination of deposit and origin?

```suggestion
//! This pallet provides:
//!
//! - Ability to enter safe mode with a privileged origin or by anyone placing a deposit;
//! - Separate origin configuration items for entering and exiting safe mode;
//! - Configuration item to set a safe mode duration, after which the system will exit safe mode;
//! - Ability to extend safe mode beyond its configured exit, if needed.
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,63 @@
...
+//! ```
+//!
+//! Entering safe mode with deposit:
+#![doc = docify::embed!("src/tests.rs", can_activate)]
+//!
+//! Entering safe mode with force:
+#![doc = docify::embed!("src/tests.rs", can_force_activate_with_config_origin)]
+//!
+//! Exiting safe mode with force:
+#![doc = docify::embed!("src/tests.rs", can_force_deactivate_with_config_origin)]
```
joepetrowski, line 65: Force is OK, but IMO "with privileged origin" is more descriptive.
```suggestion
//! Entering safe mode via privileged origin:
#![doc = docify::embed!("src/tests.rs", can_force_activate_with_config_origin)]
//!
//! Exiting safe mode via privileged origin:
#![doc = docify::embed!("src/tests.rs", can_force_deactivate_with_config_origin)]
```

substrate/frame/safe-mode/src/lib.rs
```diff
@@ -15,6 +15,63 @@
...
+//!
+//! ## Low Level / Implementation Details
+//!
+//! ### Use Cost
+//!
+//! A storage value (`EnteredUntil`) is used to store the block safe mode will be exited on.
+//! Using the call filter will require a db read of that storage on the first extrinsic.
+//! The storage will be added to the overlay and incur low cost for all additional calls.
```
joepetrowski, line 73: Line length
```suggestion
//! A storage value (`EnteredUntil`) is used to store the block safe mode will be exited on. Using
//! the call filter will require a db read of that storage on the first extrinsic. The storage will
//! be added to the overlay and incur low cost for all additional calls.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,62 @@
...
+//! Allows dynamic, chain-state-based pausing and unpausing of specific extrinsics via call filters.
+//!
+//! ## WARNING
+//!
+//! NOT YET AUDITED. DO NOT USE IN PRODUCTION.
+//!
+//! ## Pallet API
+//!
+//! See the [`pallet`] module for more information about the interfaces this pallet exposes,
+//! including its configuration trait, dispatchables, storage items, events and errors.
```
joepetrowski, line 29:
```suggestion
//! including its configuration trait, dispatchables, storage items, events, and errors.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,62 @@
...
+//!
+//! ## Overview
+//!
+//! A dynamic call filter that can be controlled with extrinsics.
+//!
+//! Pausing an extrinsic means that the extrinsic CANNOT be called again until it is unpaused.
+//! The exception is calls that use `dispatch_bypass_filter`, typically only with the root origin.
```
joepetrowski, line 36:
```suggestion
//! Pausing an extrinsic means that the extrinsic CANNOT be called again until it is unpaused. The
//! exception is calls that use `dispatch_bypass_filter`, typically only with the root origin.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,62 @@
...
+//!
+//! ### Primary Features
+//!
+//! - Calls that should never be paused can be added to a whitelist.
+//! - Separate origin configuration for pausing and pausing.
+//! - Pausing is triggered using the string representation of the call.
+//! - Pauses can target a single extrinsic or an entire pallet.
+//! - Pauses can target future extrinsics or pallets
```
joepetrowski, line 44: Structural consistency :)
```suggestion
//! - Calls that should never be paused can be added to a whitelist.
//! - Separate origins are configurable for pausing and pausing.
//! - Pausing is triggered using the string representation of the call.
//! - Pauses can target a single extrinsic or an entire pallet.
//! - Pauses can target future extrinsics or pallets.
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,62 @@
...
+//! }
+//! ```
+//!
+//! Pause specific all:
+#![doc = docify::embed!("src/tests.rs", can_pause_specific_call)]
+//!
+//! Unpause specific all:
+#![doc = docify::embed!("src/tests.rs", can_unpause_specific_call)]
+//!
+//! Pause all in a pallet:
```
joepetrowski, line 64:
```suggestion
//! Pause all calls in a pallet:
```

substrate/frame/tx-pause/src/lib.rs
```diff
@@ -15,6 +15,62 @@
...
+#![doc = docify::embed!("src/tests.rs", can_pause_all_calls_in_pallet_except_on_whitelist)]
+//!
+//! ## Low Level / Implementation Details
+//!
+//! ### Use Cost
+//!
+//! A storage map (`PausedCalls`) is used to store currently paused calls.
+//! Using the call filter will require a db read of that storage on each extrinsic.
```
joepetrowski, line 72:
```suggestion
//! A storage map (`PausedCalls`) is used to store currently paused calls. Using the call filter
//! will require a db read of that storage on each extrinsic.
```

joepetrowski (approved):
Few nits but otherwise good.

substrate/frame/safe-mode/src/lib.rs
wilwade: The combo is possible as one of the options in the following true table:

- `Force*Origin` set & Deposit = Some
- `Force*Origin` set & Deposit = None
- `Force*Origin` root/none & Deposit = Some
- (And well no reason to set both to none)

Revised the bullet points some based on your suggestions (incoming in next commit)

substrate/frame/safe-mode/src/lib.rs
wilwade: @joepetrowski (Using this thread for all the comments of this type). Here's the pattern I followed that ended up with the specific line length:

1. Each sentence should be on its own line. Found this is best for git to highlight only the changes and markdown and as far as I have seen cargo docs render it the same.
2. Run `cargo fmt` from root which is what formats the line length.

Perhaps I am missing the correct fmt command?

ggwpez: The line length in our project is default to exactly 100, the contributing docs only mention "not longer" https://github.com/paritytech/polkadot-sdk/blob/master/docs/STYLE_GUIDE.md#formatting but its actually an "exact 100".  
So its enough to write the text and then call fmt to have the correct breaks.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4460128

substrate/frame/safe-mode/src/lib.rs
joepetrowski: (1) makes sense for git, but when reading unrendered I think it's harder to read. (2) I don't think cargo does this, but am maybe wrong?

substrate/frame/safe-mode/src/lib.rs
joepetrowski: Ah, well I understood it as you could partially supply the deposit and then use some less-privileged origin.

I still think what I wrote is better. Of course the privilege origin/deposit can be practically unattainable. Even in pallets that don't allow setting `Deposit = None`, you could still configure it to be greater than the total issuance. So what you are saying in your comment applies to everything. IMO it reduces clarity to try to finesse your way around that, just document the configuration items properly.

substrate/frame/safe-mode/src/lib.rs
wilwade: I can confirm that cargo does do it, as it makes the needed changes.

Happy to go either way on the plain text. I find it easier to read, but written a lot of markdown that way so it might just be that I am used to it.

substrate/frame/safe-mode/src/lib.rs
wilwade: Here's the update from the last commit that tries to clarify. Mostly trying to highlight to someone who is looking for only one or the other that either setup is possible.

```
//! - Entering safe mode can be with a privileged origin or anyone placing a deposit.
//! - Deposit method or privileged origin may be separately disabled.
//! - Separate privileged origin configuration items for entering and exiting safe mode.
//! - Configuration item to set a safe mode block count duration, after which the system will exit
//!   safe mode.
//! - Safe mode may be extended beyond the configured exit by additional calls.
```

joepetrowski: To me yours is hard to read because the items take many different forms. Read them out loud, the third and fourth ones jump out because they are not sentences (there is no conjugated verb). (First has a typo BTW, "can be with".) I am more of a _snoot_ than others, will stop arguing from here, but this stuff bothers me :upside_down_face: 

IMO your second point is not necessary, the "or" in the first expresses it, and each configuration item's docs should give details on specific configurations. This is "Primary Features", not details about every config option or implementation details. It's very common to have config items have a setting that practically disables them. Is there a reason to highlight this at this level? I don't think the option to disable a config item would stop someone from using the pallet.

wilwade: It bothers me as well, so please continue. I'm just not great at it :)

I think the dual use cases is important and not quickly clear from the configuration options, but I agree that the "or" does the job.

How does this sound:

```
//! - Entering safe mode can be via privileged origin or anyone who places a deposit.
//! - Origin configuration items are separated for privileged entering and exiting safe mode.
//! - A configured block count duration, after which the system will exit safe mode.
//! - Safe mode may be extended beyond the configured exit by additional calls.
```

joepetrowski: I would just change the third to:
```
//! - A configurable duration sets the number of blocks after which the system will exit safe mode.
```

wilwade: Updated. Thanks for the help!

wilwade:
I believe all current comments have been addressed. Let me know if I have missed something! (And thanks for the help)

PR #2424, opened 2023-11-21
Improve fuzzer for XCMv4

louismerlin:
# Description

This PR includes improvements to the xcm-simulator-fuzzer to make it find more coverage in XCMv4.

This includes switching the fuzzing engine from [honggfuzz-rs](https://github.com/rust-fuzz/honggfuzz-rs) to both honggfuzz-rs and [afl.rs](https://github.com/rust-fuzz/afl.rs) through [ziggy](https://github.com/srlabs/ziggy), a fuzzer manager.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4419542

polkadot/xcm/xcm-simulator/fuzzer/README.md
```diff
@@ -6,35 +6,31 @@ underflows.
...
+cargo install --force ziggy cargo-afl honggfuzz grcov
 ```
 
 ## Run the fuzzer
 
 In this directory, run this command:
 
 ```
-cargo hfuzz run xcm-fuzzer
+cargo ziggy fuzz
```
ggwpez, line 17: It seems to be running on only one thread when i try this. Is it expected?  
Hongfuzz can use all threads.

polkadot/xcm/xcm-simulator/fuzzer/src/main.rs
```diff
@@ -155,6 +155,35 @@ fn run_input(xcm_messages: [XcmMessage; 5]) {
 	println!();
 
 	for xcm_message in xcm_messages {
+		// We block Transact messages because they can cause panics that are due to the test
+		// environment, and not XCM
```
ggwpez, line 159: For example?

ggwpez (approved):

polkadot/xcm/xcm-simulator/fuzzer/src/main.rs
louismerlin: I just checked, and got the following:
```
Unexpected event: RuntimeEvent::MessageQueue(
    Event::ProcessingFailed {
        ...
        error: ProcessMessageError::Unsupported,
    },
)
```
[in this line of the event processing code](https://github.com/louismerlin/polkadot-sdk/blob/80808d34fa1ae8eef6c30c1d6cc0555aa03596db/polkadot/xcm/xcm-simulator/src/lib.rs#L148) when processing an invalid Transact call.
@ggwpez @franciscoaguirre do you think this should be handled differently, e.g. by allowing Unsupported messages instead of panicking, and then removing the blocklist?

franciscoaguirre: We could allow that. The thing is arbitrary Transacts will call any type of extrinsic, with wrong attributes or anything of the sort. We could record which Transacts give an error and which ones don't and why. In that way we would know much more than just by blocklisting them

acatangiu (approved):

ggwpez:
Why was this closed?

PR #2426, opened 2023-11-21
PVF: Fix unshare "no such file or directory" error

### LINKED ISSUE #2320: PVF: refactor: replace `tmppath` functions with tempfile crate
We can use the [`tempfile` crate](https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#method.prefix) instead. We already use it for tests.

The main places `tmppath` and `tmppath_in` functions are used in production are:
1. Creating the socket dir. `with_transient_socket_path` can be eliminated since `NamedTempFile` is already cleaned up when dropped by a Rust destructor.
2. The worker dir. `impl Drop for WorkerDir` can probably be eliminated for the same reasons as above.

mrcnski:
In https://github.com/paritytech/polkadot-sdk/pull/2304 we introduced this bug where unshare will always fail, leading to a scary error message for validators.

Screenshot 2023-11-21 at 15 57 15

Not sure how this wasn't caught in testing, neither for me nor the user who reported #2304. 🙈 Sorry. I'm not sure if it's possible to include this fix in 1.4.0.

Closes #2320

### DIFF
--- a/polkadot/node/core/pvf/src/worker_intf.rs
+++ b/polkadot/node/core/pvf/src/worker_intf.rs
@@ -71,6 +71,7 @@ pub async fn spawn_with_program_path(
 
 	with_transient_socket_path(debug_id, |socket_path| {
 		let socket_path = socket_path.to_owned();
+		let worker_dir_path = worker_dir.path().to_owned();
 
 		async move {
 			let listener = UnixListener::bind(&socket_path).map_err(|err| {
@@ -91,7 +92,7 @@ pub async fn spawn_with_program_path(
 				&program_path,
 				&extra_args,
 				&socket_path,
-				&worker_dir.path,
+				&worker_dir_path,
 				security_status,
 			)
 			.map_err(|err| {
@@ -100,7 +101,7 @@ pub async fn spawn_with_program_path(
 					%debug_id,
 					?program_path,
 					?extra_args,
-					?worker_dir.path,
+					?worker_dir_path,
 					?socket_path,
 					"cannot spawn a worker: {:?}",
 					err,
@@ -108,7 +109,6 @@ pub async fn spawn_with_program_path(
 				SpawnErr::ProcessSpawn
 			})?;
 
-			let worker_dir_path = worker_dir.path.clone();
 			futures::select! {
 				accept_result = listener.accept().fuse() => {
 					let (stream, _) = accept_result.map_err(|err| {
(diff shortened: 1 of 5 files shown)


polkadot/node/core/pvf/src/worker_intf.rs
```diff
@@ -150,7 +150,42 @@ where
 	F: FnOnce(&Path) -> Fut,
 	Fut: futures::Future<Output = Result<T, SpawnErr>> + 'static,
 {
-	let socket_path = tmppath(&format!("pvf-host-{}", debug_id))
+	/// Returns a path under [`std::env::temp_dir`]. The path name will start with the given prefix.
+	///
+	/// There is only a certain number of retries. If exceeded this function will give up and return
+	/// an error.
+	pub async fn tmppath(prefix: &str) -> io::Result<PathBuf> {
```
mrcnski, line 157: The root cause was this function returning a valid path but not creating the directory. It should only be used by `with_transient_socket_path`, which requires a valid path but nothing existing at the path. Elsewhere we should be using`tempfile` to correctly create temporary files or directories, preventing this kind of mistake in the future.

alindima (approved):

.gitlab/pipeline/test.yml
```diff
@@ -29,7 +29,7 @@ test-linux-stable:
         --locked \
         --release \
         --no-fail-fast \
-        --features try-runtime,experimental \
+        --features try-runtime,experimental,ci-only-tests \
```
mrcnski, line 32: Pretty sure this used to be here!

s0me0ne-unkn0wn (approved):

.gitlab/pipeline/test.yml
mrcnski: @paritytech/ci How can we remove this feature for `test-linux-oldkernel-stable`? It's expected that this test fails on old kernels.

alindima: considering that the feature is called `ci-only-tests` and not smth like `ci-only-tests-new-kernel`, maybe we could instead check the kernel version in the test (using `uname`) and skip the test if it's too old

mrcnski: Great idea @alindima, done.

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4432563

sergejparity (approved):

alvicsam (approved):

EgorPopelyaev (approved):

PR #2428, opened 2023-11-21
Fix ParentOrSiblings

franciscoaguirre:
We were not filtering for sibling parachains, but for sibling anything


cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs
```diff
@@ -241,7 +241,7 @@ match_types! {
 	};
 	pub type ParentOrSiblings: impl Contains<MultiLocation> = {
 		MultiLocation { parents: 1, interior: Here } |
-		MultiLocation { parents: 1, interior: X1(_) }
+		MultiLocation { parents: 1, interior: X1(Parachain(_)) }
 	};
```
acatangiu, line 245: I would also rename the type to smth like `ParentOrSiblingParachains` or even `ParentRelayOrSiblingParachains` to make it clear this is to be used  by parachains for same-relay-ecosystem filters

franciscoaguirre: Agreed, I also moved this to parachains-common to know it's being used for our use-case of relay and parachains

acatangiu (approved):

bkontur (approved):

bkontur:
bot fmt

muharem (approved):
I personally prefer such configs to be located within the runtime crate, not in common.

franciscoaguirre:
It's in common just because it's a configuration that is common to all our runtimes, as a way of deduplicating code

cumulus/parachains/common/src/xcm_config.rs
```diff
@@ -120,6 +120,17 @@ impl<AssetLocation: Get<MultiLocation>> ContainsPair<MultiAsset, MultiLocation>
 	}
 }
 
+pub struct ParentRelayOrSiblingParachains;
+impl Contains<MultiLocation> for ParentRelayOrSiblingParachains {
+	fn contains(location: &MultiLocation) -> bool {
+		matches!(
+			location,
+			MultiLocation { parents: 1, interior: Here } |
```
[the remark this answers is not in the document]
joepetrowski: Yeah that works.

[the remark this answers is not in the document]
franciscoaguirre: Added docs, please re-review

joepetrowski (approved):

cumulus/parachains/common/src/xcm_config.rs
```diff
@@ -120,6 +120,20 @@ impl<AssetLocation: Get<MultiLocation>> ContainsPair<MultiAsset, MultiLocation>
 	}
 }
 
+/// Filter for the parent relay or sibling parachains.
+/// Can be used only if the current location is a parachain, since it does not verify that the parent
+/// is indeed a relaychain.
```
joepetrowski, line 125:
```suggestion
/// Filter to check if a given location is the parent Relay Chain or a sibling parachain.
///
/// This type should only be used within the context of a parachain, since it does not verify that
/// the parent is indeed a Relay Chain.
```

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -125,10 +127,6 @@ match_types! {
 		MultiLocation { parents: 1, interior: Here } |
 		MultiLocation { parents: 1, interior: X1(Plurality { .. }) }
 	};
-	pub type ParentOrSiblings: impl Contains<MultiLocation> = {
-		MultiLocation { parents: 1, interior: Here } |
-		MultiLocation { parents: 1, interior: X1(_) }
```
rphmeier, line 130: Not an XCM expert, but what does a sibling "anything" here include other than parachains? Accounts?

franciscoaguirre: Yes, it could be anything in the parent, so not really a sibling. Accounts in the parent, pallets, governance bodies, for example.

bkchr (approved):

franciscoaguirre:
bot fmt

PR #2001, opened 2023-10-24
cumulus-consensus-common: block import: `delayed_best_block` flag added

michalkucharczyk:
This PR adds the  `delayed_best_block` flag to `ParachainBlockImport`. If not set, the `params.fork_choice` is not updated (to `ForkChoiceStrategy::Custom`) during the block import. 

When `delayed_best_block` is set to `false`  all parachain blocks on the [longest fork](https://github.com/paritytech/polkadot-sdk/blob/552be4800d9e4b480f79a300fc531783e04be099/substrate/client/service/src/client/client.rs#L708-L709) will be notified as the best block, allowing transaction pool to be updated with every imported block.

Otherwise imported blocks will  not be notified as best blocks (`fork_choice=ForkChoiceStrategy::Custom(false)`), and transaction pool would be updated only with best block received from relay-chain.

Improvement for: #1202

PR #2430, opened 2023-11-21
cumulus-test-service: block import fix

michalkucharczyk:
This is follow-up for: https://github.com/paritytech/polkadot-sdk/pull/2001

Block import queue for `test-parachain` (`cumulus-test-service`) shall use delayed best block feature.

This should fixed broken zombienet tests.

### DIFF
--- a/cumulus/test/service/src/lib.rs
+++ b/cumulus/test/service/src/lib.rs
@@ -211,7 +211,8 @@ pub fn new_partial(
 		sc_service::new_full_parts::<Block, RuntimeApi, _>(config, None, executor)?;
 	let client = Arc::new(client);
 
-	let block_import = ParachainBlockImport::new(client.clone(), backend.clone());
+	let block_import =
+		ParachainBlockImport::new_with_delayed_best_block(client.clone(), backend.clone());
 
 	let registry = config.prometheus_registry();
 


michalkucharczyk:
Should the following also be updated?
https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/polkadot-parachain/src/service.rs#L344
https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/parachain-template/node/src/service.rs#L127

bkchr (approved):

skunert:
> Should the following also be updated?
> 
> https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/polkadot-parachain/src/service.rs#L344
> 
> https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/parachain-template/node/src/service.rs#L127

I think it depends on whether they are running async-backing.

The template does not, but the reference in polkadot-parachain is used by both async backing ones and non-async backing, so you need to check.

skunert (approved):

michalkucharczyk:
> > Should the following also be updated?
> > https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/polkadot-parachain/src/service.rs#L344
> > 
> > https://github.com/paritytech/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/parachain-template/node/src/service.rs#L127
> 
> I think it depends on whether they are running async-backing.
> 
> The template does not, but the reference in polkadot-parachain is used by both async backing ones and non-async backing, so you need to check.
> 
> For example [here](https://github.com/skunert/polkadot-sdk/blob/50811d6b424185e1e6c0762a8b1c95db893c6750/cumulus/polkadot-parachain/src/service.rs#L1669) we call `aura::run` which is the async backing variant, but it also gets its ParachainBlockImport from `new_partial`.

I think some import paths maybe broken. For example [relay-chain verifier](https://github.com/paritytech/polkadot-sdk/blob/ec189333845fffebd45bf023b0797d6a4f121d8d/cumulus/client/consensus/relay-chain/src/import_queue.rs#L54C1-L54C1) and [relay-or-aura verifier](https://github.com/paritytech/polkadot-sdk/blob/49874882cf3eadde01951ee48916c9811d6a8c7e/cumulus/polkadot-parachain/src/service.rs#L1262-L1275) don't set fork choice.

We already discussed that with @skunert and there are three proposals to solve this:
- set `Custom(false)` in [`ParachainBlockImport::import_block`](https://github.com/paritytech/polkadot-sdk/blob/ec189333845fffebd45bf023b0797d6a4f121d8d/cumulus/client/consensus/common/src/lib.rs#L205)
- set `fork_choice` in relay-chain verifier,
- pass bool flag to new_partial and create properly configure `ParachainBlockImport`

I am really not sure which is most elegant.

@bkchr @skunert wdyt?

bkchr:
> * set `fork_choice` in relay-chain verifier,

This.

Issue #742, opened 2023-01-05
Time Disputes


eskimor:
Latest design [here](https://github.com/paritytech/polkadot-sdk/issues/742#issuecomment-1691884835).

# Basic Idea/Design Considerations

## New objectives:

1. It is not a security threat for approval checkers to not vote for a very long
   time, because it is taking long - approval voting is secure regardless.
2. We do need some timeout for liveness only, without a dispute we would never
   finalize.
3. Backers are special, they are the only validators that are forced to commit
   to a <2s execution.
4. Approval checkers will send timing information, if it took them significantly longer than
   the backing timeout then this is put on chain.
5. We charge backers via era points for that extra time. The charging will be
   exponential and reaches the cost of raising a concluding valid dispute. Thus
   we can actually have _very_ long approval voting timeouts.
6. Voting invalid validators will not get charged for wasting everybody's time,
   if that bill has already been paid by the backers via 5. (Getting honest
   validators slashed is thus no longer possible.) Basically validators voting
   invalid only need to get charged whatever the backers were not charged yet.
7. On disputes concluding invalid, backers always get slashed 100%, voting valid
   approval checkers (and dispute participants) only if there is no (small
   enough) time report of approval checkers. This way there is an incentive to
   actually validate (and not just always vote valid  -which would be the case if we always only slashed backers).

### Questions?

1. Would it be possible for malicious approval checkers to get honest backers
   charged? Assuming enough honest approval checkers, this should not be
   possible.

### Threats/Complications:

Getting those approval votes (time) on chain:

1. How many? Only tranch 0? Is that secure, what if validators are DoSed?
2. Cherry picking?
3. If we allow for a very long approval voting timeout, we might run into issues with state pruning. Let's assume an attacker tries to get a bad candidate in. It makes it so, that it takes a very long time on honest checkers. Now if he sees that approval checkers are not his guys, the attacker reveals a better fork (not including that candidate) - that chain gets finalized. Now by the time the validation for the fork timed out, the chain's state has already been discarded (too far behind the finalized block height), which might prevent a dispute from succeeding. In fact nodes might not even participate at all, as they have not "seen" the candidate backed/included.  Both should be fixable though, by not clearing up too eagerly on finalization.

[...]

Issue #635, opened 2023-05-15
Slash approval voters on approving invalid blocks - dynamically


eskimor:
Currently we punt on slashing approval voters approving invalid blocks. This is sound from a gambler's ruin perspective, as the backers are the ones to be supposed to having skin in the game. We will always slash at least those backers.

While the security properties are maintained, there is a slight problem: Without slashing approval voters, there is no incentive left to actually do the validation. Approving without having checked is risk free, for the individual. Of course not for the network: Blindly approving puts the whole chain at risk, given that validators are heavily invested, they should have an interest in maintaining the network's security. Nevertheless there is a tragedy of the commons situation: "If it is just me not validating - I'll save a few resources and as long as everybody else is checking, everything is fine."

Right now this should be a very low risk as running modified source code is neither effort less nor risk free and also for now resources are mostly spent elsewhere. The actual validation is relatively cheap, nevertheless longer term this situation should be improved.

There are at least two possibilities:

1. Somehow let the validators [prove that they did the validation](https://github.com/paritytech/polkadot-sdk/issues/639).
2. Make approving no longer risk free - bring back the slash.

While 1 would be a more solid solution: It would just not be possible to not doing the validation without getting slashed/losing out on rewards. On the flip-side it is not even clear how this should work, there are some ideas how this could work, but any proposed solution is far from trivial and would add a lot of complexity.

Which leaves us with option 2. So, why did we remove that slash in the first place? The reason was an attack vector: If someone hacked the backing validators/was willing to sacrifice them and would not care about getting them slashed, he could craft a candidate timing out on two thirds of the validators: Thus the candidate would be deemed invalid and the attacker got up to a third of the validators slashed 100%, by only controlling as little as two backers. It is worth mentioning, that this attack is _not_ resolved by [time disputes](#6510) as time disputes only resolve disputes concluding [for the candidate](https://github.com/paritytech/polkadot-sdk/issues/742) and charging backers is also no threat if they are willing to get slashed.

So why is this a problem? This would clearly be an unjustified slash, governance could refund them. True, but slashing also means:

1. Validators will get disabled.
2. Validators will be removed from the active set.

Both of these would gain an attacker an advantage: Knocking out up to a third of all validators. With byzantine assumptions the attacker already controls up to a third. So with this attack, he would control half the validators!

# Proposed Solution

[...]

PR #2438, opened 2023-11-22
Retry timeout

### LINKED ISSUE #660: PVF: Add a new class of "possibly invalid" errors

eagr:
closes #660 

Is the retry delay being used (200ms) good enough for the retry for job timeout?

Btw, how serious are you with the `SchrodingersCandidate`? I got zero problem with the name. :)) @mrcnski @s0me0ne-unkn0wn

### DIFF
--- a/polkadot/node/core/pvf/src/execute/queue.rs
+++ b/polkadot/node/core/pvf/src/execute/queue.rs
@@ -346,10 +346,6 @@ fn handle_job_finish(
 			None,
 		),
 		Outcome::InternalError { err } => (None, Err(ValidationError::Internal(err)), None),
-		// Either the worker or the job timed out. Kill the worker in either case. Treated as
-		// definitely-invalid, because if we timed out, there's no time left for a retry.
-		Outcome::HardTimeout =>
-			(None, Err(ValidationError::Invalid(InvalidCandidate::HardTimeout)), None),
 		// "Maybe invalid" errors (will retry).
 		Outcome::WorkerIntfErr => (
 			None,
@@ -361,6 +357,10 @@ fn handle_job_finish(
 			Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::AmbiguousJobDeath(err))),
 			None,
 		),
+		// Either the worker or the job timed out. Kill the worker in either case. Treated as
+		// definitely-invalid, because if we timed out, there's no time left for a retry.
+		Outcome::HardTimeout =>
+			(None, Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::HardTimeout)), None),
 		Outcome::JobError { err } =>
 			(None, Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::JobError(err))), None),
 	};
--- a/polkadot/node/core/candidate-validation/src/lib.rs
+++ b/polkadot/node/core/candidate-validation/src/lib.rs
@@ -767,21 +767,20 @@ trait ValidationBackend {
 					} else {
 						break;
 					},
-
-				Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::JobError(_))) =>
+				Err(ValidationError::PossiblyInvalid(
+					PossiblyInvalidError::JobError(_) | PossiblyInvalidError::HardTimeout,
+				)) =>
 					if num_job_error_retries_left > 0 {
 						num_job_error_retries_left -= 1;
 					} else {
 						break;
 					},
-
 				Err(ValidationError::Internal(_)) =>
 					if num_internal_retries_left > 0 {
 						num_internal_retries_left -= 1;
 					} else {
 						break;
 					},
-
 				Ok(_) | Err(ValidationError::Invalid(_) | ValidationError::Preparation(_)) => break,
 			}
 
--- a/polkadot/node/core/candidate-validation/src/lib.rs
+++ b/polkadot/node/core/candidate-validation/src/lib.rs
@@ -651,21 +651,22 @@ async fn validate_candidate_exhaustive(
 			);
 			Err(ValidationFailed(e.to_string()))
 		},
-		Err(ValidationError::Invalid(WasmInvalidCandidate::HardTimeout)) =>
-			Ok(ValidationResult::Invalid(InvalidCandidate::Timeout)),
 		Err(ValidationError::Invalid(WasmInvalidCandidate::WorkerReportedInvalid(e))) =>
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(e))),
+
 		Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::AmbiguousWorkerDeath)) =>
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(
 				"ambiguous worker death".to_string(),
 			))),
-		Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::JobError(err))) =>
-			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(err))),
-
 		Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::AmbiguousJobDeath(err))) =>
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(format!(
 				"ambiguous job death: {err}"
 			)))),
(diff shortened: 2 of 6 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4429721

mrcnski:
> Btw, how serious are you with the `SchrodingersCandidate`? I got zero problem with the name. :)) @mrcnski @s0me0ne-unkn0wn

I love that you asked this. 😂 I vote "yes", only question is, should we include the umlaut?

polkadot/node/core/pvf/src/execute/queue.rs
```diff
@@ -361,6 +357,10 @@ fn handle_job_finish(
 			Err(ValidationError::PossiblyInvalid(PossiblyInvalidError::AmbiguousJobDeath(err))),
 			None,
 		),
+		// Either the worker or the job timed out. Kill the worker in either case. Treated as
+		// definitely-invalid, because if we timed out, there's no time left for a retry.
```
mrcnski, line 361: Could be reworded as it's potentially confusing. It is only definitely-invalid in backing where we don't retry the possibly-invalid candidates. In backing there's no time for a retry (and it's stricter).

mrcnski:
Hmm, actually I'm not sure if we should retry, at least with the current timeout which is `DEFAULT_APPROVAL_EXECUTION_TIMEOUT = 12`. That's already a long time given for execution.

One idea is to allow a retry but cut that timeout in half. The idea is that if execution is taking 6 seconds (which is already a long time) then something went wrong (i.e. maybe the worker/job communication broke down).

And speaking of communication, the host/worker comms (note, not worker/job comm) have a really long timeout (like factor of 4 of the default timeout). We shouldn't retry if that fails because there'd just be no time left. But that long timeout needs to remain long, because it's wall clock time, whereas in the job we timeout based on CPU clock time. (That probably sounds really confusing, so see https://github.com/paritytech/polkadot/issues/4217.)

But getting back to cutting the approval timeout in half, I wonder if opens up some attacks, like candidates getting through backing but timing out on some validators with slower machines, leading to disputes. 6 seconds would already be 3x the backing timeout, but I can see it happening. cc @eskimor

eagr:
You have a change of mind now or did I misunderstood your [comment](https://github.com/paritytech/polkadot-sdk/pull/2406#issuecomment-1818810225)?

12s is a really long time. How often does a timeout happen on a valid candidate in practice? If it's extremely rare, I suppose we shouldn't. In that case, I'm happy to close this as it no longer makes sense to switch the category.

mrcnski:
You understood correctly, I just forgot that we had bumped the approval timeout to 12s.

> How often does a timeout happen on a valid candidate in practice?

Not sure, I suppose my concerns are more theoretical in nature. Like if there is a validator that is 3x slower than the rest, it can potentially be exploited to create disputes. I think in practice it's more likely that a timeout happened due to some local conditions, like some deadlock, comm bug, etc... though still not very likely.

Maybe some compromise is possible, like 8s + a retry? Not sure, @eskimor would have a better idea.

eskimor:
Definitely no to shortening the timeout. We have measured fluctuations on the same machine up to a factor of 3. An additional factor of 2 does not seem to be too conservative, given that validators don't run identical machines.

Whether a retry on timeout makes sense is an interesting question:

1. We would be too slow to vote and thus approval voting would escalate because of our no-show. This is an issue, as without the retry we would have raised a dispute - which is a full escalation, but in that case at least someone will pay the bill.
2. Retrying increases the chances of us voting valid on a candidate. If a super majority finds this borderline candidate to be invalid, we just increased our chances of getting slashed significantly. Although if everybody does this, then the chances of a super majority finding the candidate invalid also gets smaller. On top of this with https://github.com/paritytech/polkadot-sdk/issues/635 in place, collectively it does not matter much if 1 or 10 validators voted valid on an invalid candidate.

2 is not much of a reason, but 1 means it is free to trigger an escalation (and break sharding). So retry on timeout is not an option ... in fact if we had the time, I don't see why we would ever retry instead of just increasing the timeout to begin with. What would be the value of retrying, instead of increasing the timeout? If you restart - any work already done (even if little) would go to waste. While with a longer timeout you keep it and have a better success chance, given the same time constraints.

eskimor:
> 12s is a really long time. How often does a timeout happen on a valid candidate in practice? 

It is.

>  If it's extremely rare

We at least hope so. That's the whole point of that long timeout. We actually should increase the noshow timeout because of above reasoning, as execution time alone is not the full story. 

For handling dishonest backers, we need [time disputes](https://github.com/paritytech/polkadot-sdk/issues/742#issuecomment-1691884835).

eskimor:
More to the point, why we really don't want retry on timeout: If backers ignore their 2s timeout, they are already exploiting network resources free of charge. Even, without escalation, by retrying we would then do this twice - doubling the cost.

mrcnski:
> We have measured fluctuations on the same machine up to a factor of 3.

Wow, a 3x fluctuation on the *same machine*? This must be wall clock time? The timeouts in question are CPU time. But still, I agree that we should have a large margin for differently-performing validators.

> in fact if we had the time, I don't see why we would ever retry instead of just increasing the timeout to begin with. What would be the value of retrying, instead of increasing the timeout?

If execution is taking that long, then it could be due to some deadlock etc. Waiting a second and then retrying could avoid some local nondeterminism. Then again, I can't think of any specific internal scenarios that would cause this to occur, except for high load on the machine, in which case retrying could make things worse.

Maybe, there is some way for the node to query whether the execution job is at least *making progress* as opposed to spinning in a loop. But I might be getting into the weeds here.

eagr:
Thanks for the explanation and pointers! @eskimor Those are so informative. Curious, what difference does it make if we no-show for longer time, if any? Say we double the timeout.

You could just close this if there's no reason to proceed. :) @mrcnski

mrcnski:
> If polkavm becomes twice as fast

Looks like it [just happened](https://forum.polkadot.network/t/announcing-polkavm-a-new-risc-v-based-vm-for-smart-contracts-and-possibly-more/3811/33).

mrcnski:
Closing. Thanks for the hustle @eagr and apologies for the spurious suggestion! Will also close the related issue as there's nothing left to do.

mrcnski:
> Maybe, there is some way for the node to query whether the execution job is at least making progress as opposed to spinning in a loop.

I suppose it can be done in PVM by checking if gas is going up. Maybe overly paranoid. I wouldn't do it unless timeouts become a source of issues.

PR #2442, opened 2023-11-22
Fixes cumulus README instructions

gpestana:
README instructions fixes to be compatible with the `polkadot-prepare` and `polkadot-execute` binary split.


cumulus/README.md
```diff
@@ -161,26 +165,26 @@ cargo build --release --bin polkadot
 # Clone
 git clone https://github.com/paritytech/polkadot-sdk
 
-# Compile
+# Compile (under `/polkadot-sdk/cumulus/polkadot-parachain`)
 cargo build --release --bin polkadot-parachain
```
bkontur, line 169:
```suggestion
cargo build --release -p polkadot-parachain-bin
```

cumulus/README.md
```diff
@@ -161,26 +165,26 @@ cargo build --release --bin polkadot
 # Clone
 git clone https://github.com/paritytech/polkadot-sdk
 
-# Compile
+# Compile (under `/polkadot-sdk/cumulus/polkadot-parachain`)
```
bkontur, line 168: I would remove this and better add there:
```
cd ./cumulus/polkadot-parachain`
cargo build --release --bin polkadot-parachain
```

or just use from the top directory:
```
cargo build --release -p polkadot-parachain-bin
```

cumulus/README.md
bkchr: Yeah this is much better!

cumulus/README.md
```diff
@@ -145,6 +145,10 @@ git clone https://github.com/paritytech/polkadot-sdk
 # Compile Polkadot
 cargo build --release --bin polkadot
 
+# Compile Polkadot's `execute` and `prepare` binaries
+cargo build --release --bin polkadot-prepare
+cargo build --release --bin polkadot-execute
```
mrcnski, line 150: Do we need the workers here? I thought they are only needed for validators. BTW, this looks a bit nicer (the parentheses start a subshell so we don't have to `cd ..` after):

```suggestion
(cd polkadot; cargo build --release)
```

bkchr: Yes, because you need to onboard your parachain somewhere.

bkchr:
@gpestana can you please fix the comments

cumulus/README.md
gpestana: Changed, thanks

PR #2447, opened 2023-11-22
CI: Disable runtime upgrade spec name check on Westend Asset Hub and fix Staking pallet migration

liamaharon:
Westend Asset Hub currently failing because the spec name is being changed next runtime upgrade (https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4413125).

This also fixes an idempotency issue with a staking pallet migration. Similar issues will be caught automatically now that we've also updated to try-runtime-cli v0.5.0 which checks for idempotency issues.

This also enables try-state checks running in the CI.


cumulus/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs
```diff
@@ -114,7 +114,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("statemine"),
 	impl_name: create_runtime_str!("statemine"),
 	authoring_version: 1,
-	spec_version: 10000,
+	spec_version: 10001,
```
liamaharon, line 117:
```suggestion
	spec_version: 10000,
```

.gitlab/pipeline/check.yml
```diff
@@ -214,6 +214,7 @@ check-runtime-migration-collectives-westend:
     PACKAGE: "collectives-westend-runtime"
     WASM: "collectives_westend_runtime.compact.compressed.wasm"
     URI: "wss://westend-collectives-rpc.polkadot.io:443"
+    COMMAND_EXTRA_ARGS: "--disable-spec-name-check"
```
bkchr, line 217: This can be removed after the collectives was upgraded?

bkchr (approved):

.gitlab/pipeline/check.yml
liamaharon: Yes

alvicsam (approved):

seadanda (approved):

gpestana (approved):

PR #2450, opened 2023-11-22
Adapt test worker to profile flag

### LINKED ISSUE #2194: PVF: Fix `cargo test -r` building debug workers
`cargo test -r` in `node/core/pvf` builds debug workers, but uses release workers for tests.

## Related

https://github.com/paritytech/polkadot-sdk/pull/2018/

eagr:
closes #2194 

cc @mrcnski

### DIFF
--- a/polkadot/node/core/pvf/build.rs
+++ b/polkadot/node/core/pvf/build.rs
@@ -1,0 +1,21 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+fn main() {
+    if let Ok(profile) = std::env::var("PROFILE") {
+        println!("cargo:rustc-cfg=build_type=\"{}\"", profile);
+    }
+}
--- a/polkadot/node/core/pvf/build.rs
+++ b/polkadot/node/core/pvf/build.rs
@@ -15,7 +15,7 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 fn main() {
-    if let Ok(profile) = std::env::var("PROFILE") {
-        println!("cargo:rustc-cfg=build_type=\"{}\"", profile);
-    }
+	if let Ok(profile) = std::env::var("PROFILE") {
+		println!("cargo:rustc-cfg=build_type=\"{}\"", profile);
+	}
 }
--- a/polkadot/node/core/pvf/build.rs
+++ b/polkadot/node/core/pvf/build.rs
@@ -16,6 +16,6 @@
 
 fn main() {
 	if let Ok(profile) = std::env::var("PROFILE") {
-		println!("cargo:rustc-cfg=build_type=\"{}\"", profile);
+		println!(r#"cargo:rustc-cfg=build_type="{}""#, profile);
 	}
 }
--- a/polkadot/node/core/pvf/src/testing.rs
+++ b/polkadot/node/core/pvf/src/testing.rs
@@ -71,8 +71,9 @@ pub fn build_workers_and_get_paths() -> (PathBuf, PathBuf) {
 			"--bin=polkadot-execute-worker",
 		];
 
-		#[cfg(build_type = "release")]
-		build_args.push("--release");
+		if cfg!(build_type = "release") {
+			build_args.push("--release");
+		}
 
 		let mut cargo = std::process::Command::new("cargo");
 		let cmd = cargo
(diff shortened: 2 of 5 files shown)


polkadot/node/core/pvf/src/testing.rs
```diff
@@ -59,21 +59,21 @@ pub fn validate_candidate(
 ///
 /// NOTE: This should only be called in dev code (tests, benchmarks) as it relies on the relative
 /// paths of the built workers.
-pub fn build_workers_and_get_paths(is_bench: bool) -> (PathBuf, PathBuf) {
+pub fn build_workers_and_get_paths(release_build: bool) -> (PathBuf, PathBuf) {
```
mrcnski, line 62: I believe we can remove this now, right?

And as always, thank you so much for the contribution!

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -117,7 +117,11 @@ pub fn build_workers_and_get_paths(is_bench: bool) -> (PathBuf, PathBuf) {
 			}
 		}
 
-		build_workers(is_bench);
+		// release build
+		#[cfg(not(debug_assertions))]
+		let release_build = true;
```
bkchr, line 122: This is not the correct check.

https://github.com/paritytech/polkadot-sdk/blob/84ddbaf68445fed23a88a086ee37bdcdbcb7356a/substrate/client/consensus/slots/build.rs#L21-L25

with

https://github.com/paritytech/polkadot-sdk/blob/84ddbaf68445fed23a88a086ee37bdcdbcb7356a/substrate/client/consensus/slots/src/lib.rs#L230 (using `release`).

is the proper approach.

bkchr (changes requested):

polkadot/node/core/pvf/src/testing.rs
bkchr: You could also just forward `profile` to the build using `--profile {profile}`

eagr: Nice! I tried tapping PROFILE in the code, not present obviously. Didn't occur to me I should try the build script. Curious, would `#[cfg(not(debug_assertions))]` not work as expected in some case or is it just too hacky?

bkchr: We are for example enable `debug_assertions` in release mode for CI. But your code would assume that it is running in `Debug`.

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -59,21 +59,21 @@ pub fn validate_candidate(
...
 			"--bin=polkadot-prepare-worker",
 			"--bin=polkadot-execute-worker",
 		];
-		if is_bench {
-			// Benches require --release. Regular tests are debug (no flag needed).
-			build_args.push("--release");
-		}
+
+		#[cfg(build_type = "release")]
+		build_args.push("--release");
```
bkchr, line 75:
```suggestion
		build_args.push(format!("--profile={}", env!("BUILD_TYPE")));
```

polkadot/node/core/pvf/build.rs
```diff
@@ -0,0 +1,21 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+fn main() {
+	if let Ok(profile) = std::env::var("PROFILE") {
+		println!(r#"cargo:rustc-cfg=build_type="{}""#, profile);
```
bkchr, line 19:
```suggestion
		println!(r#"cargo:rustc-env=BUILD_TYPE="{}""#, profile);
```

polkadot/node/core/pvf/src/testing.rs
```diff
@@ -59,21 +59,21 @@ pub fn validate_candidate(
...
 			"--package=polkadot",
 			"--bin=polkadot-prepare-worker",
 			"--bin=polkadot-execute-worker",
 		];
-		if is_bench {
-			// Benches require --release. Regular tests are debug (no flag needed).
-			build_args.push("--release");
-		}
+
+		#[cfg(build_type = "release")]
```
bkchr, line 74:
```suggestion
```

bkchr (approved):

mrcnski (approved):

polkadot/node/core/pvf/src/testing.rs
eagr: To go this way, we would need something like

```rs
fn main() {
	if let Ok(mut profile) = std::env::var("PROFILE") {
		if profile == "debug" {
			profile = String::from("dev");
		}
		println!("cargo:rustc-env=BUILD_TYPE={}", profile);
	}
}
```

So maybe just keep it as is?

polkadot/node/core/pvf/src/testing.rs
eagr: Didn't know we could that. :)

polkadot/node/core/pvf/src/testing.rs
bkchr: Yeah fine. I was mainly wondering about the `production` profile, but this is probably just mapped to `release`.

polkadot/node/core/pvf/src/testing.rs
bkchr, line 75:
```suggestion
		if cfg!(build_type = "release") {
			build_args.push("--release");
		}
```

bkchr (approved):

Issue #25, opened 2023-02-07
`GenesisConfig` in a native runtime free world


bkchr:
The `GenesisConfig` is currently one of the only things for that we still require the native runtime. It isn't compiled for wasm right now. We also require the native runtime to load a `GenesisConfig` from a JSON file and as the `GenesisConfig` can change with newer native runtimes we got the raw chain spec that only contains the key/value storage pairs of the genesis block. When we remove the native runtime we need a new way to handle the `GenesisConfig`. 

I propose that we put the `GenesisConfig` into the genesis wasm. This means that we start compiling the `GenesisConfig` for wasm, but I would put this behind some rust feature. So, the normal production builds should not include the `GenesisConfig` anymore. This is done to reduce the size of the wasm binary. I would also keep supporting the serialization of the `GenesisConfig` into JSON. If the `GenesisConfig` is part of the genesis wasm, we have some nice advantages. The first one being that we could keep the JSON serialized `GenesisConfig` in the chain spec and even distribute this chain spec as we would call into the genesis wasm to parse and generate the genesis block. This would require that we move out the genesis wasm from the actual genesis field in the chain spec to some dedicated field from where we can load the wasm binary. Another advantage would be that we could finally make the `chain-spec-builder` work properly. Currently this binary isn't really working for chain specs that aren't build from the `kitchensink-runtime`. The `chain-spec-builder` could be passed some crate name that belongs to some runtime and then it would generate the chain spec for this runtime. We could also enable certain features in the runtime etc.

To build what I outlined above, we would need the following:

- Make `GensisConfig` build in wasm. This will require supporting `serde` in the wasm build to serialize/deserialize the `GenesisConfig` to/from JSON.
- Add a new runtime api `GenesisBuilder` (better name should be found) that can be used to build the genesis config, load it etc.
- Change the chain spec to have a dedicated field for the `code`.
- Conditionally compile `GenesisConfig` into the wasm binary. By default the `GenesisConfig` should probably be compiled into the wasm binary. When doing a on-chain build there should exist a feature to disable the `GenesisConfig`. 

Part of: https://github.com/paritytech/substrate/issues/10579

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/polkadot-release-analysis-v1-2-0/4451/1

JuaniRios:
[...]

PR #2457, opened 2023-11-22
polkadot-parachain: one chain-spec for all

michalkucharczyk:
This PR removes some `ChainSpec` types which are not necessary (left-overs from #1256). Currently `ChainSpec` does not have to be generic over the specific `RuntimeGenesisConfig`, it is enough to use single type for all:
https://github.com/paritytech/polkadot-sdk/blob/9f787018857660440182142adc806954d7d07709/cumulus/polkadot-parachain/src/chain_spec/mod.rs#L53-L54

related to: https://github.com/paritytech/polkadot-sdk/issues/25


cumulus/polkadot-parachain/src/command.rs
```diff
@@ -279,39 +276,9 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
...
-				Runtime::CollectivesWestend => Box::new(
-					chain_spec::collectives::CollectivesWestendChainSpec::from_json_file(path)?,
-				),
-				Runtime::Shell =>
-					Box::new(chain_spec::shell::ShellChainSpec::from_json_file(path)?),
-				Runtime::Seedling =>
-					Box::new(chain_spec::seedling::SeedlingChainSpec::from_json_file(path)?),
-				Runtime::ContractsRococo =>
-					Box::new(chain_spec::contracts::ContractsRococoChainSpec::from_json_file(path)?),
 				Runtime::BridgeHub(bridge_hub_runtime_type) =>
```
michalkucharczyk, line 279: Probably we don't need this too, as Bridges are using the same chain-spec.

michalkucharczyk:
bot fmt

cumulus/polkadot-parachain/src/chain_spec/bridge_hubs.rs
```diff
@@ -72,24 +74,7 @@ impl BridgeHubRuntimeType {
 	pub const ID_PREFIX: &'static str = "bridge-hub";
 
 	pub fn chain_spec_from_json_file(&self, path: PathBuf) -> Result<Box<dyn ChainSpec>, String> {
```
bkchr, line 74: We don't need this function anymore?

cumulus/polkadot-parachain/src/command.rs
```diff
@@ -279,39 +276,9 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
 		path => {
 			let path: PathBuf = path.into();
 			match path.runtime() {
```
bkchr, line 274: Also this match is useless?

cumulus/polkadot-parachain/src/command.rs
bkchr: Haha, I just have commented this :P

cumulus/polkadot-parachain/src/command.rs
michalkucharczyk: [a457c64](https://github.com/paritytech/polkadot-sdk/pull/2457/commits/a457c64a84962fe0e9419947053d9ea3c01b1241)

cumulus/polkadot-parachain/src/chain_spec/bridge_hubs.rs
michalkucharczyk: [a457c64](https://github.com/paritytech/polkadot-sdk/pull/2457/commits/a457c64a84962fe0e9419947053d9ea3c01b1241)

cumulus/polkadot-parachain/src/chain_spec/bridge_hubs.rs
```diff
@@ -197,8 +176,7 @@ pub mod rococo {
 		parachains_common::rococo::currency::EXISTENTIAL_DEPOSIT;
 
 	/// Specialized `ChainSpec` for the normal parachain runtime.
-	pub type BridgeHubChainSpec = sc_service::GenericChainSpec<(), Extensions>;
-
+	pub type BridgeHubChainSpec = GenericChainSpec;
```
seadanda, line 179:
```suggestion
```
I think it could be worth removing this type from each mod and just using `GenericChainSpec` directly. That would also mean in `load_config` instead of using `rococo::BridgeHubChainSpec::from_json_bytes` etc. we just use `GenericChainSpec::from_json_bytes` for all live chains.

seadanda (commented):
Seems like a nice simplification. I think the `BridgeHubChainSpec` types can also be removed to improve it further.

cumulus/polkadot-parachain/src/chain_spec/bridge_hubs.rs
michalkucharczyk: yeah, happy to do that. I was not sure if removal of `BridgeHubChainSpec` is also preferred.

michalkucharczyk: done in: 12ad76b

bkchr (approved):

PR #2459, opened 2023-11-22
[NPoS] Use EraInfo to manipulate exposure in fast-unstake tests

Ank4n:
The FastUnstake pallet tests were previously directly modifying storage items in the pallet-staking to alter exposure. However, due to the introduction of the [new paged exposure feature](https://github.com/paritytech/polkadot-sdk/pull/1189) these tests were not testing against correct storage items. This issue resulted in a bug that I didn't catch, which has been addressed in [this fix](https://github.com/paritytech/polkadot-sdk/pull/2369).

This PR introduces a modification to how the pallet-fast-unstake handles exposure. It now utilizes `pallet-staking::EraInfo` to set or mutate Exposures.

### DIFF
--- a/substrate/frame/fast-unstake/src/tests.rs
+++ b/substrate/frame/fast-unstake/src/tests.rs
@@ -788,10 +788,12 @@ mod on_idle {
 			assert_ok!(FastUnstake::register_fast_unstake(RuntimeOrigin::signed(VALIDATOR_PREFIX)));
 
 			// but they indeed are exposed!
-			assert!(pallet_staking::ErasStakers::<T>::contains_key(
+			assert!(pallet_staking::EraInfo::<T>::get_paged_exposure(
 				BondingDuration::get() - 1,
-				VALIDATOR_PREFIX
-			));
+				&VALIDATOR_PREFIX,
+				0
+			)
+			.is_some());
 
 			// process a block, this validator is exposed and has been slashed.
 			next_block(true);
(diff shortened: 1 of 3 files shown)


substrate/frame/fast-unstake/src/tests.rs
```diff
@@ -788,10 +788,12 @@ mod on_idle {
...
 			// but they indeed are exposed!
-			assert!(pallet_staking::ErasStakers::<T>::contains_key(
+			assert!(pallet_staking::EraInfo::<T>::get_paged_exposure(
 				BondingDuration::get() - 1,
-				VALIDATOR_PREFIX
-			));
+				&VALIDATOR_PREFIX,
+				0
+			)
+			.is_some());
```
gpestana, line 796: It seems to be that probably we could have prevented this issue if instead of an `assert!` there was an `assert_eq!` that would check if the exposure is correct. Maybe a good time to add that?

gpestana (approved):

substrate/frame/fast-unstake/src/tests.rs
Ank4n: What do you mean? Checking the exposure?

gpestana: >  Checking the exposure?
yes, the return value of `fn get_paged_exposure`

georgepisaltu (approved):
LGTM

rossbulat (approved):

PR #2462, opened 2023-11-23
relay-chain-consensus: set a fork_choice

michalkucharczyk:
After #2001 the `fork_choice` strategy may remain uninitialized in the block import pipeline which uses relay-chain verifier.
Refer to https://github.com/paritytech/polkadot-sdk/pull/2430#issuecomment-1823979813 for some further discussion.

### DIFF
--- a/cumulus/client/consensus/relay-chain/src/import_queue.rs
+++ b/cumulus/client/consensus/relay-chain/src/import_queue.rs
@@ -31,7 +31,6 @@ use sp_inherents::{CreateInherentDataProviders, InherentDataProvider};
 use sp_runtime::traits::{Block as BlockT, Header as HeaderT};
 
 /// A verifier that just checks the inherents.
-
 pub struct Verifier<Client, Block, CIDP> {
 	client: Arc<Client>,
 	create_inherent_data_providers: CIDP,
--- a/cumulus/client/consensus/relay-chain/src/import_queue.rs
+++ b/cumulus/client/consensus/relay-chain/src/import_queue.rs
@@ -19,7 +19,6 @@ use std::{marker::PhantomData, sync::Arc};
 use cumulus_client_consensus_common::ParachainBlockImportMarker;
 
 use sc_consensus::{
-	block_import::ForkChoiceStrategy,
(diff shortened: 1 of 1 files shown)


cumulus/client/consensus/relay-chain/src/import_queue.rs
```diff
@@ -30,6 +31,7 @@ use sp_inherents::{CreateInherentDataProviders, InherentDataProvider};
 use sp_runtime::traits::{Block as BlockT, Header as HeaderT};
 
 /// A verifier that just checks the inherents.
+
```
michalkucharczyk, line 34:
```suggestion

```

cumulus/client/consensus/relay-chain/src/import_queue.rs
```diff
@@ -55,6 +56,8 @@ where
 		&mut self,
 		mut block_params: BlockImportParams<Block>,
 	) -> Result<BlockImportParams<Block>, String> {
+		block_params.fork_choice = Some(ForkChoiceStrategy::LongestChain);
```
michalkucharczyk, line 59: We could also set `Custom(true)` here. We had some discussion with @skunert  about this, but we don't have final conclusion on this.

`Custom(true)` would trigger more maintenance in transaction pool, making it up-to-date for all forks, but I don't know if building blocks would be actually synced with that state.

skunert: I mean this is the relay chain verifier, afaik that is pretty  niche anyway and mostly for historic blocks during sync. I think either is fine here :thinking:

bkchr: Please just bring back the old logic with "block_params.source == InitialSync`.

paritytech-cicd-pr:
**Job name**: check-runtime-migration-asset-hub-westend
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4457200

skunert (approved):

bkchr (approved):

PR #2463, opened 2023-11-23
Add `on-chain-release-build` feature for Collectives Westend

liamaharon:
Collectives Westend is missing an `on-chain-release-build` feature flag.


seadanda (approved):
LGTM, must have missed that addition in the other runtimes in my numerous updates from `master`!
Looks like you need to run zepter --fix though, as there are some whitespace issues in the `toml`.

bkchr (approved):

PR #1524, opened 2023-09-12
Modular block request handler

rahulksnv:
Submit the outstanding PRs from the old repos(these were already reviewed and approved before the repo rorg, but not yet submitted): 
Main PR: https://github.com/paritytech/substrate/pull/14014
Companion PRs: https://github.com/paritytech/polkadot/pull/7134, https://github.com/paritytech/cumulus/pull/2489

The changes in the PR:
1. ChainSync currently calls into the block request handler directly. Instead, move the block request handler behind a trait. This allows new protocols to be plugged into ChainSync.
2. BuildNetworkParams is changed so that custom relay protocol implementations can be (optionally) passed in during network creation time. If custom protocol is not specified, it defaults to the existing block handler
3. BlockServer and BlockDownloader traits are introduced for the protocol implementation. The existing block handler has been changed to implement these traits
4. Other changes:
[X] Make TxHash serializable. This is needed for exchanging the serialized hash in the relay protocol messages
[X] Clean up types no longer used(OpaqueBlockRequest, OpaqueBlockResponse)

PR #2467, opened 2023-11-23
Extract warp sync strategy from `ChainSync`

dmitry-markin:
Extract `WarpSync` (and `StateSync` as part of warp sync) from `ChainSync` as independent syncing strategy called by `SyncingEngine`. Introduce `SyncingStrategy` enum as a proxy between `SyncingEngine` and specific syncing strategies.

## Limitations
Gap sync is kept in `ChainSync` for now because it shares the same set of peers as block syncing implementation in `ChainSync`. Extraction of a common context responsible for peer management in syncing strategies able to run in parallel is planned for a follow-up PR.

## Further improvements
A possibility of conversion of `SyncingStartegy` into a trait should be evaluated. The main stopper for this is that different strategies need to communicate different actions to `SyncingEngine` and respond to different events / provide different APIs (e.g., requesting justifications is only possible via `ChainSync` and not through `WarpSync`; `SendWarpProofRequest` action is only relevant to `WarpSync`, etc.)


substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
-					fields: BlockAttributes::HEADER |
-						BlockAttributes::BODY | BlockAttributes::JUSTIFICATION,
-					from: FromBlock::Hash(header.hash()),
-					direction: Direction::Ascending,
-					max: Some(1),
-				};
-				Some((*header.number(), request))
-			},
+		let Phase::TargetBlock(header) = &mut self.phase else {
+			debug!(target: "sync", "Unexpected target block response from {peer_id}");
```
altonen, line 402:
```suggestion
			debug!(target: LOG_TARGET, "Unexpected target block response from {peer_id}");
```

substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
+			return Err(BadPeer(peer_id, rep::NOT_REQUESTED))
+		}
+
+		validate_blocks::<B>(&blocks, &peer_id, Some(request))?;
+
+		let block = blocks.pop().expect("`blocks` len checked above; qed");
+
+		let Some(block_header) = &block.header else {
+			debug!(
+				target: "sync",
```
altonen, line 429:
```suggestion
				target: LOG_TARGET,
```

substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
+				"Downloading target block failed: missing header in response from {peer_id}.",
+			);
+			return Err(BadPeer(peer_id, rep::VERIFICATION_FAIL))
+		};
+
+		if block_header != header {
+			debug!(
+				target: "sync",
```
altonen, line 437:
```suggestion
				target: LOG_TARGET,
```

substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
+				"Downloading target block failed: different header in response from {peer_id}.",
+			);
+			return Err(BadPeer(peer_id, rep::VERIFICATION_FAIL))
 		}
+
+		if block.body.is_none() {
+			debug!(
+				target: "sync",
```
altonen, line 445:
```suggestion
				target: LOG_TARGET,
```

substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
+			.iter()
+			.any(|(_, peer)| matches!(peer.state, PeerState::DownloadingProofs))
+		{
+			// Only one warp proof request at a time is possible.
+			return None
 		}
+
+		let Some((peer_id, peer)) = select_synced_available_peer(&mut self.peers, None) else {
+			return None
+		};
```
altonen, line 479:
```suggestion
		let (peer_id, peer) = select_synced_available_peer(&mut self.peers, None)?;
```

altonen: I think the borrowing issue can be worked around the similar way as in state sync

substrate/client/network/sync/src/warp.rs
```diff
@@ -232,154 +303,242 @@ where
...
+
+		// Cut the borrowing tie.
+		let target_hash = target_header.hash();
+		let target_number = *target_header.number();
+
+		let Some((peer_id, peer)) =
+			select_synced_available_peer(&mut self.peers, Some(target_number))
+		else {
+			return None
+		};
```
altonen, line 508:
```suggestion
		let (peer_id, peer) = select_synced_available_peer(&mut self.peers, Some(target_number))?;
```

substrate/client/network/sync/src/warp.rs
```diff
@@ -392,14 +551,86 @@ where
...
-				},
-				total_bytes: self.total_proof_bytes + sync.progress().size,
+			Phase::Complete => WarpSyncProgress {
+				phase: WarpSyncPhase::Complete,
+				total_bytes: self.total_proof_bytes + self.total_state_bytes,
 			},
 		}
 	}
+
+	/// Get the number of peers known to syncing.
```
altonen, line 561:
```suggestion
	/// Get the number of peers known to warp sync.
```

altonen (commented):
Very nice work! I left a bunch of comments but apart from state sync still partly being in `ChainSync` and the inability to share `BlocksCollection` between strategies (which can be addressed in a follow-up), I think it's looking good

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,463 @@
...
+		match self {
+			SyncingStrategy::WarpSyncStrategy(_) => 0,
+			SyncingStrategy::StateSyncStrategy(_) => 0,
+			SyncingStrategy::ChainSyncStrategy(strategy) => strategy.num_sync_requests(),
+		}
+	}
+
+	/// Get actions that should be performed by the owner on the strategy's behalf
+	#[must_use]
+	pub fn actions(&mut self) -> Box<dyn Iterator<Item = SyncingAction<B>>> {
```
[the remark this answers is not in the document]
dmitry-markin: Yes, we can't return just `impl Iterator` because different strategies return different `impl Iterator`'s, and even though we apply `.map()` to them to convert to the iterator of `StrategyAction`'s, these different mapped iterators remember what was the underlying iterator type and are essentially different types.

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -240,14 +243,6 @@ pub struct ChainSync<B: BlockT, Client> {
 	downloaded_blocks: usize,
 	/// State sync in progress, if any.
 	state_sync: Option<StateSync<B, Client>>,
```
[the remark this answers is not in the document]
dmitry-markin: Because of the "fast" aka "light state" sync. `ChainSync` is responsible for it, and after downloading block history (without executing blocks) it does state sync and then continues with normal block sync (executing blocks). It should be possible to refactor it in a way that we switch the startegies in a following way:
1. `ChainSync` without executing blocks
2. `StateSync`
3. "Normal" `ChainSync`

But I'm not sure if it makes sense, considering it's mostly `ChainSync` with a small period of `StateSync`.
We can revisit this in a follow-up though.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,463 @@
...
+		&mut self,
+		is_best: bool,
+		peer_id: PeerId,
+		announce: &BlockAnnounce<B::Header>,
+	) -> Option<(B::Hash, NumberFor<B>)> {
+		match self {
+			SyncingStrategy::WarpSyncStrategy(_) =>
+				Some((announce.header.hash(), *announce.header.number())),
+			SyncingStrategy::StateSyncStrategy(_) =>
+				Some((announce.header.hash(), *announce.header.number())),
```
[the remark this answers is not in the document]
dmitry-markin: Block announce validation happens in `SyncingEngine`, so it's already validated. The only exception is `ChainSync` where we do not update best hash & number if we know that the peer is in the ancestry search state.

Otherwise, updating best hash & number upon processing a validated block announcement seems not worse than updating it upon receiving a handshake, which also says nothing about what fork the peer is at.

dmitry-markin:
> Very nice work! I left a bunch of comments but apart from state sync still partly being in `ChainSync` and the inability to share `BlocksCollection` between strategies (which can be addressed in a follow-up), I think it's looking good

Another thing that should be shared between the strategies running in parallel similarly to `BlocksCollection` is the set of peers and their states. Unless we accept that it's OK to treat peers independently in different strategies and issue parallel requests to them.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4543457

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4543594

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4544158

substrate/client/network/sync/src/strategy.rs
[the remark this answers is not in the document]
altonen: Fair enough

altonen:
> Unless we accept that it's OK to treat peers independently in different strategies and issue parallel requests to them.

I don't think it's a good idea. All strategies should have the same view of connected peers. We can't have two simultaneously in-flight request to any peer so their states have to be synchronized across strategies.

If we introduce some kind of peer + blocks collection which is shared between strategies using `Arc<Mutex<Context>>` (and maybe some nice API on top of it), that would probably work.

substrate/client/network/sync/src/chain_sync.rs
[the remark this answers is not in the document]
altonen: Yes I think we have to think about this a little so a follow-up is a good idea

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4562032

substrate/client/network/sync/src/state_strategy.rs
```diff
@@ -0,0 +1,321 @@
...
+						self.state_sync.progress().size / (1024 * 1024),
+					);
+				},
+				false => {
+					error!(
+						target: LOG_TARGET,
+						"State sync failed. Falling back to full sync.",
+					);
+					// TODO: Test this scenario.
+					//       Make sure fail-over doesn't render useless warp sync tests.
```
dmitry-markin, line 234: I'll leave a comment here to ensure I don't forget to address the TODO. With `ChainSync` started even if warp sync has failed, the `warp_sync` test in `sc-network-test` might not detect the warp sync issues, because the synchronization will be performed by `ChainSync` anyway.

dmitry-markin: `sync::warp_sync` tests that we don't have a state for block 1 (target = 64), i.e. the synchronization wasn't done by full sync. Therefore not an issue.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,463 @@
...
+								.map(|(peer_id, _best_hash, best_number)| (peer_id, best_number)),
+						);
+
+						*self = Self::StateSyncStrategy(state_sync);
+					},
+					None => {
+						error!(
+							target: LOG_TARGET,
+							"Warp sync failed. Continuing with full sync."
+						);
```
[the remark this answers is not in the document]
dmitry-markin: The only case when this is currently triggered is when there is a finalized state in the DB and the warp sync is not started at all, being replaced by `ChainSync` from the very beginning. I've added a test for this scenario here: https://github.com/paritytech/polkadot-sdk/blob/3ad88a4628dfc53e68c0774d0ea09c31d210ce70/substrate/client/network/test/src/sync.rs#L1254-L1281

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4640056

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4685220

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4686595

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4686594

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4691966

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4695684

substrate/client/network/sync/src/state_strategy.rs
```diff
@@ -0,0 +1,330 @@
...
+				);
+			} else {
+				error!(
+					target: LOG_TARGET,
+					"State sync failed. Falling back to full sync.",
+				);
+			}
+
+			self.actions.push(StateStrategyAction::Finished);
+		}
```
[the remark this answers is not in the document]
dmitry-markin: Done, test added in https://github.com/paritytech/polkadot-sdk/pull/2467/commits/c0115cfefa7ecb35b2e919c803f3775871af128d.

substrate/client/network/sync/src/state_strategy.rs
```diff
@@ -0,0 +1,330 @@
...
+	pub const BAD_RESPONSE: Rep = Rep::new(-(1 << 12), "Incomplete response");
+
+	/// Reputation change for peers which send us a known bad state.
+	pub const BAD_STATE: Rep = Rep::new(-(1 << 29), "Bad state");
+}
+
+/// Action that should be performed on [`StateStrategy`]'s behalf.
+pub enum StateStrategyAction<B: BlockT> {
+	/// Send state request to peer.
+	SendStateRequest { peer_id: PeerId, request: OpaqueStateRequest },
```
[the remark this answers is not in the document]
dmitry-markin: Overall sounds good, but one thing that limits us is the `BlockDownloader` trait introduced in https://github.com/paritytech/polkadot-sdk/pull/1524. When we replace all requests by `RequestType` and `payload: Vec<u8>`, I don't think there will be an easy way to plug custom block downloading implementations.

One option that might work is to make  `BlockDownloader` run on top of requests. I.e., it generates opaque vector of bytes to send over network, and handles the opaque bytes received as a response. But this will essentially limit it to substrate request-response protocols.

@rahulksnv will this work for you, or you are using some custom protocol not based on substrate request-response protocols?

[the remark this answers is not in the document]
nazar-pc: Currently we are using `BlockDownloader` to send custom messages instead of what Substrate sends by default to implement something very similar to [Compact Blocks (BIP 0152)](https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki) and with plans to implement [Graphene](https://cryptoeconomics.cs.umass.edu/graphene/) at some point in the future.

> One option that might work is to make BlockDownloader run on top of requests. I.e., it generates opaque vector of bytes to send over network, and handles the opaque bytes received as a response

IIRC this is basically how it works already.

@rahulksnv can you confirm what I just wrote?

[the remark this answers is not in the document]
dmitry-markin: If you are fetching transactions from "mempool", downloading the missing ones, do I understand correctly that `BlockDownloader::download_blocks` doesn't necessarily map to one network request? I.e., it can perform multiple requests under the hood, accumulating the data received and returning blocks _as though_ they were downloaded in a single request?

What we are planning to do is to abstract network requests in syncing so that custom syncing protocols ("strategies") can just ask networking to perform a request on their behalf.  The networking/`SyncingEngine` is going to know nothing about the request and treat it only as a pair of protocol name and payload.

There is going to be a way to run syncing strategies in parallel (this is needed to introduce Sync 2.0 not dropping Sync 1.0), but it still serves a different purpose than a custom block downloader you are using, which plugs into standard syncing strategies.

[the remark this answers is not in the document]
rahulksnv: > if you are fetching transactions/extrinsics from "mempool", downloading the missing ones, do I understand correctly that BlockDownloader::download_blocks doesn't necessarily map to one network request? I.e., it can perform multiple requests under the hood, accumulating the data received and returning blocks as though they were downloaded in a single request?

Yes, one request can lead to multiple handshake messages between the client/server. The block downloader can be seen as a leaf level/bottom layer helper for fetching a block, using a custom protocol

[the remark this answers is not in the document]
dmitry-markin: > Yes, one request can lead to multiple handshake messages between the client/server.

Are these handshake messages implemented using standard substrate request-response protocol, or they are a completely custom implementation? I mean on the application layer?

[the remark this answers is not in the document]
rahulksnv: we currently use the request-response protocol(with protocol specific endpoint) + SCALE encoding. But this is more of an implementation detail. Some other protocol implementation could choose a different mechanism as it sees best

substrate/client/network/sync/src/chain_sync.rs
```diff
@@ -407,20 +387,10 @@ where
...
-				},
-				total_bytes: 0,
-			}),
-			(Some(sync), _, _) => Some(sync.progress()),
-			_ => None,
-		};
+		let warp_sync_progress = self.gap_sync.as_ref().map(|gap_sync| WarpSyncProgress {
+			phase: WarpSyncPhase::DownloadingBlocks(gap_sync.best_queued_number),
+			total_bytes: 0,
+		});
```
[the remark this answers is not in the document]
dmitry-markin: No sharing of `BlockCollection` yet as it's not needed to run `GapSync` in parallel with `ChainSync` (it has its own instance of `BlockCollection`), but here is a draft PR with the initial implementation of sharing of peers between the strategies: https://github.com/paritytech/polkadot-sdk/pull/2814.

I don't think time should be invested now into reviewing it as a lot could change, but hopefully it shows the general idea of what a follow-up PR may look like.

substrate/client/network/test/src/sync.rs
```diff
@@ -1249,6 +1251,35 @@ async fn warp_sync() {
 	.await;
 }
 
+/// If there is a finalized state in the DB, warp sync falls back to full sync.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
altonen, line 1255: Doesn't work with the default config?

altonen (approved):

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,484 @@
...
+		match self {
+			SyncingStrategy::WarpSyncStrategy(_) => {},
+			SyncingStrategy::StateSyncStrategy(_) => {},
+			SyncingStrategy::ChainSyncStrategy(strategy) =>
+				strategy.on_justification_import(hash, number, success),
+		}
+	}
+
+	/// Process block response.
+	pub fn on_block_response(
```
altonen, line 224: One thing I forgot to mention was that in a follow-up we should check if these `on_*_response()` could be refactored into `on_response(PeerId, ProtocolName, request: Vec<u8>, response: Vec<u8>)`.

dmitry-markin: Yeah, we have discussed this and there is this `BlockDownloader` that stands in the way. See the the discussion with Subspace people above

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,484 @@
...
+							client,
+							res.target_header,
+							res.target_body,
+							res.target_justifications,
+							false,
+							connected_peers
+								.map(|(peer_id, _best_hash, best_number)| (peer_id, best_number)),
+						);
+
+						*self = Self::StateSyncStrategy(state_sync);
```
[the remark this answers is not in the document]
dmitry-markin: In this PR only by the logic of the strategies — both `WarpSync` and `StateSync` generate one request at a time and can only finish when the response is received.

substrate/client/network/test/src/sync.rs
dmitry-markin: To be honest, I just put the same config as in other tests in this file to be consistent. It does work with the single-threaded runtime though.

substrate/client/network/sync/src/strategy/chain_sync/test.rs
skunert, line 1: silly nit: Its a bit strance that there is only the `test.rs` file in this `chain_sync` folder, maybe `chain_sync.rs` should be moved here too?

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,485 @@
...
+use std::sync::Arc;
+use warp::{EncodedProof, WarpProofRequest, WarpSync, WarpSyncAction, WarpSyncConfig};
+
+/// Corresponding `ChainSync` mode.
+fn chain_sync_mode(sync_mode: SyncMode) -> ChainSyncMode {
+	match sync_mode {
+		SyncMode::Full => ChainSyncMode::Full,
+		SyncMode::LightState { skip_proofs, storage_chain_mode } =>
+			ChainSyncMode::LightState { skip_proofs, storage_chain_mode },
+		SyncMode::Warp => ChainSyncMode::Full,
```
skunert, line 57: Do we want this branch? Should never be taken right?

dmitry-markin: Actually, it's a legit branch that's triggered: this is the mode `ChainSync` is started in after warp sync is over.

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1089,19 +1059,21 @@ where
 			peer
 		} else {
 			error!(target: LOG_TARGET, "💔 Called `on_validated_block_announce` with a bad peer ID");
-			return
+			return Some((hash, number))
```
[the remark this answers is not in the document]
dmitry-markin: It's a logic error to get here in the first place, but, if for some reason `ChainSync` is not aware of the peer, we still want to update it's best hash/number in `SyncingEngine`. May be the method comment should be updated into something like "returns best hash/number that we consider the peer to be at after receiving this block announce validation".

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -516,7 +531,7 @@ where
 				// If we are at genesis, just start downloading.
 				let (state, req) = if self.best_queued_number.is_zero() {
```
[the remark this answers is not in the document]
dmitry-markin: Best queued number is considered to be the common number in this case https://github.com/paritytech/polkadot-sdk/blob/eeb14c3334efdd72bb9313873bd52a7a19955049/substrate/client/network/sync/src/strategy/chain_sync.rs#L522
and `BlockCollection` will request for download the first different block, i.e. block 1. So, the genesis block will not be downloaded.

substrate/client/network/sync/src/strategy/chain_sync.rs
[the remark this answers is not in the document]
dmitry-markin: Duplicate comment, answered below.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,485 @@
...
+			SyncingStrategy::StateSyncStrategy(_) => {},
+			SyncingStrategy::ChainSyncStrategy(strategy) => strategy.report_metrics(),
+		}
+	}
+
+	/// Get actions that should be performed by the owner on the strategy's behalf
+	#[must_use]
+	pub fn actions(&mut self) -> Box<dyn Iterator<Item = SyncingAction<B>>> {
+		match self {
+			SyncingStrategy::WarpSyncStrategy(strategy) =>
```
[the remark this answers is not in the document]
dmitry-markin: Good point. I'll do it in a follow-up PR https://github.com/paritytech/polkadot-sdk/pull/2814 to not complicate rebasing though)

skunert (approved):

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -0,0 +1,758 @@
...
+	/// Produce state request.
+	fn state_request(&mut self) -> Option<(PeerId, OpaqueStateRequest)> {
+		if self.state_sync.is_complete() {
+			return None
+		}
+
+		if self
+			.peers
+			.iter()
+			.any(|(_, peer)| matches!(peer.state, PeerState::DownloadingState))
```
michalkucharczyk, line 287: nit: how about?
`if !self.peers.values().all(Peer::is_available)`

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -0,0 +1,758 @@
...
+		assert!(matches!(
+			&state_strategy.actions[0],
+			StateStrategyAction::ImportBlocks { origin, blocks }
+				if *origin == expected_origin && *blocks == expected_blocks,
+		));
+	}
+
+	#[test]
+	fn importing_unknown_block_doesnt_finish_strategy() {
+		let target_hash = Hash::random();
```
michalkucharczyk, line 658: Maybe you could used fixed value here?
Just in case logs from different executions need to be compared?
Same applies to other random values.

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -0,0 +1,758 @@
...
+				Ok(BlockImportStatus::ImportedUnknown(1, ImportedAux::default(), None)),
+				target_hash,
+			)],
+		);
+
+		// Strategy finishes.
+		let actions = state_strategy.actions().collect::<Vec<_>>();
+		assert_eq!(actions.len(), 1);
+		assert!(matches!(&actions[0], StateStrategyAction::Finished));
+
```
michalkucharczyk, line 754: dumb question: what would be expected result if I've called `on_block_processed` one more time here?

michalkucharczyk (approved):

michalkucharczyk:
lgtm, left some nits.

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -0,0 +1,758 @@
...
+						"Unexpected block processed: {hash} with result {result:?}.",
+					);
+					None
+				}
+			})
+			.collect::<Vec<_>>();
+
+		if !results.is_empty() {
+			// We processed the target block
+			results.iter().filter_map(|result| result.as_ref().err()).for_each(|e| {
```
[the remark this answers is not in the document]
dmitry-markin: I'm not sure that this can be done. We also need to finish the strategy if the target block import has failed, but if we filter out unsuccessful results we won't be able to distinguish between the following situations:
1. No target block was processed
2. Target block processing has failed

Am I missing something?

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: I'd like to keep the check for `PeerState::DownloadingState` for consistency with other strategies, as other peer states can be allowed should we introduce them in the future. But thanks for the hint that we can use `values()` instead of `iter()` here.

substrate/client/network/sync/src/strategy/warp.rs
```diff
@@ -0,0 +1,1440 @@
...
+	/// Notify that a new peer has connected.
+	pub fn add_peer(&mut self, peer_id: PeerId, _best_hash: B::Hash, best_number: NumberFor<B>) {
+		self.peers.insert(peer_id, Peer { best_number, state: PeerState::Available });
+
+		self.try_to_start_warp_sync();
+	}
+
+	/// Notify that a peer has disconnected.
+	pub fn remove_peer(&mut self, peer_id: &PeerId) {
+		self.peers.remove(peer_id);
```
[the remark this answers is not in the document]
dmitry-markin: I don't think so. Not even sure what `MIN_PEERS_TO_START_WARP_SYNC` should safeguard against, but kept it as it was there in the original implementation of warp sync.

substrate/client/network/sync/src/strategy/warp.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+		let target_block_request = self
+			.target_block_request()
+			.into_iter()
+			.map(|(peer_id, request)| WarpSyncAction::SendBlockRequest { peer_id, request });
+		self.actions.extend(target_block_request);
+
+		std::mem::take(&mut self.actions).into_iter()
+	}
+
```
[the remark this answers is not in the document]
dmitry-markin: Thanks, actually the docs were also missing.

substrate/client/network/sync/src/strategy/state.rs
michalkucharczyk: `any(Peer::is_downloading)` would also be nice. it is minor however, up to you...

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: `StateStrategyAction::Finished` might be generated one more time, but this won't change much. Also, it shouldn't happen in the wild as we request/import the target state only once.

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: Yeah, I also don't really like the undetermined random numbers in tests. But constructing predefined values, especially when we init 10 peers, would be real PITA.

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: So far have decided to keep `matches!` to not introduce the `is_...()` for every enum variant...

substrate/client/network/sync/src/strategy/state.rs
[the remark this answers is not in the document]
michalkucharczyk: Yeah, my mistake, but we could `filter_map` first loop into bool, and do logging on the go in the single pass. Again it is minor :),  feel free to ignore this.

substrate/client/network/sync/src/state_strategy.rs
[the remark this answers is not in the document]
altonen: I don't see why `BlockDownloader` or the entire compact block implementation couldn't be expressed as a strategy. `BlockDownloader::download_block()` could be a signal akin to `SyncingService::request_justification()` which could potentially allow us to generalize that system. But that's a future task either way and not in the scope of this PR

substrate/client/network/sync/src/strategy/chain_sync/test.rs
dmitry-markin: Personally I don't like `mod.rs` filenames because it's easier to find something called `chain_sync.rs`. So, going to keep it as is)

substrate/client/network/sync/src/strategy.rs
```diff
@@ -0,0 +1,489 @@
...
+use sp_runtime::{
+	traits::{Block as BlockT, NumberFor},
+	Justifications,
+};
+use state::{StateStrategy, StateStrategyAction};
+use std::sync::Arc;
+use warp::{EncodedProof, WarpProofRequest, WarpSync, WarpSyncAction, WarpSyncConfig};
+
+/// Corresponding `ChainSync` mode.
+fn chain_sync_mode(sync_mode: SyncMode) -> ChainSyncMode {
```
nazar-pc, line 52: Is this function really necessary? It reverted https://github.com/paritytech/substrate/pull/14465 and the fact that there are two types again makes it painful to backport https://github.com/paritytech/substrate/pull/14482 (whose goal is to allow combining Substrate's sync with a custom sync mechanism we have in Subspace).

Can this be replaced with a method on `SyncMode` that returns `true` for both `SyncMode::Full` and `SyncMode::Warp`? Or match enum against two possible variants explicitly.

dmitry-markin: What this and upcoming PRs aim to achieve is to clearly separate syncing strategies like `Warp`, `State`, `ChainSync`, etc. Introducing an atomic variable accessible by `ChainSync` that refers to warp sync would be going into the opposite direction.

On the other hand, once the refactoring is over, it should be possible to plug custom sync as a separate strategy and switch to `ChainSync` only when the initial sync is over, without the need to pause `ChainSync`. May be this will suite you?

PR #2475, opened 2023-11-23
 Disable any peer connections for parachain nodes in pov-recovery zombienet test

skunert:
I noticed that this test broke at some point. The parachain nodes should only acquire their blocks from the relay chain. But they were connecting to their peers and started fetching blocks from there.

In this test I now take additional measures so we check that each nodes really uses pov-recovery to get the blocks.

### DIFF
--- a/cumulus/zombienet/tests/0002-pov_recovery.toml
+++ b/cumulus/zombienet/tests/0002-pov_recovery.toml
@@ -48,7 +48,7 @@ add_to_genesis = false
   validator = false # full node
   image = "{{COL_IMAGE}}"
   command = "test-parachain"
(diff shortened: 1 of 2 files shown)


pepoviola (approved):

michalkucharczyk (approved):

michalkucharczyk:
In zdnsl we could also check some metrics to assert peer count is actually zero.

altonen:
How was the the test failing? If peer counts are set to 0 in tests, it doesn't fix the underlying issue, whatever it is, and it could happen in a real network, right?

skunert:
> How was the the test failing? If peer counts are set to 0 in tests, it doesn't fix the underlying issue, whatever it is, and it could happen in a real network, right?

So in this test we disable block announcements to prevent that the nodes send blocks to each other. But what I see happening that during the test nodes connect to new peers and report their best block. This way the nodes learn that the collators have a higher best-block (since their timeout for pov-recovery is lower they recover first). Then the full-nodes start requesting blocks from the collators. So here we really want to run all nodes in isolation.

At least that is my interpretation of the logs. Does that make sense @altonen? It is more a problem of this specific test setup, the node behaviour seems to be fine.

altonen:
I think the test is still broken in some way because it's quite flaky and has failed for a few pull requests that have been merged today. Are you able to reproduce this locally?

PR #2477, opened 2023-11-24
Use clone instead of fork on pvf

### LINKED ISSUE #2162: PVF worker: consider `clone` instead of `fork`
I think `fork` is okay for now, but in a follow-up we should replace it with `clone` and pass in some arguments for sandboxing, as seen here:

https://github.com/koute/polkavm/blob/11beebd06276ce9b84f335350138479e714f6caf/crates/polkavm/src/sandbox/linux.rs#L711

_Originally posted by @mrcnski in https://github.com/paritytech/polkadot-sdk/pull/1685#discussion_r1368326166_

Jpserrat:
@mrcnski Done the change on the prepare worker, once the prepare worker part is good I'll do the same for the execute worker.

This is based on https://github.com/koute/polkavm/blob/11beebd06276ce9b84f335350138479e714f6caf/crates/polkavm/src/sandbox/linux.rs#L711.

## TODO

- [x] Add a check for this capability at startup
- [x] Add prdoc mentioning the new Secure Validator Mode (optional) requirement.

## Related

Closes #2162


polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,50 +232,97 @@ pub fn worker_entrypoint(
...
+						| CloneFlags::CLONE_NEWNS
+						| CloneFlags::CLONE_NEWPID
+						| CloneFlags::CLONE_NEWUSER
+						| CloneFlags::CLONE_NEWUTS
+						| CloneFlags::from_bits_retain(libc::SIGCHLD);
+
+						// SAFETY: new process is spawned within a single threaded process. This invariant
+						// is enforced by tests. Stack size being specified to ensure child doesn't overflow
+						let result = match unsafe { nix::sched::clone(
+							Box::new(|| {
```
Jpserrat, line 252: Here I couldn't find a way to close the pipe reader and the UNIX stream. Not sure if it is needed.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4474348

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4485823

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4485824

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4489459

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4489458

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,31 +234,74 @@ pub fn worker_entrypoint(
...
-							executor_params,
-						)
-					},
-					Ok(ForkResult::Parent { child }) => {
-						// the read end will wait until all write ends have been closed,
-						// this drop is necessary to avoid deadlock
-						drop(pipe_writer);
+				cfg_if::cfg_if! {
+					if #[cfg(any(target_os = "linux"))] {
+						let stack = 1024 * 1024;
```
mrcnski, line 239: This is not enough, it should be at least the native stack limit (see `DEFAULT_CONFIG`). I think we can use `EXECUTE_THREAD_STACK_SIZE` here instead.

And now that I see it, `EXECUTE_THREAD_STACK_SIZE` doesn't correctly get updated based on the executor parameters. It should depend on the output of `params_to_wasmtime_semantics`, in other words it shouldn't be a constant anymore. The execute worker should already have access to the executor parameters. Can you make that change here? Hope it makes sense. 😅

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -331,7 +382,7 @@ struct JobResponse {
 /// - If success, pipe back `JobResponse`.
 fn handle_child_process(
 	pvf: PvfPrepData,
-	mut pipe_write: PipeWriter,
+	pipe_write: &mut PipeWriter,
```
mrcnski, line 385: I'm curious, why this change?

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: Why is it, do you get an error? I believe dropping either of those is not strictly needed, but would be nice to do.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,31 +234,74 @@ pub fn worker_entrypoint(
...
+						let stack = 1024 * 1024;
+						let mut stack: Vec<u8> = vec![0u8; stack];
+						let flags = CloneFlags::CLONE_NEWCGROUP
+						| CloneFlags::CLONE_NEWIPC
+						| CloneFlags::CLONE_NEWNET
+						| CloneFlags::CLONE_NEWNS
+						| CloneFlags::CLONE_NEWPID
+						| CloneFlags::CLONE_NEWUSER
+						| CloneFlags::CLONE_NEWUTS
+						| CloneFlags::from_bits_retain(libc::SIGCHLD);
```
mrcnski, line 248: I'm confused by this line.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,31 +234,74 @@ pub fn worker_entrypoint(
...
-							prepare_job_kind,
-							executor_params,
-						)
-					},
-					Ok(ForkResult::Parent { child }) => {
-						// the read end will wait until all write ends have been closed,
-						// this drop is necessary to avoid deadlock
-						drop(pipe_writer);
+				cfg_if::cfg_if! {
+					if #[cfg(any(target_os = "linux"))] {
```
mrcnski, line 238: Nice usage of `cfg_if`! Do we need the `any` here?

mrcnski (changes requested):
Looks really good! This will make us even more secure. 😁

After this PR, I would suggest moving away from security-related tickets. A decision has come to replace wasmtime with polkavm fairly soon: https://github.com/paritytech/polkadot-sdk/pull/2461/. Polkavm is already secured by default. I mean, it will take some time to switch as we need to do a migration, so it is good that we secured what we have already. But we can focus on other areas now. :)

polkadot/node/core/pvf/prepare-worker/src/lib.rs
Jpserrat: I don't think so. Done this way cause I was looking at how `nix` handled it. 
One unrelated point, I noticed that nix also has `pipe` in an already imported feature, what do you think about removing `os_pipe` and using the nix one to reduce dependency?

polkadot/node/core/pvf/prepare-worker/src/lib.rs
Jpserrat: I added a comment about this for this flag, but I removed it at some point, don't know why. I'm going to add this back.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
Jpserrat: I'm going to change this.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: >  The SIGCHILD flag is needed for the waitpid, if I don't pass it the waitpid call return ECHILD

Oh I see! Yes, please add a comment, it is pretty esoteric!

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: > I don't think so. Done this way cause I was looking at how nix handled it.

Probably not needed in nix either. ;) Let's remove it.

> One unrelated point, I noticed that nix also has pipe in an already imported feature, what do you think about removing os_pipe and using the nix one to reduce dependency?

Great observation! I just looked at the source for both, and it looks like `os_pipe` does a bit more to be safer and cross-platform. It also returns separate types for the reader and writer, which is better for type safety, whereas nix only returns a pair of `OwnedFd`s.

`os_pipe` is a pretty small dependency though. If you want to remove one dependency from the codebase, it could be nice to build a wrapper over `nix`'s `pipe`, that does the stuff from `os_pipe`. I'd say it's not necessary though.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: I just tested and I think it compiled fine with `mut pipe_write` (how it was before).

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: > I cannot drop either of them, because the FnMut will reuse this.

It compiles fine for me if I drop them. 🤔

Jpserrat: For me running `cargo test --target x86_64-unknown-linux-gnu` it doesn't work but with `cargo build --target x86_64-unknown-linux-gnu` it does.

Jpserrat: Both are working fine for you? Which one should I believe in? 🤣

mrcnski: What platform are you on? If you're on Mac, tests are not going to work if you build for Linux. What's the error you get?

Jpserrat:
```rust
error[E0507]: cannot move out of `stream`, a captured variable in an `FnMut` closure
   --> polkadot/node/core/pvf/prepare-worker/src/lib.rs:254:14
    |
210 |         |mut stream, worker_dir_path| {
    |          ---------- captured outer variable
...
253 |                             Box::new(|| {
    |                                      -- captured by this `FnMut` closure
254 |                                 drop(stream);
    |                                      ^^^^^^ move occurs because `stream` has type `UnixStream`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `pipe_reader`, a captured variable in an `FnMut` closure
   --> polkadot/node/core/pvf/prepare-worker/src/lib.rs:255:14
    |
226 |                 let (pipe_reader, mut pipe_writer) = os_pipe::pipe()?;
    |                      ----------- captured outer variable
...
253 |                             Box::new(|| {
    |                                      -- captured by this `FnMut` closure
254 |                                 drop(stream);
255 |                                 drop(pipe_reader);
    |                                      ^^^^^^^^^^^ move occurs because `pipe_reader` has type `PipeReader`, which does not implement the `Copy` trait

error[E0382]: borrow of moved value: `stream`
   --> polkadot/node/core/pvf/prepare-worker/src/lib.rs:215:28
    |
210 |         |mut stream, worker_dir_path| {
    |          ---------- move occurs because `stream` has type `UnixStream`, which does not implement the `Copy` trait
...
215 |                 let pvf = recv_request(&mut stream)?;
    |                                        ^^^^^^^^^^^ value borrowed here after move
...
253 |                             Box::new(|| {
    |                                      -- value moved into closure here, in previous iteration of loop

error[E0382]: use of moved value: `pipe_reader`
   --> polkadot/node/core/pvf/prepare-worker/src/lib.rs:270:10
    |
226 |                 let (pipe_reader, mut pipe_writer) = os_pipe::pipe()?;
    |                      ----------- move occurs because `pipe_reader` has type `PipeReader`, which does not implement the `Copy` trait
...
253 |                             Box::new(|| {
    |                                      -- value moved into closure here
254 |                                 drop(stream);
255 |                                 drop(pipe_reader);
    |                                      ----------- variable moved due to use in closure
...
270 |                                     pipe_reader,
    |                                     ^^^^^^^^^^^ value used here after move```

Comment (Jpserrat) [reply polkadot/node/core/pvf/prepare-worker/src/lib.rs:248]: This makes sense to me, cause if stream and pipe are being moved to the callback they won't be available for the parent, right?

File: polkadot/node/core/pvf/prepare-worker/src/lib.rs [hunk 6]
```diff
@@ -229,31 +234,74 @@ pub fn worker_entrypoint(
...
+						let stack = 1024 * 1024;
+						let mut stack: Vec<u8> = vec![0u8; stack];
+						let flags = CloneFlags::CLONE_NEWCGROUP
+						| CloneFlags::CLONE_NEWIPC
+						| CloneFlags::CLONE_NEWNET
+						| CloneFlags::CLONE_NEWNS
```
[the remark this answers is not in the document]
Jpserrat: Great observation! I'm going to add a check for the `CAP_SYS_ADMIN` before doing the clone. As the `unshare_user_namespace_and_change_root` is being used for the `CLONE_NEWUSER` and `CLONE_NEWNS`, I can remove both of them from here, right? Is that ok for you @mrcnski?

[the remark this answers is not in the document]
mrcnski: I believe `CLONE_NEW_USER` should already give us `CAP_SYS_ADMIN`:

              This flag has the same effect as the [clone(2)](https://man7.org/linux/man-pages/man2/clone.2.html)
              CLONE_NEWUSER flag.  Unshare the user namespace, so that
              the calling process is moved into a new user namespace
              which is not shared with any previously existing process.
              As with the child process created by [clone(2)](https://man7.org/linux/man-pages/man2/clone.2.html) with the
              CLONE_NEWUSER flag, the caller obtains a full set of
              capabilities in the new namespace.

This is how we are able to use `CLONE_NEWNS`, which also requires `CAP_SYS_ADMIN`, in `unshare_user_namespace_and_change_root`.

But, indeed it is possible that `clone` with these flags may fail on some systems/configurations. We may need to add a new capability to `SecurityStatus`. We can test ahead of time whether this `clone` call is expected to succeed, like we do with landlock, seccomp etc. If the check fails, and the user is not running in secure mode, we would fall back to fork here. I can implement this part if it sounds good.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4535514

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4536602

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4537320

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4537316

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4538382

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4538381

polkadot/node/core/pvf/prepare-worker/Cargo.toml
```diff
@@ -16,7 +16,7 @@ tracking-allocator = { package = "staging-tracking-allocator", path = "../../../
 tikv-jemalloc-ctl = { version = "0.5.0", optional = true }
 tikv-jemallocator = { version = "0.5.0", optional = true }
 os_pipe = "1.1.4"
```
mrcnski, line 18: Since we ported over the pipe creation, this can be removed right?

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -184,6 +221,36 @@ macro_rules! decl_worker_main {
 	};
 }
 
+// os_pipe lib
+#[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "haiku")))]
+pub fn pipe2_cloexec() -> io::Result<(libc::c_int, libc::c_int)> {
```
mrcnski, line 226: Nice, just needs:

1. A better comment with a bit more context. Something like "taken from the os_pipe crate. Copied here to reduce one dependency and because its type-safe abstractions do not play well with `nix`'s `clone`.
2. Let's remove the `ios` and `haiku` targets since we don't care about those. 😁 We only target unix systems like mac and linux.

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -30,6 +31,42 @@ use std::{
...
+// DEFAULT_NATIVE_STACK_MAX is set to 256 MiB. Not nearly enough.
+//
+// Hence we need to increase it. The simplest way to fix that is to spawn a thread with the desired
+// stack limit.
+//
+// The reasoning why we pick this particular size is:
+//
+// The default Rust thread stack limit 2 MiB + 256 MiB wasm stack.
+/// The stack size for the execute thread.
+pub const EXECUTE_THREAD_STACK_SIZE: usize = 2 * 1024 * 1024 + DEFAULT_NATIVE_STACK_MAX as usize;
```
mrcnski, line 68: Oh I'm sorry, this is specific to execute jobs. For prepare jobs we have been using 2 MiB stack sizes without any issues. My bad, brain fart on my end. Can you please remove this and pass 2 MiB to `clone`'s stack size?

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,50 +240,125 @@ pub fn worker_entrypoint(
...
+							Box::new(|| {
+								// SAFETY: pipe_writer is an open and owned file descriptor at this point.
+								let mut pipe_writer = PipeWriter::from_raw_fd(pipe_writer);
+
+								if let Err(errno) = nix::unistd::close(pipe_reader) {
+									send_child_response(&mut pipe_writer, JobResult::Err(error_from_errno("closing pipe", errno)));
+								}
+
+								if let Err(errno) = nix::unistd::close(stream_fd) {
+									send_child_response(&mut pipe_writer, JobResult::Err(error_from_errno("error closing stream", errno)));
```
mrcnski, line 273: Can we move the dropping code to the `handle_*_process` functions to clean up some duplication here? We can just pass in raw fd's to those functions, with comments to ensure on the caller-side that the correct fd's are given. This way we won't need the `PipeWriter`/`PipeReader` abstractions from `os_pipe` and can remove that dependency.

mrcnski (changes requested):
Nice! Just a few suggestions, and then it should be good!

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: Random note: one might wonder (or at least, I was wondering), instead of using the raw fd's, why not call `try_clone()` on `PipeWriter`/`PipeReader` before fork/clone? But that invokes [`dup`](https://man7.org/linux/man-pages/man2/dup.2.html) under the hood, which "allocates a new file descriptor that refers to the same open file description as the descriptor oldfd". Closing the cloned fd from `dup` in the child would not accomplish much, because the original fd would still be present in the child's address space (though, in the case of `clone`, not accessible by normal Rust code due to the borrow checker).

mrcnski:
> Do you mean, adding the clone change in there? Sure!

Hmm. It might be too much of an an implementation detail to mention in the implementer's guide. Low-level details should stay in the code, e.g. module docs. In the impl guide we should just say that the jobs have some additional sandboxing measures that workers don't have. But, we can mention clone in the top-level docs of `common/workers/security/mod.rs`. And I would move the clone args to that file also, so we can re-use them for the exec worker later. :)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4618025

Jpserrat:
Hey @mrcnski, I got back and did the fixes that you asked for. I'm going to start working on the file wrapper that you mentioned now, after that I'll do the execute worker!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4746250

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4747888

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4757401

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4758072

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -184,6 +184,36 @@ macro_rules! decl_worker_main {
 	};
 }
 
+//taken from the os_pipe crate. Copied here to reduce one dependency and
+// because its type-safe abstractions do not play well with nix's clone
+#[cfg(not(any(target_os = "macos")))]
```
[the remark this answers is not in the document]
mrcnski: Here it was fixed but below there is still an unnecessary `any`.

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -171,6 +176,66 @@ macro_rules! decl_worker_main {
...
+}
+
+pub struct PipeFd {
+	file: File,
+}
+
+impl PipeFd {
+	pub fn new(fd: i32) -> Self {
+		// SAFETY: pipe_writer is an open and owned file descriptor at this point.
+		let file = unsafe { File::from_raw_fd(fd) };
```
mrcnski, line 216: This should be an `unsafe fn`, because we can't guarantee here that the fd is valid. It's up to the caller to make sure the call is valid.

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -171,6 +176,66 @@ macro_rules! decl_worker_main {
...
+	let res = unsafe { libc::fcntl(fds[1], libc::F_SETFD, libc::FD_CLOEXEC) };
+	if res != 0 {
+		return Err(io::Error::last_os_error());
+	}
+	Ok((fds[0], fds[1]))
+}
+
+pub struct PipeFd {
+	file: File,
+}
```
mrcnski, line 211: Please document everything you've added for future readers of this code. Can you describe why this is here (i.e. justify its existence). All pub items should also be documented, and ideally the private ones too if you can!

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -171,6 +176,66 @@ macro_rules! decl_worker_main {
...
+		let file = unsafe { File::from_raw_fd(fd) };
+		PipeFd { file }
+	}
+
+	pub fn as_raw_fd(&self) -> i32 {
+		self.file.as_raw_fd()
+	}
+}
+
+impl Write for PipeFd {
```
mrcnski, line 225: Can we also implement `Read` for use in the parent process? Nice job on this, BTW!

mrcnski:
I'll work on detection of the clone capability as described [here](https://github.com/paritytech/polkadot-sdk/pull/2477/#discussion_r1406110772). I'm not sure we should enforce it though - we've already had validators, mostly running in containers, who can't enable some security features such as pivot_root. @koute What does PolkaVM do in that situation?

koute:
> @koute What does PolkaVM do in that situation?

Currently nothing. If it's not possible to start a VM with full sandboxing then the instantiation will fail. I only have a fallback for when `clone3` is not available due to a too old kernel.

The plan is to eventually:

1) find out which exact kernel version is necessary for full functionality and sandboxing, 
2) find out which exact kernel version is necessary for the bare minimum,
3) only officially support the (1) kernel (and this includes the case when the kernel is new enough, but it has some features we need disabled),
4) have a flag which will allow the VM to start with reduced security that will also work in a non-official configuration,
3) add tests for all of this to the CI.

So, basically, if everything's OK then the VM will start normally, but if it's not (assuming the kernel is not totally ancient) then it'll be necessary to pass an extra flag to `Config` to force it to start with reduced security, and the idea is that this flag will be routed to the command line so the user will have to make this decision explicitly by themselves.

(But, again, this is not implemented yet as this will be pretty easy to do and I have a lot higher priorities right now.)

polkadot/node/core/pvf/prepare-worker/src/lib.rs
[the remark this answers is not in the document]
alindima: > But, indeed it is possible that clone with these flags may fail on some systems/configurations. We may need to add a new capability to SecurityStatus.

Yes, I think we should treat these extra namespace flags as separate security features (like we do for the user namespace).

But why do we bother with switching the fork+unshare to clone? I don't think there's a big benefit in doing so (and looks like it's quite complicated to implement in a bulletproof manner)

mrcnski:
@koute Thanks, that's good to know. That sounds pretty similar to what we do with Secure Validator Mode and the insecure flag. This part at least should be pretty easy to migrate to PolkaVM later. 

I would just suggest not relying on kernel versions, but directly testing for the capabilities. The kernel version alone is not enough to tell you whether a capability is actually present - [example](https://github.com/paritytech/polkadot-sdk/issues/2728#issuecomment-1863534782).

polkadot/node/core/pvf/prepare-worker/src/lib.rs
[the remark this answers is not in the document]
mrcnski: @alindima We sandbox the worker soon after it starts, with some exceptions, because it doesn't need much I/O functionality. It seems good to keep this, even though this process doesn't directly interface with untrusted code anymore. When spawning the job process, we can restrict it even further, and doing it at `clone` seems like the appropriate time for it.

But, perhaps we could move some of the flags to the `unshare` call in the parent worker. Probably the worker doesn't need most of these capabilities. Then this `clone` call could be redundant. I can test it when I have a moment.

> (and looks like it's quite complicated to implement in a bulletproof manner)

Why do you say this?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4769297

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -182,44 +181,103 @@ pub fn worker_entrypoint(
...
-						)?
-					},
-				};
-
-				send_response(&mut stream, response)?;
+				let stream_fd = stream.as_raw_fd();
+
+				cfg_if::cfg_if! {
+					if #[cfg(target_os = "linux")] {
+						let stack_size = 2 * 1024 * 1024; // 2MiB
```
mrcnski, line 188: Remember my comment about the stack size for execution!

mrcnski (changes requested):
Looking good! Just one thing that needs to be fixed, and a couple things remaining from me:

- [ ] Check if some of all of these flags can be moved to `unshare` in the parent worker.
- [ ] Add a check for this capability at startup

polkadot/node/core/pvf/execute-worker/src/lib.rs
Jpserrat: Sorry haha, I'm gonna fix that!

koute:
> I would just suggest not relying on kernel versions, but directly testing for the capabilities. The kernel version alone is not enough to tell you whether a capability is actually present - [example](https://github.com/paritytech/polkadot-sdk/issues/2728#issuecomment-1863534782).

Obviously, the code won't actually check the kernel version (it will just see if the operation succeeds and check the error code, as that's the only way to reliably do the check), but we need to know which versions are those so that we can test them on the CI.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4770738

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -185,8 +189,7 @@ pub fn worker_entrypoint(
 
 				cfg_if::cfg_if! {
 					if #[cfg(target_os = "linux")] {
-						let stack_size = 2 * 1024 * 1024; // 2MiB
-						let mut stack: Vec<u8> = vec![0u8; stack_size];
+						let mut stack: Vec<u8> = vec![0u8; EXECUTE_THREAD_STACK_SIZE];
```
mrcnski, line 192: We need to get this value from the executor params, and the constant should be removed. Please consult my earlier comment - there's a converter function you need to use on the exec params. Hope it makes sense. :pray:

polkadot/node/core/pvf/prepare-worker/src/lib.rs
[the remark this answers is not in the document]
alindima: I'm mainly saying that the other `unshare` call received special treatment by being added as a security feature, with more robust handling in case it's not supported. I'm just saying that these extra namespace flags should be handled similarly, to preserve compatibility and avoid errors

polkadot/node/core/pvf/execute-worker/src/lib.rs
Jpserrat: One question, is to replace the `EXECUTE_THREAD_STACK_SIZE` with `deterministic_stack_limit` of the `params_to_wasmtime_semantics`, correct? 
The `deterministic_stack_limit` is an Optional value, what is going to be the default value in case of `None` return?
Should we still use the `EXECUTE_THREAD_STACK_SIZE`?

mrcnski: Yep, `deterministic_stack_limit.native_stack_max`!

Since we are spawning some threads in the process, we should add an additional 2 MiB (the default) to the process's stack size, for *each* thread's stack size. (Don't forget the process's main thread!) This should be documented, here as well as at the point where we spawn the threads, to make sure the value stays in sync with the number of threads.

In pseudo-code, `process_stack_size = wasm_native_stack_size + 2 MiB * num_threads`. Put another way, `process_stack_size = execute_stack_size + 2 MiB * num_other_threads`.*

The execute thread should continue to be spawned with 2 MiB (its own stack size) + the wasm stack size, as before. (And again, this should not be const.)

*NB, I am not sure if the process stack size is simply the sum of stack sizes of all threads, but it should be a good enough approximation for these purposes. It should suffice to account for the wasm stack size properly, as set in the executor params, and add a bit extra for the threads.

> The deterministic_stack_limit is an Optional value, what is going to be the default value in case of None return?

Good question! It's actually impossible for it to be `None` (the value comes from `DEFAULT_CONFIG`), and this is already [proven](https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/STYLE_GUIDE.md#style) in `params_to_wasmtime_semantics`. I would return the `stack_limit` from `params_to_wasmtime_semantics` separately, since we are able to do so without wrapping it in `Option`, and then the function signature itself provides the guarantee that it is present.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4793242

polkadot/node/core/pvf/common/src/error.rs
```diff
@@ -77,8 +77,8 @@ pub enum PrepareError {
 	#[codec(index = 9)]
 	ClearWorkerDir(String),
 	/// The preparation job process died, due to OOM, a seccomp violation, or some other factor.
-	JobDied { err: String, job_pid: i32 },
 	#[codec(index = 10)]
+	JobDied { err: String, job_pid: i32 },
```
mrcnski, line 81: Had a fun debug session. 🙃 My bad, this one was on me.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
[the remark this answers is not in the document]
mrcnski: > But, perhaps we could move some of the flags to the unshare call in the parent worker.

Punted on this because it could cause sudden loss of the unshare-and-pivot-root capability on some validators. It is enough for the `clone` call to have these flags to be reasonably restricted.

> I'm just saying that these extra namespace flags should be handled similarly, to preserve compatibility and avoid errors

Will do this next. Without the CI that koute suggested we don't know what are the requirements for the union of all these flags. I'm therefore inclined to make this clone-with-sandbox-flags capability optional for Secure Validator Mode. It's also not needed to reasonably sandbox networking and filesystem access (our main threat model is protecting validator keys), so seems fine not to enforce it (we would still print a warning if it's missing).

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4807242

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4807387

polkadot/node/core/pvf/execute-worker/src/lib.rs
Jpserrat: Thanks for the great explanation! Changed the stack size for the clone using 3 threads for the clone and 1 for the execute worker thread.

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -16,80 +16,48 @@
...
+	os::{fd::AsRawFd, unix::net::UnixStream},
 	path::PathBuf,
 	process,
 	sync::{mpsc::channel, Arc},
 	time::Duration,
 };
 
-// Wasmtime powers the Substrate Executor. It compiles the wasm bytecode into native code.
-// That native code does not create any stacks and just reuses the stack of the thread that
-// wasmtime was invoked from.
```
mrcnski, line 59: Instead of removing this comment, can we put it above `get_max_stack_size`? It still seems relevant and useful.

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -334,28 +368,52 @@ fn handle_child_process(
 	send_child_response(&mut pipe_write, response);
 }
 
+/// Returns clone stack size
+/// The stack size is represented by 2MiB * number of threads + native stack;
+fn get_max_stack_size(executor_params: &ExecutorParams, number_of_threads: u32) -> usize {
+	let deterministc_stack_limit = params_to_wasmtime_semantics(executor_params)
```
mrcnski, line 375:
```suggestion
	let deterministic_stack_limit = params_to_wasmtime_semantics(executor_params)
```

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -334,28 +368,52 @@ fn handle_child_process(
 	send_child_response(&mut pipe_write, response);
 }
 
+/// Returns clone stack size
```
mrcnski, line 372: It's not just for `clone` since we also use it for spawning threads.

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -182,44 +149,93 @@ pub fn worker_entrypoint(
...
+										pipe_reader,
+										stream_fd,
+										Arc::clone(&compiled_artifact_blob),
+										Arc::clone(&arc_executor_params),
+										Arc::clone(&params),
+										execution_timeout,
+										execute_worker_stack_size,
+									)
+								}),
+								stack_size as usize,
```
mrcnski: I think it's already a `usize`:

```suggestion
								stack_size,
```

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -334,28 +368,52 @@ fn handle_child_process(
 	send_child_response(&mut pipe_write, response);
 }
 
+/// Returns clone stack size
+/// The stack size is represented by 2MiB * number of threads + native stack;
+fn get_max_stack_size(executor_params: &ExecutorParams, number_of_threads: u32) -> usize {
+	let deterministc_stack_limit = params_to_wasmtime_semantics(executor_params)
+		.deterministic_stack_limit
+		.expect("the default value is always available.");
```
mrcnski, line 377: Proofs should follow a logical chain of reasoning and conclude with `qed`. Please look for examples in the repo, and let me know if it's still not clear. :)

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -271,12 +287,31 @@ fn validate_using_artifact(
...
-	executor_params: ExecutorParams,
-	params: Vec<u8>,
+	pipe_write_fd: i32,
+	pipe_read_fd: i32,
+	stream_fd: i32,
+	compiled_artifact_blob: Arc<Vec<u8>>,
+	executor_params: Arc<ExecutorParams>,
+	params: Arc<Vec<u8>>,
 	execution_timeout: Duration,
+	execute_worker_stack: usize,
```
mrcnski, line 298: For clarity:

```suggestion
	execute_worker_stack_size: usize,
```

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -182,44 +149,93 @@ pub fn worker_entrypoint(
...
+
+				cfg_if::cfg_if! {
+					if #[cfg(target_os = "linux")] {
+						use polkadot_node_core_pvf_common::worker::security;
+
+						// Number of threads here is 3:
+						// 1 - Main thread
+						// 2 - Cpu monitor thread
+						// 3 - Execute thread
+						let stack_size = get_max_stack_size(&executor_params, 3);
```
mrcnski: Nice! Please just leave a comment in the place where we spawn the threads, reminding future maintainers to update this value if the # of threads changes.

mrcnski (changes requested):
Looks good! 😀 Just needs a bit of cleanup!.And happy holidays!

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -229,50 +239,100 @@ pub fn worker_entrypoint(
...
-					%worker_pid,
-					"worker: sending result to host: {:?}",
-					result
-				);
-				send_response(&mut stream, result)?;
+				let stream_fd = stream.as_raw_fd();
+
+				cfg_if::cfg_if! {
+					if #[cfg(target_os = "linux")] {
+						let stack_size = 2 * 1024 * 1024; // 2mb
```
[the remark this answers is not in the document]
mrcnski: @Jpserrat I guess this should be updated to be 2 MiB * # of threads.

polkadot/node/core/pvf/execute-worker/src/lib.rs
Jpserrat: If the `deterministic_stack_limit` is always available, why does it need to be an optional value?

mrcnski: I believe the `Semantics` type comes from the wasmtime executor crate, and I suppose there's some reason for `deterministic_stack_limit` to be optional in its API. There are other consumers of that crate, after all. However, here we know that it's always available, so we could either return it from `params_to_wasmtime_semantics`, or we could have a solid proof here. Hope that's clear!

s0me0ne-unkn0wn: Yeah I believe we don't impose a stack limit on relay chain runtime execution, that's why it's optional.

eskimor: Proofs have to be locally reasonable, it should never break because some code somewhere else gets changed. Or put differently, whoever can (realistically) break your proof, must have seen the proof - because it is right next to it.

polkadot/node/core/pvf/execute-worker/src/lib.rs
mrcnski: Or better, instead of doing that, let's extract `3` into a constant and use the constant here in this test:

https://github.com/paritytech/polkadot-sdk/blob/3366dc2142973e74c882fab1c207458b0945ebba/polkadot/node/core/pvf/tests/it/process.rs#L376

polkadot/node/core/pvf/execute-worker/src/lib.rs
mrcnski: Yes, and since there is already a proof in `params_to_wasmtime_semantics`, I would opt for returning the value from that function.

Jpserrat: Thanks!

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -334,28 +372,87 @@ fn handle_child_process(
...
+/// typically it's configured to 8 MiB. Rust's spawned threads are 2 MiB. OTOH, the
+/// DEFAULT_NATIVE_STACK_MAX is set to 256 MiB. Not nearly enough.
+///
+/// Hence we need to increase it. The simplest way to fix that is to spawn a thread with the desired
+/// stack limit.
+///
+/// The reasoning why we pick this particular size is:
+///
+/// The default Rust thread stack limit 2 MiB + 256 MiB wasm stack.
+/// The stack size for the execute thread.
```
mrcnski, line 411: I believe this is covered by the above already, and is a bit out-of-date:

```suggestion
```

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -145,17 +113,28 @@ pub fn worker_entrypoint(
...
 
+			let executor_params: Arc<ExecutorParams> = Arc::new(executor_params);
+			let execute_worker_stack_size = get_max_stack_size(&executor_params, 1);
+			// Number of threads here is 3:
+			// 1 - Main thread
+			// 2 - Cpu monitor thread
+			// 3 - Execute thread
+			//
+			// NOTE: if the number of threads inside the child process change in the future, this
+			// value must be changed as well.
```
mrcnski, line 129: Let's move `3` to a constant (see my other comment) and move this comment to a docstring over the new constant. And please add, "The correctness of this value is enforced by a test".

Jpserrat:
I hope it's good now! 😃  
Happy holidays!

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -334,28 +374,82 @@ fn handle_child_process(
 	send_child_response(&mut pipe_write, response);
 }
 
+/// Returns stack size based on the number of threads.
+/// The stack size is represented by 2MiB * EXECUTE_WORKER_THREAD_NUMBER + native stack;
```
mrcnski, line 378:
```suggestion
/// The stack size is represented by 2MiB * number_of_threads + native stack;
```

mrcnski (approved):
I fixed a couple comments that were missed. Make sure you look over the commits, and ask questions if something is not clear!

Thanks as always for your hard work! 😊

Jpserrat:
Just noticed that I completely forgot prepare-worker 😞 
Thanks for your help!

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4822832

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4823273

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4823417

mrcnski:
`pdeathsig` seems not to work in `clone`, breaking some tests. I'm inclined to just remove it along with the associated tests. As pointed out on the [PR](https://github.com/paritytech/polkadot-sdk/pull/2643#pullrequestreview-1769563391), in the usual case the children will already exit on their own. In the malicious case, the `pdeathsig` can be easily unset, limiting its usefulness.

polkadot/node/core/pvf/common/src/worker/security/clone.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// Returns flags for `clone(2)`, including all the sandbox-related ones.
+fn clone_flags() -> CloneFlags {
+	// NOTE: CLONE_NEWUSER does not work when cloning job processes, but in Secure Validator Mode it
+	// is already set by the worker.
+	//
+	// SIGCHLD flag is used to inform clone that the parent process is
+	// expecting a child termination signal, without this flag `waitpid` function
+	// return `ECHILD` error.
+	CloneFlags::CLONE_NEWCGROUP |
```
alexggh, line 85:
```
From man page:
Only a privileged process (CAP_SYS_ADMIN) can employ CLONE_NEWCGROUP
```
Is that something we want ?
 
Not sure if that is already a hard requirement for running `polkadot` but it feels like it shouldn't be.

alexggh (approved):
Overall, looks good to me left you some comments.

tdimitrov (approved):
Good job! I really enjoyed reviewing this PR.

polkadot/node/core/pvf/common/src/worker/security/clone.rs
```diff
@@ -0,0 +1,92 @@
...
+/// SAFETY: new process should be either spawned within a single threaded process, or use only
+/// async-signal-safe functions.
+unsafe fn try_clone(cb: CloneCb, stack_size: usize, flags: CloneFlags) -> Result<Pid> {
+	let mut stack: Vec<u8> = vec![0u8; stack_size];
+
+	nix::sched::clone(cb, stack.as_mut_slice(), flags, None).map_err(|errno| Error::Clone(errno))
+}
+
+/// Returns flags for `clone(2)`, including all the sandbox-related ones.
+fn clone_flags() -> CloneFlags {
```
alindima, line 78: this can be a constant

polkadot/node/core/pvf/common/src/worker/security/clone.rs
mrcnski: Good catch! I was going to include the `CLONE_NEWUSER` clone flag, which gives us this capability. From `unshare(2)`:

              As with the child process created by [clone(2)](https://man7.org/linux/man-pages/man2/clone.2.html) with the
              CLONE_NEWUSER flag, the caller obtains a full set of
              capabilities in the new namespace.

However, for reasons I don't understand, `clone` with this flag fails if we already passed the same flag to `unshare` [here](https://github.com/paritytech/polkadot-sdk/blob/69713a718566a753ba2735cd9810546f6c0eb681/polkadot/node/core/pvf/common/src/worker/security/change_root.rs#L94), when sandboxing the parent worker process. So I was going to parametrize this function over whether the `unshare` capability is present - I just forgot to do it. If it is present, then the parent worker process already has `CAP_SYS_ADMIN`. That's also why`clone_flags` is a `fn` and not a `const` @alindima.

mrcnski: Okay, made this change. If for some reason we cannot `unshare`, we will need the `CAP_SYS_ADMIN` so I added the clone flag for that case. Hopefully there are no issues doing so in the wild, because admittedly local testing and CI have not caught all the issues in the past. However, this new capability is optional, so worst-case scenario someone sees a warning and reports it to us.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -76,6 +78,16 @@ static ALLOC: TrackingAllocator<tikv_jemallocator::Jemalloc> =
...
 
+/// The number of threads for the child process:
+/// 1 - Main thread
+/// 2 - Cpu monitor thread
+/// 3 - Memory tracker thread
+/// 3 - Prepare thread
+///
+/// NOTE: The correctness of this value is enforced by a test. If the number of threads inside
+/// the child process changes in the future, this value must be changed as well.
+pub const PREPARE_WORKER_THREAD_NUMBER: u32 = 4;
```
s0me0ne-unkn0wn, line 89: If we enable multi-threaded compilation, we cannot predict the number of threads as it's not controlled by us but by the Wasmtime. It's okay for now, but it would need reworking if we ever go for multi-threading here.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -76,6 +78,16 @@ static ALLOC: TrackingAllocator<tikv_jemallocator::Jemalloc> =
 #[global_allocator]
 static ALLOC: TrackingAllocator<std::alloc::System> = TrackingAllocator(std::alloc::System);
 
+/// The number of threads for the child process:
+/// 1 - Main thread
+/// 2 - Cpu monitor thread
+/// 3 - Memory tracker thread
+/// 3 - Prepare thread
```
s0me0ne-unkn0wn, line 85:
```suggestion
/// 4 - Prepare thread
```

polkadot/node/core/pvf/prepare-worker/src/lib.rs
```diff
@@ -306,21 +329,96 @@ struct JobResponse {
...
+	prepare_job_kind: PrepareJobKind,
+	executor_params: &Arc<ExecutorParams>,
+	worker_info: &WorkerInfo,
+	temp_artifact_dest: &Path,
+	usage_before: Usage,
+) -> Result<PrepareWorkerSuccess, PrepareError> {
+	use polkadot_node_core_pvf_common::worker::security;
+
+	// 2MiB * num_threads
+	let stack_size = 2 * 1024 * 1024 * PREPARE_WORKER_THREAD_NUMBER;
```
s0me0ne-unkn0wn, line 350: Stack size is a per-thread entity by definition. Threads cannot share stack. Thus, process stack size (to be precise, the stack size of the process's main thread) cannot depend on the number of threads. Unless I miss something, some logical mistake should be corrected here.

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -343,28 +425,70 @@ fn handle_child_process(
...
+///
+/// Typically on Linux the main thread gets the stack size specified by the `ulimit` and
+/// typically it's configured to 8 MiB. Rust's spawned threads are 2 MiB. OTOH, the
+/// DEFAULT_NATIVE_STACK_MAX is set to 256 MiB. Not nearly enough.
+///
+/// Hence we need to increase it. The simplest way to fix that is to spawn an execute thread with
+/// the desired stack limit. We must also make sure the job process has enough stack for *all* its
+/// threads. This function can be used to get the stack size of either the execute thread or execute
+/// job process.
+fn get_max_stack_size(executor_params: &ExecutorParams, number_of_threads: u32) -> usize {
```
s0me0ne-unkn0wn, line 467: 1) See the comment on the preparation worker stack size; it shouldn't depend on the number of threads;
2) You could get the full semantics and extract the stack size here. There is no need to return the stack separately from `params_to_wasmtime_semantics()` as that's not an optimization: you process the full set of executor parameters to find out the stack size anyway.

polkadot/node/core/pvf/src/security.rs
```diff
@@ -348,6 +361,50 @@ async fn check_seccomp(
...
+/// We do this check by spawning a new process and trying to sandbox it. To get as close as possible
+/// to running the check in a worker, we try it... in a worker. The expected return status is 0 on
+/// success and -1 on failure.
+async fn check_can_do_secure_clone(
+	#[cfg_attr(not(target_os = "linux"), allow(unused_variables))]
+	prepare_worker_program_path: &Path,
+) -> SecureModeResult {
+	cfg_if::cfg_if! {
+		if #[cfg(target_os = "linux")] {
+			match tokio::process::Command::new(prepare_worker_program_path)
```
s0me0ne-unkn0wn, line 375: AFAIR we're running workers in some kind of clean environment (pruning environment variables, etc.) In theory, running this check in a different environment could result in a different outcome of the check (think about `LD_PRELOAD`, `LD_LIBRARY_PATH`, and other things that cannot be talked about in polite society). I can hardly imagine how it can be an attack vector, but it still could be considered to run the check and the worker itself in the same environment; it's just safer.

s0me0ne-unkn0wn (changes requested):
Wow, good job, it's very appreciated!
Left some comments on things that are doubtful to me. It's possible I'm just not getting the reasoning behind them, but right now, some of them look to me like they should be fixed. Please follow up in discussions.
Either way, nicely done!

polkadot/node/core/pvf/common/src/worker/mod.rs
```diff
@@ -171,6 +191,83 @@ macro_rules! decl_worker_main {
...
+	let res = unsafe { libc::fcntl(fds[1], libc::F_SETFD, libc::FD_CLOEXEC) };
+	if res != 0 {
+		return Err(io::Error::last_os_error())
+	}
+	Ok((fds[0], fds[1]))
+}
+
+/// A wrapper around a file descriptor used to encapsulate and restrict
+/// functionality for pipe operations.
+pub struct PipeFd {
```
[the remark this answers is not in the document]
Jpserrat: Great point! Fixed the implementations.
`PipeFd` is just to restrict the file functions.

polkadot/node/core/pvf/prepare-worker/src/lib.rs
mrcnski: You are right, I made a [little experiment](https://github.com/mrcnski/clone-stack-experiment) to correct my understanding. So for the clone stack size we can always just use 2 MiB, or is that even overkill?

polkadot/node/core/pvf/common/src/executor_interface.rs
```diff
@@ -169,8 +172,8 @@ pub fn params_to_wasmtime_semantics(par: &ExecutorParams) -> Semantics {
 			ExecutorParam::PvfExecTimeout(_, _) => (), /* Not used here */
 		}
 	}
-	sem.deterministic_stack_limit = Some(stack_limit);
-	sem
+	sem.deterministic_stack_limit = Some(stack_limit.clone());
+	(sem, stack_limit)
```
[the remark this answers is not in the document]
mrcnski: > You could clone it from semantics anytime from the caller's side as well

We were doing that, but it required a proof with non-local reasoning as it was behind an `Option`. As we are guaranteed to have the stack limit, I changed it so that the function signature itself provides the guarantee that it exists. Let me know if you have any ideas for refactoring it!

polkadot/node/core/pvf/src/security.rs
mrcnski: Hmm that complicates things as we don't know here what other security features are available and will be enacted. It seems reasonable to always clear the env vars in all the checks, and also have that be the first sandboxing step done - anything else we can do?

polkadot/node/core/pvf/prepare-worker/src/lib.rs
s0me0ne-unkn0wn: Indeed, it is a shame that `nix` strictly requires a pre-allocated stack here. That's because it uses libc's `clone()` wrapper instead of raw `clone3()` syscall. You wouldn't need to specify explicit stack at all as you're not sharing virtual address space with the cloned process; the stack of the parent process would just be cloned as well as all the rest.

Still, we have what we have, and the child process requires some reasonable stack. Offhand, 2 Mb looks okay to me.

polkadot/node/core/pvf/src/security.rs
s0me0ne-unkn0wn: I believe that's enough, and it's not that critical anyway, can be a follow-up.

polkadot/node/core/pvf/common/src/executor_interface.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Well, fair enough. I'm trying to invent something around `Infallible` result out of my head, but cannot give birth to it right away. Let's leave it for the next refactoring then :)

polkadot/node/core/pvf/src/security.rs
mrcnski: Makes sense to do it here. We will miss the next merge window, which is in less than three hours - which means we can take our time with this PR. :)

polkadot/node/core/pvf/common/src/worker/mod.rs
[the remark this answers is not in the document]
mrcnski: Exactly. We could pass around `RawFd` instead, but having a separate type just makes it more clear that these fd's correspond to pipes and not some other kind of fd. The commit looks good  @Jpserrat!

s0me0ne-unkn0wn (approved):
Yes, that looks good now! Thank you!

Issue #586, opened 2023-08-03
Consider refactoring subsystem initialization in overseer


mrcnski:
Currently we include all subsystems in `RealOverseerGen`, irregardless of whether they are needed or not. A couple of them are only initialized based on passed-in parameters. E.g. there is a conditional for whether pvf-checker is enabled, and if not we spawn a `DummySubsystem`, which is a bit clumsy.

## Proposal

We could instead nix `RealOverseerGen` and have three new structs that implement `OverseerGen`:

1. `ValidatorOverseerGen`
2. `CollatorOverseerGen`
3. `FullNodeOverseerGen`

Then for each we only initialize the subsystems that are relevant to that kind of node.

Edit: in the future there can be nodes that are both validators and collators. See https://github.com/paritytech/polkadot/pull/7566#discussion_r1279753180. Perhaps a capabilities-based approach would be better, or a builder pattern, e.g. `new().is_validator(true).is_collator(true)`.

I'm not too familiar with the code so looking for feedback. cc @ordian 

## Followup

After this has been done we can move the worker binary check (where we call `determine_workers_paths`) into `CandidateValidation::new`. This would ensure that the workers are only required for nodes that actually need them (validators).

Issue #2458, opened 2023-11-22
could not open either /var/log/audit/audit.log


tugytur:
With the release `polkadot 1.4.0-00395a23d12` when a validator becomes a p/v I get the following logs

```
2023-11-22 22:12:00 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:00 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1440
2023-11-22 22:12:00 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:00 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:02 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:02 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:05 💤 Idle (39 peers), best: #20673517 (0x172c…3e4f), finalized #20673515 (0xd810…4f52), ⬇ 2.0MiB/s ⬆ 1.5MiB/s
2023-11-22 22:12:06 ✨ Imported #20673518 (0x84ee…bf74)
2023-11-22 22:12:06 could not open either /var/log/audit/audit.log or /var/log/syslog for reading audit logs worker_pid=1383
2023-11-22 22:12:06 ✨ Imported #20673518 (0xeacd…27ac)
2023-11-22 22:12:10 💤 Idle (39 peers), best: #20673518 (0x84ee…bf74), finalized #20673515 (0xd810…4f52), ⬇ 1.5MiB/s ⬆ 2.7MiB/s
```

Processes are:

```
1383  0.2  0.0 283788 20212 ?        S    20:28   0:14 /usr/local/bin/polkadot-execute-worker execute-worker --node-impl-version 1.4.0 --socket-path /tmp/pvf-host-executeucSLBix89h --worker-dir-path /var/blockchain/data/chains/ksmcc3/db/full/pvf-artifacts/worker-dir-execute-wUmMklebcN --can-enable-landlock --can-enable-seccomp
[...]

PR #2486, opened 2023-11-24
PVF: Add Secure Validator Mode

### LINKED ISSUE #881: PVF host: Specialize on Linux, support macOS
(This is not an actionable issue, just a platform for discussion and linking from other issues)

To my knowledge, we do not support Windows explicitly. Maybe mingw may work, but MSVC almost surely won't work. That's fine since Windows has WSL for those who need to run Polkadot there.

There is quite a bunch of people who are using macOS for development either of Polkadot or parachains. While running Polkadot is possible inside a docker container, it comes with many inconveniences:

- `polkadot-launch` does not support docker, 
- macOS does not support docker natively so it spins up a virtual machine, which can eat a decent amount of memory and CPU, and then on top of that the host and the guest do not share the resources.
- profiling inside of a docker is not the best experience on macOS either

So therefore we probably want to keep supporting polkadot directly on macOS.

However, PVF host or other parts of wasm execution may want to use special features of Linux for performance, security or other reasons. To name a few:

1. Linux supports uffd which is useful for low-latency instantiation in wasmtime. Maybe macOS' mach layer supports for something like that, but it would be take additional effort to implement that.
2. we are interested in security hardening of the worker processes https://github.com/paritytech/polkadot-sdk/issues/882. Even if macOS supports something of that sort, it would take us to reimplement the same for it.
3. we may need to constrain the workers with amount of resources they can take. The best we can hope on macOS is probably `rlimit` (and the last time I looked I wanted to use cgroups, Linux only feature). And again if there are required facilities, we would have to reimplement everything for macOS again.

Thus, it may make sense to specialize our implementation efforts on Linux for production deployments. We can implement good enough fallbacks that would allow deploying simple localnets, but running validators on macOS will be discouraged.

### LINKED ISSUE #1444: Secure validator mode

### LINKED ISSUE #2495: PVF: "failed to read dir" error running `cargo bench`
Getting the following error after #1918:

```
Nov 24 15:44:09.323 ERROR parachain::pvf: failed to read dir "/tmp/.tmpxxQbgX" err=Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The PR has not been released yet, so we should find out whether this affects production (and also have a test to reproduce this).

The weird thing is, we do successfully create the dir beforehand with `create_dir_all`. But I've reproduced the error from `read_dir` on two machines already. 🤔

```rs
// Make sure that the cache path directory and all its parents are created.
if let Err(err) = tokio::fs::create_dir_all(cache_path).await {
    // NOTE: This log is not in `master`, I added it to verify that the dir exists.
    if err.kind() != io::ErrorKind::AlreadyExists {
        gum::error!(
            target: LOG_TARGET,
            ?err,
            "failed to create dir {:?}",
            cache_path,
        );
        return
    }
}

let mut dir = match tokio::fs::read_dir(cache_path).await {
    Ok(dir) => dir,
    Err(err) => {
        gum::error!(
            target: LOG_TARGET,
            ?err,
            "failed to read dir {:?}",
            cache_path,
        );
        return
    },
};
```

cc @eagr if you have any ideas. I'll try to get to it later today.

mrcnski:
## Summary

Running `--validator` without all security features available (with some exceptions) requires the `--insecure-validator-i-know-what-i-do` flag.

## Overview

We've recently added Linux-only security features. This PR makes them "required" to enforce protection on validators (and thus the network), though there is a CLI flag to bypass the restrictions.

#### If all capabilities are present, or the insecure flag is passed

Screenshot 2023-11-24 at 14 44 30

#### If some optional capabilities are missing

Screenshot 2023-11-24 at 14 34 29

#### If some required capabilities are missing

<details>

Screenshot 2023-11-21 at 19 52 44

Screenshot 2023-11-09 at 11 20 59
</details>

Yeah, this is really ugly, because we return `SubsystemError` from `candidate-validation`. We could move the check to before subsystem initialization, like we do for the worker version check. But I believe we want subsystems to be in charge of their own initialization and requirements (see https://github.com/paritytech/polkadot-sdk/issues/586.)

### Exceptions

Either one of the `landlock` or the `unshare-and-change-root capability` are allowed to be missing if the other is present, because either provides FS security. We decided to be lax about this as Landlock requires Linux 5.13+ and many validators are not yet on this version. See https://github.com/paritytech/polkadot-sdk/issues/1444#issuecomment-1808992442.

### Other

In this PR I also did the following. (It might be good to review commit-by-commit.)

- Shutdown the worker when we can’t enable security features that we were able to enable in the host.
- Add `secure_validator_mode` to `SecurityStatus`, pass it to workers this way.
- Pass security options to workers over the socket instead of CLI args. (Looks much cleaner.)
- Cleaned up a couple of old TODOs.
- Fix check_seccomp and check_landlock errors not being logged to stderr
  (involved a bit of a refactor)

Some refactors:

- Refactored `spawn_with_program_path`
- Refactored worker data (see “Worker data” at #1576). This was driving me crazy.
- Move change_root to its own file
- Refactor change_root errors

## TODO

- [ ] Should we merge this for 1.5? I vote yes, but objections are welcome:

There was an informative warning that was shown to validators so that they could prepare to meet the new security requirements. However, due to https://github.com/paritytech/polkadot-sdk/pull/2426, we announced to validators that they need to ignore the warning for one release. We could show the fixed warning in 1.5 and merge this in 1.6, but we may not have time with on-demand coming. See [release calendar](https://www.notion.so/paritytechnologies/Release-Calendar-cc269daa514f4dc996702e50689d1bca).

However, we recently decided that Landlock is optional and dropped the requirement for x86-64 (since we are [not doing full seccomp](https://github.com/paritytech/polkadot-sdk/pull/2461)). Almost no validators should actually need the insecure flag, so I believe that more warning time is not needed.

- [ ] How to abort subsystem initialization?

Currently it is pretty ugly (see screenshots above under "If some required capabilities are missing"). But I believe we want subsystems to be in charge of their own initialization and requirements (see https://github.com/paritytech/polkadot-sdk/issues/586.) We can move the check out of the subsystem, or make that shutdown less ugly maybe...

## Related

Closes #1444
Closes #881
Closes #2495

Previous attempt: https://github.com/paritytech/polkadot/pull/7073


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4479592

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4479870

polkadot/node/core/pvf/common/src/worker/security/mod.rs
```diff
@@ -27,143 +27,30 @@
...
 
-use crate::{worker::WorkerKind, LOG_TARGET};
+use crate::{worker::WorkerInfo, LOG_TARGET};
 
+#[cfg(target_os = "linux")]
+pub mod change_root;
 #[cfg(target_os = "linux")]
 pub mod landlock;
-
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```
mrcnski, line 36: @alindima to confirm, if we're only blocking `socket` and `io_uring`, can we remove the restriction on `target_arch = "x86_64"`? (Context: we're dropping the seccomp whitelist idea, see https://github.com/paritytech/polkadot-sdk/pull/2461 for explanation.)

mrcnski: Nvm, I dig up this comment. We're waiting on a CI job for aarch64: https://github.com/paritytech/polkadot-sdk/pull/2009#discussion_r1371695338

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4481347

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4489718

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4507508

polkadot/node/core/pvf/src/security.rs
```diff
@@ -70,6 +65,60 @@ pub async fn check_security_status(config: &Config) -> Result<SecurityStatus, St
 	Ok(security_status)
 }
 
+/// Contains the full security status including error states.
+struct FullSecurityStatus {
+	partial: SecurityStatus,
```
alindima, line 66: full/partial is rather confusing naming to me.

I think it would be better if `SecurityStatus` would have the `SecureModeResult` as fields instead of the booleans. I assume errors are strictly correlated to the security features

polkadot/node/core/pvf/common/src/worker/security/change_root.rs
```diff
@@ -59,20 +57,12 @@ pub fn enable_for_worker(worker_info: &WorkerInfo) -> Result<()> {
...
-			pid: std::process::id(),
-			kind: WorkerKind::CheckPivotRoot,
-			version: None,
-			worker_dir_path,
-		})
+	try_restrict(&WorkerInfo {
+		pid: std::process::id(),
+		kind: WorkerKind::CheckPivotRoot,
+		version: None,
+		worker_dir_path,
```
alindima, line 64: why was this not caught before?

polkadot/node/core/pvf/tests/it/main.rs
```diff
@@ -417,19 +417,26 @@ async fn prepare_can_run_serially() {
...
+		// The version will look something like "5.15.0-87-generic".
+		let version = sysinfo.linux_kernel.unwrap();
+		let version_split: Vec<&str> = version.split(".").collect();
+		let major: u32 = version_split[0].parse().unwrap();
+		let minor: u32 = version_split[1].parse().unwrap();
+		if major >= 6 {
+			true
+		} else {
+			minor >= 13
+		}
```
alindima, line 431: this'll return true for 4.13, which it shouldn't

polkadot/node/core/pvf/src/security.rs
```diff
@@ -71,12 +127,17 @@ enum SecureModeError {
...
-			CannotEnableLandlock(_) => true,
+			// Landlock is present on relatively recent Linuxes. This is optional if the unshare
+			// capability is present, providing FS sandboxing a different way.
+			CannotEnableLandlock(_) => security_status.can_unshare_user_namespace_and_change_root,
+			// seccomp should be present on all modern Linuxes unless it's been disabled.
 			CannotEnableSeccomp(_) => false,
-			CannotUnshareUserNamespaceAndChangeRoot(_) => false,
+			// Should always be present on modern Linuxes. If not, Landlock also provides FS
+			// sandboxing, so don't enforce this.
+			CannotUnshareUserNamespaceAndChangeRoot(_) => security_status.can_enable_landlock,
```
alindima, line 140: I don't see why we allow any security features to not be enabled in secure mode.
I view secure mode as an all-or-nothing type thing, with our safest options and recommendations. Otherwise, why bother having a secure mode when we could instead do a best-effort security and just log the errors, without killing the process.

polkadot/node/core/pvf/common/src/worker/security/change_root.rs
mrcnski: What do you mean @alindima?

polkadot/node/core/pvf/tests/it/main.rs
mrcnski: I thought about it, but it's labeled as a `ci-only-tests` test, and I doubt we'd ever run CI with such an old kernel.. 🤷 I can change it just for sheer correctness.

polkadot/node/core/pvf/common/src/worker/security/change_root.rs
alindima: I'm referring to the supposed error when pivoting root in a multi-threaded process

polkadot/node/core/pvf/src/security.rs
mrcnski: I see what you mean. Our approach has changed a bit over time, but we settled on this for mostly practical reasons. Since we have two FS sandboxing features, a validator can be missing one and still be reasonably safe. On the other hand, we "require" seccomp because it's our only networking sandboxing feature.

You can see in [telemetry](https://telemetry.polkadot.io/#stats/0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3) that a large amount of validators (like at least 1/3) is on Ubuntu 20.04.6 LTS. (A bit better than when I last checked in May, it was more like 50%.) Enforcing 5.13+ would "require" a *lot* of people to upgrade on a timeline that may not be convenient for them.

Remember, if a validator can't run securely he has to either upgrade his setup (which may be painful), use the insecure flag (which is not ideal), or stay on an old node version (not great either). So, we tried to minimize "breakage". We've also communicated already that only one of the FS features will be "required". We could always switch to enforcing everything, and communicate it properly in advance, but with ~Godzilla~ PolkaVM coming it may not be worth it.

A final note. Some security capabilities being present on some machines and missing on others may lead to indeterminism that can be exploited to cause disagreements/disputes/stalls. However, we are so far from determinism anyway that this was, in the end, a low priority of the security work. If we had more time, we could have done that virtualization work while iteratively making the security "requirements" more strict. And I have been putting "require" in quotes because, as Basti has said multiple times, we should not require anything. None of this is in the spec, and validators can always build from source with the barriers removed.

mrcnski: > we are so far from determinism anyway

To clarify, if an attacker can execute arbitrary code, there are already many other means by which he can obtain and exploit randomness. We decided that protecting the chain was not feasible with this work, and we have governance to deal with that sort of thing, so we prioritized securing validator machines.

polkadot/node/core/pvf/common/src/worker/security/change_root.rs
mrcnski: Are you referring to the comments? The comment is already presnt in [master](https://github.com/paritytech/polkadot-sdk/blob/cfa19c37e6/polkadot/node/core/pvf/common/src/worker/security/mod.rs#L40), I just added it to the new `pub fn`s in the refactor. This was definitely caught before, it prevents this from being unit-tested in Rust's default test apparatus which is multi-threaded.

polkadot/node/core/pvf/src/security.rs
alindima: > You can see in [telemetry](https://telemetry.polkadot.io/#stats/0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3) that a large amount of validators (like at least 1/3) is on Ubuntu 20.04.6 LTS

The telemetry page unfortunately show 70% of the kernel versions as "unknown" :( 

According to [this page](https://ubuntu.com/about/release-cycle#ubuntu-kernel-release-cycle), Ubuntu 20.04.5 should have at least kernel version 5.15.

> use the insecure flag (which is not ideal)

Well then, what's the purpose of having the insecure flag? The insecure flag should still apply security sandboxing on a best-effort basis, is that right?

Anyway, those were my 2 cents, I won't be against merging it as it is either :)

polkadot/node/core/pvf/common/src/worker/security/change_root.rs
alindima: I was referring to [this commit](https://github.com/paritytech/polkadot-sdk/pull/2486/commits/ec9bd4b52c54ffd408c4c6d8119d0f9fc3b44e95). I got confused because I was reviewing commit-by-commit

polkadot/node/core/pvf/src/security.rs
eskimor: I think Alin has a point. Even in insecure mode, still enabling in best-effort makes senses. For secure mode: As @mrcnski said, we have multiple layers of restricting filesystem access. If we assume each and every one of them is sufficient by itself, it makes sense to say the validator is secure as long as at least one of them can be enabled.

mrcnski: > The telemetry page unfortunately show 70% of the kernel versions as "unknown" :(

There was [an issue about it](https://github.com/paritytech/substrate-telemetry/issues/535) which has been completed, but I guess not deployed yet.

> According to [this page](https://ubuntu.com/about/release-cycle#ubuntu-kernel-release-cycle), Ubuntu 20.04.5 should have at least kernel version 5.15.

I got it from [here](https://askubuntu.com/questions/517136/list-of-ubuntu-versions-with-corresponding-linux-kernel-version), but I guess that is wrong. 🤷 Someone in the comments there even proved it is wrong. I guess Ubuntu "patch" versions can bump the Linux kernel version? Anyway, that is good news. I can raise a follow-up for enforcing it.

> Well then, what's the purpose of having the insecure flag? 

We want validators to be secure because a validator getting compromised is a big deal, but we have the constraint that we can't actually require it. So we just make it inconvenient and scary to run insecurely.

> I think Alin has a point. Even in insecure mode, still enabling in best-effort makes senses.

Of course, the flag only affects how we treat errors and not whether we attempt to secure ourselves. :) I can try to make this more clear.

polkadot/node/core/pvf/src/security.rs
mrcnski: Yeah, I had thought about that. What I didn't like about it is that instead of just looping over errors (which we do in a couple places) you have to check each struct field manually. That's less nice and also more error-prone, because I don't know how to do that exhaustively.

I don't know if I did it an ideal way, but `FullSecurityStatus` is designed to be easy and safe to extend, with a hopefully-nice API. And we are already thinking about [extending the security features](https://github.com/paritytech/polkadot-sdk/pull/2477#discussion_r1406110772).

Another point, is that the exact errors are not needed and unused elsewhere, so storing them in `SecurityStatus` would just make it harder to use.

polkadot/node/core/pvf/src/security.rs
alindima: > I got it from here, but I guess that is wrong. 🤷 Someone in the comments there even proved it is wrong. I guess Ubuntu "patch" versions can bump the Linux kernel version? Anyway, that is good news. I can raise a follow-up for enforcing it.

Well that's older I would assume. I assume ubuntu LTS should get updates that include kernel upgrades.

> We want validators to be secure because a validator getting compromised is a big deal, but we have the constraint that we can't actually require it. So we just make it inconvenient and scary to run insecurely.

Of course, I was suggesting to encourage them to run with all layers.

> If we assume each and every one of them is sufficient by itself, it makes sense to say the validator is secure as long as at least one of them can be enabled.

Agreed, but then why do we have landlock is we think chroot is enough 😛

alindima (approved):

polkadot/node/core/pvf/src/security.rs
mrcnski: > Agreed, but then why do we have landlock is we think chroot is enough 😛

Indeed, birdcage [removed landlock](https://github.com/phylum-dev/birdcage/pull/47) because they deemed it was redundant on top of mount namespaces.

There are a few reasons we have it:

1. A lot of this sandboxing work was a research project for me, and I found out about namespaces/pivot_root after landlock. (Still don't fully understand the former tbh.) As I understand from the above discussion, namespaces alone are merely "good enough".
2. Landlock provides finer-grained control (didn't end up being useful for us, but gives us more flexibility for the design space in the future).
3. Landlock is still expanding its capabilities and will only get more powerful with time.
4. I believe it's possible that unshare/pivot_root may not be available on some systems, so it's nice to have a fallback.
5. I don't see any downsides to having *both*, and defense-in-depth has been our philosophy since the beginning stages. It's also the approach advocated by the landlock dev in the above PR:

>  layering of mount/Landlock definitely works. I'm suggesting to use namespaces, then seccomp (locking down the availability of user namespaces and uncommon syscalls), then Landlock (partial restrictions but could still help with fine-grained access control, if supported by the running system).

alindima: > I believe it's possible that unshare/pivot_root may not be available on some systems, so it's nice to have a fallback.

I think that's highly unlikely

That being said, sounds fair 👍🏻

mrcnski: Thanks for the review @alindima!

s0me0ne-unkn0wn (approved):
Looks really good!

My only doubt is why we don't enforce the requirement for development mode (running `--dev --alice` on an insecure host warns about security but doesn't terminate). I understand it's for a better developer experience. But I ask myself again, how good is it that the DX diverges so much from the UX? We almost never test things in a production context during development. That may result in some undiscovered pitfalls and the wider the gap between DX and UX, the more pitfalls may fit into it. Just a general thought to think about.

eskimor:
> Looks really good!
> 
> My only doubt is why we don't enforce the requirement for development mode (running `--dev --alice` on an insecure host warns about security but doesn't terminate). I understand it's for a better developer experience. But I ask myself again, how good is it that the DX diverges so much from the UX? We almost never test things in a production context during development. That may result in some undiscovered pitfalls and the wider the gap between DX and UX, the more pitfalls may fit into it. Just a general thought to think about.

Good remark! Especially we should notice when the UX sucks ... The problem is, some developers us MAC and there secure mode does not work - right?

mrcnski:
> We almost never test things in a production context during development.

It's a really good point, but it's not realistic for devs to test everything. I think between the merge window closing and the actual release, a dedicated person/process should be doing actual testing of the node and we should be open to backporting fixes. I didn't mention it before because I didn't want to sound like I was making lame excuses for things like [this](https://github.com/paritytech/polkadot-sdk/pull/2426) and [this](https://github.com/paritytech/polkadot-sdk/pull/2304). But like eskimor says, I and many other devs work on a Mac. I do my due diligence by testing on a Linux VPS, but we can't expect all devs to do that (and my VPS account is root so it didn't catch [some things](https://github.com/paritytech/polkadot-sdk/issues/2458).)

Thanks for the review! 😊

mrcnski:
@pepoviola  zombienet tests are failing. I don't know how to check it but i guess it's because the new argument is required in CI. Can you help me update zombienet in CI to include the fix?

pepoviola:
> @pepoviola zombienet tests are failing. I don't know how to check it but i guess it's because the new argument is required in CI. Can you help me update zombienet in CI to include the fix?

yes,  we need a new version with the fix we merge in zombienet. Let me create a new one and update the version here.
Thx!

mrcnski:
Thanks @pepoviola! Can we re-generate the `cmd`? And not to rush you, but just note that I have to press the merge button by Wednesday to get this into the next release.

pepoviola:
Hi @mrcnski, last commit fix the upgrade test.
Thx!

polkadot/node/core/pvf/src/security.rs
mrcnski: > Indeed, birdcage removed landlock because they deemed it was redundant on top of mount namespaces.

I found out that a different project, Emilua, has taken the [opposite approach](https://lore.kernel.org/landlock/20231205.2bd704a93e21@gnoack.org/T/#t):

> Linux namespaces are too convoluted and it became clear how inappropriate they really are to just create sandboxes. Starting from the just newly released version, Landlock and Capsicum were added as alternatives to Linux namespaces.

Their [sandboxing docs](https://docs.emilua.org/api/0.5/tutorial/sandboxes.html) are really good. It's a bit too late for my work to benefit from them, oh well.

matthewmarcus:
Hey @mrcnski. Not sure if this is the right place to share this and thus ask for some guidance. So please pardon me if this is out of line (I'm not a regular github user).

After updating our Kusama validator to the latest version (1.7.0 - and now 1.7.1), we're getting errors (see below) similar to those you shared above in the original post. We're running our validator on a bare metal box running Ubuntu 20.04 with the latest kernel (6.7.5). Additionally, we have SECURE BOOT disabled in the BIOS.

We thought b/c we were running an older kernel version (5.15.0-88-generic) that updating to the latest version would solve our problem. Alas, it did not. We are able to run our node using the "insecure" flag, but obviously that isn't preferable and we'd like to keep our validator secure and not put the network at risk in any way.

Might you (or anyone else) be able to give us some guidance on what we need to do to alleviate our issue?

Errors:

`Running validation of malicious PVF code has a higher risk of compromising this machine.
   - Cannot enable landlock (ABI 1), a Linux 5.13+ kernel security feature: not available: Could not fully enable: NotEnforced
   - Cannot unshare user namespace and change root, which are Linux-specific kernel security features: not available: unshare user and mount namespaces: Operation not permitted (os error 1)
   - Optional: Cannot call clone with all sandboxing flags, a Linux-specific kernel security features: not available: could not clone, errno: EPERM: Operation not permitted`

mrcnski:
Off the top of my head I'm not sure what the problem could be. What architecture are you on? Only x86-64 is supported.

@s0me0ne-unkn0wn did the security code change in the latest release? Maybe @maksimryndin's refactor broke something?

PR #2487, opened 2023-11-24
Do not pollute global base path with export genesis/wasm

bkchr:
Otherwise the user may runs into weird errors if there is already a db.


michalkucharczyk (approved):
Reminds me my 1st problem :) (https://github.com/paritytech/substrate/pull/12254)

davxy (approved):

Issue #1389, opened 2023-09-04
[xcm-emulator] Restructure Parachains Integration Tests


NachoPal:
More info here: https://github.com/paritytech/cumulus/issues/2892

gilescope:
proposal 3: depending on the runtimes could be feature gated. This would allow the asset-hub-westend tests to only depend on the "westend-runtime", "penpal-runtime" and "asset-hub-runtime" features.

bkontur:
I started branch/PR https://github.com/paritytech/polkadot-sdk/pull/1693 with some TODOs I found on the way (feel free to continue or create new branch/PR).

**More suggestions:**

- [ ] `send_transact_native_from_system_para_to_relay_fails` - should go directly to the relevant runtime as unit-test (no need to be in xcm emulator)
- [ ] `relay_sets_system_para_xcm_supported_version` - does not test anything relevant to AssetHub, should go to the relevant relay runtime tests or as unit-test directly to the relay runtime 
- [ ] `open_hrmp_channel_between_paras_works` - does not test anything relevant to AssetHub, should go to the relevant relay runtime tests

**HRMP related suggestions:**
- [ ]  `force_open_hrmp_channel_for_system_para_works`:
- does not test anything relevant to AssetHub, we should add there relevant asserts and calls like:
  - should try to execute `PenpalKusamaA::send(msg) -> assert fails`
  - open HRMP channel to AssetHubKusama
  - should try to execute `PenpalKusamaA::send(msg) -> assert ok`
  - or some storage check on parachain side, that HRMP is correctly opened
  - check TODOs [here](https://github.com/paritytech/polkadot-sdk/pull/1693)
 - another problem is that we open HRMP channels on startup `new_block -> hrmp_channel_parachain_inherent_data`, so for this test we should check real `force_open_hrmp_channel` process
- [ ] `send_xcm_from_para_to_system_para_paying_fee_with_assets_works`
  - does not open HRMP channel (because of opening on startup `new_block`), so this test is not "complete"
  - maybe if we joing `force_open_hrmp_channel_for_system_para_works` + `send_xcm_from_para_to_system_para_paying_fee_with_assets_works` it could be one big relevant integration test

Well, yes, that automatic opening HRMP channel on `new_block` for every para is questionable, it kind of mist test.

PR #2488, opened 2023-11-24
asset-hub-westend-integration-tests: add more asset transfers tests

acatangiu:
Just adds more tests.


cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -159,19 +263,58 @@ fn reserve_transfer_native_asset_from_system_para_to_relay_fails() {
...
+		>(test.args.assets.clone(), 0, test.args.weight_limit, test.args.beneficiary, test.args.dest)
+	});
+
+	let sender_balance_after = test.sender.balance;
+	let receiver_balance_after = test.receiver.balance;
+
+	// Sender's balance is reduced
+	assert_eq!(sender_balance_before - amount_to_send - delivery_fees, sender_balance_after);
+	// Receiver's balance is increased
+	assert!(receiver_balance_after > receiver_balance_before);
```
joepetrowski, line 302: Why not?
```suggestion
	assert_eq!(receiver_balance_after, receiver_balance_before + amount_to_send - delivery_fees);
```

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/swap.rs
```diff
@@ -44,12 +44,6 @@ fn swap_locally_on_chain_using_local_assets() {
 			100_000_000_000_000,
 		));
 
-		assert_ok!(<AssetHubRococo as AssetHubRococoPallet>::Balances::force_set_balance(
-			<AssetHubRococo as Chain>::RuntimeOrigin::root(),
-			AssetHubRococoSender::get().into(),
-			100_000_000_000_000,
-		));
```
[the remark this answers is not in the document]
acatangiu: balance needed, but no need to fund here, `AssetHubRococoSender` and other accounts come prefunded from genesis

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -14,7 +14,31 @@
...
 use crate::*;
-use asset_hub_westend_runtime::xcm_config::XcmConfig;
+use asset_hub_westend_runtime::xcm_config::XcmConfig as AssetHubWestendXcmConfig;
+use penpal_runtime::xcm_config::XcmConfig as PenpalWestendXcmConfig;
+use westend_runtime::xcm_config::XcmConfig as WestendXcmConfig;
+
+fn relay_to_para_sender_assertions(t: RelayToParaTest) {
+	type RuntimeEvent = <Westend as Chain>::RuntimeEvent;
+
+	Westend::assert_xcm_pallet_attempted_complete(Some(Weight::from_parts(864_610_000, 8_799)));
```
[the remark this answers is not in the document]
acatangiu: sure, I'll change all tests to that approach

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
acatangiu: because there's also `BuyExecution` actual spent amount which we don't know (even if we did we don't want to hardcode here as weights change)

at most I would tighten this check to `receiver_balance_before < receiver_balance_after && receiver_balance_after < receiver_balance_before + amount_to_send - delivery_fees`

svyatonik (approved):

cumulus/parachains/integration-tests/emulated/chains/parachains/testing/penpal/src/lib.rs
```diff
@@ -67,5 +68,6 @@ decl_test_parachains! {
 // Penpal implementation
 impl_accounts_helpers_for_parachain!(PenpalA);
 impl_assets_helpers_for_parachain!(PenpalA, Rococo);
+impl_assets_helpers_for_parachain!(PenpalB, Westend);
```
bkontur, line 71: I am just thinking if it would be better for clarity to rename:
PenpalA -> PenpalRococo/PenpalForRococo
PenpalB -> PenpalWestend/PenpalForWestend
?
or maybe to split: `chains/parachains/testing/penpal` -> `chains/parachains/testing/penpal-rococo` and `chains/parachains/testing/penpal-westend`? Then it wont fetch both rococo and westend dependency where just one of those is needed.

wdyt?

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -159,19 +263,58 @@ fn reserve_transfer_native_asset_from_system_para_to_relay_fails() {
 	});
 }
 
+/// Reserve Transfers of native asset from Relay to Parachain should work
```
bkontur, line 266: it looks like that this test has nothing to do with AssetHub, so it should be placed elsewhere, e.g.:
`/tests/assets/westend/src/tests/reserve_transfer.rs`
or
`/tests/westend/src/tests/reserve_transfer.rs`
or
`/tests/relays/westend/src/tests/reserve_transfer.rs`
or
elsewhere?

bkontur (approved):

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
acatangiu: You have a point, but I don't see a good alternative - we have a lot of tests in both `assets` and `bridges` involving the relay chains.. pure `relay` tests are very few (right now it'd just be this one discussed here).

So I'm keeping it under `assets` - it's not an `AssetHub` test, yes, but still it is an `assets` test if you ask me..

Also `cumulus/parachains/integration-tests/emulated` already has toooo many targets and crates to build if you ask me, I really don't want to add more unless really necessary/useful.

acatangiu: This whole split of tests based on chains is not the best for integration tests that by definition use more than one chain.. we'll keep having these types of discussions, unless we start structuring them by topic/scenario, rather than chains involved.

bkontur: Yeah, I agree, we should restructure them - there is an older @NachoPal 's issue: https://github.com/paritytech/polkadot-sdk/issues/1389.
There are also various tests for AssetHubs that are not directly related to 'assets,' such as setting the XCM version or opening an HRMP channel. Some of these tests are not even related to the AssetHub itself, for example, there are tests using only the relay chain and penpal.

PR #2489, opened 2023-11-24
pallet-xcm: fix benchmarking

acatangiu:
Use non-zero weight limit in benchmarking `pallet_xcm::execute()` so the given XCM actually executes.


bkchr (approved):
I just had pushed the commit :D

joepetrowski (approved):

acatangiu:
all "short-benchmarks" are now passing, so looks like this was the only issue 👍

ggwpez (approved):
🤦‍♂️  lol

PR #2490, opened 2023-11-24
Release Process

ggwpez:
Writing down the processes to do our releases.

Status: please review & approve so we can go ahead.


docs/RELEASE.md
```diff
@@ -0,0 +1,62 @@
+# Processes
+
+The following processes are necessary to actualize our releases. Each process has a *Cadence* on which it must execute and an *Responsible* that is responsible for autonomously doing so and reporting back any error in the RelEng<sup>1</sup> channel.
+
+## Mainline Release
+
+Cadence: every two weeks. Responsible: Release Team.
+
+This process aims to release the `stable` branch as a *Mainline* release every two weeks. It needs to start three days before the proposed release deadline to allow leeway for unexpected issues. This process should eventually be automated.
```
bkchr, line 9: If this should be automated at some point, I would not add any extra "leeway". If something unexpected happens, I think it is fine that it takes longer, but "unexpected" should not be the norm. Otherwise, we need to check the process and improve it.

docs/RELEASE.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Steps
+
+1. [ ] Check if process [Clobbering](#clobbering) needs to happen and do so first, if that is the case.
+1. [ ] Check out the latest commit of `stable`.
+2. [ ] Verify all CI checks of that commit.
+3. [ ] Announce that commit as cutoff *Mainline* for a release in the General<sup>2</sup> chat.
```
bkchr, line 16: As there should be way less merges to `stable`, I would say that the release is actually done as some pr against the `stable` branch. IMO this gives more visibility and also makes it more clear on what is happening.

docs/RELEASE.md
```diff
@@ -0,0 +1,62 @@
...
+4. [ ] Bump the semver of all crates <!-- FAIL-CI: We need some better process here on how to do it exactly -->
+5. [ ] Abort the release process and announce so in General if there are no bumps needed.
+6. [ ] Create a merge request to `stable` with the proposed SemVer bumps.
```
bkchr, line 19: Should be obsolete with what I said above.

docs/RELEASE.md
```diff
@@ -0,0 +1,62 @@
...
+The following script is provided to do the clobbering.
+
+```bash
+git checkout polkadot-sdk -b audited
+mv polkadot-sdk polkadot-sdk-audited
+git checkout polkadot-sdk -b stable
+cp -r polkadot-sdk-audited/* polkadot-sdk
+cd polkadot-sdk
+git commit -a -m "Clobber with latest audited"
+git push
```
bkchr, line 56:
```suggestion
git checkout stable
git reset --hard audited
git push -f
```

Shorter and ensures we don't have any stale files :P

docs/RELEASE.md
```diff
@@ -0,0 +1,62 @@
...
+
+This process aims to release the `stable` branch as a *Mainline* release every two weeks. It needs to start three days before the proposed release deadline to allow leeway for unexpected issues. This process should eventually be automated.
+
+### Steps
+
+1. [ ] Check if process [Clobbering](#clobbering) needs to happen and do so first, if that is the case.
+1. [ ] Check out the latest commit of `stable`.
+2. [ ] Verify all CI checks of that commit.
+3. [ ] Announce that commit as cutoff *Mainline* for a release in the General<sup>2</sup> chat.
+4. [ ] Bump the semver of all crates <!-- FAIL-CI: We need some better process here on how to do it exactly -->
```
bkchr, line 17: This should be really not done. IMO we should bump the crate versions as we merge stuff to `unstable`.

I had written here some things: https://github.com/paritytech/polkadot-sdk/pull/2366

I mean I know that is not a perfect thing, but probably the best we can do. IMO this should not be done by a single person or after the fact as this is otherwise too hard to follow.

bkchr (commented):
Thank you! 

Have done a first pass and left some comments. IMO, we should make the process less depended on posting something into some internal channels. This would be totally opaque to the outside and not that good.

Otherwise it already is already okay!

One thing we should think about is to name things which are "unstable" internal api. Aka things that can be merged to master faster even when it requires a major version bump. I'm thinking about the Polkadot node internals that are not really exposed.

docs/RELEASE.md
ggwpez: Yea I was about to change it to exactly that as well 😆, the first version was just copied from the hackmd.

docs/RELEASE.md
ggwpez: But if we bump the version of each crate (and all its transitive dependants) in each MR, then we will have huge merge conflicts.  
I think if we instead note the version bump in the `prdoc` then it can happen automatically at release process.

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
...
+Audits are conducted to ensure the absence of severe or exploitable bugs. MRs are generally merged without audit into the `master` branch. The `audited` tag is used to track the latest audited commit of the `master` branch. This means that audits need to happen in order of being merged.  
+This is an optimistic approach that lets us develop with greater speed, while requiring (possibly) large refactors in the failure case.
+
+Audits can be deferred if the logic is gated by an `experimental` feature or marked as "Not Production Ready" within the first line of doc. Such changes should be queued manually before these warnings are removed.
+
+## General Guidelines for what to Audit
+
+There is no single one-fits-all rule. Generally we should audit important logic that could immediately be used on production networks. If in doubt; ask in chat or in the Merge Request.
+
+## Requesting an Audit
```
ggwpez, line 12: @0xJayPi i copied this from the chat. I think its important to mention the process for this somewhere in the repo, so we can easily link it to (outsider) contributors.

docs/RELEASE.md
ggwpez: > IMO, we should make the process less depended on posting something into some internal channels. This would be totally opaque to the outside and not that good.

Yea right. The channel pings would be mostly for us to react to issues and monitoring, but having it as comments in a MR is probably better.

docs/RELEASE.md
bkchr: > It would also require is to backport the version bumps from crates-io, leading to even more conflicts.

Hmm? Why would we need to backport version bumps? I mean the version isn't bumping automatically. 



> I think if we instead note the version bump in the `prdoc` then it can happen automatically at release process.

If you change something deep inside the tree, I don't want to copy past 300 crate names and their bumps to a prdoc file...



> But if we bump the version of each crate (and all its transitive dependants) in each MR, then we will have huge merge conflicts.

We should not bump in every pr. Bumps should only be done as they are required between releases.

ggwpez:
>One thing we should think about is to name things which are "unstable" internal api. Aka things that can be merged to master faster even when it requires a major version bump. I'm thinking about the Polkadot node internals that are not really exposed.

So it would only change the behaviour of a polkadot node, but not the behaviour of any of our public APIs?  
Is the use-case for this just bug fixes to the node or also adding features?  
Not sure if we need to include CLI flags, would probably be good if we consider them part of our external API.

docs/RELEASE.md
ggwpez: >Hmm? Why would we need to backport version bumps? I mean the version isn't bumping automatically.

Yea I see. I thought about using `prdoc` to not the semver bump of each crate that i changed, and a script can then take the maximum of all bumps and apply them.

>If you change something deep inside the tree, I don't want to copy past 300 crate names and their bumps to a prdoc file...

It could suffice to note the changed crates - all dependant ones could bump automatically.

>We should not bump in every pr. Bumps should only be done as they are required between releases.

Okay yea i read it in the other MR from you. So:
- developer changes one crate and bumps that.  
- CI reminds him to not bump multiple times but to also bump all dependant crates
- CI checks that there are no obvious and unintended breaking changes

docs/RELEASE.md
```diff
@@ -0,0 +1,61 @@
+# Processes
+
+The following processes are necessary to actualize our releases. Each process has a *Cadence* on which it must execute and an *Responsible* that is responsible for autonomously doing so and reporting back any error in the RelEng<sup>1</sup> channel.
+
+## Mainline Release
+
+Cadence: every two weeks. Responsible: Release Team.
+
+This process aims to release the `release` branch as a *Mainline* release every two weeks. It needs to start three days before the proposed release deadline to allow leeway for unexpected issues. This process should eventually be automated.
```
mrcnski, line 9: So the merge window is now three days and not a week?

bkchr: No. Two weeks minus three days is 11 days. But taking the "missing three days" from the previous release cycle into account, you end up with 14 days again.

docs/RELEASE.md
bkchr: > It could suffice to note the changed crates - all dependant ones could bump automatically.

Bumping the dependent crates is not that easy. I mean you only need to bump the dependent crates if you re-export anything as public api.

README.md
```diff
@@ -35,6 +35,21 @@ Polkadot.
 
 Cumulus is a set of tools for writing Substrate-based Polkadot parachains.
 
+## Releases
+
+> [!NOTE]  
+> Our release process is still Work-In-Progress and may not yet reflect the aspired outline here.
+
+The Polkadot-SDK has two release channels: `mainline` and `nightly`. Production software is advised to only use `mainline`. `nightly` is meant for tinkers to try out the latest features. The detailed release process is described in [RELEASE.md](docs/RELEASE.md)
```
joepetrowski, line 43:
```suggestion
The Polkadot-SDK has two release channels: `mainline` and `nightly`. Production software is advised to only use `mainline`. `nightly` is meant for tinkerers to try out the latest features. The detailed release process is described in [RELEASE.md](docs/RELEASE.md)
```

README.md
```diff
@@ -35,6 +35,21 @@ Polkadot.
...
+
+### Mainline
+
+`mainline` releases have a support duration of **three months**. In this period, the release will not have any breaking changes but only receive bug- and security fixed on a **two week** cadence.
```
joepetrowski, line 47:
```suggestion
`mainline` releases have a support duration of **three months**. In this period, the release will not have any breaking changes but only receive bug and security fixes on a **two week** cadence.
```

README.md
```diff
@@ -35,6 +35,21 @@ Polkadot.
...
+
+### Nightly
+
+`nightly` releases are released every night with possibly breaking changes and offer **no SemVer** guarantee between each other. They have pre-releases version numbers in the format `major.0.0-nightlyYYMMDD`.
```
joepetrowski, line 51:
```suggestion
`nightly` releases are released every night with possibly breaking changes and offer **no SemVer** guarantee between each other. They have pre-release version numbers in the format `major.0.0-nightlyYYMMDD`.
```

joepetrowski (commented):
Would be nice to add a note about testnet updates. Do they update on nightly or mainline?

bkchr:
> Would be nice to add a note about testnet updates. Do they update on nightly or mainline?

They need to be mentioned explicitly! Good point!

docs/RELEASE.md
ggwpez: Okay I will remove it.

docs/RELEASE.md
ggwpez: But when we change something in `sp-io` for example, without modifying the public API but some implementation detail, then we should bump all dependants, or?  
I am not sure if its easy to manually select which crates to bump, since there is the chance of missing some. If we always bump all dependants then maybe we bump too many, but at least we never miss any.

docs/RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+The versioning of the node is done 99% of the time by only incrementing the `minor` version. 
+The `major` version is only bumped for special releases and the `patch` can be used for an 
+out of band release that fixes some critical bug. The node version is not following SemVer. 
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in 
+`polkadot/node/primitives/src/lib.rs`.
+
+### Westend & Rococo
+
+For the these networks, we only increment the `spec_version`. The spec version is also following
```
EgorPopelyaev, line 28: Just as addition, we do check and bump `transaction_version` as we did before for both as well

docs/RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+3. [ ] Announce that commit as cutoff *Mainline* for a release in the General<sup>2</sup> chat.
+4. [ ] Bump the semver of all crates <!-- FAIL-CI: We need some better process here on how to do it exactly -->
+5. [ ] Abort the release process and announce so in General if there are no bumps needed.
+6. [ ] Create a merge request to `release` with the proposed SemVer bumps.
+7. [ ] Announce this merge request in the *General* channel to quickly gather reviews.
+8. [ ] Merge it into `release`.
+9. [ ] Verify all CI checks.
+10. [ ] Announce the intent to do a *Mainline* release from the resulting commit hash in RelEng.
+11. [ ] <!-- The release team has internal checklists for QA i think, should we mention this? -->
+12. [ ] Release all crates to crates.io using [parity-publish](https://github.com/paritytech/parity-publish).
```
liamaharon, line 74: Why release every crate even if no changes? I am not sure what advantages `parity-publish` has over a tool like https://github.com/crate-ci/cargo-release if we are following proper versioning practices.

ggwpez: I think the other tools were tired out by the Release team and did not work with our massive workspace.  
I will change the sentence to only publish the changed crates yes - but in the end i dont care how the publishing happens 🙈

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
+# Audit
+
+Audits are conducted to ensure the absence of severe or exploitable bugs. MRs are generally merged without audit into the `master` branch. The `audited` tag is used to track the latest audited commit of the `master` branch. This means that audits need to happen in order of being merged.
```
athei, line 3: Why is `audited` a tag and not a branch? Changing where a tag points to breaks with expectations. Why not use a branch, instead? A branch is meant to point to a commit and is expected to change where it points to. I guess there are reasons for both ways. Just looking for a rational.

docs/RELEASE.md
liamaharon: If they didn't work as intended for our workspace I'd hope we opened some issues in their repos to report the bugs, so the problems can be fixed?

Anyway, all these tools are new or under active development so hopefully at least one of them works for us now.

bkchr: @Morganamilo can probably answer this. However, I also think that our workspace size is quite unique :P

docs/RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+### Westend & Rococo
+
+For the these networks, we only increment the `spec_version`. The spec version is also following
+the node version. The schema is as follows: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to the release in this repository. 
+
+## Backports
+
+Backports should most of the time not be required. We should only backport critical bug fixes and then release the fixed crates. There should be no need to backport anything from a release branch.
+
+When a backport is required for some previous release, it is the job of the developer (assuming it is some internal person) that has created the initial PR to create the backports. After the backports are done, it is important to ensure that the crate release is being done. We should backport fixes to the releases of the last 6 months.
```
gabreal, line 35: if i understand correctly the stable branch will be clobbered every 3 months and there will be no branch left pointing to a paritcular release. how will it be possible to backport changes to the oldstable release?

docs/AUDIT.md
ggwpez: Yes I am not sure what is better here. A branch would make sense since it is going to change often. But on the other hand, a tag is used to reference a commit on another branch, which is also what we are doing.  

Dont know if this is the best approach, but i think we will find out quickly if it needs improving.

docs/RELEASE.md
ggwpez: I changed it now to make the developer update the crates versions in their Merge Requests (where necessary).

liamaharon: FYI working on a tool for this https://github.com/liamaharon/cargo-workspace-version-tools

docs/RELEASE.md
gabreal: In understand that the idea is not to force people building on top of polkadot-sdk not to deal with a breaking change every two weeks. if there is no backporting then there will be such a breaking change every 3 months. users will be required to upgrade right once a new release has been made if there will be no way to backport changes to the previous stable release.

ggwpez: Yes, this is also being discussed in the forum https://forum.parity.io/t/stable-frame-releases/2142/31

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+This process aims to bring branch `release` in sync with the latest audited commit of `master`. It is not done via a Merge Request but rather by just copying files. It should be automated.
+
+The following script is provided to do the clobbering.
+
+```bash
+git fetch
+git checkout release
+git reset --hard origin/audited
+git push --force release
+```
```
liamaharon, line 105: I'm not sure about this clobber process, I think it will cause us to lose commit history of old releases?

Instead, I suggest to make clobbering mean creating a new commit on the `release` branch to make the tracked files match **exactly** with what's at the `audited` tag, but keep the history of the `release` branch intact.


This could be achieved like this:

```bash
# Switch to release branch
git checkout release

# Delete all tracked files in the working directory
git ls-files -z | xargs -0 rm -f

# Find and delete any empty directories
find . -type d -empty -delete

# Checkout all files from master into the working directory
git checkout master -- .

# Stage, commit, and push the working directory which now matches master 1:1
git add .
git commit -m "some release message"
git push
```

This method is similar to what Gav proposed at https://hackmd.io/XD28Qh0nSzylw1v9ktALMg?both, but will make sure any deleted files from `master` are also deleted when we make the changes to release.

README.md
```diff
@@ -35,6 +35,21 @@ Polkadot.
 
 Cumulus is a set of tools for writing Substrate-based Polkadot parachains.
 
+## Releases
+
+> [!NOTE]  
+> Our release process is still Work-In-Progress and may not yet reflect the aspired outline here.
+
+The Polkadot-SDK has two release channels: `mainline` and `nightly`. Production software is advised to only use `mainline`. `nightly` is meant for tinkerers to try out the latest features. The detailed release process is described in [RELEASE.md](docs/RELEASE.md)
```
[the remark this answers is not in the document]
ggwpez: @gavofyork originally named it `Mainline`, but i agree that we could get away here with one word less.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
+# Release
+
+The output of a release are the `polkadot` node, runtimes for the Westend & Rococo networks and new versions of the crates published to `crates.io`.
+
+## Versioning
+
+We are releasing multiple different things from this repository in one release, but 
+we don't want to use the same version for everything. Thus, in the following we explain
+the versioning story for the crates, node and Westend & Rococo. To easily refer to a release, we shall use the node version of it.
```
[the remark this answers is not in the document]
ggwpez: Yea i think this would be easier as well. I copied this from @bkchr MR here https://github.com/paritytech/polkadot-sdk/pull/2366.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+Cadence: (possibly) each Merge Request. Responsible: Developer that opened the MR.
+
+Following SemVer isn't easy, but there exists [a guide](https://doc.rust-lang.org/cargo/reference/semver.html) in the Rust documentation that explains the small details on when to bump what. This process should be augmented with CI checks that utilize [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) and/or [`cargo-public-api`](https://github.com/Enselic/cargo-public-api). They must also pay attention to downstream dependencies that require a version bump, because they export the changed API.
+
+### Steps
+
+1. Developer opens a Merge Request with changed crates against `master`.
+2. They bump all changed crates according to SemVer.
+3. They bump all crates that export any changed types in their *public API*.
+4. They also bump all crates that inherit logic changes from relying on one of the bumped crates.
```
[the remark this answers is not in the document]
ggwpez: I was trying to describe the normal workflow of merging a standard Merge request.  

Currently we never touch the `Cargo.toml` version. But in the future we would need to increment it according to SemVer. This can be a `major`, `minor` or `patch` bump - depending on the change.  

Otherwise when would we bump these versions? We dont really "merge" to the release branch, unless its a backport.

docs/RELEASE.md
ggwpez: >I'm not sure about this clobber process, I think it will cause us to lose commit history of old releases?

Yes. But we dont backport to them. Or is your concern that the code will not easily be accessible?  
Then we could tag it with the version.

docs/RELEASE.md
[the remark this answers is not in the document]
liamaharon: My understanding was

- merge to `master`: don't need to do bumps on merge. nightly releases will be made nightly bumping the semver with a new `-nightlyYYYY-MM-DD` suffix. I don't see benefit of bumping X.Y.Z between nightly releases, because the code will have a breaking bump (-nightly change) soon anyway. 
- merge to `release`: we need to bump according to X.Y.Z semver, because we need to make new releases for the changed crates.

docs/RELEASE.md
liamaharon: Yes my concern is the code from old releases can be completely removed. 

Using `push --force` and relying only on tags seems dangerous to me, if we forget to make a tag or accidentally delete/overwrite one we lose commit history on remote. 

With my approach we get all the same benefits (I think?) without the danger of accidentally losing history. 

I'm not sure what it has to do with backporting.

docs/RELEASE.md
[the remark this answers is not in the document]
liamaharon: This is how I understood it

```
stable
v1.0.0 -> v1.0.1 -------------------------------> v2.0.0
                                                    ^
                                             CLOBBER|
master                                              |
v2.0.0-nightly2023-01-01 -> ... v2.0.0-nightly2023-04-01 -> v3.0.0-nightly2023-04-02 -> ...
```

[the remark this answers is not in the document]
liamaharon: > But then how do we keep track of these in the meantime?

Keep track of what in the meantime?

> Note that we dont need to bump multiple times on master if there was no release so far.

Yup, it should only create a new nightly build if there have been changes. That should be automated. 

> We would also need to bullet proof automate these version bump then. I would also like to do this with prdoc and so on, but @bkchr seems to be in favour of just bumping them directly.

I assume you mean automate version bumps in stable?

Patch bumps should be easy enough to automate. I plan to add a subcommand to https://github.com/liamaharon/cargo-workspace-version-tools where you specify a crate to patch, and it'll identify all dependants of that crate (direct or via other crates). Once it has all the dependants, it bumps all the Cargo.toml versions. Then in CI, it should detect Cargo.toml version bumps and release those crates (I can look into a script for this if the release team is unable to).

We just need to be very careful to not accidentally make a breaking change in these patches. I'm thinking we use a combination of those two semver check libraries we've discussed to inspect the public api before/after. Maybe a bot can post a diff of the public API in the PR to make it easier to view.

pavelsupr:
@bkchr @ggwpez, 

Security wanted to propose one enhancement, targeted on the release pipelines from the integrity perspective. 

Looks like, this proposition doesn't affect the subject of this thread too much, but it can cause some minor adjustments for this thread.

Could you please check [this issue](https://github.com/paritytech/security/issues/18) for the details, and comment on that issue directly? Since this PR is public. Here we can post the final outcomes only.

docs/RELEASE.md
liamaharon: > Bumping the dependent crates is not that easy. I mean you only need to bump the dependent crates if you re-export anything as public api.

This isn't true when making breaking version bumps. Types are not the same between major versions, even if they have the exactly same method signature and api https://doc.rust-lang.org/cargo/reference/resolver.html#version-incompatibility-hazards

I just stumbled across this https://github.com/dtolnay/semver-trick/tree/master which 

> refers to publishing a breaking change to a Rust library without requiring a coordinated upgrade across its downstream dependency graph.

Kinda hacky, but what we need to not be stuck in the hell of major bumping dependant crates.. Would need to some time spent thinking about how to implement this with minimum friction, however.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+1. Check out the latest commit of branch `master`.
+2. Verify all CI checks of that commit.
+3. Compare this commit to the latest `nightly*` tag and abort if there are no changes detected.
+4. Set the version of all crates to `major.0.0-nightlyYYMMDD` where `major` is the last released `major` version of that crate plus one.
+5. Tag this commit as `nightlyYYMMDD`.
+9. Do a dry-run release to ensure that it *should* work.
+7. Push this tag (the commit will not belong to any branch).
+8. Announce the intent to do a *Nightly* release from that tag in the RelEng chat.
+9. Release all crates that had changed since the last nightly release to crates.io.
+10. Create a release on GitHub.
```
EgorPopelyaev, line 90: How should those nightly releases look like on github? The same way as the current releases? I mean, should the node binaries and runtimes be built and attached as well? And what about `polkadot-parachain` (a.k.a `cumulus`) binary and the parachains runtimes, should they also follow the same cadence and process?

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+👉 The public API of our library crates is defined as all public items that are not inside a `__private` module.
+
+### Node
+
+The versioning of the node is done 99% of the time by only incrementing the `minor` version. 
+The `major` version is only bumped for special releases and the `patch` can be used for an 
+out of band release that fixes some critical bug. The node version is not following SemVer. 
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in 
+`polkadot/node/primitives/src/lib.rs`.
```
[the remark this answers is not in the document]
bkchr: > 1. Why doesn't the node follow semver?

What would be the point and what would be included in this? RPC? Cli? Some internal changes?



> 2\. Why do we put the node version in `NODE_VERSION` instead of just in the Cargo.toml?

https://github.com/paritytech/polkadot-sdk/pull/1495#issuecomment-1715483786

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+For the these networks, in addition to incrementing the Cargo.toml version we also increment the `spec_version` and sometimes the `transaction_version`. The spec version is also following
+the node version. Its schema is: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to the release in this repository. 
+
+## Backports
+
+Backports should be rare. We should only backport [critical bug fixes](#bug-and-security-fix) and then release the fixed crates. There should be no need to backport anything from a release branch.
+
+# Processes
+
+The following processes are necessary to actualize our releases. Each process has a *Cadence* on which it must execute and an *Responsible* that is responsible for autonomously doing so and reporting back any error in the RelEng<sup>1</sup> channel.
```
bkchr, line 37:
```suggestion
The following processes are necessary to actualize our releases. Each process has a *Cadence* on which it must execute and a *Responsible* that is responsible for autonomously doing so and reporting back any error in the RelEng<sup>1</sup> channel.
```

docs/RELEASE.md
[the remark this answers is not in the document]
bkchr: > Patch bumps should be easy enough to automate

This isn't the problem. The problem is more to bump versions properly for breaking changes.



> I would also like to do this with prdoc and so on, but @bkchr seems to be in favour of just bumping them directly.

I'm really not happy with either solution. Prdoc will require to copy X crates names and the version bump it requires. While on the other side we may bump a crate version too often in the `Cargo.toml` on master. I have pushed back for the prdoc option, but if you insist I release my "block" on this. I was imagining that we have some extra CI job that informs us if we for example want to bump the major version multiple times on master.


Orrr.... We just scrap versions completely from master and ensure that when we backport stuff to stable that versions are bumped properly. We could use these semver tools to help us on these bumps. I didn't tried them yet, but we can see how good they work and if not improve them on the way. WDYT?

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+Cadence: every two weeks. Responsible: Release Team.
+
+This process aims to release the `release` branch as a *Mainline* release every two weeks. It should eventually be automated.
+
+### Steps
+
+1. Check if process [Clobbering](#clobbering) needs to happen and do so, if that is the case.
+2. Check out the latest commit of `release`.
+3. Verify all CI checks of that commit.
+4. Update the `CHANGELOG.md` version and date. <!-- Here we could aggregate the PrDocs into the CHANGELOG -->
```
bkchr, line 63:
```suggestion
4. Update the `CHANGELOG.md` version, date and compile the content using the prdoc files.
```

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+
+Cadence: every two weeks. Responsible: Release Team.
+
+This process aims to release the `release` branch as a *Mainline* release every two weeks. It should eventually be automated.
+
+### Steps
+
+1. Check if process [Clobbering](#clobbering) needs to happen and do so, if that is the case.
+2. Check out the latest commit of `release`.
+3. Verify all CI checks of that commit.
```
bkchr, line 62: Why? This should already have done the last time something got merged into release?

docs/RELEASE.md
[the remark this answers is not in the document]
bkchr: Whatever we do, we loose "visibility" any way on master on which code exactly is on crates.io.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+5. Open a Merge Request against `release` for visibility.
+6. Check if there were any changes since the last release and abort, if not.
+7. Run `cargo semver-checks` and `cargo public-api` again to ensure that there are no SemVer breaks.
+8. Internal QA from the release team can happen here.
+9.  Do a dry-run release to ensure that it *should* work.
+10. Merge it into `release`.
+11. Verify all CI checks.
+12. Comment that a *Mainline* release will happen from the merged commit hash.
+13. Release all changed crates to crates.io.
+14. Create a release on GitHub.
```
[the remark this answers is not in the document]
bkchr: 100%!! All of this should be done by some bot. Only the final merge should be  done by a human. Dry run of a release would just be CI `cargo build --release` and then run all checks. We should also potentially add more checks as required.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+9. Do a dry-run release to ensure that it *should* work.
+10. Merge it into `release`.
+11. Verify all CI checks.
+12. Comment that a *Mainline* release will happen from the merged commit hash.
+13. Release all changed crates to crates.io.
+14. Create a release on GitHub.
+
+## Nightly Release
+
+Cadence: every day at 00:00 UTC+1. Responsible: Release Team
```
bkchr, line 77: CET superiority!!! HAHA ;D

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+
+This process aims to release the `master` branch as a *Nightly* release. The process can start at 00:00 UTC+1 and should automatically do the following steps.
+
+1. Check out the latest commit of branch `master`.
+2. Verify all CI checks of that commit.
```
bkchr, line 82: Again, too late :P

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+3. Compare this commit to the latest `nightly*` tag and abort if there are no changes detected.
+4. Set the version of all crates to `major.0.0-nightlyYYMMDD` where `major` is the last released `major` version of that crate plus one.
+5. Tag this commit as `nightlyYYMMDD`.
+9. Do a dry-run release to ensure that it *should* work.
+7. Push this tag (the commit will not belong to any branch).
+8. Announce the intent to do a *Nightly* release from that tag in the RelEng chat.
```
bkchr, line 88: And then you assume that someone answers them at 3am? :P 

This entire process should be 100% automated. Just push the shit out there. No questions asked or whatever.

docs/RELEASE.md
bkchr: Nightly should be only for crates or @ggwpez? So, we probably only need some tag on github.

docs/RELEASE.md
bkchr: > if we forget to make a tag or accidentally delete/overwrite one we lose commit history on remote.

Git is a decentralized version control system. Whoever would have pushed this, can just go back using `git reflog`.

bkchr: Whatever way we are using, we will need to extend the steps anyway as we will also need to bump the crate versions properly etc. Similar to what we do for the mainline release.

docs/RELEASE.md
```diff
@@ -0,0 +1,126 @@
...
+`polkadot/node/primitives/src/lib.rs`.
+
+### Westend & Rococo
+
+For the these networks, in addition to incrementing the Cargo.toml version we also increment the `spec_version` and sometimes the `transaction_version`. The spec version is also following
+the node version. Its schema is: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to the release in this repository. 
+
+## Backports
+
+Backports should be rare. We should only backport [critical bug fixes](#bug-and-security-fix) and then release the fixed crates. There should be no need to backport anything from a release branch.
```
bkchr, line 33: What about new features or performance improvements? Not having any progress for 3 months is quite long. I mean as long as these new things to do not break APIs or whatever.

bkchr (commented):
Generally looking good.

Still some questions around backporting. I think we should also backport other things, that do not break APIs. If in between there was a big change on master, we can probably stop backporting for most of the stuff until we do next clobber. We also should highlight on what is internal api, like certain crates from Polkadot node and thus they are fine to break/do major releases in between on the release branch.

And what about westend? This will follow the `release` branch?

docs/RELEASE.md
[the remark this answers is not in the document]
liamaharon: > What would be the point and what would be included in this? RPC? Cli? Some internal changes?

from the PoV of node operators, with node versions I would expect

- Patch: I can blindly update my node. No new features, something like a bug fix or perf improvement
- Minor: I can blindly update my node. Maybe there are some new CLI args / RPC endpoints, but no breakages
- Major: I should be careful updating my node. There have been breaking changes, it could be something about the CLI args or RPC format has changed

If you ask node operators, I think they would appreciate semver signalling like this?

docs/RELEASE.md
liamaharon: Yeah if there's something we think would be useful to get out earlier makes sense to add it to both `stable` and `master`.

docs/RELEASE.md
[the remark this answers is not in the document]
liamaharon: > Whatever we do, we loose "visibility" any way on master on which code exactly is on crates.io.

CI has to pick a commit to deploy from anyway, it shouldn't be hard for CI every nightly release to bump the `Cargo.toml` versions YYYY-MM-DD, commit it, and tag the commit?

> Orrr.... We just scrap versions completely from master and ensure that when we backport stuff to stable that versions are bumped properly. We could use these semver tools to help us on these bumps. I didn't tried them yet, but we can see how good they work and if not improve them on the way. WDYT?

My understanding from the forum post is this was the plan yes: `master` is never bumped manually, it's only bumped by the CI every nightly release.

For example, if a crate version in `stable` is 1.2.1, `master` would be set to 2.0.0-nightlyYYYY-MM-DD. 

Then every 6th release, we remove the `-nightlyYYYY-MM-DD` from the versions and that becomes our new `stable` branch. 

This way, we don't need to worry at all about breaking changes when we merge to `master`.

A downside of this process is that we blindly push major bumps for every crates every 6th release even if there were no breaking changes. HOWEVER this is still 10x better than what we do now, so I think it makes sense to get this process out now and make the diminishing returns improvements later iteratively, rather than trying to get something perfect all at once.

docs/RELEASE.md
liamaharon: > Git is a decentralized version control system. Whoever would have pushed this, can just go back using git reflog.

If I'm trying to debug an old release I really don't want to be DMing people asking for them to push their old code up for me.. It should just be in the remote? What's the downside to my approach or Gav's approach where we retain commit history on `stable`?

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: I changed it to `stableYYMMDD` now, so that it is consistent with the `nightlyYYMMDD` tag.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4661483

docs/RELEASE.md
ggwpez: We can also backport features/improvements, but only after they were audited. I will add a section for it.

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: It could work for `master`, and it does sound quite a lot easier for a start. But then we loose the ability to backport features/improvements, since we have no overview when a breaking change merged to master.

docs/RELEASE.md
ggwpez: Yea it should not be needed, it mostly in case to spot a broken master CI (which happens from time to time).

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: Yea sure. That is why i wrote `It should eventually be automated`. All of this can be automated.  

With dry run i meant `cargo release --dry-run` for all crates in the correct order. Not sure if this is possibly without mocking crates-io (like JP did) but i hope so.

docs/RELEASE.md
ggwpez: So with your approach we still have the backport MRs into the `release` branch?  
Can we not just push a tag with the release commit? I imagine the `release` branch would look quite weird otherwise, since we have all the backport MRs and then these "clobber commits" that basically revert and reapply a lot of stuff.

docs/RELEASE.md
ggwpez: Yes AFAIK its only for crates. We could create a pre-release, but just a tag should be fine for now, or?

ggwpez:
>And what about westend? This will follow the release branch?

added: `The Westend testnet will be updated to the new runtime version immediately after a mainline release happened`.
In theory we could also update it on nightly, but starting with the `mainline` is good for now i think.

>We also should highlight on what is internal api, like certain crates from Polkadot node and thus they are fine to break/do major releases in between on the release branch.

Can you give me a list or something?

docs/RELEASE.md
bkchr: Yeah good point, did had thought about the missing history that much!

docs/RELEASE.md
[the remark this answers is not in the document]
bkchr: > Then every 6th release, we remove the `-nightlyYYYY-MM-DD` from the versions and that becomes our new `stable` branch.

So we would still bump only major every 3 months? 



> On the other hand if we put the crate bumps into the PRdoc, then we can easily backport stuff that has only `minor/patch` in it. Would be even easier than bumping the `Cargo.toml` directly, since the Prdoc gives us the semver-diff.

I mean I see the advantage and maybe we should do it and live with copying over the bumps to the prdoc. The problem is still that we could forget to mention things that are require a bump. I would really first like to try out the tooling and then see what we should do. I mean the tooling probably already works for most of the things we are doing as breaking changes, like changing a function name or removing a parameter.

docs/RELEASE.md
```diff
@@ -0,0 +1,136 @@
...
+we don't want to use the same version for everything. Thus, in the following we explain
+the versioning story for the crates, node and Westend & Rococo. To easily refer to a release, it shall be named by its date in the form `stableYYMMDD`.
+
+## Crate
+
+We try to follow SemVer<sup>1</sup> as best as possible for versioning our crates. SemVer requires a piece of software to first declare a public API. The public API of the Polkadot SDK is hereby declared as the sum of all crates' public APIs, with the exception of the following internal crates:
+
+- @bkchr the list of node internals please
+
+The public API of our library crates is declared as all public items that are not inside a `__private` module.
```
gavofyork, line 22: ...And also not marked explicitly as "unstable" either in the first line of documentation or within the API entity's name or path.

docs/RELEASE.md
```diff
@@ -0,0 +1,136 @@
...
+The `major` version is only bumped for special releases and the `patch` can be used for an 
+out of band release that fixes some critical bug. The node version is not following SemVer. 
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in 
+`polkadot/node/primitives/src/lib.rs`.
+
+## Westend & Rococo
+
+For the these networks, in addition to incrementing the Cargo.toml version we also increment the `spec_version` and sometimes the `transaction_version`. The spec version is also following
+the node version. Its schema is: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to the release in this repository.
```
gavofyork, line 36: I seem to remember our versioning has historically been MMmmppxx

docs/RELEASE.md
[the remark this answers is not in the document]
gavofyork: > So we would still bump only major every 3 months?

In my proposal, yes.

In my proposal:

- Major bumps on `release` every 3 months when the Clobber happens. SemVer respected.
- Minor bumps on `release` every 2 weeks for any crate touched or any crate depending on such a crate. No backports are allowed to make breaking changes. SemVer respected.
- Patch bumps on `release` for out-of-band security fixes which have been reviewed as non-breaking. SemVer respected.
- Nightly just has the next major version with a `-nightly` suffix. SemVer respected.

docs/RELEASE.md
ggwpez: There was some discussion to rather go with triplets so that it formats nicer as Rust integer, because of the underscores.  
Or do you prefer the other schema?

docs/RELEASE.md
liamaharon: > So with your approach we still have the backport MRs into the release branch?

What makes you think we need to backport? I think there's still some confusion. 

My suggestion is very simple; every major release to `stable`, a single new commit is added to make it match the `master` branch 1:1 (clobbering). This can be entirely automated and does not require any rebasing/merge conflicts. No existing commits are overridden.

docs/RELEASE.md
gavofyork: Don't have a huge opinion, but we have in the past put release candidates on-chain, particularly in testnets, and this needed a version number in between two patch releases.

docs/RELEASE.md
gavofyork: This sounds very much like my original (The Clobber) proposal.

liamaharon: Yep @gavofyork, it's a slightly tweaked version of your original proposal. See https://github.com/paritytech/polkadot-sdk/pull/2490#discussion_r1414045936.

> This method is similar to what Gav proposed at https://hackmd.io/XD28Qh0nSzylw1v9ktALMg?both, but will make sure any deleted files from master are also deleted when we make the changes to release.

ggwpez: Okay I will change it.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827795

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827796

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827794

docs/RELEASE.md
ggwpez: Hm, we could change the spec-version to a `(u8, u8, u8, u8)` tuple with these `MMM, mmm, ppp, xxx` semantic.  
Currently we just assume one ever-increasing integer, but since its structured information, IMHO it would make sense to represent it as such.

Otherwise yea, MMmmppxx would also work.

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: > I would really first like to try out the tooling and then see what we should do. I mean the tooling probably already works for most of the things we are doing as breaking changes, like changing a function name or removing a parameter.

Okay then lets try the native bumping without Prdoc for now.

README.md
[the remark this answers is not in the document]
ggwpez: Changed it now to `stable` and the `release` branch as well.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4888004

docs/RELEASE.md
```diff
@@ -0,0 +1,155 @@
...
+- @bkchr the list of node internals please
+
+Inductively, the public API of our library crates is declared as all public items that are neither:
+- Inside a `__private` module
+- Documented as "unstable" or "experimental" in the first line of docs
+- Bear `unstable` or `experimental` in their absolute path
+
+## Node
+
+The versioning of the node is done most of the time by only incrementing the `minor` version.
```
kianenigma, line 30: I could not get until the end of this paragraph that this is referring to the polkadot node. 

```suggestion
The versioning of the Polkadot node is done most of the time by only incrementing the `minor` version. 
```

docs/RELEASE.md
```diff
@@ -0,0 +1,155 @@
...
+- Documented as "unstable" or "experimental" in the first line of docs
+- Bear `unstable` or `experimental` in their absolute path
+
+## Node
+
+The versioning of the node is done most of the time by only incrementing the `minor` version. 
+The `major` version is only bumped for special releases and the `patch` can be used for an 
+out of band release that fixes some critical bug. The node version is not following SemVer. 
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in
```
kianenigma, line 34:
```suggestion
interface or similar. The node version is declared in the [`NODE_VERSION`](https://paritytech.github.io/polkadot-sdk/master/polkadot_node_primitives/constant.NODE_VERSION.html) variable.

File: docs/RELEASE.md [hunk 33]
```diff
@@ -0,0 +1,155 @@
...
+- Bear `unstable` or `experimental` in their absolute path
+
+## Node
+
+The versioning of the node is done most of the time by only incrementing the `minor` version. 
+The `major` version is only bumped for special releases and the `patch` can be used for an 
+out of band release that fixes some critical bug. The node version is not following SemVer. 
+This means that the version doesn't express if there are any breaking changes in the CLI 
+interface or similar. The node version is declared in the `NODE_VERSION` variable in 
+`polkadot/node/primitives/src/lib.rs`.
```
kianenigma, line 35:
```suggestion
```

docs/RELEASE.md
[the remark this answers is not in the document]
kianenigma: I see Liam's point, that it would have been good to use SemVer to signal the behavior changes of the node. 

In my understanding the reason we don't do that is that we have always targeted Polkadot node 1.0 to be the white paper. Now that it has been delivered, we have two options: 

1. Indefinitely keep the major to 1 to adhere to the previous standard. 
2. Pivot into SemVer. All major versions higher than 1 adhere to Polkadot 1. But, then, if someday there is an official Polkadot 2 white paper and node for it, it would be a bit of a pickle. 

At the moment, I am leaning towards option 1. I think most operators are already used to reading changelogs and applying any CLI changes.

docs/RELEASE.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Westend & Rococo
+
+For the these networks, in addition to incrementing the `Cargo.toml` version we also increment the
+`spec_version` and sometimes the `transaction_version`. The spec version is also following the node
```
kianenigma, line 40: Is `transaction_version` really being used/enforced?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4912613

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4912594

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4914014

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4913997

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4914016

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4914015

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+4. Tag this commit as `nightlyYYMMDD`.
+5. Do a dry-run release to ensure that it *should* work.
+6. Push this tag (the commit will not belong to any branch).
+8. Release all crates that had changed since the last nightly release to crates.io.
+
+## Clobbering
+
+Cadence: every 6th release (~3 months). Responsible: Release Team
+
+This process aims to bring branch `stable` in sync with the latest audited commit of `master`. It is not done via a Merge Request but rather by just copying files. It should be automated.
```
eskimor, line 111: What is the problem with merging?

eskimor (commented):
Not sure this process is suited for relay chain development.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
+# Release
+
+The outputs of a release are the `polkadot` and `polkadot-parachain` node binaries, the runtimes for Westend & Rococo and their system parachains, and new crate versions published to `crates.io`.
+
+# Setup
+
+We have two branches: `master` and `stable`. `master` is the main development branch where normal merge requests are opened. Developers need to mostly only care about this branch.
```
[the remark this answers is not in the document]
liamaharon: Non-breaking changes still can be added to stable on a more frequent cadence.

Fellowship could also if they wish use the latest audited nightly version if they want to incorporate breaking changes.

docs/RELEASE.md
liamaharon: Copying files is not quite the best term.. It just means nuke the working tree with whatever is the most recently audited commit in nightly, instead of trying to do merge conflict resolution or whatever.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
+# Release
+
+The outputs of a release are the `polkadot` and `polkadot-parachain` node binaries, the runtimes for Westend & Rococo and their system parachains, and new crate versions published to `crates.io`.
+
+# Setup
+
+We have two branches: `master` and `stable`. `master` is the main development branch where normal merge requests are opened. Developers need to mostly only care about this branch.  
+The `stable` branch contains a version of the code that is ready to be released. Its contents are always audited. Merging to it is restricted to [Backports](#backports).
```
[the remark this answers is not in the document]
liamaharon: https://github.com/paritytech/polkadot-sdk/pull/2490#discussion_r1454579198 does this answer your q?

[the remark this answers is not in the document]
eskimor: Not really what would the workflow for relay chain devs look like? Like which branch would we usually base our PRs on? If I read this document right, it would be master and then we would backport to stable? And this for all Polkadot PRs?

I feel like the monorepo is biting us here a bit. From the perspective of Polkadot a more consuming perspective would make sense in my opinion. Basically we just use substrate as a normal dep, with its releases.  Develop Polkadot and whenever we actually need a new substrate feature (or need security fixes) we update our dependency.

docs/RELEASE.md
eskimor: Which would prohibit us from working on stable directly for Polkadot - correct?

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: >Fellowship could also if they wish use the latest audited nightly version if they want to incorporate breaking changes before the next stable release.

Yea true, did not think about this but the new process is meant to not slow us down. I guess audited nightly could work.

docs/RELEASE.md
ggwpez: Stable cannot have changes that were not in master first. Problem are the git conflicts and nuking is easier.

>Which would prohibit us from working on stable directly for Polkadot - correct?

Yes

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: >I feel like the monorepo is biting us here a bit. From the perspective of Polkadot a more consuming perspective would make sense in my opinion

Could be... but i think if we go with audited nightly then it should be very close to what we currently have.  
But its not really that nice, since currently we update the Runtimes to a proper Polkadot-SDK release, in the future we would update then to a nightly release.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+
+Describes how developers should merge bug and security fixes.
+
+### Steps
+
+1. Developer opens a Merge Request with a bug or security fix.
+2. They have the possibility to mark the MR as such, and does so.
+3. Audit happens with priority.
+4. It is merged into `master`.
+5. It is automatically back-ported to `stable`.
```
alindima, line 148: you mean we'll have automation for this? Most likely the backport will have conflicts or even logic changes that don't always result in git conflicts

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+## Crate Bumping
+
+Cadence: (possibly) each Merge Request. Responsible: Developer that opened the MR.
+
+Following SemVer isn't easy, but there exists [a guide](https://doc.rust-lang.org/cargo/reference/semver.html) in the Rust documentation that explains the small details on when to bump what. This process should be augmented with CI checks that utilize [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) and/or [`cargo-public-api`](https://github.com/Enselic/cargo-public-api). They must also pay attention to downstream dependencies that require a version bump, because they export the changed API.
+
+### Steps
+
+1. Developer opens a Merge Request with changed crates against `master`.
+2. They bump all changed crates according to SemVer.
```
alindima, line 68: what about having multiple PRs that modify the same crates. I guess we're supposed to bump them only once within a 3-month period right? This can be a bit cumbersome and prone to human error

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
+# Audit
+
+Audits are conducted to ensure the absence of severe or exploitable bugs. Merge Requests are generally merged into the `master` branch without audit. The `audited` tag is used to track the latest audited commit of the `master` branch. This means that audits need to happen in order of being merged.
```
alindima, line 3: if audits need to happen in the order of being merged to master, does the priority even matter? whichever one is merged first has the highest priority and blocks all new stable releases if I understand correctly.

docs/RELEASE.md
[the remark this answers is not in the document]
gavofyork: I don't see how this could push any feature further back than 3 months and only then in the very worst case. Given our RC development schedules tend to be of the order of years it doesn't seem all that relevant.

docs/RELEASE.md
gavofyork: In the trivial case I guess the back port could be made automatically. But in general it would likely be two separate (similar) PRs.

gavofyork (approved):

docs/AUDIT.md
ggwpez: Yes. Prioritizing would become "merging sooner".

docs/RELEASE.md
ggwpez: Yes good observation. There was some discussion on if we maybe rather use `PrDoc` to indicate SemVer bumps, but then we need additional tooling to translate that into crate bumps and it does not play well with other tools.  

We would probably need a CI that prevents double bumps, but i agree that it could be a point of friction.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+Cadence: every 6th release (~3 months). Responsible: Release Team
+
+This process aims to bring branch `stable` in sync with the latest audited commit of `master`. It is not done via a Merge Request but rather by just copying files. It should be automated.  
+The following script is provided to do the clobbering. Note that it keeps the complete history of all past clobbering processes.
+
+```bash
+# Ensure we have the latest remote data
+git fetch
+# Switch to the release branch
+git checkout release
```
muharem, line 118: meant `stable` probably here

muharem (approved):

liamaharon (approved):
Some comments but LGTM to merge and we can iron out wrinkles as we go

README.md
joepetrowski, line 43:
```suggestion
The Polkadot-SDK has two release channels: `mainline` and `nightly`. Production software is advised to only use `mainline`. `nightly` is meant for tinkerers to try out the latest features. The detailed release process is described in [RELEASE.md](docs/RELEASE.md).
```

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
+# Audit
+
+Audits are conducted to ensure the absence of severe or exploitable bugs. Merge Requests are generally merged into the `master` branch without audit. The `audited` tag is used to track the latest audited commit of the `master` branch. This means that audits need to happen in order of being merged.  
+This is an optimistic approach that lets us develop with greater speed, while requiring (possibly) large refactors in the failure case.
+
+Audits can be deferred if the logic is gated by an `experimental` feature or marked as "Not Production Ready" within the first line of doc. Such changes should be queued manually before these warnings are removed.
+
+## General Guidelines for what to Audit
+
+There is no single one-fits-all rule. Generally we should audit important logic that could immediately be used on production networks. If in doubt; ask in chat or in the Merge Request.
```
joepetrowski, line 10:
```suggestion
There is no single one-fits-all rule. Generally we should audit important logic that could immediately be used on production networks. If in doubt, ask in chat or in the Merge Request.
```

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+# Versioning
+
+We are releasing multiple different things from this repository in one release, but we don't want to
+use the same version for everything. Thus, in the following we explain the versioning story for the
+crates, node and Westend & Rococo. To easily refer to a release, it shall be named by its date in
+the form `stableYYMMDD`.
+
+## Crate
+
+We try to follow SemVer<sup>1</sup> as best as possible for versioning our crates. SemVer requires a piece of software to first declare a public API. The public API of the Polkadot SDK is hereby declared as the sum of all crates' public APIs, with the exception of the following internal crates:
```
joepetrowski, line 19: There is a mix of wrapped and unwrapped lines in this doc. Personally I prefer wrapped, but can the doc have a consistent style?

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+
+- @bkchr the list of node internals please
```
joepetrowski, line 21: Still todo before merge?

docs/RELEASE.md
joepetrowski: Yes

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+interface or similar. The node version is declared in the [`NODE_VERSION`](https://paritytech.github.io/polkadot-sdk/master/polkadot_node_primitives/constant.NODE_VERSION.html) variable.
+
+## Westend & Rococo
+
+For the these networks, in addition to incrementing the `Cargo.toml` version we also increment the
+`spec_version` and sometimes the `transaction_version`. The spec version is also following the node
+version. Its schema is: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This
+versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to
+the release in this repository.  
+The Westend testnet will be updated to the new runtime version immediately after a *Stable* release happened.
```
joepetrowski, line 43: I'm not sure what happened to the _other new_ testnet, but I thought that one was supposed to follow more stable releases, and Westend we would update more frequently to test latest features?

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Requesting an Audit
+
+1. Add the PR to the project `Security Audit (PRs) - SRLabs`
+2. Set status to Backlog
+3. Assign priority, considering the universe of PRs currently in the backlog, or just leave it as TBD
+4. Add the component
```
0xJayPi, line 17:
```suggestion
4. Add the component
```

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+Cadence: (possibly) each Merge Request. Responsible: Developer that opened the MR.
+
+Following SemVer isn't easy, but there exists [a guide](https://doc.rust-lang.org/cargo/reference/semver.html) in the Rust documentation that explains the small details on when to bump what. This process should be augmented with CI checks that utilize [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) and/or [`cargo-public-api`](https://github.com/Enselic/cargo-public-api). They must also pay attention to downstream dependencies that require a version bump, because they export the changed API.
+
+### Steps
+
+1. Developer opens a Merge Request with changed crates against `master`.
+2. They bump all changed crates according to SemVer.
+3. They bump all crates that export any changed types in their *public API*.
+4. They also bump all crates that inherit logic changes from relying on one of the bumped crates.
```
bkchr, line 70: I mean this is some sort of non exhaustive list on what you need to bump. IMO, we should just have step 2. You could add some remark that people should not forget crates that import the bumped crates.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+
+This process aims to release the `stable` branch as a *Stable* release every two weeks.
+
+### Steps
+
+1. Check if process [Clobbering](#clobbering) needs to happen and do so, if that is the case.
+2. Check out the latest commit of `stable`.
+3. Update the `CHANGELOG.md` version, date and compile the content using the prdoc files.
+4. Open a Merge Request against `stable` for visibility.
+5. Check if there were any changes since the last release and abort, if not.
```
bkchr, line 84: Sounds like it should be step 2. :P

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+6. Run `cargo semver-checks` and `cargo public-api` again to ensure that there are no SemVer breaks.
```
bkchr, line 85:
```suggestion
```

This is being done continuously by the continuous integration aka CI :P

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+7. Internal QA from the release team can happen here.
+8. Do a dry-run release to ensure that it *should* work.
```
bkchr, line 87:
```suggestion
```

What should work? Building the binary? Sounds like bullshit to me, again CI is building the stuff all the time.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+10. Comment that a *Stable* release will happen from the merged commit hash.
```
bkchr, line 88: Comment where?

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+11. Release all changed crates to crates.io.
+12. Create a release on GitHub.
```
bkchr, line 90: These two steps should be one and this should be automatic, similar to how the fellowship stuff works.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+13. Notify Devops so that they can update Westend to the new runtime.
```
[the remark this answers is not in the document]
bkchr: Tagging is automatically done by doing a release, see my comments above.

bkchr, line 91:
```suggestion
```

As written somewhere previously, we want that Westend always gets the latest nightly every week or every second week. It is a test net.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+
+This process aims to release the `master` branch as a *Nightly* release. The process can start at 00:00 UTC+1 and should automatically do the following steps.
+
+1. Check out the latest commit of branch `master`.
+2. Compare this commit to the latest `nightly*` tag and abort if there are no changes detected.
+3. Set the version of all crates to `major.0.0-nightlyYYMMDD` where `major` is the last released `major` version of that crate plus one.
+4. Tag this commit as `nightlyYYMMDD`.
+5. Do a dry-run release to ensure that it *should* work.
+6. Push this tag (the commit will not belong to any branch).
+8. Release all crates that had changed since the last nightly release to crates.io.
```
bkchr, line 105: Basically we would need to release all of the crates or not? Or the tool that checks if something changed in a crate, needs to take into consideration the dependency graph. Aka we changed something deep inside the graph and thus all crates that depend on this crate are also bumped.

Which brings up another point, we can not just do what you said in point 3. We only need to bump the version of the crates that changed. We will then also need to depend on some of the older nightly releases in crates. In the end we will end up with every crate being on a different nightly tag, depending on when it was modified the last time. I mean this works, but just highlights again that we need some tooling to ensure that we use the correct crate versions for every crate in some downstream project.

docs/RELEASE.md
bkchr, line 111:
```suggestion
This process aims to bring branch `stable` in sync with the latest audited commit of `master`. It is not done via a Pull Request but rather by just copying files. It should be automated.  
```

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+## Bug and Security Fix
+
+Cadence: n.a. Responsible: Developer
+
+Describes how developers should merge bug and security fixes.
+
+### Steps
+
+1. Developer opens a Merge Request with a bug or security fix.
+2. They have the possibility to mark the MR as such, and does so.
```
bkchr, line 145:
```suggestion
1. Developer opens a Pull Request with a bug or security fix.
2. They have the possibility to mark the PR as such, and does so.
```

bkchr (approved):
Left some comments, in general it looks good :+1:

README.md
```diff
@@ -35,6 +35,21 @@ Polkadot.
...
+
+### Mainline
+
+`mainline` releases have a support duration of **three months**. In this period, the release will not have any breaking changes but only receive bug and security fixes on a **two week** cadence.
+
+### Nightly
+
+`nightly` releases are released every night from the `master` branch, potentially with breaking changes. They have pre-release version numbers in the format `major.0.0-nightlyYYMMDD`.
```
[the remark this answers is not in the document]
ggwpez: Yes i think having nightly releases right from the start is unrealistic.  
It is still put in here as "how it *should* be", but i dont assume to see nightly releases soon.

docs/AUDIT.md
```diff
@@ -0,0 +1,17 @@
...
+
+## General Guidelines for what to Audit
+
+There is no single one-fits-all rule. Generally we should audit important logic that could immediately be used on production networks. If in doubt; ask in chat or in the Merge Request.
+
+## Requesting an Audit
+
+1. Add the PR to the project `Security Audit (PRs) - SRLabs`
+2. Set status to Backlog
+3. Assign priority, considering the universe of PRs currently in the backlog, or just leave it as TBD
```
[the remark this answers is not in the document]
ggwpez: I just copied what the SecOps told me. I think its not a good process, but they should take ownership of this doc eventually.

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: Yes **definitely** need to keep them consistent. I will create an issue.

1. Not sure. I still dont think that we should bump major every three months even without breaking changes.
2. ✅ yea we can use that every night to temporarily create the nightly versions.

ggwpez: Yes will do. I think we have a doc, but for now it way copy&paste from different docs.

docs/RELEASE.md
ggwpez: In the MR. But maybe that is obvious 🤷‍♂️

docs/RELEASE.md
ggwpez: Yes but there are lots of build errors that only arrive when building with `cargo publish --dry-run`.  
For example the location of the README or other relative paths gets messed up within the workspace, since it builds each crate in isolation.  
Eventually i dont mind *how* the release team ensures that it all works, but i think for a start it is fine to have in here.

docs/RELEASE.md
ggwpez: Oh okay, yea that is better.

docs/RELEASE.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Processes
+
+The following processes are necessary to actualize our releases. Each process has a *Cadence* on which it must execute and a *Responsible* that is responsible for autonomously doing so and reporting back any error in the RelEng<sup>2</sup> channel. All processes should be automated as much as possible.
+
+## Crate Bumping
+
+Cadence: (possibly) each Merge Request. Responsible: Developer that opened the MR.
+
+Following SemVer isn't easy, but there exists [a guide](https://doc.rust-lang.org/cargo/reference/semver.html) in the Rust documentation that explains the small details on when to bump what. This process should be augmented with CI checks that utilize [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) and/or [`cargo-public-api`](https://github.com/Enselic/cargo-public-api). They must also pay attention to downstream dependencies that require a version bump, because they export the changed API.
```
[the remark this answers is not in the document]
ggwpez: Probably we should add another chapter for this. Would do it afterwards though.  
The CI checks only see changes in the API definitions, not in their changed implementation or changed intent.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5006436

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5006437

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5006435

docs/RELEASE.md
```diff
@@ -0,0 +1,146 @@
...
+version. Its schema is: `M_mmm_ppp` and for example `1_002_000` is the node release `1.2.0`. This
+versioning has no further meaning, and is only done to map from an on chain `spec_version` easily to
+the release in this repository.  
+The Westend testnet will be updated to the new runtime version immediately after a *Stable* release happened.
+
+# Backports
+
+**From `master` to `stable`**
+
+Backports in this direction can be anything that is audited and either `minor` or a `patch` bump. [Security fixes](#bug-and-security-fix) should be prioritized over additions or improvements. Crates that are declared as internal API can also have major version bumps through backports.
```
ggwpez, line 48: @bkchr how do we declare crates as internal?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5006581

docs/RELEASE.md
ggwpez: Yea this is just the list of things that should happen, not how they should happen.

docs/RELEASE.md
ggwpez: >Basically we would need to release all of the crates or not? Or the tool that checks if something changed in a crate, needs to take into consideration the dependency graph. 

AFAIK parity-publish does exactly this.

> We will then also need to depend on some of the older nightly releases in crates. In the end we will end up with every crate being on a different nightly tag, depending on when it was modified the last time.

Ah crap. Did not think about this.

>I mean this works, but just highlights again that we need some tooling to ensure that we use the correct crate versions for every crate in some downstream project.

Yea this sounds like a nightmare. We could also create an overarching `polkadot-sdk` crate that locks in all crate versions. Otherwise some custom bumper CLI...

docs/RELEASE.md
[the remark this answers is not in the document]
ggwpez: created https://github.com/paritytech/polkadot-sdk/issues/3031

README.md
```diff
@@ -35,6 +35,26 @@ Polkadot.
...
+> Our release process is still Work-In-Progress and may not yet reflect the aspired outline here.
+
+The Polkadot-SDK has two release channels: `stable` and `nightly`. Production software is advised to only use `stable`.
+`nightly` is meant for tinkerers to try out the latest features. The detailed release process is described in
+[RELEASE.md](docs/RELEASE.md).
+
+### Stable
+
+`stable` releases have a support duration of **three months**. In this period, the release will not have any breaking
+changes. It will receive receive bug, security fixes, performance fixes and new non-breaking features on a **two week**
```
jsidorenko, line 50:
```suggestion
changes. It will receive bug fixes, security fixes, performance fixes and new non-breaking features on a **two week**
```

docs/AUDIT.md
```diff
@@ -0,0 +1,22 @@
+# Audit
+
+Audits are conducted to ensure the absence of severe or exploitable bugs. Merge Requests are generally merged into the
```
jsidorenko, line 3: Shall we use `Pull Requests` instead of `Merge Requests` everywhere? As I can see, in some places we have 'Merge Requests', while in other 'Pull Requests'. For devs not experienced with Gitlab, the term 'Merge Request' could cause confusion.

ggwpez: Oh, i tried to make it uniform but got it messed up again. Thanks, will fix.

PR #2493, opened 2023-11-24
Zombienet: add polkadot-debug image publish as needs for cumulus tests

pepoviola:
Add `build-push-image-polkadot-debug` job to needs since we changed to use the polkadot-debug image from the current branch for cumulus test and we need to be sure that the image is ready. Fix issues like https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4481059

cc: @bkchr


michalkucharczyk (approved):
Good catch, I missed that...

bkchr (approved):

PR #2497, opened 2023-11-26
Refactors staking and `stake-tracker` so that target and voter lists include idle stakers

gpestana:
In this PR we change the current assumptions so that `TargetList` and `VoterList` **also** keep track of the idle stakers (w.g. chilled) and the target scores are tracked and updated even while the validator is chilled. The idle stakers will be removed when the ledger unbonds.

This allows the stake-tracker to keep track of the chilled validators and respective score so that a re-validate of a chilled validator is brought up with the correct target score (i.e. self stake + all current nominations in the system).

### DIFF
--- a/substrate/frame/staking/src/mock.rs
+++ b/substrate/frame/staking/src/mock.rs
@@ -602,7 +602,7 @@ impl ExtBuilder {
 		let mut ext = self.build();
 		ext.execute_with(test);
 		ext.execute_with(|| {
-			Staking::do_try_state(System::block_number())
+			let _ = Staking::do_try_state(System::block_number())
 				.map_err(|err| println!(" 🕵️‍♂️  try_state failure: {:?}", err))
 				.unwrap();
 		});
@@ -881,7 +881,10 @@ pub(crate) fn balances(who: &AccountId) -> (Balance, Balance) {
 pub(crate) fn stake_tracker_sanity_tests() -> Result<(), &'static str> {
 	use sp_staking::StakingInterface;
 
-	assert_eq!(Nominators::<Test>::count() + Validators::<Test>::count(), VoterBagsList::count());
+	assert_eq!(
+		Nominators::<Test>::count() + Validators::<Test>::count(),
+		VoterBagsList::iter().filter(|v| Staking::status(&v) != Ok(StakerStatus::Idle)).count() as u32,
+	);
 
 	// recalculate the target's stake based on voter's nominations and compare with the score in the
 	// target list.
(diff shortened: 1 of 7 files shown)


gpestana:
bot fmt

substrate/frame/staking/src/mock.rs
```diff
@@ -881,7 +881,12 @@ pub(crate) fn balances(who: &AccountId) -> (Balance, Balance) {
 pub(crate) fn stake_tracker_sanity_tests() -> Result<(), &'static str> {
 	use sp_staking::StakingInterface;
 
-	assert_eq!(Nominators::<Test>::count() + Validators::<Test>::count(), VoterBagsList::count());
+	assert_eq!(
+		Nominators::<Test>::count() + Validators::<Test>::count(),
+		VoterBagsList::iter()
+			.filter(|v| Staking::status(&v) != Ok(StakerStatus::Idle))
```
Ank4n, line 887: may be better to reduce it to a tuple of (nominator_count, validator_count) and then assert on each count separately.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -832,7 +833,9 @@ impl<T: Config> Pallet<T> {
 		let mut nominators_taken = 0u32;
 		let mut min_active_stake = u64::MAX;
 
-		let mut sorted_voters = T::VoterList::iter();
+		// voter list also contains chilled/idle voters, filter those.
+		let mut sorted_voters =
+			T::VoterList::iter().filter(|v| Self::status(&v) != Ok(StakerStatus::Idle));
```
Ank4n, line 838: Might be more explicit to filter like this?
```suggestion
			T::VoterList::iter().filter(|v| Self::status(&v) == Ok(StakerStatus::Nominator));
```

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -940,7 +943,9 @@ impl<T: Config> Pallet<T> {
 		let mut all_targets = Vec::<T::AccountId>::with_capacity(final_predicted_len as usize);
 		let mut targets_seen = 0;
 
-		let mut targets_iter = T::TargetList::iter();
+		// target list also contains chilled/idle validators, filter those.
+		let mut targets_iter =
+			T::TargetList::iter().filter(|t| Self::status(&t) != Ok(StakerStatus::Idle));
```
Ank4n, line 948: Similarly
```suggestion
			T::TargetList::iter().filter(|t| Self::status(&t) != Ok(StakerStatus::Validator));
```

substrate/frame/staking/src/mock.rs
Ank4n: To expand, you can have assertions for count of nominators and validators separately.

Ank4n: Never mind what I said, its wrong.

PR #2510, opened 2023-11-27
polkadot: disable block authoring backoff on production networks

andresilva:
Currently the polkadot node will backoff from block authoring if finality starts lagging. This PR disables this mechanism on production networks (polkadot and kusama) and adds a flags to optionally force enabling it.

### DIFF
--- a/cumulus/client/relay-chain-inprocess-interface/src/lib.rs
+++ b/cumulus/client/relay-chain-inprocess-interface/src/lib.rs
@@ -286,6 +286,7 @@ fn build_polkadot_full_node(
 			grandpa_pause: None,
 			// Disable BEEFY. It should not be required by the internal relay chain node.
 			enable_beefy: false,
+			force_authoring_backoff: false,
 			jaeger_agent: None,
 			telemetry_worker_handle,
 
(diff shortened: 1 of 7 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4505302

bkchr (approved):

acatangiu (approved):

tomaka:
What's the motivation for this change?

This is pretty problematic for light clients.
If finality starts lagging and blocks continue being produced at a normal rate, the number of blocks that must be tracked increases linearly, and for light clients there will be a point when there isn't enough memory anymore to hold all the not-finalized-yet blocks.

Similarly, the new JSON-RPC API has been designed with the assumption that it's possible to hold all non-finalized blocks in memory.

burdges:
Adversaries wind up controlling the chain when honest nodes backoff, which adversaries can trigger.  Alistair and I raised this concern earlier, but we finally discussed the alternatives in https://github.com/paritytech-secops/srlabs_findings/issues/227

bkchr:
For a proper solution we will require to add a finality tracker on chain. This should be hopefully be achievable with beefy as it is lightweight enough and then the runtime can take care to scale down the number of produced blocks on finality lags.

tomaka:
I've discussed this in DM with André for what it's worth.

eskimor:
> This is pretty problematic for light clients.

Also for validators, for the same reasons.

PR #2513, opened 2023-11-27
Add UniqueMembership trait implemented for a non-fungible

olanod:
A general purpose trait to manage "memberships". Memberships are unique and they are implemented for `nonfungible_v2` unlocking the use case of transferability.


substrate/frame/support/src/traits/members.rs
```diff
@@ -391,3 +394,141 @@ impl<T: Clone + Ord> ChangeMembers<T> for () {
...
+		*self = Self(self.0.saturating_sub(n.get()).max(Self::MIN.0))
+	}
+}
+impl From<u8> for GenericRank {
+	fn from(value: u8) -> Self {
+		Self(value)
+	}
+}
+
+const MEMBERSHIP_INFO_ATTRIBUTE: [u8; 10] = *b"memberinfo";
```
jsidorenko, line 489: What do you think about adding `MEMBERSHIP_INFO_ATTRIBUTE` to  the [`PalletAttributes` enum](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/nfts/src/types.rs#L355)?
Maybe we should extract that enum to a common space and rename to `NftSystemAttributes`..

olanod: Extracting it somewhere in frame-support would be better :) we shouldn't create a dependency with the nfts pallet which is also why I made the ItemConfig an associated type.

PR #1654, opened 2023-09-20
Adds `stake-tracker` pallet

gpestana:
This PR adds the `stake-tracker` pallet without integrating it with the staking and bags lists pallets.

The stake-tracker pallet implements the `OnStakingUpdate` trait to listen to staking events and multiplexes those events to one or multiple types (e.g. pallets). Initially, the stake tracker pallet is used as a degree of indirection to maintain the target and voter semi-sorted lists up to date.

The final goal is to achieve a design where the targets and voters lists are updated at each relevant staking event through the stake-tracker pallet. The list reads in staking are performed directly through the `SortedListProvider`.

stake-tracker

**To finalize**
- [ ] Consider a multi-block event processing to handle updates on slashes and rewards for both targets and voters (unconfirmed, perhaps on a follow-up PR).

Related and step towards https://github.com/paritytech/polkadot-sdk/issues/443

PR #2514, opened 2023-11-28
Initialize `TotalValueUnbonding`

### LINKED ISSUE #441: pallet-staking: track total unstaking amount
In staking, we have a mapping as such: `AccountId -> Ledger` whereby `Ledger { active, total }`. What we need is a tracker over the `sum(Ledger::active)` and `sum(Ledger::total)` in the entire map. Subtracting the two should give the total unbonding amount. 

As for the implementation, once #1654 is done, staking will have fully implement the `OnStakingUpdate`. This can be implemented as a component that is implementing `OnStakingUpdate` and using that. 

I prefer an implementation that is actually just a `stuct` implementing the types, which can easily be embedded in a pallet. Then, we can decide to position this in the existing staking pallet or elsewhere. 

I imagine we can use it to: 

1. future fancier unstaking process.
2. implement things like preventing all of the DOTs to be unstaked at once etc.

eagr:
closes #441 

Adding a `TotalValueUnbonding` to on-chain storage for tracking the sum of unbonding balances across pools, which
* increases on unbonding/unstaking
* decreases on withdrawing unbonded funds (not when funds are unbonded)

These are implemented on `BondedPool` coz that's where the TVL ops go. I gather it'd be better to sorta put the similar logic in one place.

Verifications are done with
```rs
ensure!(LVU == sum(total_stake - active_stake))
ensure!(LVU <= sum(member.unbonding_balance()))
```

LMK if I'm mistaken about anything. :))

TODOs
* [ ] update docs (after confirming the implementation)


substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -47,6 +57,70 @@ pub mod versioned {
...
+
+			let members_balance_unbonding = PoolMembers::<T>::iter()
+				.map(|(_, member)| member.unbonding_balance())
+				.reduce(|acc, balance| acc + balance)
+				.unwrap_or_default();
+
+			ensure!(
+				TotalValueUnbonding::<T>::get() <= members_balance_unbonding,
+				"TotalValueUnbonding cannot surpass the sum of unbonding funds of members across all pools.",
+			);
```
liamaharon, line 117: Why's this `<=` not `==`?

substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -47,6 +57,70 @@ pub mod versioned {
...
+
+	impl<T: Config> MigrateToV8<T> {
+		fn calculate_tvu() -> BalanceOf<T> {
+			BondedPools::<T>::iter()
+				.map(|(id, inner)| {
+					let bonded_account = BondedPool { id, inner: inner.clone() }.bonded_account();
+					let total_stake = T::Staking::total_stake(&bonded_account).unwrap_or_default();
+					let active_stake =
+						T::Staking::active_stake(&bonded_account).unwrap_or_default();
+					total_stake - active_stake
```
liamaharon, line 73: Should use safe math here I think

liamaharon:
cc @gpestana @Ank4n

substrate/frame/nomination-pools/src/migration.rs
eagr: `active_stake <= total_stake && active_stake >= 0` should always hold right? If that's the contract, I would rather let the migration fail than carry on with the wrong states in this case.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4510096

substrate/frame/nomination-pools/src/migration.rs
eagr: TBH, I'm not sure. :)) I'm just following the pattern above: `TotalValueLocked <= members_balance_total`

substrate/frame/nomination-pools/src/migration.rs
liamaharon: It wouldn't fail, it would silently underflow :P

substrate/frame/nomination-pools/src/tests.rs
```diff
@@ -4333,6 +4378,8 @@ mod withdraw_unbonded {
 					pool_events_since_last_call(),
 					vec![Event::Withdrawn { member: 30, pool_id: 1, points: 5, balance: 5 }]
 				);
+				// FIXME eagr actual value 5?
```
eagr, line 4381: Does this mean a bug in my code or I misunderstand the test steps?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4526803

eagr:
Some things I'm uncertain about.

Are the added assertions in the tests excessive or even necessary? Providing the added `ensure!` assertions seem to run on every test case. If those aren't necessary, I'm happy to remove them.

How do we determine whether the migration could be carried out in one block?

gpestana (changes requested):
High level, this implementation for exposing the total value unbonding in the context of nomination pools seems fine. However, I think it would be better to implement this at a higher lever in the stack. Nomination pools are just a keyless staker account, module a few cosmetic differences. This said, I think it would be great if we could implement this so that both nomination pools, validators and nominations would expose this value.

To achieve this, we probably need to implement this logic at the staking pallet level and include this functionality as part of the `StakingInterface` trait. wdyt?

eagr:
Thanks for the advices! A couple questions.

* Are `TVL` and `TVU` (if one should go upstream so should the other) useful enough to justify upstreaming them? I mean, there's also a risk of bloating the base trait with features that people may not need.
* How about using a new trait for those features? Off the top of my head, something like `trait SwarmStakingInterface`, or maybe `trait SwarmStakingInterface: StakingInterface`.

I don't see the whole picture, so I'd trust your judgements.

kianenigma:
Just reading the description, this is a misunderstanding from the original issue. What I ment to request is total unbonding at the staking level, not nomination pools.

In other words, in staking, we have a mapping as such: `AccountId -> Ledger` whereby `Ledger { active, total }`. What we need is a tracker over the `sum(Ledger::active)` and `sum(Ledger::total)`. Subtracting the two should give the total unbonding amount. 

As for the implementation, once #1654 is done, staking will have fully implement the `OnStakingUpdate`. This can be implemented as a component that is implementing `OnStakingUpdate` and using that.

kianenigma:
I think the work here is valuable and should be replicated at the right abstraction level, but since this PR is certainly not going to be merged as-is, I will close this.

PR #2515, opened 2023-11-28
Set `frame_system::LastRuntimeUpgrade` after running `try-runtime` migrations

liamaharon:
Sets `frame_system::LastRuntimeUpgrade` after running try-runtime migrations to better emulate real behavior. 

This fixes an issue where migrations using the spec version to determine whether to execute can incorrectly fail idempotency checks. 

@s0me0ne-unkn0wn noticed this issue with the session key migration introduced in https://github.com/paritytech/polkadot-sdk/pull/2265.

### DIFF
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -362,9 +362,13 @@ where
 		Ok(frame_system::Pallet::<System>::block_weight().total())
 	}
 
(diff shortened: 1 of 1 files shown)


s0me0ne-unkn0wn (approved):
LVGTM, thank you!

ggwpez (approved):

PR #2524, opened 2023-11-28
Update schnorrkel to 0.11.4

andresilva:


polkadot/node/core/approval-voting/Cargo.toml
```diff
@@ -31,14 +31,14 @@ sp-consensus = { path = "../../../../substrate/primitives/consensus/common", def
...
-rand_core = "0.5.1"
+rand_core = "0.6.2"
 rand_chacha = { version = "0.3.1" }
 rand = "0.8.5"
 
 [dev-dependencies]
 async-trait = "0.1.57"
 parking_lot = "0.12.0"
-rand_core = "0.5.1" #                                                                       should match schnorrkel
+rand_core = "0.6.2" # should match schnorrkel
```
davxy, line 41:
```suggestion
```
Already in `dependencies`

davxy (approved):

andresilva:
We should do a burn-in on this, in particular try to sync both Kusama and Polkadot from scratch to make sure there's no issues.

altonen:
Libp2p devs were nice and accepted to backport a fix to 0.51 which works around the WebRTC issue. We didn't even enable the feature in Polkadot but the dependencies got pulled in anyway.

This is now fixed in 0.51.4 so if you want, this PR could be rebased on top master with just a bump in the libp2p patch version

cc @koute @michalkucharczyk @skunert

andresilva:
@altonen Unfortunately it still doesn't solve our problem because of `snow` (coming from `libp2p-noise`), which requires `subtle` 2.4.1 (https://github.com/mcginty/snow/blob/main/Cargo.toml#L41).

I guess an alternative with potentially faster turnaround would be to release a new schnorrkel version that only requires subtle `2.4`? @burdges does that make sense or do we really need `2.5`?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4660834

burdges:
Published schnorrkel 0.11.4 with subtle 2.4.1

andresilva:
I synced polkadot with this PR and didn't find any issue. Will sync kusama as well and if no issue is found I think we can merge this.

alexggh:
Some additional positive datapoints I've got for this PR, I've been working on some benchmarks(https://github.com/paritytech/polkadot-sdk/pull/2621) for approval-distribution and approval-voting subsystems and I gave this PR a spin a few scenarios that I can confirm work:
1. `0.11.4` seems to be backwards compatible with assignments and approvals generated with `0.9.1`, tested around 400K messages generated with `0.9.1`.
2. At least on my machine there seems to be a speed-up on average of about **14-15%** in the time it takes to process `2574667` messages.
## Times for 0.9.1
```
    Total approval-distribution CPU usage 21.57s
    CPU usage per block 2.16s
    Total approval-voting CPU usage 34.51s
    CPU usage per block 3.45s
```
## Times for 0.11.4
```
    Total approval-distribution CPU usage 21.67s
    CPU usage per block 2.17s
    Total approval-voting CPU usage 34.44s
    CPU usage per block 3.44s
```

andresilva:
Thanks for testing @alexggh. I believe that might be a result of this schnorrkel release using an updated version of `curve25519-dalek` which includes [this](https://github.com/dalek-cryptography/curve25519-dalek/pull/523).

burdges:
Yes, that's the verification speed up, but signing maybe slighly faster too, due to ThreadRng vs OSRng.  I'll try to speed up VRF verification a little bit more too in the future

andresilva:
Synced kusama as well, no issues. I think we can go ahead and merge this.

bkchr (approved):

PR #2531, opened 2023-11-28
[ci] Run gitspiegel trigger with merge conflicts

alvicsam:
Currently gitspiegel trigger won't run if there is merge conflict. This PR fixes it.

close https://github.com/paritytech/gitspiegel/issues/183

### DIFF
--- a/.github/workflows/gitspiegel-trigger.yml
+++ b/.github/workflows/gitspiegel-trigger.yml
@@ -6,7 +6,8 @@ name: gitspiegel sync
 # and running the workflow is protected by GitHub
 
 on:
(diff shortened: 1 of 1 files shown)


AndWeHaveAPlan (approved):

rcny (approved):

pavelsupr (approved):
Does it make any sense to preserve the previous `pull_request` trigger after this change?

alvicsam:
> Does it make any sense to preserve the previous `pull_request` trigger after this change?

Yes, because it doesn't work without it for some reasons.

Issue #2530, opened 2023-11-28
[RPC-Spec-V2] chainHead: Backport error codes from spec


lexnv:
Backport error codes from: https://github.com/paritytech/json-rpc-interface-spec/pull/117:

- `-32800` two or more subscriptions requested for chainHead_follow
- `-32801` block hash not reported by chainHead_follow or block hash has been unpinned
- `-32802` chainHead_follow started with withRuntime == false
- `-32803` chainHead_follow did not generate an operationWaitingForContinue event

The following are errors defined in the [JSON-RPC spec](https://www.jsonrpc.org/specification#error_object):
- `-32602` The provided parameter isn't one of the expected values, has different format or is missing
- `-32603` Internal server error:
  - The server was unable to generate a valid subscription ID (https://github.com/paritytech/polkadot-sdk/blob/21f1811c6600d8a7fe043592ff34dcb79284d583/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs#L198-L204)
  - The server cannot interpret the block: hardware error, ram/disk error, corrupt database, broken code logic (https://github.com/paritytech/polkadot-sdk/blob/21f1811c6600d8a7fe043592ff34dcb79284d583/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs#L306-L309)

PR #2539, opened 2023-11-29
chainHead: Backport error codes from spec

lexnv:
This PR backports the error codes from the spec.

This relies on two specs for defining the error codes:
- Our rpc-spec-v2 https://github.com/paritytech/json-rpc-interface-spec.
- JSON-RPC spec https://www.jsonrpc.org/specification#error_object.

To better describe the error codes, they are divided into two separate modules `rpc_spec_v2` and `json_rpc_spec` respectively.

The `InvalidSubscriptionID` and `FetchBlockHeader` are merged into the JSON-RPC spec `INTERNAL_ERROR`.
While the other error codes are adjusted from spec.

Errors that are currently in use:
- -32801 block hash not reported by chainHead_follow or block hash has been unpinned
- -32802 chainHead_follow started with withRuntime == false
- -32803 chainHead_follow did not generate an operationWaitingForContinue event

The following are errors defined in the [JSON-RPC spec](https://www.jsonrpc.org/specification#error_object):
- -32602 The provided parameter isn't one of the expected values, has different format or is missing
- -32603 Internal server error

Note: Error `-32801` must be introduced and generated by the outstanding https://github.com/paritytech/polkadot-sdk/issues/1505

Closes: https://github.com/paritytech/polkadot-sdk/issues/2530


skunert (approved):
Nice :ok_hand:

substrate/client/rpc-spec-v2/src/chain_head/error.rs
```diff
@@ -22,52 +22,62 @@ use jsonrpsee::{
...
-			Error::InvalidContinue => ErrorObject::owned(INVALID_CONTINUE, msg, None::<()>),
+			Error::InvalidBlock =>
+				ErrorObject::owned(rpc_spec_v2::INVALID_BLOCK_ERROR, msg, None::<()>),
+			Error::InvalidRuntimeCall(_) =>
+				ErrorObject::owned(rpc_spec_v2::INVALID_RUNTIME_CALL, msg, None::<()>),
+			Error::InvalidContinue =>
+				ErrorObject::owned(rpc_spec_v2::INVALID_CONTINUE, msg, None::<()>),
+			Error::InvalidParam(_) =>
+				ErrorObject::owned(json_rpc_spec::INVALID_PARAM_ERROR, msg, None::<()>),
+			Error::InternalError(_) =>
```
niklasad1, line 79: Technically according https://www.jsonrpc.org/specification#error_object the message should be "Internal error" but I guess this is fine anyway.

niklasad1 (approved):

PR #2281, opened 2023-11-12
Rococo & Westend People Chain

joepetrowski:
Rococo and Westend runtimes for the "People Chain". This chain contains the Identity pallet with plans to migrate all related data from the Relay Chain.

Changes `IdentityInfo` to:

- Remove `additional_fields`.
- Add `github` and `discord` as first class fields. From scraping chain data, these were the only two additional fields used (for the Fellowship and Ambassador Program, respectively).
- Rename `riot` to `matrix`.

Note: This will use the script in https://github.com/paritytech/polkadot-sdk/pull/2025 to generate the genesis state.

TODO:

- [x] https://github.com/paritytech/polkadot-sdk/pull/1814 and integration of the Identity Migrator pallet for migration.
- [x] Tests: https://github.com/paritytech/polkadot-sdk/pull/2373

PR #2540, opened 2023-11-29
Add Rococo People <> Rococo Bulletin bridge support to Rococo Bridge Hub

svyatonik:
This PR adds [Rococo People](https://github.com/paritytech/polkadot-sdk/pull/2281) <> [Rococo Bulletin](https://github.com/zdave-parity/polkadot-bulletin-chain) to the Rococo Bridge Hub code. There's a couple of things left to do here:
- [x] add remaining tests - it'd need some refactoring in the `bridge-hub-test-utils` - will do in a separate PR;
- [x] actually run benchmarks for new messaging pallet (do we have bot nowadays?).

The reason why I'm opening it before this ^^^ is ready, is that I'd like to hear others opinion on how to deal with hacks with that bridge. Initially I was assuming that Rococo Bulletin will be the 1:1 copy of the Polkadot Bulletin (to avoid maintaining multiple runtimes/releases/...), so you can see many `PolkadotBulletin` mentions in this PR, even though we are going to bridge with the parallel chain (`RococoBulletin`). That's because e.g. pallet names from `construct_runtime` are affecting runtime storage keys and bridges are using runtime storage proofs => it is important to use names that the Bulletin chain expects.

But in the end, this hack won't work - we can't use Polkadot Bulletin runtime to bridge with Rococo Bridge Hub, because Polkadot Bulletin expects Polkadot Bridge hub to use `1002` parachain id and Rococo Bridge Hub seats on the `1013`. This also affects storage keys using in bridging, so I had to add the [`rococo` feature](https://github.com/svyatonik/polkadot-bulletin-chain/blob/add-bridge-pallets/runtime/Cargo.toml#L198) to the Bulletin chain. So now we can actually alter its runtime and adapt it for Rococo.

So the question here is - what's better for us here
- to leave everything as is (seems hacky and non-trivial);
- change Bulletin chain runtime when `rococo` feature is used - e.g. use proper names there (`WithPolkadotGrandpa` -> `WithRococoGrandpa`, ...)
- add another set of pallets to the Bulletin chain runtime to bridge with Rococo and never use them in production. Similar to hack that we had in Rococo/Wococo

cc @acatangiu @bkontur @serban300 

also cc @joepetrowski as the main "client" of this bridge

---

A couple words on how this bridge is different from the Rococo <> Westend bridge:
- it is a bridge with a chain that uses GRANDPA finality, not the parachain finality (hence the tests needs to be changed);
- it is a fee-free bridge. So `AllowExplicitUnpaidExecutionFrom<Equals<SiblingPeople>>` + we are not paying any rewards to relayers (apart from compensating transaction costs).


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4536033

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
```diff
@@ -0,0 +1,267 @@
...
+use xcm_builder::{BridgeBlobDispatcher, HaulBlobExporter};
+
+parameter_types! {
+	/// Maximal number of entries in the unrewarded relayers vector at the Rococo Bridge Hub. It matches the
+	/// maximal number of unrewarded relayers that the single confirmation transaction at Rococo Bulletin Chain
+	/// may process.
+	pub const MaxUnrewardedRelayerEntriesAtInboundLane: bp_messages::MessageNonce =
+		bp_polkadot_bulletin::MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX;
+	/// Maximal number of unconfirmed messages at the Rococo Bridge Hub. It matches the maximal number of
+	/// uncinfirmed messages that the single confirmation transaction at Rococo Bulletin Chain may process.
```
acatangiu, line 62:
```suggestion
	/// unconfirmed messages that the single confirmation transaction at Rococo Bulletin Chain may process.
```

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
```diff
@@ -0,0 +1,267 @@
...
+	/// message is following.
+	pub FromRococoPeopleToRococoBulletinRoute: SenderAndLane = SenderAndLane::new(
+		ParentThen(X1(Parachain(RococoPeopleParaId::get().into()))).into(),
+		XCM_LANE_FOR_ROCOCO_PEOPLE_TO_ROCOCO_BULLETIN,
+	);
+
+	/// XCM message that is never sent.
+	pub NeverSentMessage: Option<Xcm<()>> = None;
+}
+pub const XCM_LANE_FOR_ROCOCO_PEOPLE_TO_ROCOCO_BULLETIN: LaneId = LaneId([0, 0, 0, 0]);
```
acatangiu, line 98: nit: let's use some "non-default" value:
```suggestion
pub const XCM_LANE_FOR_ROCOCO_PEOPLE_TO_ROCOCO_BULLETIN: LaneId = LaneId([0, 0, 0, 1]);
```

(`XCM_LANE_FOR_ASSET_HUB_WESTEND_TO_ASSET_HUB_ROCOCO` is set to `LaneId([0, 0, 0, 2])`)

Also maybe we should use some name that implies it is bidirectional? Maybe `XCM_LANE_BETWEEN_ASSET_HUB_WESTEND_AND_ASSET_HUB_ROCOCO`

acatangiu (approved):
Haven't looked at the tests, but everything else looks good as is.

acatangiu:
> So the question here is - what's better for us here
    - to leave everything as is (seems hacky and non-trivial)
    -  change Bulletin chain runtime when rococo feature is used - e.g. use proper names there (WithPolkadotGrandpa -> WithRococoGrandpa, ...)
     - add another set of pallets to the Bulletin chain runtime to bridge with Rococo and never use them in production. Similar to hack that we had in Rococo/Wococo

If I understand correctly, the question only affects the Bulletin chain runtime (which is deployed on both Polkadot and Rococo), and how it defines/sees the bridges (to PolkadotBH or RococoBH depending on where itself is deployed), right?

IMO any of the 3 options is good. I expect that when Bulletin chain will be launched, it will also be moved to fellowship and we'll keep a Rococo copy here, and the problem "goes away". So in the meantime, I'm fine with whatever option is easiest for us :)

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
svyatonik: It has already propagated to 4+ runtimes and scripts, so it'll be painful to do that now, so I'd use it as is

svyatonik:
> > So the question here is - what's better for us here
> > - to leave everything as is (seems hacky and non-trivial)
> > -  change Bulletin chain runtime when rococo feature is used - e.g. use proper names there (WithPolkadotGrandpa -> WithRococoGrandpa, ...)
> > - add another set of pallets to the Bulletin chain runtime to bridge with Rococo and never use them in production. Similar to hack that we had in Rococo/Wococo
> 
> If I understand correctly, the question only affects the Bulletin chain runtime (which is deployed on both Polkadot and Rococo), and how it defines/sees the bridges (to PolkadotBH or RococoBH depending on where itself is deployed), right?
> 
> IMO any of the 3 options is good. I expect that when Bulletin chain will be launched, it will also be moved to fellowship and we'll keep a Rococo copy here, and the problem "goes away". So in the meantime, I'm fine with whatever option is easiest for us :)

Ok, thanks. I'm leaving as is, because it is already deployed somewhere

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4686733

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4686734

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -1004,6 +1063,41 @@ impl_runtime_apis! {
 				}
 			}
 
+			impl BridgeMessagesConfig<bridge_to_bulletin_config::WithRococoBulletinMessagesInstance> for Runtime {
+				fn is_relayer_rewarded(_relayer: &Self::AccountId) -> bool {
+					// we do not pay any rewards in this bridge
```
bkontur, line 1068: nit: `is_relayer_rewarded == true` => "we do not pay any rewards" is little bit kind of confusing, but probably we can live with that

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -68,6 +69,7 @@ parameter_types! {
 	pub const MaxAssetsIntoHolding: u32 = 64;
 	pub TreasuryAccount: AccountId = TREASURY_PALLET_ID.into_account_truncating();
 	pub RelayTreasuryLocation: MultiLocation = (Parent, PalletInstance(rococo_runtime_constants::TREASURY_PALLET_ID)).into();
+	pub SiblingPeople: MultiLocation = (Parent, Parachain(1004)).into();
```
bkontur, line 72: I think instead of `1004` we should use rococo's system parachain constant, I would add it like here:
https://github.com/paritytech/polkadot-sdk/pull/2281/files#diff-723e030bc6a22c68b3606b0463863382f2b3c82af20c7b36da8d889199e1b86eR115-R116

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/tests.rs
```diff
@@ -346,3 +348,202 @@ mod bridge_hub_rococo_tests {
...
+				match RuntimeEvent::decode(&mut &runtime_event_encoded[..]) {
+					Ok(RuntimeEvent::XcmpQueue(event)) => Some(event),
+					_ => None,
+				}
+			}),
+			XCM_LANE_FOR_ROCOCO_PEOPLE_TO_ROCOCO_BULLETIN,
+			|| (),
+		)
+	}
+	/*
```
bkontur, line 435: can we uncomment this now?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/// Dispatches received XCM messages from other bridge.
+type FromRococoBulletinMessageBlobDispatcher = BridgeBlobDispatcher<
+	XcmRouter,
+	BridgeHubRococoUniversalLocation,
+	BridgeRococoToRococoBulletinMessagesPalletInstance,
+>;
+
+/// Export XCM messages to be relayed to the other side.
+pub type ToRococoBulletinHaulBlobExporter = HaulBlobExporter<
```
bkontur, line 115: I think we could use `pallet_xcm_bridge_hub` here also like for to_westend stuff: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_westend_config.rs#L129C47-L129C70

bkontur (approved):
nice, lgtm, just few very nits which will be solved after rebase to actual master

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
```diff
@@ -0,0 +1,267 @@
...
+		= XCM_LANE_FOR_ROCOCO_PEOPLE_TO_ROCOCO_BULLETIN;
+	/// Priority boost that the registered relayer receives for every additional message in the message
+	/// delivery transaction.
+	///
+	/// It is determined semi-automatically - see `FEE_BOOST_PER_MESSAGE` constant to get the
+	/// meaning of this value.
+	pub PriorityBoostPerMessage: u64 = 182_044_444_444_444;
+
+	/// Identifier of the sibling Rococo People parachain.
+	pub RococoPeopleParaId: cumulus_primitives_core::ParaId = 1004.into();
```
bkontur, line 87: also better to use rococo's system parachain constants mentioned in previous comment

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
svyatonik: Well, we can't test that the relayer is rewarded if we don't pay any rewards :) I could change a comment, if you have some better version :)

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
svyatonik: Yep, let's wait until https://github.com/paritytech/polkadot-sdk/pull/2281 is merged - I'll do final polishing afterwards

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/tests.rs
svyatonik: I've been pretty sure that I did that already :confused:  Lost that commit somewhere, so did it again from the scratch :)

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
svyatonik: Yep, added the other instance of BH pallet

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -483,39 +488,56 @@ construct_runtime!(
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 36,
 
-		// BridgeHubRococo uses:
-		//  - BridgeWestendGrandpa
-		//  - BridgeWestendParachains
-		//  - BridgeWestendMessages
-		//  - BridgeRelayers
+		// Bridge relayers pallet, used by several bridges here.
```
svyatonik, line 491: I've changed the grouping here - this runtime is no longer shared by Rococo and Wococo, so imo it makes sense to group by bridged chain.

svyatonik:
@bot help

svyatonik:
bot cancel 5-a34ea1ac-b42a-424e-ac12-0d58da204a4d

svyatonik:
bot bench cumulus-bridge-hubs --runtime=bridge-hub-rococo --pallet=pallet_bridge_messages

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
svyatonik: Actually, let's add it here - I've forgot that this PR is based on master, so IIUC there's nothing wrong with adding it here

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_to_bulletin_config.rs
svyatonik: done

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4726740

PR #2542, opened 2023-11-29
remove timepoint from multisig

### LINKED ISSUE #2541: Improve developer experience by removing `timepoint` from `multisig`
### What is `timepoint`?
- In multisig pallet, we have `timepoint` parameter for `as_multi` and `approve_as_multi` functions. 
- Multisig operations can be uniquely identified with `AccountId, Hash` pair, and they are stored in storage as `StorageDoubleMap`. And we don't need `timepoint` to uniquely identify multisig operations.
- I believe the only advantage of having `timepoint` is:
    - say, someone created a multisig operation (`M`), with AccoundId = A, and Hash = H
    - then, this multisig operation gets approved, and dispatched -> after dispatch, it will get removed from the storage
    - but, one of the approvers of `M` (say Alice) forgot to approve it (and it's ok. It got dispatched, which means it got enough approvals already)
    - after some time, another multisig operation (`M'`) is created, with AccountId = A, and Hash = H (so basically, it is the same multisig operation again)
    - and Alice, continues to be careless by: 
        - not submitting approval on time
        - not being aware of the multisig operation got dispatched already
        - not being aware of a new one is already waiting for approval
    -  and then she decides to approve the multisig operation `M`, but in fact, she will be approving `M'`
    - and although `M'` is basically the exact copy of `M`, it will somehow be a problem that Alice approved `M'`.
    - so, yes, having a `timepoint` parameter prevents that highly unlikely and probably harmless situation... But at what cost?
 

### The problem
Having a `timepoint` parameter is significantly hindering the developer experience for multisig pallet. Due to following reasons:
1. providing the timepoint is an arduous task. Let me elaborate: 
    - no function is exposed in the library for finding the `timepoint` of a multisig operation. 
    - see this example for how many steps are required for finding the `timepoint`: https://github.com/paritytech/txwrapper-core/tree/768bb445beb2907582b2d5e13ade3be5d995af3e/packages/txwrapper-examples/multisig#timepoint-of-approveasmulti--storage
    - this complexity is probably discouraging developers from using multisig pallet and susbtrate
2. it is not obvious what `timepoint`s purpose is in the pallet. It is confusing for most of the developers to see `timepoint`

### The solution
Simply getting rid of `timepoint` in multisig pallet.

And in the scenario where Alice is too careless, and approves `M'` instead of `M`, and this becomes a problem for her, I think it is Alice to blame, not this pallet.

To me, it seems like: **_to make maybe 1 or 2 user/s happy, we are making 99.9% of the users unhappy._**

ozgunozerk:
# Description

This PR aims to improve developer experience by solving #2541. A detailed explanation is provided in the issue itself.

- What does this PR do?
    Removes `timeout` from `multisig pallet`
    It will significantly improve the developer experience
    This will be a breaking change I'm afraid.  

Fixes #2541 

# Checklist

  required) **_(it should probably be labeled with T2 for this)_**
- [x] I have made corresponding changes to the documentation
- [x] I have removed related tests

### DIFF
--- a/substrate/frame/multisig/src/benchmarking.rs
+++ b/substrate/frame/multisig/src/benchmarking.rs
@@ -79,7 +79,7 @@ benchmarks! {
 		// Whitelist caller account from further DB operations.
 		let caller_key = frame_system::Account::<T>::hashed_key_for(&caller);
 		frame_benchmarking::benchmarking::add_to_whitelist(caller_key.into());
-	}: as_multi(RawOrigin::Signed(caller), s as u16, signatories, None, call, Weight::zero())
+	}: as_multi(RawOrigin::Signed(caller), s as u16, signatories, call, Weight::zero())
 	verify {
 		assert!(Multisigs::<T>::contains_key(multi_account_id, call_hash));
 	}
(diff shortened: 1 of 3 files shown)


ozgunozerk:
Closing in favor of #2735

PR #2561, opened 2023-11-30
Add `recorded_keys` function to get recorded keys from the proof recorder

ParthDesai:
# Description

- What does this PR do?
This PR adds function to get recorded keys from proof recorder instance
This change is required to get the keys accessed by the trie backend during the runtime execution. The keys are already tracked by proof recorder, just aren't exposed publicly.
The changes are implemented by adding a public function in proof recorder that simply clones the `recorded_keys` field. It is pure addition of function and AFAIK does not affect anything.

### DIFF
--- a/substrate/primitives/trie/src/recorder.rs
+++ b/substrate/primitives/trie/src/recorder.rs
@@ -105,6 +105,13 @@ impl<H: Hasher> Clone for Recorder<H> {
 }
 
 impl<H: Hasher> Recorder<H> {
+	/// Keys for which we have recorded the trie nodes till now.
+	/// Note that this does not modify the internals, rather returns snapshot of the recorded keys.
+	/// There can be multiple storage root in case we are tracking more than one tries.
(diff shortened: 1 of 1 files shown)


substrate/primitives/trie/src/recorder.rs
```diff
@@ -105,6 +105,13 @@ impl<H: Hasher> Clone for Recorder<H> {
 }
 
 impl<H: Hasher> Recorder<H> {
+	/// Keys for which we have recorded the trie nodes till now.
+	/// Note that this does not modify the internals, rather returns snapshot of the recorded keys.
+	/// There can be multiple storage root in case we are tracking more than one tries.
```
bkchr, line 110:
```suggestion
	/// Returns [`RecordedForKey`] per recorded key per trie.
	///
	/// There are multiple tries when working with e.g. child tries.
```

skunert (approved):

ParthDesai:
@bkchr Review bot is still failing. How can we fix it?

ggwpez:
I pinged it internally to get approval for the bot. But maybe Basti force merges it 😉

Issue #2245, opened 2023-11-09
[remote-externalities] Batch calls scraping child-tree data


liamaharon:
Child-tree data is currently fetched sequentially.

Most chains to date don't have that many child-trees so this wasn't an issue, but the Rococo Contracts chain having almost 3000 child tries making this serial process way slower than it needs to be.

The keys to scrape appear to all be known ahead of time, so we should be able to batch the requests.

brunopgalvao:
https://github.com/paritytech/polkadot-sdk/blob/ea4085ab7448bb557a1558a25af164cf364e88d6/substrate/utils/frame/remote-externalities/src/lib.rs#L767-L775

1. Would you say this is where I should be looking at?
2. Can you recommend an easy way to test/run child trie remote externalities?

liamaharon:
> Would you say this is where I should be looking at?

Yes exactly 

> Can you recommend an easy way to test/run child trie remote externalities?

Add a test that scrapes live state similar to this: https://github.com/paritytech/polkadot-sdk/pull/1985/files#diff-f73dfe42c5919a5ebddb4df3294fea8e06d1b4dc0f0bba175a2c555223202cf5R1543-R1565

A remote with lots of child tries is `wss://rococo-contracts-rpc.polkadot.io:443`.

brunopgalvao:
I have some code (https://github.com/paritytech/polkadot-sdk/pull/2584) and it compliles! Now I want to run the tests but when I run the tests locally or using `wss://rococo-contracts-rpc.polkadot.io:443` or `wss://rpc.polkadot.io:443` I get the following error:
```
Dec 04 01:34:52.398 ERROR remote-ext: Error = Transport(Server returned an error status code: 429) 
```
Do I need to be behind a VPN for the Polkadot RPC nodes to work?

Locally, I am using the contracts node:
```
./target/debug/substrate-contracts-node --rpc-max-request-size 10000000 --rpc-max-response-size 10000000
```
```
TEST_WS=ws://127.0.0.1:9944 cargo test --features=remote-test -p frame-remote-externalities -- --nocapture
```

ggwpez:
What is the status of this @liamaharon @brunopgalvao ?

brunopgalvao:
> What is the status of this @liamaharon @brunopgalvao ?

I would very much like to continue working on this issue. I got an error here: https://github.com/paritytech/polkadot-sdk/issues/2245#issuecomment-1837936900

[...]

PR #2584, opened 2023-12-02
refactor to concurrently request child keys

brunopgalvao:
Closes https://github.com/paritytech/polkadot-sdk/issues/2245

> Child-tree data is currently fetched sequentially.
> 
> Most chains to date don't have that many child-trees so this hasn't been an issue, but the Rococo Contracts chain has almost 3000 child tries making this serial approach unnecessarily add a really long time to scraping keys.

This PR makes it so requests for child keys are concurrent rather than sequential.


brunopgalvao:
I am no longer with Parity I have gone rogue and decentralized therefore I no longer have permissions to push to this branch.

Re-opening from my repo here: https://github.com/paritytech/polkadot-sdk/pull/3367

PR #2599, opened 2023-12-04
More tests and checks confirming that `ledger.controller` is always correct.

gpestana:
A bonded ledger fetched with the `StakingLedger` implementation exposes a method `ledger.controller()` that returns the controller of the ledger. However, that controller is computed and stored under the `ledger.controller` field on the fly - i.e when the ledger is fetched from storage using the `StakingLedger::get` method. The controller field is never stored in storage. 

This PR add a few more tests checks and improves the ledger try-state checks to make sure these invariants hold true.


gpestana:
bot fmt

rossbulat (approved):

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1853,7 +1853,17 @@ impl<T: Config> Pallet<T> {
 
 	fn check_ledgers() -> Result<(), TryRuntimeError> {
 		Bonded::<T>::iter()
-			.map(|(_, ctrl)| Self::ensure_ledger_consistent(ctrl))
+			.map(|(stash, ctrl)| {
+				// `ledger.controller` is never stored in raw storage.
+				let raw = Ledger::<T>::get(stash).unwrap_or_else(|| {
+					Ledger::<T>::get(ctrl.clone())
+						.expect("try_check: bonded stash/ctrl does not have an associated ledger")
```
Ank4n, line 1860: Should we just log an error or panic? Since try state checks are mostly caused by older issues, we may not want CI to fail for these errors?

Ank4n (approved):

georgepisaltu (approved):

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1853,7 +1853,17 @@ impl<T: Config> Pallet<T> {
...
+				});
+				ensure!(raw.controller == None, "raw storage controller should be None");
```
kianenigma, line 1862:
```suggestion
				ensure!(raw.controller.is_none(), "raw storage controller should be None");
```

kianenigma (commented):
Is there any inconsistent/wrong state in a production network?

Issue #2195, opened 2023-11-07
PVF worker: refactor worker/job errors


mrcnski:
As part of https://github.com/paritytech/polkadot-sdk/pull/1685 I did a big refactor of errors, but they need some more work. I left it out of that PR because it was already turning into a monster.

As I was refactoring errors in #1685 I noticed some security issues. Some errors from the execute job were being treated as internal, even though malicious code can send back arbitrary errors! That's why it's important that the error handling here is robust and accurate.

### Refactor execute `WorkerResponse`

We should refactor `WorkerResponse` for more idiomatic error handling:

`type WorkerResult = Result<WorkerResponse, WorkerError>`.

That is, move the `Ok` variant into a `WorkerResponse` type and the rest of the variants into
`WorkerError`. See e.g. `JobResponse` and `JobError`.

### Move more of the execute `WorkerIntfError`s to `InternalValidationError`

Now that the worker and job are separate processes, some worker-specific errors can potentially be treated as internal now.

### Have `JobResponse`/`JobError` types for prepare like we did for execute

Right now preparation returns a `PrepareResult`, which gets turned into an `Outcome`, which gets turned into a `PrepareError` again. Note that in prepration itself, many of the variants of `PrepareError` are never constructed. We should have a `JobError` type that limits the possible errors of a job, like we did for execution jobs.

### Don't allow prepare jobs to raise non-deterministic errors

It's an existing issue, and not so critical for preparation as for execution, so I left it out of the PR. It's not critical because we treat all preparation errors as internal because we assume that real errors were caught and filtered during prechecking.

But still, we should get the story right. It could make prechecking more secure. So, preparation jobs should not trigger internal preparation errors. See `NonDeterministicPrepareError`.

## Related

Follow-up of https://github.com/paritytech/polkadot-sdk/pull/1685

s0me0ne-unkn0wn:
(1) and (2) have been addressed in #4071.

PR #2604, opened 2023-12-04
PVF: Refactor execute worker errors, treat more as internal

mrcnski:
Addresses points (1) and (2) of https://github.com/paritytech/polkadot-sdk/issues/2195. Recommended to review individual commits rather than the whole diff.

Instead of addressing points (3) and (4), I plan to move on after this PR. Preparation is not so security-critical, and it's also not clear how preparation will change with PolkaVM.

@eagr: Heads up, in case you're working on the error refactor.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4595336

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4597579

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4829255

polkadot/node/core/pvf/execute-worker/src/lib.rs
```diff
@@ -149,10 +154,26 @@ pub fn worker_entrypoint(
 			let worker_pid = process::id();
 			let artifact_path = worker_dir::execute_artifact(&worker_dir_path);
 
-			let Handshake { executor_params } = recv_execute_handshake(&mut stream)?;
+			let Handshake { executor_params } = match recv_execute_handshake(&mut stream) {
```
s0me0ne-unkn0wn, line 157: Is this change (and the following similar changes) worth it? It's a complication, not a big one, but still, a couple of dozens of lines. Does it make things better? I believe not. Neither handshake nor request contains untrusted data, and an error would mean everything went terribly wrong. It doesn't make much sense to me to communicate that back to the host. If the communication is so much broken at this point, is it even possible to communicate it back?

s0me0ne-unkn0wn (approved):
Looks good to me, brings much more structure into error processing. Left a comment with my doubts, but it's not a blocker.

polkadot/node/core/pvf/execute-worker/src/lib.rs
mrcnski: Good point! I can revert the `HostCommunication` errors here.

mrcnski: Looks like we can use `map_err` to save one line. :P

s0me0ne-unkn0wn: Okay, one more thought, then. The worker main loop is a closure that returns `io::Result<Never>` to the `run_worker()` function. Maybe those errors should be handled on the `run_worker()` level? That would still allow us to use `?` in the main loop and keep error handling simple. Not sure it makes total sense, but it probably is worth considering.

mrcnski: Thanks, that's a great idea. Right now we quit the workers on `io::Error` but continue looping (most of the time) on other kinds of errors. But it should be that if any error occurs, the worker always quits. Then we could make `run_worker` generic over both ExecuteResult and PrepareResult.

The question is, can the worker always just quit if an error occurred?
- execute: yes, an `Err` response always means that the worker dies (at least, it will after this refactor)
- prepare: no, it’s not so clear in the un-refactored code here but we return an error if compilation fails. We still want to keep the current behavior (keeping the worker alive in that case).

I’d like to refactor the prepare errors which would enable the `run_worker` refactor. But it’s more work which is probably going to become obsolete soon. It would also create more merge conflicts with the other PRs. So I will drop this idea to refactor `run_worker` as well, although it’s a good idea and I would like to do it. 

We can raise a follow-up just in case, and I'll push what I already implemented. There is some duplication of code, but this code may not need to be maintained for much longer, anyway, and it should make things a bit more robust in the meantime.

mrcnski: I realize that `send_error` wrapping the error in an `io::Error` is pretty weird. I thought about doing the refactor described above, but then I remembered that killing the prepare-worker on error causes the host to [send a `FromPool::Rip` signal](https://github.com/paritytech/polkadot-sdk/blob/dbd8d20b25b29abf138e3ad35c40df4c55bd7c5c/polkadot/node/core/pvf/src/prepare/pool.rs#L201-L205). This may race with the actual error sent to the host from the worker before it dies, though I'm not sure what exactly happens. (Note that there is no race in the equivalent [execute-worker code](https://github.com/paritytech/polkadot-sdk/blob/3d847e333d4da3c3e234b37723db9b99948c3a6f/polkadot/node/core/pvf/src/execute/queue.rs#L275-L279), so the code we have now works - internal errors should get reported.) Clearly this also requires a deeper look, so it should also be a follow-up. It can be addressed later if some of this code remains when PolkaVM is integrated.

mrcnski:
bot fmt

bkchr:
@s0me0ne-unkn0wn what is the status of this?

s0me0ne-unkn0wn:
@bkchr this is a useful PR, but it diverged a lot from the current master and needs a lot of love to push it to merge. I wonder if any of the external contributors want to sort it out. CC @eagr @Jpserrat @maksimryndin

maksimryndin:
> @bkchr this is a useful PR, but it diverged a lot from the current master and needs a lot of love to push it to merge. I wonder if any of the external contributors want to sort it out. CC @eagr @Jpserrat @maksimryndin

Hi @s0me0ne-unkn0wn, I can take this issue

s0me0ne-unkn0wn:
@maksimryndin great, very much appreciated! I'll assign it to you then. Don't forget to publish your Kusama address!

s0me0ne-unkn0wn:
> What is the best way to handle the divergence: start a new branch and bring all the changes or try to resolve conflicts in this one?

It's up to you, I don't have a strong opinion here, it would depend on how much effort is needed to resolve those conflicts. Try it, maybe you'll find out that a new branch is a good idea :)

maksimryndin:
> @maksimryndin great, very much appreciated! I'll assign it to you then. Don't forget to publish your Kusama address!

@s0me0ne-unkn0wn I think it is ready for review https://github.com/paritytech/polkadot-sdk/pull/4071. I wonder as we change an encoding of messages between the host and workers, it shouldn't be a problem as the upgrade is done both for the main binary and workers, right?

Jpserrat:
Hey @s0me0ne-unkn0wn, I was on vacation this last month 😞 . Let me know if you need help with another issue!

PR #1631, opened 2023-09-19
Upgrade libp2p to 0.52.4

altonen:
Upgrade libp2p to 0.52.4, including a fix: 

* Set Kademlia to server mode (https://github.com/paritytech/substrate/pull/14703)

### TODO
- [x] Fix 3 zombienet tests failing:
  - [x] `zombienet-substrate-0002-validators-warp-sync`
  - [ ] ~`zombienet-polkadot-functional-0005-parachains-disputes-past-session`~ The test is also flaky in other PRs and is not required for CI to succeed.
  - [x] `zombienet-polkadot-functional-0009-approval-voting-coalescing`
- [x] Uncomment and update to the actual libp2p API tests in [`substrate/client/network/src/protocol/notifications/handler.rs`](https://github.com/paritytech/polkadot-sdk/blob/7331f1796f1a0b0e9fb0cd7bf441239ad9664595/substrate/client/network/src/protocol/notifications/handler.rs#L1009).
- [x] When upgrading `multihash` crate as part of libp2p upgrade to version v0.19.1, uncomment the conversion code at https://github.com/paritytech/polkadot-sdk/blob/7547c4942a887029c11cbcfd5103f6d8315ab95c/substrate/client/network/types/src/multihash.rs#L159
- [x] Perform a burn-in.

PR #2611, opened 2023-12-04
Stabilize pov-recovery zombienet

skunert:
Smoldot sometimes stops reporting finalized blocks to us. Since we are recovering from the relay chain with a huge delay on full nodes, we can not rely on import notifications to set the best block. Because sometimes we also do not receive finality notifications, the height check in zombienet fails.

Proper solution is to update smoldot, there have been some changes since the version we use. But upgrade is blocked by version conflict which will be resolved with https://github.com/paritytech/polkadot-sdk/pull/1631.

### DIFF
--- a/cumulus/zombienet/tests/0002-pov_recovery.zndsl
+++ b/cumulus/zombienet/tests/0002-pov_recovery.zndsl
@@ -13,7 +13,6 @@ alice: reports block height is at least 20 within 600 seconds
 charlie: reports block height is at least 20 within 600 seconds
 one: reports block height is at least 20 within 800 seconds
 two: reports block height is at least 20 within 800 seconds
-three: reports block height is at least 20 within 800 seconds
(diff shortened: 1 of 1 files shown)


cumulus/zombienet/tests/0002-pov_recovery.zndsl
```diff
@@ -13,7 +13,6 @@ alice: reports block height is at least 20 within 600 seconds
 charlie: reports block height is at least 20 within 600 seconds
 one: reports block height is at least 20 within 800 seconds
 two: reports block height is at least 20 within 800 seconds
-three: reports block height is at least 20 within 800 seconds
```
michalkucharczyk, line 16: some comment with ref to 1631 would be nice.

michalkucharczyk (approved):

davxy (approved):

PR #2612, opened 2023-12-04
Fix slot_duration divide by zero panic in rococo-parachain runtime

Sophia-Gold:
Adds the fix in https://github.com/paritytech/polkadot-sdk/pull/2039 to the rococo-parachain runtime.


alindima (approved):
I don't think this fixes a divide by zero in this case, because the MinimumPeriod was always non-zero before this PR. This actually adds the experimental 0 MinimumPeriod

alindima:
> Yes, right. Bradley upgraded glutton slightly differently than I did. I actually think using zero does make slightly more sense (including as a reference for parachain times) so although this PR isn't 100% necessary, I'm leaving it open.

Makes sense, that's why I approved it 👍🏻

mrcnski (approved):

PR #2564, opened 2023-11-30
Relax approval requirements on CI files

bkchr:
- Remove the old Cumulus checks as these files do not exist anymore
- Relax requirements on CI files to also give two core devs the right to change these files
- Relax requirements on the review bot configuration itself as well.

PR #2620, opened 2023-12-05
Remove `testnets-common`

seadanda:
`testnets-common` was introduced recently to start to separate testnet parachain configurations from those of Polkadot and Kusama.

The `locks-review` and `polkadot-review` requirements are removed from `parachains-common` in https://github.com/paritytech/polkadot-sdk/pull/2564 and there are [plans](https://github.com/paritytech/polkadot-sdk/pull/2564#discussion_r1410882115) to move the Polkadot and Kusama contents of that package to the fellowship, `testnets-common` is no longer needed.

This PR removes the crate and replaces uses of it in `collectives-westend`, the only place it is currently used.


seadanda:
bot fmt

joepetrowski (approved):

PR #2626, opened 2023-12-05
[NTFs] Emit CollectionMaxSupplySet on collection create

### LINKED ISSUE #2293: [NFTs] `create` should emit `CollectionMaxSupplySet ` when `max_supply` has value in config

### Context 

When someone creates an collection with `config.max_supply = 1`.
The chain should emit `CollectionMaxSupplySet` same as this function `set_collection_max_supply `

https://github.com/paritytech/polkadot-sdk/blob/604704a84ce690eca93fd053abec1bb0a2b0ee32/substrate/frame/nfts/src/lib.rs#L1656

https://github.com/paritytech/polkadot-sdk/blob/5f4ce8026693b537beaee3aea5161ee34bac7ace/substrate/frame/nfts/src/features/create_delete_collection.rs#L67

### Expected 

I expect that `CollectionMaxSupplySet` should be emitted then `config.max_supply` has a value.

### Steps to reproduce

[](https://user-images.githubusercontent.com/9987732/282414710-907d8b3c-a98f-4872-8b4c-142cdb58e935.png)

jsidorenko:
Closes #2293 

if the max_supply is set during the collection creation, we emit the `CollectionMaxSupplySet` event

### DIFF
--- a/substrate/frame/nfts/src/features/create_delete_collection.rs
+++ b/substrate/frame/nfts/src/features/create_delete_collection.rs
@@ -67,11 +67,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		CollectionConfigOf::<T, I>::insert(&collection, config);
 		CollectionAccount::<T, I>::insert(&owner, &collection, ());
 
+		Self::deposit_event(event);
+
 		if let Some(max_supply) = config.max_supply {
 			Self::deposit_event(Event::CollectionMaxSupplySet { collection, max_supply });
 		}
 
-		Self::deposit_event(event);
 		Ok(())
 	}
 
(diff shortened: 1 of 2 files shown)


substrate/frame/nfts/src/features/create_delete_collection.rs
```diff
@@ -66,6 +66,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		CollectionConfigOf::<T, I>::insert(&collection, config);
 		CollectionAccount::<T, I>::insert(&owner, &collection, ());
+
+		if let Some(max_supply) = config.max_supply {
+			Self::deposit_event(Event::CollectionMaxSupplySet { collection, max_supply });
```
ggwpez, line 71: Is this emitted before or after the Collection Creation event?

jsidorenko: Before, you think it's better to put it after?

ggwpez: Yea, otherwise someone listening for events would possibly not be able to match it up.

bkchr (approved):

joepetrowski (approved):

ggwpez (approved):

PR #2629, opened 2023-12-05
Change utility::as_derivative index to u32

olanod:
This PR proposes to change the index in `utility::as_derivative` to be a `u32` instead of `u16`. More than being able to derive  `u16::MAX`+ accounts, `u32` can map better to existing values that need the extra bits. Personally I would use it to derive accounts for [indexed pluralities](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/src/v3/junction.rs#L129) from a parachain that are represented with a `u32`.

### DIFF
--- a/substrate/frame/utility/src/lib.rs
+++ b/substrate/frame/utility/src/lib.rs
@@ -266,7 +266,7 @@ pub mod pallet {
 		})]
 		pub fn as_derivative(
 			origin: OriginFor<T>,
-			index: u16,
+			index: u32,
 			call: Box<<T as Config>::RuntimeCall>,
 		) -> DispatchResultWithPostInfo {
 			let mut origin = origin;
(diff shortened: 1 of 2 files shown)


xlc:
This is a tx version breaking change and we should avoid them if possible.

I had similar requirement and the solution is simply `utility::as_derivative(idx1, utility::as_derivative(idx2, call))` and we can essentially support arbitrary long index with this pattern.

olanod:
Maybe I am being too optimistic but in practice I don't see this change really breaking people's code 🤔 so as a "minor breaking change" could be acceptable? We could also have it as a separate extrinsic but it doesn't feel worth it.

joepetrowski:
>  "minor breaking change" could be acceptable?

Transaction version is a single integer: it's either breaking or it's not. There is no degree.

xlc:
a tx breaking change is a tx breaking change. a new ledger app is required for it. and new ledger app costs time and money.

olanod:
Alright understandable, no breaking breaking change then. Maybe I'll approach the idea differently with a separate extrinsic(`as_derivative_big(_, index: u64, _)`?), the thing with the `BodyId::Index(u32)` is that it might not be an incrementing counter so even one single body in a parachain can already have an index greater than `u16::MAX`. 
This trick with `as_derivative` I want to pair it with a location converter specific to indexed pluralities that uses `Utility::derivative_account_id` to get an account for the plurality from the parachain account, having the interesting property that pluralities get their account in the remote system while the parachain can also control the account.

bkchr:
Going to close this. You can solve this with the trick that Bryan mentioned.

PR #2632, opened 2023-12-05
Remove hashbrown from trie cache.

cheme:
Using hashmap instead (hashset do not expose entry), to get the default random hasher her.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18002,7 +18002,6 @@ dependencies = [
  "array-bytes 6.1.0",
  "criterion 0.4.0",
  "hash-db",
- "hashbrown 0.13.2",
  "lazy_static",
  "memory-db",
  "nohash-hasher",
--- a/substrate/primitives/trie/Cargo.toml
+++ b/substrate/primitives/trie/Cargo.toml
@@ -20,7 +20,6 @@ harness = false
 [dependencies]
 ahash = { version = "0.8.2", optional = true }
 codec = { package = "parity-scale-codec", version = "3.6.1", default-features = false }
-hashbrown = { version = "0.13.2", optional = true }
 hash-db = { version = "0.16.0", default-features = false }
 lazy_static = { version = "1.4.0", optional = true }
 memory-db = { version = "0.32.0", default-features = false }
@@ -50,7 +49,6 @@ std = [
 	"ahash",
 	"codec/std",
 	"hash-db/std",
-	"hashbrown",
 	"lazy_static",
 	"memory-db/std",
 	"nohash-hasher",
--- a/substrate/primitives/trie/src/cache/shared_cache.rs
+++ b/substrate/primitives/trie/src/cache/shared_cache.rs
@@ -148,7 +148,7 @@ pub struct SharedValueCacheLimiter {
 	heap_size: usize,
 
 	/// A set with all of the keys deduplicated to save on memory.
-	known_storage_keys: HashMap<Arc<[u8]>, ()>,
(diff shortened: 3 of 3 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4616929

koute:
Can you explain why are we changing this? Is it because the hash map from `hashbrown` by default only generates a completely new seed once at startup? In which case we could probably just do:

```
use ahash::AHashSet as HashSet;
```

instead, which would still use `ahash` (which is going to be faster than the default hash in Rust's `std`), but it will use its [`RandomState`](https://github.com/tkaitchuck/aHash/blob/9f6a2ad8b721fd28da8dc1d0b7996677b374357c/src/random_state.rs#L202) which will always give a unique seed to each hash map/hash set.

Here's an example program to demonstrate:

```rust
use core::hash::BuildHasher;
use core::hash::Hasher;

fn main() {
    println!("ahash::AHashSet");
    for _ in 0..2 {
        let mut h = ahash::AHashSet::<i32>::new();
        let mut hs = h.hasher();
        let mut hh = hs.build_hasher();
        hh.write_u32(1234);
        println!("  {:?}", hh.finish());
    }

    println!("std::collections::HashSet");
    for _ in 0..2 {
        let mut h = std::collections::HashSet::<i32>::new();
        let mut hs = h.hasher();
        let mut hh = hs.build_hasher();
        hh.write_u32(1234);
        println!("  {:?}", hh.finish());
    }

    println!("hashbrown::HashSet");
    for _ in 0..2 {
        let mut h = hashbrown::HashSet::<i32>::new();
        let mut hs = h.hasher();
        let mut hh = hs.build_hasher();
        hh.write_u32(1234);
        println!("  {:?}", hh.finish());
    }
}
```

Result:

```
ahash::AHashSet
  9713553790381145631
  1617804629700195850
std::collections::HashSet
  12330284507346958173
  15536535817085373979
hashbrown::HashSet
  3076938710444694377
  3076938710444694377
```

bkchr:
Yeah, I also told @cheme the same.

cheme:
Yes, it is totally right to use hashbrown. (when I write this I did check hashbrown code on a local fork that was reallly outdated and it was still using comptime_rand, but it is not the case anymore since 2020 :).

I only let the PR to avoid having hashbrown dependency here (one less dep in the Cargo.toml), but it really don't have to be merge, so just to lower deps number.

koute:
> I only let the PR to avoid having hashbrown dependency here (one less dep in the Cargo.toml), but it really don't have to be merge, so just to lower deps number.

Well, ultimately even if you do remove the `hashbrown` dependency it won't actually get removed, because `schnellru` will always depend on `hashbrown`, so we'll still be using it transitively anyway.

Anyway, you can remove `hashbrown` if you want (as the `HashSet` in `std` is `hashbrown` anyway); I'd just like to keep the faster hasher from `ahash` as-is, since this part of the code is speed sensitive.

cheme:
Of course, this should not change the actual generated code (only make sense so user/reader don't ask themselve why we have hashbrown and also for version mgmt even if if we want to strictly check speed at each version update it may make sense to keep hashbrown).

bkchr:
> I'd just like to keep the faster hasher from `ahash` as-is, since this part of the code is speed sensitive.

Can you please do this @cheme?

cheme:
> > I'd just like to keep the faster hasher from `ahash` as-is, since this part of the code is speed sensitive.
> 
> Can you please do this @cheme?

Sure (I kind of miss the info :see_no_evil:)

PR #2643, opened 2023-12-06
PVF: ensure job processes are cleaned up, add tests

mrcnski:
Fixes a potential memory leak.

`PR_SET_PDEATHSIG` is used to terminate children when the parent dies. Note that this is subject to a race. There seems to be a raceless alternative [here](https://stackoverflow.com/a/42498370/6085242), but the concern is small enough that a bit more complexity doesn't seem worth it. Left a bit more info in the code comment.


mrcnski:
Just ran into this: noticed I had a few zombie processes on my VPS running near 100% CPU. I guess they were running the `halt` PVF, left over from some previous test run gone awry. I'm scared to look at my usage bill. 😰

s0me0ne-unkn0wn:
> I think a child can kill its parent?

Yes, if permissions allow. That may be tricky. Normally, it's always possible for a child to kill its parent unless different `setuid()`s are involved. But in a sophisticated environment (I mean namespace unsharing, cgroup settings etc.), more factors may be in place. The best way to know is to test, as always.

s0me0ne-unkn0wn (approved):
Looks good :+1:

mrcnski:
Thanks for the review! Yeah, with the [clone changes](https://github.com/paritytech/polkadot-sdk/pull/2477) the child will have its own PID namespace. It should be impossible for the child to kill its parent. But, I was curious whether our `kill_parent_node_in_emergency` function even works. I remember testing it, but that was probably also on a root environment--the internet says this shouldn't normally work.

s0me0ne-unkn0wn:
> But, I was curious whether our `kill_parent_node_in_emergency` function even works. I remember testing it, but that was probably also on a root environment--the internet says this shouldn't normally work.

I definitely tested it as non-root, but that was well before the unsharing stuff was introduced... I should re-test it and get rid of it if it's useless now.

PR #2651, opened 2023-12-07
Unique Usernames in Identity Pallet

joepetrowski:
This PR allows _username authorities_ to issue unique usernames that correspond with an account. It also provides two-way lookup, that is from `AccountId` to a single, "primary" `Username` (alongside `Registration`) and multiple unique `Username`s to an `AccountId`.

Key features:

- Username Authorities added (and removed) via privileged origin.
- Authorities have a `suffix` and an `allocation`. They can grant up to `allocation` usernames. Their `suffix` will be appended to the usernames that they issue. A suffix may be up to 7 characters long.
- Users can ask an authority to grant them a username. This will take the form `myusername.suffix`. The entire name (including suffix) must be less than or equal to 32 alphanumeric characters.
- Users can approve a username for themselves in one of two ways (that is, authorities cannot grant them arbitrarily):
  - Pre-sign the entire username (including suffix) with a secret key that corresponds to their `AccountId` (for keyed accounts, obviously); or
  - Accept the username after it has been granted by an authority (it will be queued until accepted) (for non-keyed accounts like pure proxies or multisigs).
- The system does not require any funds or deposits. Users without an identity will be given a default one (presumably all fields set to `None`). If they update this info, they will need to place the normal storage deposit.
- If a user does not have any username, their first one will be set as `Primary`, and their `AccountId` will map to that one. If they get subsequent usernames, they can choose which one to be their primary via `set_primary_username`.
- There are some state cleanup functions to remove expired usernames that have not been accepted and dangling usernames whose owners have called `clear_identity`.

TODO:

- [x] Add migration to runtimes
- [x] Probably do off-chain migration into People Chain genesis
- [x] Address a few TODO questions in code (please review)


xlc:
I kinda get it, but it will be great if someone can put some higher level intended usage of this pallet. e.g. I guess it will be deployed to the people chain? and how does this integrates with other chains? e.g. I guess one of the goal is able to allow transfer to `xlc.dot` to improve the UX? but how does it work across all the different chains? yeah XCM integration is for sure out of scope of this PR but there is no other issues / discussions about this feature else where.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4797984

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+		/// Add an `AccountId` with permission to grant usernames with a given `suffix` appended.
+		/// The authority can grant up to `allocation` usernames. To top up their allocation, they
+		/// should just issue (or request via governance) a new `add_username_authority` call.
+		#[pallet::call_index(15)]
+		#[pallet::weight(T::WeightInfo::add_username_authority())]
+		pub fn add_username_authority(
+			origin: OriginFor<T>,
+			authority: AccountIdLookupOf<T>,
+			suffix: Vec<u8>,
+			allocation: u32,
```
gpestana, line 979: Should there be a maximum allocation allowed per authority?

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+		#[pallet::call_index(16)]
+		#[pallet::weight(T::WeightInfo::remove_username_authority())]
+		pub fn remove_username_authority(
+			origin: OriginFor<T>,
+			authority: AccountIdLookupOf<T>,
+		) -> DispatchResult {
+			T::UsernameAuthorityOrigin::ensure_origin(origin)?;
+			let authority = T::Lookup::lookup(authority)?;
+			UsernameAuthorities::<T>::remove(&authority);
+			Self::deposit_event(Event::AuthorityRemoved { authority });
```
gpestana, line 1005: If the authority didn't exist before, this event may be misleading. Unless line 1003 returns early in that case.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+						maybe_authority.as_mut().ok_or(Error::<T>::NotUsernameAuthority)?;
+					ensure!(properties.allocation > 0, Error::<T>::NoAllocation);
+					properties.allocation.saturating_dec();
+					Ok(properties.suffix.clone())
+				},
+			)?;
+
+			// Verify input length before allocating a Vec with the user's input.
+			// `<` instead of `<=` because it needs one element for the point
+			// (`username` + `.` + `suffix`)
```
gpestana, line 1034:
```suggestion
			// (`username` + `.` + `suffix`).
```

substrate/frame/identity/src/lib.rs
```diff
@@ -166,7 +187,7 @@ pub mod pallet {
 		_,
 		Twox64Concat,
 		T::AccountId,
-		Registration<BalanceOf<T>, T::MaxRegistrars, T::IdentityInformation>,
+		(Registration<BalanceOf<T>, T::MaxRegistrars, T::IdentityInformation>, Option<Username>),
```
gpestana, line 190: Any special reason for not including the `Option<Username>` within `Registration`? from my pov, it seems like a cleaner approach.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+					ensure!(properties.allocation > 0, Error::<T>::NoAllocation);
+					properties.allocation.saturating_dec();
+					Ok(properties.suffix.clone())
+				},
+			)?;
+
+			// Verify input length before allocating a Vec with the user's input.
+			// `<` instead of `<=` because it needs one element for the point
+			// (`username` + `.` + `suffix`)
+			ensure!(
```
gpestana, line 1035: I'd move this check into `fn validate_unsername(username, suffix)`.

substrate/frame/identity/src/lib.rs
```diff
@@ -213,6 +234,33 @@ pub mod pallet {
...
+	/// primary username.
+	#[pallet::storage]
+	#[pallet::getter(fn username)]
+	pub(super) type AccountOfUsername<T: Config> =
+		StorageMap<_, Twox64Concat, Username, T::AccountId, OptionQuery>;
+
+	/// Usernames that an authority has granted, but that the account controller has not confirmed
+	/// that they want it. Used primarily in cases where the `AccountId` cannot provide a signature
+	/// because they are a pure proxy, multisig, etc. In order to confirm it, they should call
+	/// `accept_username`.
```
gpestana, line 256:
```suggestion
	/// [`Call::accept_username`].
```

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+			Self::insert_username(&who, username.clone());
+			Self::deposit_event(Event::UsernameSet { who: who.clone(), username });
+			Ok(Pays::No.into())
+		}
+
+		/// Remove an expired username approval. The call must include the full username, as in
+		/// `username.suffix`.
+		#[pallet::call_index(19)]
+		#[pallet::weight(T::WeightInfo::remove_expired_approval())]
+		pub fn remove_expired_approval(
```
gpestana, line 1096: I'm not a big fan of `on_idle` storage clearing, but since there's no deposit on the pre-approvals maybe we could have the clearing of expired pre approvals running on on_idle.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+			} else {
+				Err(Error::<T>::NoUsername.into())
+			}
+		}
+
+		/// Set a given username as the primary. The username should include the suffix.
+		#[pallet::call_index(20)]
+		#[pallet::weight(T::WeightInfo::set_primary_username())]
+		pub fn set_primary_username(origin: OriginFor<T>, username: Username) -> DispatchResult {
+			// ensure `username` maps to `origin` (i.e. has already been set by an authority)
```
gpestana, line 1115:
```suggestion
			// ensure `username` maps to `origin` (i.e. has already been set by an authority).
```

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,181 @@ pub mod pallet {
...
+
+		/// Remove a username that corresponds to an account with no identity. Exists when a user
+		/// gets a username but then calls `clear_identity`.
+		#[pallet::call_index(21)]
+		#[pallet::weight(T::WeightInfo::remove_dangling_username())]
+		pub fn remove_dangling_username(
+			origin: OriginFor<T>,
+			username: Username,
+		) -> DispatchResultWithPostInfo {
+			// ensure `username` maps to `origin` (i.e. has already been set by an authority)
```
gpestana, line 1136:
```suggestion
			// ensure `username` maps to `origin` (i.e. has already been set by an authority).
```

gpestana (dismissed):
Is there any reason for not adding the `Option<Username>` into the `Registration` type in the `IdentityOf` storage map? It seems more clear to me.

substrate/frame/identity/src/lib.rs
joepetrowski: The allocation needs to come from governance anyway, so IMO that's just something to cause trouble (i.e. failing referenda).

substrate/frame/identity/src/lib.rs
joepetrowski: It would need to select usernames though. I think we can write something offchain for this.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+				Error::<T>::UsernameTaken
+			);
+
+			// Insert or queue.
+			let who = T::Lookup::lookup(who)?;
+			if let Some(s) = signature {
+				// Account has pre-signed an authorization. Verify the signature provided and grant
+				// the username directly.
+				let encoded = Encode::encode(&bounded_username.to_vec());
+				Self::validate_signature(&encoded, &s, &who)?;
```
ggwpez, line 1065: I guess there is no real threat of the authority re-using these signatures? Otherwise the account nonce could be included to improve it a bit.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+				Vec::with_capacity(username.len().saturating_add(suffix.len()).saturating_add(1));
+			full_username.extend(username);
+			full_username.extend(b".");
+			full_username.extend(suffix);
+			let bounded_username =
+				Username::try_from(full_username).map_err(|_| Error::<T>::InvalidUsername)?;
+
+			// Usernames must be unique. Ensure it's not taken.
+			ensure!(
+				!AccountOfUsername::<T>::contains_key(bounded_username.clone()),
```
ggwpez, line 1055: Maybe also check `PendingUsernames`.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+		/// Accept a given username that an `authority` granted. The call must include the full
+		/// username, as in `username.suffix`.
+		#[pallet::call_index(18)]
+		#[pallet::weight(T::WeightInfo::accept_username())]
+		pub fn accept_username(
+			origin: OriginFor<T>,
+			username: Username,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+			let (approved_for, _) =
```
ggwpez, line 1083: Need to check the expiration date?

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+		}
+
+		/// Set a given username as the primary. The username should include the suffix.
+		#[pallet::call_index(20)]
+		#[pallet::weight(T::WeightInfo::set_primary_username())]
+		pub fn set_primary_username(origin: OriginFor<T>, username: Username) -> DispatchResult {
+			// ensure `username` maps to `origin` (i.e. has already been set by an authority).
+			let who = ensure_signed(origin)?;
+			ensure!(
+				AccountOfUsername::<T>::get(username.clone()).is_some(),
```
ggwpez, line 1117:
```suggestion
				AccountOfUsername::<T>::contains_key(&username),
```

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+		#[pallet::call_index(21)]
+		#[pallet::weight(T::WeightInfo::remove_dangling_username())]
+		pub fn remove_dangling_username(
+			origin: OriginFor<T>,
+			username: Username,
+		) -> DispatchResultWithPostInfo {
+			// ensure `username` maps to `origin` (i.e. has already been set by an authority).
+			let _ = ensure_signed(origin)?;
+			let who =
+				AccountOfUsername::<T>::take(username.clone()).ok_or(Error::<T>::NoUsername)?;
```
ggwpez, line 1138:
```suggestion
				AccountOfUsername::<T>::take(&username).ok_or(Error::<T>::NoUsername)?;
```

substrate/frame/identity/src/lib.rs
```diff
@@ -925,7 +1179,99 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	/// Calculate the deposit required for an identity.
+	fn calculate_identity_deposit(info: &T::IdentityInformation) -> BalanceOf<T> {
+		let bytes = info.encoded_size() as u32;
+		let byte_deposit = T::ByteDeposit::get().saturating_mul(<BalanceOf<T>>::from(bytes));
+		T::BasicDeposit::get().saturating_add(byte_deposit)
+	}
+
+	/// Validate that a username conforms to allowed characters/format. The function will validate
```
ggwpez, line 1192:
```suggestion
	/// Validate that a username conforms to allowed characters/format.
	///
	/// The function will validate
```
Normally we do one sentence and then a newline followed by more explanations. AFAIK it renders nicer in the docs.

substrate/frame/identity/src/lib.rs
```diff
@@ -925,7 +1179,99 @@ impl<T: Config> Pallet<T> {
...
+		// Verify input length before allocating a Vec with the user's input. `<` instead of `<=`
+		// because it needs one element for the point (`username` + `.` + `suffix`).
+		if let Some(l) = length {
+			ensure!(l < USERNAME_MAX_LENGTH, Error::<T>::InvalidUsername);
+		}
+		ensure!(
+			username.iter().all(|byte| byte.is_ascii_alphanumeric()),
+			Error::<T>::InvalidUsername
+		);
+		Ok(())
```
ggwpez, line 1206: I guess usernamesshould not be empty either?

substrate/frame/identity/src/lib.rs
```diff
@@ -925,7 +1179,99 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Validate that a username conforms to allowed characters/format. The function will validate
+	/// the characters in `username` and that `length` (if `Some`) conforms to the limit. It is not
+	/// expected to pass a fully formatted username here (i.e. one with any protocol-added
+	/// characters included, such as a `.`).
+	fn validate_username(username: &Vec<u8>, length: Option<u32>) -> DispatchResult {
+		// Verify input length before allocating a Vec with the user's input. `<` instead of `<=`
+		// because it needs one element for the point (`username` + `.` + `suffix`).
+		if let Some(l) = length {
+			ensure!(l < USERNAME_MAX_LENGTH, Error::<T>::InvalidUsername);
```
ggwpez, line 1200: Why is this `USERNAME_MAX_LENGTH` hard-coded? 32 could be too low for some use-cases.

substrate/frame/identity/src/lib.rs
```diff
@@ -925,7 +1179,99 @@ impl<T: Config> Pallet<T> {
...
+		}
+		ensure!(
+			username.iter().all(|byte| byte.is_ascii_alphanumeric()),
```
ggwpez, line 1203: Not sure if we should maybe normalize the usernames to lower case, just like URLs. Otherwise `bob.dot` and `Bob.dot` would be different.

substrate/frame/identity/src/lib.rs
```diff
@@ -925,7 +1179,99 @@ impl<T: Config> Pallet<T> {
...
+			),
+		};
+
+		// Enter in identity map. Note: In the case that the user did not have a pre-existing
+		// Identity, we have given them the storage item for free. If they ever call
+		// `set_identity` with identity info, then they will need to place the normal identity
+		// deposit.
+		IdentityOf::<T>::insert(&who, (reg, Some(primary_username)));
+		// Enter in username map.
+		AccountOfUsername::<T>::insert(username.clone(), &who);
```
ggwpez, line 1261: Shouldnt this only be done if `new_is_primary`?

substrate/frame/identity/src/lib.rs
```diff
@@ -943,7 +1289,7 @@ impl<T: Config> Pallet<T> {
 	pub fn reap_identity(who: &T::AccountId) -> Result<(u32, u32, u32), DispatchError> {
 		// `take` any storage items keyed by `target`
 		// identity
-		let id = <IdentityOf<T>>::take(&who).ok_or(Error::<T>::NotNamed)?;
+		let (id, _maybe_username) = <IdentityOf<T>>::take(&who).ok_or(Error::<T>::NotNamed)?;
```
ggwpez, line 1292:
```suggestion
		let (id, _maybe_username) = <IdentityOf<T>>::take(&who).ok_or(Error::<T>::NoIdentity)?;
```

substrate/frame/identity/src/migration.rs
```diff
@@ -0,0 +1,112 @@
...
+use super::*;
+use frame_support::{migrations::VersionedMigration, pallet_prelude::*, traits::OnRuntimeUpgrade};
+use log;
+
+#[cfg(feature = "try-runtime")]
+use codec::{Decode, Encode};
+#[cfg(feature = "try-runtime")]
+use sp_runtime::TryRuntimeError;
+
+pub mod versioned {
```
ggwpez, line 27: Is this the new recommended way? This `versioned` module seems kind of pointless, since we already prefixed the unversioned migration.

substrate/frame/identity/src/migration.rs
```diff
@@ -0,0 +1,112 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Storage migrations for the Identity pallet.
+
+use super::*;
+use frame_support::{migrations::VersionedMigration, pallet_prelude::*, traits::OnRuntimeUpgrade};
+use log;
```
ggwpez, line 20:
```suggestion
```

substrate/frame/identity/src/types.rs
```diff
@@ -240,6 +240,9 @@ pub trait IdentityInformationProvider:
 	/// Check if an identity registered information for some given `fields`.
 	fn has_identity(&self, fields: Self::FieldsIdentifier) -> bool;
 
+	/// Construct a default instance of `Self`.
+	fn default() -> Self;
```
ggwpez, line 244: Should rather require a `+ Default` for the trait.

substrate/frame/identity/src/types.rs
```diff
@@ -319,6 +324,29 @@ pub struct RegistrarInfo<
 	pub fields: IdField,
 }
 
+/// The number of usernames that an authority may allocate.
+type Allocation = u32;
+/// The maximum allowed length of a suffix.
+const SUFFIX_MAX_LENGTH: u32 = 7;
```
ggwpez, line 330: Can this also be configurable? There is no justification in the docs of why this is a good value.

substrate/frame/identity/src/types.rs
```diff
@@ -319,6 +324,29 @@ pub struct RegistrarInfo<
...
+pub struct AuthorityProperties {
+	/// The suffix added to usernames granted by this authority. Will be appended to usernames; for
+	/// example, a suffix of `wallet` will result in `.wallet` being appended to a user's selected
+	/// name.
+	pub suffix: Suffix,
+	/// The number of usernames remaining that this authority can grant.
+	pub allocation: Allocation,
+}
+
+/// The maximum allowed length of a username, _including_ the suffix.
```
ggwpez, line 345: And the delimiter?

substrate/frame/identity/src/types.rs
ggwpez, line 330:
```suggestion
pub const SUFFIX_MAX_LENGTH: u32 = 7;
```

substrate/frame/identity/src/types.rs
```diff
@@ -319,6 +324,29 @@ pub struct RegistrarInfo<
...
+pub(crate) const USERNAME_MAX_LENGTH: u32 = 32;
```
ggwpez, line 346:
```suggestion
pub const USERNAME_MAX_LENGTH: u32 = 32;
```
I think these should be pub, otherwise it will be difficult for externals to even discover these values.

substrate/frame/identity/src/lib.rs
```diff
@@ -331,36 +412,34 @@ pub mod pallet {
 			info: Box<T::IdentityInformation>,
 		) -> DispatchResultWithPostInfo {
```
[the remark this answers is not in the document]
joepetrowski: No, it wouldn't make sense. The user can't set it, it needs to be granted.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		/// Add an `AccountId` with permission to grant usernames with a given `suffix` appended.
+		/// The authority can grant up to `allocation` usernames. To top up their allocation, they
+		/// should just issue (or request via governance) a new `add_username_authority` call.
```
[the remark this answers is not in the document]
joepetrowski: You mean a new call like `increase_allocation`?

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			authority: AccountIdLookupOf<T>,
+			suffix: Vec<u8>,
+			allocation: u32,
+		) -> DispatchResult {
+			T::UsernameAuthorityOrigin::ensure_origin(origin)?;
+			let authority = T::Lookup::lookup(authority)?;
+			// We don't need to check the length because it gets checked when casting into a
+			// `BoundedVec`.
+			Self::validate_username(&suffix, None).map_err(|_| Error::<T>::InvalidSuffix)?;
```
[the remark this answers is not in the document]
joepetrowski: It could be either one IMO. An invalid suffix would result in every username with it attached being invalid.

substrate/frame/identity/src/lib.rs
joepetrowski: How would they re-use them? It would still be for the same username, and it'd be bad UX if an account can't transact between signing a message and the authority submitting it.

substrate/frame/identity/src/lib.rs
joepetrowski: IMO no, the expiration is to make sure storage gets cleaned up. If it's expired, they got lucky, but the end result is the same, that the username is removed from `PendingUsernames`.

substrate/frame/identity/src/lib.rs
joepetrowski: No strong reason, it could be configurable.

substrate/frame/identity/src/lib.rs
joepetrowski: No, primary refers to the username that `IdentityOf` maps to, i.e. the `AccountId => Username` mapping. But multiple usernames can map to the same `AccountId`.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4834331

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4834645

substrate/frame/identity/src/lib.rs
joepetrowski: The `Registration` only contains information associated with a registrar. It makes sense to keep the username separate, but without the need to introduce another storage map.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +966,180 @@ pub mod pallet {
...
+			// since we already ensured it is below the max length.
+			let mut full_username =
+				Vec::with_capacity(username.len().saturating_add(suffix.len()).saturating_add(1));
+			full_username.extend(username);
+			full_username.extend(b".");
+			full_username.extend(suffix);
+			let bounded_username =
+				Username::try_from(full_username).map_err(|_| Error::<T>::InvalidUsername)?;
+
+			// Usernames must be unique. Ensure it's not taken.
```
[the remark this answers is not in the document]
joepetrowski: If the authority re-issues it, yes. But having their account killed would still slash them and remove the username from storage (has been updated to clear that).

substrate/frame/identity/src/lib.rs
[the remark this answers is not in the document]
ggwpez: Yea generally for changing the config, increase/decrease alloc and changing suffix.  
Then we dont have to rely on overwriting existing entries in that case.

substrate/frame/identity/src/lib.rs
```diff
@@ -213,6 +234,33 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// A map of the accounts who are authorized to grant usernames.
+	#[pallet::storage]
+	#[pallet::getter(fn authority)]
+	pub(super) type UsernameAuthorities<T: Config> =
+		StorageMap<_, Twox64Concat, T::AccountId, AuthorityProperties, OptionQuery>;
+
+	/// Reverse lookup from `username` to the `AccountId` that has registered it. The value should
```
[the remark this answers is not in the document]
ggwpez: It will be refactored for a different use-case anyway, so should be fine.

substrate/frame/identity/src/lib.rs
ggwpez: Yea i was just wondering how it was meant. It could also be meant as insurance to the authority that their grant is only valid for a limited time, since they cannot retract it.

ggwpez (approved):

substrate/frame/identity/src/lib.rs
joepetrowski: Ah I see. Yeah in that case we're relying on people to actively patrol expired names and remove them so they don't linger forever.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +1012,185 @@ pub mod pallet {
...
+
+		/// Remove an expired username approval. The call must include the full username, as in
+		/// `username.suffix`.
+		#[pallet::call_index(19)]
+		#[pallet::weight(T::WeightInfo::remove_expired_approval())]
+		pub fn remove_expired_approval(
+			origin: OriginFor<T>,
+			username: Username<T>,
+		) -> DispatchResultWithPostInfo {
+			let _ = ensure_signed(origin)?;
```
gupnik, line 1154: Shouldn't this increase the allocation of the authority (as the username as not used)?

substrate/frame/identity/src/lib.rs
```diff
@@ -331,36 +448,34 @@ pub mod pallet {
...
+					Registration {
+						info: *info,
+						judgements: BoundedVec::default(),
+						deposit: Zero::zero(),
+					},
+					None,
+				),
 			};
 
+			let new_deposit = Self::calculate_identity_deposit(&id.info);
```
gupnik, line 472: Seems like a good candidate for using the `Consideration` API.

substrate/frame/identity/src/lib.rs
gupnik: Task API can be used for this.

substrate/frame/identity/src/types.rs
```diff
@@ -253,6 +253,8 @@ pub trait IdentityInformationProvider:
 ///
 /// NOTE: This is stored separately primarily to facilitate the addition of extra fields in a
 /// backwards compatible way through a specialized `Decode` impl.
+//
+// TODO: Perhaps just move `username` into a field of `Registration`?
```
gupnik, line 257: Probably good to create an issue for this?

gupnik (approved):

substrate/frame/identity/src/types.rs
joepetrowski: I actually removed it, just haven't pushed the commit yet.

substrate/frame/identity/src/lib.rs
joepetrowski: I think there is another effort to migrate this whole pallet. Didn't want to mix that into this.

substrate/frame/identity/src/lib.rs
```diff
@@ -886,6 +1012,185 @@ pub mod pallet {
...
+			// Ensure origin is a Username Authority and has an allocation. Decrement their
+			// allocation by one.
+			let sender = ensure_signed(origin)?;
+			let suffix = UsernameAuthorities::<T>::try_mutate(
+				&sender,
+				|maybe_authority| -> Result<Suffix<T>, DispatchError> {
+					let properties =
+						maybe_authority.as_mut().ok_or(Error::<T>::NotUsernameAuthority)?;
+					ensure!(properties.allocation > 0, Error::<T>::NoAllocation);
+					properties.allocation.saturating_dec();
```
seadanda, line 1084: Should this be done at the point just before the username is inserted/queued, rather than here? There are a few different ways it could fail between here and insertion.

I suppose it just depends whether the allocation limit is intended as pressure for registrars to make fewer mistakes to limit pointless work being done, vs purely as a storage limit

seadanda (approved):

substrate/frame/identity/src/lib.rs
joepetrowski: The whole transaction should be atomic though.

ggwpez: Yes and there no re-entrancy possible here either.

seadanda: Ah I didn't realise that

Polkadot-Forum:

https://forum.polkadot.network/t/people-chain-launch-and-identity-migration-plan/5930/3

substrate/frame/identity/src/migration.rs
```diff
@@ -0,0 +1,124 @@
...
+				}
+			}
+			if !interrupted {
+				log::info!("all {} identities migrated", translated);
+			}
+
+			weight.saturating_accrue(T::DbWeight::get().reads_writes(translated, translated));
+			weight.saturating_accrue(T::DbWeight::get().writes(1));
+			weight
+		}
```
wischli, line 111: This migration is missing a storage version bump from v0 to v1. How can this slip through if even `try-runtime` prints an error after applying the migration?

```
[2024-05-29T08:46:53Z INFO  runtime::identity::migration::v1] pre-upgrade state contains '35' identities.
[2024-05-29T08:46:53Z INFO  runtime::identity::migration::v1] running storage migration from version 0 to version 1.
[2024-05-29T08:46:53Z INFO  pallet_identity::migration::v1] all 35 identities migrated
[2024-05-29T08:46:53Z INFO  pallet_identity::migration::v1] post-upgrade expects '35' identities to have been migrated.
[2024-05-29T08:46:53Z INFO  runtime::identity::migration::v1] migrated all identities.
[2024-05-29T08:46:53Z ERROR runtime::frame-support] Identity: On chain storage version StorageVersion(0) doesn't match current storage version StorageVersion(1).
```

I hope this is an exception because manually bumping storage versions should be avoided at all costs.

wischli: Update: Sorry, I see this is covered by the `versioned` module wrapper.

PR #2653, opened 2023-12-07
[NFTs] Fix consumers issue

jsidorenko:
When we call the `set_accept_ownership` method, we increase the number of account consumers, but then we don't decrease it on collection transfer, which leads to the wrong consumers number an account has.


substrate/frame/nfts/src/features/transfer.rs
```diff
@@ -124,38 +124,39 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 			// Update account ownership information.
 			CollectionAccount::<T, I>::remove(&details.owner, &collection);
-			CollectionAccount::<T, I>::insert(&owner, &collection, ());
+			CollectionAccount::<T, I>::insert(&new_owner, &collection, ());
 
-			details.owner = owner.clone();
-			OwnershipAcceptance::<T, I>::remove(&owner);
+			details.owner = new_owner.clone();
+			OwnershipAcceptance::<T, I>::remove(&new_owner);
+			frame_system::Pallet::<T>::dec_consumers(&new_owner);
```
joepetrowski, line 156: Shouldn't this be the old owner?

substrate/frame/nfts/src/tests.rs
```diff
@@ -614,8 +614,13 @@ fn transfer_owner_should_work() {
 			Nfts::transfer_ownership(RuntimeOrigin::signed(account(1)), 0, account(2)),
 			Error::<Test>::Unaccepted
 		);
+		assert_eq!(System::consumers(&account(2)), 0);
+
 		assert_ok!(Nfts::set_accept_ownership(RuntimeOrigin::signed(account(2)), Some(0)));
+		assert_eq!(System::consumers(&account(2)), 1);
+
 		assert_ok!(Nfts::transfer_ownership(RuntimeOrigin::signed(account(1)), 0, account(2)));
+		assert_eq!(System::consumers(&account(2)), 1); // one consumer is added due to deposit repatriation
```
joepetrowski, line 623: It looks unchanged

jsidorenko: 1) `set_accept_ownership` increases the number of consumers
2) `transfer_collection` will need to decrease that number, but it additionally repatriates the deposit, which adds one consumer. So, without that fix, the new owner will have 2 consumers added instead of just one on collection transfer

bkchr (approved):

ggwpez (approved):
Can it be that this is already causing reference counting issues on-chain?  
So maybe we need a migration as follow-up to correct them?

jsidorenko:
> Can it be that this is already causing reference counting issues on-chain? So maybe we need a migration as follow-up to correct them?

Is there a way to detect and get a proper count of those buggy references?

bkchr:
> > Can it be that this is already causing reference counting issues on-chain? So maybe we need a migration as follow-up to correct them?
> 
> Is there a way to detect and get a proper count of those buggy references?

We would need to write some script and scrape these buggy references manually.

PR #2660, opened 2023-12-08
Update proc-macro-crate

bkchr:


substrate/frame/contracts/fixtures/Cargo.toml
```diff
@@ -16,7 +16,7 @@ anyhow = "1.0.0"
 [build-dependencies]
 parity-wasm = "0.45.0"
 tempfile = "3.8.1"
-toml = "0.8.8"
+toml = "0.8.2"
```
bkchr, line 19: I needed to pin to `toml_edit` to `0.20.3` to not break semver by requiring an higher version of rustc..

dmitry-markin (approved):

Issue #2662, opened 2023-12-08
PVF: potential "could not create a temporary directory" error


mrcnski:
It is possible to run into this warning. It is caused by us doing security checks in the PVF cache directory (`/tmp/.tmpIcLriO` in the warning) before creating this cache directory.

```
WARN parachain::pvf: 🚨 Some security issues have been detected. 
Running validation of malicious PVF code has a higher risk of compromising this machine.
  - Optional: Cannot unshare user namespace and change root, which are Linux-specific kernel security features: could not create a temporary directory in "/tmp/.tmpIcLriO": No such file or directory (os error 2) at path "/tmp/.tmpIcLriO/check-can-unshare-4XnFgA"
```

## Severity: annoyance

This should only occur if the PVF cache path (`/tmp/.tmpIcLriO` above) doesn't exist, i.e. the node is running for the first time.

On Linux 5.13+ it will be only a warning. On Linux 5.12- it's an error forcing shutdown of the node.

**Mitigation:** The validator will have to make sure the directory exists and restart the node.

This is a mild annoyance and should be mentioned in the release notes, but not worth a backport.

yayoi-k:
I found that simply restarting the node resolves the 'no such file or directory' error.  My error message was slightly different than what was described.  In my error, the first portion 'could not create temp directory in' references '.../pvf-artifacts'

`could not create a temporary directory in "/home/polkadot/.local/share/polkadot/chains/ksmcc3/db/full/pvf-artifacts": No such file or directory (os error 2) at path "/home/polkadot/.local/share/polkadot/chains/ksmcc3/db/full/pvf-artifacts/check-can-unshare-M7pL8f"`

mrcnski:
Good to hear. Thanks for letting us know @yayoi-k!

PR #2663, opened 2023-12-08
PVF: fix unshare "could not create temporary directory"; refactor

mrcnski:
The fix is just a couple lines in `host.rs`. The rest of the PR is a refactor. The refactor makes the code less verbose and should give the error messages more context.

Closes https://github.com/paritytech/polkadot-sdk/issues/2662


alexggh (approved):

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -201,13 +201,22 @@ impl Artifacts {
...
 		let mut artifacts = Self { inner: HashMap::new() };
-		artifacts.insert_and_prune(cache_path).await;
+		match artifacts.insert_and_prune(cache_path).await {
+			Ok(()) => (),
+			Err(err) => gum::error!(
+				target: LOG_TARGET,
+				"could not initialize artifacts cache: {}",
+				err,
+			),
+		}
```
lexnv, line 215: nit: No need to explicitly match here

```suggestion
		artifacts.insert_and_prune(cache_path).await.map_err(|err| gum::error!(
				target: LOG_TARGET,
				"could not initialize artifacts cache: {err}",
			));
```

lexnv (approved):

PR #2668, opened 2023-12-09
Add paseo network `--chain` option and chain spec

al3mart:
After some conversations around the topic of launching a community staging network ( [forum post 1](https://forum.polkadot.network/t/a-new-test-network-for-polkadot/4325), [forum post 2](https://forum.polkadot.network/t/the-new-polkadot-community-testnet/4956) ). And Erin having introduced the community testnet in the [last episode of AAG](https://www.youtube.com/watch?v=uzJEGVG78_E&t=1612s).

I would like to include the `paseo` options as one of the chains available for launching using the polkadot node.

Note that the chain spec implementation is not yet included at this stage.

UPDATE:

Merged master with rebase enabled and introduced quite some noise in this PR.
Relevant changes are in:
- `polkadot/node/service/chain-specs/paseo.json`
- `polkadot/node/service/src/chain_spec.rs`
- `polkadot/cli/src/command.rs`


polkadot/cli/src/command.rs
```diff
@@ -91,6 +91,7 @@ impl SubstrateCli for Cli {
 			"polkadot" => Box::new(service::chain_spec::polkadot_config()?),
 			name if name.starts_with("polkadot-") && !name.ends_with(".json") =>
 				Err(format!("`{name}` is not supported anymore as the polkadot native runtime no longer part of the node."))?,
+			"paseo" => Box::new(service::chain_spec::paseo_config()?),
```
ggwpez, line 94: We probably also want a `dev` config for local testing.

bkchr (changes requested):
I don't see any need for this. You can just pass chain spec. Why would you need this?

al3mart:
> I don't see any need for this. You can just pass chain spec. Why would you need this?

While it is true that there is no need for it, infra maintainers expressed a preference for having the chain spec option embedded in the node.

If there aren't string arguments against including this and it doesn't set any unwanted precedents, I'd say that it doesn´t hurt to have this as part of the node.

bkchr:
> infra maintainers expressed a preference for having the chain spec option embedded in the node.

Why? Then people need to open prs here again to update the bootnodes etc. IMO it would be better to have this external. Maybe the bare minimum could be the raw version, but clearly not more, but even for the raw version I'm skeptical given what I said before.

lovelaced:
@bkchr for developer experience/tutorial purposes it would make more sense to just be able to download the binary and pass an argument rather than have someone have to locate the chainspec and download it and then pass it to the binary.

lovelaced:
> I mean we will probably not point beginner level docs to do this and it is also not rocket science :P 

Perhaps, but I'm mindful that Paseo is likely going to be many new developers' first interaction with Polkadot as it's likely to be the entrypoint for both parachain and dapp development and would like it to be as painless as possible. Bootnodes are unlikely to change often.

xlc:
i don’t know why this is necessary. if you really want to offer a binary / docker image for beginners. just build a new one on a different repo.

lovelaced:
Maybe I'm missing something here, what is the reasoning for not adding this?

xlc:
if there is no good reason to add to this repo, that’s a good reason to not to add to this repo. 

this is polkadot-sdk and IMO already too big and contains too many unnecessary thins to make maintenance complicated

lovelaced:
> if there is no good reason to add to this repo, that’s a good reason to not to add to this repo.
> 
> this is polkadot-sdk and IMO already too big and contains too many unnecessary thins to make maintenance complicated

Of course, and we definitely don't want to make anything *worse* - but consider this as a replacement for Rococo, which will be removed in the coming months, and this is the official polkadot binary used for all development purposes. The future maintenance burden of this should fall on the proposer. I would consider the ability to easily launch an important testnet with the official binary a good reason, but of course it is not yet launched so if the network needs to prove itself first that's reasonable.

hitchhooker:
where do you hide networks chain spec?

al3mart:
I have added Paseo's raw chainspec to the code base.

This should still keep the noise out of this repo. Happy to wait to get it merged if the network needs to proof useful to the ecosystem. Although, having this option available would be very friendly to infra maintainers.

bkchr (approved):

bkchr:
I'm fine with moving on with this. I mean I'm still not 100% convinced, but the current diff is not that much, so it is fine.

ggwpez (approved):

polkadot/node/service/chain-specs/paseo.json
```diff
@@ -4,9 +4,20 @@
   "chainType": "Live",
   "bootNodes": [
     "/dns/paseo.bootnode.amforc.com/tcp/30333/wss/p2p/12D3KooWFD81HC9memUwuGMLvhDDEfmXjn6jC4n7zyNs3vToXapS",
-    "/dns/paseo.bootnode.amforc.com/tcp/30344/p2p/12D3KooWFD81HC9memUwuGMLvhDDEfmXjn6jC4n7zyNs3vToXapS"
+    "/dns/paseo.bootnode.amforc.com/tcp/30344/p2p/12D3KooWFD81HC9memUwuGMLvhDDEfmXjn6jC4n7zyNs3vToXapS",
+    "/dns/paseo-bootnode.radiumblock.com/tcp/30333/p2p/12D3KooWADeaC8zag4Qrb4GosSn65MmfVZztRPMaBdgZnQqXRo",
+    "/dns/paseo-bootnode.radiumblock.com/tcp/30335/wss/p2p/12D3KooWeayZC8zag4Qrb4GosSn65MmfVZztRPMaBdgZnQqXRo",
```
hitchhooker: @al3mart  [same fix](https://github.com/paseo-network/runtimes/commit/6b2adf9f557e2041e94a7c6aeeaa0dde344a03a1) required here or going to get [parse errors](https://github.com/paseo-network/runtimes/issues/18) using --chain paseo.

al3mart: Thanks @hitchhooker! I have pushed the fix

Cargo.toml
```diff
@@ -175,6 +175,7 @@ members = [
 	"polkadot/node/service",
 	"polkadot/node/subsystem",
 	"polkadot/node/subsystem-bench",
+	"polkadot/node/subsystem-bench",
```
jsidorenko, line 178:
```suggestion
```

PR #2390, opened 2023-11-17
Set clippy lints in workspace (requires rust 1.74)

gilescope:
We currently use a bit of a hack in `.cargo/config` to make sure that clippy isn't too annoying by specifying the list of lints.

There is now a stable way to define lints for a workspace. The only down side is that every crate seems to have to opt into this so there's a *few* files modified in this PR.

Dependencies:

- [x] PR that upgrades CI to use rust 1.74 is merged.

PR #2669, opened 2023-12-09
[ci] Configure clippy to fail on warnings

bkontur:
**Before this PR:** clippy has one warning, but did not fail:
https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4672812
```
warning: field `cache_dir` is never read
  --> polkadot/node/core/pvf/benches/host_prepare_rococo_runtime.rs:32:2
   |
30 | struct TestHost {
   |        -------- field in this struct
31 |     // Keep a reference to the tempdir as it gets deleted on drop.
32 |     cache_dir: tempfile::TempDir,
   |     ^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
warning: `polkadot-node-core-pvf` (bench "host_prepare_rococo_runtime") generated 1 warning
```

**After fix in this PR:** clippy fails on warnings.
https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4672949


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4672949

.gitlab/pipeline/check.yml
```diff
@@ -5,6 +5,7 @@ cargo-clippy:
     - .common-refs
     - .pipeline-stopper-artifacts
   script:
+    # The removal of `RUSTFLAGS` is related to the configuration of `cargo-clippy` in `.cargo/config.toml`.
```
liamaharon, line 8: Can we be more specific here?

polkadot/node/core/pvf/benches/host_prepare_rococo_runtime.rs
```diff
@@ -29,6 +29,7 @@ const TEST_PREPARATION_TIMEOUT: Duration = Duration::from_secs(30);
 
 struct TestHost {
 	// Keep a reference to the tempdir as it gets deleted on drop.
```
liamaharon, line 31:
```suggestion
	// Keep a reference to the tempdir otherwise it gets deleted on drop.
```

liamaharon (approved):

sergejparity (approved):

.gitlab/pipeline/check.yml
liamaharon: also pinging @athei and @gilescope https://github.com/paritytech/substrate/pull/9694/files#diff-9a4f3e4537ebb7474452d131b0d969d89a51286f4269aac5ef268e712be17268R7

gilescope: If we've updated the compiler to 1.74 then there's a clippy PR that we can land that removes our usage of the hack in .cargo/config.toml as we can now store lints in the workspace toml file. I think in my PR there's a couple of new crates that need to have their Cargo.tomls tweaked and then it should run green. https://github.com/paritytech/polkadot-sdk/pull/2390

bkontur: @gilescope 
Giles, cool, thank you for the comment :)
Do you plan to finish #2390? If not, I could. Just, please, let me know.


CI has 1.74:
```
$ cargo --version
cargo 1.74.0 (ecb9851af 2023-10-18)
...
$ cargo +nightly --version
cargo 1.75.0-nightly (b4d18d4bd 2023-10-31)
```

Issue #1833, opened 2023-10-10
Inconsistent Transferable Balance Calculation Logic


ahmadkaouk:
An inconsistency has been identified in the Polkadot SDK concerning the logic used for computing transferable balances. This divergence may mislead users about the actual spendable balance, potentially leading to transaction failures due to inadequate fee coverage.

The inconsistency arises from the utilization of different formulas:
1. Old Formula: `transferable = free - max(frozen, reserved)`.
2. New Formula: `transferable = free - (frozen - reserved)`.

This originates from the `CurrencyAdapter`, which relies on the outdated `Currency` trait as illustrated  [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/transaction-payment/src/payment.rs#L84). The new transferable balance is computed utilizing the `reducible_balances` function defined [here](https://github.com/paritytech/polkadot-sdk/blob/e8da320734ae44803f89dd2b35b3cfea0e1ecca1/substrate/frame/balances/src/impl_fungible.rs#L44).

The discrepancy is particularly noticeable in the `eth.getBalance` method in Frontier, where the new formula is employed. Under certain circumstances, a positive balance may be displayed, indicating available funds. However, transactions may fail due to an insufficient balance to cover the fees, given that the old formula is utilized for fee calculations.

- Is the discrepancy in the logic for calculating "transferable balance" concerning fees a deliberate design or an oversight?
- Is there a rationale behind not extending the new logic to fee calculations, and if so, what is the reasoning behind this approach?

kianenigma:
IN any case, an item missing (and rather high priority) in https://github.com/paritytech/polkadot-sdk/issues/226 is then migrate `CurrencyAdapter` to also work with `fungible`, I added it there.

michalisFr:
@ahmadkaouk Just to point that if the account cannot be reaped (which should be the case when there's a hold and/or freeze) or we don't want it to be reaped, the transferable balance calculation is this:
`transferable = free - max(frozen - reserved, ED)`
since now the ED is a requirement over the free balance, not the total balance.

liamaharon:
Spendable balance indeed I think should be `transferable = free - max(frozen, reserved)`, NOT `transferable = free - (frozen - reserved)` (@bkchr can you confirm?).

I will work on this.

liamaharon:
@ahmadkaouk I have a couple of questions: 

1. Could you please clarify why you determined this regression to have originated in the `CurrencyAdaptor`?

[...]

PR #2683, opened 2023-12-11
Ref docs frame currency

juangirini:
Closes https://github.com/paritytech/polkadot-sdk-docs/issues/49

polkadot address: 12zshjtBEmK43evpHZXyQEVHkfVhW7J5sUvs64EmusSCxPUp


docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,99 @@
...
 //!
-//! Notes:
+//! TODO:
 //!
-//! - History, `Currency` trait.
-//! - `Hold` and `Freeze` with diagram.
-//! - `HoldReason` and `FreezeReason`
-//! 	- This footgun: https://github.com/paritytech/polkadot-sdk/pull/1900#discussion_r1363783609
+//! - [x] History, `Currency` trait.
+//! - [ ] `Hold` and `Freeze` with diagram.
```
juangirini, line 6: not sure what kind of diagram would make it clear... maybe some bars diagram with Freeze and Hold balance overlapped?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,99 @@
...
+//! - [x] `HoldReason` and `FreezeReason`
+//! - [ ] This footgun: https://github.com/paritytech/polkadot-sdk/pull/1900#discussion_r1363783609
```
juangirini, line 8: I don't think we should include discussions in the ref docs, as from my point of view this section would get obsolete as soon as the discussion reached an agreement, in which case we could document the output. But I am open to arguments

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4687288

docs/sdk/src/reference_docs/frame_currency.rs
liamaharon: It should not only be holds and freeze overlapped, but also free, spendable, untouchable, see https://github.com/paritytech/polkadot-sdk/issues/1833#issuecomment-1805764506

However, I don't think it should live here... These are concepts related to the Balances pallet, so I think should live in the Balances pallet docs.

docs/sdk/src/reference_docs/frame_currency.rs
liamaharon: This is a fungible / fungibles thing, should probably be documented in the fungible docs?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,99 @@
...
-//! Notes:
+//! TODO:
 //!
-//! - History, `Currency` trait.
-//! - `Hold` and `Freeze` with diagram.
-//! - `HoldReason` and `FreezeReason`
-//! 	- This footgun: https://github.com/paritytech/polkadot-sdk/pull/1900#discussion_r1363783609
+//! - [x] History, `Currency` trait.
+//! - [ ] `Hold` and `Freeze` with diagram.
+//! - [x] `HoldReason` and `FreezeReason`
```
liamaharon, line 7: These are specifically related to fungible, should be documented in the fungible docs?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,99 @@
...
+//! *Freezes* can overlap with *Holds*. Since *Holds* are designed to be infallibly slashed, this
+//! means that any logic using a *Freeze* must handle the possibility of the frozen amount being
+//! reduced, potentially to zero. A permissionless function should be provided in order to allow
+//! bookkeeping to be updated in this instance. E.g. some balance is frozen when it is used for
+//! voting, one could use held balance for voting, but nothing prevents this frozen balance from
+//! being reduced if the overlapping hold is slashed.
+//!
+//! Both *Holds* and *Freezes* require an identifier, `HoldReason` and `FreezeReason` respectively,
+//! which is configurable and is expected to be an enum aggregated across all pallet instances of
+//! the runtime.
```
liamaharon, line 77: All this I feel belongs in fungible pallet docs?

liamaharon (commented):
I think it's worth defining a clear goal of what this ref doc should achieve, then outlining the sections and writing each sentence to ensure they are maximally fulfilling this goal. 

I understood it as something to disambiguate the different traits, trait implementations, and terminology we have for handling tokens. We should be careful not to let the scope creep and start writing documentation for the actual traits / implementations here.

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,99 @@
 //! FRAME Currency Abstractions and Traits
```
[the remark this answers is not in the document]
juangirini: This does not cover NFTs which are also tokens. I don't see this potential for confusion, specially it the person reads the document

juangirini:
> I think it's worth defining a clear goal of what this ref doc should achieve, then outlining the sections and writing each sentence to ensure they are maximally fulfilling this goal.
> 
> I understood it as something to disambiguate the different traits, trait implementations, and terminology we have for handling tokens. We should be careful not to let the scope creep and start writing documentation for the actual traits / implementations here.

Indeed, we have different goals in mind for this document. We definitely don't want to talk (only) about the Currency trait here, but what is Currency in FRAME if not a very important use case of the fungible traits?
What would be a valid structure for this document from your point of view?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4730425

docs/sdk/src/reference_docs/frame_currency.rs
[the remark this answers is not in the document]
liamaharon: Maybe it does make sense to also to discuss NFT traits and trait impls in this doc. It would be similarly helpful to disambiguate the different non-fungible token pallets and traits, just like it's helpful to disambiguate the different fungible token pallets and traits. 

If we don't I still don't think it should be called 'currency'. A 'currency' is a specific type of fungible tokens, and this doc so far explores traits and trait impls that can be used for much more than just currencies.

liamaharon:
> what is Currency in FRAME if not a very important use case of the fungible traits?

Not sure what you mean here? 

> Indeed, we have different goals in mind for this document. 

What is/was your goal for this doc? 

> What would be a valid structure for this document from your point of view?

Primarily I think the goal of an overarching doc for token traits and trait impls should be to allow a reader with zero prior knowledge of Substrate come away with 

- A good understanding of the patterns used for tokens in Substrate (traits and trait impls), and why we use these patterns
- A good understanding of which traits and pallets exist, the differences between the traits and pallets, and how to choose a trait or pallet for their use case

This could look something loosely like

```markdown
## Tokens in Substrate

briefly explains 
- how Substrate follows the pattern of traits and trait implementation for tokens
- why this approach was chosen

### Fungible Traits and Pallets

list and compare fungible traits and pallets

### Non-fungible Traits and Pallets

list and compare non-fungible traits and pallets
```

I don't think this needs to be long. Each sentence should be genuinely informative and contribute to the goal of the document.

Let me know what you think, I'm open to other ideas.

juangirini:
> > what is Currency in FRAME if not a very important use case of the fungible traits?
> 
> Not sure what you mean here?

This is supposed to be a Currency documentation, being currency a fungible token I think we should talk about both of them.
 
> > Indeed, we have different goals in mind for this document.
> 
> What is/was your goal for this doc?

Give a high level overview of currency and fungible traits.

> 
```md
## Tokens in Substrate

briefly explains 
- how Substrate follows the pattern of traits and trait implementation for tokens
- why this approach was chosen

### Fungible Traits and Pallets

list and compare fungible traits and pallets

### Non-fungible Traits and Pallets

list and compare non-fungible traits and pallets
```

I don't think we should generalize it to "all" tokens, and therefore I wouldn't touch NFT in the Currency documentation, that should belong somewhere else.
The current PR already explains the `Fungible Traits and Pallets` part, this could of course improved and adapted.
I like the idea of the first part you propose as long as we stick to currency
> - how Substrate follows the pattern of traits and trait implementation for tokens
> - why this approach was chosen
WDYT? Do you want to add that to the PR?

juangirini:
The more broader/generic tokens document will be taken care of in https://github.com/paritytech/polkadot-sdk-docs/issues/70

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4739247

JuaniRios:
Just wanted to comment that we spent 2 days researching the move from currency to fungible/fungibles, and it was very hard to understand.
This doc will definitely help us and other people in the future :)

Our main pain points that we had to clear up:

### Do freezes make the same changes to `AccountData` storage as locks?

Yes

### Should all previous locks be freezes, and all reserves be now holds?

No, because previously a lock could assume that the free balance wouldn't be reduced below that amount. But now if a hold happens, then you are considering also the frozen/locked balance, which might lead to unexpected behaviour.
[This is why pallet-staking had to go from lock to hold](https://arc.net/l/quote/kbdcfgjg)

### If a pallet used a currency trait, can another pallet use a fungible/fungibles trait?

Depends very much on the expected behavior. For example if one pallet is using locks, but you are ok with that amount being considered in another pallet's hold, then you could do that.
In our case we could use pallet-vesting which uses `LockableCurrency`, because if we were to replace it with the fungible trait it would be a Freeze (we don't mind having a hold on the frozen amount). And from our understanding freezes and locks produce the same outcome on the `AccountData` storage.

### What are the bookkeeping functions mentioned by gav?

In [here](https://arc.net/l/quote/zsnzqtwg) and [here](https://arc.net/l/quote/bqnwnfgt) Gav explains that since a frozen balance cannot guarantee that the free will ever go below that (due to slashing on holds), then we should add a bookkeeping function which deals with the case where we find less free balance than the frozen one.
We are not sure what this means or how it would look like in the code.

It is mentioned that this is implemented in the `conviction-voting` pallet, but we couldn't find where that happened. 

### Why is reducable_balance when Polite calculated like that?
Here [Liam's visualisation](https://arc.net/l/quote/xvqdehna) came in handy.

I think we would benefit from also explaining how the currency and fungible traits relate to the AccountData storage.

If I got anything wrong from my self-answered questions, please let me know :)

juangirini:
> Just wanted to comment that we spent 2 days researching the move from currency to fungible/fungibles, and it was very hard to understand. This doc will definitely help us and other people in the future :)
> 
> Our main pain points that we had to clear up:
> 
> ### Do freezes make the same changes to `AccountData` storage as locks?
> Yes
> 
> ### Should all previous locks be freezes, and all reserves be now holds?
> No, because previously a lock could assume that the free balance wouldn't be reduced below that amount. But now if a hold happens, then you are considering also the frozen/locked balance, which might lead to unexpected behaviour. [This is why pallet-staking had to go from lock to hold](https://arc.net/l/quote/kbdcfgjg)
> 
> ### If a pallet used a currency trait, can another pallet use a fungible/fungibles trait?
> Depends very much on the expected behavior. For example if one pallet is using locks, but you are ok with that amount being considered in another pallet's hold, then you could do that. In our case we could use pallet-vesting which uses `LockableCurrency`, because if we were to replace it with the fungible trait it would be a Freeze (we don't mind having a hold on the frozen amount). And from our understanding freezes and locks produce the same outcome on the `AccountData` storage.
> 
> ### What are the bookkeeping functions mentioned by gav?
> In [here](https://arc.net/l/quote/zsnzqtwg) and [here](https://arc.net/l/quote/bqnwnfgt) Gav explains that since a frozen balance cannot guarantee that the free will ever go below that (due to slashing on holds), then we should add a bookkeeping function which deals with the case where we find less free balance than the frozen one. We are not sure what this means or how it would look like in the code.
> 
> It is mentioned that this is implemented in the `conviction-voting` pallet, but we couldn't find where that happened.
> 
> ### Why is the output of reducable_balance when the parameter fortitude=Polite calculated like that?
> Here [Liam's visualisation](https://arc.net/l/quote/xvqdehna) came in handy.
> 
> I think we would benefit from also explaining how the currency and fungible traits relate to the AccountData storage.
> 
> If I got anything wrong from my self-answered questions, please let me know :)

Hey @JuaniRios thanks for your valuable input. We've loosely spoken about fungible/s traits in this doc as there's another one more specific to it https://github.com/paritytech/polkadot-sdk-docs/issues/70. I am cc'ing @liamaharon as he's working on it

liamaharon:
Thank you @JuaniRios, this is super useful! I have a draft PR for token doc here, will add a TODO there to make sure all these points have been covered. If you have any other suggestions for these docs you can comment in that pr :)

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
...
+//! different forms of digital assets within the blockchain environment, ensuring that the economic
+//! aspects of the chain are robust and adaptable to different use cases.
 //!
-//! - History, `Currency` trait.
-//! - `Hold` and `Freeze` with diagram.
-//! - `HoldReason` and `FreezeReason`
-//! 	- This footgun: <https://github.com/paritytech/polkadot-sdk/pull/1900#discussion_r1363783609>
+//! ## The Currency Trait
+//!
+//! The [`Currency`](../../../frame_support/traits/tokens/currency/index.html) trait was initially
```
kianenigma, line 10: FYI you can also list links in the bottom, if it helps with readability.

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
...
+//! introduced in Substrate to manage the native token balances. This trait was later deprecated in
```
kianenigma, line 11: Reason: the interface is only sensible in a system that has only one currency type. There is no such thing in it as "id".

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
-//! FRAME Currency Abstractions and Traits
+//! # FRAME Currency Abstractions and Traits
 //!
-//! Notes:
+//! Currency related traits in FRAME provide standardized interfaces for implementing various
+//! currency functionalities. These traits enable developers to create, transfer, and manage
+//! different forms of digital assets within the blockchain environment, ensuring that the economic
+//! aspects of the chain are robust and adaptable to different use cases.
```
kianenigma, line 6: It is worth noting that the fundamental reason to have these traits is to allow pallets to utilize a token without relying on the implementation detail. Or else I could have a perfectly fine runtime without these traits as well.

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
...
+//! for managing currency types, providing a streamlined approach for single-asset operations.
+//! Fungible is currently preferred over currency as the latter is deprecated.
+//!
+//! #### Holds and Freezes
+//!
+//! Learn more about this two concepts in
+//! [`frame_support::traits::tokens::fungible::hold`][1] and
+//! [`frame_support::traits::tokens::fungible::freeze`][2].
+//!
+//! ## Pallet Balances
```
kianenigma, line 40:
```suggestion
//! ## Balances Pallet
```

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
...
+//! The [`pallet_balances`](../../../pallet_balances/index.html) is a key component in FRAME. It
+//! is designed for managing native token balances. It plays a crucial role in tracking and
+//! manipulating the balance of accounts, providing functionalities essential for a wide range of
+//! financial operations. The key functions of
+//! [`pallet_balances`](../../../pallet_balances/index.html) include transferring tokens between
+//! accounts, checking account balances, and adjusting balances for staking or fees. This pallet
+//! implements the [`fungible`](../../../frame_support/traits/tokens/fungible/index.html)
```
kianenigma, line 47: I think the most important detail came last; What I think is the most important piece of info here is that `pallet-balances` and `pallet-assets` provide implementations for all of the aforementioned traits. We can list those here (while linking to the proper `impl` block in the rust-docs).

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,51 @@
...
+//! on the [Tokens Horizon issue](https://github.com/paritytech/polkadot-sdk/issues/327). The
+//! [`Currency`](../../../frame_support/traits/tokens/currency/index.html) trait is still available
+//! in Substrate, but it is recommended to use the **fungible** traits instead. The
+//! [deprecation PR](https://github.com/paritytech/substrate/pull/12951) has a dedicated section on
+//! upgrading from **Currency** to **fungible**. Besides, this [issue](https://github.com/paritytech/polkadot-sdk/issues/226)
+//! lists the pallets that have been upgraded to the **fungible** traits, and the ones that are
+//! still using the [`Currency`](../../../frame_support/traits/tokens/currency/index.html) trait.
+//! There one could find the relevant code examples for upgrading.
+//!
+//! ## Fungible Traits
```
kianenigma, line 28: You are not going to briefly talk about `Fungibles`?

substrate/frame/support/src/traits/tokens/fungible/freeze.rs
```diff
@@ -15,7 +15,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! # Freeze
```
kianenigma, line 18: This is indeed the right place to document this. Any methods that can also be improved?

substrate/frame/support/src/traits/tokens/fungible/freeze.rs
```diff
@@ -15,7 +15,26 @@
...
+//! this frozen balance from being reduced if the overlapping hold is slashed.
+//!
+//! ```
+//! |__total__________________________________|
+//! |__on_hold__|_____________free____________|
+//! |__________frozen___________|
+//! |__on_hold__|__ed__|  
+//! ```
+//!
+//! Freezes require a `Reason`, which is configurable and is expected to be an enum aggregated
```
kianenigma, line 36: I don't anyone who doesn't know `RuntimeFreezeReason` can understand this. Don't have a better option in mind now, but maybe we can do better.

kianenigma (approved):
This is a good start, but I expect a lot more substance on this particular topic. I expect this to be in https://github.com/paritytech/polkadot-sdk/pull/2802? if so, we can merge this already.

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: I see `fungibles` making sense in multiple assets scenario but not here, there is this other [PR](https://github.com/paritytech/polkadot-sdk/pull/2802) that covers more generic tokens topics

juangirini:
> This is a good start, but I expect a lot more substance on this particular topic. I expect this to be in #2802? if so, we can merge this already.

Yes, we have split the topic into two sections, this one focused only on "currency" as a native token, and a more generic one covered by that other PR

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
-//! FRAME Currency Abstractions and Traits
+//! # FRAME Currency Abstractions and Traits
 //!
-//! Notes:
+//! Currency related traits in FRAME provide standardized interfaces for implementing various
+//! currency functionalities. The fundamental reason to have these traits is to allow pallets to
+//! utilize a token without relying on the implementation detail. Or else we could have a perfectly
```
gpestana, line 5:
```suggestion
//! utilize tokens without relying on the implementation detail. Or else we could have a perfectly
```

gpestana, line 5: > Or else we could have a perfectly fine runtime without these traits.

I don't quite get what this sentence means to say. Could you elaborate/rephrase?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
...
+//! fine runtime without these traits.
+//! These traits enable developers to create, transfer, and manage different forms of digital assets
```
gpestana, line 7: > different forms of digital assets

Nit but the first thing I thought when I read this was NFTs and I believe the `Currency` traits are not the best abstractions.

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
...
+//! and adaptable to different use cases.
+
 //!
-//! - History, `Currency` trait.
-//! - `Hold` and `Freeze` with diagram.
-//! - `HoldReason` and `FreezeReason`
-//! 	- This footgun: <https://github.com/paritytech/polkadot-sdk/pull/1900#discussion_r1363783609>
+//! ## The Currency Trait
+//!
+//! The [`Currency`][4] trait was initially introduced in Substrate to manage the native token
```
gpestana, line 14: > [`Currency`][4]

Nit but this is the first time I'm seeing this type of referencing in our docs, maybe we should just add the URL in line? I find it easier to just click on the word/expression directly if I want more details.

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
...
+//! upgraded to the **fungible** traits, and the ones that are still using the [`Currency`][4]
+//! trait. There one could find the relevant code examples for upgrading.
+//!
+//! ## Fungible Traits
+//!
+//! The [`fungible`][3] traits are designed for managing currency types, providing a streamlined
+//! approach for single-asset operations. Fungible is currently preferred over currency as the
+//! latter is deprecated.
+//!
+//! #### Holds and Freezes
```
gpestana, line 36:
```suggestion
//! ### Holds and Freezes
```

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
...
+//!
+//! Learn more about this two concepts in
+//! [`frame_support::traits::tokens::fungible::hold`][1] and
+//! [`frame_support::traits::tokens::fungible::freeze`][2].
```
gpestana, line 40:
```suggestion
//! [`frame_support::traits::tokens::fungible::freeze`].
```

I believe the link is added automatically in this case?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
...
+//!
+//! ## Balances Pallet
+//! The [`pallet_balances`](../../../pallet_balances/index.html) is a key component in FRAME. It
```
gpestana, line 43: This linking looks better to me, maybe we can stick with that and remove the references in the other cases?

docs/sdk/src/reference_docs/frame_currency.rs
```diff
@@ -1,8 +1,55 @@
-//! FRAME Currency Abstractions and Traits
+//! # FRAME Currency Abstractions and Traits
```
gpestana, line 1: I think it would be useful to have a table with the comparing the Currency vs Fungible/s terms and a description on how they differ (e.g. locks vs holds, etc). This is usually a place where people get confused and it would be great if we could have a single place to refresh the memory.

kianenigma:
@juangirini I wonder if you intend to finish the last bits of this, as there seems to be little left? else we can consider mentoring someone else to finish it. 

In case you do, feel free to leave your DOT address in the PR description :)

juangirini:
> @juangirini I wonder if you intend to finish the last bits of this, as there seems to be little left? else we can consider mentoring someone else to finish it.
> 
> In case you do, feel free to leave your DOT address in the PR description :)

@kianenigma, yes I'll finish the PR, there is very little left to do. Thanks for the reminder

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: It should be removed as it is redundant, what I meant is that you don't need these traits if your runtime doesn't require a currency

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: Yes, it's a tradeoff... this format helps readability when there are long / too many links
https://github.com/paritytech/polkadot-sdk/pull/2683#discussion_r1444902876

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: Not really as `frame_support` is not a dependency of `polkadot-sdk-docs`

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: I have removed them where they made the lines to long or difficult to read

docs/sdk/src/reference_docs/frame_currency.rs
juangirini: I agree we can improve this document by adding some more information, maybe we can keep that for another PR, wdyt?

docs/sdk/src/reference_docs/frame_currency.rs
liamaharon: Should be able to add it as a dep?

kianenigma:
@liamaharon is looking into how this PR can be incorporated into his more recent and broader PR around tokens, but nonetheless submitting a tip for gratitude.

liamaharon:
Closing in favour of https://github.com/paritytech/polkadot-sdk/pull/2802

Issue #2627, opened 2023-12-05
Make Beefy work with GRANDPA warp sync


serban300:
We should make it possible to start a node with beefy enabled when using GRANDPA warp sync

bkchr:
We currently only use `Grandpa` as provider for the warp sync proofs. For Beefy we probably need to extend the proofs to also include proofs from Beefy. Or better, only use beefy for proofing the warp from state X to state Y.

bkchr:
And isn't this a duplicate of https://github.com/paritytech/polkadot-sdk/issues/1118?

serban300:
Sorry, I should have added more details in the description. 

Right now if the node is started with `--sync warp` we just disable the BEEFY client. The requirement here is to be able to start a node with BEEFY client + grandpa warp sync. This would mean to wait for GRANDPA warp sync to finish and start the BEEFY client after all the headers have been imported. 

It would be a first step. After that we can implement actual BEEFY warp sync (#1118).

bkchr:
> This would mean to wait for GRANDPA warp sync to finish and start the BEEFY client after all the headers have been imported

After all headers from genesis? Why do we need this?

serban300:
Because BEEFY tries to walk back from the best finalized header down to BEEFY genesis and then sync all the BEEFY justifications in this range. If headers are missing, it fails. We need to fix this somehow.

acatangiu:
> After all headers from genesis? Why do we need this?

To be exact, to initialize BEEFY worker it _currently_ needs headers in backend in range `[max(beefy_genesis_block, best_beefy_finalized_in_aux_db), best_grandpa_finalized]`.

When WarpSyncing there is a "jump" between last sync and current head - GRANDPA gets its mandatory blocks justifications through the `WarpSyncProof` - BEEFY will do the same with #1118 - but before that, we can do a quick fix where BEEFY can walk back the chain over the "jumped" blocks, look at header digests to find out which are mandatory blocks then on-demand request justifications for those.

This ^ should be fairly easy to do and will allow us to remove the "disallow `--sync warp` when BEEFY" limitation.

bkchr:
The problem is that gap sync (sync responsible for syncing missing headers/blocks between the warp sync target and genesis) starts syncing from genesis AFAIR. I mean we could probably change this, but right now your quick fix would not work.

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

[...]

PR #1329, opened 2023-08-31
BEEFY: add support for slashing validators signing forking commitments (without ancestry proofs)

Lederstrumpf:
# Description
Migrates & builds on https://github.com/paritytech/substrate/pull/14744 - see its description & discussion for history.

For proving a fork equivocation, this PR's equivocation proofs currently includes headers that the verifier calculates hashes of and compares with on-chain `<frame_system::Pallet<T>>::block_hash(block_number)`. Since this mapping has a horizon of [4096 blocks](https://github.com/lederstrumpf/polkadot-sdk/blob/80656b2fe9d11d665b51db167daa6803307a303f/polkadot/runtime/common/src/lib.rs#L84), equivocations predating this horizon cannot be slashed. While this thwarts attacks assuming a 2/3rds of validators honestly participate in BEEFY finalization and at least one honest relayer can update the beefy light client at least once every 4096 blocks, this is clearly only a temporary solution. Should not be merged until 
1. ancestry proofs are integrated (#1441)
2. BEEFY goes live on Kusama (because going live with slashing that assumes 2/3rd of validators are honest is better than going live with no fork equivocation slashing at all)

# Proposed Solution
## Runtime
- two new extrinsic submission calls in `pallet_beefy`: `submit_unsigned_vote_equivocation_report` (previously `report_equivocation`) for vote equivocations (`VoteEquivocationProof`), and `submit_unsigned_fork_equivocation_report` for fork equivocations (`ForkEquivocationProof`), the latter whether detected in votes, `SignedCommitment`, or `VersionedFinalityProof`
- verify commitment's `mmr_root` != on-chain `mmr_root`
report offense to staking so offending vote author / `SignedCommitment` signatories get slashed
## Client-side
adds "fisherman" capabilities to voter gossip - on detecting votes for historical forks, it builds the required proof of misbehavior and submits report.

Closes https://github.com/paritytech/polkadot-sdk/issues/1120

PR #2689, opened 2023-12-12
BEEFY: Support compatibility with Warp Sync - Allow Warp Sync for Validators

serban300:
Resolves #https://github.com/paritytech/polkadot-sdk/issues/2627

Initializes voter _after_ headers sync finishes in the background.

This enables the BEEFY gadget to work with `--sync warp` (GRANDPA warp sync).

Not sure if this can be tested:
- with zombienet I don't think it's feasible. We would need to wait for a very long chain to be built (days)
- still looking for the network sync unit tests but couldn't find anything relevant yet


acatangiu:
> with zombienet I don't think it's feasible. We would need to wait for a very long chain to be built (days)

Why is that? In a zombienet test we control the chainspec and the node parameters so we can run with very short sessions/eras/epochs, very aggressive pruning, etc.
Two nodes running a couple of minutes, then a third node warp syncing to catch up, then checking BEEFY is also working after another minute should do it.

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -292,21 +295,35 @@ pub async fn start_beefy_gadget<B, BE, C, N, P, R, S>(
...
+			},
+		};
+
+		if let Err(e) =
+			wait_for_headers_sync(&*backend, beefy_genesis, best_grandpa.clone(), HEADER_SYNC_DELAY)
+				.await
+		{
+			error!(target: LOG_TARGET, "Error: {:?}. Terminating.", e);
+			return
+		}
```
acatangiu, line 318: This is expensive (especially when the chain is long), so we should do it _only_ when necessary. Most of the time BEEFY gadget is initialized from aux db from some existing state, in which case it had already handled `[beefy_genesis, best_beefy]` sub-chain.

You should move this logic after loading persisted state, but before finishing worker initialization and set the start of the range to `max(beefy_genesis, best_beefy)`.

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -388,6 +405,53 @@ where
...
+async fn wait_for_parent_header<B, BE>(
+	backend: &BE,
+	current: <B as Block>::Header,
+	delay: Duration,
+) -> ClientResult<<B as Block>::Header>
+where
+	B: Block,
+	BE: Backend<B>,
+{
+	let blockchain = backend.blockchain();
```
acatangiu, line 417: nit: I think you could directly pass `blockchain` reference
```suggestion
async fn wait_for_parent_header<B, BE>(
	blockchain: &BE::Blockchain,
	current: <B as Block>::Header,
	delay: Duration,
) -> ClientResult<<B as Block>::Header>
where
	B: Block,
	BE: Backend<B>,
{
```

acatangiu (changes requested):

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -78,6 +79,8 @@ mod tests;
 
 const LOG_TARGET: &str = "beefy";
 
+const HEADER_SYNC_DELAY: Duration = Duration::from_secs(30);
```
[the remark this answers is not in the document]
serban300: Done

substrate/client/consensus/beefy/src/lib.rs
serban300: Good point. Changed the strategy to calling `wait_for_parent_header()` directly from `initialize_voter_state()`

substrate/client/consensus/beefy/src/lib.rs
serban300: Done

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -388,6 +405,53 @@ where
...
+{
+	let blockchain = backend.blockchain();
+	loop {
+		match blockchain.header(*current.parent_hash())? {
+			Some(parent) => return Ok(parent),
+			None => {
+				info!(
+					target: LOG_TARGET,
+					"🥩 Parent of header {} not found. BEEFY gadget waiting for header sync to finish ...",
+					current.number()
```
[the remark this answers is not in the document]
serban300: Done

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -388,6 +405,53 @@ where
...
+				);
+				tokio::time::sleep(delay).await;
+			},
+		}
+	}
+}
+
+async fn wait_for_headers_sync<B, BE>(
```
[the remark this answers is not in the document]
serban300: Removed this function, but added a detailed comment to `wait_for_parent_header()`

serban300:
> > with zombienet I don't think it's feasible. We would need to wait for a very long chain to be built (days)
> 
> Why is that? In a zombienet test we control the chainspec and the node parameters so we can run with very short sessions/eras/epochs, very aggressive pruning, etc. Two nodes running a couple of minutes, then a third node warp syncing to catch up, then checking BEEFY is also working after another minute should do it.

I ran a rococo-local test chain these days for testing these changes and judging by those experiments I think we need at least 10-20k blocks in order to have to wait for the header warp sync to finish. For less than 10-20k blocks, all the headers are synced before the BEEFY gadget starts.

I tried to play a bit with the block times and I couldn't get below 4 seconds without getting errors that go like this:
```
Creating inherent data took more time than we had left for slot 851194616 for block 0x6924fbb0c1641fd0507da0f5185c9964fb89c9cd57d970030b4c1d680d70740f
```
I will experiment a bit more with this, but even if we could get to 500 milliseconds per block, we would get to 20k blocks in 2-3 hours. And I don't know if it's feasible to get under 500 ms per block.

acatangiu:
> For less than 10-20k blocks, all the headers are synced before the BEEFY gadget starts.

I see..

Can we start a zombienet test from some predefined state/db? Instead of starting from zero and building from genesis? That way we could have a node starting from db with a 50k blocks chain, and another from zero and warpsync to 50k.

serban300:
> > For less than 10-20k blocks, all the headers are synced before the BEEFY gadget starts.
> 
> I see..
> 
> Can we start a zombienet test from some predefined state/db? Instead of starting from zero and building from genesis? That way we could have a node starting from db with a 50k blocks chain, and another from zero and warpsync to 50k.

I see that there are zombienet tests for warp sync that start from db snapshots. Probably it would work for our use case as well. Will try.

acatangiu:
> > > For less than 10-20k blocks, all the headers are synced before the BEEFY gadget starts.
> > 
> > 
> > I see..
> > Can we start a zombienet test from some predefined state/db? Instead of starting from zero and building from genesis? That way we could have a node starting from db with a 50k blocks chain, and another from zero and warpsync to 50k.
> 
> I see that there are zombienet tests for warp sync that start from db snapshots. Probably it would work for our use case as well. Will try.

Nice! We should reuse same test then - just add BEEFY verifications to it 👍

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -381,18 +392,51 @@ where
 			state.set_min_block_delta(min_block_delta);
 			info!(target: LOG_TARGET, "🥩 Loading BEEFY voter state from db: {:?}.", state);
 			Some(Ok(state))
-		})
-		// No valid voter-state persisted, re-initialize from pallet genesis.
-		.unwrap_or_else(|| {
-			initialize_voter_state(backend, runtime, beefy_genesis, best_grandpa, min_block_delta)
-		})
+		});
+	if let Some(persisted_state) = maybe_persisted_state {
```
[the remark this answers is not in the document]
serban300: Done

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -381,18 +392,51 @@ where
 			state.set_min_block_delta(min_block_delta);
 			info!(target: LOG_TARGET, "🥩 Loading BEEFY voter state from db: {:?}.", state);
 			Some(Ok(state))
```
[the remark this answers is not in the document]
serban300: Done

acatangiu (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4784164

acatangiu:
I tested this by warp syncing a local validator node:
- warp sync Rococo to catch up after node was offline for a couple of hours (ran same test a few times incrementally)
- warp sync Westend from scratch (empty db)

In both cases, BEEFY initialized correctly, node finalized all BEEFY mandatory blocks from warp sync window, then continued with participation in up-to-date consensus.

@paritytech/bridges-core please review - I want to merge this within current release window as it greatly helps with BEEFY "adoption" for Kusama and Polkadot validators.

I've opened https://github.com/paritytech/polkadot-sdk/issues/2787 for adding/enhancing zombienet tests to cover it - we can do that after the holidays.

substrate/client/consensus/beefy/src/lib.rs
```diff
@@ -532,7 +586,7 @@ where
 	Err(ClientError::Backend(err_msg))
 }
 
-fn expect_validator_set<B, BE, R>(
+async fn expect_validator_set<B, BE, R>(
```
svyatonik, line 589: I've been suggesting to use this function from the https://github.com/paritytech/polkadot-sdk/pull/1329, but looks like it isn't safe now, because it may block callers and if used on a critical path (e.g. during block import), we may halt the regular client operations. I suggest adding a comment to that function, mentioning that it uses `sleep` (asybc) inside, so shall be used with care

svyatonik (approved):

substrate/client/consensus/beefy/src/lib.rs
acatangiu: done

PR #2695, opened 2023-12-12
Add dummy impls of `fungible` and `fungibles` trait families

franciscoaguirre:
In `Currency`, we have a dummy impl that we can use for mocks or examples where we only want to satisfy the trait bounds. I added the same dummy implementations to `fungible` and `fungibles` regular traits.


franciscoaguirre:
bot fmt

seadanda (approved):

gupnik (approved):

PR #2700, opened 2023-12-13
Contracts: Fix legacy provider count

pgherveou:
With the breaking currency API update (https://github.com/paritytech/substrate/pull/12951) you now always need to hold at least the ed as free balance. That was not the case before. Contracts accounts for example would typically have some amount reserved as deposit and hold 0 balance.

The logic in `try_mutate_account` below (more specifically the meaning of `did_provide`) is only valid "post-update" and without this update, bumps the provider count (in our case from 1 to 2) when we unreserve the deposit.

https://github.com/paritytech/polkadot-sdk/blob/10451b921f4ddd82ee15b3641eb3aa49d8761008/substrate/frame/balances/src/lib.rs#L931-L933

This PR updates the `ensure_upgraded` logic to make sure this is fixed when we migrate the accounts to the new logic.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4711218

pgherveou:
bot fmt

substrate/frame/balances/src/tests/currency_tests.rs
```diff
@@ -513,6 +513,26 @@ fn unreserving_balance_should_work() {
...
+		assert_eq!(System::<Test>::account(&1).providers, 1);
+
+		// Set up a legacy account where free = 0 reserved = 100.
+		EXISTENTIAL_DEPOSIT.with(|v| *v.borrow_mut() = 0);
+		assert_ok!(Balances::reserve(&1, 100));
+		EXISTENTIAL_DEPOSIT.with(|v| *v.borrow_mut() = 1);
+		assert_eq!(System::<Test>::account(&1).providers, 1);
+
+		Balances::unreserve(&1, 100);
+		assert_eq!(System::<Test>::account(&1).providers, 1); // <- this fails
```
pgherveou, line 530:
```suggestion
		assert_eq!(System::<Test>::account(&1).providers, 1);
```

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4713904

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4728047

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4728048

substrate/frame/balances/src/lib.rs
```diff
@@ -911,8 +911,8 @@ pub mod pallet {
 			let result = T::AccountStore::try_mutate_exists(who, |maybe_account| {
 				let is_new = maybe_account.is_none();
 				let mut account = maybe_account.take().unwrap_or_default();
-				let did_provide =
-					account.free >= Self::ed() && Self::have_providers_or_no_zero_ed(who);
+				let did_provide = (account.reserved + account.free) >= Self::ed() &&
```
liamaharon, line 914: Just to confirm: so the `ed` can be part of `reserved`?

pgherveou: With the breaking currency api update you now always need to hold at least the ed as free balance. 
That was not the case before. Contracts accounts for example would typically have some amount reserved as deposit and hold 0 balance. 

The code touched here does not take this scenario into account, and migrating these old accounts will wrongly bump the providers count 1. 

The changes here attempt to fix it. I need to look at these failing ci jobs though

liamaharon: I see, thanks. Let's also add a comment to explain.

pgherveou, line 914:
```suggestion
				let did_provide = account.free >= Self::ed() &&
```

Rolling back this is the wrong fix.
I believe, that we need to fix that in the ensure_upgraded function.

If an account that is being upgraded does not provide for itself (free < ed), we should move ed from reserved to the free balance when we migrate the account, or decrement it's provider count, so that the logic here does not offset the provider count by 1.

substrate/frame/balances/src/lib.rs
```diff
@@ -911,8 +911,8 @@ pub mod pallet {
...
+					Self::have_providers_or_no_zero_ed(who);
```
pgherveou, line 915:
```suggestion
				let did_provide =
					account.free >= Self::ed() && Self::have_providers_or_no_zero_ed(who);

```

substrate/frame/balances/src/lib.rs
```diff
@@ -913,6 +913,7 @@ pub mod pallet {
 				let mut account = maybe_account.take().unwrap_or_default();
 				let did_provide =
 					account.free >= Self::ed() && Self::have_providers_or_no_zero_ed(who);
+
```
pgherveou, line 916:
```suggestion
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4749914

pgherveou:
@liamaharon @gavofyork you mind taking a look at this?
I updated the PR description and the fix since last time.

substrate/frame/balances/src/lib.rs
```diff
@@ -773,7 +773,14 @@ pub mod pallet {
 				}
 				let _ = system::Pallet::<T>::inc_consumers_without_limit(who).defensive();
 			}
-			// Should never fail - we're only setting a bit.
+
+			// Ensure the account holds at least ED of free balance.
+			if a.free < Self::ed() {
+				let shortfall = Self::ed() - a.free;
+				a.free = Self::ed();
+				a.reserved = a.reserved.defensive_saturating_sub(shortfall);
```
liamaharon, line 781: This function is already deployed, so the fix won't 'activate' for accounts who already are on the new logic. 

When we enter this code path, we're guaranteed reserved will be >= shortfall right? I'm wary of accidentally creating tokens out of thin air and messing up total issuance values. Even if it's fine for us, maybe it won't be for parachains who could have adjusted EDs.

pgherveou: The previous approach was wrong, It was breaking a few tests.
Once an account is upgraded you don't run into this bug, so I think the logic really needs to live here.

You only run into this bug for legacy account, that match the scenario described (pre-upgrade free-balance < e.d).
The bug here is that you end up with a provider off by 1 when this code run post-upgrade, so I think the impacted accounts need to be identified and a migration script should be written for those.

pgherveou:
👋  @paritytech/frame-coders can someone review this ?

substrate/frame/balances/src/lib.rs
ggwpez: You mean we need another migration to fix the first one?

pgherveou: Liam was asking about a previous commit in this PR that was trying to fix this differently.
To be clear, accounts that have been updated with this lazy migration logic and fall into this category will end up with a provider count off by one. To make sure they get reaped when their funds get below ED,  I believe we will need a sudo transaction to fix up the provider count.

liamaharon: So it seems this pr will not completely resolve the provider count issue in its existing state because it does not do anything to existing accounts?

I'm probably missing context, what benefit do we achieve by only partially fixing this?

pgherveou: This pr fix the issue, that's important for chain that haven't migrated yet, and since it's a lazy migration this also fix the issue for accounts that have not been touched yet on chain that have already migrated. 

Whether it's worth making another migration for accounts already impacted, can be decided and done separately.

liamaharon: What is the reason it's more important to fix for chains that haven't migrated than chains that have already migrated? 

I'm fine to implement this fix, but am concerned not to just 'forget' fixing this for chains which have already migrated. This would be a good time to consider how to do that, otherwise we'll likely forget about it. 

Also is this still primarily about fixing the provider count or is it now primarily about fixing the ed, which has the side effect of the provider count not messing up later?

pgherveou: It's just fixing the provider count. This is is more urgent to stop an ongoing issue rather than fixing accounts that can only be done with some kind of root privileges. If the numbers of accounts impacted is not that big it's not adding much bloat to the chain's db size and it might be fine in some cases to do just let it be

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4939299

substrate/frame/balances/src/lib.rs
liamaharon: Did we consider fixing how `did_provide` is derived instead of messing with account balances? Perhaps that would be a more direct approach, if this pr really is about fixing the provider count and account eds is just a way to do that.

pgherveou: Yes I [tried](https://github.com/paritytech/polkadot-sdk/pull/2700/commits/2f847113a22bd08bca53179fdbe296afc7041fb0), but it was not right. You don't want to add  migration logic outside of the ensure_upgrade migration function, or this is going to be a mess.

ggwpez: FYI Kusama, Polkadot, Rococo got 100% of their accounts upgraded by a script, but sure its important for parachains.

substrate/frame/balances/src/lib.rs
```diff
@@ -768,12 +769,19 @@ pub mod pallet {
 						"account with a non-zero reserve balance has no provider refs, account_id: '{:?}'.",
 						who
 					);
-					a.free = a.free.max(Self::ed());
+					a.free = a.free.max(ed);
```
ggwpez, line 772: Btw would this not mess up the `TotalIssuance`?

substrate/frame/balances/src/lib.rs
```diff
@@ -768,12 +769,19 @@ pub mod pallet {
...
 				}
 				let _ = system::Pallet::<T>::inc_consumers_without_limit(who).defensive();
 			}
-			// Should never fail - we're only setting a bit.
+
+			// Ensure the account holds at least ED of free balance.
+			let free = a.free;
+			if free < ed && a.reserved >= ed - free {
+				a.free = ed;
+				a.reserved -= ed - free;
```
ggwpez, line 782: We cannot just change the reserve of an account, reserved balances are GUARANTEED to be slashable.  
To me it looks like this should reap accounts that are below ED. Contract accounts would then need to be air-dropped... which is a pain.  
Alternative is to just "give" them `ed` for free. But Total issuance needs to be updated AFAIK.

substrate/frame/balances/src/lib.rs
liamaharon: > Yes I [tried](https://github.com/paritytech/polkadot-sdk/pull/2700/commits/2f847113a22bd08bca53179fdbe296afc7041fb0), but it was not right. You don't want to add migration logic outside of the ensure_upgrade migration function, or this is going to be a mess.

How is this more hacky than the currently proposed approach?

substrate/frame/balances/src/lib.rs
liamaharon: I don't get why not to fix the `did_provide` logic (like https://github.com/paritytech/polkadot-sdk/pull/2700/commits/2f847113a22bd08bca53179fdbe296afc7041fb0) rather than faffing about changing balances which comes with a whole new set of considerations and risks?

substrate/frame/balances/src/lib.rs
pgherveou: It should not as it is now it just move funds from reserved to free

substrate/frame/balances/src/lib.rs
pgherveou: This is not correct the invariant on the new logic is that `account.free >= Self::ed()`
This previous fix was breaking some other tests. I can lookup which one later on to let you know.

substrate/frame/balances/src/lib.rs
pgherveou: > FYI Kusama, Polkadot, Rococo got 100% of their accounts upgraded by a script, but sure its important for parachains.

when you say script you mean some code added to the runtime that runs on_idle ?

substrate/frame/balances/src/lib.rs
pgherveou: I guess what we could try is to lower the provider_count by one knowing that it will be bumped by 1 later on, not sure if that's a safe approach or not

substrate/frame/balances/src/lib.rs
ggwpez: No, i mean a python script 😆 You can see the 1.1M events here: https://polkadot.subscan.io/event?page=1&time_dimension=date&module=balances&event_id=upgraded

PR #2706, opened 2023-12-14
fix overflow in `balance_to_point` conversion

### LINKED ISSUE #416: Nomination pools: points to balance and balance to points rounding inaccuracy
It has been [reported](https://hackmd.io/OHDPAxEHQq6O6EP_9zzCag) that a conversion of `points_to_balance` in nominations pools returned less balance than expected (in Westend). In summary, an account joined a pool with 4 WND and -- even though there was no unbonding or slashes in the pool -- the result of calling the `points_to_balance` with the received points when joining is 399.999 mWND.

The runtime API is using the same logic for calculating the `points_to_balance` and `balance_to_points` as used when accounts join/leave pools.

SCJangra:
Closes #416

As I mentioned in the above issue `balance_to_points` conversion currently overflows the `Balance` types even before computing the actual points for the given tokens. This fixes that,

### DIFF
--- a/substrate/frame/nomination-pools/src/lib.rs
+++ b/substrate/frame/nomination-pools/src/lib.rs
@@ -2955,9 +2955,12 @@ impl<T: Config> Pallet<T> {
 			},
 			(false, false) => {
 				// Equivalent to (current_points / current_balance) * new_funds
-				balance(u256(current_points).saturating_mul(u256(new_funds)))
-					// We check for zero above
-					.div(current_balance)
+				balance(
+					u256(current_points)
+						.saturating_mul(u256(new_funds))
+						// We check for zero above
+						.div(u256(current_balance)),
+				)
 			},
 		}
 	}
--- a/substrate/frame/nomination-pools/src/tests.rs
+++ b/substrate/frame/nomination-pools/src/tests.rs
@@ -292,6 +292,28 @@ mod bonded_pool {
 			assert_ok!(pool.ok_to_join());
 		});
 	}
+
+	#[test]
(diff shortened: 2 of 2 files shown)


gpestana (changes requested):
Thanks for the PR, it looks like a fix for this issue. One improvement could be adding some tests that ensure that the conversion is working as expected and the bounds are safe and that would fail with the code before the PR, wdyt?

As a suggestion, something along these lines:

```rust
        #[test]
	fn points_and_balance_conversions_are_safe() {
		ExtBuilder::default().build_and_execute(|| {
			let bonded_pool = BondedPool::<Runtime> {
				id: 123123,
				inner: BondedPoolInner {
					commission: Commission::default(),
					member_counter: 1,
					points: u128::MAX,
					roles: DEFAULT_ROLES,
					state: PoolState::Open,
				},
			};
			StakingMock::set_bonded_balance(bonded_pool.bonded_account(), u128::MAX);

			// max out the points and balance of the pool and make sure the conversion works as
			// expected ans does not overflow.
			assert_eq!(bonded_pool.balance_to_point(u128::MAX), u128::MAX);
			assert_eq!(bonded_pool.points_to_balance(u128::MAX), u128::MAX);
		})
	}
```

SCJangra:
@gpestana I've shamelessly copied the code you posted into tests, I can't think of any more tests specifically for this issue. Let me know if you need anything else changed/added.

ggwpez (approved):

gpestana (approved):
Great, thanks!

bkchr (approved):

Issue #523, opened 2023-01-27
Announce "capabilities" over the network?


bkchr:
Currently we expect that we can sync from all full nodes all blocks or request warp proofs from all full nodes. When people start using things like `--block-pruning` these assumptions are not correct anymore. As there are currently not that much nodes running with these settings, retrying the download on from a different node probably succeeds. However, we can not expect this to happen in the future and may in general switch to "prune everything" on default. Stuff like all the block/bodies from each block isn't that useful too keep. 

So, we may should change the sync handshake to contain information until which block a node has data. Warp syncing probably changes any way when we have nodes (like light clients) that only need to sync from a certain snapshot. All data before these snapshots is not important anymore.

tomaka:
I think that one good way to do this is to reserve some kind of prefix or key in the DHT where all the nodes are archives nodes, in a similar way to https://github.com/polkadot-fellows/RFCs/pull/8.

If we allow ourselves to modify the libp2p Kademlia protocol, then we can for example switch to 64 bytes DHT keys, where the first 32 bytes are a prefix that indicates a capability, and the last 32 bytes are the actual PeerId.

If we don't allow ourselves to modify the libp2p Kademlia protocol, then we can use the existing "providers" system, where the PeerIds close to a certain key representing a capability have the responsibility to store a list of PeerIds that implement this capability.
The disadvantage of that second method is that only 20 peers can be stored at a time (20 is a network-wide constant that we could tweak, but it has to be a constant).

This approach has an issue: it can easily be eclipse attacked by malicious nodes hijacking the list. The reason why the normal DHT can't be eclipse attacked is purely because we put a few honest bootnodes in the chain spec that guarantee that all honest nodes can be reached in one way or another. We don't have that luxury here (unless we add additional stuff in the chain spec, which is not a great solution).

In order to bypass this eclipse attack, we can rotate the "capability prefix" using the on-chain Babe/Sassafras randomness. In other words, this would rotate in a random way the list of PeerIds that are responsible for storing the list of nodes that have a certain capability. This doesn't completely solve the problem, but it gives a random chance for an eclipse attack to fail every time the session changes, meaning that in finite time we should be able to find an honest node.

Since we would need to know the current on-chain randomness in order to find where nodes with a certain capability are located, this means that all nodes must have the capability of bringing you to the head of the chain.
In other words, all nodes must always have the "has warp sync proof", "has recent blocks headers", and "has recent block storage" capabilities.

[...]

PR #2710, opened 2023-12-14
Allow skipping block history download in warp sync

### LINKED ISSUE #8: warp sync without download histrocial blocks
Any docs about it? Can't find them.

To my understanding, it will download finality proof and put the best block to latest block and download historical blocks.
Is it possible to not download historical blocks? I tried with `--blocks-pruning=128 --state-pruning=128` and that doesn't seem to have any impacts.

Basically I want to minimize the resource requirement of a parachain node and I don't want to download and store all the relaychain data.

tmpolaczyk:
Adds a new sync mode `--warp-no-block-history` that is exactly the same as `--warp`, but doesn't download block history.

Close #8

### DIFF
--- a/cumulus/client/service/src/lib.rs
+++ b/cumulus/client/service/src/lib.rs
@@ -453,16 +453,15 @@ where
 	RCInterface: RelayChainInterface + Clone + 'static,
 	IQ: ImportQueue<Block> + 'static,
 {
-	let warp_sync_params = match parachain_config.network.sync_mode {
-		SyncMode::Warp => {
-			let target_block = warp_sync_get::<Block, RCInterface>(
-				para_id,
-				relay_chain_interface.clone(),
-				spawn_handle.clone(),
-			);
-			Some(WarpSyncParams::WaitForTarget(target_block))
-		},
-		_ => None,
+	let warp_sync_params = if parachain_config.network.sync_mode.is_warp() {
+		let target_block = warp_sync_get::<Block, RCInterface>(
+			para_id,
+			relay_chain_interface.clone(),
+			spawn_handle.clone(),
+		);
+		Some(WarpSyncParams::WaitForTarget(target_block))
+	} else {
+		None
 	};
 
 	let block_announce_validator = match sybil_resistance_level {
--- a/cumulus/client/service/src/lib.rs
+++ b/cumulus/client/service/src/lib.rs
@@ -40,7 +40,7 @@ use sc_consensus::{
 	import_queue::{ImportQueue, ImportQueueService},
 	BlockImport,
 };
-use sc_network::{config::SyncMode, NetworkService};
+use sc_network::NetworkService;
 use sc_network_sync::SyncingService;
 use sc_network_transactions::TransactionsHandlerController;
 use sc_service::{Configuration, NetworkStarter, SpawnTaskHandle, TaskManager, WarpSyncParams};
--- a/substrate/client/cli/src/arg_enums.rs
+++ b/substrate/client/cli/src/arg_enums.rs
@@ -231,8 +231,10 @@ pub enum SyncMode {
 	Fast,
 	/// Download blocks without executing them. Download latest state without proofs.
 	FastUnsafe,
-	/// Prove finality and download the latest state.
+	/// Prove finality and download the latest state. Download block history as well.
 	Warp,
+	/// Prove finality and download the latest state. Does not download block history.
+	WarpNoBlockHistory,
 }
 
 impl Into<sc_network::config::SyncMode> for SyncMode {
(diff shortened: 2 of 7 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4724646

tomaka:
I don't know whether it is already the case as I'm not familiar with the sync code, but we must make sure that, no matter what, the block headers and justifications of session change blocks are downloaded anyway.

Otherwise it won't be possible for other nodes to warp sync from your node, which means that if enough people run with this option we could completely break warp syncing altogether.

dmitry-markin:
> but we must make sure that, no matter what, the block headers and justifications of session change blocks are downloaded anyway.

As far as I can read the code, that's not the case — as a result of warp sync we only import target block body, state, and justifications. So it won't be possible to warp sync from a node that skipped block history phase.

Also, it needs to be carefully evaluated if adding nodes not able to provide syncing from genesis is safe for network. And likely address https://github.com/paritytech/polkadot-sdk/issues/523 first.

So, right now adding an option to turn off gap sync (block history phase) sounds risky.

tmpolaczyk:
Thanks for your comments. Is the justifications problem something that affects only the relaychain, or does it affects parachains as well? Because I only need this feature for syncing parachains, so I could modify the pr to error if trying to sync a relaychain.

Otherwise is there any way to make this a hidden sync mode or something? Because for my use case it doesn't matter if other nodes are not able to sync from this one.

tomaka:
> Also, it needs to be carefully evaluated if adding nodes not able to provide syncing from genesis is safe for network. And likely address https://github.com/paritytech/polkadot-sdk/issues/523 first.

I've recently noticed that more and more nodes refuse to serve warp sync requests. Currently, around 20% of warp sync requests fail on Westend, and after a small investigation it seems that it's because they target nodes that have blocks pruning enable.

While it's not alarming yet, it could become alarming if people enable the mode that this PR adds.

tomaka:
> Because for my use case it doesn't matter if other nodes are not able to sync from this one.

The problem that we're talking about is a tragedy of the commons type of problem.
If you alone enable this mode then there's no problem, but the more people enable it the worse the syncing will become for everyone.

girazoki:
can we make it under a feature flag? Our usecase is rotating-collators that connect to a full-node that has the history of a chain. The collators are not intended to store the history.

Now using full-sync but then enabling block-prunning has the same implications as this PR right?

tomaka:
An alternative way is that nodes that have this mode enabled do not participate in the DHT at all.

Light clients currently do not participate in the DHT at all because they would otherwise be polluting the network similar to how nodes that have this mode enabled would be polluting the network.

> Now using full-sync but then enabling block-prunning has the same implications as this PR right?

Yes, but that doesn't make it correct. See https://github.com/paritytech/polkadot-sdk/issues/2733.

bkchr (changes requested):
This will first require: https://github.com/paritytech/polkadot-sdk/issues/2738 and https://github.com/paritytech/polkadot-sdk/issues/2733 

I would also not add a new warp sync mode and instead this should be decided based on the blocks-pruning/state pruning flags.

tmpolaczyk:
@bkchr regarding block prunning, I was thinking the same. I just tested that running `--sync warp --blocks-pruning=100` does still download the full block history. So in that case we want to only download `HEADER | BODY | JUSTIFICATION` of the latest 100 blocks, and `HEADER | JUSTIFICATION` of the full block history. Is that correct?

bkchr:
> So in that case we want to only download `HEADER | BODY | JUSTIFICATION` of the latest 100 blocks

Not sure I would care that much about this part, but it would be for sure the cleanest implementation. We will need some kind of implementation that handles the "dynamic nature", as the node will continue to import and finalize new blocks. This means, the "ultimate" last block to download moves all the time. However, it should not be that complicated to implement and also downloading one more block is probably fine.

> and `HEADER | JUSTIFICATION` of the full block history.

The beauty of https://github.com/paritytech/polkadot-sdk/issues/2738 would be that we do not need to download them, as we already downloaded them as part of the warp proof. When we have support for syncing from a checkpoint, we will not have all the justifications back from the genesis anymore. We could then discuss to download them or we would expect that all nodes always use some recent checkpoint.

dmitry-markin:
To give a heads up — as the syncing is currently being heavily refactored in preparation for Sync 2.0, it's better to wait at least for https://github.com/paritytech/polkadot-sdk/pull/2467 and follow-up warp sync strategy PRs to be merged before implementing improvements to warp sync. Otherwise we'll end up with a lot of conflicts.

tmpolaczyk:
Ok, I will close this PR given that it needs to be rewritten anyway. If after sync 2.0 this functionality is still missing I will open a new PR. Thank you all for your input!

PR #2711, opened 2023-12-14
pallet-nomination-pools: Enable function for `fuzzing` feature as well

bkchr:
The function is called by `do_try_state` which is also enabled for the `fuzzing` feature.


bkchr:
bot fmt

ggwpez (approved):

Issue #150, opened 2023-07-24
Adjust `GenesisBuilder` naming


michalkucharczyk:
`GenesisBuilder` trait needs some naming adjustments.

              LGTM other than a grumble about naming which I hope to see fixed before this is stabilized.

_Originally posted by @kianenigma in https://github.com/paritytech/substrate/pull/14310#pullrequestreview-1541606306_

Refer also to: https://github.com/paritytech/substrate/pull/14131#pullrequestreview-1435453355

part of: https://github.com/paritytech/polkadot-sdk/issues/25

kianenigma:
To re-iterate, I have two points: 

1. We are using the word "config", where it is meant to just represent "state". For example `fn build_config(config: Vec<u8>, param: Vec<u8>) -> sp_genesis_builder::Result` which is now part of all runtimes is doing _nothing_ other than building the initial "state". Why are we calling this "config"? there is no "configuration" involved. 
2. And, the former point is important, because is FRAME `Config` means something specific, and I worried that people would mix the two.

JoshOrndorff:
This is relevant to what we are doing in https://github.com/Off-Narrative-Labs/Tuxedo/pull/127 (cc @muraca) and my question at https://substrate.stackexchange.com/questions/10105/extrinsics-in-genesis-block

IDK if it is in scope for this issue, but we are wanting to allow chain launchers to specify genesis transactions (yes transactions, not just state) in their chain specs. But have so far relied on a storage hack.

michalkucharczyk:
This issue is only about re-naming the components that are already in place (e.g : [`GenesisBuilder API`](https://github.com/paritytech/polkadot-sdk/blob/3e980219862d58aa52fd0b4dfff907d69ce92790/substrate/primitives/genesis-builder/src/lib.rs#L38), which were implemented to add support for `GenesisConfig` in wasm runtimes. (refer to: https://github.com/paritytech/polkadot-sdk/issues/25). 

Adding support for genesis extrinsic is definitely out of scope of this ticket.

bkchr:
> IDK if it is in scope for this issue, but we are wanting to allow chain launchers to specify genesis transactions (yes transactions, not just state) in their chain specs. But have so far relied on a storage hack.

This is already supported, as I have answered you on SE.

michalkucharczyk:
New naming proposed in https://github.com/paritytech/polkadot-sdk/pull/2714 (cc: @kianenigma)

[...]

Issue #1984, opened 2023-10-23
Decouple node chain-specs from runtime types


michalkucharczyk:
Removing references to runtime's `RuntimeGenesisConfig` struct on the node side is not enough to decouple the node/client from the native-runtime entirely.

There are still some runtime-related left-overs required on the node side to build the chain-spec. Examples of the types (or consts) that one needs to know are as follows ([example](https://github.com/paritytech/polkadot-sdk/blob/7241a8db7b3496816503c6058dae67f66c666b00/polkadot/node/service/src/chain_spec.rs#L808-L830)):
- `AccountId`,
- `HostConfiguration`,
- `SessionKeys` (BabeId, GrandpaId etc)
-  `westend::BABE_GENESIS_EPOCH_CONFIG`

This applies to every complex type defined in runtime. 

These types need to be known at time of creating the json chain-spec. One needs to know what the types are to properly put them into the json format of runtime genesis config. Assuming that we don't want to craft json files manually, this requires building the helper tool for each runtime (as it was done for fellowship runtimes [here](https://github.com/polkadot-fellows/runtimes/pull/78)) or incorporating the runtime deps into client code.

To overcome this problem, I propose to move the creation of the JSON representation of the runtime genesis config into the runtime itself, and exposing multiple configurations through the `GenesisBuilder` API.
`GenesisBuilder` implementation may encapsulate any number of predefined named configurations (e.g. "dev", "testnet", "staging", "alice"), and make them accessible via the runtime API.
The named configuration within runtime would be provided in the form of JSON patch (meaning that only the customized fields would be included into it). Refer to [`sc_chain_spec`](https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/index.html#genesis-initial-runtime-state) doc to find out more about genesis config patch format

The `GenesisBuilder` API could be updated as follows (draft):
```rust
fn get_preset_names() -> Vec<String>;
fn get_preset(name: Option<String>) -> serde_json::Value;
fn build_state(value: serde_json::Value);
```

The `ChainsSpecBuilder` (introduced in #1256) would be extended with one additional method: `with_named_genesis_config_patch`, allowing chain-spec construction with the following code:
```rust
type RococoChainSpec = ChainSpecBuilder<(), Extensions>;

RococoChainSpec::builder(rococo::WASM_BINARY, Extensions::default()) 
   .with_name("Rococo Staging Testnet")
   .with_id("rococo_staging_testnet")
[...]

PR #2714, opened 2023-12-14
`GenesisConfig` presets for runtime

michalkucharczyk:
The runtime now can provide a number of predefined presets of `RuntimeGenesisConfig` struct. This presets are intended to be used in different deployments, e.g.: `local`, `staging`, etc, and should be included into the corresponding chain-specs.

Having `GenesisConfig` presets in runtime allows to fully decouple node from runtime types (the problem is described in #1984).

**Summary of changes:**
- The `GenesisBuilder` API was adjusted to enable this functionality (and provide better naming - #150):
   ```rust
    fn preset_names() -> Vec<PresetId>;
    fn get_preset(id: Option<PresetId>) -> Option<serde_json::Value>; //`None` means default
    fn build_state(value: serde_json::Value);
    pub struct PresetId(Vec<u8>);
   ```

- **Breaking change**:   Old `create_default_config` method was removed, `build_config` was renamed to `build_state`. As a consequence a node won't be able to interact with genesis config for older runtimes. The cleanup was made for sake of API simplicity. Also IMO maintaining compatibility with old API is not so crucial.
- Reference implementation was provided for `substrate-test-runtime` and `rococo` runtimes. For rococo new [`genesis_configs_presets`](https://github.com/paritytech/polkadot-sdk/blob/3b41d66b97c5ff0ec4a1989da5ffd8b9f3f588e3/polkadot/runtime/rococo/src/genesis_config_presets.rs#L530) module was added and is used in `GenesisBuilder` [_presets-related_](https://github.com/paritytech/polkadot-sdk/blob/3b41d66b97c5ff0ec4a1989da5ffd8b9f3f588e3/polkadot/runtime/rococo/src/lib.rs#L2462-L2485) methods.

- The `chain-spec-builder` util was also improved and allows to ([_doc_](https://github.com/paritytech/polkadot-sdk/blob/3b41d66b97c5ff0ec4a1989da5ffd8b9f3f588e3/substrate/bin/utils/chain-spec-builder/src/lib.rs#L19)):
   - list presets provided by given runtime (`list-presets`),
   - display preset or default config provided by the runtime (`display-preset`),
   - build chain-spec using named preset (`create ... named-preset`),

- The `ChainSpecBuilder` is extended with [`with_genesis_config_preset_name`](https://github.com/paritytech/polkadot-sdk/blob/3b41d66b97c5ff0ec4a1989da5ffd8b9f3f588e3/substrate/client/chain-spec/src/chain_spec.rs#L447) method which allows to build chain-spec using named preset provided by the runtime. Sample usage on the node side [here](https://github.com/paritytech/polkadot-sdk/blob/2caffaae803e08a3d5b46c860e8016da023ff4ce/polkadot/node/service/src/chain_spec.rs#L404).

Implementation of #1984.
fixes: #150
part of: #25


michalkucharczyk:
This requires #2044.

Allows some nice functionality:
```
./chain-spec-builder list-presets -r wbuild/substrate-test-runtime/substrate_test_runtime.compact.wasm
Known presets are:
[
    "foobar",
    "staging",
]
```

```
./chain-spec-builder display-preset -r wbuild/substrate-test-runtime/substrate_test_runtime.compact.wasm -p "staging"
{"balances":{"balances":[["5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",1000000000000000],["5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",1000000000000000]]},"substrateTest":{"authorities":["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY","5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL"]}}
```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4729765

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -49,6 +57,22 @@ sp_api::decl_runtime_apis! {
...
+		/// Otherwise returns a JSON representation of the built-in, named, pre-configured `GenesisConfig` value.
+		#[api_version(2)]
+		fn get_preset(id: Option<sp_std::vec::Vec<u8>>) -> Option<sp_std::vec::Vec<u8>>;
+
+		/// Returns a list of available builtin `GenesisConfig` presets.
+		///
+		/// The presets from the list can be queried with [`GenesisBuilder::get_preset`] method. If no named presets are
+		/// provided by the runtime the list is empty.
+		#[api_version(2)]
+		fn preset_names() -> sp_std::vec::Vec<sp_runtime::RuntimeString>;
```
michalkucharczyk, line 76: @bkchr I'd appreciate feedback on naming. `config` is not advised (see #150), so maybe `preset` will do the job?

I am also not sure if we need to keep `create_default_config` method. Maybe we can remove it?

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -49,6 +57,22 @@ sp_api::decl_runtime_apis! {
...
+		#[renamed("build_config", 2)]
+		fn build_state(json: sp_std::vec::Vec<u8>) -> Result;
+
+		/// Returns a JSON blob representation of the builtin `GenesisConfig` identified by `id`.
+		///
+		/// If `id` is `None` the function instantiates the default `GenesisConfig` struct for the runtime and serializes it into a JSON
+		/// blob. It returns a `Vec<u8>` containing the JSON representation of the default `GenesisConfig`.
+		/// Otherwise returns a JSON representation of the built-in, named, pre-configured `GenesisConfig` value.
+		#[api_version(2)]
+		fn get_preset(id: Option<sp_std::vec::Vec<u8>>) -> Option<sp_std::vec::Vec<u8>>;
```
michalkucharczyk, line 69: maybe:
```suggestion
		fn get_preset(id: Option<RuntimeString>) -> Option<sp_std::vec::Vec<u8>>;
```

michalkucharczyk: I had some troubles with using RuntimeString in [match](https://github.com/paritytech/polkadot-sdk/blob/249b8136ad19addd97e73e01f426e990d543718a/substrate/test-utils/runtime/src/lib.rs#L758-L760) , don't really see much benefit from having it here.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4737846

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4737881

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4737882

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4737994

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4738698

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4738837

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4739656

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4739926

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4740203

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4740238

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4740712

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: `String` :facepalm:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4742299

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4826028

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5515002

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5516048

substrate/primitives/genesis-builder/src/lib.rs
ggwpez: >I am also not sure if we need to keep create_default_config method. Maybe we can remove it?

I am currently using that in the new chain-agnostic runtime benchmarker (https://github.com/paritytech/polkadot-sdk/pull/3512), how would i know what preset is the default?

substrate/primitives/genesis-builder/src/lib.rs
ggwpez: If the `preset_names` return Strings and the `get_preset` accepts `Vec<u8>` then it is up to the API user to figure out that he is supposed to UTF8 encode that?  
Maybe use `Vec<u8>` or `RuntimeString` everywhere?

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: `get_preset(None)` would return default.

michalkucharczyk:
> The `get_preset` does not quite align with the `create_default_config` that is currently used. Maybe `create_preset(name ..)` or add an additional `default_preset_name()`?

`create_default_config` was not most fortunate name probably. _create_ meant instantiate and serialize to json.
Currently my reasoning is: `get_preset` seems to be more appropriate - there are some presets embedded in runtime (so we can _get_ them), and serializing default is cheap so _get_ fits here well. 

But I am open to suggestions.

ggwpez:
> `create_default_config` was not most fortunate name probably. _create_ meant instantiate and serialize to json.

Yea i was slightly confused by the naming when seeing it first time 😅  
I dont have strong opinion on how to name them, but it should be consistent.

michalkucharczyk:
> > `create_default_config` was not most fortunate name probably. _create_ meant instantiate and serialize to json.
> 
> Yea i was slightly confused by the naming when seeing it first time 😅 I dont have strong opinion on how to name them, but it should be consistent.

To be on the same page: I think that maybe we can remove V1 with `create_default_config`. I don't think there is any requirement for maintaining this. If the old function is removed then `get_preset` maybe more appropriate.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5516962

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5517600

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5520991

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5522451

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: Did some cleanup here. 

I dropped old API and introduced new functions as I assume the node will not support old runtimes when it comes to spec generating.

Introduced `PresetId`.

michalkucharczyk: Did some cleanup here. 

I dropped old API and introduced new functions as I assume the node will not support old runtimes when it comes to spec generating.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5574908

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5575818

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5576295

michalkucharczyk:
One more thought. I overlooked this...

I think we need also to put production preset (in the raw form) into the runtime.

Imagine a scenario when you just want to run a local node which connects to production network.

The `cli` interface must be exactly the same as for other presets. We cannot expect used to go somewhere and find the json file for production.

It can be easily addressed in `GenesisBuilder` API. `get_preset(id:PresetId)` method could return the json blob wrapped into some enum `enum { Default, Patch, Raw }`, which would indicate how exactly blob should be used.

- `default` - can be used a base for patch,
- `patch` - set of key-value pairs that should be merged into default config,
- `raw` - raw representation of state, which does need to be passed to `build_state` function.

Having this we could embed [this file](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/service/chain-specs/polkadot.json) into the runtime. Today it is done on the node side: https://github.com/paritytech/polkadot-sdk/blob/9d2963c29d9b7ea949851a166e0cb2792fc66fff/polkadot/node/service/src/chain_spec.rs#L95-L97

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,90 @@
...
+use alloc::vec::Vec;
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
 
 /// The result type alias, used in build methods. `Err` contains formatted error message.
 pub type Result = core::result::Result<(), sp_runtime::RuntimeString>;
 
+/// Wrapper representing preset id.
+#[derive(Decode, Encode, Debug, TypeInfo, PartialEq)]
+pub struct PresetId(Vec<u8>);
```
kianenigma, line 59: I think this is fine, if you want to be pedantic you can use something like `BoundedVec<u8>` to have a limit on the size, or use a fixed size byte array as identifier.

polkadot/runtime/rococo/src/genesis_config_presets.rs
```diff
@@ -0,0 +1,544 @@
...
+		BeefyId,
+	)>,
+	root_key: AccountId,
+	endowed_accounts: Option<Vec<AccountId>>,
+) -> serde_json::Value {
+	let endowed_accounts: Vec<AccountId> = endowed_accounts.unwrap_or_else(testnet_accounts);
+
+	const ENDOWMENT: u128 = 1_000_000 * ROC;
+
+	serde_json::json!({
```
kianenigma, line 180: If you want to be really pedantic about these refactors, you can check the output of this against a previous output that is generated from the node side and assert that while migrating to using runtime presets, everything remains backwards compatible.

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -2460,12 +2460,26 @@ sp_api::impl_runtime_apis! {
...
+			if let Some(id) = id {
+				genesis_config_presets::get_preset(id)
+			} else {
+				get_preset::<RuntimeGenesisConfig>(&None)
+			}
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn preset_names() -> Vec<PresetId> {
```
kianenigma, line 2475: minor annoyance that the list has to be duplicated between here and `genesis_config_preset.rs`, but I don't have a better solution off the top of my head either 🤷

polkadot/runtime/rococo/src/genesis_config_presets.rs
michalkucharczyk: This was already done when migrating to `json!` format. Test were added in #1256 and remove in this commit: https://github.com/paritytech/polkadot-sdk/pull/1256/commits/f4d3f32cb8c974a3b6b06ca8248daa99f3d0266e

polkadot/runtime/rococo/src/lib.rs
michalkucharczyk: We could have a `Map` defined somewhere once. The preset_names would be `map.keys()` maybe.

substrate/client/chain-spec/src/genesis_config_builder.rs
```diff
@@ -84,30 +84,46 @@ where
...
+
 		let call_result = self
-			.call(&mut ext, "GenesisBuilder_build_config", &config.to_string().encode())
+			.call(&mut ext, "GenesisBuilder_build_state", &json_pretty_str.encode())
 			.map_err(|e| format!("wasm call error {e}"))?;
 
 		BuildResult::decode(&mut &call_result[..])
-			.map_err(|e| format!("scale codec error: {e}"))??;
+			.map_err(|e| format!("scale codec error: {e}"))?
+			.map_err(|e| format!("{e} for blob:\n{}", json_pretty_str))?;
```
michalkucharczyk, line 126: This _pretty json_ ideally should be embedded into returned `Error` struct. For now it allows for better error finding by CLI user as @davxy pointed out during meeting.

michalkucharczyk: <details>
  <summary>Example</summary>
  
  ```
  Invalid JSON blob: BadLength at line 28 column 8 for blob:
  {
    "assignedSlots": {
      "config": null,
      "maxPermanentSlots": 0,
      "maxTemporarySlots": 0
    },
    "authorityDiscovery": {
      "keys": []
    },
    "babe": {
      "authorities": [],
      "epochConfig": {
        "allowed_slots": "PrimaryAndSecondaryVRFSlots",
        "c": [
          1,
          4
        ]
      }
    },
    "balances": {
      "balances": [
        [
          "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
          1000000000000000000
        ],
        [
          "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694tyx",
          1000000000000000000
        ],
  ```
</details>

substrate/bin/utils/chain-spec-builder/src/lib.rs
```diff
@@ -28,60 +28,94 @@
...
+//!	```bash
+//! chain-spec-builder create -r runtime.wasm default
 //! ```
-//!
-//! _Note:_ [`GenesisBuilder::create_default_config`][sp-genesis-builder-create] runtime function is
+//! 
+//! _Note:_ [`GenesisBuilder::get_preset`][sp-genesis-builder-get-preset] runtime function is
 //! called.
 //!
 //!
```
kianenigma, line 42: I've always wondered if we can use `docify` to document CLIs or not. The only reasonable hack is to write a cli-test that performs an action eg `chain-spec-builder display-preset -r runtime.wasm` then display it here.

substrate/bin/utils/chain-spec-builder/src/lib.rs
```diff
@@ -137,6 +173,7 @@ enum GenesisBuildAction {
 	Patch(PatchCmd),
 	Full(FullCmd),
 	Default(DefaultCmd),
+	NamedPreset(NamedPresetCmd),
```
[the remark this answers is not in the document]
michalkucharczyk: It seems to be fine already:
```
$ ./chain-spec-builder create --help
Create a new chain spec by interacting with the provided runtime wasm blob

Usage: chain-spec-builder create [OPTIONS] --runtime-wasm-path <RUNTIME_WASM_PATH> <COMMAND>

Commands:
  patch         Patches the runtime's default genesis config with provided patch
  full          Build the genesis config for runtime using provided json file. No defaults will be used
  default       Gets the default genesis config for the runtime and uses it in ChainSpec. Please note that default genesis config may not be valid. For some runtimes initial values should be added there (e.g. session keys, babe epoch)
  named-preset  Uses named preset provided by runtime to build the chains spec
  help          Print this message or the help of the given subcommand(s)

Options:
  -n, --chain-name <CHAIN_NAME>
          The name of chain [default: Custom]
  -i, --chain-id <CHAIN_ID>
          The chain id [default: custom]
  -r, --runtime-wasm-path <RUNTIME_WASM_PATH>
          The path to runtime wasm blob
  -s, --raw-storage
          Export chainspec as raw storage
  -v, --verify
          Verify the genesis config. This silently generates the raw storage from genesis config. Any errors will be reported
  -h, --help
          Print help
```

substrate/client/chain-spec/src/chain_spec.rs
```diff
@@ -1010,6 +1035,38 @@ mod tests {
 	}
 
 	#[docify::export]
```
kianenigma, line 1037: now sure where it is used, but happy to see this :)

michalkucharczyk: https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/index.html#building-a-chain-specification

substrate/client/chain-spec/src/genesis_config_builder.rs
```diff
@@ -84,30 +84,46 @@ where
...
+	/// provided by the `runtime`.
+	pub fn get_named_preset(&self, id: Option<&String>) -> core::result::Result<Value, String> {
 		let mut t = BasicExternalities::new_empty();
 		let call_result = self
-			.call(&mut t, "GenesisBuilder_create_default_config", &[])
+			.call(&mut t, "GenesisBuilder_get_preset", &id.encode())
 			.map_err(|e| format!("wasm call error {e}"))?;
-		let default_config = Vec::<u8>::decode(&mut &call_result[..])
+
+		let named_preset = Option::<Vec<u8>>::decode(&mut &call_result[..])
```
kianenigma, line 103: has anyone demystified why `decode` needs a `&mut`? :))) I have always known it by heart but never got a good explanation.

kianenigma: (I am assuming it never mutates the original buffer, maybe it does?)

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,92 @@
...
 //! Substrate genesis config builder
 //!
-//! This Runtime API allows to construct `RuntimeGenesisConfig`, in particular:
-//! - serialize the runtime default `RuntimeGenesisConfig` struct into json format,
-//! - put the RuntimeGenesisConfig struct into the storage. Internally this operation calls
-//!   `GenesisBuild::build` function for all runtime pallets, which is typically provided by
-//!   pallet's author.
-//! - deserialize the `RuntimeGenesisConfig` from given json blob and put `RuntimeGenesisConfig`
-//!   into the state storage. Allows to build customized configuration.
+//! This module provides means to interact with `RuntimeGenesisConfig`. Runtime provides a default
```
kianenigma, line 22: a holistic question: is this crate only capable of providing all of its functionality to non-FRAME-based runtimes? 
If no, we can move it to FRAME and possibly make life easier. 
If yes, it might be worth explaining that `RuntimeGenesisConfig` is the name that FRAME-based runtimes have for their Genesis struct and it is not a "magic" word. 

Not an important point, and ofc we don't really have FRAME alternatives, but is is good to think about this :)

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,92 @@
...
+	fn try_from(v: &'a PresetId) -> core::result::Result<&'a str, Self::Error> {
+		core::str::from_utf8(&v.0)
+	}
+}
+
 sp_api::decl_runtime_apis! {
 	/// API to interact with RuntimeGenesisConfig for the runtime
 	pub trait GenesisBuilder {
-		/// Creates the default `RuntimeGenesisConfig` and returns it as a JSON blob.
+		/// Build `RuntimeGenesisConfig` from a JSON blob not using any defaults and store it in the
```
kianenigma, line 77: For example in a NON-frame based runtime, you can still use this API, and you receive this JSON, and you somehow put it in the storage, and there is no `RuntimeGenesisConfig` involved.

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,92 @@
...
+		/// storage.
 		///
-		/// This function instantiates the default `RuntimeGenesisConfig` struct for the runtime and serializes it into a JSON
-		/// blob. It returns a `Vec<u8>` containing the JSON representation of the default `RuntimeGenesisConfig`.
-		fn create_default_config() -> alloc::vec::Vec<u8>;
+		/// This function deserializes the full `RuntimeGenesisConfig` from the given JSON blob and
```
kianenigma, line 80:
```suggestion
		/// In the case of a FRAME-based runtime, this function deserializes the full `RuntimeGenesisConfig` from the given JSON blob and
```

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,92 @@
...
+}
+
+impl<'a> TryFrom<&'a PresetId> for &'a str {
+	type Error = core::str::Utf8Error;
+	fn try_from(v: &'a PresetId) -> core::result::Result<&'a str, Self::Error> {
+		core::str::from_utf8(&v.0)
+	}
+}
+
 sp_api::decl_runtime_apis! {
```
kianenigma, line 74: Does this resonate with you as a good representation of the user flow? 

https://mermaid.live/edit#pako:eNqFUU1LxDAU_Cvlndsl_di0zcGLiiB6qTeNLLF97QbapCQp7lr63822InhYfKc3w8w8HjNDrRsEBm2vP-ujMC54qrgK_NyfHBol-g09oEIr7a1WrewerVYbXU3KyQEDrv6agii6-c8TRR26w2jQorsiv3LbWz8m2TcH64TDi7dK3n5i37mCEAY0g5CNf2y-ZHBwRxyQA_Nrg62YeseBq8VLxeT0y1nVwFrRWwxhGhufeidFZ8Twy45CvWrtsTPTBoHNcAJWpjualHtS0iQr85QUIZyBxXmyK2hapgXJaZESki0hfK0BZNUnNE49v6c0i0PARjptnrcu1kqWbz59g20

kianenigma: the narration would be: You always need a genesis config json file. You can either get this from the runtime by generating it from a preset (so the outcome of a preset is a genesis config), or from other extranl sources, like `chain-spec-builder`.

templates/parachain/runtime/src/lib.rs
```diff
@@ -756,12 +756,16 @@ impl_runtime_apis! {
...
 	impl sp_genesis_builder::GenesisBuilder<Block> for Runtime {
-		fn create_default_config() -> Vec<u8> {
-			create_default_config::<RuntimeGenesisConfig>()
+		fn build_state(config: Vec<u8>) -> sp_genesis_builder::Result {
+			build_state::<RuntimeGenesisConfig>(config)
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
```
kianenigma, line 763: Given that these are templates usable in testing and education, it would be a good idea to add presets to them. Do you intend in doing this? if not I can open a branch on top of this PR.

kianenigma (approved):
LGTM, looking forward to having this as it helps a lot with a proper omni-node story.

polkadot/runtime/rococo/src/lib.rs
kianenigma: Yeah, it has to be a mutable static variable AFAIK.

substrate/bin/utils/chain-spec-builder/bin/main.rs
```diff
@@ -71,6 +81,36 @@ fn main() -> Result<(), String> {
...
+				.into_iter()
+				.map(|preset| {
+					String::from(
+						TryInto::<&str>::try_into(&preset)
+							.unwrap_or_else(|_| "cannot display preset id")
+							.to_string(),
+					)
+				})
+				.collect();
+			println!("Known presets are:\n{presets:#?}");
```
skunert, line 102: Tiny nit: I was playing a bit with the chain-spec-builder, and thought about the output here. IMO we could have this as json output as it is to have it machine-readable (e.g. you can forward to a file and read it in another script). But then we should probably remove the "Known presets" part.

Or have it for manual use, but I think then it would be preferable to just have a plain list without anything, but really no strong opinion.

substrate/client/chain-spec/src/chain_spec.rs
skunert: Hmm, but that uses `build_chain_spec_with_patch_works` and not `generate_chain_spec_with_named_preset_works`, but both are exported via docify, did you intend to replace this maybe?

skunert (approved):
Looks really nice, only silly nits.

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: This is not on-chain so I see no point in imposing size limits here. Why would you use `BoundedVec` here?

substrate/client/chain-spec/src/genesis_config_builder.rs
michalkucharczyk: AIRC it is kinda [read pointer](https://github.com/paritytech/parity-scale-codec/blob/ca68d2edbe3c3b1bfcb2272202c62eb80d146195/src/codec.rs#L120) into the input buffer which must be moved during read process.

substrate/client/chain-spec/src/chain_spec.rs
michalkucharczyk: just blindly copied, fixed. `build_chain_spec_with_patch_works` is still used.

templates/parachain/runtime/src/lib.rs
michalkucharczyk: So my intention here was to get to consensus about the general approach. Once agreed upon and merged I planned to migrate all the runtimes one by one (including fellowship runtimes).

I'll appreciate help with this.

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: Yes.

But my mental model is:
-  _external_ is somehow hand-crafted or edited/patched genesis config json file (e.g. containing manually added endowed account or parachain configuration for zombie network).
- `chain-spec-builder` is just an util that helps us to build the chain-spec using different source, one being runtime's `GenesisBuilder`.

substrate/bin/utils/chain-spec-builder/bin/main.rs
michalkucharczyk: done in: 23067cdc4fe165d4686f527f1aff63e4e665eb96

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: It can be even trickier. In runtime interface description we shall not refer to any implementation details.
The `Vec<u8>` does not necessarily need to be a JSON blob. If we want to keep the doc super-generic and non opinionated we should rewrite it, and make a special section how is it implemented for FRAME?

Not sure what approach should we take here.

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: I added some note for non-FRAME runtimes.

michalkucharczyk: 92e6824b1a2

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: 92e6824b1a2

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: done in: 71589d9ef0a

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5655840

polkadot/runtime/rococo/src/lib.rs
michalkucharczyk: Maybe we could use [`async::Lazy`](https://docs.rs/once_cell/latest/once_cell/unsync/struct.Lazy.html) for this?

If that is fine for you, this could be taken to the follow-up PR. We either-way need to re-work the rest of runtimes so we can figure out something and fix rococo/substrate-test-runtime.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5726678

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5726829

substrate/primitives/genesis-builder/src/lib.rs
```diff
@@ -19,39 +19,107 @@
...
+		)
+	}
+}
+
+impl<'a> TryFrom<&'a PresetId> for &'a str {
+	type Error = core::str::Utf8Error;
+	fn try_from(v: &'a PresetId) -> core::result::Result<&'a str, Self::Error> {
+		core::str::from_utf8(&v.0)
+	}
+}
```
bkchr, line 87: Why isn't this just `RuntimeString`?

substrate/frame/support/src/genesis_builder_helper.rs
```diff
@@ -19,26 +19,37 @@
...
 }
+
+/// Get the default `GenesisConfig` as a JSON blob if `id` is None.
+///
+/// No named presets are supported. For more info refer to
+/// [`sp_genesis_builder::GenesisBuilder::get_preset`].
+pub fn get_preset<GC>(id: &Option<PresetId>) -> Option<Vec<u8>>
+where
+	GC: BuildGenesisConfig + Default,
+{
```
bkchr, line 45: This naming is totally weird in contrast to the documentation and the code.

`fn get_default_preset() -> Option<Vec<u8>>` probably makes more sense here.

bkchr: Or pass some closure that should handle the `Some(whatever)` cases.

bkchr (commented):
Just skimmed this very quickly. Looks good from what I have seen. Left 2 comments on things that should be improved.

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: Could be easily done with this extension: #3942 (initially I didn't want to add this).

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1648,12 +1648,16 @@ impl_runtime_apis! {
...
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
+			get_preset::<RuntimeGenesisConfig>(id)
+		}
+
+		fn preset_names() -> Vec<sp_genesis_builder::PresetId> {
+			vec![]
```
ggwpez, line 1660: So the number of available presets is `preset_names().len() + 1` since the Default is modelled as `None`?  
I guess its fine.

michalkucharczyk: Yes. 

But the `default` itself is not a preset as the other ones. It is a default value which may not guarantee that _runtime_ will work correctly (it may miss some important settings, e.g. initial authorities).  It is a base upon which the presets are applied to create operational initial state.

Initially there was dedicated function for default, but `None` also should work fine here. (ref: https://github.com/paritytech/polkadot-sdk/issues/62#issuecomment-1775078514)

substrate/test-utils/runtime/src/lib.rs
```diff
@@ -725,12 +728,44 @@ impl_runtime_apis! {
...
-		fn create_default_config() -> Vec<u8> {
-			create_default_config::<RuntimeGenesisConfig>()
+		fn build_state(config: Vec<u8>) -> sp_genesis_builder::Result {
+			build_state::<RuntimeGenesisConfig>(config)
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn get_preset(params: &Option<PresetId>) -> Option<Vec<u8>> {
+			Some(if let Some(params) = params {
```
ggwpez, line 736: I think you can un-nest this function once with `if let .. else`.

substrate/test-utils/runtime/src/lib.rs
```diff
@@ -1290,21 +1323,60 @@ mod tests {
 		fn default_config_as_json_works() {
 			sp_tracing::try_init_simple();
 			let mut t = BasicExternalities::new_empty();
-			let r = executor_call(&mut t, "GenesisBuilder_create_default_config", &vec![]).unwrap();
-			let r = Vec::<u8>::decode(&mut &r[..]).unwrap();
+			let r = executor_call(&mut t, "GenesisBuilder_get_preset", &None::<&PresetId>.encode())
+				.unwrap();
+			let r = Option::<Vec<u8>>::decode(&mut &r[..])
+				.unwrap()
+				.expect("defuault config is there");
```
ggwpez, line 1330:
```suggestion
				.expect("default config is there");
```

ggwpez (approved):
LGTM from what i can judge.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
ggwpez: Ah i see. I thought that the default would be valid, okay 😄

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1648,12 +1648,16 @@ impl_runtime_apis! {
 	}
 
 	impl sp_genesis_builder::GenesisBuilder<Block> for Runtime {
-		fn create_default_config() -> Vec<u8> {
-			create_default_config::<RuntimeGenesisConfig>()
+		fn build_state(config: Vec<u8>) -> sp_genesis_builder::Result {
```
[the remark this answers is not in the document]
michalkucharczyk: It is a `json` representation of  `RuntimeGenesisConfig` not preset name.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1648,12 +1648,16 @@ impl_runtime_apis! {
...
+			build_state::<RuntimeGenesisConfig>(config)
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
+			get_preset::<RuntimeGenesisConfig>(id)
+		}
+
+		fn preset_names() -> Vec<sp_genesis_builder::PresetId> {
```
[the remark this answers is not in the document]
michalkucharczyk: I'd prefer to keep `preset_names`. I can rename `PresetId` to `PresetName`, or get rid of this type as I am gonna to do following in a follow up commit:
```rust
pub type PresetId = RuntimeString;
```

wdyt?

cumulus/parachains/runtimes/people/people-westend/src/lib.rs
```diff
@@ -850,12 +850,16 @@ impl_runtime_apis! {
...
+			build_state::<RuntimeGenesisConfig>(config)
 		}
 
-		fn build_config(config: Vec<u8>) -> sp_genesis_builder::Result {
-			build_config::<RuntimeGenesisConfig>(config)
+		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
+			get_preset::<RuntimeGenesisConfig>(id)
+		}
+
+		fn preset_names() -> Vec<sp_genesis_builder::PresetId> {
```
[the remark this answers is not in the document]
michalkucharczyk: Some runtimes are gonna overwrite this implementation as they are providing presets. How would that work with macro?

polkadot/node/service/src/chain_spec.rs
```diff
@@ -682,7 +402,7 @@ pub fn rococo_staging_testnet_config() -> Result<RococoChainSpec, String> {
 	.with_name("Rococo Staging Testnet")
 	.with_id("rococo_staging_testnet")
 	.with_chain_type(ChainType::Live)
-	.with_genesis_config_patch(rococo_staging_testnet_config_genesis())
+	.with_genesis_config_preset_name("staging_testnet")
```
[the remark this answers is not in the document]
michalkucharczyk: We need flexibility here. We cannot predict all genesis config presets and put them in enum. We also need to allow runtime devs to define their own presets (e.g. for their own testnets or CI tests).

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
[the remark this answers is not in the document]
ggwpez: ups 🙃

polkadot/node/service/src/chain_spec.rs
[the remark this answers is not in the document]
ggwpez: It could have `Benchmarking` and `Named(RuntimeString)` variants? But i dont want to prevent this MR from merging, just ideas.

substrate/test-utils/runtime/src/lib.rs
michalkucharczyk: reworked in a39c4bb

substrate/frame/support/src/genesis_builder_helper.rs
michalkucharczyk: makes sense, done in: a773494

substrate/primitives/genesis-builder/src/lib.rs
michalkucharczyk: done in: c374a37
Maybe we don't need extra type at all....

polkadot/node/service/src/chain_spec.rs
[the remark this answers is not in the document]
michalkucharczyk: > How would the benchmark-runner know what preset to pick?

How is it done today?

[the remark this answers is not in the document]
ggwpez: It uses the default. But that meant that i had to change the default for a few pallets to be the worst cast.  
Not sure what is best - either to have a dedicated preset - or to add more traits to the pallets to overwrite the config as preparation for benchmarks.   
For now i can probably continue to use the default, as it seems to work okay-ish.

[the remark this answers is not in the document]
michalkucharczyk: > It could have `Benchmarking` and `Named(RuntimeString)` variants? But i dont want to prevent this MR from merging, just ideas.

It is inline with my previous proposal. Maybe something to consider in the future:
https://github.com/paritytech/polkadot-sdk/pull/2714#issuecomment-2015125305

cumulus/parachains/runtimes/people/people-westend/src/lib.rs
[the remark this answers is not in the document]
ggwpez: Yea what i mean is that the macro is not smart enough to fill in missing functions with default impls.  
With a derive macro that would be simple enough but with a proc macro its overkill.

Polkadot-Forum:

https://forum.polkadot.network/t/stabilizing-polkadot/7175/24

PR #2715, opened 2023-12-14
xcm-builder: describe local location

muharem:
`DescribeLocation` implementation for local xcm locations.

Currently, our runtimes lack a conversion mechanism for translating local root/pallet/plurality xcm locations into account IDs. However, there are instances where this functionality is needed - https://github.com/polkadot-fellows/runtimes/pull/125.

In this PR, I propose three options to address this issue:
1. Include `(0, Here)` description into the `DescribeFamily` type for the root location. Utilize terminals like `DescribePalletTerminal` and `DescribeBodyTerminal` independently for the remaining locations without any additional prefixes.
2. Introduce a type similar to `DescribeFamily` called `DescribeLocal`.
3. Introduce a more generic type called `DescribeWithPrefix`.

### DIFF
--- a/polkadot/xcm/xcm-builder/src/location_conversion.rs
+++ b/polkadot/xcm/xcm-builder/src/location_conversion.rs
@@ -15,7 +15,7 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use crate::universal_exports::ensure_is_remote;
-use frame_support::traits::Get;
+use frame_support::traits::{Contains, Get};
 use parity_scale_codec::{Compact, Decode, Encode};
 use sp_io::hashing::blake2_256;
 use sp_runtime::traits::{AccountIdConversion, TrailingZeroInput, TryConvert};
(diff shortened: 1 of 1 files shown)


gavofyork:
This totally misses the point. What has been added is merely a confusing duplicate of `DescribeTerminus`.

muharem:
> The Plurality and pallet instance stuff is probably fine (though plurality might better be described with readable language rather than just encoding some value (which isn't necessarily expected to stay the same between XCM versions)).
> 
> The rest totally misses the point. What has been added is merely a confusing duplicate of `DescribeTerminus`.

I see, I wasn't certain whether to utilize `DescribeTerminus` and other `Terminals` for local locations without a prefix. Specifically, `DescribeTerminus` results an empty byte slice. And `DescribeLocal` was intended to be used in conjunction with existing terminals, only to have a prefix.

However, if I can simply use `DescribeTerminus` to describe `(0, Here)`, then we do not need any of this.

gavofyork:
(And actually, we already have `DescribePalletTerminal` for pallet instances.)

muharem:
> (And actually, we already have `DescribePalletTerminal` for pallet instances.)

yes, and for pluralities - `DescribeBodyTerminal`. I meant only to add `LocalChain` prefix to them, something like,
- `DescribeLocal<(DescribeTerminus, DescribePalletTerminal, DescribeBodyTerminal)>`
- or `DescribeWithPrefix<.., b"LocalChain", (DescribeTerminus, DescribePalletTerminal, DescribeBodyTerminal)>`

but never mind, without prefix also looks good to me

PR #2454, opened 2023-11-22
[WIP] Migrate from `SignedExtension` to `TransactionExtension`

georgepisaltu:
Follow up effort on top of https://github.com/paritytech/polkadot-sdk/pull/2280

- [x] implement `TransactionExtension` for all `SignedExtension` `impl` instances
- [x] remove `AsTransactionExtension` adapter
- [ ] mark `SignedExtension` as deprecated and remove all `SignedExtension` impl` instances
- [ ] clean up any `SignedExtension` remnants in docs or other names

PR #2726, opened 2023-12-15
Add benchmarks to transaction extensions

georgepisaltu:
Follow up effort on top of https://github.com/paritytech/polkadot-sdk/pull/2280

This PR implements benchmarking code for transaction extensions:

- ChargeTransactionPayment
- ChargeAssetTxPayment (asset-tx-payment)
- ChargeAssetTxPayment (asset-conversion-tx-payment)
- CheckWeight
- CheckTxVersion
- CheckSpecVersion
- CheckNonce
- CheckNonZeroSender
- CheckMortality
- CheckGenesis
- CheckOnlySudoAccount
- PrevalidateAttests

Extensions which still don't have benchmarks but could probably continue to use the default impl of `Weight::zero()` or call some existing weight function in the case of wrappers:

- Test extensions
   - DummyExtension
   - WatchDummy
- Wrappers
   - GenericTransactionExtension
   - TransactionExtension (chain-polkadot-bulletin)
- Bridge extensions (see discussion below)
   - RefundTransactionExtensionAdapter

Each extension's `TransactionExtensionBase::weight` impl needs to use the resulting weight function and adapt existing runtimes' benchmarking configuration to these changes. So far, this has been done for:

- [x] ChargeTransactionPayment
- [x] ChargeAssetTxPayment (asset-tx-payment)
- [x] ChargeAssetTxPayment (asset-conversion-tx-payment)
- [x] CheckWeight
- [x] CheckTxVersion
- [x] CheckSpecVersion
- [x] CheckNonce
- [x] CheckNonZeroSender
- [x] CheckMortality
- [x] CheckGenesis
- [x] CheckOnlySudoAccount
- [x] PrevalidateAttests
- [ ] ~RefundTransactionExtensionAdapter~


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4739789

svyatonik:
Re `RefundTransactionExtensionAdapter` - it is a bit complicated case && we already have some benchmarking there. I haven't checked - where this weight will be used. If it'll be added to weight of all runtime calls (as a base extrinsic weight or something like that), then it won't work for us - extensions does non-trivial things. So couple of details:
1) first thing that bridge extensions do (in `pre_dispatch`/`validate`), is that they check whether the `call` is the call of one of bridge pallets or batch of those calls **by matching.** So for most calls the weight should be near zero and adding it to the base tx weight would mean that the signer overpays and we waste block space;
2) then it does some checks in  `pre_dispatch` that are then repeated by the calls themselves. They're mostly about reading tdb values and comparing them to arguments. Important thing there is that since those reads + comparisons are then duplicated, we do not account `weight` spent on this here, in extension;
3) if those checks ^^^ are failed, then we do not include transaction to the block, so no weight is spent;
4) in `post_dispatch` there are two possible options that could happen and actions on both branches are non-trivial. That's why we add `max` of those two weights to the weight of our bridge call. So if we'll add it to the base tx weight, it would mean duplication.

So the main question here is whether this weight you got through benchmarks is added to the base tx weight. If so, then please just left all as is (no additional benchmarks required for our extensions) - just return `Weight::zero()` from where it is required. Otherwise, please ping one of us (@paritytech/bridges-core ) - we'll help solve any issues here. Actually, I've been working on adapting bridge extensions to new scheme here: https://github.com/paritytech/polkadot-sdk/tree/sv-gav-tx-ext, but I think it was lost somewhere

georgepisaltu:
> So the main question here is whether this weight you got through benchmarks is added to the base tx weight. If so, then please just left all as is (no additional benchmarks required for our extensions) - just return Weight::zero() from where it is required. Otherwise, please ping one of us (@paritytech/bridges-core ) - we'll help solve any issues here.

Regarding `RefundTransactionExtensionAdapter`, I agree the benchmarking there is complicated. I wanted to first come up with a worst case benchmark just to know it can be done, refining it to consume only what it should would have been incremental work. It could probably be done with "choice" arguments to the weight function to signal which code path would execute so we could choose an appropriate weight function. I'll try to come up with a POC for what is described above, but if it proves too difficult we may resort to accounting for this extension's weight in a different way and just return `Weight::zero()` in this extension's implementation.

> Actually, I've been working on adapting bridge extensions to new scheme here: https://github.com/paritytech/polkadot-sdk/tree/sv-gav-tx-ext, but I think it was lost somewhere

This part is already done, check out the [PR doing it](https://github.com/paritytech/polkadot-sdk/pull/2454) and merging into the [original PR](https://github.com/paritytech/polkadot-sdk/pull/2280). Feel free to create additional PRs with any fixes/improvements.

georgepisaltu:
> So the main question here is whether this weight you got through benchmarks is added to the base tx weight. If so, then please just left all as is (no additional benchmarks required for our extensions) - just return Weight::zero() from where it is required.

Something I forgot to address in my original comment: in the current implementation I'm just aiming to have a weight associated with running the whole `validate` - `prepare` - `post_dispatch` flow in the trait impl (which is obviously flawed in the current iteration of the `RefundTransactionExtensionAdapter` benches). The `TransactionExtensionBase::weight()` function isn't hooked to any fee/weight management yet. My current understanding is that when run on-chain, these will need to consume weight on top of the base TX weight, but nothing stops off-chain workers from running `validate` while disregarding the weight. What I mean is that the weight function is there to associate a weight with each `validate` - `prepare` - `post_dispatch` flow; it's up to the extension user to do something (or nothing) with it. In our reference runtimes I think we will be charging weight for extensions, so if bridges deals with the weight of these extensions separately, having the weight function return `Weight::zero()` is probably correct.

svyatonik:
> My current understanding is that when run on-chain, these will need to consume weight on top of the base TX weight

That is what worries me. As I said, for non-bridge transactions, our extensions are just (almost) noops in term of weight. So if we'll have non-zero weights for that tx and it'll be added by default to the tx base weight (like `bas_weight = base_weight + SignedExtensions::pre_dispatch_weight + SignedExtensions::validate_weight + SignedExtensions::post_dispatch_weight`), that'd be a problem. I'd just leave some big detailed comment (we may do that) on how to plugin this extension to the runtime with bridge pallets and just go with the "weight function return `Weight::zero()`" approach.

georgepisaltu:
@svyatonik  at least for now, I reverted the benches for the bridge extensions altogether [here](2c8a093603dde7bbb966b51dd628bd47dff91003).

georgepisaltu:
bot bench-all substrate

substrate/frame/transaction-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,62 @@
...
+    BalanceOf<T>: Send + Sync + From<u64>,
+)]
+mod benchmarks {
+	use super::*;
+
+	#[benchmark]
+	fn charge_transaction_payment() {
+		let caller: T::AccountId = whitelisted_caller();
+		<T::OnChargeTransaction as OnChargeTransaction<T>>::endow_account(&caller, u32::MAX.into());
+		let ext: ChargeTransactionPayment<T> = ChargeTransactionPayment::from(10u64.into());
```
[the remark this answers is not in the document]
svyatonik: Actually, is adding some outcome test possible here?

substrate/frame/system/src/extensions/weights.rs
```diff
@@ -0,0 +1,197 @@
...
+		// Minimum execution time: 7_283_000 picoseconds.
+		Weight::from_parts(8_115_000, 0)
+			.saturating_add(Weight::from_parts(0, 0))
+	}
+	fn check_nonce() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 41_448_000 picoseconds.
+		Weight::from_parts(42_500_000, 0)
```
svyatonik, line 96: It should read the account data (from the Account map), so maybe something wrong with benchmarks code?

svyatonik (commented):
Bridges part is ok - we handle it in our own way, so returning zero is find for us. Regarding the rest - I see that somewhere benchmakrs have zero `proof_size` and zero db reads+writes, where there sholud be. IIUC the current base transaction weight will be replaced with sum of weights of all signed extensions. So I think we need to fix that ^^^.

Wild guess: maybe it is zero, because `whitelisted_caller` is used? Let's try to use something else?

georgepisaltu:
> Bridges part is ok - we handle it in our own way, so returning zero is find for us. Regarding the rest - I see that somewhere benchmakrs have zero `proof_size` and zero db reads+writes, where there sholud be. IIUC the current base transaction weight will be replaced with sum of weights of all signed extensions. So I think we need to fix that ^^^.
> 
> Wild guess: maybe it is zero, because `whitelisted_caller` is used? Let's try to use something else?

These are just dummy implementations until I get some real results, I just copied them from a different benchmark to have it compile. you can see above I'm trying to get the bot to run the benches. I ran some of them locally, but I didn't commit the results since I would need to overwrite them with the hardware reference ones anyway.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/weights/pallet_transaction_payment.rs
```diff
@@ -0,0 +1,37 @@
...
+
+/// Weight functions for `pallet_transaction_payment`.
+pub struct WeightInfo<T>(PhantomData<T>);
+impl<T: frame_system::Config> pallet_transaction_payment::WeightInfo for WeightInfo<T> {
+	fn charge_transaction_payment() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `57`
+		//  Estimated: `0`
+		// Minimum execution time: 3_322_000 picoseconds.
+		Weight::from_parts(3_577_000, 0)
```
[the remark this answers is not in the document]
georgepisaltu: `pallet_transaction_payment` now has real weights.

substrate/frame/system/src/extensions/weights.rs
georgepisaltu: `pallet_transaction_payment` now has real weights.

substrate/frame/transaction-payment/src/benchmarking.rs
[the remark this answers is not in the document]
georgepisaltu: `pallet_transaction_payment` now has real weights. I'll also add an event assertion.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/weights/pallet_transaction_payment.rs
```diff
@@ -25,13 +50,18 @@ use core::marker::PhantomData;
 /// Weight functions for `pallet_transaction_payment`.
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> pallet_transaction_payment::WeightInfo for WeightInfo<T> {
+	/// Storage: `TransactionPayment::NextFeeMultiplier` (r:1 w:0)
+	/// Proof: `TransactionPayment::NextFeeMultiplier` (`max_values`: Some(1), `max_size`: Some(16), added: 511, mode: `MaxEncodedLen`)
```
svyatonik, line 54: :+1:

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/weights/frame_system_extensions.rs
```diff
@@ -0,0 +1,121 @@
...
+		// Minimum execution time: 732_000 picoseconds.
+		Weight::from_parts(2_875_000, 0)
+			.saturating_add(Weight::from_parts(0, 0))
+	}
+	fn check_nonce() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 3_627_000 picoseconds.
+		Weight::from_parts(6_322_000, 0)
```
svyatonik, line 88: I still suppose that the `CheckNonce` should read from the `Accounts` map (https://github.com/paritytech/polkadot-sdk/pull/2726#discussion_r1432632910), so its `proof_size` shall be non-zero (and also it should have at least `1` db read op). My guess was that it is zero, because you're using `whitelisted_caller`. Do you think we should count that weight here?

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/weights.rs
```diff
@@ -0,0 +1,82 @@
...
+#![allow(unused_imports)]
+#![allow(missing_docs)]
+
+use frame_support::weights::Weight;
+use core::marker::PhantomData;
+
+/// Weight functions needed for pallet-asset-conversion-tx-payment.
+pub trait WeightInfo {
+	fn charge_asset_tx_payment_zero() -> Weight;
+    fn charge_asset_tx_payment_native() -> Weight;
```
svyatonik, line 29: Something wrong with formatting here

substrate/frame/transaction-payment/asset-tx-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,135 @@
...
+				.is_ok());
+		}
+	}
+
+	#[benchmark]
+	fn charge_asset_tx_payment_asset() {
+		let caller: T::AccountId = whitelisted_caller();
+		let (fun_asset_id, asset_id) = T::create_asset_id_parameter(1);
+		T::setup_balances_and_pool(fun_asset_id.clone(), caller.clone());
+		let tip = 10u32.into();
```
svyatonik, line 109: Perhaps add the similar `minimum_balance` here?

svyatonik (approved):
LGTM modulo yet missing event test in `pallet-transaction-payment` and similar stuff in the assets payment pallet. I'm also unsure about weight of `CheckNonce` - IMO it should have non-zero `proof_size`, because it reads from accounts map. And since we are measuring base transaction weight, we should not use the `whitelisted_caller` there (in other benchmarks it is fine, because it will be already counted by the `CheckNonce`).

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/weights.rs
georgepisaltu: Fixed.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/weights/frame_system_extensions.rs
georgepisaltu: I removed the whitelist for all extensions here and ran the benches again to see what reads/writes they actually do. Looking at the results, I think it's fine to run with normal accounts for all extensions, not just `CheckNonce`.

substrate/frame/transaction-payment/asset-tx-payment/src/benchmarking.rs
georgepisaltu: I think it's fine with the hardcoded `10` tip either way, all benches run fine. Depending on how quickly this PR moves, I might have time to make this change too since I would need to re-run all the benches for this pallet to make sure they work with the change.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4846365

georgepisaltu:
> missing event test in `pallet-transaction-payment` and similar stuff in the assets payment pallet

I added an event test in `pallet-transaction-payment`. However, for the other 2 similar pallets, `asset-conversion-tx-payment` and `asset-tx-payment`, asserting the event is not trivial since we can't just do a `match` on the `enum` variant, so we'd have to get all the `enum` parameters correct, which would mean re-running the code in `validate`, `prepare` and `post_dispatch` to reconstruct the values for the actual converted fees, which is difficult and would require duplicating and maintaining the logic in a benchmark specific function. Judging from the bench results (and my testing while writing this), the correct code is running in each function, so I'd be inclined to leave this as is for now so we can get this merged and revisit this sometime in the future.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4846648

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4846934

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4847770

cumulus/parachains/runtimes/coretime/coretime-rococo/src/weights/frame_system_extensions.rs
```diff
@@ -0,0 +1,121 @@
...
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 561_000 picoseconds.
+		Weight::from_parts(2_705_000, 0)
+			.saturating_add(Weight::from_parts(0, 0))
+	}
+	fn check_nonce() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
```
ggwpez, line 86: Weird that it does not have a proof weight here. Since it should access the `System.:Account` map.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -176,6 +176,7 @@ impl frame_system::Config for Runtime {
 	type Version = Version;
 	type AccountData = pallet_balances::AccountData<Balance>;
 	type SystemWeightInfo = weights::frame_system::WeightInfo<Runtime>;
+	type SystemExtensionsWeightInfo = weights::frame_system_extensions::WeightInfo<Runtime>;
```
ggwpez, line 179: DQ but why are they called `SystemExtensions` and not `TransactionExtensions`?

substrate/frame/system/src/extensions/check_genesis.rs
```diff
@@ -59,6 +59,10 @@ impl<T: Config + Send + Sync> TransactionExtensionBase for CheckGenesis<T> {
 	fn implicit(&self) -> Result<Self::Implicit, TransactionValidityError> {
 		Ok(<Pallet<T>>::block_hash(BlockNumberFor::<T>::zero()))
 	}
+	fn weight(&self) -> sp_weights::Weight {
+		use super::WeightInfo;
+		T::SystemExtensionsWeightInfo::check_genesis()
```
ggwpez, line 64:
```suggestion
		<T::SystemExtensionsWeightInfo as super::WeightInfo>::check_genesis()
```

substrate/frame/system/benchmarking/src/extensions.rs
```diff
@@ -0,0 +1,213 @@
...
+		let len = 0_usize;
+		let ext = CheckNonce::<T>::from(1u32.into());
+		let info = DispatchInfo { weight: Weight::zero(), ..Default::default() };
+		let call: T::RuntimeCall = frame_system::Call::remark { remark: vec![] }.into();
+
+		#[block]
+		{
+			ext.test_run(RawOrigin::Signed(caller.clone()).into(), &call, &info, len, |_| {
+				Ok(().into())
+			})
```
ggwpez, line 125:
```suggestion
			ext.test_run(RawOrigin::Signed(caller.clone()).into(), &call, &info, len, |_| Ok(().into()))
```

substrate/frame/system/src/lib.rs
```diff
@@ -347,6 +348,9 @@ pub mod pallet {
 			/// Weight information for the extrinsics of this pallet.
 			type SystemWeightInfo = ();
 
+			/// Weight information for the extensions of this pallet.
```
ggwpez, line 351: This line of docs would be more helpful down below in the `trait Config`.

substrate/frame/system/src/lib.rs
```diff
@@ -550,6 +554,8 @@ pub mod pallet {
 
 		type SystemWeightInfo: WeightInfo;
```
ggwpez, line 555: This also has no comment 🤔

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,140 @@
...
+use frame_benchmarking::v2::*;
+use frame_support::{
+	dispatch::{DispatchInfo, PostDispatchInfo},
+	pallet_prelude::*,
+	traits::fungibles::Inspect,
+};
+use frame_system::RawOrigin;
+use sp_runtime::traits::{AsSystemOriginSigner, DispatchTransaction, Dispatchable};
+
+pub trait ExtConfig: Config {
```
ggwpez, line 31:
```suggestion
pub trait BenchConfig: Config {
```
Or something similar is more expressive IMHO.

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,140 @@
...
+		<<Self as Config>::OnChargeAssetTransaction as OnChargeAssetTransaction<Self>>::AssetId,
+	);
+	fn setup_balances_and_pool(
+		asset_id: <<Self as Config>::Fungibles as Inspect<Self::AccountId>>::AssetId,
+		account: Self::AccountId,
+	);
+}
+
+#[benchmarks(where
+	T: Config + Send + Sync + ExtConfig,
```
ggwpez, line 45:
```suggestion
	T: Send + Sync + ExtConfig,
```
Already implied.

substrate/frame/transaction-payment/asset-tx-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,135 @@
...
+		<<Self as Config>::OnChargeAssetTransaction as OnChargeAssetTransaction<Self>>::AssetId,
+	);
+	fn setup_balances_and_pool(
+		asset_id: <<Self as Config>::Fungibles as Inspect<Self::AccountId>>::AssetId,
+		account: Self::AccountId,
+	);
+}
+
+#[benchmarks(where
+	T: Config + Send + Sync + ExtConfig,
```
ggwpez, line 45: ditto

substrate/frame/transaction-payment/asset-tx-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,135 @@
...
+use frame_benchmarking::v2::*;
+use frame_support::{
+	dispatch::{DispatchInfo, PostDispatchInfo},
+	pallet_prelude::*,
+	traits::fungibles::Inspect,
+};
+use frame_system::RawOrigin;
+use sp_runtime::traits::{AsSystemOriginSigner, DispatchTransaction, Dispatchable};
+
+pub trait ExtConfig: Config {
```
ggwpez, line 31: ditto

substrate/frame/transaction-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,81 @@
...
+fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
+	let events = frame_system::Pallet::<T>::events();
+	let system_event: <T as frame_system::Config>::RuntimeEvent = generic_event.into();
+	// compare to the last event record
+	let EventRecord { event, .. } = &events[events.len() - 1];
+	assert_eq!(event, &system_event);
+}
+
+#[benchmarks(where
+	T: Config,
```
ggwpez, line 36: Why do we not need the ugly `Send + Sync` here?

substrate/frame/transaction-payment/src/payment.rs
```diff
@@ -191,4 +197,14 @@ where
 		}
 		Ok(())
 	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	fn endow_account(who: &T::AccountId, amount: Self::Balance) {
+		let _ = C::deposit_creating(who, amount);
```
ggwpez, line 203:
```suggestion
		assert!(C::deposit_creating(who, amount).peek() != amount, "Could not endow account.");
```
The benchmark tests actually fail like this 🤔

substrate/utils/frame/benchmarking-cli/src/pallet/template.hbs
```diff
@@ -22,7 +22,11 @@ use core::marker::PhantomData;
 
 /// Weight functions for `{{pallet}}`.
 pub struct WeightInfo<T>(PhantomData<T>);
+{{#if (eq pallet "frame_system_extensions")}}
```
ggwpez, line 25: Looks a bit hacky, see my review comment

ggwpez (commented):
Code looks good.

Just about the design: why did you opt to create a new `SystemExtensionsWeightInfo` instead of just making it part of the `WeightInfo` of `frame-system`?  
It causes a lot of changes (also breaking) and i dont really see the advantage.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
georgepisaltu: It's not really `SystemExtensions`, it's more like `System Extensions WeightInfo`, if that makes sense. Could have been `SystemTransactionExtensionsWeightInfo` but it felt too long.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/weights/pallet_transaction_payment.rs
```diff
@@ -0,0 +1,67 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Autogenerated weights for `pallet_transaction_payment`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2023-12-21, STEPS: `2`, REPEAT: `2`, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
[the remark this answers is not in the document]
georgepisaltu: There are a bunch of weight files which were just copied, some of them are outdated as well. Everything will need to be benched again before a release.

georgepisaltu:
> Just about the design: why did you opt to create a new `SystemExtensionsWeightInfo` instead of just making it part of the `WeightInfo` of `frame-system`? It causes a lot of changes (also breaking) and i dont really see the advantage.

I actually think it's cleaner having them separate since the code being benchmarked in extensions does very different things compared to the system pallet, even though they both depend on `frame_system::Config`. Main advantage is this way you can easily choose to ignore extension weights, which would be backwards compatible with what we have now, and have real weights for the rest of the system stuff. Given how the current `frame_system` weights are implemented now, with the benches in a different crate, I think it fits fine.

substrate/utils/frame/benchmarking-cli/src/pallet/template.hbs
georgepisaltu: It doesn't feel hacky to me, it's basically the same as the model for the `frame_system` weights in the other templates. They also get special treatment because of the name, like in `substrate/.maintain/frame-weight-template.hbs`.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
ggwpez: `SystemWeightInfo` is already a misnomer since we normally use `WeightInfo` so maybe `ExtrnsionsWeightInfo` is enough?

substrate/frame/transaction-payment/src/payment.rs
georgepisaltu: Isn't it supposed to be `assert!(C::deposit_creating(who, amount).peek() == amount, "Could not endow account.");` in your example? As far as I understand, if `deposit_creating` succeeded, the returned positive imbalance would be equal to the `amount` parameter, as in the entire amount was minted successfully.

polkadot/runtime/common/src/claims.rs
```diff
@@ -591,6 +595,12 @@ impl<T: Config> Pallet<T> {
 	}
 }
 
+#[cfg(feature = "runtime-benchmarks")]
+pub trait ExtConfig: Config {
```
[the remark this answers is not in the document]
georgepisaltu: Done.

polkadot/runtime/common/src/claims.rs
```diff
@@ -1521,6 +1548,11 @@ mod benchmarking {
 	}
 
 	benchmarks! {
+		where_clause { where <T as frame_system::Config>::RuntimeCall: IsSubType<Call<T>>,
+			T: Config + Send + Sync + ExtConfig,
```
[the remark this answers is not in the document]
georgepisaltu: We need it for extensions holding a `PhantomData`, but not for the other ones. I removed it where applicable.

substrate/frame/system/src/extensions/check_genesis.rs
georgepisaltu: Done.

substrate/frame/system/benchmarking/src/extensions.rs
georgepisaltu: My formatter wants to keep it as is. Are you using nightly on your `fmt`?

substrate/frame/system/src/lib.rs
georgepisaltu: Done.

substrate/frame/system/src/lib.rs
georgepisaltu: Fixed. Also added one to the line above it.

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/benchmarking.rs
georgepisaltu: Renamed.

substrate/frame/transaction-payment/asset-conversion-tx-payment/src/benchmarking.rs
georgepisaltu: Removed.

substrate/frame/transaction-payment/src/benchmarking.rs
georgepisaltu: Because it doesn't have a `PhantomData`.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
georgepisaltu: Renamed all instances to `SystemWeightInfo`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4850950

substrate/frame/transaction-payment/src/payment.rs
ggwpez: Oh yea, i was playing around with the code before. 👍

substrate/frame/system/benchmarking/src/extensions.rs
ggwpez: I use the same version as the CI, currently: `cargo +nightly-2023-11-01 fmt`.

substrate/utils/frame/benchmarking-cli/src/pallet/template.hbs
ggwpez: Yea but its not neccecary here since you can move the TE benchmarks to a `frame-system-ext-benchmarks` crate.

ggwpez (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4882504

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4883968

PR #2735, opened 2023-12-18
relaxing timepoint constraint for multisig

### LINKED ISSUE #2541: Improve developer experience by removing `timepoint` from `multisig`

ozgunozerk:
# Description

This PR aims to improve developer experience by solving #2541. A detailed explanation is provided in the issue itself.

- What does this PR do?
    Relaxes the `timeout` parameter for `multisig pallet` as suggested by @gavofyork 
    It will significantly improve the developer experience
    This won't be a breaking change :)

Fixes #2541 

# Checklist

- [x] My PR follows the [labeling requirements](CONTRIBUTING.md#Process) of this project (at minimum one label for `T`
  required) **_(it should probably be labeled with T2 for this)_**
- [x] I have made corresponding changes to the documentation
- [x] I have added related tests


xlc (commented):
please avoid tx version breaking change. you can add a new extrinsic and keep the old one remain as it is.

ozgunozerk:
Hi @xlc , just to be sure, you meant `cancel_as_multi`?

xlc:
yes. any changes to existing extrinsic signature is tx breaking change

ozgunozerk:
1. implemented a new extrinsic called `cancel_as_multi_without_timepoint`.
2. extracted the reusable code to a private function, to reduce code duplication between `cancel_as_multi` and `cancel_as_multi_without_timepoint`
3. added benchmark item for `cancel_as_multi_without_timepoint`
4. modified tests accordingly

substrate/frame/multisig/src/lib.rs
```diff
@@ -469,29 +470,38 @@ pub mod pallet {
...
+		///
+		/// ## Complexity
+		/// - `O(S)`.
+		/// - Up to one balance-reserve or unreserve operation.
+		/// - One passthrough operation, one insert, both `O(S)` where `S` is the number of
+		///   signatories. `S` is capped by `MaxSignatories`, with weight being proportional.
+		/// - One encode & hash, both of complexity `O(S)`.
+		/// - One event.
+		/// - I/O: 1 read `O(S)`, one remove.
+		/// - Storage: removes one item.
```
kianenigma, line 494:
```suggestion
		/// Same as [`Pallet::cancel_as_multi`], but with an optional timepoint
		///
		/// <explain the difference risk factors here>
```

kianenigma (approved):
LGTM, needs audit and a clear documentation both on the method and the pallet top level docs explaining the tradeoff of what you gain by using the timepoint and what not. In other words, the scenario explained in the issue.

substrate/frame/multisig/src/lib.rs
ozgunozerk: Tried to address this. Let me know if further clarification/elaboration is wanted

ozgunozerk:
@kianenigma @xlc, anything left for this PR that you want me to do? Or are we just waiting for the audit?

gupnik (changes requested):
I was expecting that we would modify the `MultiSig` struct to have `when` as optional. Then, we would ensure that it is mandatory to provide the timepoint if it was `Some` at the time of creation of the multisig.

The current change allows users to pass None for those multisigs as well that might want to enforce a timepoint check.

ozgunozerk:
@gupnik I agree, I think your suggestion will be quite beneficial for the developers. Implemented the suggested way.

Before I also modify the tests accordingly, I wanted to get confirmation on my assumptions/choices.

**Problem:** during the creation of the multisig, how can the creator (the first approver) can select whether the multisig will be with or without timepoint?

**Solutions:**
- add a boolean flag for this to `approve_as_multi` and `as_multi` functions
    - did not go with this one, since it will be a breaking change, and will be useless for further approvals. I thought this would unnecessarily complicate the UX.
- use the current `maybe_timepoint` for this flag (whether to enable timepoint or not during the creation of the multisig). 
    - went with this approach
    - supplying `Some(timepoint)` means that the creator wants to enable `timepoint` feature for this multisig
    - supplying  `None` means that the creator wants to disable `timepoint` feature for this multisig

However, there are some _important_ details:
1. supplying `Some(timepoint)` for the creation -> the actual value for `timepoint` does not serve any purpose, and it can be anything. Because, during creation, we will use `Self::timepoint()` function to get the timepoint. I've explained this in inline documentation, so I think it won't be confusing for those who read the description of the function.
2. I wanted to somehow allow the creator to not supply any timepoint, yet able to enable the `timepoint` security measure, but without introducing a breaking change, I couldn't do it.
3. since we may get a `timepoint` even during the creation now, the previous `UnexpectedTimeout` error is now gone. I don't know if this counts towards introducing a **_breaking change_**. 
4. everything else now should work as described in the suggestion. If the multisig is created with a timepoint, we require the following approvals to provide a timepoint as well. If the multisig is created without a timepoint, the timepoint field in the following approvals will be ignored, so it can be anything, we don't care.

If you confirm all the above, I can proceed to updating tests accordingly.

gupnik:
> use the current `maybe_timepoint` for this flag (whether to enable timepoint or not during the creation of the multisig).
> * went with this approach

Makes sense 👍 

> 1. supplying `Some(timepoint)` for the creation -> the actual value for `timepoint` does not serve any purpose, and it can be anything. Because, during creation, we will use `Self::timepoint()` function to get the timepoint. I've explained this in inline documentation, so I think it won't be confusing for those who read the description of the function.

Yeah, I don't see any issue with this, given that previously we were enforcing this as `None` during creation only via the docs.

> 3\. since we may get a `timepoint` even during the creation now, the previous `UnexpectedTimeout` error is now gone. I don't know if this counts towards introducing a **_breaking change_**.

This should be fine.

Thanks @ozgunozerk. I believe we would also need a storage migration here.

ozgunozerk:
Completed implementing migrations for this change. It's my first time writing a migration, so better to double check 😅 
Will update tests soon (today or tomorrow)

ozgunozerk:
Done!

substrate/frame/multisig/src/migrations.rs
```diff
@@ -85,3 +88,119 @@ pub mod v1 {
 		}
 	}
 }
+
```
gupnik, line 91: We should use `VersionedMigration` here. @liamaharon can provide guidance if needed.

gupnik (approved):

ozgunozerk:
@liamaharon 
I tried to implement multi-block migrations as presented in the example you provided.

There are few obstacles I've stumbled upon, they are all marked with `TODO:` in the codebase.
Please find them in the review part, I'll explicitly comment them as self-review

substrate/frame/multisig/src/migrations/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// # Multi-Block Migrations Module
+///
+/// This module showcases a simple use of the multi-block migrations framework.
+pub mod v2;
```
ozgunozerk, line 21: I only applied multi-block migration approach to the new migration logic (v2). There was an already existing migration logic for v0 to v1, but I did not touch it, since I think it's outside the scope of this PR.

substrate/frame/multisig/src/migrations/v2/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+		pub deposit: Balance,
+		/// The account who opened it (i.e. the first to approve it).
+		pub depositor: AccountId,
+		/// The approvals achieved so far, including the depositor. Always sorted.
+		pub approvals: BoundedVec<AccountId, MaxApprovals>,
+	}
+
+	#[storage_alias]
+	pub type Multisigs<T: Config> = StorageDoubleMap<
+		_, // TODO: empty prefix is not allowed, what value should be supplied here?
```
ozgunozerk, line 59: with `#[storage_alias]` macro, empty prefix was not allowed, I don't know what to supply there or does it even matter?

substrate/frame/multisig/src/migrations/v2/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+		T::AccountId,
+		Blake2_128Concat,
+		[u8; 32],
+		OldMultisig<BlockNumberFor<T>, BalanceOf<T>, T::AccountId, T::MaxSignatories>,
+	>;
+}
+
+use crate::MaxEncodedLen;
+use crate::{Decode, Encode};
+// TODO: did not want to touch frame/support/src/migrations.rs
```
ozgunozerk, line 70: in the examples, `MigrationId` was there in `frame/support/src/migrations.rs`. I didn't modify that file for granular and small-scoped PR.

substrate/frame/multisig/src/migrations/v2/weights.rs
```diff
@@ -0,0 +1,84 @@
+// This file is part of Substrate.
```
ozgunozerk, line 1: did not touch this file at all, don't know how to generate weights for this purpose

substrate/frame/multisig/src/lib.rs
```diff
@@ -469,29 +503,23 @@ pub mod pallet {
...
-			Self::deposit_event(Event::MultisigCancelled {
-				cancelling: who,
-				timepoint,
-				multisig: id,
-				call_hash,
-			});
-			Ok(())
+		/// Same as [`Pallet::cancel_as_multi`], but without a timepoint.
+		///
+		/// If the multisig is created with a `timepoint`, use [`Pallet::cancel_as_multi`],
```
ggwpez, line 511: Why would the timepoint not be passed as `Option`?

substrate/frame/multisig/src/migrations/v2/mod.rs
ggwpez: `Pallet<T>`

substrate/frame/multisig/src/migrations/v2/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+// so I moved it here, please double check
+#[derive(MaxEncodedLen, Encode, Decode)]
+pub struct MigrationId<const N: usize> {
```
ggwpez, line 73: Why do you need to copy the definition? You can just import it?

substrate/frame/multisig/src/lib.rs
ozgunozerk: https://github.com/paritytech/polkadot-sdk/pull/2735/files#diff-a99723ce864d6d9e58c74c4a3246a8c9ba573564258190a799d65496537b8880R483-R484

I'd gladly turn it to `Option`, but then it would be a breaking change?

substrate/frame/multisig/src/migrations/v2/weights.rs
ggwpez: Yea we can re-generate them as last step.

substrate/frame/multisig/src/migrations/v2/mod.rs
ozgunozerk: there is no `MigrationId` in that location (nothing to import), and I did not want to add that struct to `frame/support/src/migrations.rs/migration.rs` to keep the PR granular

substrate/frame/multisig/src/migrations/mod.rs
ggwpez: The existing `MigrateToV1` is written in a way that will not make it work anymore, so i *guess* its fine to delete it.  
Its also 1,5 yrs old, so hopefully most people updates already.

substrate/frame/multisig/src/lib.rs
ozgunozerk: is there a consensus on this? Are we introducing a breaking change, or keeping things as is?

substrate/frame/multisig/src/migrations/v2/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+				// We can just insert here since the old and the new map share the same key-space.
+				// Otherwise it would have to invert the concat hash function and re-hash it.
+				Multisigs::<T>::insert(last_key1, last_key2, new_multisig);
+				cursor = Some((last_key1, last_key2)) // Return the processed key as the new cursor.
+			} else {
+				cursor = None; // Signal that the migration is complete (no more items to process).
+				break;
+			}
+		}
+		Ok(cursor)
```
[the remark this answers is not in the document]
ozgunozerk: any update on this @liamaharon ?

ozgunozerk:
@liamaharon @ggwpez pinging once more in case notification got lost in the inbox

substrate/frame/multisig/src/migrations/v2/mod.rs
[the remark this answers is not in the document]
liamaharon: StorageVersion::new(2).put::<crate::Pallet<T>>();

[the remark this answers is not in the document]
ozgunozerk: @liamaharon the problem I'm facing is I don't know where to put `StorageVersion::new(2).put::<crate::Pallet>();`. It should be put as the last step of the migration (if migration succeeded). 

- I couldn't find a relevant function in the trait `SteppedMigration`. 
- Putting it in `step()` function seems counter intuitive because it will be called multiple times, and may not be finalized.
- tried to track down the initial caller of `step()` function for the migration, but failed to do so. See my effort below in the details.

<details>

`SteppedMigration::step()` <- `SteppedMigration::transactional_step()` <- `SteppedMigrations::nth_transactional_step()` <- `exec_migration()` <- `progress_mbms()` <- `MultiStepMigrator::step()` <- ???

Should I also implement `MultiStepMigrator` for the pallet? Even then, which code piece is actually triggering the migration part? I'm a bit confused at this point.

</details>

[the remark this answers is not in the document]
liamaharon: Can you put it immediately before you set the cursor to None?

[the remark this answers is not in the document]
liamaharon: > Should I also implement MultiStepMigrator for the pallet? Even then, which code piece is actually triggering the migration part? I'm a bit confused at this point.

Not required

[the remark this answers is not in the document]
ozgunozerk: > Can you put the new storage version immediately before you set the cursor to None?
yeah makes sense 🤦 that place makes sure migration is finished, thanks for pointing it out! I was looking for it in the wrong places

ozgunozerk:
@ggwpez should we progress into generating weights?

substrate/frame/multisig/src/lib.rs
```diff
@@ -109,7 +124,9 @@ where
 	MaxApprovals: Get<u32>,
 {
 	/// The extrinsic when the multisig operation was opened.
-	when: Timepoint<BlockNumber>,
+	/// `timepoint` is an optional extra security measure, which can be enabled by the creator of
+	/// the multisig during the creation by supplying a `timepoint` for the `maybe_when` field.
```
ggwpez, line 128: Personally i think that the level of security should be decided upon by the approvers - not the creator.  
Now the creator can weaken the security of the other participants by using `None`.  

When the timepoint is always present, then the approvers can decide ad-hoc if they want to approve just this single call by sending `Some(_)`, or if they want to also approve all future versions of this by using `None`.  
This is very similar to immortal transactions.

substrate/frame/multisig/src/lib.rs
ggwpez: Okay new function is fine.

substrate/frame/multisig/src/lib.rs
Tbaut: As a UI dev using these pallets a lot, this sounds like worsening the DX:
- I need to understand all this, when I didn't before
- I need to take a decision for my users whether to pass the time-point or not, or pass this complexity down to my users by having a checkbox, and an explanation.
- the default, and easiest route, not caring about this, is not the safest unfortunately

Zooming out, this PR was supposed to make DX better. My opinion is that it's not the case. As an implementer, I used to send `None` with the first tx, and managed to find the timepoint for the subsequent approvals, it 1 call to make. It was actually easier than having to think about all this IMHO.

ozgunozerk: @Tbaut, valid points, I mostly agree, but mostly. 

Here is my take (from the issue #2541):
>
>1. providing the timepoint is an arduous task. Let me elaborate: 
>    - no function is exposed in the library for finding the `timepoint` of a multisig operation. 
>    - see this example for how many steps are required for finding the `timepoint`: https://github.com/paritytech/txwrapper-core/tree/768bb445beb2907582b2d5e13ade3be5d995af3e/packages/txwrapper-examples/multisig#timepoint-of-approveasmulti--storage
>    - this complexity is probably discouraging developers from using multisig pallet and substrate
>2. it is not obvious what `timepoint`s purpose is in the pallet. It is confusing for most of the developers to encounter `timepoint`

In short, with the current state of this PR, we achieve the following:
- purpose of the `timepoint` is clearly explained.
- BIG WIN: approving a multisig way easier (due to `timepoint` is not a necessity anymore for the most use cases)

--- 

As to your points:
> I need to understand all this, when I didn't before.

I'd argue against you have to understand more. This PR does not introduce a new concept. `Timepoint` was already there, and the purpose of the `timepoint` has not changed. We are just making it optional because for the most multisigs, timepoint is an overkill. In other words, this PR just makes the purpose of everything clear imo.

> I need to take a decision for my users whether to pass the time-point or not, or pass this complexity down to my users by having a checkbox, and an explanation -> worse UX.

Yes, you need to. But examples and reasonings behind those are explained clearly in the docs. I'd again argue against worse UX, because previously we still had to provide timepoint, and the reason for that was not clear. The explanation for the `timepoint` should have been there in the UX in the first place imo.

> the default, and easiest route, not caring about this, is not the safest unfortunately

Yes, if anyone has a suggestion on the defaults, I'm all ears.

> Zooming out, this PR was supposed to make DX better. 

I believe it still does, a lot!

> As an implementer, I used to send None with the first tx, and managed to find the timepoint for the subsequent approvals, it's 1 call to make. It was actually easier than having to think about all this IMHO.

If you don't care about the `timepoint`, you can still send `None` and everybody else will send `None`. Previously, this was not possible, even for the multisigs that do not need `timepoint`, every approver had to provide `timepoint` and finding the `timepoint` is not easy as I explained above from the perspective of the approvers.

---

From my perspective, when zoomed out:
- the functionality of this pallet is much more clear.
- for the majority of the use cases, approvers will have a much better time.

Tbaut: I should probably have said it from the beginning. I think this first assumption, that it's hard to get the timepoint of a multisig tx, is IMHO, wrong. Maybe because of the api you're using. It's literally one state call, that you must do in any case to know the currently pending multisig txs. Then the info is in the storage, here's an example of a pending multisig tx:
image

> Yes, you need to. [decide or pass down this complexity to users]

Let's not confuse things. DX is my experience implementing this, UX is the experience of users of my application. I want to build applications that have good UX, and I don't want to feature creep them by adding options that users have no idea about. I should take the best decision for them. If the best decision is the safest decision at no additional cost for my users, and a minimal cost for me (my DX, my time) then I should include the timepoint... because.. it isn't hard IMHO.

Now I'm not willing to die on this hill, I just left this comment to give my opinion. To me it's a feature creep. I've implemented this in JS, there's been millions of hard things, retrieving the timepoint wasn't one of them. It's `api.query.multisig.multisigs.entries(address)` -> iterate over the stroage, then `storage[1].toJson().when`. I may be missing something, it may be hard in other languages, but it would mean that getting the pending multisig txs is also hard to start with.

ozgunozerk: I understand your concerns. I just wanted to fully explain my reasoning above, that's all!
Thanks for the conversation, appreciated 👍 🙏 

This PR is old, and the related issue was opened when I first tried to understand the `multisig` pallet back then. I still think `timepoint` as is, will be confusing for newcomers. And maybe I don't know how to do it, but from what I recall, one cannot easily approve a multisig in a browser (polkadotjs app). As you mentioned, they have to execute some code afaik:

> It's api.query.multisig.multisigs.entries(address) -> iterate over the stroage, then storage[1].toJson().when.

And I just think this requirement shouldn't be there in the first place for the majority of the multisigs (or at least, there should be an explanation on the UI side on why timepoint is required).

But after all, these are my subjective takes :)

Tbaut: > one cannot easily approve a multisig in a browser (polkadotjs app)

You're mistaking. UIs have all they need to find and fill the timepoint. If Pjs apps isn't good at something, it's not with the timepoint, it's with the `callData`. Which is a whole other beast. And this is worked around by many UIs including one I build, Multix, [more info about this workaround here](https://github.com/ChainSafe/Multix/wiki/Multix-integration). Now this issue is the subject of an RFC: https://github.com/polkadot-fellows/RFCs/pull/74

PR #2741, opened 2023-12-18
SDK docs ref cli

juangirini:
Closes https://github.com/paritytech/polkadot-sdk-docs/issues/53

### DIFF
--- a/docs/sdk/src/reference_docs/cli.rs
+++ b/docs/sdk/src/reference_docs/cli.rs
@@ -47,7 +47,7 @@
 //! - `--raw`: Converts the chain specification into a raw format with encoded storage keys.
 //! - `> customSpecRaw.json`: Outputs to customSpecRaw.json.
 //!
-//! #### Starting the First Node (Bootnode) in a Private Network
+//! #### Starting the First Node in a Private Network
 //! ```bash
 //! ./target/debug/node-template \
 //!   --base-path /tmp/node01 \
@@ -58,8 +58,7 @@
 //!   --telemetry-url "wss://telemetry.polkadot.io/submit/ 0" \
 //!   --validator \
 //!   --rpc-methods Unsafe \
-//!   --name MyNode01 \
-//!   --password-interactive
+//!   --name MyNode01
 //! ```
 //!
 //! - `--base-path`: Sets the directory for node data.
@@ -71,7 +70,6 @@
 //! - `--validator`: Indicates the node’s participation in block production.
 //! - `--rpc-methods Unsafe`: Allows potentially unsafe RPC methods.
 //! - `--name`: Sets a human-readable name for the node.
-//! - `--password-interactive`: Interactive password input for key operations.
 //!
 //! #### Adding a Second Node to the Network
 //! ```bash
@@ -93,7 +91,8 @@
 //! - `--rpc-port`: HTTP port for JSON-RPC.
 //! - `--telemetry-url`: Endpoint for sending telemetry data.
 //! - `--validator`: Indicates the node’s participation in block production.
-//! - `--bootnodes`: Specifies the address of the first node (bootnode) for peer discovery.
+//! - `--bootnodes`: Specifies the address of the first node for peer discovery. Nodes should find
+//!   each other using mDNS. This command needs to be used if they don't find each other.
 //!
 //! ---
 //!
--- a/docs/sdk/src/reference_docs/cli.rs
+++ b/docs/sdk/src/reference_docs/cli.rs
@@ -1,8 +1,8 @@
 //! # Substrate CLI
 //!
 //! Let's see some examples of typical CLI arguments used when setting up and running a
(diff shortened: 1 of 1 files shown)


gpestana (approved):

docs/sdk/src/reference_docs/cli.rs
```diff
@@ -1,7 +1,105 @@
...
+//! - `--base-path`: Sets the directory for node data.
+//! - `--chain`: Specifies the chain specification file.
+//! - `--port`: TCP port for peer-to-peer communication.
+//! - `--ws-port`: WebSocket port for RPC.
+//! - `--rpc-port`: HTTP port for JSON-RPC.
+//! - `--telemetry-url`: Endpoint for sending telemetry data.
+//! - `--validator`: Indicates the node’s participation in block production.
+//! - `--rpc-methods Unsafe`: Allows potentially unsafe RPC methods.
+//! - `--name`: Sets a human-readable name for the node.
+//! - `--password-interactive`: Interactive password input for key operations.
```
bkchr, line 74:
```suggestion
```

Not sure why you should mention this at the very beginning.

docs/sdk/src/reference_docs/cli.rs
```diff
@@ -1,7 +1,105 @@
...
+//! ```bash
+//! ./target/release/node-template \
+//!   --base-path /tmp/bob \
+//!   --chain local \
+//!   --bob \
+//!   --port 30334 \
+//!   --rpc-port 9946 \
+//!   --telemetry-url "wss://telemetry.polkadot.io/submit/ 0" \
+//!   --validator \
+//!   --bootnodes /ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp
```
bkchr, line 86: Nodes should find each other using mdns. Maybe we should only say that this command needs to be used if they don't find each other.

docs/sdk/src/reference_docs/cli.rs
```diff
@@ -1,7 +1,105 @@
...
+//!   --base-path /tmp/node01 \
+//!   --chain ./customSpecRaw.json \
+//!   --port 30333 \
+//!   --ws-port 9945 \
+//!   --rpc-port 9933 \
+//!   --telemetry-url "wss://telemetry.polkadot.io/submit/ 0" \
+//!   --validator \
+//!   --rpc-methods Unsafe \
+//!   --name MyNode01 \
+//!   --password-interactive
```
bkchr, line 62:
```suggestion
```

docs/sdk/src/reference_docs/cli.rs
```diff
@@ -1,7 +1,105 @@
...
+//! --disable-default-bootnode \
+//! > customSpecRaw.json
+//! ```
+//!
+//! - `--chain=customSpec.json`: Uses the custom chain specification as input.
+//! - `--disable-default-bootnode`: Disables the default bootnodes in the node template.
+//! - `--raw`: Converts the chain specification into a raw format with encoded storage keys.
+//! - `> customSpecRaw.json`: Outputs to customSpecRaw.json.
+//!
+//! #### Starting the First Node (Bootnode) in a Private Network
```
bkchr, line 50:
```suggestion
//! #### Starting the First Node in a Private Network
```

If you want to mention boot node, I would do it somewhere below in a sentence

bkchr (approved):
Some nitpicks. Otherwise looks good for a broad overview.

docs/sdk/src/reference_docs/cli.rs
liamaharon, line 74: Why do we document these here? Aren't they documented in the cli rustdoc?

liamaharon (approved):

docs/sdk/src/reference_docs/cli.rs
```diff
@@ -1,7 +1,104 @@
...
+//! - `--help`: Displays the available CLI arguments.
 //!
-//! - Command line arguments of a typical substrate based chain, and how to find and learn them.
-//! - How to extend them with your custom stuff.
+//! #### Starting a Local Substrate Node in Development Mode
+//! ```bash
+//! ./target/release/node-template \
+//! --dev
+//! ```
+//! - `--dev`: Runs the node in development mode, using a pre-defined development chain
```
kianenigma, line 18: Should link to [this](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/cli/src/params/shared_params.rs#L40) Rust type to avoid any repetition or inaccurate info. 

In other words, there should only be one source of truth for what `--dev` means, which is [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/cli/src/params/shared_params.rs#L40). A short paraphrasing is okay, but let's link to it. 

Same can be applied for technically every other arg mentioned here. Please make sure at least the important ones are linked to the proper Rust type, and that the Rust type is well documented.

PR #2760, opened 2023-12-19
use a single source for simple-mermaid dependency

andresilva:
this breaks cargo vendor which is necessary for building polkadot with nix


andresilva:
I will probably backport this to the v1.5.0 release branch, otherwise can't update the polkadot nixpkgs package.

bkchr (approved):

dmitry-markin (approved):

PR #2763, opened 2023-12-20
Cleanup bridges tests: with-grandpa-chain case

svyatonik:
related to https://github.com/paritytech/parity-bridges-common/issues/2739


bridges/primitives/runtime/src/chain.rs
```diff
@@ -199,7 +199,7 @@ pub trait Chain: Send + Sync + 'static {
 }
 
 /// A trait that provides the type of the underlying chain.
-pub trait UnderlyingChainProvider {
+pub trait UnderlyingChainProvider: Send + Sync + 'static {
```
bkontur, line 202: @svyatonik just out of curiosity, what exactly was the reason for adding `Send + Sync + 'static`?

bkontur (approved):

bridges/primitives/runtime/src/chain.rs
svyatonik: That's to avoid duplicating those lines:
```rust
	<MB as MessageBridge>::BridgedChain: Send + Sync + 'static,
	<MB as MessageBridge>::ThisChain: Send + Sync + 'static,
```
Which in turn are required, because `Chain` needs to be `Send + Sync + 'static` and when we use `UnderlyingChainProvider` we are implementing all `Chain` associated types, but we can't guarantee that it'll be `Send + Sync + 'static`. Alternative would be to `impl<T: UnderlyingChainProvider + Send + Sync + 'statuc> Chain for T { ... }`, but imo it is better

bkontur: aha, ok, thank you for explanation :)

georgepisaltu (approved):

PR #2777, opened 2023-12-21
Try to set `beacon-spec-mainnet` as default on runtime and not binary

bkontur:


bkontur:
@claravanstaden @alistair-singh @vgeddes guys, please, check this

vgeddes:
I think the issue with this approach is that it enables `beacon-spec-mainnet` by default, but does not provide a straightforward way to disable it at compilation time on the the command-line.

For our local E2E testing, we do require that the feature is disabled.

That said, I have an idea that we can try perhaps we can use cargo [patches](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) to override the default features for the `snowbridge-ethereum-beacon-client` crate.

bkontur:
> I think the issue with this approach is that it enables `beacon-spec-mainnet` by default, but does not provide a straightforward way to disable it at compilation time on the the command-line.
> 
> For our local E2E testing, we do require that the feature is disabled.

I am almost 100% sure, that dependency should be "production-ready" by default and not "test-ready".

@alistair-singh @vgeddes 

Anyway, if you need something compile-time-dependent then `polkadot-parachain` binary is a wrong place - it could work locally or with zombienet, when runtime is compiled/bundled with `polkadot-parachain`. But for real/production deployment it won't work, because runtimes are compiled with `srtool` which does not use `polkadot-parachain` binary, e.g. for `polkadot-fellows`: https://github.com/polkadot-fellows/runtimes/releases/tag/v1.0.1
```
These runtimes were built with rustc 1.73.0 (cc66ad468 2023-10-03) using [srtool v0.12.0](https://github.com/paritytech/srtool)
```

---

> That said, I have an idea that we can try perhaps we can use cargo [patches](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) to override the default features for the `snowbridge-ethereum-beacon-client` crate.

If I saw correctly, the feature `beacon-spec-mainnet` just enables/disables some "constants". I think there is no need to use cargo patches.

You can check `polkadot-sdk` and `prod_or_fast!` macro with `feature = fast-runtime` and maybe use it instead `beacon-spec-mainnet`.
If it is works for you, we could add this `fast-runtime` to the `polkadot-parachain` (sure, not as a default) like `polkadot` binary has [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/Cargo.toml#L61C45-L61C45).

acatangiu (approved):

svyatonik (approved):

PR #2782, opened 2023-12-21
Duplicated runtime version in AH-Rococo runtime ?

al3mart:
This MR removes a presumably duplicated runtime version definition.

EDIT:
As @bkontur points out the version has `state_version: 0`, for `#[cfg(not(feature = "state-trie-version-1"))]`
I overlooked that `not` in the feature.
This PR can be closed

### DIFF
--- a/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
+++ b/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
@@ -118,19 +118,6 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	state_version: 1,
 };
 
-#[cfg(not(feature = "state-trie-version-1"))]
-#[sp_version::runtime_version]
-pub const VERSION: RuntimeVersion = RuntimeVersion {
-	spec_name: create_runtime_str!("statemine"),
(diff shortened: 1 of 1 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4787528

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -118,19 +118,6 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	state_version: 1,
 };
 
-#[cfg(not(feature = "state-trie-version-1"))]
```
bkontur, line 121: I am not sure if it is still needed, but I wouldn't say it is a duplicate, they have different `state_version`:
`state_version: 1,` for `#[cfg(feature = "state-trie-version-1")]`
`state_version: 0,` for `#[cfg(not(feature = "state-trie-version-1"))]`

al3mart:
Oh, you are right, I totally missed that `not` in the feature!

Cool, then maybe we can just close the PR

PR #2786, opened 2023-12-22
Update Safe Call Filters

joepetrowski:
- Updates all safe call filters to allow `system::authorize_upgrade`.
- Updates Coretime safe call filter to allow `sudo` and `system_set_storage`.
- Update Coretime teleports to allow teleportation with other system chains.

### DIFF
--- a/cumulus/parachains/runtimes/coretime/coretime-rococo/src/xcm_config.rs
+++ b/cumulus/parachains/runtimes/coretime/coretime-rococo/src/xcm_config.rs
@@ -28,7 +28,7 @@ use pallet_xcm::XcmPassthrough;
 use parachains_common::{
 	impls::ToStakingPot,
 	xcm_config::{
-		AllSiblingSystemParachains, ConcreteNativeAssetFrom, ParentRelayOrSiblingParachains,
+		AllSiblingSystemParachains, ConcreteAssetFromSystem, ParentRelayOrSiblingParachains,
 		RelayOrOtherSystemParachains,
 	},
 	TREASURY_PALLET_ID,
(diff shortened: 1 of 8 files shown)


cumulus/parachains/runtimes/coretime/coretime-rococo/src/xcm_config.rs
```diff
@@ -140,15 +140,22 @@ impl Contains<RuntimeCall> for SafeCallFilter {
...
+				pallet_xcm::Call::force_xcm_version { .. } |
+					pallet_xcm::Call::force_default_xcm_version { .. }
+			) | RuntimeCall::System(
+				frame_system::Call::set_heap_pages { .. } |
 						frame_system::Call::set_code { .. } |
 						frame_system::Call::set_code_without_checks { .. } |
-						frame_system::Call::kill_prefix { .. },
-				) | RuntimeCall::ParachainSystem(..) |
+						frame_system::Call::authorize_upgrade { .. } |
+						frame_system::Call::authorize_upgrade_without_checks { .. } |
```
seadanda, line 151: Should we be marking this one with a warning as well? Likewise in the other runtimes.

seadanda (approved):

bkchr (approved):

cumulus/parachains/runtimes/coretime/coretime-rococo/src/xcm_config.rs
joepetrowski: IMO no, the function itself has a warning. It does meet the criteria of being "safe" for the `SafeCallFilter`.

PR #2800, opened 2023-12-23
Ensure force_set_metadata controls if asset is live + code reformat to avoid code duplications

0xMenna01:
- The force_set_metadata function does not control whether the asset is live. 
- To avoid code duplication I reused the do_set_metadata function for both force_set_metadata and set_metadata, differentiating between origins.

### DIFF
--- a/substrate/frame/assets/src/functions.rs
+++ b/substrate/frame/assets/src/functions.rs
@@ -86,7 +86,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			// allow accidental usage of all consumer references which could cause grief.
 			if !frame_system::Pallet::<T>::can_inc_consumer(who) {
 				frame_system::Pallet::<T>::dec_consumers(who);
-				return Err(Error::<T, I>::UnavailableConsumer.into())
+				return Err(Error::<T, I>::UnavailableConsumer.into());
 			}
 			ExistenceReason::Consumer
 		};
@@ -133,24 +133,24 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			None => return DepositConsequence::UnknownAsset,
 		};
 		if increase_supply && details.supply.checked_add(&amount).is_none() {
-			return DepositConsequence::Overflow
+			return DepositConsequence::Overflow;
 		}
 		if let Some(account) = Account::<T, I>::get(id, who) {
 			if account.status.is_blocked() {
-				return DepositConsequence::Blocked
+				return DepositConsequence::Blocked;
 			}
 			if account.balance.checked_add(&amount).is_none() {
-				return DepositConsequence::Overflow
+				return DepositConsequence::Overflow;
 			}
 		} else {
 			if amount < details.min_balance {
-				return DepositConsequence::BelowMinimum
+				return DepositConsequence::BelowMinimum;
 			}
 			if !details.is_sufficient && !frame_system::Pallet::<T>::can_accrue_consumers(who, 2) {
-				return DepositConsequence::CannotCreate
+				return DepositConsequence::CannotCreate;
 			}
 			if details.is_sufficient && details.sufficients.checked_add(1).is_none() {
-				return DepositConsequence::Overflow
+				return DepositConsequence::Overflow;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


substrate/frame/assets/src/functions.rs
```diff
@@ -953,46 +953,59 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				Error::<T, I>::NoPermission
+			);
 
 			let old_deposit = metadata.take().map_or(Zero::zero(), |m| m.deposit);
-			let new_deposit = Self::calc_metadata_deposit(&name, &symbol);
 
-			if new_deposit > old_deposit {
-				T::Currency::reserve(from, new_deposit - old_deposit)?;
+			let deposit = if is_force_origin {
+				old_deposit
```
joepetrowski, line 982: Why?

substrate/frame/assets/src/types.rs
```diff
@@ -35,6 +35,31 @@ pub(super) type AssetAccountOf<T, I> = AssetAccount<
 pub(super) type ExistenceReasonOf<T, I> =
 	ExistenceReason<DepositBalanceOf<T, I>, <T as SystemConfig>::AccountId>;
 
+/// Whether the set metadata operation is coming from a `force_origin` or not.
+// If it's a force origin then `is_frozen` is provided for the asset as well.
+pub(super) enum MetadataRequestOf<AccountId> {
+	ForceOrigin(bool),
+	Origin(AccountId),
+}
```
joepetrowski, line 43: IMO this is tries to conflate too many things and doesn't express the information clearly (and the below `impl` likewise returns types that aren't intuitive to the user). I think it would be a better API to remove this type and change the signature of `do_set_metadata` to something like this, and leave it up to the caller to supply the correct arguments when calling it:
```rust
pub(super) fn do_set_metadata(
	id: T::AssetId,
	// `Some` if called by an origin that converts to an `AccountId`.
	// Must equal the asset class's `owner`. `None` if called by a
	// privileged function that bypasses any permissions (that is,
	// the origin has already been checked and has the privilege to
	// call this function).
	maybe_owner: Option<AccountId>,
	name: Vec<u8>,
	symbol: Vec<u8>,
	decimals: u8,
	// Whether or not the asset class is frozen.
	is_frozen: bool,
) -> DispatchResult 
```

joepetrowski (commented):
See comments, I think setting the metadata by force should also set the deposit to 0, such that it refunds an existing deposit (and if the owner sets new metadata, of course they'd have to put down the full deposit).

Also can you update formatting to our version (i.e. remove trailing `;` that you added)?

substrate/frame/assets/src/functions.rs
0xMenna01: That was just to reproduce the implementation already in place. If the call is made from a force origin, it used the old deposit.

0xMenna01:
> lso can you update formatting to our version (i.e. remove trailing `;` that you added)?

These suggestions would change the actual design of the set_metadata by force. If the force origin cannot change the original deposit by design, why do you think it should be set to zero and refund an existing deposit ?

substrate/frame/assets/src/functions.rs
joepetrowski: OK fair enough. I think with the force origin behind governance, giving it no deposit is sensible.

joepetrowski:
> > lso can you update formatting to our version (i.e. remove trailing `;` that you added)?
> 
> These suggestions would change the actual design of the set_metadata by force. If the force origin cannot change the original deposit by design, why do you think it should be set to zero and refund an existing deposit ?

I don't think that's to do with formatting? Anyway, the idea of the deposit is to avoid abusing some feature for "free" storage. But `force_` operations are not permissionless, they go through governance. So by nature using `force` methods achieves the goal of limiting resource use. If the asset owner comes along later and sets new metadata then they would need to pay the going deposit rate.

substrate/frame/assets/src/functions.rs
```diff
@@ -86,7 +86,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			// allow accidental usage of all consumer references which could cause grief.
 			if !frame_system::Pallet::<T>::can_inc_consumer(who) {
 				frame_system::Pallet::<T>::dec_consumers(who);
-				return Err(Error::<T, I>::UnavailableConsumer.into())
+				return Err(Error::<T, I>::UnavailableConsumer.into());
```
ggwpez, line 89: Please no unrelated format changes - they just slow down the review process.

bkchr:
Stale

PR #2803, opened 2023-12-24
`cumulus-primitives-parachain-inherent`: Split into two crates

bkchr:
This splits `cumulus-primitives-parachain-inherent` into two crates, the previous `cumulus-primitives-parachain-inherent` and a new `cumulus-client-parachain-inherent`. The idea behind this is to move the `create_at` logic into the client crate. This removes quite a lot of unrelated dependencies from the runtime std build and thus, makes the compilation faster. On my Laptop the compilation is goes down by one minute for `asset-hub-rococo-runtime`. I also assume that the full build of the entire workspace probably can be speed-up a little bit, because more stuff can be compiled in parallel.


bkchr:
bot fmt

michalkucharczyk (approved):

skunert (approved):

davxy (approved):

PR #2811, opened 2023-12-26
Broker pallet: fix interlacing

Szegoo:
With the current code, when a user interlaces their region, the end result will be three regions in the state:
- the non-interlaced region
- first part of the interlaced region
- second part of the interlaced region

The existing implementation retains the non-interlaced region in the state, leading to a problematic scenario:

1. User 1 acquires a region from the market.
2. User 1 then interlaces this region.
3. Subsequently, User 1 transfers one part of the interlaced regions to User 2.
Despite this transfer, User 1 retains the ability to assign the entire original non-interlaced region, which is inconsistent with the fact that they no longer own one of the interlaced parts.

This PR resolves the issue by removing the original region, ensuring that only the two new interlaced regions remain in the state.

### DIFF
--- a/prdoc/pr_2811.prdoc
+++ b/prdoc/pr_2811.prdoc
@@ -1,0 +1,13 @@
+title: "Interlacing removes the region on which it is performed."
+
+doc:
+  - audience: Runtime User
+    description: |
+      The current implementation of the broker pallet does not remove
+      the region on which the interlacing is performed. This can create
+      a vulnerability, as the original region owner is still allowed to 
+      assign a task to the region even after transferring an interlaced
+      part of it.
+
+crates:
+  - name: "pallet-broker"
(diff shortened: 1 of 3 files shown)


cuteolaf (approved):

bkchr (approved):

ggwpez (approved):

seadanda (approved):

Szegoo:
@bkchr If no issues with the PR may we proceed with merging it?

MarBro33:
> With the current code, when a user interlaces their region, the end result will be three regions in the state:
> - the non-interlaced region
> - first part of the interlaced region
> - second part of the interlaced region
> 
> The existing implementation retains the non-interlaced region in the state, leading to a problematic scenario:
> 
> 1. User 1 acquires a region from the market.
> 2. User 1 then interlaces this region.
> 3. Subsequently, User 1 transfers one part of the interlaced regions to User 2.
> Despite this transfer, User 1 retains the ability to assign the entire original non-interlaced region, which is inconsistent with the fact that they no longer own one of the interlaced parts.
> 
> This PR resolves the issue by removing the original region, ensuring that only the two new interlaced regions remain in the state.

PR #2814, opened 2023-12-26
Share peers between syncing strategies

dmitry-markin:
Introduce the ability to share peers between syncing strategies and reserve them for requests. This is needed to run `GapSync` as a separate strategy, and, ultimately, run Sync 2.0 alongside with `ChainSync`.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4806947

substrate/client/network/sync/src/strategy.rs
```diff
@@ -71,6 +72,74 @@ pub struct SyncingConfig {
...
+	fn remove_peer(&self, peer_id: &PeerId) {
+		self.peers.lock().remove(peer_id);
+	}
+
+	fn available_peers(&self) -> Vec<PeerId> {
+		self.peers
+			.lock()
+			.iter()
+			.filter_map(
+				|(peer_id, status)| if status.is_available() { Some(*peer_id) } else { None },
```
altonen, line 105:
```suggestion
				|(peer_id, status)| status.is_available().then_some(*peer_id),
```

substrate/client/network/sync/src/strategy.rs
```diff
@@ -264,114 +350,145 @@ where
...
 	pub fn num_downloaded_blocks(&self) -> usize {
-		match self {
-			SyncingStrategy::WarpSyncStrategy(_) => 0,
-			SyncingStrategy::StateSyncStrategy(_) => 0,
-			SyncingStrategy::ChainSyncStrategy(strategy) => strategy.num_downloaded_blocks(),
+		if let Some(ref chain_sync) = self.chain_sync {
+			chain_sync.num_downloaded_blocks()
+		} else {
+			0
 		}
```
altonen, line 414:
```suggestion
        self.chain_sync.map_or(0, |chain_sync| chain_sync.num_downloaded_blocks())
```

substrate/client/network/sync/src/strategy.rs
```diff
@@ -264,114 +350,145 @@ where
...
 	pub fn num_sync_requests(&self) -> usize {
-		match self {
-			SyncingStrategy::WarpSyncStrategy(_) => 0,
-			SyncingStrategy::StateSyncStrategy(_) => 0,
-			SyncingStrategy::ChainSyncStrategy(strategy) => strategy.num_sync_requests(),
+		if let Some(ref chain_sync) = self.chain_sync {
+			chain_sync.num_sync_requests()
+		} else {
+			0
 		}
```
altonen, line 423:
```suggestion
        self.chain_sync.map_or(0, |chain_sync| chain_sync.num_sync_requests())
```

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -468,130 +475,159 @@ where
...
+				best_hash,
+				best_number,
+				state: PeerSyncState::New,
+			},
+		);
 	}
 
-	#[must_use]
-	fn add_peer_inner(
+	fn handle_new_peers(
```
altonen, line 490: Documentation

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1330,16 +1369,11 @@ where
...
-				// update the request if the new one is available
-				Ok(Some(request)) => ChainSyncAction::SendBlockRequest { peer_id, request },
-				// this implies that we need to drop pending response from the peer
-				Err(bad_peer) => ChainSyncAction::DropPeer(bad_peer),
-			};
+			self.add_peer(peer_id, p.best_hash, p.best_number);
 
-			self.actions.push(action);
+			// since the request is not a justification, remove it from pending responses
+			// TODO: check `PeerSyncState`?
```
altonen, line 1375: ?

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1535,98 +1572,149 @@ where
...
-						if queue.contains(hash) {
-							BlockStatus::Queued
+		self.peer_pool
+			.available_peers()
+			.into_iter()
+			.filter_map(|peer_id| {
+				if let Some(peer) = self.peers.get_mut(&peer_id) {
+					if !peer.state.is_available() || !allowed_requests.contains(&peer_id) {
+						return None
+					}
```
altonen, line 1583: This check should not be needed if it's iterating over available peers

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1649,13 +1737,34 @@ where
...
-				if peer.state.is_available() && peer.common_number >= sync.target_number() {
-					peer.state = PeerSyncState::DownloadingState;
-					let request = sync.next_request();
-					trace!(target: LOG_TARGET, "New StateRequest for {}: {:?}", id, request);
-					self.allowed_requests.clear();
-					return Some((*id, OpaqueStateRequest(Box::new(request))))
+			for peer_id in self.peer_pool.available_peers() {
+				if let Some(peer) = self.peers.get_mut(&peer_id) {
+					if peer.state.is_available() && peer.common_number >= sync.target_number() {
+						if self.peer_pool.try_reserve_peer(&peer_id) {
```
altonen, line 1743: Same here, this should not be possible/it shouldn't be necessary to check that a peer is available in two different places.

substrate/client/network/sync/src/strategy/warp.rs
```diff
@@ -464,12 +474,30 @@ where
...
+			// Find a random available peer that is synced as much as peer majority and is above
 			// `min_best_number`.
-			for (peer_id, peer) in self.peers.iter_mut() {
-				if peer.state.is_available() && peer.best_number >= threshold {
-					peer.state = new_state;
-					return Some(*peer_id)
+			for peer_id in self.peer_pool.available_peers() {
+				if let Some(peer) = self.peers.get_mut(&peer_id) {
+					if peer.state.is_available() && peer.best_number >= threshold {
+						if self.peer_pool.try_reserve_peer(&peer_id) {
```
altonen, line 482: I think this iterating over available peers/reserving a peer has to be refactored. The availability state of a peer should be stored in one place (`PeerPool`) and `try_reserve_peer()` should just be `reserved_peer()` and made infallible. I was playing around with some code that would allow doing that with an iterator but it went nowhere because of the `MutexGuard` we hold.

I think there are two ways to fix this. `PeerPool` has an explicit `lock()` function and each strategy locks it when it wants to query next available peer. This way you can iterate over `PeerPool`s available peers, check whether the next peer satisfies the best number check and if it does, `reserved_peer()` *always* succeeds in reserving the peer since we hold a lock.

Another way, which I think might be worth exploring, is storing some additional state to `PeerPool` which is utilized by all strategies (such as the best number) and then having a function `reserve_next_available_peer(target: B::Number) -> Option<PeerId>` which iterates over all available peers and tries to find a peer whose `best_number > target`, reserves it and returns its `PeerId`. Whether the idea of storing additional information to `PeerPool` is too leaky remains to be seen.

Making `try_reserve_peer()` infallible would make much of the related code cleaner and technically (with Sync 2.0 in mind) failing to reserve a peer which was just previously iterated over as an available peer is not worth a `debug_assert!()` because Sync 1.0 and 2.0 might iterate over the available peers at the same time and try to reserve the same peer, only one succeeding so there should be a way to retry and not crash.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4912205

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4912169

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4918136

substrate/client/network/sync/src/strategy.rs
dmitry-markin: Unfortunately, `map()` consumes `self`, so this won't work. The closest possible solution is probably converting option into iterator, but it's quite ugly and would be even uglier if you need not only to call a function, but return a result.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4919365

substrate/client/network/sync/src/strategy.rs
altonen: what about `self.chain_sync.as_ref().map()`?

substrate/client/network/sync/src/strategy/chain_sync.rs
dmitry-markin: Not every `PeerSyncState` state has an associated pending response. On the other hand, removing pending response when there is actually no active one is harmless, so to not complicate the code I think we don't have to check `PeerSyncState`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4920046

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5063312

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5063313

substrate/client/network/sync/src/strategy/warp.rs
dmitry-markin: I followed the first approach, as different strategies use different approaches for peer selection. For example, `ChainSync` selects peers in a completely different way than warp & state sync.

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -304,10 +306,21 @@ impl<B: BlockT> StateStrategy<B> {
...
+					peer.state = new_state;
+					return Some(*peer_id)
+				}
+			} else {
+				warn!(
+					target: LOG_TARGET,
+					"State inconsistency: peer {peer_id} is in the pool of connected peers, \
+					 but not known to `StateStrategy`.",
+				);
+				debug_assert!(false);
```
altonen, line 323: What about flipping testing the other way: this code iterates over its peers, check if a peer satisfies `peer.best_number >= threshold` and then it calls `PeerPool::try_reserve_peer()` and if returns true, the peer is reserved and if returns false, it continues to the next peer. I think it would clean up this code and would hold the lock only while trying to reserve a peer.

substrate/client/network/sync/src/strategy/warp.rs
```diff
@@ -487,12 +492,23 @@ where
...
 		let median = targets[targets.len() / 2];
 		let threshold = std::cmp::max(median, min_best_number.unwrap_or(Zero::zero()));
-		// Find a random peer that is synced as much as peer majority and is above
+		// Find a random available peer that is synced as much as peer majority and is above
 		// `min_best_number`.
-		for (peer_id, peer) in self.peers.iter_mut() {
-			if peer.state.is_available() && peer.best_number >= threshold {
-				peer.state = new_state;
-				return Some(*peer_id)
+		for mut available_peer in self.peer_pool.lock().available_peers() {
```
altonen, line 497: same here

substrate/client/network/sync/src/strategy/state.rs
```diff
@@ -751,4 +868,6 @@ mod test {
 		// No more actions generated.
 		assert_eq!(state_strategy.actions().count(), 0);
 	}
+
+	// TODO: test that peers are freed from `PeerPool` when responses are received.
```
altonen, line 872: todo

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -468,130 +476,160 @@ where
...
+								}),
+								Some(ancestry_request::<B>(common_best)),
+							)
+						} else {
+							trace!(
+								target: LOG_TARGET,
+								"`ChainSync` is aware of a new peer {peer_id} with unknown best \
+								 hash {best_hash} ({best_number}), but can't start ancestry search \
+								 as the peer is reserved by another syncing strategy.",
+							);
```
altonen, line 587: The peer might also not exist in `PeerPool`?

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -468,130 +476,160 @@ where
...
+				best_number,
+				state: PeerSyncState::New,
+			},
+		);
 	}
 
-	#[must_use]
-	fn add_peer_inner(
+	/// Process new peers assigning proper states and initiating requests.
+	fn handle_new_peers(
```
altonen, line 492: It's unclear to me why this function is needed and why can't the new peers be handled in `ChainSync::block_requests()`

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1312,34 +1353,35 @@ where
...
+					self.add_peer(peer_id, peer_sync.best_hash, peer_sync.best_number);
+				},
+				PeerSyncState::AncestorSearch { .. } |
+				PeerSyncState::DownloadingNew(_) |
+				PeerSyncState::DownloadingStale(_) |
+				PeerSyncState::DownloadingGap(_) |
+				PeerSyncState::DownloadingState => {
+					self.add_peer(peer_id, peer_sync.best_hash, peer_sync.best_number);
+					self.actions.push(ChainSyncAction::CancelRequest { peer_id });
+					self.peer_pool.lock().free_peer(&peer_id);
```
altonen, line 1368: I wonder if here's a race condition. The peer is immediately freed but the cancellation is postponed until `SyncingEngine` processes the event so if some other strategy selects the peer while the request hasn't been canceled, it could result in two in-flight requests. Maybe the strategy/`SyncingStrategy` should be notified by `SyncingEngine` once the request is actually canceled so that freeing the peer is safe.

altonen (commented):
Did a first pass and left some comments but I will have to go over it again.

I'm not super excited about `peer_best_blocks` or of the fact that we store peers both in `PeerPool` and in each strategy and then deal with all the possible inconsistencies that follow from that on runtime. Ideally `PeerPool` would store strategy-specific data but I don't know how feasible that is and I'll do some testing. I also think we may experience yet unknown issues if `GapSync`, `ChainSync` and `Sync2`, or whatever its name will be, have independent (and potentially differing) views of peers. If peer best and common number were stored in `PeerPool`, all strategies could query and update them and we'd get rid of `peer_best_blocks`. What do you think?

I also think `allowed_requests` has to go because `ChainSync` is now checking the availability of a peer three times and the likelyhood that at one point during future refactorings one of them will go out of sync with the others is non-zero.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -92,15 +94,68 @@ pub enum SyncingAction<B: BlockT> {
...
-	StateSyncStrategy(StateStrategy<B>),
-	ChainSyncStrategy(ChainSync<B, Client>),
+pub struct SyncingStrategy<B: BlockT, Client> {
+	config: SyncingConfig,
+	client: Arc<Client>,
+	warp: Option<WarpSync<B, Client>>,
+	state: Option<StateStrategy<B>>,
+	chain_sync: Option<ChainSync<B, Client>>,
+	peer_pool: Arc<Mutex<PeerPool>>,
+	peer_best_blocks: HashMap<PeerId, (B::Hash, NumberFor<B>)>,
```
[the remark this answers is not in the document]
dmitry-markin: It's needed to seed the peers when switching between the strategies. Otherwise, for example, the state strategy won't be aware of the best hash/number (and won't know what peer to request a state from) until the block announcement is received from a peer.

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -468,130 +476,160 @@ where
...
-			Err(bad_peer) => self.actions.push(ChainSyncAction::DropPeer(bad_peer)),
-		}
+		self.peers.insert(
+			peer_id,
+			PeerSync {
+				peer_id,
+				common_number: Zero::zero(),
+				best_hash,
+				best_number,
+				state: PeerSyncState::New,
```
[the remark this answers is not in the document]
dmitry-markin: A new peer and an `Available` peer are handled differently in the code, and the idea was to postpone peer initialization that can lead to requests (`handle_new_peers()`) until `actions()`, so that we can call strategies in the specific order and implement the priorities of reserving the peers (e.g., Sync 2.0, then `ChainSync`, then `GapSync`).

substrate/client/network/sync/src/strategy/chain_sync.rs
dmitry-markin: I'll look if this can be simplified.

substrate/client/network/sync/src/strategy/chain_sync.rs
dmitry-markin: That's possible only due to logic error in the code, as the peers are added to the strategies and `PeerPool` at the same time.

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: We can do that, but now it doesn't make a big difference as requests are generated after `select!` in `SyncingEngine` where all the strategies are called serially, so the lock is always available.

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: Actually already resolved.

dmitry-markin, line 872:
```suggestion
```

substrate/client/network/sync/src/extra_requests.rs
```diff
@@ -17,17 +17,17 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::{
-	request_metrics::Metrics,
-	strategy::chain_sync::{PeerSync, PeerSyncState},
-	LOG_TARGET,
+	peer_pool::PeerPool, request_metrics::Metrics, strategy::chain_sync::PeerSync, LOG_TARGET,
```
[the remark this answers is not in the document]
altonen: Yes I know you haven't named this and the original comment came out maybe a bit too harsh, sorry about that.

Maybe the file could be called `justification_requests.rs` with an explanation at the top of the file mentioning why this extra piece of code is needed to handle justifications?

substrate/client/network/sync/src/strategy/state.rs
altonen: Would get rid of this `else` branch with `debug_assert!(false)` though

substrate/client/network/sync/src/strategy/chain_sync.rs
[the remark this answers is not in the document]
altonen: But `ChainSync::block_requests()` is also only called in `ChainSync::actions()` so is it not possible to insert the new peers as `PeerState::Available` and then deal with all peers the same way?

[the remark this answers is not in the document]
dmitry-markin: I will need to look into this, as new peers are put into ancestry search state by default, and `ChainSync::block_requests()` deals with peers for which we already know the common number. May be we can simplify it, but it's not instantly obvious.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -123,37 +178,59 @@ where
...
-				strategy.add_peer(peer_id, best_hash, best_number),
-		}
+		self.peer_pool.lock().add_peer(peer_id);
+		self.peer_best_blocks.insert(peer_id, (best_hash, best_number));
+
+		self.warp.iter_mut().for_each(|s| s.add_peer(peer_id, best_hash, best_number));
+		self.state.iter_mut().for_each(|s| s.add_peer(peer_id, best_hash, best_number));
+		self.chain_sync
+			.iter_mut()
+			.for_each(|s| s.add_peer(peer_id, best_hash, best_number));
```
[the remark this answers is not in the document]
dmitry-markin: Fixed

substrate/client/network/sync/src/strategy.rs
```diff
@@ -264,226 +337,205 @@ where
...
+			chain_sync.report_metrics();
+		}
+	}
+
+	/// Let `WarpSync` know about target block header
+	pub fn set_warp_sync_target_block_header(
+		&mut self,
+		target_header: B::Header,
+	) -> Result<(), ()> {
+		if let Some(ref mut warp) = self.warp {
```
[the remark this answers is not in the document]
dmitry-markin: Fixed

substrate/client/network/sync/src/strategy.rs
```diff
@@ -264,226 +337,205 @@ where
...
+						Ok(chain_sync) => chain_sync,
+						Err(e) => {
+							error!(target: LOG_TARGET, "Failed to start `ChainSync`.");
+							return Err(e)
+						},
+					};
+					// Let `ChainSync` know about connected peers.
+					self.peer_best_blocks.iter().for_each(|(peer_id, (best_hash, best_number))| {
+						chain_sync.add_peer(*peer_id, *best_hash, *best_number)
+					});
```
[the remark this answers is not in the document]
dmitry-markin: Done

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5088674

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -690,6 +728,7 @@ where
 				blocks.reverse()
 			}
 			self.allowed_requests.add(peer_id);
```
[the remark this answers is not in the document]
dmitry-markin: We can't remove `allowed_requests` yet, as they are used to block block requests when downloading state during fast sync. It should be safe to get rid of them once we refactor fast sync into a sequence of `ChainSync` -> `State` -> `ChainSync` strategies. You can argue that we can already replace `allowed_requests` with a simple switch, but I'd rather keep them as they are as the logic of masking peers during block request is also not obvious to me (and probably buggy), so IMO it's safer to address this separately.

substrate/client/network/sync/src/extra_requests.rs
```diff
@@ -288,6 +288,7 @@ impl<'a, B: BlockT> Matcher<'a, B> {
 	pub(crate) fn next(
 		&mut self,
 		peers: &HashMap<PeerId, PeerSync<B>>,
+		peer_pool: &Arc<Mutex<PeerPool>>,
```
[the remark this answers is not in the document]
dmitry-markin: I don't think `RwLock` makes sense here, as we are mostly interested in reserving peers, so read access is not prevalent.

substrate/client/network/sync/src/strategy/state.rs
dmitry-markin: Done

substrate/client/network/sync/src/strategy/chain_sync.rs
dmitry-markin: It's even worse than that. When another strategy initiates the request, the old one is automatically dropped. And when we finally cancel the request, we can cancel a legitimate request of another strategy.

I'm not super happy with any kind of ACKing and bookkeeping, but it looks like we need to keep track of pending cancellations in every strategy and free peers in `PeerPool` when something like `on_request_cancelled()` is heard from `SyncingEngine`.

altonen: Is there a reason why pending cancellations couldn't be stored in `SyncingStrategy` and once the request has been canceled, `SyncingEngine` calls `SyncingStrategy` which releases the peer?

dmitry-markin: This means intercepting request cancellation actions in `SyncingStrategy` on their way to `SyncingEngine`. Not the best design ever, but this is probably better then dealing with it in every strategy.

altonen: If we store peer's best and common blocks in `PeerPool`, we'd get rid of `peer_best_blocks`, allow `SyncingEngine` to replace its `HashMap<PeerId, Peer<B>>` with `PeerPool` and also allow it to free the peer once the request is canceled.

But now that I'm thinking this whole approach again, is there any fundamental reason why `GapSync` and `ChainSync` couldn't send a request to the same peer at the same time? I believe this limitation only applies to the sending end, namely the strategies, and made sense before when everything was one state machine. The request handlers shouldn't care how many requests from the same peer are in the queue. If `GapSync` and `ChainSync` are separate, is there a valid reason why they couldn't send simultaneous requests to a peer? We must ascertain that two copies of the same request are not sent because that'd get us banned, implying an existence of some kind of `PeerPool` for shared data, but is there a reason why they couldn't send two different requests at the same time? So the limitation of one request per peer still applies but now it's per strategy.

dmitry-markin: > is there any fundamental reason why GapSync and ChainSync couldn't send a request to the same peer at the same time?

Valid point, there is nothing in the block request handler that forbids "simultaneous" requests. But if we get rid of the global `PeerPool`, we'll need to introduce a way to cancel specific requests, as otherwise strategies could cancel requests of each other.

altonen: Each strategy could be identified by a unique key and when strategy returns `StrategyAction::SendRequest { PeerId, Request }`, `SyncingStrategy` would convert it to `SyncingAction::SendRequest { PeerId, Key, Request }` and `PendingResponses` would keep track of responses with `(PeerId, Key)`.

dmitry-markin: Yes, that's what I had in mind.

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -468,130 +489,160 @@ where
...
-						best_hash,
-						best_number,
-						state: PeerSyncState::Available,
-					},
-				);
-				self.allowed_requests.add(&peer_id);
-				Ok(None)
-			},
-		}
+					self.allowed_requests.add(&peer_id);
```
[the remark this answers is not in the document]
dmitry-markin: I don't completely understand the logic behind it too, it looks like the only useful thing it's doing is blocking block request during state download in fast sync.

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1649,13 +1724,27 @@ where
...
-				if peer.state.is_available() && peer.common_number >= sync.target_number() {
-					peer.state = PeerSyncState::DownloadingState;
-					let request = sync.next_request();
-					trace!(target: LOG_TARGET, "New StateRequest for {}: {:?}", id, request);
-					self.allowed_requests.clear();
-					return Some((*id, OpaqueStateRequest(Box::new(request))))
+			for mut available_peer in self.peer_pool.lock().available_peers() {
+				let peer_id = available_peer.peer_id();
+				if let Some(peer) = self.peers.get_mut(&peer_id) {
+					if peer.state.is_available() && peer.common_number >= sync.target_number() {
```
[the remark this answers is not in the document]
dmitry-markin: Yes, should not be needed.

dmitry-markin:
@skunert Thanks for reviewing the PR, but I'm about to publish another one that should completely supersede it. So, please don't spend more time on reviewing this one for now.

dmitry-markin:
Closing in favor of https://github.com/paritytech/polkadot-sdk/pull/3224.

PR #2818, opened 2023-12-28
add alloc::string to sp-std

yjhmelody:
Sorry for forgetting to add explanations.

I feel that in the current environment, this restriction no longer applies. More and more libraries support no_std, and they use the String type. For example, wasmi itself also uses it, and the primitives crate of our project also uses String.


xlc:
any explanations?

ggwpez (changes requested):
?

PR #1619, opened 2023-09-18
Updating glutton for async backing

BradleyOlson64:
Applied changes from the [User Update Guide](https://docs.google.com/document/d/1WQijD3bZTCsudOyPcDvugv659nCa2hEp2b_8eRU0h-Q), diverging in the node side where service.rs is different for `polkadot-parachain` than in the parachain template.

Issue #1566, opened 2023-09-14
Error after upgrading polkadot-parachain to v1.1.0-polkadot on Asset Hub


ilhanu:
I've updated 6 system chains with ansible-galaxy. From v1.0.0 to v1.1.0, I'm using the binary that is published in the releases.  5 out of 6 upgrades went smooth. One of the machines is giving an error which I can't solve. 

```
2023-09-14 13:25:32.761  INFO main sc_cli::runner: Polkadot parachain
2023-09-14 13:25:32.761  INFO main sc_cli::runner: ✌️  version 1.1.0-783579b4a49
2023-09-14 13:25:32.761  INFO main sc_cli::runner: ❤️  by Parity Technologies <admin@parity.io>, 2017-2023
2023-09-14 13:25:32.761  INFO main sc_cli::runner: 📋 Chain specification: Polkadot Asset Hub
....
....
2023-09-14 13:13:14.603  WARN tokio-runtime-worker sc_service::builder: [Parachain] The NetworkStart returned as part of `build_network` has been silently dropped
Error: Service(Client(RuntimeApiError(Application(Execution(Other("Exported method AuraApi_slot_duration is not found"))))))
asset-hub-dot.service: Main process exited, code=exited, status=1/FAILURE
asset-hub-dot.service: Failed with result 'exit-code'.
```

There are no other warnings or errors in the start-up logs after the starting. After that it just keeps rebooting.

I have deleted the DB, since it was a redundant node and tried to re-sync it, but the error pertains. 

The exec parameters
```
COMMON="\
--base-path /home/statemint/.local/share/polkadot \
--detailed-log-output"
RC_CHAIN="--chain polkadot"
RC_ADDR="\
--listen-addr=/ip4/0.0.0.0/tcp/30338 \
--public-addr=/ip4/xxxx/tcp/30338"
RC_CONNECTIONS="--in-peers 25 --out-peers 25"
RC_DB="\
--database paritydb \"
RC_PRUNING="--state-pruning=256"
RC_METRICS="\
--prometheus-external \
--prometheus-port 9620"
RC_RPC="--rpc-port 10000"


PC_NAME="--name 'Staker Space [2]'"
PC_ROLE_SPECIFIC="\
--collator"
PC_CHAIN="--chain asset-hub-polkadot"
PC_REMOTE_RC_URLS=""
PC_ADDR="\
--listen-addr=/ip4/0.0.0.0/tcp/31555 \
[...]

PR #2826, opened 2023-12-29
Enable async backing on asset-hub-rococo

mrcnski:
The goal is to move all system chains on Rococo (+ other testnets?) to use async backing.

Starting with `asset-hub-rococo` to get feedback, before I do the rest.

## Related

Example: https://github.com/paritytech/polkadot-sdk/pull/1619/
Guide: https://github.com/w3f/polkadot-wiki/blob/master/docs/maintain/maintain-guides-async-backing.md


cumulus/polkadot-parachain/src/service.rs
```diff
@@ -1410,7 +1410,9 @@ where
 	<<AuraId as AppCrypto>::Pair as Pair>::Signature:
 		TryFrom<Vec<u8>> + std::hash::Hash + sp_runtime::traits::Member + Codec,
 {
-	start_node_impl::<RuntimeApi, _, _, _>(
+	// TODO: is this right? Seems okay since asset hub node is a "shell" node. Can Kusama start
+	// using this right away?
+	start_basic_lookahead_node_impl::<RuntimeApi, _, _, _>(
```
mrcnski, line 1415: I wasn't sure about this because asset-hub-kusama also calls this line. Would it break something? I would guess we should we test this change on rococo first actually.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4817223

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4817957

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4817958

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -712,9 +740,11 @@ impl pallet_aura::Config for Runtime {
 	type AuthorityId = AuraId;
 	type DisabledValidators = ();
 	type MaxAuthorities = ConstU32<100_000>;
-	type AllowMultipleBlocksPerSlot = ConstBool<false>;
+	// TODO: should this be true? Guide says false, but see:
+	// https://github.com/paritytech/polkadot-sdk/pull/1619/files#diff-17c0e44911184254502a1d0b157c58e706eaa938cd0b1fe34c0456e541066a2fR234
+	type AllowMultipleBlocksPerSlot = ConstBool<true>;
```
joepetrowski, line 745: Yeah interesting, `false` makes more sense to me. @eskimor ?

Actually, guide says `true`: https://github.com/w3f/polkadot-wiki/blob/master/docs/maintain/maintain-guides-async-backing.md#phase-3---activate-async-backing

cumulus/polkadot-parachain/src/service.rs
joepetrowski: I don't know this node code well enough, but my understanding is that it needs the `lookahead`?

joepetrowski (commented):
Happy to try it on Rococo 🤠

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
mrcnski: Yeah, I figure that's the intended final value. It's just a bit confusing because we set it to `false` before that, in step 1.5.

cumulus/polkadot-parachain/src/service.rs
seadanda: You're right, Kusama and Polkadot AH both call this too. A change here will affect how `polkadot-parachain` runs any of them, so if you want to only move testnets over in the next released version without moving Kusama and Polkadot AH onto async backing you'll need two separate functions.

mrcnski: Thanks! So I'll copy/paste this and `start_node_impl` into new functions e.g. `start_lookahead_node_impl` with the necessary changes from the guide. It's a bit ugly but I guess soon the non-async-backing `start_asset_hub_node` can be removed.

mrcnski: @seadanda Looks like `AuraParams` is missing `collation_request_receiver`, which is `Some` in this case. Should I add it, and copy the code that uses it from `basic_aura` to `lookahead::aura`?

seadanda: Yea it would be nice to find a cleaner way to combine them, the diff is very small between the four `_impl` functions, but nothing jumps out at me immediately because of the slightly different constraints.

Not too sure on your second comment, I'm not familiar with the implementation of lookahead vs basic.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -137,6 +136,30 @@ pub fn native_version() -> NativeVersion {
...
+);
+
+/// Maximum number of blocks simultaneously accepted by the Runtime, not yet included
+/// into the relay chain.
+const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
+/// How many parachain blocks are processed by the relay chain per parent. Limits the
+/// number of blocks authored per slot.
+const BLOCK_PROCESSING_VELOCITY: u32 = 1;
+/// Relay chain slot duration, in milliseconds.
+const RELAY_CHAIN_SLOT_DURATION_MILLIS: u32 = 6000;
```
bkontur, line 152: why not to import `parachains_common::rococo::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS` instead?

mrcnski: Seems reasonable, I was just following the guide and [this PR](https://github.com/paritytech/polkadot-sdk/pull/1619/files#diff-17c0e44911184254502a1d0b157c58e706eaa938cd0b1fe34c0456e541066a2f). Maybe someone more familiar with runtimes has insight, if it makes sense to keep these values separate or not.

cumulus/polkadot-parachain/src/service.rs
```diff
@@ -1530,6 +1723,157 @@ where
...
+					collator_key,
+					para_id,
+					overseer_handle,
+					slot_duration,
+					relay_chain_slot_duration,
+					proposer,
+					collator_service,
+					authoring_duration: Duration::from_millis(1500),
+					// TODO: is the below needed?
+					//collation_request_receiver: Some(request_stream),
```
mrcnski, line 1860: Is this still needed here (see `start_asset_hub_node`), or is it only for transitioning to aura above?

cumulus/polkadot-parachain/src/service.rs
mrcnski: I'm getting a really long error message now. 

```
error[E0277]: the trait bound `<RuntimeApi as ConstructRuntimeApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_service::client::client::Client<sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, LocalCallExecutor<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, WasmExecutor<(sp_io::storage::HostFunctions, sp_io::default_child_storage::HostFunctions, sp_io::misc::HostFunctions, sp_io::wasm_tracing::HostFunctions, sp_io::offchain::HostFunctions, sp_io::crypto::HostFunctions, sp_io::hashing::HostFunctions, sp_io::allocator::HostFunctions, sp_io::panic_handler::HostFunctions, sp_io::logging::HostFunctions, sp_io::trie::HostFunctions, sp_io::offchain_index::HostFunctions, sp_io::transaction_index::HostFunctions)>>, sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, RuntimeApi>>>::RuntimeApi: cumulus_primitives_aura::AuraUnincludedSegmentApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>` is not satisfied
    --> cumulus/polkadot-parachain/src/service.rs:1864:80
     |
1864 |                 aura::run::<Block, <AuraId as AppCrypto>::Pair, _, _, _, _, _, _, _, _, _>(params)
     |                 -------------------------------------------------------------------------- ^^^^^^ the trait `cumulus_primitives_aura::AuraUnincludedSegmentApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>` is not implemented for `<RuntimeApi as ConstructRuntimeApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_service::client::client::Client<sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, LocalCallExecutor<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, WasmExecutor<(sp_io::storage::HostFunctions, sp_io::default_child_storage::HostFunctions, sp_io::misc::HostFunctions, sp_io::wasm_tracing::HostFunctions, sp_io::offchain::HostFunctions, sp_io::crypto::HostFunctions, sp_io::hashing::HostFunctions, sp_io::allocator::HostFunctions, sp_io::panic_handler::HostFunctions, sp_io::logging::HostFunctions, sp_io::trie::HostFunctions, sp_io::offchain_index::HostFunctions, sp_io::transaction_index::HostFunctions)>>, sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, RuntimeApi>>>::RuntimeApi`
     |                 |
     |                 required by a bound introduced by this call
     |
note: required by a bound in `cumulus_client_consensus_aura::collators::lookahead::run`
    --> /Users/marcinparity/Repos/github.com/paritytech/polkadot-sdk-3/cumulus/client/consensus/aura/src/collators/lookahead.rs:127:61
     |
113  | pub fn run<Block, P, BI, CIDP, Client, Backend, RClient, CHP, SO, Proposer, CS>(
     |        --- required by a bound in this function
...
127  |         AuraApi<Block, P::Public> + CollectCollationInfo<Block> + AuraUnincludedSegmentApi<Block>,
     |                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `run`
help: consider further restricting the associated type
     |
1751 |         TryFrom<Vec<u8>> + std::hash::Hash + sp_runtime::traits::Member + Codec, <RuntimeApi as ConstructRuntimeApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_service::client::client::Client<sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, LocalCallExecutor<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, sc_client_db::Backend<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>, WasmExecutor<(sp_io::storage::HostFunctions, sp_io::default_child_storage::HostFunctions, sp_io::misc::HostFunctions, sp_io::wasm_tracing::HostFunctions, sp_io::offchain::HostFunctions, sp_io::crypto::HostFunctions, sp_io::hashing::HostFunctions, sp_io::allocator::HostFunctions, sp_io::panic_handler::HostFunctions, sp_io::logging::HostFunctions, sp_io::trie::HostFunctions, sp_io::offchain_index::HostFunctions, sp_io::transaction_index::HostFunctions)>>, sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, RuntimeApi>>>::RuntimeApi: cumulus_primitives_aura::AuraUnincludedSegmentApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>
     |                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I think in essence it's saying that `AuraUnincludedSegmentApi<Block>` bound is missing for `RuntimeApi`, but I'm lost on how to fix it. Maybe after the transition to aura, a new client needs to be constructed?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4897141

cumulus/polkadot-parachain/src/service.rs
mrcnski: @georgepisaltu Do you have any ideas on this or https://github.com/paritytech/polkadot-sdk/pull/2826/#pullrequestreview-1816111722?

georgepisaltu: For this long error, you just need to add the correct trait bound, `cumulus_primitives_aura::AuraUnincludedSegmentApi<Block>`, [here](https://github.com/paritytech/polkadot-sdk/blob/f7306d327e0eee0a641ac43628ab525b0096ba7c/cumulus/polkadot-parachain/src/service.rs#L1409).

In the case of AH collators, the runtime initially started as shell and then later transitioned into the real runtime which had Aura. Therefore, because the genesis is already set and will never change, AH collators will always have to account for this edge case, meaning they `start_asset_hub_node` will always be different from the rest. More details [here](https://github.com/paritytech/polkadot-sdk/issues/1566) and [here](https://github.com/paritytech/polkadot-sdk/pull/1788).

The `collation_request_receiver` was added because there is already a relay chain collation request stream initialized for the shell consensus, and it can't be initialized again so it needs to be reused when transitioning into Aura. More details [here](https://github.com/paritytech/polkadot-sdk/pull/1911).

cumulus/polkadot-parachain/src/service.rs
georgepisaltu: I see that the lookahead implementation doesn't use the [relay chain driven collation request stream](https://github.com/paritytech/polkadot-sdk/blob/f7306d327e0eee0a641ac43628ab525b0096ba7c/cumulus/client/consensus/aura/src/collators/basic.rs#L122), but instead uses [`cumulus_client_collator::initialize_collator_subsystems`](https://github.com/paritytech/polkadot-sdk/blob/f7306d327e0eee0a641ac43628ab525b0096ba7c/cumulus/client/consensus/aura/src/collators/lookahead.rs#L148), which still [sends a message](https://github.com/paritytech/polkadot-sdk/blob/f7306d327e0eee0a641ac43628ab525b0096ba7c/cumulus/client/collator/src/lib.rs#L248) which will [error as a double initialization](https://github.com/paritytech/polkadot-sdk/blob/f7306d327e0eee0a641ac43628ab525b0096ba7c/polkadot/node/collation-generation/src/lib.rs#L136). To fix this, you will probably have to create a special message to override an existing config with the lookahead specific one.

mrcnski: ⭐ 

Thanks @georgepisaltu for walking me through these issues - super helpful! What do you think would be a reasonable change here? I'm thinking we could add a new parameter to `initialize_collator_subsystems` like `reinitialize: bool` and if it's true, send a new message like `CollationGenerationMessage::Reinitialize` instead of `CollationGenerationMessage::Initialize` which doesn't have an initialization limit error. Do you think this makes sense or is there a better way to do it?

mrcnski:
Raised https://github.com/w3f/polkadot-wiki/pull/5506 with some fixes to the guide. In a sec I'll push a commit here removing the corresponding TODOs.

cumulus/polkadot-parachain/src/service.rs
georgepisaltu: > we could add a new parameter to `initialize_collator_subsystems` like `reinitialize: bool`

Sounds good to me.

> send a new message like `CollationGenerationMessage::Reinitialize` instead of `CollationGenerationMessage::Initialize` which doesn't have an initialization limit error

Also sounds good. The other option I see is to have a new `override` or `force` parameter in the `Initialize` message, which, if set, ignores any existing config and replaces it with the new one.

But I think creating a new message for this is better because it's backwards compatible.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4915003

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4917329

eskimor:
@joepetrowski please ping me when on Rococo :nerd_face: :rocket:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4933240

joepetrowski (approved):

mrcnski:
@seadanda @georgepisaltu It's ready for final review. 😏

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
```diff
@@ -55,15 +55,15 @@ use frame_system::{
...
 use parachains_common::{
 	impls::DealWithFees,
 	message_queue::*,
 	westend::{consensus::*, currency::*, fee::WeightToFee},
 	AccountId, AssetIdForTrustBackedAssets, AuraId, Balance, BlockNumber, Hash, Header, Nonce,
 	Signature, AVERAGE_ON_INITIALIZE_RATIO, DAYS, HOURS, MAXIMUM_BLOCK_WEIGHT,
-	NORMAL_DISPATCH_RATIO, SLOT_DURATION,
+	NORMAL_DISPATCH_RATIO,
 };
+pub use parachains_common::{westend::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS, SLOT_DURATION};
```
seadanda, line 66: Why are you re-exporting this? From what I can see it's fine to just use it without needing to be pub.

Changing it from re-exporting the full module to just these two consts is a breaking change anyway.

I recently removed these `pub use`s from the fellowship runtimes, maybe worth doing here too, as I don't see why anybody would add the whole runtime as a dependency just to get access to `parachains_common`.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -94,8 +94,9 @@ use parachains_common::{
 	impls::DealWithFees,
 	rococo::{consensus::*, currency::*, fee::WeightToFee},
 	AccountId, Balance, BlockNumber, Hash, Header, Nonce, Signature, AVERAGE_ON_INITIALIZE_RATIO,
-	HOURS, MAXIMUM_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO, SLOT_DURATION,
+	HOURS, MAXIMUM_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO,
 };
+pub use parachains_common::{rococo::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS, SLOT_DURATION};
```
seadanda, line 99: same with this
```suggestion
use parachains_common::{rococo::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS, SLOT_DURATION};
```

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-westend/src/lib.rs
```diff
@@ -83,8 +83,9 @@ use parachains_common::{
 	impls::DealWithFees,
 	westend::{consensus::*, currency::*, fee::WeightToFee},
 	AccountId, Balance, BlockNumber, Hash, Header, Nonce, Signature, AVERAGE_ON_INITIALIZE_RATIO,
-	HOURS, MAXIMUM_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO, SLOT_DURATION,
+	HOURS, MAXIMUM_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO,
 };
+pub use parachains_common::{westend::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS, SLOT_DURATION};
```
seadanda, line 88: Same here:
```suggestion
use parachains_common::{westend::consensus::RELAY_CHAIN_SLOT_DURATION_MILLIS, SLOT_DURATION};
```

cumulus/polkadot-parachain/src/command.rs
```diff
@@ -706,6 +704,16 @@ pub fn run() -> Result<()> {
...
 
+				    AssetHubRococo =>
+						crate::service::start_asset_hub_lookahead_node::<
+						RuntimeApi,
+							AuraId,
+						>(config, polkadot_config, collator_options, id, hwbench)
+						.await
+						.map(|r| r.0)
+						.map_err(Into::into),
+
```
seadanda, line 715:
```suggestion
```

polkadot/node/subsystem-types/src/messages.rs
```diff
@@ -832,6 +832,8 @@ pub enum ProvisionerMessage {
 pub enum CollationGenerationMessage {
 	/// Initialize the collation generation subsystem
 	Initialize(CollationGenerationConfig),
+	/// Reinitialize the collation generation subsystem, overriding the existing config
```
seadanda, line 835:
```suggestion
	/// Initialize the collation generation subsystem.
	Initialize(CollationGenerationConfig),
	/// Reinitialize the collation generation subsystem, overriding the existing config.
```

cumulus/polkadot-parachain/src/service.rs
```diff
@@ -1530,6 +1531,158 @@ where
 	.await
 }
 
+/// Start a shell node which should later transition into an Aura powered parachain node. Asset Hub
+/// uses this because at genesis, Asset Hub was on the `shell` runtime which didn't have Aura and
+/// needs to sync and upgrade before it can run `AuraApi` functions.
+///
+/// Uses the lookahead collator to support async backing.
+pub async fn start_asset_hub_lookahead_node<RuntimeApi, AuraId: AppCrypto + Send + Codec + Sync>(
```
seadanda, line 1539:
```suggestion
#[sc_tracing::logging::prefix_logs_with("Parachain")]
pub async fn start_asset_hub_lookahead_node<RuntimeApi, AuraId: AppCrypto + Send + Codec + Sync>(
```

seadanda (approved):
Looks good! I just have a few small queries.

georgepisaltu (approved):

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
mrcnski: Looks like it was exported for use in the tests. I didn't notice that, IMO exports should not be interspersed with imports.

mrcnski: Should I add back to the `pub use ... as common` to prevent the breaking change, or just remove this line?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4960193

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4960585

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
seadanda: I think you could go with either option for that:

1. Removing enforces importing directly from common, where it is actually defined and is a breaking changed but if documented is completely legit IMO. The polkadot and kusama runtimes have just done this.

2. Leaving it in as is doesn't actually cost you anything and is the minimal diff for this PR, but continues with an unnecessary re-export

Up to you on that one, I was just highlighting that your previous changes were not purely internal to the runtime.

mrcnski: I'm fixing this and it looks like a bunch of tests were using the re-exports. We will have to remember to change it back in https://github.com/paritytech/polkadot-sdk/pull/2949/ once the runtimes define their own SLOT_DURATION, so that the tests pick up the right constants. Not a big deal, worst case scenario if we forget to do that is some tests will fail.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4982851

PR #2834, opened 2024-01-02
proposer: return optional block

rphmeier:
This opens up the proposer to only optionally create blocks. Nodes may only make blocks when there are transactions or the chain is scheduled.


bkchr (approved):

bkchr:
bot fmt

skunert (approved):

sandreim (approved):

PR #2839, opened 2024-01-02
Refactors the `OnStakingUpdate` trait

gpestana:
This PR refactors the `OnStakingUpdate` trait to make its usage simpler and less error prone in the context of pallet-staking and the [pallet-stake-tracker](https://github.com/paritytech/polkadot-sdk/pull/1933). The main goal is to make readily available more information/data required by the `OnStakingUpdate` listener (e.g. new stake on `on_stake_update`) through the interface methods.

With this refactor, the event emitter needs to explicitly call the events with more information about the event. The advantage is that this new interface design prevents potential issues with data sync, e.g. the event emitter does not necessarily need to update the staking state before emitting the events and the `OnStakingUpdate` implementor does not need to rely as much on the staking interface, making the interface less error prone. 

### Examples

Currently, on the `OnStakingUpdate` implemetor, we have

```rust
impl OnStakingUpdate<AccountId, Balance> for X {

  // snip..
 
  // the implementor will most likely need to fetch the nominations of the new nominator.
  fn on_nominator_add(_who: &AccountId) {

    // if the event emitter did not update the nominations at this point, the data is out of sync.
    let nominations = stakingInterface::get_nominations(); 
  }
}
```

with the new interface design, the implementor will have all the most important information implicitly in the method interface:

```rust
impl OnStakingUpdate<AccountId, Balance> for X {

  // snip..

  fn on_nominator_add(_who: &AccountId, nominations: Vec<AccountId>) {
    // no need to call into staking interface or other storage to fetch the nominations of the new nominator.
  }
}
```

Although the new design is less error prone, it is more opinionated and the design is highly affected by how the staking pallet will use this interface. Evidently, it is not possible to cover all the inputs that all the `OnStakingUpdate` implementors may need (depending on the use case), but the refactor aims at striking a good balance between safety for the usage in the context of staking and generality.

### DIFF
--- a/substrate/primitives/staking/src/lib.rs
+++ b/substrate/primitives/staking/src/lib.rs
@@ -95,24 +95,34 @@ pub struct Stake<Balance> {
 /// pre-action data that is needed needs to be passed to interface methods. The rest of the data can
 /// be retrieved by using `StakingInterface`.
 #[impl_trait_for_tuples::impl_for_tuples(10)]
-pub trait OnStakingUpdate<AccountId, Balance> {
+pub trait OnStakingUpdate<AccountId, Balance, ValidatorPrefs> {
 	/// Fired when the stake amount of someone updates.
 	///
 	/// This is effectively any changes to the bond amount, such as bonding more funds, and
 	/// unbonding.
-	fn on_stake_update(_who: &AccountId, _prev_stake: Option<Stake<Balance>>) {}
+	fn on_stake_update(
+		_who: &AccountId,
+		_prev_stake: Option<Stake<Balance>>,
+		_stake: Stake<Balance>,
+	) {
+	}
 
 	/// Fired when someone sets their intention to nominate.
 	///
 	/// This should never be fired for existing nominators.
-	fn on_nominator_add(_who: &AccountId) {}
+	fn on_nominator_add(_who: &AccountId, _nominations: Vec<AccountId>) {}
 
 	/// Fired when an existing nominator updates their nominations.
 	///
 	/// Note that this is not fired when a nominator changes their stake. For that,
 	/// `on_stake_update` should be used, followed by querying whether `who` was a validator or a
 	/// nominator.
-	fn on_nominator_update(_who: &AccountId, _prev_nominations: Vec<AccountId>) {}
+	fn on_nominator_update(
+		_who: &AccountId,
+		_prev_nominations: Vec<AccountId>,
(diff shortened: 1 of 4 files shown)


gpestana:
bot fmt

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4828489

rossbulat (approved):
LGTM, useful additions.

substrate/primitives/staking/src/lib.rs
```diff
@@ -123,15 +133,15 @@ pub trait OnStakingUpdate<AccountId, Balance> {
 	/// Fired when someone sets their intention to validate.
 	///
 	/// Note validator preference changes are not communicated, but could be added if needed.
-	fn on_validator_add(_who: &AccountId) {}
+	fn on_validator_add(_who: &AccountId, _self_stake: Option<Stake<Balance>>) {}
 
 	/// Fired when an existing validator updates their preferences.
 	///
 	/// Note validator preference changes are not communicated, but could be added if needed.
 	fn on_validator_update(_who: &AccountId) {}
```
Ank4n, line 141: Wouldn't this also need old and new self stake params?

substrate/primitives/staking/src/lib.rs
```diff
@@ -100,19 +100,29 @@ pub trait OnStakingUpdate<AccountId, Balance> {
 	///
 	/// This is effectively any changes to the bond amount, such as bonding more funds, and
 	/// unbonding.
-	fn on_stake_update(_who: &AccountId, _prev_stake: Option<Stake<Balance>>) {}
+	fn on_stake_update(
+		_who: &AccountId,
+		_prev_stake: Option<Stake<Balance>>,
```
Ank4n, line 105: Should we have both `_prev_stake` and `_stake` as same type? Either `Stake<Balance>` or `Option<Stake<Balance>>`. If `Stake<Balance>`, we can pass a zero value.

Ank4n (approved):

gpestana:
Closing as these changes have been included in the stake tracker PR https://github.com/paritytech/polkadot-sdk/pull/1933 and only make sense in that context.

gpestana:
Closing as these changes have been included in the stake tracker PR https://github.com/paritytech/polkadot-sdk/pull/1933 and only make sense in that context.

Issue #2836, opened 2024-01-02
orchestra-proc-macro warning in build log


ggwpez:
Seeing a log of these recently (also in the CI https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4827444):  
```pre
[orchestra-proc-macro-0.3.3/src/graph.rs:207] err_msg = "Expected value after assignment"
Hetscher/Hiccup: Failed to parse dotfile
caused by: Expected value after assignment
expander: writing orchestra-expansion-c5de4377c228.rs
```

Dont know how bad that is, but the build and clippy still work.

PR #2848, opened 2024-01-04
Remove `dotgraph` feature

bkchr:
Apparently the implementation is broken and is generating warnings in the build. Not sure this was used any way by anyone.

Closes: https://github.com/paritytech/polkadot-sdk/issues/2836

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7166,15 +7166,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "layout-rs"
-version = "0.1.1"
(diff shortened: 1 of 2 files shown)


sandreim (approved):
This is actually a bug that needs fixed in orchestra. I created a [ticket](https://github.com/paritytech/orchestra/issues/67) for it . I think it is fine to remove the feature for now.

ggwpez (approved):

bkchr:
> This is actually a bug that needs fixed in orchestra.

Yes I know, but do you actually use this? I mean we have the warning about cycles there since 2 years and no one cares...

PR #2853, opened 2024-01-04
Update missing worker binaries error

mrcnski:
Please let me know if this would be a better UX. Should we have specific links in the error message?

### DIFF
--- a/polkadot/node/service/src/lib.rs
+++ b/polkadot/node/service/src/lib.rs
@@ -239,7 +239,7 @@ pub enum Error {
 	InvalidWorkerBinaries { prep_worker_path: PathBuf, exec_worker_path: PathBuf },
 
 	#[cfg(feature = "full-node")]
-	#[error("Worker binaries could not be found, make sure polkadot was built and installed correctly. Please see the readme or wiki for the latest instructions. If you ran with `cargo run`, please run `cargo build` first. Searched given workers path ({given_workers_path:?}), polkadot binary path ({current_exe_path:?}), and lib path (/usr/lib/polkadot), workers names: {workers_names:?}")]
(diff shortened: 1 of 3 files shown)


polkadot/node/service/src/lib.rs
```diff
@@ -239,7 +239,7 @@ pub enum Error {
 	InvalidWorkerBinaries { prep_worker_path: PathBuf, exec_worker_path: PathBuf },
 
 	#[cfg(feature = "full-node")]
-	#[error("Worker binaries could not be found, make sure polkadot was built/installed correctly. If you ran with `cargo run`, please run `cargo build` first. Searched given workers path ({given_workers_path:?}), polkadot binary path ({current_exe_path:?}), and lib path (/usr/lib/polkadot), workers names: {workers_names:?}")]
+	#[error("Worker binaries could not be found, make sure polkadot was built and installed correctly. Please see the readme or wiki for the latest instructions. If you ran with `cargo run`, please run `cargo build` first. Searched given workers path ({given_workers_path:?}), polkadot binary path ({current_exe_path:?}), and lib path (/usr/lib/polkadot), workers names: {workers_names:?}")]
```
ggwpez, line 242: Maybe make a case-decision that describes it for local builds and for downloaded distribution paths?  
Like: if you have the source code locally, run `cargo build`  
if you downloaded it, then also download the two other bins and mark them as executable.  

But yea, linking to some readme maybe good.

bkchr (approved):

polkadot/node/service/src/lib.rs
mrcnski: Do you mean in the error message? Having too much specific info seems like it will be hard to keep up-to-date. Maybe just link to the readme - let me know if that's better.

And did you have to mark them executable? On Mac I just have to accept a pop-up that asks if they are safe to run - should this step be added to the readme?

ggwpez (approved):

polkadot/node/service/src/lib.rs
ggwpez: I mean the `chmod +x`, otherwise there is an error that they are not executable. I guess its normal when downloading a binary 🤷‍♂️

Issue #2568, opened 2023-11-30
`0002-validators-warp-sync` test failing


michalkucharczyk:
The warp sync in [following configuration](https://github.com/paritytech/polkadot-sdk/blob/64361ac19a3c74ec50329403e45a3a98fc40ac5b/substrate/zombienet/0002-validators-warp-sync/test-validators-warp-sync.toml) is sometimes failing.

### Failure
Sometimes [bob](https://github.com/paritytech/polkadot-sdk/blob/64361ac19a3c74ec50329403e45a3a98fc40ac5b/substrate/zombienet/0002-validators-warp-sync/test-validators-warp-sync.toml#L17) is trying do download the state for some old block. And since it cannot download it, it cannot accomplish warp-sync. Here is what it says:
```
...
2023-11-29 13:36:05.204  INFO tokio-runtime-worker substrate: ⏩ Warping, Downloading state, 0.11 Mib (0 peers), best: #0 (0xedb3…0fa9), finalized #0 (0xedb3…0fa9), ⬇ 24.3kiB/s ⬆ 1.8kiB/s
2023-11-29 13:36:10.205  INFO tokio-runtime-worker substrate: ⏩ Warping, Downloading state, 0.11 Mib (0 peers), best: #0 (0xedb3…0fa9), finalized #0 (0xedb3…0fa9), ⬇ 2.7kiB/s ⬆ 2.9kiB/s
2023-11-29 13:36:15.206  INFO tokio-runtime-worker substrate: ⏩ Warping, Downloading state, 0.11 Mib (0 peers), best: #0 (0xedb3…0fa9), finalized #0 (0xedb3…0fa9), ⬇ 2.1kiB/s ⬆ 2.5kiB/s
2023-11-29 13:36:20.207  INFO tokio-runtime-worker substrate: ⏩ Warping, Downloading state, 0.11 Mib (0 peers), best: #0 (0xedb3…0fa9), finalized #0 (0xedb3…0fa9), ⬇ 2.3kiB/s ⬆ 2.5kiB/s
2023-11-29 13:36:25.208  INFO tokio-runtime-worker substrate: ⏩ Warping, Downloading state, 0.11 Mib (0 peers), best: #0 (0xedb3…0fa9), finalized #0 (0xedb3…0fa9), ⬇ 2.1kiB/s ⬆ 2.5kiB/s
[...]

Issue #2787, opened 2023-12-22
Enhance WarpSync zombienet tests with BEEFY validations


acatangiu:
BEEFY + (GRANDPA) WarpSync was enabled in https://github.com/paritytech/polkadot-sdk/pull/2689

Add zombienet Warp Sync regression tests (needs rebuilding db used in existing zombienet tests to include BEEFY state)

Done:
- [x] Added BEEFY support to kitchensink
- [x] Published [PR](https://github.com/paritytech/polkadot-sdk/pull/2854) that adds some basic BEEFY checks to the warp sync tests

Summarizing remaining work:

- [x] investigate the issue with `0002-validators-warp-sync` where BEEFY doesn't catch-up.
- [x] investigate this comment: https://github.com/paritytech/polkadot-sdk/pull/2854#discussion_r1447343237 => not needed anymore. It's related to https://github.com/paritytech/polkadot-sdk/issues/2568

serban300:
Experimented with multiple solutions. No matter what, we need a pretty big database (>50k blocks) in order to make sure that the headers sync doesn't finish before BEEFY starts. Otherwise the test wouldn't cover much.

I think the best strategy would be to generate a rococo-local database with > 50k blocks and with an epoch duration that isn't too small (for example 30-60 min) in order to make sure that BEEFY catches up fast with the chain.

Started a zombienet with this strategy. Will need to leave it running for a couple of days in order to generate a big enough db.

serban300:
Added BEEFY support to the kitchensink runtime. Now trying to extend the current warp sync tests (which were also built using the kitchensink runtime) with BEEFY. 

Generated a couple of snapshots and experimenting with them. Most of the tests work. Just having some issues with `0002-validators-warp-sync` where the chain doesn't finalize blocks if I start it from my snapshots, or sometimes bob doesn't even warp sync (it gets stuck with 0 peers - all peers refuse the connection). Not sure if it's something related to the way I generate the snapshots. Investigating.

serban300:
Found the cause of the issue where for `0002-validators-warp-sync` the chain doesn't finalize blocks, or doesn't warp sync. We need to generate the snapshot using `--state-pruning archive --blocks-pruning archive`. But still for `0002-validators-warp-sync` BEEFY doesn't catch-up.

I will shortly publish a PR that adds some basic BEEFY checks to the warp sync tests first. And I would like to investigate the remaining issue related to `0002-validators-warp-sync` separately.

[...]

PR #2854, opened 2024-01-04
[WIP] Add BEEFY finality to warp sync zombienet tests

serban300:
Will address https://github.com/paritytech/polkadot-sdk/issues/2787

For the moment opening this PR only to debug some zombienet features inside the CI


substrate/zombienet/0001-basic-warp-sync/README.md
```diff
@@ -10,17 +10,18 @@ The `dave` node executed with `--sync warp` syncs database with the rest of the
...
-chain = "gen-db"
+chain = "local"
 
 chain_spec_path = "chain-spec.json"
 
   [[relaychain.nodes]]
   name = "alice"
   validator = true
-
+  args = ["--state-pruning archive --blocks-pruning archive"]
```
acatangiu, line 23: why is this needed? to my understanding we should need neither state nor block bodies for warp syncing..

substrate/zombienet/0001-basic-warp-sync/generate-warp-sync-database.toml
```diff
@@ -1,16 +1,17 @@
...
 # refer to ./README.md for more details on how to create snapshot and spec
-chain = "gen-db"
+chain = "local"
 chain_spec_path = "chain-spec.json"
 
 
   [[relaychain.nodes]]
   name = "alice"
   validator = true
+  args = ["--state-pruning archive --blocks-pruning archive"]
```
acatangiu, line 14: same question here

acatangiu (approved):
I didn't go into detailed review of `chain-spec.json` changes.

Everything else looks good.

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: I'm not sure why it's needed, but without it the nodes from which the warp proof is downloaded generate some errors that go something like `State for block ... already discarded` when they are requested for warp sync fragments. And the warp sync freezes. But good point. I can investigate this separately. Will take a note.

serban300:
@michalkucharczyk  just adding you as a reviewer in case you want to take a look on this PR that brings some changes to the zombienet tests for warp sync, since judging by git blame looks like you added these tests

substrate/zombienet/0001-basic-warp-sync/generate-warp-sync-database.toml
serban300: Answered above

.gitlab/pipeline/zombienet/substrate.yml
```diff
@@ -50,23 +57,29 @@ zombienet-substrate-0000-block-building:
...
       --local-dir="${LOCAL_DIR}/0001-basic-warp-sync"
       --test="test-warp-sync.zndsl"
 
 zombienet-substrate-0002-validators-warp-sync:
   extends:
-    - .zombienet-substrate-common
+    - .zombienet-substrate-warp-sync-common
+  before_script:
+    - !reference [.zombienet-substrate-warp-sync-common, before_script]
+    - cp ${LOCAL_DIR}/0001-basic-warp-sync/chain-spec.json ${LOCAL_DIR}/0002-validators-warp-sync
```
michalkucharczyk, line 71: hm, running locally requires remembering about this copy. This can become frustrating. Maybe better do a link?

substrate/zombienet/0001-basic-warp-sync/README.md
michalkucharczyk: Do you see this log `State for block ... already discarded` every time?
I was investigating that some time ago, and it was happening sometimes (10%, maybe 20% depending on CPU load):
https://github.com/paritytech/polkadot-sdk/issues/2568

When preparing then database snapshot for warp-sync test we should keep all blocks, so adding this option could be desirable.

substrate/zombienet/0001-basic-warp-sync/README.md
```diff
@@ -32,33 +33,30 @@ The zombienet shall be executed with the following command, and run for some per
...
+data/chains/local_testnet/
+data/chains/local_testnet/db/
+data/chains/local_testnet/db/full/
+...
+data/chains/local_testnet/db/full/000469.log
 ```
 
 Also refer to: [zombienet#578](https://github.com/paritytech/zombienet/issues/578)
 
-The `raw` chain-spec shall also be saved: `db-test-gen/gen-db-raw.json`.
```
michalkucharczyk, line 60: We still need to keep raw chain-spec, right? I think keeping this note maybe worthy for future maintainers.

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: > Do you see this log `State for block ... already discarded` every time?

Yes, for me this was happening very often. Definitely more than 50% of the runs. Maybe also because the new db is bigger.

michalkucharczyk: Sorry, second thought: we don't actually need this params. It will indirectly fix #2568, but the problem will still be there. The database actually does not need to be full archive.

michalkucharczyk: Better workaround would be to forbid Bob to connect to Alice.

serban300: Sounds good. Do you know how this can be done ? IS there a config for it ? And would this prevent bob to connect to alice just for getting the warp sync ? Would they still be able to connect to each other in order to produce blocks ?

substrate/zombienet/0001-basic-warp-sync/test-warp-sync.zndsl
```diff
@@ -13,18 +13,16 @@ charlie: reports peers count is at least 3 within 60 seconds
...
 
 
-# db snapshot has 12133 blocks
-dave: reports block height is at least 1 within 60 seconds
-dave: reports block height is at least 12132 within 60 seconds
-dave: reports block height is at least 12133 within 60 seconds
+# db snapshot has {{DB_BLOCK_HEIGHT}} blocks
+alice: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
+bob: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
+charlie: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
```
michalkucharczyk, line 19: 12333 check meant that alice,bob,charlie are building blocks. I'd prefer to keep this check. Please add also `greater then`?

michalkucharczyk: Sorry, you are right, in this test it is not needed. I see that some numbers could be messed, so we can fix numbers (or use `greater then`) in tests with validators. Thanks.

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: Thanks @skunert ! Will try.

.gitlab/pipeline/zombienet/substrate.yml
michalkucharczyk, line 71: Is `cp` a leftover now?

serban300: Yes, sorry, removed

substrate/zombienet/0003-block-building-warp-sync/test-block-building-warp-sync.zndsl
```diff
@@ -13,18 +13,19 @@ charlie: reports peers count is at least 3 within 60 seconds
...
+# db snapshot has {{DB_BLOCK_HEIGHT}} blocks
+alice: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
+bob: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
+charlie: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
 
-alice: reports block height is at least 12133 within 60 seconds
-bob: reports block height is at least 12133 within 60 seconds
-charlie: reports block height is at least 12133 within 60 seconds
+dave: reports block height is at least 1 within 60 seconds
+dave: reports block height is at least {{DB_BLOCK_HEIGHT}} within 60 seconds
```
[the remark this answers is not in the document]
serban300: Done

substrate/zombienet/0003-block-building-warp-sync/test-block-building-warp-sync.zndsl
```diff
@@ -13,18 +13,20 @@ charlie: reports peers count is at least 3 within 60 seconds
...
```
michalkucharczyk, line 22: we may also want to check other validators are not stuck:
```suggestion
alice: reports block height is greater than {{DB_BLOCK_HEIGHT}} within 60 seconds
bob: reports block height is greater than {{DB_BLOCK_HEIGHT}} within 60 seconds
charlie: reports block height is greater than {{DB_BLOCK_HEIGHT}} within 60 seconds
```

serban300: Done

michalkucharczyk (approved):

michalkucharczyk:
Please also re-upload db-snapshot created w/o archive mode.

.gitlab/pipeline/zombienet/substrate.yml
serban300: Actually this is needed : https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4887696 .

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: > Please also re-upload db-snapshot created w/o archive mode.

@michalkucharczyk I just checked this and indeed it solves the warp sync issue, but the chain isn't finalizing blocks because bob can't talk to alice. Is it ok if I leave the archide db for the moment ? I added a comment that this is a workaround and that it should be removed after fixing https://github.com/paritytech/polkadot-sdk/issues/2568

.gitlab/pipeline/zombienet/substrate.yml
michalkucharczyk: Hm, could you explain why? We have a link there, right?

serban300: I guess it's because when we do `/home/nonroot/zombie-net/scripts/ci/run-test-local-env-manager.sh --local-dir="${LOCAL_DIR}/0001-basic-warp-sync"` only the `${LOCAL_DIR}/0002-validators-warp-sync` is made available to the test somehow (uploaded to the container or something). So even if the link is there, the `${LOCAL_DIR}/0001-basic-warp-sync` folder is not present

michalkucharczyk: makes sense. Thank you.

substrate/zombienet/0001-basic-warp-sync/README.md
```diff
@@ -10,17 +10,17 @@ The `dave` node executed with `--sync warp` syncs database with the rest of the
...
-chain = "gen-db"
+chain = "local"
 
 chain_spec_path = "chain-spec.json"
 
   [[relaychain.nodes]]
   name = "alice"
   validator = true
-
+
```
Lederstrumpf, line 23: apologies: whitespace nit 🙈

Lederstrumpf (approved):
lgtm; would be nice if [@skunert's comment](https://github.com/paritytech/polkadot-sdk/pull/2854#discussion_r1447447558) can be integrated (in a followup)

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: Fixed

substrate/zombienet/0001-basic-warp-sync/README.md
serban300: @michalkucharczyk sorry for the ping. WDYT ? Would you have any concern about merging this PR as it is and then switching to a pruned snapshot in a future PR? After https://github.com/paritytech/polkadot-sdk/issues/2568 is fixed ?

michalkucharczyk: I'd prefer to have DB without archive mode (because this is more close to scenario used widely in production).

But yeah, let's use archive DB here. I have some ideas for fixing #2568. Please also leave comment there (just not to forget about this). I would also appreciate if you leave the actual toml file that was used for generating DB in repo  (with appropriate comment).

Thank you.

serban300: Ok thanks ! Done:

- Added comment to #2568 here: https://github.com/paritytech/polkadot-sdk/issues/2568#issuecomment-1886745731
- The toml file that was used for generating DB is already updated and it also has a TODO comment about removing the workaround

michalkucharczyk: ty!

serban300:
Yes, looks like we should increase the timeout. Will do

PR #2856, opened 2024-01-05
Kitchensink chain: Add BEEFY support

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/2787

Adding BEEFY support to the kitchensink chain in order to be able to extend the current warp sync zombienet tests with BEEFY enabled


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4847371

substrate/bin/node/cli/src/cli.rs
```diff
@@ -41,6 +41,12 @@ pub struct Cli {
 	#[arg(long)]
 	pub no_hardware_benchmarks: bool,
 
+	/// Disable the BEEFY gadget.
+	///
+	/// Currently enabled by default.
+	#[arg(long)]
+	pub no_beefy: bool,
```
acatangiu, line 48: I don't think we need this, kitchensink can always use BEEFY

substrate/bin/node/cli/src/service.rs
```diff
@@ -357,11 +382,27 @@ pub fn new_full_base(
 	config: Configuration,
 	mixnet_config: Option<sc_mixnet::Config>,
 	disable_hardware_benchmarks: bool,
+	disable_beefy: bool,
```
acatangiu, line 385: same here

```suggestion
```

acatangiu (approved):

substrate/bin/node/cli/src/cli.rs
serban300: I don't know what would be better. I saw that there is also the option to disable grandpa in kitchensink and I thought that it would be good to also have `--no-beefy`. Personally I would keep it, but I don't have strong preferences. If you think it's better to remove it, I can remove it.

acatangiu: the grandpa one is an old relic which was also planned to be removed (but low on prio list so it never gets done :P)

serban300: Oh, ok. Removed `-no-beefy` then.

substrate/bin/node/runtime/src/lib.rs
```diff
@@ -1573,6 +1578,30 @@ impl pallet_mmr::Config for Runtime {
...
+	/// - `major` (3 bits)
+	/// - `minor` (5 bits)
+	///
+	/// `major` should be updated only if decoding the previous MMR Leaf format from the payload
+	/// is not possible (i.e. backward incompatible change).
+	/// `minor` should be updated if fields are added to the previous MMR Leaf, which given SCALE
+	/// encoding does not prevent old leafs from being decoded.
+	///
+	/// Hence we expect `major` to be changed really rarely (think never).
+	/// See [`MmrLeafVersion`] type documentation for more details.
```
bkchr, line 1594: Why is this documentation copied everywhere? What is the purpose? This should be above the `MmrLeafVersion` type, if this isn't there yet.

bkchr (approved):

substrate/bin/node/runtime/src/lib.rs
serban300: Sorry, I copied it blindly. Fixed.

PR #2862, opened 2024-01-05
Return latest known relay chain block number in `on_initialize` etc

bkchr:
This changes `RelaychainDataProvider` to return the latest known relay chain block number in `on_initialize` et all, aka when the `validation_data` wasn't yet set by the inherent.

### DIFF
--- a/cumulus/pallets/parachain-system/src/lib.rs
+++ b/cumulus/pallets/parachain-system/src/lib.rs
@@ -1709,7 +1709,7 @@ impl<T: Config> BlockNumberProvider for RelaychainDataProvider<T> {
 	fn current_block_number() -> relay_chain::BlockNumber {
 		Pallet::<T>::validation_data()
 			.map(|d| d.relay_parent_number)
-			.unwrap_or_default()
+			.unwrap_or_else(|| Pallet::<T>::last_relay_block_number())
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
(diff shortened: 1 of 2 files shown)


cumulus/pallets/parachain-system/src/lib.rs
```diff
@@ -1683,14 +1683,27 @@ pub trait RelaychainStateProvider {
...
-impl<T: Config> BlockNumberProvider for RelaychainBlockNumberProvider<T> {
+/// Implements [`BlockNumberProvider`] and [`RelaychainStateProvider`] that returns relevant relay
+/// data fetched from validation data.
+///
+/// NOTE: When validation data is not available (e.g. within `on_initialize`):
+///
+/// - [`current_relay_chain_state`](Self::current_relay_chain_state): Will return the default value
+///   of [`RelayChainState`].
+/// - [`current_block_number`](Self::current_block_number): Will return
+///   [`Pallet::last_relay_block_number()`].
```
seadanda, line 1703: This would still return zero if validation data is `None`.

Should the below instead be something like:

```
	fn current_block_number() -> relay_chain::BlockNumber {
		Pallet::<T>::validation_data()
			.map(|d| d.relay_parent_number)
			.unwrap_or_else(|| LastRelayChainBlockNumber::<T>::get())
	}
```

bkchr: Am I dumb? 🤔 Did I just document this properly and didn't do the code change? 👀

bkchr: But you are right. 😬

seadanda (approved):

PR #2868, opened 2024-01-07
NFTs: move deposit to new owner on trade

### LINKED ISSUE #2851: NFT: Question, why is the deposit not transferred to the new owner when buying an NFT?
Hi, so when we set some metadata for an NFT we reserve some funds depending on the length of the metadata, and the same happens for setting attributes.

When someone buys an NFT they pay whatever price the seller has set. But I find it strange that even after selling an NFT, the original owner still has to pay the metadata and attribute deposit. It makes sense to me that when someone sells an NFT their reserved tokens are freed and the same amount of buyer's tokens are reserved. I am wondering why that is not the case.

Thanks

SCJangra:
Closes #2851

Move item deposit and metadata deposit to the new owner when buying an NFT.

This currently does not move the attribute deposit because there can be any number of attributes on an item, and I am not sure about looping through all of them. One solution could be to ask the seller which attributes are moved in `set_price` extrinsic. I am open to other suggestions as well.


SCJangra:
@jsidorenko Could I get some input on this?

jsidorenko:
> @jsidorenko Could I get some input on this?

Sorry, I had a crazy week. I will check your PR first thing on Monday!

SCJangra:
@jsidorenko No problem at all!

jsidorenko:
I've checked the logic - it looks correct! A good catch on checking and touching the collection's owner data ;)
I have some comments on the syntax. Do you mind if I push the fixes straight into your PR? Or I can fork your fork and open a separate PR :) or just write as a code suggestion..

SCJangra:
@jsidorenko Feel free to push directly to this PR!

jsidorenko:
> @jsidorenko Feel free to push directly to this PR.

Done, I've extracted your changes into a separate method and connected to atomic swaps

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5020262

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5020440

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5020422

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5020441

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5020439

SCJangra:
Thank you. Just a quick question: you mentioned that, earlier the deposit was moved but later it was changed. Was the attribute deposit also moved then?

jsidorenko:
Nope, only the main item deposit got moved.
In general, moving the attribute deposits could become a heavy operation if there are a lot of attributes. I think, if that'd become a real UX issue, we should add a new extrinsic for such transfer. As you've suggested above it could follow the "propose -> accept" model like we have with collections.

jsidorenko (approved):

SCJangra:
Hi @jsidorenko, any particular reason why this is not approved? is it blocked by something else?

substrate/frame/nfts/src/features/transfer.rs
```diff
@@ -231,4 +232,68 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		new_owner: T::AccountId,
+		collection_details: &mut CollectionDetailsFor<T, I>,
+		item_details: &mut ItemDetailsFor<T, I>,
+	) -> DispatchResult {
+		let ItemDepositOf::<T, I> { amount: item_deposit, account: item_depositor } =
+			item_details.deposit.clone();
+
+		if new_owner != item_depositor {
+			T::Currency::unreserve(&item_depositor, item_deposit);
+			T::Currency::reserve(&new_owner, item_deposit)?;
```
gui1117, line 261: Maybe it is common practice now, but this means that the caller must rollback the storage if it wants to handle the error.

PR #2870, opened 2024-01-07
pallet-contracts-fixtures: Only build RISCV when the feature is enabled

bkchr:
This disables building the RISCV fixtures by default. They still require a custom build rustc version. When there are actual tests for RISCV, the feature can be enabled in CI. Also fixes some unwraps that assume again that people are using rustup...

### DIFF
--- a/substrate/frame/contracts/fixtures/Cargo.toml
+++ b/substrate/frame/contracts/fixtures/Cargo.toml
@@ -23,3 +23,7 @@ toml = "0.8.2"
 twox-hash = "1.6.3"
 polkavm-linker = "0.3.0"
 anyhow = "1.0.0"
+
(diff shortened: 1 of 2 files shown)


seadanda (approved):

ggwpez (approved):

pgherveou (approved):
> Also fixes some unwraps that assume again that people are using rustup...

Anyway we can make it fail on CI so it does not happen again?

bkchr:
> > Also fixes some unwraps that assume again that people are using rustup...
> 
> Anyway we can make it fail on CI so it does not happen again?

I mean it is not that annoying yet, that it is worth the time to setup the CI in this way 😬

PR #1768, opened 2023-10-02
Add `asset_sufficient` getter for fungibles

pandres95:
# Description

This Pull Request introduces a method inside the `fungibles::Inspect` trait called `asset_sufficient`. This method is useful since pallet implementors can benefit from having this (so far missing) piece of information, either for testing checks or for handling accounts touching on behalf of others.

### Use Case: Checking whether an asset was appropriately created as sufficient

A pallet might have a rule that allows creating assets, but only marking one of these (i.e. the first) as sufficient. A typical test for this would resemble something like this

```rs
#[test]
fn it_works() {
    new_test_ext().execute_with(|| {
        setup();

        // Can register a new asset
        assert_ok!(
            Communities::create_asset(RuntimeOrigin::signed(COMMUNITY_ADMIN), COMMUNITY, ASSET_B, 1)
        );

        // Can register additional assets
        assert_ok!(
            Communities::create_asset(RuntimeOrigin::signed(COMMUNITY_ADMIN), COMMUNITY, ASSET_C, 1)
        );

        // First asset owned by the community is sufficient by default
        assert_sufficiency(COMMUNITY, ASSET_B, 1, true);

        // Additional assets owned by the community are not sufficient
        // by default
        assert_sufficiency(COMMUNITY, ASSET_C, 1, false);
    });
}
```

Without this getter, implementing `assert_sufficiency` would require (as shown [here](https://github.com/virto-network/virto-node/blob/d7707099d811d94dbb0a13d2b2098c9659446615/pallets/communities/src/tests/helpers.rs#L16-L50)) the following steps:

1. Forcing access to the storage via `sp_io::storage::get`.
2. Encoding a `AssetDetails` struct that resembles the expected one, since `is_sufficient` is a private field.
3. Asserting equality for both structs.

With the getter, it would be something as simple as this, no further illegal accesses required:

```rs
        // First asset owned by the community is sufficient by default
        assert_eq(Assets::asset_sufficient(ASSET_B), Some(true));

        // Additional assets owned by the community are not sufficient
        // by default
        assert_eq(Assets::asset_sufficient(ASSET_C), Some(false));
```

## Implementation

The implementation is actually really simple.

1. Define the method under the `fungibles::Inspect` trait. The method signature conveniently returns `Option<bool>`: to be `None` implies that the asset does not exist, so it's also a shortcut for `asset_exists(asset) && asset_sufficient(asset)`, reducing code.
[...]

PR #2872, opened 2024-01-07
feat(pallet-assets): define and implement sufficients traits for `pallet-assets`

pandres95:
This pull request supersedes #1768, and implements methods to access and mutate `is_sufficient` value of an asset.

---

# Description

This Pull Request introduces two traits under `pallet-asset`: a trait called `sufficients::Inspect`, and a trait called `sufficients::Mutate`. These methods are useful since pallet implementors can benefit from having this (so far missing) piece of information, either for testing checks or for handling accounts touching on behalf of others.

### Use Case: Checking whether an asset was appropriately created as sufficient

A pallet might have a rule that allows creating assets, but only marking one of these (i.e. the first) as sufficient. A typical test for this would resemble something like this

```rs
#[test]
fn it_works() {
    new_test_ext().execute_with(|| {
        setup();

        // Can register a new asset
        assert_ok!(
            Communities::create_asset(RuntimeOrigin::signed(COMMUNITY_ADMIN), COMMUNITY, ASSET_B, 1)
        );

        // Can register additional assets
        assert_ok!(
            Communities::create_asset(RuntimeOrigin::signed(COMMUNITY_ADMIN), COMMUNITY, ASSET_C, 1)
        );

        // First asset owned by the community is sufficient by default
        assert_sufficiency(COMMUNITY, ASSET_B, 1, true);

        // Additional assets owned by the community are not sufficient
        // by default
        assert_sufficiency(COMMUNITY, ASSET_C, 1, false);
    });
}
```

Without this getter, implementing `assert_sufficiency` would require (as shown [here](https://github.com/virto-network/virto-node/blob/d7707099d811d94dbb0a13d2b2098c9659446615/pallets/communities/src/tests/helpers.rs#L16-L50)) the following steps:

1. Forcing access to the storage via `sp_io::storage::get`.
2. Encoding a `AssetDetails` struct that resembles the expected one, since `is_sufficient` is a private field.
3. Asserting equality for both structs.

With the getter, it would be something as simple as this, no further illegal accesses required:

```rs
        // First asset owned by the community is sufficient by default
        assert_eq(Assets::is_sufficient(ASSET_B), true);

        // Additional assets owned by the community are not sufficient
        // by default
        assert_eq(Assets::is_sufficient(ASSET_C), false);
```

### Use Case: Checking sufficiency for an existing asset

Some pallets might see it useful to check whether an asset is sufficient, for some cases where it might be useful (e.g. some restricted versions of DEXes where pool implementors would like to check that those assets are sufficient).

Without the getter, this would require building the pallet as an extension of the Assets pallet. Some tricks like the one explained in the use case above would be impractical.

## Implementation

The implementation is actually really simple.

1. Define the traits under `substrate/frame/pallets/assets/src/traits.rs`: `sufficients::Inspect` and `sufficients::Mutate`. The method signature returns `bool`, defaults to `false` when the asset does not exist.
2. Implement on the `impl_sufficients.rs` file, getting the asset's details, and mapping the response as `asset.is_sufficient`.

# Checklist

  required)
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)

### DIFF
--- a/substrate/frame/assets/src/functions.rs
+++ b/substrate/frame/assets/src/functions.rs
@@ -998,7 +998,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		})
 	}
 
-	/// Do set sufficiency
+	/// Set the sufficiency of an asset class
+	///
+	/// ### Errors
+	///
+	/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
 	pub(super) fn do_set_sufficiency(asset_id: T::AssetId, is_sufficient: bool) -> DispatchResult {
 		Asset::<T, I>::try_mutate(asset_id, |maybe_asset| {
 			if let Some(asset) = maybe_asset {
--- a/substrate/frame/assets/src/functions.rs
+++ b/substrate/frame/assets/src/functions.rs
@@ -998,6 +998,18 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		})
 	}
 
+	/// Do set sufficiency
+	pub(super) fn do_set_sufficiency(asset_id: T::AssetId, is_sufficient: bool) -> DispatchResult {
+		Asset::<T, I>::try_mutate(asset_id, |maybe_asset| {
+			if let Some(asset) = maybe_asset {
+				asset.is_sufficient = is_sufficient;
+				Ok(())
+			} else {
+				Err(Error::<T, I>::Unknown)?
+			}
+		})
+	}
+
 	/// Calculate the metadata deposit for the provided data.
 	pub(super) fn calc_metadata_deposit(name: &[u8], symbol: &[u8]) -> DepositBalanceOf<T, I> {
 		T::MetadataDepositPerByte::get()
--- a/substrate/frame/assets/src/impl_sufficiency.rs
+++ b/substrate/frame/assets/src/impl_sufficiency.rs
@@ -1,5 +1,5 @@
 use crate::{
-	traits::sufficients::{Inspect, Mutate},
+	traits::sufficiency::{Inspect, Mutate},
 	Asset, Config, Pallet,
 };
 
--- a/substrate/frame/assets/src/impl_sufficiency.rs
+++ b/substrate/frame/assets/src/impl_sufficiency.rs
@@ -1,15 +1,34 @@
+// This file is part of Substrate.
+
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Assets pallet's `StoredMap` implementation.
+
 use crate::{
-	traits::sufficiency::{Inspect, Mutate},
+	traits::sufficiency::{IsSufficient, SetSufficiency},
 	Asset, Config, Pallet,
 };
 
-impl<T: Config<I>, I: 'static> Inspect<<T as Config<I>>::AssetId> for Pallet<T, I> {
+impl<T: Config<I>, I: 'static> IsSufficient<<T as Config<I>>::AssetId> for Pallet<T, I> {
 	fn is_sufficient(asset_id: <T as Config<I>>::AssetId) -> bool {
 		Asset::<T, I>::get(asset_id).map(|asset| asset.is_sufficient).unwrap_or(false)
 	}
 }
 
-impl<T: Config<I>, I: 'static> Mutate<<T as Config<I>>::AssetId> for Pallet<T, I> {
+impl<T: Config<I>, I: 'static> SetSufficiency<<T as Config<I>>::AssetId> for Pallet<T, I> {
 	fn make_sufficient(asset_id: <T as Config<I>>::AssetId) -> sp_runtime::DispatchResult {
 		Pallet::<T, I>::do_set_sufficiency(asset_id, true)
 	}
--- a/substrate/frame/assets/src/traits.rs
+++ b/substrate/frame/assets/src/traits.rs
@@ -1,4 +1,4 @@
-pub mod sufficients {
+pub mod sufficiency {
 	use sp_runtime::DispatchResult;
 
 	/// Trait for providing the sufficient state of an asset.
--- a/substrate/frame/assets/src/traits.rs
+++ b/substrate/frame/assets/src/traits.rs
@@ -3,17 +3,24 @@ pub mod sufficients {
 
 	/// Trait for providing the sufficient state of an asset.
 	pub trait Inspect<AssetId> {
-		/// Returns whether an asset is sufficient or not
+		/// Returns whether an asset is sufficient or not.
 		fn is_sufficient(asset_id: AssetId) -> bool;
 	}
 
 	/// Trait for mutating the sufficient state of an asset
 	pub trait Mutate<AssetId> {
-		/// Sets the `is_sufficient` value of an asset.
+		/// Makes the asset to be sufficient.
 		///
 		/// ### Errors
 		///
 		/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
-		fn set_sufficient(asset_id: AssetId, is_sufficient: bool) -> DispatchResult;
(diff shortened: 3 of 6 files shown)


pandres95:
Per @muharem's [suggestion][1], I decided to close #1768 and continue here. Ready to gather feedback (I know this PR might come with some controversy).

[1]: https://github.com/paritytech/polkadot-sdk/pull/1768#discussion_r1366792128

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	/// Trait for mutating the sufficient state of an asset
+	pub trait Mutate<AssetId> {
+		/// Sets the `is_sufficient` value of an asset.
+		///
+		/// ### Errors
+		///
+		/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
+		fn set_sufficient(asset_id: AssetId, is_sufficient: bool) -> DispatchResult;
```
joepetrowski, line 17: I would either name this as such or split into two functions: `make_sufficient(asset_id)` and `make_insufficient(asset_id)`.
```suggestion
		fn set_sufficiency(asset_id: AssetId, is_sufficient: bool) -> DispatchResult;
```

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,19 @@
+pub mod sufficients {
+	use sp_runtime::DispatchResult;
+
+	/// Trait for providing the sufficient state of an asset.
+	pub trait Inspect<AssetId> {
+		/// Returns whether an asset is sufficient or not
```
joepetrowski, line 6:
```suggestion
		/// Returns whether an asset is sufficient or not.
```

substrate/frame/assets/src/traits.rs
pandres95: I like the `make_sufficient`/`make_insufficient` approach better. While a bit verbose now, it might fit better in a situation where something is needed to be done on each change (i.e. should anything happen when making a sufficient asset insufficient?).

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,26 @@
...
+
+	/// Trait for providing the sufficient state of an asset.
+	pub trait Inspect<AssetId> {
+		/// Returns whether an asset is sufficient or not.
+		fn is_sufficient(asset_id: AssetId) -> bool;
+	}
+
+	/// Trait for mutating the sufficient state of an asset
+	pub trait Mutate<AssetId> {
+		/// Makes the asset to be sufficient.
```
joepetrowski, line 12:
```suggestion
		/// Makes the asset sufficient.
```

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,26 @@
...
+		/// ### Errors
+		///
+		/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
+		fn make_sufficient(asset_id: AssetId) -> DispatchResult;
+
+		/// Makes the asset to be insufficient.
+		///
+		/// ### Errors
+		///
+		/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
```
joepetrowski, line 23: Implementation specific.

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,26 @@
+pub mod sufficients {
+	use sp_runtime::DispatchResult;
+
+	/// Trait for providing the sufficient state of an asset.
```
joepetrowski, line 4:
```suggestion
	/// Trait for providing the sufficiency of an asset.
```

substrate/frame/assets/src/functions.rs
```diff
@@ -998,6 +998,18 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		})
 	}
 
+	/// Do set sufficiency
```
joepetrowski, line 1001:
```suggestion
	/// Set the sufficiency of an asset class.
	///
	/// ### Errors
	///
	/// - [`Unknown`][crate::Error::Unknown] when the asset ID is unknown.
```

joepetrowski: This is where you should document the error case.

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,26 @@
+pub mod sufficients {
```
joepetrowski, line 1: `sufficients` isn't really a good name because the mod has nothing to do with the behavior of sufficient assets, but merely setting the `sufficient` property to `true` or `false`.

I would therefore call it `sufficiency`.

I also question whether this should be in the Assets pallet at all and not frame-support like the [Account Touch](https://github.com/paritytech/polkadot-sdk/blob/0b74812/substrate/frame/support/src/traits/misc.rs#L1181) trait.

joepetrowski: @muharem and @liamaharon have done some recent work on this pallet, maybe they can comment.

pandres95: I get it. Just resolved all the other changes.

As for whether `sufficiency` should be on Assets pallet and not on `frame-support`, my hint was this @muharem's [comment](https://github.com/paritytech/polkadot-sdk/pull/1768#discussion_r1366792128):

> not sure if this concept is general enough, I cannot really say.

This is because this concept might not be general enough, but is somewhat specific to Assets. Instead, the most general concept, actually, is the [`AccountTouch`](https://github.com/paritytech/polkadot-sdk/blob/0b74812/substrate/frame/support/src/traits/misc.rs#L1181) trait.

> but an alternative can be a separate trait on Assets pallet level, which lets you inspect some additional characteristics of an asset.

That's what I based on to close #1768 and open this PR instead. 😁

substrate/frame/assets/src/impl_sufficiency.rs
```diff
@@ -0,0 +1,20 @@
+use crate::{
```
muharem, line 1: maybe `traits` and `impls`? otherwise the list of files will grow here pretty quick. 
I also have only strong opinion on keeping our folders/files/modules style consistent, since there is no yet agreement on it, I would just try to keep it close to what we have now, and now we tend to have less files than many.

muharem, line 1: new files missing the header

substrate/frame/assets/src/traits.rs
```diff
@@ -0,0 +1,18 @@
...
+	use sp_runtime::DispatchResult;
+
+	/// Trait for providing the sufficiency of an asset.
+	pub trait Inspect<AssetId> {
+		/// Returns whether an asset is sufficient or not.
+		fn is_sufficient(asset_id: AssetId) -> bool;
+	}
+
+	/// Trait for mutating the sufficiency of an asset
+	pub trait Mutate<AssetId> {
```
muharem, line 11: I would try to come up with a different namings for new traits. otherwise it will create more confusions. also I see the possibility for these traits moving support module

muharem (commented):
I clearly see use cases for a trait to read the is_sufficient flag. But I am not fully agree with the use case you giving in the description - "Checking whether an asset was appropriately created as sufficient". When it is tests within the package, I think it is fine to access the internals of a package (e.g. storage). And you should not need to test it in that way from outside.

I also would like to question if you have a real use case, where you need to alter is_sufficient flag. if you look to the public api of the pallet, the altering api - `force_asset_status` has more parameters  then just one flag. I can easily imagine someone coming and introducing new function with all those parameters. What I am trying to say, it is good to have real use cases for introducing such contracts. And not trying to say that one better than another.

substrate/frame/assets/src/impl_sufficiency.rs
pandres95: Would this mean also moving `impl_fungibles` and `impl_stored_map` onto an `impls` file/folder? I see the point on making the file structure lighter, but seeing that other impls are not that light (i.e. `impl_fungibles` is 311 lines) a bit of _not-flatness_ might be good for the sake of readibility from time to time.

My suggestion:

```
impls/
├─ fungibles.rs
├─ stored_maps.rs
├─ sufficiency.rs
├─ mod.rs
traits.rs
```

pandres95:
Reply to https://github.com/paritytech/polkadot-sdk/pull/2872#discussion_r1446871916
> also I see the possibility for these traits moving support module
 
I see it's possible to come up with a more general solution, by considering a more general trait that works alongside `CanTouch` and would be `CanProvide`/`SetProvider` (a hook).

This would general enough, and would enable every possible entity out there to be a provider or set that possibility. Now, I'm not sure of whether that would be achievable without some ambiguity (on entity ID, for example) so I think sticking to sufficiency for now would help @muharem. Maybe if we get enough feedback on a way around, I could move forward with a more general solution that can go on `frame-support` in the future.

pandres95:
Closing due to inactivity.

If someone is interested about this proposal in the future, just ping me.

PR #2886, opened 2024-01-09
Remove bounds from `PrevalidateAttests` struct definition

ggwpez:
Removing some bounds as it came up in https://github.com/paritytech/polkadot-sdk/pull/2726 while still keeping `Send + Sync` capabilities.

### DIFF
--- a/polkadot/runtime/common/src/claims.rs
+++ b/polkadot/runtime/common/src/claims.rs
@@ -1467,12 +1467,6 @@ mod tests {
 			);
 		});
 	}
-
-	#[test]
-	fn prevalidate_assets_is_send_and_sync() {
-		fn is_send_and_sync<T: Send + Sync>() {}
-		is_send_and_sync::<PrevalidateAttests<Test>>();
-	}
 }
 
 #[cfg(feature = "runtime-benchmarks")]
--- a/polkadot/runtime/common/src/claims.rs
+++ b/polkadot/runtime/common/src/claims.rs
@@ -591,11 +591,9 @@ impl<T: Config> Pallet<T> {
 /// otherwise free to place on chain.
 #[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
 #[scale_info(skip_type_params(T))]
-pub struct PrevalidateAttests<T: Config + Send + Sync>(sp_std::marker::PhantomData<T>)
-where
-	<T as frame_system::Config>::RuntimeCall: IsSubType<Call<T>>;
+pub struct PrevalidateAttests<T>(core::marker::PhantomData<fn(T)>);
 
-impl<T: Config + Send + Sync> Debug for PrevalidateAttests<T>
+impl<T: Config> Debug for PrevalidateAttests<T>
 where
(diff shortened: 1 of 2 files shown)


georgepisaltu (approved):

polkadot/runtime/common/src/claims.rs
```diff
@@ -591,11 +591,9 @@ impl<T: Config> Pallet<T> {
 /// otherwise free to place on chain.
 #[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
 #[scale_info(skip_type_params(T))]
-pub struct PrevalidateAttests<T: Config + Send + Sync>(sp_std::marker::PhantomData<T>)
-where
-	<T as frame_system::Config>::RuntimeCall: IsSubType<Call<T>>;
+pub struct PrevalidateAttests<T>(core::marker::PhantomData<fn(T)>);
```
bkchr, line 594: Not sure we really need this :P

bkchr (approved):

polkadot/runtime/common/src/claims.rs
ggwpez: WDYM?

bkchr: Yes

ggwpez: But if i use `PhantomData<T>` then its not `Send + Sync` and if i remove the PD then it errors `parameter T is never used`.  
Signed Extension does require it though.

bkchr: Why does it needs to be Send and Sync?

ggwpez: The `SignedExtension` trait [requires](https://github.com/paritytech/polkadot-sdk/blob/9d80735a7cccb7c52280ea057177d92e17203c51/substrate/primitives/runtime/src/traits.rs#L1477) it, i did not check why.  
The new TEs will also [require](https://github.com/paritytech/polkadot-sdk/blob/e8bbdeb6dd6a55455dc31669351cf8acd602f7c9/substrate/primitives/runtime/src/traits/transaction_extension/mod.rs#L48) it.

polkadot/runtime/common/src/claims.rs
```diff
@@ -620,7 +618,7 @@ where
 	}
 }
 
-impl<T: Config + Send + Sync> SignedExtension for PrevalidateAttests<T>
+impl<T: Config> SignedExtension for PrevalidateAttests<T>
```
bkchr, line 621: If you would just kept the bounds here, it would have been enough. 

But we can go with the phantomdata trick.

polkadot/runtime/common/src/claims.rs
```diff
@@ -1469,6 +1467,12 @@ mod tests {
 			);
 		});
 	}
+
+	#[test]
+	fn prevalidate_assets_is_send_and_sync() {
+		fn is_send_and_sync<T: Send + Sync>() {}
+		is_send_and_sync::<PrevalidateAttests<Test>>();
+	}
```
bkchr, line 1475:
```suggestion
```

Not a requirement.

Issue #2399, opened 2023-11-19
PVF: Re-check file integrity before voting against; document


mrcnski:
Random thought, could we add an `# Artifact integrity` section here? It could explain that artifacts must remain valid to satisfy the `SAFETY` constraints of `execute_artifact`. Therefore we do two things: check file integrity and version compatibility on host startup, and also before voting against a candidate we re-check the file integrity. cc @eagr

_Originally posted by @mrcnski in https://github.com/paritytech/polkadot-sdk/pull/1918#discussion_r1390940354_

Jpserrat:
@mrcnski Is this issue about re-checking the file integrity and adding the `# Artifact integrity`?
Why the re-checking is needed? Do we not vote against it in case of failure?

mrcnski:
> @mrcnski Is this issue about re-checking the file integrity and adding the # Artifact integrity?

Yes, exactly! We still need to do both, re-checking was not implemented yet.

> Why the re-checking is needed? Do we not vote against it in case of failure?

We had the idea to re-check the file hash of the artifact before execution, to make sure it wasn't corrupted. However, this would add a few ms to each execution. As an optimization, we can only re-check after execution, if the execution failed. Reason being, that corruption should be very very rare, and if it happens then the execution will fail, so we only need to check it there. Hope that makes sense. Thanks for asking, for sure it's not obvious so should be documented. 👍

mrcnski:
If we re-check and it's corrupted, we should return a specific error that indicates that the host should purge this artifact from the artifacts cache.

eagr:
I'm about to take a shot at the integrity re-check. Wondering if you have started on this? @Jpserrat

Jpserrat:
@eagr I haven't, you can take it

eskimor:
>  Do we not vote against it in case of failure?

We do and this is precisely the problem. If the file got corrupted on disk, we would currently vote invalid on the candidate and raise a dispute, despite the issue being our hard disk and not the candidate. We would get slashed and disabled for raising a dispute on a valid candidate. Obviously there should be a big fat warning in the logs if a corruption is detected.

PR #2895, opened 2024-01-09
PVF: Remove artifact persistence across restarts

mrcnski:
Considering the complexity of https://github.com/paritytech/polkadot-sdk/pull/2871 and the discussion therein, as well as the further complexity introduced by the hardening in https://github.com/paritytech/polkadot-sdk/pull/2742, as well as the eventual replacement of wasmtime by PolkaVM, it seems best to remove this persistence as it is creating more problems than it solves.

## Related

Closes https://github.com/paritytech/polkadot-sdk/issues/2863


polkadot/node/core/pvf/src/prepare/worker_interface.rs
```diff
@@ -232,8 +229,14 @@ async fn handle_response(
...
-	let artifact_id = ArtifactId::from_pvf_prep_data(pvf);
-	let artifact_path = artifact_id.path(cache_path, &checksum);
+	let unique_id = {
+		use rand::RngCore;
+		let mut bytes = [0u8; 64];
+		rand::thread_rng().fill_bytes(&mut bytes);
+		hex::encode(&bytes)
+	};
+	let file_name = format!("{}_0x{}", unique_id, checksum);
+	let artifact_path = cache_path.join(file_name);
```
mrcnski, line 239: This is now the only place that the path is created. It gets returned to the host and stored in the `Artifacts` table.

I kept the checksum as it will still be needed by execution for https://github.com/paritytech/polkadot-sdk/issues/2399 - but now that I look at it, probably better to instead add the checksum to `ArtifactPathId`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4876949

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4876966

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4876968

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4877335

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4877334

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4877336

s0me0ne-unkn0wn (approved):
We need more "+38 -379" PRs!

bkchr (approved):

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -193,121 +150,18 @@ impl Artifacts {
...
-			if err.kind() != io::ErrorKind::AlreadyExists {
-				return Err(format!("failed to create dir {cache_path:?}: {err}"))
-			}
-		}
-
-		let mut dir = tokio::fs::read_dir(cache_path)
-			.await
-			.map_err(|err| format!("failed to read dir {cache_path:?}: {err}"))?;
+		// First delete the entire cache. Nodes are long-running so this should populate shortly.
+		let _ = tokio::fs::remove_dir_all(cache_path).await;
```
koute, line 161: Instead of `remove_dir_all` could we have a preset unique file extension for the cached files (e.g. maybe `.pvfcache`) and then over a directory's contents and only delete files with that given extension? That should only be a couple of extra lines but would be a lot safer if something goes wrong with where the `cache_path` points to.

`remove_dir_all` always makes me nervous because it's essentially an `rm -Rf`, so if for some reason `cache_path` would be something that also contains other stuff (e.g. in the worst case someone would want to have a cache in their `/home/$user` :P) this would also unnecessarily delete all of it.

mrcnski: Another very sensible suggestion. :) Should we try to clean up any existing artifacts with the current naming scheme? e.g. delete any files starting with `wasmtime_` which would be compatible with the [current](https://github.com/paritytech/polkadot-sdk/pull/1918/files#diff-8b70b8fcbc7a957f4e0cca743522649ffc8d2ee24ef49531e8318392bdffddbbR74) scheme as well as the [one before artifact persistence](https://github.com/paritytech/polkadot-sdk/pull/1918/files#diff-8b70b8fcbc7a957f4e0cca743522649ffc8d2ee24ef49531e8318392bdffddbbL78). We also keep the self-contained "worker dirs" in there, which should be deleted:

Screenshot 2024-01-10 at 08 53 59

Maybe not ideal, but indeed seems better than nuking the whole directory.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -193,121 +150,18 @@ impl Artifacts {
...
+		let _ = tokio::fs::create_dir_all(cache_path).await;
```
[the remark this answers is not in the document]
mrcnski: We should make sure the dir exists no? Or does it look like there are two lines with `remove_dir_all`? 🙃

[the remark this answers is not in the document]
bkchr: Ohh, apparently I did not read correctly :sweat_smile:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4884206

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -194,120 +171,35 @@ impl Artifacts {
...
+
+		// Delete any leftover artifacts and worker dirs from previous runs. We don't delete the
+		// entire cache directory in case the user made a mistake and set it to e.g. their home
+		// directory. This is a best-effort to do clean-up, so ignore any errors.
+		if let Ok(paths) = fs::read_dir(cache_path) {
+			for path in paths.flat_map(|res| res.map(|e| e.path())) {
+				let file_name = match path.file_name().and_then(|f| f.to_str()) {
+					Some(f) => f,
+					None => continue,
+				};
```
koute, line 191:
```suggestion
				let Some(file_name) = path.file_name().and_then(|f| f.to_str()) else { continue };
```

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -194,120 +171,35 @@ impl Artifacts {
...
-		if let Err(err) = tokio::fs::create_dir_all(cache_path).await {
-			if err.kind() != io::ErrorKind::AlreadyExists {
-				return Err(format!("failed to create dir {cache_path:?}: {err}"))
+		let _ = tokio::fs::create_dir_all(cache_path).await;
+
+		// Delete any leftover artifacts and worker dirs from previous runs. We don't delete the
+		// entire cache directory in case the user made a mistake and set it to e.g. their home
+		// directory. This is a best-effort to do clean-up, so ignore any errors.
+		if let Ok(paths) = fs::read_dir(cache_path) {
+			for path in paths.flat_map(|res| res.map(|e| e.path())) {
```
koute, line 187: You can reduce the level of indentation and simplify this a bit with something like this:

```rust
for entry in fs::read_dir(cache_path).into_iter().flatten().flatten() {
    let path = entry.path();
```

koute (approved):

alexggh (approved):

s0me0ne-unkn0wn (approved):

PR #2899, opened 2024-01-10
Improve storage monitor API

nazar-pc:
This removes the need to unnecessarily provide a very specific data structure `DatabaseSource` and removes huge `sc-client-db` dependency from storage monitor. It is now possible to use storage monitor with any path.

P.S. I still strongly dislike that it pulls `clap` dependency for such a small feature, but many other crates do as well, so nothing special here.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4879577

bkchr (approved):

bkchr:
@nazar-pc please write a prdoc.

substrate/client/storage-monitor/src/lib.rs
```diff
@@ -70,10 +69,10 @@ impl StorageMonitorService {
 	/// Creates new StorageMonitorService for given client config
 	pub fn try_spawn(
 		parameters: StorageMonitorParams,
-		database: DatabaseSource,
+		path: Option<&Path>,
```
michalkucharczyk, line 72: `Option` is not needed here, could just be `Path`.

nazar-pc: Yes, but then if/else needs to be checked at call site + there is logging that would need to be copied as well. I looked at that and decided to keep it an option.

michalkucharczyk: Hm.. `None` may not always mean/require logging, it could be handled in other way. And IMO handling `None` seems like part of caller responsibility.

michalkucharczyk (approved):

substrate/client/storage-monitor/src/lib.rs
nazar-pc: Sure, done in last commit

PR #2905, opened 2024-01-10
Fix coretime migration

antonva:
The XCM instruction limit of 100 was being hit by the 1 large message before. The weight ref time also needed to be bumped for these extrinsics.

### DIFF
--- a/polkadot/runtime/parachains/src/coretime/migration.rs
+++ b/polkadot/runtime/parachains/src/coretime/migration.rs
@@ -142,8 +142,8 @@ mod v_coretime {
 			let new_core_count = assigner_coretime::Pallet::<T>::session_core_count();
 			ensure!(new_core_count == prev_core_count, "Total number of cores need to not change.");
 			ensure!(
-				dmp_queue_size == prev_dmp_queue_size + 4,
-				"There should have been four enqueued DMP messages."
+				dmp_queue_size > prev_dmp_queue_size,
+				"There should have been enqueued at least one DMP messages."
 			);
 
 			Ok(())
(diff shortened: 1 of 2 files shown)


polkadot/runtime/parachains/src/coretime/migration.rs
```diff
@@ -142,8 +142,8 @@ mod v_coretime {
 			let new_core_count = assigner_coretime::Pallet::<T>::session_core_count();
 			ensure!(new_core_count == prev_core_count, "Total number of cores need to not change.");
 			ensure!(
-				dmp_queue_size == prev_dmp_queue_size + 1,
-				"There should have been enqueued one DMP message."
+				dmp_queue_size == prev_dmp_queue_size + 4,
+				"There should have been four enqueued DMP messages."
 			);
```
ggwpez, line 148: Will the migration execute again or be prevented by `!Self::already_migrated()`?  
In general it is a bit dangerous to amend an existing migration without bumping any version.

antonva: The migration will not execute again on Rococo, tested via `try-runtime`. The messages will be submitted manually on Rococo.

polkadot/runtime/parachains/src/coretime/migration.rs
```diff
@@ -142,8 +142,8 @@ mod v_coretime {
 			let new_core_count = assigner_coretime::Pallet::<T>::session_core_count();
 			ensure!(new_core_count == prev_core_count, "Total number of cores need to not change.");
 			ensure!(
-				dmp_queue_size == prev_dmp_queue_size + 1,
-				"There should have been enqueued one DMP message."
+				dmp_queue_size == prev_dmp_queue_size + 4,
+				"There should have been four enqueued DMP messages."
```
bkchr, line 146:
```suggestion
				dmp_queue_size > prev_dmp_queue_size,
				"There should have been enqueued at least one DMP messages."
```

bkchr (approved):

eskimor (approved):

seadanda (approved):

PR #2908, opened 2024-01-10
backing: Back-off from backing on approval checking lag

alexggh:
If approval checking falls behind a certain threshold, it means that the network could not process the assignments and approvals needed for approving the block fast enough, so we need to back-off from creating new work, to give the opportunity of the approvals subsystems to catch up.

Continuously, creating new work is not a good idea because of the way approvals subsystems work, so if the system is slow on processing the assignments and approvals for the current block, either because we are behind on work from previous blocks or because the network is slow, validators will simply trigger new tranches which in turn causes more delays so we are going to create the conditions for the system to never catch up and fall behind.

Hence, why we need a mechanism to ensure that instead of falling more and more behind we actually allow the system to automatically catch up and start working in optimal conditions. This PR achieves that, by abstaining from backing new candidates if the node is behind on approvals beyond a certain threshold.

## TODO:
[ ] Is approval checking lag the right signal for deciding to back-off on backing ?
[ ] Test behaviour in real-world conditions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4123,7 +4123,7 @@ dependencies = [
  "polkadot-primitives",
  "polkadot-service",
  "polkadot-test-client",
- "prioritized-metered-channel",
+ "prioritized-metered-channel 0.5.1",
  "sc-cli",
  "sc-client-api",
  "sc-sysinfo",
@@ -8987,8 +8987,7 @@ checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
 [[package]]
 name = "orchestra"
 version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "46d78e1deb2a8d54fc1f063a544130db4da31dfe4d5d3b493186424910222a76"
+source = "git+https://github.com/paritytech/orchestra?branch=alexggh/expose_meters#0bcf7b7a05da7b5770824fee83794b5443f01e19"
 dependencies = [
  "async-trait",
  "dyn-clonable",
@@ -8996,7 +8995,7 @@ dependencies = [
  "futures-timer",
  "orchestra-proc-macro",
  "pin-project",
- "prioritized-metered-channel",
+ "prioritized-metered-channel 0.5.0",
  "thiserror",
  "tracing",
 ]
@@ -9004,8 +9003,7 @@ dependencies = [
 [[package]]
 name = "orchestra-proc-macro"
 version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d035b1f968d91a826f2e34a9d6d02cb2af5aa7ca39ebd27922d850ab4b2dd2c6"
+source = "git+https://github.com/paritytech/orchestra?branch=alexggh/expose_meters#0bcf7b7a05da7b5770824fee83794b5443f01e19"
 dependencies = [
  "expander 2.0.0",
  "indexmap 2.0.0",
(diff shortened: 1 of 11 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4888126

Overkillus:
Undeniably there is a need for a back-off mechanism. Also if there will be a back-off mechanism it definitely should limit the the entry-point of more candidates which is backing. Those two aspects I think pretty much everyone agrees on.

As Sandreim points out:
> this approach only considers the view of a single node.

This is also a problem because it effectively makes it an opt-in back-off scheme. Malicious nodes can continue backing. This is an issue when considering cases where all honest nodes back-off due to a large approval-lag / big approval-queue but malicious nodes continue backing and they lengthen this backed-off state on purpose because it allows them to monopolize backing. 

Fundamentally this issue is very very similar to the discussions around the finality lag and block-authoring back-off. I have very similar fears. 

I'm definitely not saying that finality proofs are the only way, but they are the nicest and cleanest way. That allows for easy audits and hopefully less mistakes on weird edge-cases as well as no concerns around balancing opt-in systems which can be abused.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4938609

alexggh:
# Try 2

## Changes:
- Updated the implementation with @sandreim suggestion where the backing subsystem backs-off from backing the candidate they are assigned to if approval-distribution informed them they are overloaded.
- Modified the signal we use to decide if we should back-off, instead of using the ApprovalCheckingLag, make ApprovalDistribution notify backing system when they have a high-backlog of unprocessed messages, I think this a better signal than ApprovalCheckingLag, because ApprovalCheckingLag in the presence of no-shows, but if we are able to process the messages we receive there is no reason to worry and back-off from backing.

## Addressing concerns:

### Use on-chain finality proofs.
Agree that having finality proofs on-chain would allow us to implement this in the cleanest and safest way, where we can prove that all nodes respected the behaviours, however that is a longer-term solution that would take some significant efort and resources to implement, in the meanwhile I do think that having this short-term solution would give us enough benefits and they won't make the situation worse.

### What if malicious nodes don't comply ?

#### Would the mechanism achieve its purpose ?
 Our network assumes 2/3 honest validators, so I would say that should be enough to allow nodes to catch up on work, we have bigger problems if we can't handle approving work backed by only 1/3 of the nodes.
 
 #### Can malicious nodes control/lenghten this state somehow ?
`approval-distribution` has spam mechanisms which prevents nodes to be overwhelmed with bogus messages, so this state would be trigger only if approval-distribution gets overwhelmed with valid expected messages, but since 2/3 of validators would back-off from backing their allocated parachains we should actually catch-up.

#### What will uncompliant nodes gain from this ?
Not much, since this just prevents honest validators from Seconding, they would still be allow to issue Valid statements if a candidate has been Seconded(done by malicious nodes), so there is no change in the mechanism of checking the integrity of a candidate and also in the era points rewards, since both Seconded and Valid statements are rewarded the same.

### How does affect liveness of System Parachains ?
Since System Parachains are doing important work on behalf of the network, this PR proposes that we make an exception for  System Parachains and continue seconding them.

@Overkillus @sandreim @eskimor Let me know what you think.

bkchr:
@alexggh do we still want this?

alexggh:
> @alexggh do we still want this?

No, it is deprecated in favor of: https://github.com/paritytech/polkadot-sdk/issues/4632

PR #2917, opened 2024-01-11
Update `sp-runtime::DigestItem` documentation

vgeddes:
* Updated description about `DigestItem::Other` which is now utilized by the BridgeHub system parachain. So it is no longer unsupported or experimental.
* Added a warning about ABI compatibility - Development teams working on Polkadot & Kusama relay chains should be careful about adding & using new digest items.


gpestana (commented):
Although I think this is an important warning and information, I'm not sure if it should be under a primitives repo. This warning would be better added to the Polkadot and Kusama runtimes, where the digest items are defined, perhaps?

PR #2927, opened 2024-01-15
Fix contracts compilation with `CARGO_TARGET_DIR` set

nazar-pc:
In case `CARGO_TARGET_DIR` is set, build artifacts were in the wrong place and `build.rs` was failing. With `CARGO_TARGET_DIR=/home/nazar-pc/.cache/cargo/target`:
```
error: failed to run custom build command for `pallet-contracts-fixtures v1.0.0 (/web/subspace/polkadot-sdk/substrate/frame/contracts/fixtures)`

Caused by:
  process didn't exit successfully: `/home/nazar-pc/.cache/cargo/target/debug/build/pallet-contracts-fixtures-35d534f7ac3009e0/build-script-build` (exit status: 1)
  --- stderr
  Error: Failed to read "/tmp/.tmpiYwXfv/target/wasm32-unknown-unknown/release/dummy.wasm"

  Caused by:
      Can't read from the file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The file was actually in `/home/nazar-pc/.cache/cargo/target/wasm32-unknown-unknown/release/dummy.wasm`.

### DIFF
--- a/substrate/frame/contracts/fixtures/build.rs
+++ b/substrate/frame/contracts/fixtures/build.rs
@@ -203,6 +203,7 @@ fn invoke_wasm_build(current_dir: &Path) -> Result<()> {
 
 	let build_res = Command::new(env::var("CARGO")?)
 		.current_dir(current_dir)
+		.env("CARGO_TARGET_DIR", current_dir.join("target").display().to_string())
 		.env("CARGO_ENCODED_RUSTFLAGS", encoded_rustflags)
 		.args(["build", "--release", "--target=wasm32-unknown-unknown"])
 		.output()


nazar-pc:
I believe it is not necessary there due to [`env_clear`](https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.env_clear) call there

athei:
You are right. We should probably do the same for the wasm build then. It seems more robust to remove all variables and only forward/set what we need. But let's wait for @pgherveou opinion.

pgherveou:
> You are right. We should probably do the same for the wasm build then. It seems more robust to remove all variables and only forward/set what we need. But let's wait for @pgherveou opinion.

I think I tried that and it was failing for some reason. I did not investigated much more since it was working without clearing.
let met check this quickly again

pgherveou:
> > You are right. We should probably do the same for the wasm build then. It seems more robust to remove all variables and only forward/set what we need. But let's wait for @pgherveou opinion.
> 
> I think I tried that and it was failing for some reason. I did not investigated much more since it was working without clearing. let met check this quickly again

seems to work ¯\_(ツ)_/¯, let me test in docker, push an update and see if that pass CI

substrate/frame/contracts/fixtures/build.rs
```diff
@@ -203,6 +203,7 @@ fn invoke_wasm_build(current_dir: &Path) -> Result<()> {
 
 	let build_res = Command::new(env::var("CARGO")?)
 		.current_dir(current_dir)
+		.env("CARGO_TARGET_DIR", current_dir.join("target").display().to_string())
```
pgherveou, line 206: shouldln't we just clear it if it set, instead of setting it ? 
or the fixtures will end up here instead of OUT_DIR?

nazar-pc: I haven't thought of it right away, but clearing should work as well

athei:
> > > You are right. We should probably do the same for the wasm build then. It seems more robust to remove all variables and only forward/set what we need. But let's wait for @pgherveou opinion.
> > 
> > 
> > I think I tried that and it was failing for some reason. I did not investigated much more since it was working without clearing. let met check this quickly again
> 
> seems to work ¯_(ツ)_/¯, let me test in docker, push an update and see if that pass CI
> 
> nvm, does not work in CI, it seems to pick up the wrong clang binary

This is weird. We don't depend on C code in our test fixtures?

substrate/frame/contracts/fixtures/build.rs
athei: It has to end up wherever you expect to find them in the rest of your script. In this case `current_dir/target` is the same as the default. Otherwise the CI would fail. So deleting should have the same effect. But let's find out why clearing the environment doesn't work.

athei:
Okay yes we pull in a lot of stuff for macro support. Try to forward `CC` and `CXX` to the command.

bkchr:
BTW, this is all already solved [here](https://github.com/paritytech/polkadot-sdk/blob/127b9bec15f8a83964ef6049209df75da47b1479/substrate/utils/wasm-builder/src/wasm_project.rs#L729-L741). Not sure what is that complicated to just copy this..

substrate/frame/contracts/fixtures/build.rs
pgherveou: let's stick with the current fix

if we wish to clear env, This works as well

```rust
﻿		.envs([
			("PATH", env::var("PATH").unwrap_or_default()),
			("CARGO_HOME", env::var("CARGO_HOME").unwrap_or_default()),
			("RUSTC", env::var("RUSTC").unwrap_or_default()),
		])
```

pgherveou (approved):

athei (approved):

bkchr (approved):

PR #2929, opened 2024-01-15
feat: add ApplyAuthorizedUpgradeOrigin

mrshiposha:
This PR introduces an origin check to `apply_authorized_upgrade` (and to `enact_authorized_upgrade` in the ParachainSystem).
By default, this check is set to `EnsureAlways` to ensure backward compatibility.

The check is helpful for better control over upgrade timings. For instance, the technical committee could initiate the upgrade authorized by the community. The committee could do upgrades in the predetermined timeframe.

Also, some upgrades might require additional maintenance right after their enactment. In this scenario, a "maintenance mode" could be enabled to turn off some chain functionality. To minimize the time of the "maintenance", it should be enabled right before the upgrade's enactment. It is only possible if the time of the enactment is predetermined.

### DIFF
--- a/bridges/snowbridge/parachain/pallets/ethereum-beacon-client/src/mock.rs
+++ b/bridges/snowbridge/parachain/pallets/ethereum-beacon-client/src/mock.rs
@@ -205,6 +205,7 @@ pub mod mainnet {
 	impl frame_system::Config for Test {
 		type BaseCallFilter = frame_support::traits::Everything;
 		type OnSetCode = ();
+		type ApplyAuthorizedUpgradeOrigin = frame_system::EnsureAlways;
 		type BlockWeights = ();
 		type BlockLength = ();
 		type DbWeight = ();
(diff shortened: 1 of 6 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4913306

joepetrowski:
I am not really sure about this PR. An upgrade authorization comes from root, and by adding a new origin to the runtime application, you are trusting that origin to do what root authorized to do. The scenario you give is reasonable but also benevolent, the technical committee could also refuse to apply some upgrade (e.g., one that removes the committee).

Substrate has lots of other tools that facilitate scheduling things like a maintenance mode. With both the Democracy and Referenda pallets you can set enactment block numbers or just use the Scheduler pallet to coordinate. For example, you could batch setting maintenance mode with the upgrade authorization, or schedule the authorization to happen at a certain time and co-schedule maintenance mode with it.

bkchr:
Yeah, I also don't see any reason. If you want to have some maintenance mode, you should look into `pallet-txpause` and `pallet-safe-mode`. You should enable them by the runtime upgrade automatically to have your maintenance mode.

PR #2935, opened 2024-01-15
Introduce Notification block pinning limit

### LINKED ISSUE #19: sync: "Unable to pin block for finality notification"
I'm running with the command `cargo run --release -- --sync fast-unsafe`* and seeing a bunch of these errors:

```
Unable to pin block for finality notification. hash: 0x05bb…e8db, Error: UnknownBlock: State already discarded for `0x05bb4525e4d7cd1eb622573e5a8d433b7c8f3bf0919bcdf6eff9851f4e37e8db` 
```

*: Background: Was trying to get a synced node as fast as possible for testing purposes. Realized later that warp sync is faster.

skunert:
While investigating some pruning issues I found some room for improvement in the notification pin handling.

**Problem:** It was not possible to define an upper limit on notification pins. The block pinning cache has a limit, but only handles bodies and justifications.

After this PR, bookkeeping for notifications is managed in the pinning worker. A limit can be defined in the worker. If that limit is crossed, blocks that were pinned for that notification are unpinned, which now affects the state as well as bodies and justifications. The pinned blocks cache still has a limit, but should never be hit.

closes #19


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4922884

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4922918

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+		Self { unpin_message_rx, task_backend, pinned_blocks }
+	}
+
+	pub fn handle_announce_message(&mut self, hash: Block::Hash) {
+		if let Some(entry) = self.pinned_blocks.get_or_insert(hash, Default::default) {
+			*entry = *entry + 1;
+		}
+	}
+
+	pub fn handle_unpin_message(&mut self, hash: Block::Hash) {
```
bkchr, line 146:
```suggestion
	fn handle_announce_message(&mut self, hash: Block::Hash) {
		if let Some(entry) = self.pinned_blocks.get_or_insert(hash, Default::default) {
			*entry = *entry + 1;
		}
	}

	fn handle_unpin_message(&mut self, hash: Block::Hash) {
```

Don't need to be `pub`?

bkchr:
> **Problem:** It was not possible to define an upper limit on notification pins. The block pinning cache has a limit, but only handles bodies and justifications.

I'm confused. Why do we need two times bookkeeping to handle this? Can we not just fix it in the block pinning cache?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4927203

substrate/client/db/src/pinned_blocks_cache.rs
```diff
@@ -20,7 +20,7 @@ use schnellru::{Limiter, LruMap};
 use sp_runtime::{traits::Block as BlockT, Justifications};
 
 const LOG_TARGET: &str = "db::pin";
-const PINNING_CACHE_SIZE: usize = 1024;
+const PINNING_CACHE_SIZE: usize = 2048;
```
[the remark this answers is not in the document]
skunert: This should be larger than the limit set in the notification pinning worker, since this is the maximum number of pinned block bodies overall: Notifications + RPC API + other uses. The number of pins in the state-db is not limited at all, it trusts users entirely to unpin by themselves. Here it is left it as a last resort.
If you don't like the value specifically, we can reduce both the values here and in the worker.

skunert:
> > **Problem:** It was not possible to define an upper limit on notification pins. The block pinning cache has a limit, but only handles bodies and justifications.
> 
> 
> 
> I'm confused. Why do we need two times bookkeeping to handle this? Can we not just fix it in the block pinning cache? 

The pinning cache is for everything pinned by the backend. For example the RPC API will also pin things.

I want that we have a defined limit on notifications pinned. And for that we need the extra bookkeeping.

substrate/client/service/src/client/client.rs
```diff
@@ -416,25 +431,16 @@ where
...
-						backend.unpin_block(message);
-					} else {
-						log::debug!("Terminating unpin-worker, backend reference was dropped.");
-						return
-					}
-				}
-				log::debug!("Terminating unpin-worker, stream terminated.")
-			}
-			.boxed(),
+		let unpin_worker = crate::client::notification_pinning::NotificationPinningWorker::new(
```
davxy, line 439: Nit. maybe you can do the `Arc::downgrade(&backend)` directly in the `NotificationPinningWorker::new` constructor.
Super-Nit : can you import `NotificationPinningWorker` to the scope? :-D

substrate/client/service/src/client/client.rs
```diff
@@ -326,19 +325,35 @@ where
...
-				};
+				} else {
+					let _ = self
+						.unpin_worker_sender
+						.unbounded_send(UnpinWorkerMessage::AnnouncePin(notification.hash))
+						.map_err(|e| {
+							log::debug!(
+								"Unable to send AnnouncePin worker message for finality: {e}"
+							)
+						});
```
davxy, line 340: Why if we fail to `pin_block` we don't consider this an `error` anymore?
Also what if the `upbounded_send` fails? Shouldn't this be logged as an error? Is ok to ignore the error here?

substrate/client/service/src/client/client.rs
```diff
@@ -326,19 +325,35 @@ where
...
 						"Unable to pin block for import notification. hash: {}, Error: {}",
 						notification.hash, err
 					);
+				} else {
+					let _ = self
+						.unpin_worker_sender
+						.unbounded_send(UnpinWorkerMessage::AnnouncePin(notification.hash))
+						.map_err(|e| {
+							log::debug!("Unable to send AnnouncePin worker message for import: {e}")
+						});
```
davxy, line 356: What about offloading the `backend.pin_block()` call to the `UnpinWorker` when it receives the `AccouncePin` message (which at this point would become `UnpinWorkerMessage::Pin`? 

In this way we completely delegate pins management to the Worker (which is also in charge of calling `unpin`. Also removes a couple of lines of duplicated logic from here.

Of course, in this case pinning becomes a "post-import" async action. Is that bearable?

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+		} else {
+			log::debug!(target: LOG_TARGET, "Received unpin message for already unpinned block. hash = {hash:?}");
+		};
+	}
+
+	/// Start working on the received messages.
+	///
+	/// The worker maintains a map which keeps track of the pinned blocks and their reference count.
+	/// Depending upon the received message, it acts to pin/unpin the block.
+	pub async fn work(mut self) {
```
davxy, line 168:
```suggestion
	pub async fn run(mut self) {
```

Looks like this is a more widespread naming *convention*. But feel free to ignore

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+	fn on_cleared(&mut self) {}
+
+	fn on_grow(&mut self, _new_memory_usage: usize) -> bool {
+		true
+	}
+}
+
+/// Worker for the handling of notification pinning.
+///
+/// It receives messages from a receiver and pins/unping based on the incoming messages.
```
davxy, line 118:
```suggestion
/// It receives messages from a receiver and pins/unpins based on the incoming messages.
```

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+			*refcount = *refcount - 1;
+			if *refcount == 0 {
+				self.pinned_blocks.remove(&hash);
+			}
+			if let Some(backend) = self.task_backend.upgrade() {
+				log::debug!(target: LOG_TARGET, "Reducing pinning refcount for block hash = {hash:?}");
+				backend.unpin_block(hash);
+			} else {
+				log::debug!(target: LOG_TARGET, "Terminating unpin-worker, backend reference was dropped.");
+				return
```
davxy, line 157: In this case we're not terminating the worker. Maybe we want to return a `Result` here to terminate the `work/run` loop below?

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+			}
+		} else {
+			log::debug!(target: LOG_TARGET, "Received unpin message for already unpinned block. hash = {hash:?}");
+		};
```
davxy, line 161:
```suggestion
		}
```

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,341 @@
...
+				UnpinWorkerMessage::AnnouncePin(hash) => self.handle_announce_message(hash),
+				UnpinWorkerMessage::Unpin(hash) => self.handle_unpin_message(hash),
+			}
+		}
+		log::debug!(target: LOG_TARGET, "Terminating unpin-worker, stream terminated.")
+	}
+}
+
+#[cfg(test)]
+impl<Block: BlockT, Back: Backend<Block>> NotificationPinningWorker<Block, Back> {
```
davxy, line 180:
```suggestion
impl<Block: BlockT, Back: Backend<Block>> NotificationPinningWorker<Block, Back> {
```

This whole impl block can be moved into `mod tests` scope

substrate/client/service/src/client/notification_pinning.rs
skunert: Good catch!

substrate/client/service/src/client/client.rs
skunert: I think this would indeed be more elegant. 

> Of course, in this case pinning becomes a "post-import" async action. Is that bearable?

Exactly, that is the point of not doing it in the worker. I remember having some discussions about this and we want to make absolutely sure that the block is pinned before the notifications goes out to all listeners. So I would like to keep it this way.

substrate/client/service/src/client/client.rs
skunert: When doing fast-sync it happens that we receive finality notifications for blocks that don't have state. We had an issue abouit these error messages causing confusion, so its debug now.

Made the unbounded send an error  however, thats a good point :+1:

substrate/client/service/src/client/client.rs
davxy: Yeah. I also see a couple of potential annoyances:

1) Any issue if we pin a block in the backend but the `unbounded_send` fails for any reason? 
2) A potential race condition: what if the `UnpinHandle::drop()` (-> and thus the `Unpin` message) is called before the worker receives the `AnnouncePin` message?  In this case maybe the pin counter in the worker will be :
   - first tried to be decremented (potentially decrementing the counter of another instance)
   - then in the best case will be re-incremented. But there is the possibility that the increment message will be lost as well

So 2 may introduce some nasty bugs maybe.
What about making pin request to the worker a sync message? So we wait before proceed sending out the import notification.

skunert: Regarding 1:
I think it is not very realistic. It is an unbounded queue. Behind this channel is `async-channel`, and the docs state that sending can fail if the queue is full or the other side is closed. First one will not happen since its unbounded, closing could happen but means that the worker is dead.

Regarding 2:
Even if drop is called before the worker calls gets the `AnnouncePin` message, the drop message needs to reach the worker. I checked the `async-channel` docs again and sadly it does not seem to give any ordering guarantees. The channels from `tokio` however are mpsc and FIFO, so maybe I should switch to that. Would that address your concern?

> What about making pin request to the worker a sync message? So we wait before proceed sending out the import notification.

Hmm that would be an option. However I am more concerned that this would delay this rather critical codepath while holding the import lock. IMO that would be more concerning than the two edge cases above.

davxy: > The channels from tokio however are mpsc and FIFO,

Well, I think would be a fair assumption for mpsc channels to be FIFO. But as you said I can't find any explicit statement on the doc. I asked https://github.com/smol-rs/async-channel/issues/6

> Hmm that would be an option. However I am more concerned that this would delay this rather critical codepath while holding the import lock. IMO that would be more concerning than the two edge cases above.

Fair enough.

skunert: > Well, I think would be a fair assumption for mpsc channels to be FIFO.

I mean the async-channel ones are mpmc. My proposal was to switch to tokio channels because we only require a single consumer and they explicitly state it.

davxy: I see. But would be interesting to understand how hey are working when there is a single consumer.
(the unpin worker is the only consumer right?)

skunert: Yep, unpin worker ist the only one

substrate/client/service/src/client/notification_pinning.rs
```diff
@@ -0,0 +1,353 @@
...
+use sc_client_api::{Backend, UnpinWorkerMessage};
+
+use sc_utils::mpsc::TracingUnboundedReceiver;
+use schnellru::Limiter;
+use sp_runtime::traits::Block as BlockT;
+
+const LOG_TARGET: &str = "db::notification_pinning";
+const NOTIFICATION_PINNING_LIMIT: usize = 1024;
+
+/// A limiter which automatically unpins blockst hat leave the data structure.
```
andresilva, line 41:
```suggestion
/// A limiter which automatically unpins blocks that leave the data structure.
```

andresilva (approved):

PR #2937, opened 2024-01-15
Feature gate to minimize `polkadot-sdk-docs` build time

NukeManDan:
Typically in webdev and docsdev, instant rendering of HTML in "dev" or "watching" mode is possible. Sadly AFAIK rustdocs does not support this kind of workflow and requires code compilation simply to view HTML. Rendering HTML only should _not_ require building the universe. Without any ability prior to this PR to optionally build the crates referenced in the `polkadot-sdk-docs` crate, you would need to have a successful build of all deps. 

This PR introduces a feature gate to allow minimal HTML only development of the `polkadot-sdk-docs` crate independent of it's internally referenced crates, taking the startup time to open the rendered HTML for it from many minutes to seconds on my machine. 

Only the docs crate now is ~35 crate deps for mermaid and docify, and building all deps is ~1716 in total.

## Todo

It still takes SOOOOO long to build, even with `SKIP_WASM_BUILD=1` hangs on:
- librocksdb-sys
- wasm-opt-sys
- kitchensink-runtime(build)

- [ ] does cumulus.rs need the deny internal links lint?
- [ ] Check `polkadot-sdk-docs` build time plot to see what crates are stalling build most <https://doc.rust-lang.org/beta/cargo/reference/timings.html> and try and gate them too.

### DIFF
--- a/docs/sdk/src/meta_contributing.rs
+++ b/docs/sdk/src/meta_contributing.rs
@@ -134,10 +134,14 @@
 //!
 //! ## How to Develop Locally
 //!
-//! To view the docs specific [`crate`] locally for development, including the correct HTML headers
-//! injected, run:
+//! To expedite documentation development and view _only_ the docs specific [`crate`] locally for development, including the correct HTML headers injected, run:
 //!
 //! ```sh
+//! # (Fastest) Only build the documentation crate
+//! SKIP_WASM_BUILD=1 RUSTDOCFLAGS="--html-in-header $(pwd)/docs/sdk/headers/toc.html" cargo doc -p polkadot-sdk-docs --no-deps --open
+//! # (Full SDK) All documentation, no deps
+//! SKIP_WASM_BUILD=1 RUSTDOCFLAGS="--html-in-header $(pwd)/docs/sdk/headers/toc.html" cargo doc -p polkadot-sdk-docs --no-deps --open
+//! # (Full SDK) All documentation, with external deps
 //! SKIP_WASM_BUILD=1 RUSTDOCFLAGS="--html-in-header $(pwd)/docs/sdk/headers/toc.html" cargo doc -p polkadot-sdk-docs --no-deps --open
 //! ```
 //!
--- a/docs/sdk/src/meta_contributing.rs
+++ b/docs/sdk/src/meta_contributing.rs
@@ -134,13 +134,14 @@
 //!
 //! ## How to Develop Locally
 //!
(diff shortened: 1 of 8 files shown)


docs/sdk/src/polkadot_sdk/cumulus.rs
```diff
@@ -42,7 +42,6 @@
 //!
 //! [FRAME]: crate::polkadot_sdk::frame_runtime
 
-#![deny(rustdoc::broken_intra_doc_links)]
```
NukeManDan, line 45: Unclear if this lint must live here, if so it may need to be feature gated as it breaks the minimal docs build

kianenigma:
I have to inspect what is done here a bit further, but fwiw if you run `cargo doc -p polkadot-sdk-docs --no-deps` it is pretty fast (except for the first time) to recompile, and I doubt if it gets any faster than that as this will ignore all dependencies (again, except for the first run). 

With this workflow, I do most of my writing with `--no-deps`, and only when I want to check my linking I will remove the `--no-dep` to see if all links work.

In general, if this is only an improvement on the first build, assuming you don't have any `target` dir cached, I don't think it is worth it.

docs/sdk/src/meta_contributing.rs
```diff
@@ -134,14 +134,18 @@
...
-//! injected, run:
+//! To expedite documentation development and view _only_ the docs specific [`crate`] locally for
+//! development, including the correct HTML headers injected, run:
 //!
 //! ```sh
+//! # (Fastest) HTML ONLY - Only build the documentation crate, with MINIMAL code compilation.
+//! RUSTDOCFLAGS="--html-in-header $(pwd)/docs/sdk/headers/toc.html" cargo doc -p polkadot-sdk-docs --no-default-features -F min-docs --no-deps --open
+//!
+//! # SDK Docs only - Minimal build - Only build the documentation crate, including code compilation for examples.
 //! SKIP_WASM_BUILD=1 RUSTDOCFLAGS="--html-in-header $(pwd)/docs/sdk/headers/toc.html" cargo doc -p polkadot-sdk-docs --no-deps --open
```
NukeManDan, line 145: This is the command suggested to get the minimal docs build. I did try this again on present master, and what I was experiencing was what I think was very long linking times in the stage of building the full kitchensink runtime and node... I can't seem to reproduce now, so I agree this extra complexity isn't worth it to get fast HTML output to inspect while doing docs dev.

PR #2946, opened 2024-01-16
Rococo Identity Migration Part 2 + Bug Fix

joepetrowski:
Order:

- [x] Start People Chain
    - [RPC node](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frococo-people-rpc.polkadot.io#/explorer)
- [x] Upgrade Rococo Relay (`EnsureRoot` -> `EnsureSigned`) (v1,006,002)
    - Done [here](https://rococo.subscan.io/extrinsic/0xef07e0f9dbb2b9e829305f132e6ce45d291239286e409177e20895e6687daa6c)
- [x] Migrate all identities
    - Done, see extrinsics from [this account](https://rococo.subscan.io/account/5FyNYrBwndvBttTkGUqGGCRAXtBH4Mh8xELDaxaFywTsjDKb)
- [x] Upgrade Rococo People (remove call filter) (v1,006,002)
    - Authorized [here](https://rococo.subscan.io/extrinsic/0xedf6a80229bd411b7ed8d3a489a767b0f773bed5c49239987a294c293a35b98b)

With added:

- [x] Upgrade Rococo People to fix `poke_deposit` bug (v1,006,001)
    - Authorized [here](https://rococo.subscan.io/extrinsic/0xd1dc3cd6e8274bd0196f8d9f13ed09f6e9c76e6a40f9786a1629f4cb22cf948d)

Note:

It's also possible to remove the Identity Migrator pallet from both the Relay Chain and the parachain at this time. I will leave them in for now to preserve the test cases until we run them on Kusama/Polkadot. We will also want a follow up to remove all Identity-related state from the Relay Chain.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4933781

cumulus/parachains/runtimes/people/people-rococo/src/xcm_config.rs
```diff
@@ -169,13 +169,17 @@ impl Contains<RuntimeCall> for SafeCallFilter {
...
-						frame_system::Call::kill_prefix { .. },
-				) | RuntimeCall::ParachainSystem(..) |
+			RuntimeCall::PolkadotXcm(
+				pallet_xcm::Call::force_xcm_version { .. } |
+					pallet_xcm::Call::force_default_xcm_version { .. }
+			) | RuntimeCall::System(
+				frame_system::Call::set_heap_pages { .. } |
+					frame_system::Call::set_code { .. } |
+					frame_system::Call::set_code_without_checks { .. } |
+					frame_system::Call::authorize_upgrade { .. } |
```
seadanda, line 179: Maybe the ability to upgrade the runtime like this vs just setcode deserves a mention in the changelog/prdoc, WDYT?

seadanda (approved):

cumulus/parachains/runtimes/people/people-rococo/src/xcm_config.rs
joepetrowski: This is the normal way to upgrade a parachain though. It's just the move of this call from Parachain System (all of which is included) to System (which specifies certain calls). So it's just preserving functionality.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4936085

substrate/frame/identity/src/lib.rs
```diff
@@ -1377,34 +1377,49 @@ impl<T: Config> Pallet<T> {
...
-				let new_subs_deposit = Self::subs_deposit(subs_of.len() as u32);
-				Self::rejig_deposit(&target, *current_subs_deposit, new_subs_deposit)?;
-				*current_subs_deposit = new_subs_deposit;
-				Ok(new_subs_deposit)
-			},
-		)?;
+		let has_identity = IdentityOf::<T>::contains_key(&target);
+		// This should really always be true, but don't insert random storage on the off chance it's
+		// not.
+		let new_id_deposit = if has_identity {
```
georgepisaltu, line 1383: This is redundant code because if there is no identity, the function will fail on `let (reg, _) = identity_of.as_mut().ok_or(Error::<T>::NoIdentity)?;`. It works with `IdentityOf` but not with `SubsOf` because `IdentityOf` is `OptionQuery` but `SubsOf` is `ValueQuery`. Guarding the subs rejig with a `SubsOf::<T>::contains_key(&target)` is enough IMO.

joepetrowski: Ah yeah, you are right.

georgepisaltu (approved):

ggwpez (approved):

substrate/frame/identity/src/lib.rs
```diff
@@ -1395,16 +1395,21 @@ impl<T: Config> Pallet<T> {
...
+					let new_subs_deposit = Self::subs_deposit(subs_of.len() as u32);
+					Self::rejig_deposit(&target, *current_subs_deposit, new_subs_deposit)?;
+					*current_subs_deposit = new_subs_deposit;
+					Ok(new_subs_deposit)
+				},
+			)?
+		} else {
+			// If the item doesn't exist, there is no "old" deposit, and the new one is zero, so no
+			// need to call rejig, it'd just be zero -> zero.
+			Zero::zero()
```
gpestana, line 1411: Based on the identity migrator impl, calling `Call::poke_deposit` with a non existing target will never pay any fees -- is this the expected behaviour? I guess it should be fine since the call is gated by root.

gpestana (approved):

substrate/frame/identity/src/lib.rs
joepetrowski: Yeah it is gated by root, and the only use in the impl of `reap_identity` would return an error if the target did not exist, so it shouldn't end up calling into `poke_deposit` at all.

PR #2949, opened 2024-01-16
Enable async backing on all testnet system chains

mrcnski:
Built on top of https://github.com/paritytech/polkadot-sdk/pull/2826/ which was a trial run.

Guide: https://github.com/w3f/polkadot-wiki/blob/master/docs/maintain/maintain-guides-async-backing.md


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4943591

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4983877

mrcnski:
I've tested all these chains and I see 6s blocks everywhere! Ready for a final review now.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4989961

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4994407

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4994405

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4995218

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4995236

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4995237

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -151,7 +151,7 @@ const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
 const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
 /// How many parachain blocks are processed by the relay chain per parent. Limits the
 /// number of blocks authored per slot.
-const BLOCK_PROCESSING_VELOCITY: u32 = 1;
+const BLOCK_PROCESSING_VELOCITY: u32 = 2;
```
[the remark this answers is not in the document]
mrcnski: Ah, I was just trying different values when 6s block times weren't working, and i forgot about this part:

> When elastic scaling has been released velocities greater than 1 will be supported.

I'll change it back, but this seems not to have any detrimental effect, in local testing at least. Thanks!

prdoc/pr_2949-async-backing-on-all-testnet-system-chains.prdoc
```diff
@@ -0,0 +1,21 @@
+title: Enable async backing on all testnet system chains
+
+doc:
+  - audience: Runtime User
+    description: |
+      Async backing has been enabled on all testnet system chains: asset-hub-westend,
+      bridge-hub-westend, bridge-hub-rococo, collectives-westend, contracts-rococo,
+      coretime-westend, coretime-rococo, people-westend, and people-rococo. These now target 6s
+      block times.
```
seadanda, line 9: On coretime at least, this will change the configuration required to achieve the same sale `leadin_length` and `interlude_length` durations in real time (they should both double). I think this should be included in the changelog, as for running chains the configuration will need to be updated after the runtime upgrade but before the end of the current region.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5003932

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5049615

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5055474

cumulus/parachains/runtimes/people/people-westend/src/lib.rs
```diff
@@ -136,6 +136,28 @@ pub fn native_version() -> NativeVersion {
...
+const BLOCK_PROCESSING_VELOCITY: u32 = 1;
+
+/// This determines the average expected block time that we are targeting.
+/// Blocks will be produced at a minimum duration defined by `SLOT_DURATION`.
+/// `SLOT_DURATION` is picked up by `pallet_timestamp` which is in turn picked
+/// up by `pallet_aura` to implement `fn slot_duration()`.
+///
+/// Change this to adjust the block time.
+pub const MILLISECS_PER_BLOCK: u64 = 6000;
+pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Well, I'd assume that they are parameters adjustable per parachain, and in theory, they could diverge in the future :thinking: 

Not sure if that makes sense though.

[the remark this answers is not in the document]
ordian: Yes, but they were previously using the same constants imported from the same place. If there's a need to override a value for some parachain, it can simply not import it declare it's own constant.

rphmeier (approved):

cumulus/parachains/runtimes/people/people-westend/src/lib.rs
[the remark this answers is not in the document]
seadanda: I was thinking the same. It would fit in well in the new parachains-constants crate, Should be straightforward to abstract this out but could also be a follow up PR.

ordian (approved):

seadanda (approved):

cumulus/parachains/runtimes/people/people-westend/src/lib.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Fair enough! Moved them.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5096089

svyatonik (approved):

bridges/primitives/chain-bridge-hub-cumulus/src/lib.rs
```diff
@@ -57,6 +57,12 @@ const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(constants::WEIGHT_REF_TI
 	.saturating_div(2)
 	.set_proof_size(polkadot_primitives::MAX_POV_SIZE as u64);
 
+/// We allow for 2 seconds of compute with a 6 second average block.
+const MAXIMUM_BLOCK_WEIGHT_FOR_ASYNC_BACKING: Weight = Weight::from_parts(
+	constants::WEIGHT_REF_TIME_PER_SECOND.saturating_mul(2),
+	polkadot_primitives::MAX_POV_SIZE as u64,
+);
```
bkchr, line 64: Why do we still have this extra const?

bkchr: Forumlated differently, why do we need `BlockWeightsForAsyncBacking` and `BlockWeights`?

svyatonik: It is used by our relayer (lives in a separate repo still) and now we have: (1) testnet bridge hubs, which are using async backing, so we'll be using this constant for those chains (2) production bridge hubs (Kusama and Polkadot), where we still need to this without-async-backing version until those are migrated to async backing too

bkchr: Ahh okay, I forgot this again.

bkontur (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5096660

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5096659

acatangiu (approved):

prdoc/pr_2949-async-backing-on-all-testnet-system-chains.prdoc
seadanda: Sorry missed the notification - see here:
https://github.com/paritytech/polkadot-sdk/blob/452f81366870436634f72e2a187179e5a9ad9894/substrate/frame/broker/src/types.rs#L261

In the configuration these two types are the parachain block number, whereas the `advance_notice` is the relay chain block number.

When we set the config on Rococo it was set with 12s block times in mind for these two values. I wonder does it make more sense to just change these to be the relay block number as well

bkchr: Ahh okay!

bkchr: We should rewrite this to also use the relay chain block number.

bkchr:
BTW, when we merge this we can not upgrade any of the system chains until someone has written a migration from 12s to 6s block time. This will require code on the node side and the runtime side.

rphmeier:
> BTW, when we merge this we can not upgrade any of the system chains until someone has written a migration from 12s to 6s block time

Can you summarize what is missing?

bkchr:
- https://github.com/paritytech/polkadot-sdk/blob/05cdf41c8391725a0d80088d8a21b3e65de0eb2e/cumulus/client/consensus/aura/src/collators/lookahead.rs#L98-L99 these and similar `slot_duration` settings need to be dynamic to react to changes on chain.
- I actually thought that we need some migration for: https://github.com/paritytech/polkadot-sdk/blob/05cdf41c8391725a0d80088d8a21b3e65de0eb2e/substrate/frame/aura/src/lib.rs#L180 But I realized as we go down with the slot duration, the slots will jump to a much higher slot and thus the check that the slot increases will be correct.

rphmeier:
> these and similar slot_duration settings need to be dynamic to react to changes on chain.

If configured correctly, you don't necessarily need to have 6s slots for 6s blocks.

With an unincluded segment length of 3, you can allow up to 3 blocks per 12s slot and always have something ready.

bkchr:
The slot duration is used to calculate the slot. If the node and runtime side do not match, the blocks will be rejected.

rphmeier:
and has it changed in any of the chains? I just took another look and it seems to continue using the genesis slot duration of 12 seconds.

bkchr:
> and has it changed in any of the chains? I just took another look and it seems to continue using the genesis slot duration of seconds, right?

For me it looks like this changed: https://github.com/paritytech/polkadot-sdk/pull/2949/files#diff-95384194e1d261e31cd6d6326580597925d02e6877fa546c6379b8f31bf6a7c3R134
See the diff, the original import of `SLOT_DURATION` was changed everywhere.

> as I wrote above, you can still enable and benefit from asynchronous backing with a 12s slot time

Yeah I get what you said, but I also have to say that it makes stuff more "weird". But maybe it is just me that likes constant slots ;)

cumulus/pallets/collator-selection/src/benchmarking.rs
```diff
@@ -394,7 +394,7 @@ mod benchmarks {
 		register_validators::<T>(c);
 		register_candidates::<T>(c);
 
-		let new_block: BlockNumberFor<T> = 1800u32.into();
+		let new_block: BlockNumberFor<T> = 3600u32.into();
```
bkchr, line 397: Not your fault as the benchmark before was not correct. However, changing this will break the benchmark for anyone downstream that doesn't change the slot duration to 6 seconds. This needs to made generic.

bkchr:
If someone says that they have tested this and it works, I'm fine. I just don't that we break stuff again and then need to fix it again etc.

cumulus/pallets/collator-selection/src/benchmarking.rs
georgepisaltu: Working on a fix.

georgepisaltu: Fix [here](https://github.com/paritytech/polkadot-sdk/pull/3189).

s0me0ne-unkn0wn:
> See the diff, the original import of `SLOT_DURATION` was changed everywhere.

BTW, that's right, but TBH, I don't like the result. Having that `SLOT_DURATION` and everything in `parachain_common` is very error-prone; it's easy to screw things up and import `SLOT_DURATION` from one place and `HOUR` from another and get a working parachain that has the wrong perception of time. I'd say with the introduction of async backing, those constants couldn't be called "common" anymore and thus shouldn't be present in a package named `whatever_common`. But that can be addressed in a follow-up.

bkchr:
I mean this `parachain_common` crate is only for these test chains we have there. `SLOT_DURATION` is also not defined anywhere else. I don't think that this is a huge problem. Consolidating this stuff between the huge number of test runtimes we have makes sense IMO.

PR #2957, opened 2024-01-17
Update Wasm benchmarks

Robbepop:
In https://github.com/paritytech/polkadot-sdk/pull/2941 we found out that the new Wasmi (register) is very effective at optimizing away certain benchmark bytecode constructs in a way that created an unfair advantage over Wasmi (stack) which yielded our former benchmarks to be ineffective at properly measuring the performance impact.

This PR adjusts both affected benchmarks to fix the stated problems.
Affected are
- `instr_i64const` -> `instr_i64add`: Renamed since it now measures the performance impact of the Wasm `i64.add` instruction with locals as inputs and outputs. This makes it impossible for Wasmi (register) to aggressively optimize away the entire function body (as it previously did) but still provides a way for Wasmi (register) to shine with its register based execution model.
- `call_with_code_per_byte`: Now uses `local.get` instead of `i32.const` for the `if` condition which prevents Wasmi (register) to aggressively optimizing away whole parts of the `if` creating an unfair advantage.

cc @athei


Robbepop:

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -2582,19 +2582,24 @@ benchmarks! {
...
-	// the same amount of time. We call this weight `w_base`.
-	// The weight that would result from the respective benchmark we call: `w_bench`.
+	// We add two `i64` integers from local variables and store the result into yet another
+	// local variable. The combination of this computation is our weight base `w_base`.
 	//
-	// w_base = w_i{32,64}const = w_drop = w_bench / 2
+	// The weight that would result from the respective benchmark we call: `w_bench`.
 	#[pov_mode = Ignored]
-	instr_i64const {
+	instr_i64add {
```
pgherveou, line 2590: if we change things here, we also need to update
 
https://github.com/paritytech/polkadot-sdk/blob/6c3c1b6f26ea14daed1140110ff057776f8ec9c1/substrate/frame/contracts/src/schedule.rs#L398-L404

Robbepop: Ah thanks, I was already wondering about this!

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -2582,19 +2582,24 @@ benchmarks! {
...
 	}: call(origin, instance.addr, 0u32.into(), Weight::MAX, None, vec![])
 
-	// We make the assumption that pushing a constant and dropping a value takes roughly
-	// the same amount of time. We call this weight `w_base`.
-	// The weight that would result from the respective benchmark we call: `w_bench`.
+	// We add two `i64` integers from local variables and store the result into yet another
+	// local variable. The combination of this computation is our weight base `w_base`.
 	//
-	// w_base = w_i{32,64}const = w_drop = w_bench / 2
+	// The weight that would result from the respective benchmark we call: `w_bench`.
```
pgherveou, line 2588: We can just remove this comment now

substrate/frame/contracts/src/benchmarking/mod.rs
pgherveou: yes, let's see if that makes CI happy 🤞

Robbepop:
bot cancel

Robbepop:

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4941369

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4941335

Robbepop:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4941590

substrate/frame/contracts/src/schedule.rs
```diff
@@ -396,10 +396,12 @@ impl Default for Limits {
...
 impl<T: Config> Default for InstructionWeights<T> {
-	/// We price both `i64.const` and `drop` as `instr_i64const / 2`. The reason
-	/// for that is that we cannot benchmark either of them on its own.
+	/// We price all of `i64.add`, `local.get` and `local.set` as `instr_i64add / 2`. The reason
+	/// for that is that we cannot benchmark either of them on their own and division by 2 makes
+	/// for a decent compromise between the optimization capabilities of Wasmi (stack) and
+	/// Wasmi (register).
 	fn default() -> Self {
-		Self { base: cost_instr!(instr_i64const, 1), _phantom: PhantomData }
+		Self { base: cost_instr!(instr_i64add, 1), _phantom: PhantomData }
```
pgherveou, line 404: Since cost_instr macro is only used here, since we use wami gas metering
we should simplify it to 
```rust
macro_rules! cost_instr {
	() => {
		cost_instr_no_params!(instr_i64add)
			.saturating_sub((cost_instr_no_params!(instr_i64add) / 2))
	};
}
```
or just inlined it here. That would make the comment above easier to understand as well

pgherveou (approved):

substrate/frame/contracts/src/schedule.rs
Robbepop: I agree but the bench bot still reports a bug for usage of `instr_i64add` so I am reluctant to more changes until this is resolved. (i have no idea what is wrong atm)

Robbepop: Or is this bug to be expected since we changed the set of benchmark test cases between `master` and this PR?

pgherveou: > This is weird to me since we change an automatically generated file but code review recommended to change this to make the CI happy so I do.

It's only updated when you run the benchmark or ask the bot to do it with `bot bench substrate-pallet --pallet=pallet_contracts`. But I believe the code needs to compile first, I usually add a default impl in the WeightInfo trait to make it work

Robbepop:

athei (commented):
Looks good. We just need to resolve the situation in the schedule and decide which factor we want to apply to the benchmark result.

Robbepop:
> Looks good. We just need to resolve the situation in the schedule and decide which factor we want to apply to the benchmark result.

Always happy to remove code.
I was suggesting the factor 2 and still am as a compromise. We might change it later anyways.

athei:
The code on master has to be correct for every commit. Since we are merging this into master without the register machine the factor should not be a compromise but reflect the performance of the stack machine we are still using when this PR will be merged. So please use the stack machine factor in #2957. You can update it in #2941.

Robbepop:
> The code on master has to be correct for every commit. Since we are merging this into master without the register machine the factor should not be a compromise but reflect the performance of the stack machine we are still using when this PR will be merged. So please use the stack machine factor in this PR. You can update it in #2941.

done

Robbepop:

substrate/frame/contracts/src/schedule.rs
```diff
@@ -396,10 +383,13 @@ impl Default for Limits {
...
 
 impl<T: Config> Default for InstructionWeights<T> {
-	/// We price both `i64.const` and `drop` as `instr_i64const / 2`. The reason
-	/// for that is that we cannot benchmark either of them on its own.
+	/// We execute 4 different instructions to emulate the `i64.add` therefore
+	/// we have to divide the actual computed gas costs by 4 to have a rough
+	/// estimate as to how expensive each single executed instruction is going to be.
 	fn default() -> Self {
-		Self { base: cost_instr!(instr_i64const, 1), _phantom: PhantomData }
+		let instr_cost = cost_args!(instr_i64add, 1).ref_time() as u32;
```
athei, line 390: This should be the same.

```suggestion
		let instr_cost = cost!(instr_i64add).ref_time() as u32;
```

Robbepop:
This is ready to merge! Awaiting approvals.

athei:
> call_with_code_per_byte: Now uses local.get instead of i32.const for the if condition which prevents Wasmi (register) to aggressively optimizing away whole parts of the if creating an unfair advantage.

Your change will also (correctly) affect the other two benchmarks which are proportional to the contract size `instantiate_with_code`and `upload_code`. Just wanted to point this out. Hmm I hope we do not double charge in this case. But I don't think so.

By the way: You don't need to re-benchmark after just changing the `Schedule`. The benchmarks results are an **input** to the `Schedule` derivation. But of course best practice to run after every code change just be safe.

> This is ready to merge! Awaiting approvals.

Have you used `subweight` to compare results? Is everything as expected?

Robbepop:
@athei and me resolved some questions in chat.

The benchmarks have some weird outliers which we need to analyze. It does not really make sense that the changes introduced in this PR are the root cause for them which is why I just started https://github.com/paritytech/polkadot-sdk/pull/2969 in order to run benchmarks with no changes and see about their results to confirm.

More info: the new `instr_i64add` benchmark seems to have improved by 20% which is counterintuitive since we actually doubled the amount of instruction that Wasmi (stack) has to execute.

Robbepop:
I have an explanation for the performance difference between the old `instr_i64const` and the new `instr_i64add`.
Recap: Since `instr_i64add` has roughly double the amount of instructions than `instr_i64const` we assumed that it would also run twice as slow, however, benchmarks showed that it in fact runs roughly 20% faster. Why is that?

An explanation is the following: Wasmi (stack) has optimized `i64.const` instructions for `i64` (and `i32`) values in range `i32::MIN..i32::MAX`. For all other values of `i64` it needs to load the constant value from a global constant pool which is slow because it isn't local and not cache friendly (more or less random access).
However, in the old `instr_i64const` benchmark all of the `i64.const` constant values were generated in `i64::MIN..i64::MAX` uniformely making it extremely unlikely (1/2^32) that the constant could be small enough for Wasmi (stack) to use the optimized instruction. Therefore the former `instr_i64const` benchmark tested the absolute and kinda unrealistic worst case possible for `i64.const` instructions all the time whereas the new `instr_i64add` has a more realistic approach and Wasmi (stack) can execute all the instructions in a cache friendly manner making the results more realistic.

substrate/frame/contracts/src/schedule.rs
```diff
@@ -396,10 +383,13 @@ impl Default for Limits {
...
-	/// We price both `i64.const` and `drop` as `instr_i64const / 2`. The reason
-	/// for that is that we cannot benchmark either of them on its own.
+	/// We execute 4 different instructions to emulate the `i64.add` therefore
+	/// we have to divide the actual computed gas costs by 4 to have a rough
+	/// estimate as to how expensive each single executed instruction is going to be.
 	fn default() -> Self {
-		Self { base: cost_instr!(instr_i64const, 1), _phantom: PhantomData }
+		let instr_cost = cost!(instr_i64add).ref_time() as u32;
+		let base = instr_cost.saturating_sub(instr_cost / 4);
+		Self { base, _phantom: PhantomData }
```
athei, line 392: This code does not device the cost by `4`. It does multiple it with `0.75`. So either the comment or the code is wrong. (Or I am wrong)

athei (changes requested):
Given your findings we need to keep continuing using big numbers to exercise this worst case. While it is "unrealistic" in practise that big numbers are used it is very easy for an attacker to fill a contract with these instructions. If we to profit from this optimization it has to be accounted for in the wasmi gas model. Not by optimistically benchmarking for that optimization. Benchmarks have to be for the worst case.

While the benchmark itself only improves by 20% you have to keep in mind that we are dividing that by 4 to account for the increase of instructions. So you actually have to compare what is being put into the schedule and not only the raw benchmarks results.

Before we divided by 2. So with this PR we are actually using 40% less gas per instruction.

Robbepop:
In defense of the new benchmark using proper `i64.add` instructions as base weight I have to disagree.
The issue is that the previous benchmark was accidentally using a worst-case for the Wasmi (stack), however, under Wasmi (register) this might totally behave differently since both Wasmi versions handle these big constants differently.
Therefore the old benchmark was only implicitly testing the worst case and only by accident since I am sure that the person who implemented this was not even aware of this Wasmi optimization for small constant values.

If we want to do a proper job with accounting for the worst case instruction we probably should at least be explicit about our intent and use `i64.load` or `i64.store` or ideally both with random access patterns and a huge linear memory area. As discussed in chat this should behave extremely similar to the previous `i64.const` with big values but is explicit about the worst-case random access and not implicit.

Though, there technically are even worse offenders with respect to performance but it always varies from engine to engine. For example Wasmi (stack) is very efficient at handling function calls whereas Wasmi (register) is much more efficient at handling computation.

At the very moment where we are about to change the underlying Wasm execution engine soon and furthermore where a thorough Polkadot SDK built-in gas model is in sight, I don't see a huge gain in trying to do a splendid job here right now.

If division by 4 is a problem then division by 2 (as planned by me originally) would be a good enough compromise.

Robbepop:
After some discussion in chat we agreed to use the proposed `i64.{load,store}` based benchmark since it is kinda similar to what was done previously by accident and kinda models some worst case behavior for a base instruction with random memory access patterns as described earlier.

@athei  I just implemented this agreed design test case.

Robbepop:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4975457

Robbepop:

Robbepop:
Applied your review suggestions.

Rendered benchmark results:
image

Though unfortunately even with `--threshold 0` the new `instr_i64_load_store` is missing.

athei (approved):

PR #2958, opened 2024-01-17
Pvf thiserror

### LINKED ISSUE #2157: PVF: refactor errors to use `thiserror`
We're already using [`thiserror`](https://github.com/dtolnay/thiserror) in some places, so we should consider expanding its usage to reduce boilerplate. For example, we could add `#[from]` attributes for the variants of [`ValidationError`](https://github.com/paritytech/polkadot-sdk/blob/f84b89710b/polkadot/node/core/pvf/src/error.rs#L21), making it easier to construct. For sure another candidate is [`PrepareError`](https://github.com/paritytech/polkadot-sdk/blob/f84b89710b/polkadot/node/core/pvf/common/src/error.rs#L27), we could provide an auto conversion for e.g. `IoErr`, reducing the boilerplate whenever it's constructed. Then instead of a bunch of `map_err`s on `std::io::Error`, we can just use `?`.

See [here](https://github.com/paritytech/polkadot-sdk/blob/f84b89710b/polkadot/node/core/pvf/common/src/worker/security/seccomp.rs#L90-L98) for an example thiserror usage. cc @eagr since you were interested in some refactorings. :)

maksimryndin:
resolve #2157 

- [x] fix broken doc links
- [x] fix codec macro typo https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/core/pvf/common/src/error.rs#L81 (see the comment below)
- [x] refactor `ValidationError`, `PrepareError` and related error types to use `thiserror` crate

## `codec` issue

`codec` macro was mistakenly applied two times to `Kernel` error (so it was encoded with 10 instead of 11 and the same as `JobDied`). The PR changes it to 11 because

- it was an initial goal of the code author
- Kernel is less frequent than JobDied so in case of existing error encoding it is more probable to have 10 as JobDied than Kernel

See https://github.com/paritytech/parity-scale-codec/issues/555

----
polkadot address: 13zCyRG2a1W2ih5SioL8byqmQ6mc8vkgFwQgVzJSdRUUmp46


mrcnski:
> codec macro was mistakenly applied two times to Kernel error (so it was encoded with 10 instead of 11 and the same as JobDied). I changed it to 11

Thank you! You're absolutely right, and sorry if it caused any confusion. It is already being fixed in https://github.com/paritytech/polkadot-sdk/pull/2477, and I've opened an issue about it on [parity-scale-codec](https://github.com/paritytech/parity-scale-codec/issues/555). Appreciate you fixing it here as well!

polkadot/node/core/pvf/common/src/error.rs
```diff
@@ -32,35 +31,43 @@ pub type PrecheckResult = Result<(), PrepareError>;
...
+	#[error("preparation: {0}")]
 	Preparation(String),
 	/// Instantiation of the WASM module instance failed.
 	#[codec(index = 2)]
+	#[error("runtime construction: {0}")]
 	RuntimeConstruction(String),
 	/// An unexpected error has occurred in the preparation job.
 	#[codec(index = 3)]
+	#[error("panic: {0}")]
 	JobError(String),
```
mrcnski, line 51:
```suggestion
	#[error("prepare job error: {0}")]
	JobError(String),
```

(That's my bad, I forgot to update this line after renaming `Panic` -> `JobError`.

polkadot/node/core/pvf/src/error.rs
```diff
@@ -27,35 +27,42 @@ pub enum ValidationError {
 	/// pre-checking enabled only valid runtimes should ever get enacted, so we can be
 	/// reasonably sure that this is some local problem on the current node. However, as this
 	/// particular error *seems* to indicate a deterministic error, we raise a warning.
+	#[error(transparent)]
 	Preparation(PrepareError),
 	/// The error was raised because the candidate is invalid. Should vote against.
-	Invalid(InvalidCandidate),
+	#[error(transparent)]
```
mrcnski, line 33: Can we have specific error messages here instead of the fall-through `transparent`?

mrcnski (approved):
Looks great, thank you! If you want a tip, you can include your kusama or polkadot address in the PR body:

https://github.com/paritytech/substrate-tip-bot#pull-request-body

I'm interested in whether we can now better log some of these errors that didn't implement the `Display` trait before. We should be able to log them with `{}` instead of `{:?}`. I'll look through the codebase later for such instances.

polkadot/node/core/pvf/common/src/error.rs
maksimryndin: For sure:) I've noticed it but from [the book description](https://paritytech.github.io/polkadot-sdk/book/node/utility/pvf-host-and-workers.html) it was quite logical

polkadot/node/core/pvf/src/error.rs
maksimryndin: Let me explain the motivation:)

1. `InternalValidationError` already has `validation: ` prefix in its messages
2. `ValidationError` has 1-1 match between enum variants and wrapped sub errors
3. As a novice to the codebase, I am not sure how much freedom I have over the error messages

So I decided to stick with the safest solution in my opinion. But of course, if you think we should expand messages on wrapper, I will do :)

polkadot/node/core/pvf/common/src/error.rs
maksimryndin: For consistency we could also make a `prepare: ` prefix for other messages in `PrepareError` if it is allowed

maksimryndin: > I'm interested in whether we can now better log some of these errors that didn't implement the Display trait before. We should be able to log them with {} instead of {:?}. I'll look through the codebase later for such instances.

It is good point! Thank you! I will check these instances within the PR

s0me0ne-unkn0wn (approved):
A good cleanup! Makes things more clear.

If you ever want to continue working on cleaning up the error processing, we have a long-waiting #752 ;) (although I believe it should be updated and discussed further).

polkadot/node/core/pvf/common/src/error.rs
maksimryndin: @mrcnski I've checked the usage of `{:?}` within `pvf/` and changed several ones to `{}`. And also changed `PrepareError` messages a little bit for consistency.

maksimryndin:
> A good cleanup! Makes things more clear.
> 
> If you ever want to continue working on cleaning up the error processing, we have a long-waiting #752 ;) (although I believe it should be updated and discussed further).

Thank you!:) Perhaps, I will also try that but later. The coding part is not a problem here but my lack of domain knowledge and lack of usage experience which would benefit to the mentioned task.

polkadot/node/core/pvf/src/error.rs
mrcnski: Good points! If you want though, you can change this:

- For preparation, add a message like `preparation error for validation: {0}`
- For internal error, maybe change the message to `internal validation error: {0}`

> As a novice to the codebase, I am not sure how much freedom I have over the error messages

Appreciate the humility and respect for the codebase. :)

mrcnski (approved):
Looks awesome, just one optional comment and then we can merge!

polkadot/node/core/pvf/src/error.rs
mrcnski: Also not sure if the change is necessary TBH so I'll leave it to your judgment. I think it's fine to err on the side of being a bit too verbose: error messages are rare and the more info they contain, the better.

maksimryndin: Agree! I've [decided to use](https://github.com/paritytech/polkadot-sdk/pull/2958/commits/471a768493f2b3629bffb7ff59cc63cc963bf317) `candidate validation:` prefix for `ValidationError`. Thus it looks consistent

PR #2962, opened 2024-01-17
Allow to set a worst case buy execution fee asset and weight

girazoki:
Adds `worst_case_buy_execution` to the Config trait of `pallet-xcm-benchmarks` with a default implementation that mimics the code that existed previous to this PR.

Rationale: not allowing to set the `WeightLimit` and the `FeeAsset` might mean that we dont benchmark the worst case, as with `WeightLimit::Unlimited` the `Trader` does not even execute: https://github.com/paritytech/polkadot-sdk/blob/c01dbebeaa6394691974de46dd2d41a582f6a4c2/polkadot/xcm/xcm-executor/src/lib.rs#L833

The new configurable function allows projects to customize the parameters with which the benchmark is run to make sure they account for the worst-case scenario


polkadot/xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -77,11 +77,13 @@ benchmarks! {
 		let mut executor = new_executor::<T>(Default::default());
 		executor.set_holding(holding);
 
-		let fee_asset = AssetId(Here.into());
+		// The worst case we want for buy execution in terms of
+		// fee asset and weight
+		let (fee_asset, weight_limit) = T::worst_case_buy_execution()?;
```
franciscoaguirre, line 82: I definitely like the idea of benchmarking with `WeightLimit::Limited` so we can hit more logic. There's now a `fee_asset` configuration item that is used to buying execution in the benchmark for `refund_surplus`, this could be merged with that to include the weight_limit.

We also have to make sure this is in holding, don't know if there's a good way to do it.

polkadot/xcm/pallet-xcm-benchmarks/src/generic/mod.rs
```diff
@@ -100,6 +106,13 @@ pub mod pallet {
 				crate_version: <frame_system::Pallet<Self> as frame_support::traits::PalletInfoAccess>::crate_version(),
 			}
 		}
+
+		/// The worst case buy execution weight limit and
+		/// asset to trigger the Trader::buy_execution in the XCM executor
+		/// By default returns ((AssetId(Here.into()), 100_000_000u128), Unlimited)
+		fn worst_case_buy_execution() -> Result<(Asset, WeightLimit), BenchmarkError> {
+			Ok(((AssetId(Junctions::Here.into()), 100_000_000u128).into(), WeightLimit::Unlimited))
```
franciscoaguirre, line 114: I don't think we need a default for this. More often than not the worst case is `Limited` and we have many examples that don't buy fees with `Here`

polkadot/xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
girazoki: ok I can try to do this!

girazoki: For holding, yes we rely on `worst_case_holding` having the fees, and I guess that if it is not the benchmark will fail. So I think this is good enough.

The other way we could do it is by injecting the fee_asset into holding directly, without using `worst_case_holding` at all

girazoki: @franciscoaguirre I dont see the `fee_asset` setup in `refund_surplus`. all I see is that the `total_surplus` is artificially set https://github.com/paritytech/polkadot-sdk/blob/c01dbebeaa6394691974de46dd2d41a582f6a4c2/polkadot/xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs#L131

polkadot/xcm/pallet-xcm-benchmarks/src/generic/mod.rs
girazoki: What should I do in this PR for all the polkadot runtimes? Use the same values they had before? Because I dont think I know wll the polkadot/cumulus  runtimes as well to know what is the worst usecase for each of them..

girazoki:
@franciscoaguirre do you have any more input on this? If I dont provide the default implementation I dont think I know enough of the runtimes myself to implement the worst case scenario in each of the cumulus/relay runtimes

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5265397

RomarQ:
@girazoki is this still applicable?

acatangiu:
Closing as stale, please reopen if needed.

girazoki:
I think it is still aplicable, I willl open a new PR.

girazoki:
Sorry that I forgot about this and it became stale

Issue #2960, opened 2024-01-17
error when compiling Substrate node-template


ausipome:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

error[E0449]: visibility qualifiers are not permitted here
    --> substrate/frame/contracts/src/lib.rs:1247:3
     |
1247 |         environmental!(executing_contract: bool);
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: trait items always share the visibility of their trait
     = note: this error originates in the macro `environmental` (in Nightly builds, run with -Z macro-backtrace for more

### Steps to reproduce

This happens when compiling the node-template in the Substrate folder on WSL. I have tried to compile using the latest nightly tool chain and also  rustc 1.75.0-nightly as the latest release has been tested against this.

PR #2986, opened 2024-01-18
pallet-contracts: Fix compilation on latest nightly

bkchr:
Apparently they changed detection for visibility identifiers on traits, which broke more than it should. There is an issue open: https://github.com/rust-lang/rust/issues/119924 The easy solution for us is to move the declaration of the global variable outside of the trait.

Closes: https://github.com/paritytech/polkadot-sdk/issues/2960

### DIFF
--- a/substrate/frame/contracts/src/lib.rs
+++ b/substrate/frame/contracts/src/lib.rs
@@ -1228,6 +1228,9 @@ struct InternalOutput<T: Config, O> {
 	result: Result<O, ExecError>,
 }
 
+// Set up a global reference to the boolean flag used for the re-entrancy guard.
+environmental!(executing_contract: bool);
+
 /// Helper trait to wrap contract execution entry points into a single function
 /// [`Invokable::run_guarded`].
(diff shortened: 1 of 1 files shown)


pgherveou (approved):

ggwpez (approved):
So its only important inside traits?  
Since otherwise we should also change it [here](https://github.com/paritytech/polkadot-sdk/blob/dcc76525d98a097be152293eee3389f6af5c4404/substrate/frame/message-queue/src/lib.rs#L1459).

athei (approved):

bkchr:
> So its only important inside traits? Since otherwise we should also change it [here](https://github.com/paritytech/polkadot-sdk/blob/dcc76525d98a097be152293eee3389f6af5c4404/substrate/frame/message-queue/src/lib.rs#L1459).

Yes only in traits AFAIK.

fmease:
As the relevant rustc dev speaking, sorry for the inconvenience, I will fix this as soon as possible when time permits.

Issue #2963, opened 2024-01-17
Impossible to create chain spec with large balance


nazar-pc:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

After https://github.com/paritytech/polkadot-sdk/pull/1256 `ChainSpec::from_genesis` is deprectated, but both serializing manually constructed `RuntimeGenesisConfig` explicitly and using `json!` macro fails for large balances like `u128:MAX / 2`:
```
Error("number out of range", line: 0, column: 0)
```

Same balance still works great with `ChainSpec::from_genesis`, but the way it actually does that is extremely convoluted, I was not yet able to figure out why it succeeds.

### Steps to reproduce

Try to set `u128::MAX / 2` balance for one of the accounts at genesis

bkchr:
@nazar-pc does this fixes your issue?

bkchr:
Yeah fine, @michalkucharczyk you want to open a pr to enable it somewhere in our stack? Maybe somewhere near the `GenesisConfig` stuff?

PR #2987, opened 2024-01-18
`GenesisBuilder`: `arbitrary_precision` feature enabled for `serde_json`

michalkucharczyk:
[`arbitrary_precision`](https://github.com/serde-rs/json/blob/6d44b9fac9269b4decf76acac5d68e8ec9d10c58/Cargo.toml#L69-L75) feature allows to (de-)serialize big numbers w/o error. 
For some details refer also to https://github.com/paritytech/polkadot-sdk/pull/1256#discussion_r1455564450

fixes: #2963

### DIFF
--- a/substrate/primitives/genesis-builder/Cargo.toml
+++ b/substrate/primitives/genesis-builder/Cargo.toml
@@ -19,7 +19,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 sp-api = { path = "../api", default-features = false }
 sp-runtime = { path = "../runtime", default-features = false }
 sp-std = { path = "../std", default-features = false }
-serde_json = { version = "1.0.111", default-features = false, features = ["alloc"] }
+serde_json = { version = "1.0.111", default-features = false, features = ["alloc", "arbitrary_precision"] }
 
 [features]
 default = ["std"]


koute (approved):
Hmm.... as an alternative maybe we should consider serializing those numbers as strings instead, if possible?

The reason being that this is *not* really portable, so anyone who will try to load a JSON containing such numbers in any other language, using any other library, or just using `serde_json` without this flag will silently get incorrect results (if the library doesn't explicitly produce an error in such case, and I think almost all of them don't?).

davxy:
json spec doesn't pose any limit to the number of bits for the `number` type (https://www.json.org/json-en.html).

As `u128` is native to Rust and thus doesn't really need bignum support. I propose to:
1. accept this
2. (optional) propose a `serde_json` patch to exten numbers to u128.

However, I also understand @koute perplexities, i.e. popular libs and other lang implementations typically support up to u64. But this is parsed by a runtime built in Rust right? :-)

davxy (approved):

michalkucharczyk:
> Hmm.... as an alternative maybe we should consider serializing those numbers as strings instead, if possible?
> 
> The reason being that this is _not_ really portable, so anyone who will try to load a JSON containing such numbers in any other language, using any other library, or just using `serde_json` without this flag will silently get incorrect results (if the library doesn't explicitly produce an error in such case, and I think almost all of them don't?).

Would that means that also 64-bit integers would need to be serialized as `String`? (https://www.rfc-editor.org/rfc/rfc8259#section-6):
   > Note that when such software is used, numbers that are integers and
   are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
   sense that implementations will agree exactly on their numeric
   values.

This for sure needs to be solved at some point.

michalkucharczyk:
> json spec doesn't pose any limit to the number of bits for the `number` type (https://www.json.org/json-en.html).
> 
> As `u128` is native to Rust and thus doesn't really need bignum support. I propose to:
> 
> 1. accept this
> 2. (optional) propose a `serde_json` patch to exten numbers to u128.
> 
> However, I also understand @koute perplexities, i.e. popular libs and other lang implementations typically support up to u64

See discussion here: https://github.com/serde-rs/json/issues/846

FWIS actually `serde_json` does not support u128 :), that is why we need arbitrary precision.

koute:
> Would that means that also 64-bit integers would need to be serialized as `String`?

For perfect compatibility with everything, yes. (In languages where numbers are `f64` there's only 52 bits of mantissa, so you can't encode full 64-bit ints with full precision.)

It is an open question whether we care, but it is a footgun that should be at the very least documented.

bkchr (approved):

bkchr:
> It is an open question whether we care, but it is a footgun that should be at the very least documented.

You mean that other languages would fail to parse such a number?

michalkucharczyk:
> > It is an open question whether we care, but it is a footgun that should be at the very least documented.
> 
> You mean that other languages would fail to parse such a number?

It's matter of json lib implementation. Some will do (ruby,python), some not (nodejs). From what I found this c++ json lib [nlohmann/json](https://github.com/nlohmann/json) is not able to parse it.

PR #2990, opened 2024-01-18
better debugging for accountid32 in debug build

pgherveou:


ggwpez (approved):

pgherveou:
I am also for using debug_assertions instead of a custom "force-debug" flag here to make verbose debug impl the default in debug build.

@bkchr  thoughts?

https://github.com/paritytech/polkadot-sdk/blob/0c166ae0194b7337f1c151434483b78b31235549/substrate/primitives/debug-derive/src/impls.rs#L61-L65

bkchr:
We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.

xermicus (approved):

smiasojed (approved):

pgherveou:
> We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.

that's a 8kb change for me when compiling node_template_runtime.compact.wasm size without  debug-derive and this

substrate/primitives/core/src/crypto.rs
```diff
@@ -593,13 +593,21 @@ impl std::fmt::Display for AccountId32 {
 }
 
 impl sp_std::fmt::Debug for AccountId32 {
-	#[cfg(feature = "std")]
+	#[cfg(any(feature = "std", debug_assertions))]
```
liamaharon, line 596: does this also work?

```suggestion
	#[cfg(any(feature = "std", debug_assertions, test))]
```

liamaharon:
> > We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.
> 
> that's a 8kb change for me when compiling node_template_runtime.compact.wasm size without debug-derive

what's the % increase on both node-template and rococo-runtime?

substrate/primitives/core/src/crypto.rs
pgherveou: that should not be necessary ? 
All tests with the exception of integration and benchmark should run with std and debug_assertions on.

pgherveou:
> > > We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.
> > 
> > 
> > that's a 8kb change for me when compiling node_template_runtime.compact.wasm size without debug-derive
> 
> what's the % increase on both node-template and rococo-runtime?

|  wasm                                                                                 | branch - debug-derive  | master     |  diff     |  %      |
|------------------------------------------------------|-------------------------|----------|-------|-------|
| rococo_runtime.compact.compressed.wasm                 | 1549113                         |  1528406 | 20707 |  1.35  |
| node_template_runtime.compact.compressed.wasm   |  243764                         |   243338  | 426    |  0.17   |

liamaharon:
> We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.

@pgherveou I removed from the merge queue because maybe we want to do this instead, since only very small increase to runtime size. @bkchr sound good?

pgherveou:
> > We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.
> 
> @pgherveou I removed from the merge queue because maybe we want to do this instead, since only very small increase to runtime size. @bkchr sound good?

fine for me, this can be done in a follow up, but if we are fine with the size bump then we can use this patch instead, or maybe just get rid of debug-derive?

liamaharon:
Yeah that's what I meant just use your patch then we get it all done in this pr instead of needing to make another one, get reviews again, etc :)

athei:
> We should really check the size impact of just having `Debug` everywhere. And then probably just enable it.

Agreed. The use of `Debug` vs `RuntimeDebug` seems really arbitrary. Do you suggest just getting rid of `RuntimeDebug` all together if the size impact is not too bad after compression?

athei:
I am not really sure what is being discussed here. Your PR does derive `Debug` for the type unconditionally. We can only format ss58 if serde is available. So this is covered. We won't be changing the whole of substrate over to `Debug` in this PR. So I don't get why it was removed from the merge queue. Can you elaborate @liamaharon ?

liamaharon:
> We won't be changing the whole of substrate over to Debug in this PR.

I must have misunderstood Basti's comment. Feel free to merge if this is ready.

ggwpez:
> We won't be changing the whole of substrate over to `Debug` in this PR.

Yea we would first have to re-check the size impact again as well.

athei:
> > We won't be changing the whole of substrate over to Debug in this PR.
> 
> I must have misunderstood Basti's comment. Feel free to merge if this is ready.

I understood it more as a general thing we should do.  But not necessarily in this PR. Could be wrong though. Shouldn't stop us here. So let's go on and merge this. Eradication of `RuntimeDebug` should be a separate PR.

athei (approved):

bkchr (approved):

bkchr:
Created an issue: https://github.com/paritytech/polkadot-sdk/issues/3005

PR #2999, opened 2024-01-19
Make `Slot` and `SlotDuration` transparent

nazar-pc:
I have a use case that for safety requires all contained data structures to be `#[repr(C)]` or `#[repr(transparent)]` and it seemed like non-invasive change.


substrate/primitives/consensus/slots/src/lib.rs
```diff
@@ -135,6 +136,7 @@ impl From<Slot> for u64 {
 	Eq,
 	TypeInfo,
 )]
+#[repr(transparent)]
 pub struct SlotDuration(u64);
```
ggwpez, line 140: Does this change the `PassBy` that is required for the RuntimeAPI?  
I assume not. But maybe it would have made more sense to use `PassByInner` from the beginning 🙈

bkchr: `PassBy` is required by the runtime interface, aka when calling into the host from the runtime api. I don't know any host function that uses `Slot` nor `SlotDuration` :P

bkchr (approved):

ggwpez (approved):

PR #3002, opened 2024-01-19
PoV Reclaim Runtime Side

skunert:
# Runtime side for PoV Reclaim

## Implementation Overview
- Hostfunction to fetch the storage proof size has been added to the PVF. It uses the size tracking recorder that was introduced in my previous PR.
- Mechanisms to use the reclaim HostFunction have been introduced.
  - 1. A SignedExtension that checks the node-reported proof size before and after application of an extrinsic. Then it reclaims the difference.
  - 2. A manual helper to make reclaiming easier when manual interaction is required, for example in `on_idle` or other hooks.
- In order to utilize the manual reclaiming, I modified `WeightMeter` to support the reduction of consumed weight, at least for storage proof size.

## How to use
To enable the general functionality for a parachain:
1. Add the SignedExtension to your parachain runtime. 
2. Provide the HostFunction to the node
3. Enable proof recording during block import

## TODO
- [x] PRDoc


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4972760

cumulus/parachains/common/src/storage_weight_reclaim.rs
```diff
@@ -0,0 +1,517 @@
...
+/// // Perform some operations that consume storage weight.
+///
+/// // Reclaim unused storage weight.
+/// if let Some(reclaimed_weight) = storage_reclaim.reclaim() {
+///     log::info!("Reclaimed {} weight", reclaimed_weight);
+/// }
+/// ```
+pub struct StorageWeightReclaimer<T: Config> {
+	previous_proof_weight: u64,
+	previous_reported_proof_size: Option<u64>,
```
dmitry-markin, line 62: Nit: is it on purpose that one of the variables called `proof_weight`, and another `proof_size`? I can see it as convention that `weight` refers to benchmarked "weight" and `size` refers to actually consumed proof size, but below in `post_dispatch()` there is a variable called `consumed_weight`.

skunert: I tried to use proof_size as the absolute number that is reported by the host function. Weight as the amount consumed or reported by the runtime facilities.

cumulus/parachains/common/src/storage_weight_reclaim.rs
```diff
@@ -0,0 +1,517 @@
...
+		let (Some(current_storage_proof_size), Some(initial_proof_size)) =
+			(get_proof_size(), self.previous_reported_proof_size)
+		else {
+			return None;
+		};
+		let used_weight = current_weight.saturating_sub(self.previous_proof_weight);
+		let used_storage_proof = current_storage_proof_size.saturating_sub(initial_proof_size);
+		let reclaimable = used_weight.saturating_sub(used_storage_proof);
+		log::trace!(target: LOG_TARGET, "Reclaiming storage weight. benchmarked_weight: {used_weight}, consumed_weight: {used_storage_proof}, reclaimable: {reclaimable}");
+		frame_system::BlockWeight::<T>::mutate(|current| {
```
skunert, line 87: I will remove this, there should not happen modification of the storage here.

dmitry-markin (approved):
Looks good to me, but it's a good idea for somebody else to have a closer look at `validate_block` and that the host function is registered everywhere where needed, as I lack the needed knowledge to check this.

cumulus/parachains/common/src/storage_weight_reclaim.rs
```diff
@@ -0,0 +1,497 @@
...
+	/// Creates a new `StorageWeightReclaimer` instance and initializes it with the current storage
+	/// weight and reported proof size from the node.
+	pub fn start(weight_meter: &WeightMeter) -> StorageWeightReclaimer {
+		let previous_remaining_proof_weight = weight_meter.remaining().proof_size();
+		let previous_reported_proof_size = get_proof_size();
+		Self { previous_remaining_proof_weight, previous_reported_proof_size }
+	}
+
+	/// Check the consumed storage weight and calculate the consumed excess weight.
+	fn reclaim(&mut self, remaining_weight: Weight) -> Option<Weight> {
```
davxy, line 74: (Opinionated nit) Personally, I would find it easier to follow if the following variable names were used:
- `initial_proof_size` -> `previous_storage_proof_size`
- `used_storage_proof` -> `reported_used_size`
- `used_weight` -> `expected_used_size`

davxy (approved):
Mostly nitpicks

cumulus/polkadot-parachain/src/service.rs
```diff
@@ -68,11 +68,15 @@ use substrate_prometheus_endpoint::Registry;
...
-type HostFunctions = sp_io::SubstrateHostFunctions;
+type HostFunctions =
+	(sp_io::SubstrateHostFunctions, cumulus_client_service::storage_proof_size::HostFunctions);
 
 #[cfg(feature = "runtime-benchmarks")]
-type HostFunctions =
-	(sp_io::SubstrateHostFunctions, frame_benchmarking::benchmarking::HostFunctions);
+type HostFunctions = (
+	sp_io::SubstrateHostFunctions,
+	cumulus_client_service::storage_proof_size::HostFunctions,
```
davxy, line 77: If we propose this host function "by default" then maybe we also need to suggest to add the new `StorageWeigthReclaim`  signed extension in the runtime signed extensions tuple?
Otherwise what is its usage?

cumulus/parachain-template/node/src/service.rs
```diff
@@ -40,7 +40,10 @@ use substrate_prometheus_endpoint::Registry;
 pub struct ParachainNativeExecutor;
 
 impl sc_executor::NativeExecutionDispatch for ParachainNativeExecutor {
-	type ExtendHostFunctions = frame_benchmarking::benchmarking::HostFunctions;
+	type ExtendHostFunctions = (
+		cumulus_client_service::storage_proof_size::HostFunctions,
```
davxy, line 44: If we propose this host function in the template. Maybe we should also need to add the `StorageWeigthReclaim` signed the runtime signed extensions tuple? (i.e. [here](https://github.com/paritytech/polkadot-sdk/blob/966a886439f4c5ef0bc81f32b8a189d758a2505f/cumulus/parachain-template/runtime/src/lib.rs#L101))

Otherwise what is its usage?

michalkucharczyk:
Code/concept looking good to me. Just a random thought: did you consider adding signed extensions to one of test runtimes (e.g. substrate-test-runtime or cumulus-test-runtime) to add some client-level testing involving more components?

substrate/primitives/weights/src/weight_meter.rs
```diff
@@ -149,6 +149,11 @@ impl WeightMeter {
 	pub fn can_consume(&self, w: Weight) -> bool {
 		self.consumed.checked_add(&w).map_or(false, |t| t.all_lte(self.limit))
 	}
+
+	/// Reclaim the given weight.
+	pub fn reclaim_proof_size(&mut self, s: u64) {
+		self.consumed.saturating_reduce(Weight::from_parts(0, s));
+	}
```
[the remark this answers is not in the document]
skunert: I was thinking about this, however the `WeightMeter` was not offering any mechanism to reduce the consumed weight at all. I think because usually, once you have consumed some weight there is no way to get it back. I want to prevent users from even thinking about reclaiming ref time. The current API makes it very clear that you shall only ever reclaim proof size.

cumulus/parachain-template/node/src/service.rs
skunert: > IIUC the StorageWeightReclaimer alone can't work correctly if we don't also return the extra weight to frame_system::BlockWeight ( by means of this signed extension hooks right?
Or I'm missing some other means of doing this? E.g. the user directly altering frame_system::BlockWeight in some other hook.

The `StorageWeightReclaimer` is a separate mechanism from the `SignedExtension`. It is designed to be used inside the `on_idle` hook, where weight management happens manually via `WeightMeter`. Users need to check whether the operation they plan fits into the remaining weight. In the end, `on_idle` will return the consumed weight and it will be accounted for after the hook finished here:
https://github.com/paritytech/polkadot-sdk/blob/150a360c410d9a051ac04022bb69eb1458396db6/substrate/frame/executive/src/lib.rs#L607-L614

This means that the weightreclaimer should only modify the value in the weightmeter, not the storage value itself. 

> If we propose this host function in the template. Maybe we should also need to add the StorageWeigthReclaim signed extension to the runtime signed extensions tuple? (i.e. [here](https://github.com/paritytech/polkadot-sdk/blob/966a886439f4c5ef0bc81f32b8a189d758a2505f/cumulus/parachain-template/runtime/src/lib.rs#L101))
>
> Otherwise what is its usage?

This is a good point, I will add it.

cumulus/pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -167,9 +174,11 @@ where
 			.replace_implementation(host_default_child_storage_next_key),
 		sp_io::offchain_index::host_set.replace_implementation(host_offchain_index_set),
 		sp_io::offchain_index::host_clear.replace_implementation(host_offchain_index_clear),
+		cumulus_primitives_proof_size_hostfunction::storage_proof_size::host_storage_proof_size
+			.replace_implementation(host_storage_proof_size),
 	);
 
-	run_with_externalities::<B, _, _>(&backend, || {
+	run_with_externalities_and_recorder::<B, _, _>(&backend, &mut recorder, || {
```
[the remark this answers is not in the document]
skunert: I mean we need to supply the recorder everywhere where host functions are potentially being called. And since the `CheckInherent` type is user supplied, host functions could be called there. 

The two closures are separate because we want to start with a fresh overlay and externalities. But this is a good point, I now reset the recorder too between these two code sections.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5238557

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5262394

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5266732

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5267164

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5271690

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5273918

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5273917

cumulus/primitives/storage-weight-reclaim/src/lib.rs
```diff
@@ -0,0 +1,516 @@
...
+		let consumed_weight = post_dispatch_proof_size.saturating_sub(pre_dispatch_proof_size);
+
+		let storage_size_diff = benchmarked_weight.abs_diff(consumed_weight as u64);
+		frame_system::BlockWeight::<T>::mutate(|current| {
+			if consumed_weight > benchmarked_weight {
+				log::error!(
+					target: LOG_TARGET,
+					"Benchmarked storage weight smaller than consumed storage weight. benchmarked: {benchmarked_weight} consumed: {consumed_weight}"
+				);
+				current.accrue(Weight::from_parts(0, storage_size_diff), info.class)
```
bkchr, line 169:
```suggestion
				current.saturating_accrue(Weight::from_parts(0, storage_size_diff), info.class)
```

cumulus/primitives/storage-weight-reclaim/src/lib.rs
```diff
@@ -0,0 +1,516 @@
...
+			} else {
+				log::trace!(
+					target: LOG_TARGET,
+					"Reclaiming storage weight. benchmarked: {benchmarked_weight}, consumed: {consumed_weight}"
+				);
+				current.reduce(Weight::from_parts(0, storage_size_diff), info.class)
```
bkchr, line 175:
```suggestion
				current.saturating_reduce(Weight::from_parts(0, storage_size_diff), info.class)
```

cumulus/primitives/storage-weight-reclaim/src/lib.rs
```diff
@@ -0,0 +1,516 @@
...
+		_info: &sp_runtime::traits::DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> Result<Self::Pre, sp_runtime::transaction_validity::TransactionValidityError> {
+		Ok(get_proof_size())
+	}
+
+	fn post_dispatch(
+		pre: Option<Self::Pre>,
+		info: &DispatchInfoOf<Self::Call>,
+		_post_info: &PostDispatchInfoOf<Self::Call>,
```
bkchr, line 144: If the call returned some weight, [this logic](https://github.com/paritytech/polkadot-sdk/blob/e675bb44d5d9840f11dfa6f23bfeabc27589cbf5/substrate/frame/system/src/extensions/check_weight.rs#L262-L266) will be triggered in `CheckWeight`. This would ultimately lead to reducing the proof size twice. Thus, we need to take this into account here.

bkchr: Otherwise we would require to have a forced order of `CheckWeight` and this extension.

cumulus/pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -265,6 +274,19 @@ fn validate_validation_data(
...
 
+/// Run the given closure with the externalities and recorder set.
+fn run_with_externalities_and_recorder<B: BlockT, R, F: FnOnce() -> R>(
+	backend: &TrieBackend<B>,
+	recorder: &mut SizeOnlyRecorderProvider<HashingFor<B>>,
+	execute: F,
+) -> R {
+	let mut overlay = sp_state_machine::OverlayedChanges::default();
+	let mut ext = Ext::<B>::new(&mut overlay, backend);
+	recorder.reset();
```
bkchr, line 285: Actually realizing that my comment last time was not correct :see_no_evil: 

Given that we reset the `recorder` here, it is fine to also use this function for `check_inherents`. So can you do this? :D It should not be used there, but better than having it fail the validation.

bkchr (approved):
Some last nitpicks, otherwise it should be good to go :+1: 

Nice work :+1:

cumulus/primitives/storage-weight-reclaim/src/lib.rs
skunert: These are already saturating: https://github.com/paritytech/polkadot-sdk/blob/ca382f32033d8350d49934b30680d8c30dd9bdfc/substrate/frame/support/src/dispatch.rs#L475-L487

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5300234

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5307234

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5311174

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5311172

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5314083

AndWeHaveAPlan (approved):

alvicsam (approved):

PR #3013, opened 2024-01-22
add NotBefore variant for DispatchTime

xlc:
Useful for many cases.
For example, we cannot use `DispatchTime::At` for a proposal in OpenGov with block before the end of max voting time, otherwise there is a risk that the proposal will be executed after the `At` block and therefore cannot be dispatched at all.
`After` is not a good thing is some cases before it could be dispatched at any time (e.g. holiday).
`NotBefore` allow us to specify a date and if for some reason the voting takes the full duration, the proposal can still be enacted.

Let me know what do you think and then I will finish this.

TODO: 
- [ ] some tests
- [ ] PRdoc?

### DIFF
--- a/substrate/frame/support/src/traits/schedule.rs
+++ b/substrate/frame/support/src/traits/schedule.rs
@@ -40,13 +40,16 @@ pub enum DispatchTime<BlockNumber> {
 	At(BlockNumber),
 	/// After specified number of blocks.
 	After(BlockNumber),
+	/// At or after specified block.
+	NotBefore(BlockNumber),
 }
 
-impl<BlockNumber: Saturating + Copy> DispatchTime<BlockNumber> {
+impl<BlockNumber: Saturating + Copy + Ord> DispatchTime<BlockNumber> {
 	pub fn evaluate(&self, since: BlockNumber) -> BlockNumber {
 		match &self {
 			Self::At(m) => *m,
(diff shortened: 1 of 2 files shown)


substrate/frame/support/src/traits/schedule.rs
```diff
@@ -40,13 +40,16 @@ pub enum DispatchTime<BlockNumber> {
 	At(BlockNumber),
 	/// After specified number of blocks.
 	After(BlockNumber),
+	/// At or after specified block.
+	NotBefore(BlockNumber),
```
xlc, line 44: let me know if you can think a better name

joepetrowski:
> For example, we cannot use DispatchTime::At for a proposal in OpenGov with block before the end of max voting time

Yes, we can. The Referenda pallet uses `At` with the behavior you describe:

https://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.6.0/substrate/frame/referenda/src/lib.rs#L877

I suppose we could make this logic first class with scheduler, but it's quite simple and if `now` is already past the previously stated intention for dispatch time, then whatever is doing the scheduling can decide if it's invalid and should be dropped or should be rescheduled.

xlc:
Oh wasn't aware of this behaviour so this isn't necessary. Let me know if I should continue this.

PR #3022, opened 2024-01-22
Add unit impl for XcmAssetTransfers trait

franciscoaguirre:
If an Executor is required in some pallet's config, a mock might be provided where `Executor = ()`. `()` already implements `ExecuteXcm`, but not `XcmAssetTransfers`, which is also related. This PR just fixes that so you can skip having to create a whole xcm configuration, which is non-trivial right now.


franciscoaguirre:
bot fmt

acatangiu (approved):

PR #3029, opened 2024-01-23
Snowbridge Ethereum Deneb fork preparation

claravanstaden:
- Prepares for the Deneb hardfork on Sepolia testnet on 31 January (needs to be deployed to Rococo before then)
- Removes `beacon-minimal-spec` flag for simpler config
- Adds test comments


cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -203,7 +203,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("bridge-hub-rococo"),
 	impl_name: create_runtime_str!("bridge-hub-rococo"),
 	authoring_version: 1,
-	spec_version: 1_006_000,
+	spec_version: 1_007_000,
```
svyatonik, line 206: I have no idea - whether it should be bumped here, or during release?

svyatonik (commented):
Don't want to block this PR, because most of changes are in the Snowfork pallets and are already reviewed. But need to ask a question before.

Looking at Eth hardfork times, won't it be hard for us to follow that schedule in deployment, given current deployment pipeline (fix in Snowfork repo -> update in `polkadot-sdk` -> wait for crates publish -> upgrade dependencies in the fellowship repo -> runtime upgrade)? Like `Paris` has been activated 9 days after the previous hardfork (`Bellatrix`). Do you have some safety switch to turn off bridge if we'll fail to deploy on time? Or we'll be trying to do that on time.

Also I remember some accidental hard forks that have happened because of differences in client implementations. Do you also have some solution for bridge fixes in such case (e.g.: force switch forks). Just curious

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
claravanstaden: Erm I'm also not sure. I wonder now if it shouldn't be a minor bump, like `1_006_001`? @bkontur? 😅

bridges/snowbridge/parachain/runtime/test-common/Cargo.toml
```diff
@@ -89,7 +89,7 @@ snowbridge-system-runtime-api = { path = "../../pallets/system/runtime-api", def
 [dev-dependencies]
 static_assertions = "1.1"
 bridge-hub-test-utils = { path = "../../../../../cumulus/parachains/runtimes/bridge-hubs/test-utils" }
-bridge-runtime-common = { path = "../../../../bin/runtime-common", features = ["integrity-test"] }
+bridge-runtime-common = { path = "../../../../../bridges/bin/runtime-common", features = ["integrity-test"] }
```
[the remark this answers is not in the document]
claravanstaden: My bad, fixed in https://github.com/paritytech/polkadot-sdk/pull/3029/commits/57f9b19e69f71acd7bacf17d8bc48989d0b85199.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
bkontur: I would not change `spec_version` here, last time for 1.6.0 tag/release, they did it in the release branch and when released they backport that to master, so I would better revert this

bkontur (approved):
I didn't really check your subtree stuff, at the first glance looks ok,
I saw you removed minimal/mainet features, is it possible to remove this:
```
fast-runtime = [
	"bridge-hub-rococo-runtime/fast-runtime",
]
```
and change to:
```
fast-runtime = [ ]
```
in the `polkadot-parachain` Cargo.toml?

claravanstaden:
@svyatonik thanks for raising this, this is a valid point. A few things to note. 

> Looking at Eth hard-fork times, won't it be hard for us to follow that schedule in deployment, given current deployment pipeline (fix in Snowfork repo -> update in `polkadot-sdk` -> wait for crates publish -> upgrade dependencies in the fellowship repo -> runtime upgrade)? Like `Paris` has been activated 9 days after the previous hard-fork (`Bellatrix`). 

After the Ethereum merge, there are 2 Ethereum chains, the execution chain (the "old" Ethereum chain as it was before the merge) and the consensus chain (the new Beacon chain). For the Merge, it required an upgrade on both chains. Paris was the hard-fork on the execution chain and Bellatrix was the hardfork on the consensus chain. So they were essentially the same upgrade, just on the two different chains. We do not interact with the execution chain for our Ethereum light client, so we would just have prepared for the Bellatrix upgrade. Never say never, but it highly unlikely for hard-fork to be days from each other.

> Also I remember some accidental hard forks that have happened because of differences in client implementations. Do you also have some solution for bridge fixes in such case (e.g.: force switch forks). Just curious

Iirc those hard-forks were related to the execution chain, and it has not been required since the consensus chain was launched. Consensus client changes, so far, have been planned months in advance and usually have required code changes (like the changes in the PR, to accommodate changes in the ExecutionHeaderPayload struct). We could have merged these changes in months ago but we decided to hold off since we had other higher priorities. 

This is usually how these hard-forks work:
- Code changes announced here: https://github.com/ethereum/consensus-specs
- We make the required code changes
- Ethereum announces the expected epoch that the hard-fork will take place on

If we had to, we could configure the hard-fork versions via a governance call. I am not sure if we want to do that now though. We do run the risk that the time between the mainnet epoch announcement and fork is shorther than the Parity release train. cc @vgeddes 

> Do you have some safety switch to turn off bridge if we'll fail to deploy on time? Or we'll be trying to do that on time.

We can [halt the light client](https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/parachain/pallets/ethereum-client/src/lib.rs#L265) if need be, but it would obviously be preferable not to have "downtime" where we can't process messages. If there should be an unexpected Ethereum hard-fork because of the bug, we could halt the bridge until we configured the new fork version (cc @yrong please correct me if I am wrong).

claravanstaden:
> I didn't really check your subtree stuff, at the first glance looks ok, I saw you removed minimal/mainet features, is it possible to remove this:
> 
> ```
> fast-runtime = [
> 	"bridge-hub-rococo-runtime/fast-runtime",
> ]
> ```
> 
> and change to:
> 
> ```
> fast-runtime = [ ]
> ```
> 
> in the `polkadot-parachain` Cargo.toml?

@bkontur it is unfortunately not possible because the fast-runtime feature needs to be propagated to the bridge hub runtime to toggle between the different fork versions for local testing and Rococo.

svyatonik (approved):

bkontur:
> > I didn't really check your subtree stuff, at the first glance looks ok, I saw you removed minimal/mainet features, is it possible to remove this:
> > ```
> > fast-runtime = [
> > 	"bridge-hub-rococo-runtime/fast-runtime",
> > ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and change to:
> > ```
> > fast-runtime = [ ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in the `polkadot-parachain` Cargo.toml?
> 
> @bkontur it is unfortunately not possible because the fast-runtime feature needs to be propagated to the bridge hub runtime to toggle between the different fork versions for local testing and Rococo.

I thought you removed that https://github.com/paritytech/polkadot-sdk/pull/3029/files#diff-e4f9240a3174984f0bd0785096798ab4025b60a91ebf4c4afd6e8bffb425d487L98-L99
or is there other place where it is needed?

claravanstaden:
> I thought you removed that https://github.com/paritytech/polkadot-sdk/pull/3029/files#diff-e4f9240a3174984f0bd0785096798ab4025b60a91ebf4c4afd6e8bffb425d487L98-L99
or is there other place where it is needed?

@bkontur yip, it is still needed in the BridgeHub runtime to configure the fork versions: https://github.com/paritytech/polkadot-sdk/pull/3029/files#diff-f5e5608adb7096db0c665bb20e5298c113904c48b16633684a7cad72d95b4962R595

bridges/snowbridge/parachain/pallets/ethereum-client/src/config/mod.rs
```diff
@@ -54,3 +45,10 @@ pub const PUBKEY_SIZE: usize = 48;
 pub const SIGNATURE_SIZE: usize = 96;
 
 const_assert!(SYNC_COMMITTEE_BITS_SIZE == SYNC_COMMITTEE_SIZE / 8);
+
+pub const SLOTS_PER_EPOCH: usize = 32;
+pub const EPOCHS_PER_SYNC_COMMITTEE_PERIOD: usize = 256;
+pub const SYNC_COMMITTEE_SIZE: usize = 512;
+pub const SYNC_COMMITTEE_BITS_SIZE: usize = SYNC_COMMITTEE_SIZE / 8;
```
ggwpez, line 52: Always nice to have one line of doc on public stuff 😋

ggwpez (approved):
Approving for the git dependency cleanup.

bridges/snowbridge/parachain/pallets/ethereum-client/src/mock.rs
```diff
@@ -98,187 +89,90 @@ pub fn get_message_verification_header() -> CompactExecutionHeader {
...
 
-#[cfg(not(feature = "beacon-spec-minimal"))]
-pub mod mainnet {
-	use super::*;
+impl frame_system::Config for Test {
+	type BaseCallFilter = frame_support::traits::Everything;
+	type OnSetCode = ();
+	type BlockWeights = ();
+	type BlockLength = ();
+	type DbWeight = ();
```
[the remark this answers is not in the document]
claravanstaden: I replaced it here. However, our other mock modules relies on `AccountId` = `AccountId32`, and this macro seems to rely on `AccountId` being `u64`.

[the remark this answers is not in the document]
ggwpez: The macro allows to overwrite custom ones, so you can just put `type AccountId = AccountId32` in there like usual.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5053088

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5053089

bridges/snowbridge/parachain/pallets/ethereum-client/src/mock.rs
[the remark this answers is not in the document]
claravanstaden: Yip I tried that but got this error:

`#[derive_impl(frame_system::config_preludes::TestDefaultConfig as frame_system::DefaultConfig)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `AccountId32`, found `u64`
`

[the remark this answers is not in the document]
svyatonik: We're using `TestDefaultConfig` + `AccountId = u64` e.g. here: https://github.com/paritytech/polkadot-sdk/blob/84a246c24a5d3c410a2e4f7d63c0215decddb2b1/bridges/bin/runtime-common/src/mock.rs#L148

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5057824

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5058096

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5058459

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/tests.rs
```diff
@@ -222,6 +223,72 @@ mod bridge_hub_westend_tests {
...
+			collator_session_keys(),
+			bp_bridge_hub_rococo::BRIDGE_HUB_ROCOCO_PARACHAIN_ID,
+			Box::new(|call| RuntimeCall::System(call).encode()),
+			vec![
+				(snowbridge_pallet_outbound_queue::Nonce::<Runtime>::hashed_key_for::<ChannelId>(
+					channel_id_one,
+				)
+				.to_vec(), 0u64.encode()),
+				(snowbridge_pallet_inbound_queue::Nonce::<Runtime>::hashed_key_for::<ChannelId>(
+					channel_id_one,
```
bkontur, line 243:
```suggestion
					channel_id_two,
```

@claravanstaden typo or intention?

cumulus/parachains/runtimes/test-utils/src/test_cases.rs
```diff
@@ -91,3 +91,55 @@ pub fn change_storage_constant_by_governance_works<Runtime, StorageConstant, Sto
...
+		.with_session_keys(collator_session_key.session_keys())
+		.with_para_id(runtime_para_id.into())
+		.with_tracing()
+		.build();
+	runtime.execute_with(|| {
+		initialize_storage();
+	});
+	runtime.execute_with(|| {
+		// encode `kill_storage` call
+		let kill_storage_call = runtime_call_encode(frame_system::Call::<Runtime>::set_storage {
```
bkontur, line 124: @claravanstaden small nits, name `kill_storage` here is kind of misleading, because it does `set_storage`.
Also I think that this introduced `set_storage_keys_by_governance_works` is not necessary and its usage can be replaced with four calls of `change_storage_constant_by_governance_works` in-place.

claravanstaden: Ok cool! Would it be OK if I create an internal ticket for this? Just because the Ethereum Deneb fork is tomorrow.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/tests.rs
claravanstaden: I think this is intentional because if I change it, the test fails. @alistair-singh can provide more context here and perhaps add a comment why.

claravanstaden:
@ggwpez several runtime migration check CI tasks are failing with:

> Other("Max message size for channel is too large. This means that the V5 migration can be front-run and an attacker could place a large message just right before the migration to make other messages un-decodable. Please either increase `MaxPageSize` or decrease the `max_message_size` for this channel.")

It is for the `asset-hub-rococo`, `asset-hub-westend`, `bridge-hub-rococo`, `bridge-hub-westend` and `collectives-westend` runtimes. (e.g. https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5065025)

Could you please point me in the right direction to correct this config?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/tests.rs
alistair-singh: This test assets that only `channel_one` nonces are reset and that `channel_two` nonces are left alone.

cumulus/parachains/runtimes/test-utils/src/test_cases.rs
alistair-singh: Originally I used kill_storage to reset nonces so this is leftover naming from that and needs to change. 

The issue with `change_storage_constant_by_governance_works` is that it is bound to storage items which implement [Get\<StorageConstantType>](https://github.com/paritytech/polkadot-sdk/pull/3029/files/8bb432a4e8ad52ed4a9b34e81f6ce27a82be81b7#diff-4db90058dd96cd36eaa6d1751e8937386185c594d86dca5f2bdc689554430f97R42). This limits its usage to storage constants that created in `parameter_types!` macro and only tests single keys. This will not work for pallets with a StorageMap. `set_storage_keys_by_governance_works` does this in a more general way by allowing two closures, one to initialize storage before modification, and one to asset storage after modification.

The tests that use `change_storage_constant_by_governance_works` could be re-implemented in terms of `set_storage_keys_by_governance_works` as it is more general, but that would affect 7 runtime tests which use it. Should we rather do that?

claravanstaden:
@ggwpez Thanks for the fix! 🙌🏻 Can this PR be merged or do we need to wait for https://github.com/paritytech/polkadot-sdk/pull/3117?

cumulus/parachains/runtimes/test-utils/src/test_cases.rs
bkontur: @alistair-singh yes, you're right, `change_storage_constant_by_governance_works` was dedicated more for `parameter_types! { pub storage..` stuff. Or maybe `change_storage_constant_by_governance_works` can internally just call `set_storage_keys_by_governance_works`, anyway, if we could make it nicer, lets do it, but this is very low prio,

@claravanstaden we don't have access to [your](https://linear.app/snowfork/issue/SNO-840), it is related to the `polkadot-sdk` code directly, so it is also fine to create issue here.

Issue #1570, opened 2023-09-14
Add APIs for block pruning manually


nazar-pc:
As discussed in https://github.com/paritytech/substrate/issues/14758 there are currently APIs to finalize blocks (`Finalizer::apply_finality()` and `Finalizer::finalize_block()`) and blocks pruning is at a fixed offset from that, but we have a use case in Subspace where we want to have that offset dynamic.

Essentially we need an API to prune blocks manually and decouple it from block finalization (finalized blocks will be much newer than pruned blocks). The same for state.

While at it having API to prune headers and `--headers-pruning` CLI argument would have been both nice to have as well since currently they are not pruned and not allow for node to occupy bounded amount of space.

kianenigma:
This would be great. I think it is a good UX improvement to allow full nodes to run with all 3 pruning types set, and it should only occupy a more or less constant amount of space on disk.

nazar-pc:
We have a downstream issue at Subspace (https://github.com/subspace/subspace/issues/2114) and might be able to sponsor work on this if someone is interested.

kayabaNerve:
I'm specifically interested in an API where clients can issue over RPC a 'safe-to-prune' command to advance which blocks are pruned. This would allow reclaiming disk space after indexing, yet guarantees the ability to perform indexing.

If this issue is to be focused on header-pruning or a time-based prune (compared to current block count based prune), I'd be happy to open a new issue, yet this sounds close enough it may be best to simply tag in here.

(I hadn't prior opened an issue as it seemed easy enough I'd just write an impl myself at one point, yet I'm commenting here now as this sounds like a prerequisite I assumed would already be sufficiently available, but by this issue's existence, isn't.)

nazar-pc:
FYI @shamil-gadelshin on Subspace team will be working on this once he is back after holidays.

shamil-gadelshin:
My current plan is to introduce a CLI flag similar to state and block pruning. I'm going to remove data from db, in-memory header and header metadata caches. The current fork calculation and pruning depend on the block headers for already pruned blocks so I'm going to add a temporary in-memory cache for "block headers marked for pruning" and prune them with some delay.

nazar-pc:
> The current fork calculation and pruning depend on the block headers for already pruned blocks so I'm going to add a temporary in-memory cache for "block headers marked for pruning" and prune them with some delay.

What if node restarts?

shamil-gadelshin:
[...]

PR #3033, opened 2024-01-23
Add block header pruning.

shamil-gadelshin:
This PR introduces block header pruning for blockchain-db. Original issue: https://github.com/paritytech/polkadot-sdk/issues/1570
This is a draft PR to show a possible solution while having several open questions.

Pruning block headers include:
- removing data from DB using `columns::KEY_LOOKUP`, `columns::HEADER`, and block hash.
- removing data from block header in-memory cache
- removing data from header metadata in-memory cache (the last two using `remove_header_metadata`)

The PR consists of two parts:
- change stale fork calculation: we needed this change because the previous algorithm relied on the block header persistence
- block header pruning:  we enable it using the new CLI argument `---prune-block-headers` which defaults to false

Open questions:
- What is the correct command to test the whole solution locally?
- Are there other places to clean data from during the block header pruning?
- The test `prune_blocks_on_finalize_and_reorg` marks several blocks in a row as finalized and commits the transaction later in contrast with `prune_blocks_on_finalize` which finalized every block in a separate transaction. If we modify `prune_blocks_on_finalize_and_reorg` test to query a header from the pruned fork we will get one, however, the database won't contain it. It seems that the cache gets polluted by dirty reads originating in meta updates because meta updates are saved separately and after the transaction. Is there a way to prevent it? I appreciate an advice here.

cc @nazar-pc @bkchr

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20091,9 +20091,9 @@ checksum = "222a222a5bfe1bba4a77b45ec488a741b3cb8872e5e499451fd7d0129c9c7c3d"
 
 [[package]]
 name = "thiserror"
-version = "1.0.50"
+version = "1.0.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9a7210f5c9a7156bb50aa36aed4c95afb51df0df00713949448cf9e97d382d2"
+checksum = "6e3de26b0965292219b4287ff031fcba86837900fe9cd2b34ea8ad893c0953d2"
 dependencies = [
  "thiserror-impl",
 ]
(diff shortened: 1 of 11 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5008692

shamil-gadelshin:
Closed in favor separate PRs:  https://github.com/paritytech/polkadot-sdk/pull/3962

Polkadot-Forum:

https://forum.polkadot.network/t/block-header-pruning/7198/1

PR #3042, opened 2024-01-24
statement-distribution: Ensure we deactivate leaves before activating

bkchr:
As activation can fail, we ensure that we don't miss deactivation of leaves.

### DIFF
--- a/polkadot/node/network/statement-distribution/src/lib.rs
+++ b/polkadot/node/network/statement-distribution/src/lib.rs
@@ -319,8 +319,10 @@ impl<R: rand::Rng> StatementDistributionSubsystem<R> {
 				if let Some(ref activated) = activated {
 					let mode = prospective_parachains_mode(ctx.sender(), activated.hash).await?;
 					if let ProspectiveParachainsMode::Enabled { .. } = mode {
-						v2::handle_active_leaves_update(ctx, state, activated, mode).await?;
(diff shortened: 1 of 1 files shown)


sandreim (approved):

polkadot/node/network/statement-distribution/src/lib.rs
```diff
@@ -319,8 +319,10 @@ impl<R: rand::Rng> StatementDistributionSubsystem<R> {
 				if let Some(ref activated) = activated {
 					let mode = prospective_parachains_mode(ctx.sender(), activated.hash).await?;
 					if let ProspectiveParachainsMode::Enabled { .. } = mode {
-						v2::handle_active_leaves_update(ctx, state, activated, mode).await?;
+						// Ensure we deactivate first, before we activate the new block.
+						// Activation can fail and then we would skip the deactivation.
 						v2::handle_deactivate_leaves(state, &deactivated);
+						v2::handle_active_leaves_update(ctx, state, activated, mode).await?;
```
rphmeier, line 325: This may cause issues as ordering was somewhat sensitive here. 

Reference counts for active leaves may drop to zero and then be incremented back to 1 again. It would be more reliable to just ensure that we `handle_deactivate_leaves` even if `handle_active_leaves_update` fails.

PR #3049, opened 2024-01-24
Fix treasury benchmarks when no SpendOrigin

bgallois:
### Issue

It was impossible to benchmark the pallet_treasury when `SpendOrigin = frame_support::traits::NeverEnsureOrigin<u64>;` was specified.

### Done

- [x] Use `weight = 0` for all extrinsics that are un-callable with no `SpendOrigin`.
- [x] Fix benchmarks for extrinsics requiring a Spend even if  `SpendOrigin = frame_support::traits::NeverEnsureOrigin<u64>;`


bgallois:
I'm unable to sign the CLA: `Content Encoding Error. An error occurred during a connection to cla.parity.io. `

substrate/frame/treasury/src/benchmarking.rs
```diff
@@ -215,6 +215,8 @@ mod benchmarks {
 		Ok(())
 	}
 
+	// This benchmark is short-circuited if `SpendOrigin` cannot provide
```
ggwpez, line 218:
```suggestion
	/// This benchmark is short-circuited if `SpendOrigin` cannot provide
```
Should be possible to use real doc comments here.

ggwpez (approved):
Thanks!

ggwpez:
> I'm unable to sign the CLA: `Content Encoding Error. An error occurred during a connection to cla.parity.io. `

Hm, can you try again?  
Not sure whom to ping here @mutantcornholio maybe?

bgallois:
> > I'm unable to sign the CLA: `Content Encoding Error. An error occurred during a connection to cla.parity.io. `
> 
> Hm, can you try again? Not sure whom to ping here @mutantcornholio maybe?

Still no luck.

bkchr (approved):

bkchr (changes requested):
This is actually not correct. You are right that the benchmarks are not correct the way they are written, but these calls clearly don't have zero weight when `SpendOrigin` fails. If you look at the functions of the dispatchables, you will see that we access `Spends` aka reading from the DB which is one of the most expensive operations we have.

bgallois:
> This is actually not correct. You are right that the benchmarks are not correct the way they are written, but these calls clearly don't have zero weight when `SpendOrigin` fails. If you look at the functions of the dispatchables, you will see that we access `Spends` aka reading from the DB which is one of the most expensive operations we have.

In fact, these extrinsics should not be callable when `SpendOrigin = frame_support::traits::NeverEnsureOrigin<u64>;`. In this case, we can authorize weight=0 so that the benchmark can pass for this pallet. Another solution is to benchmark the path until failure occurs and use this weight as the default weight.

ggwpez:
>You are right that the benchmarks are not correct the way they are written, but these calls clearly don't have zero weight when SpendOrigin fails.

Yes true. But he is fixing the case that `SpendOrigin` is configured to be impossible to pass.  
Otherwise the `try_success_origin` for benchmarks is advised to return something that always works by using a [feature flag](https://github.com/paritytech/polkadot-sdk/blob/0226b55f9ffff28821458a71c84eb5d3e75f1794/cumulus/parachains/runtimes/collectives/collectives-westend/src/ambassador/origins.rs#L132). Although normally its better to go through the normal setup to also benchmark the effort of the `try_success_origin`.

bkchr:
https://github.com/paritytech/polkadot-sdk/blob/acd043bc5fc9acfa384b69c33e341f925ef250a7/substrate/frame/treasury/src/lib.rs#L813-L815

https://github.com/paritytech/polkadot-sdk/blob/acd043bc5fc9acfa384b69c33e341f925ef250a7/substrate/frame/treasury/src/lib.rs#L856-L861

https://github.com/paritytech/polkadot-sdk/blob/acd043bc5fc9acfa384b69c33e341f925ef250a7/substrate/frame/treasury/src/lib.rs#L910-L911

As we can see, there are reads. If `SpendOrigin` isn't set, this prevents that any value is stored in `Spends`. However, it doesn't prevent that these other functions are doing a read.

bgallois:
> > You are right that the benchmarks are not correct the way they are written, but these calls clearly don't have zero weight when SpendOrigin fails.
> 
> Yes true. But he is fixing the case that `SpendOrigin` is configured to be impossible to pass. There should be no DB reads in that case, right? Otherwise the `try_success_origin` for benchmarks is advised to return something that always works by using a [feature flag](https://github.com/paritytech/polkadot-sdk/blob/0226b55f9ffff28821458a71c84eb5d3e75f1794/cumulus/parachains/runtimes/collectives/collectives-westend/src/ambassador/origins.rs#L132). Although its better to go through the normal setup to also benchmark the effort of the `try_success_origin`.

Alright, but in this case, we need to adapt the three benchmarks(check_status, void_spend and payout) that always failed not because of a `SpendOrigin` check, but because the extrinsic will fail when reading the `Spend` storage, which will always be empty. For these, the weight will be one signed origin check and then one read to the database.

bkchr:
> Alright, but in this case, we need to adapt the three benchmarks(check_status, void_spend and payout) that always failed not because of a `SpendOrigin` check,

Yes, I actually wanted this. I should have been more clear :D

mutantcornholio:
> > I'm unable to sign the CLA: `Content Encoding Error. An error occurred during a connection to cla.parity.io. `
> 
> Hm, can you try again? Not sure whom to ping here @mutantcornholio maybe?

Thanks for report, I'm on it

substrate/frame/treasury/src/benchmarking.rs
```diff
@@ -104,12 +104,39 @@ fn create_spend_arguments<T: Config<I>, I: 'static>(
...
 }
 
+// Use this `spend` insertion to bypass the `SpendOrigin` check. This is helpful
+// for benchmarking `payout`, `check_status`, and `void_spend`, which require
+// a spend in storage even if `SpendOrigin` cannot provide a successful origin.
+fn force_add_spend<T: Config<I>, I: 'static>(
+	asset_kind: Box<T::AssetKind>,
+	amount: AssetBalanceOf<T, I>,
+	beneficiary: Box<BeneficiaryLookupOf<T, I>>,
+) -> Result<(), &'static str> {
```
bkchr, line 114: This is not correct. What you should do is to skip adding something to `Spend` when `T::SpendOrigin::try_successful_origin()` fails. 

So,
```
		if let Ok(origin) = T::SpendOrigin::try_successful_origin() {
		Treasury::<T, _>::spend(
			origin,
		force_add_spend::<T, _>(
			Box::new(asset_kind.clone()),
			amount,
			Box::new(beneficiary_lookup),
			None,
		)?;
}
```

is the logic I would expect for these benchmarks below.

bgallois: If we omit adding something to the spend, the benchmark will fail during execution with the 'InvalidIndex' error. While it is possible to bypass this error using a solution like [this](https://substrate.stackexchange.com/questions/8215/benchmark-failing-extrinsics), it appears error-prone.

As extrinsic benchmark cover worst case scenario, It seems not far-fetched that even with `SpendOrigin = frame_support::traits::NeverEnsureOrigin<u64>` we take this impossible case.

mutantcornholio:
@bgallois could you try signing the CLA again please?

bgallois:
> @bgallois could you try signing the CLA again please?

Thanks it is working :)

substrate/frame/treasury/src/benchmarking.rs
bkchr: I mean we should be able to use the same condition as above to check either on success or error?

CC @ggwpez

bgallois: I know that the `#[extrinsic_call]` cannot be included inside a conditional block.

If there is a way to perform a benchmark with a condition, I have never seen it.

ggwpez: You can use `#[block]` and put the `if` inside:  
```rust
#[block]
{
    if condition {
        ...
    } else {
        ...
    }
}
```
Condition should be a `bool` already to not mess up the timing. Otherwise you can create a new benchmark just for the failing case and then use the maximum of the two as weight.

bgallois: Quite interesting, I will implement that. I see in the documentation that "the #[block] syntax is broad and will benchmark any code contained within the block the attribute is attached to".  So, we consider the if condition check negligible in the timing, right?

ggwpez: Yea negligible, but since we measure worst case timing it is actually not harming us since it should only slow it down.

bgallois: I understand, but I have a conceptual issue with these benchmarks because the worst-case scenario for them (most general case) is when the extrinsic finds a spend and performs what it should do. In the (edge) case where `SpendOrigin = frame_support::traits::NeverEnsureOrigin<u64>;`, no spend can ever be added to the Storage. As a result, with this setting, these three extrinsics should never be used by clients, except in cases of error calls.

Complicating the benchmark with conditionals in a solution that slightly overestimate the weight versus charging the most general case where a spend is found in storage (also overestimating weight if this is an impossible case to obtain in production) tends to yield the same result.

I'm not sure if I'm explaining myself very clearly.

bkchr: > As a result, with this setting, these three extrinsics should never be used by clients, except in cases of error calls.

Yes, no one should use them. However, an attacker will may want to use them to spam your chain. This is the reason that it needs a proper weight, aka account for the read.

bgallois: The more general case (processing the Spend) as a weight that is superior to the edge case (one read then an error), thus posing no risk of spamming…

bkchr: > The worst-case scenario occurs when a valid spend is found and paid.

But this isn't the worst-case when you have `SpendOrigin = NeverEnsureOrigin<u64>;`. Then the worst-case is basically the best case as well and just means that there is only one read of the storage.

bgallois: > > The worst-case scenario occurs when a valid spend is found and paid.
> 
> But this isn't the worst-case when you have `SpendOrigin = NeverEnsureOrigin<u64>;`. Then the worst-case is basically the best case as well and just means that there is only one read of the storage.

From the extrinsic point of view, it is the worst path because there is no mention of `SpendOrigin` in its implementation. In all the benchmarks that I found in Substrate, the worst path is determined from the extrinsic perspective, rather than considering all the (sometimes complicated) interference of all the parameters that can play a role outside the extrinsic implementation (and that can be quite complicated to anticipate when writing the benchmark, see this error for example).

I think that the implementation you're asking is suboptimal by adding unnecessary complexity to the benchmark for something that doesn't pose any threat to the chain but add a slight overweight. However, you seem very committed to this specific approach, so I will move this pull request to draft status and I will implement it according to your preferences when I have some time.

bkchr: Not sure your arguments are correct here. I mean the problem here is that the **benchmark** is failing because it is written in an incorrect way. The benchmark is "abusing" `SpendOrigin` to setup the state and the benchmark is failing because `SpendOrigin` is set to `NeverEnsureOrigin`. Now if we fix the **benchmark**, we should do it properly that it generates the correct weights. The thing I'm proposing here is also not really that complicated and can be implemented in some minutes.

bgallois: Yes, I agree that the benchmark is failing because it was written incorrectly. In my opinion, this failure directly stems from not approaching it from the extrinsic point of view, as I explained before. 

No, it's not complicated. However, since **I'm paid to work on another Substrate chain, this fix isn't a priority in my work schedule** that is quite full at the moment. Feel free to complete it if you want it done rapidly (i.e. before the end of a week).

Hugo-Trentesaux:
> The thing I'm proposing here is also not really that complicated and can be implemented in some minutes.

@bkchr, those minutes transformed into months and the benchmarks are still broken.

If you are still interested by a fix, we have it on our fork: https://github.com/duniter/duniter-polkadot-sdk/commits/duniter-substrate-v1.14.0/ (latest commit here).

bkchr (approved):

bkchr:
@Hugo-Trentesaux I have done now the changes as I had requested.

ggwpez (approved):

bgallois:
Thank you for the fix.

There is still a small attack vector if a chain is improbably (mis)configured with `SpendOrigin` set to `NeverOrigin`, but `RejectOrigin` can be successful. In this case, the code from `remove_approval` could be executed without any weight at all.

bkchr:
Ty! https://github.com/paritytech/polkadot-sdk/pull/5713

PR #3057, opened 2024-01-25
Serialize/Deserialize trait implemented in no-std for XCM assets

fgamundi:
Continues the work of https://github.com/paritytech/polkadot/pull/7312. `MultiLocation` was modified to implement serialize in no-std in that PR, but not the assets types.
This PR does the same for assets, and this also allows to inject them in pallets that require them in genesis.

### DIFF
--- a/polkadot/xcm/src/v3/multiasset.rs
+++ b/polkadot/xcm/src/v3/multiasset.rs
@@ -51,9 +51,20 @@ use scale_info::TypeInfo;
 
 /// A general identifier for an instance of a non-fungible asset class.
 #[derive(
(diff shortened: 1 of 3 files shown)


girazoki:
@KiChjang @franciscoaguirre we are experimenting with a pallet that would inject `AssetId` in genesis and we need the serialize to be derived. This PR does that and adds it to pretty much all XCM types, is this something that is OK for you?

KiChjang (approved):
Should be ok, I don't see major issues.

bkchr (approved):

PR #3060, opened 2024-01-25
Add retry mechanics to `pallet-scheduler`

### LINKED ISSUE #3014: Scheduler: Retry a Task on a Call Failure
### Problem
The scheduler pallet, utilized for setting and execution delayed tasks/calls, currently abandons the task upon it's call failure. In specific situations (refer to the use case: [PR](https://github.com/paritytech/polkadot-sdk/pull/2939), [place in code](https://github.com/paritytech/polkadot-sdk/blob/e42cfa80c5833d44181e4f69cfe9a4955eab000f/cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/treasury.rs#L382)), there is a requirement to implement tasks retries within a specified time period.

### Possible Solution 
Introduce optional retry setting for a task:
``` rust
#[pallet::storage]
pub type Retries<T: Config> = StorageMap<
    _,
    _,
    TaskAddress<BlockNumberFor<T>>,
    // (total retries, retries left, period)
    (u8, u8, BlockNumberFor<T>),
>;

/// Users assess the possibility of overlapping between a periodic task and retries on their own.
pub fn set_retry(origin Origin,  when: BlockNumber, index: u32, retries: u8, period: BlockNumber);
pub fn set_retry_named(origin Origin, id: TaskName, retries: u8, period: BlockNumber);

fn service_agenda(...) {
  .....
  if call_failed {
     // check if any retries 
     // if any, schedule new task to retry and update/remove the retry record
  } else {
     // check if any retries 
     // if any, remove the retry record and create new if a task is periodic
  }
  .....
}
```

georgepisaltu:
Fixes #3014 

This PR adds retry mechanics to `pallet-scheduler`, as described in the issue above.

Users can now set a retry configuration for a task so that, in case its scheduled run fails, it will be retried after a number of blocks, for a specified number of times or until it succeeds.

If a retried task runs successfully before running out of retries, its remaining retry counter will be reset to the initial value. If a retried task runs out of retries, it will be removed from the schedule.

Tasks which need to be scheduled for a retry are still subject to weight metering and agenda space, same as a regular task. Periodic tasks will have their periodic schedule put on hold while the task is retrying.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5026755

polkadot/runtime/westend/src/weights/pallet_scheduler.rs
```diff
@@ -204,4 +204,28 @@ impl<T: frame_system::Config> pallet_scheduler::WeightInfo for WeightInfo<T> {
...
 	}
+	/// Dummy impl
+	fn check_retry(s: u32, ) -> Weight {
+		Weight::from_parts(2_963_000, 0)
+			.saturating_add(Weight::from_parts(0, 159279))
+			.saturating_add(Weight::from_parts(909_557, 0).saturating_mul(s.into()))
+			.saturating_add(T::DbWeight::get().reads(1))
+			.saturating_add(T::DbWeight::get().writes(1))
+	}
+	/// Dummy impl
```
ggwpez, line 215:
```suggestion
	/// Dummy impl - FAIL-CI
```
then the CI fails and reminds you 😄

georgepisaltu: I plan on running the benches locally and committing the results soon. We will need to run them again on reference hardware before a release anyway but just to make sure everything works.

georgepisaltu: Done ✅

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+		///
+		/// Tasks which need to be scheduled for a retry are still subject to weight metering and
+		/// agenda space, same as a regular task. Periodic tasks will have their periodic schedule
+		/// put on hold while the task is retrying.
+		#[pallet::call_index(6)]
+		#[pallet::weight(<T as Config>::WeightInfo::set_retry(T::MaxScheduledPerBlock::get()))]
+		pub fn set_retry(
+			origin: OriginFor<T>,
+			when: BlockNumberFor<T>,
+			index: u32,
```
ggwpez, line 482: I think there is a convenience type to reference a task: `TaskAddress<BlockNumberFor<T>>`

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+			retries: u8,
+			period: BlockNumberFor<T>,
+		) -> DispatchResult {
+			T::ScheduleOrigin::ensure_origin(origin.clone())?;
+			let origin = <T as Config>::RuntimeOrigin::from(origin);
+			let agenda = Agenda::<T>::get(when);
+			match agenda.get(index as usize) {
```
ggwpez, line 489: You can also do something like this `let task = agenda.get(index as usize).flatten().ok_ok(NotFound)?;`.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+					if matches!(
+						T::OriginPrivilegeCmp::cmp_privilege(origin.caller(), &task.origin),
+						Some(Ordering::Less) | None
+					) {
+						return Err(BadOrigin.into())
+					},
+				_ => return Err(Error::<T>::NotFound.into()),
+			}
+			Retries::<T>::insert(
+				(when, index),
```
ggwpez, line 500: What happens when a retry is already scheduled and the retry config is changed?  
Do you think its important to be able to change the retry config?

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1320,52 @@ impl<T: Config> Pallet<T> {
...
+						Some(task)
+					},
+				}
+			},
+			None => Some(task),
+		}
+	}
 }
 
 impl<T: Config> schedule::v2::Anon<BlockNumberFor<T>, <T as Config>::RuntimeCall, T::PalletsOrigin>
```
ggwpez, line 1371: I think it would be easier to just include an optional `RetryConfig` as part of the `schedule` and `schedule_named` call instead of having another call, or?

mordamax:
bot bench-all pallet -v PIPELINE_SCRIPTS_REF=mak-bench-all-pallet --pallet=pallet_scheduler

substrate/frame/scheduler/src/lib.rs
georgepisaltu: I saw that but I decided to go with the current convention in the code, as seen in [`cancel`](https://github.com/paritytech/polkadot-sdk/blob/73fd8cd717c4df533e3c4fbaa2ee4f06504bf978/substrate/frame/scheduler/src/lib.rs#L357).

substrate/frame/scheduler/src/lib.rs
georgepisaltu: It would need an additional `map` because of the borrow and it would look like:

```
let task = agenda
	.get(index as usize)
	.map(Option::as_ref)
	.flatten()
	.ok_or(Error::<T>::NotFound)?;
```

I think this is a tiny bit uglier, but it might be better because we wouldn't be using the wildcard in the `match`. Let me know what you think.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: > What happens when a retry is already scheduled and the retry config is changed?

The next time the task comes up to be serviced the retry config would be updated. The current implementation allows you to change the config on the fly (e.g. you want to increase the period because something else in the system changed, but you don't want to deregister then register the task again).

> Do you think its important to be able to change the retry config?

I think it's useful but not necessary. That being said, I don't see a reason to forbid it. You pay for each execution either way, and it is simpler than to keep a separate state of the retry config per attempt just to restrict this behavior.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: I think this is a niche mechanic and I wouldn't burden the `schedule` and `schedule_named` extrinsics with this, they already hide enough complexity as it stands. It's simpler to test as well. Also, doing it separately like this also has the nice bonus of being backwards compatible. Do you feel strongly about this?

mordamax:
bot help

substrate/frame/scheduler/src/lib.rs
ggwpez: But then in the event its a `TaskAddress`? I think it should be consistent...

substrate/frame/scheduler/src/lib.rs
ggwpez: This works, but not that big of a difference:  
```rust
let Some(task) = agenda.get(index as usize).and_then(|x| x.as_ref()) else {
	return Err(Error::<T>::NotFound.into())
};
	
if matches!(
	T::OriginPrivilegeCmp::cmp_privilege(origin.caller(), &task.origin),
	Some(Ordering::Less) | None
) {
	return Err(BadOrigin.into())
}

File: substrate/frame/scheduler/src/lib.rs (hunk 5 above)
Comment (ggwpez) [reply substrate/frame/scheduler/src/lib.rs:1371]: Hm okay. Yea its probably not used that often.

File: substrate/frame/scheduler/src/lib.rs (hunk 4 above)
Comment (ggwpez) [reply substrate/frame/scheduler/src/lib.rs:500]: Okay. Its not an issue, just wanted to hear your reasoning for it 😄

File: substrate/frame/scheduler/src/lib.rs (hunk 2 above)
Comment (georgepisaltu) [reply substrate/frame/scheduler/src/lib.rs]: Changed it to use `TaskAddress`.

File: prdoc/pr_3060.prdoc [hunk 6]
```diff
@@ -0,0 +1,16 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Add retry mechanics to `pallet-scheduler`
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Fixes #3014
```
ggwpez, line 9:
```suggestion
```
The prdocs should be self contained so that we can eventually compile them into a CHANGELOG

substrate/frame/scheduler/src/lib.rs
georgepisaltu: I went with the one above.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+			T::ScheduleOrigin::ensure_origin(origin.clone())?;
+			let origin = <T as Config>::RuntimeOrigin::from(origin);
+			let (when, index) = task;
+			let agenda = Agenda::<T>::get(when);
+			match agenda.get(index as usize) {
+				Some(Some(task)) =>
+					if matches!(
+						T::OriginPrivilegeCmp::cmp_privilege(origin.caller(), &task.origin),
+						Some(Ordering::Less) | None
+					) {
```
ggwpez: I see this copy-pasted a few times instead of being an `ensure_privilege` function, but does not need to be fixed in this MR.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+						return Err(BadOrigin.into())
+					},
+				_ => return Err(Error::<T>::NotFound.into()),
+			}
+			Retries::<T>::insert(
+				(when, index),
+				RetryConfig { total_retries: retries, remaining: retries, period },
+			);
+			Ok(())
```
ggwpez, line 504: Maybe this should emit an event since there is otherwise no way for an UI to see that a retry config was added?

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+						Some(Ordering::Less) | None
+					) {
+						return Err(BadOrigin.into())
+					},
+				_ => return Err(Error::<T>::NotFound.into()),
+			}
+			Retries::<T>::insert(
+				(when, agenda_index),
+				RetryConfig { total_retries: retries, remaining: retries, period },
+			);
```
ggwpez, line 544: Also event probably.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +462,86 @@ pub mod pallet {
...
+		/// If a retried task runs successfully before running out of retries, its remaining retry
+		/// counter will be reset to the initial value. If a retried task runs out of retries, it
+		/// will be removed from the schedule.
+		///
+		/// Tasks which need to be scheduled for a retry are still subject to weight metering and
+		/// agenda space, same as a regular task. Periodic tasks will have their periodic schedule
+		/// put on hold while the task is retrying.
+		#[pallet::call_index(7)]
+		#[pallet::weight(<T as Config>::WeightInfo::set_retry_named(T::MaxScheduledPerBlock::get()))]
+		pub fn set_retry_named(
```
ggwpez, line 520: The way to remove a retry config is just to set it to `0`, or?

substrate/frame/scheduler/src/lib.rs
ggwpez, line 1: Line 195 the function maybe also needs to include `WeightInfo::check_retry`, i did not look closely.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1322,52 @@ impl<T: Config> Pallet<T> {
 		let _ = weight.try_consume(call_weight);
 		Ok(result)
 	}
+
+	/// Check if a task has a retry configutation in place and, if so, try to reschedule it.
```
ggwpez, line 1326:
```suggestion
	/// Check if a task has a retry configuration in place and, if so, try to reschedule it.
```

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1322,52 @@ impl<T: Config> Pallet<T> {
...
+	/// If the task was successfully rescheduled for later, this function will return `None`.
+	/// Otherwise, if the task was not rescheduled, either because there was no retry configuration
+	/// in place, there were no more retry attempts left, or the agenda was full, this function
+	/// will return the task so it can be handled somewhere else.
+	fn check_retry(
+		now: BlockNumberFor<T>,
+		when: BlockNumberFor<T>,
+		agenda_index: u32,
+		task: ScheduledOf<T>,
+	) -> Option<ScheduledOf<T>> {
```
ggwpez, line 1337: It does not make a big difference, but semantically it seems that the return type of this function is `Result<(), ScheduledOf<T>>`.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1322,52 @@ impl<T: Config> Pallet<T> {
...
 		Ok(result)
 	}
+
+	/// Check if a task has a retry configutation in place and, if so, try to reschedule it.
+	///
+	/// If the task was successfully rescheduled for later, this function will return `None`.
+	/// Otherwise, if the task was not rescheduled, either because there was no retry configuration
+	/// in place, there were no more retry attempts left, or the agenda was full, this function
+	/// will return the task so it can be handled somewhere else.
+	fn check_retry(
```
ggwpez, line 1332: `schedule_retry`? The function actually *does* something and not just check things.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1322,52 @@ impl<T: Config> Pallet<T> {
...
+						// Reinsert the retry config to the new address of the task after it was
+						// placed.
+						Retries::<T>::insert(
+							address,
+							RetryConfig { total_retries, remaining, period },
+						);
+						None
+					},
+					Err((_, task)) => {
+						// TODO: Leave task in storage somewhere for it to be
```
ggwpez, line 1357: I think its not so nice that this part appears twice...

ggwpez (commented):
Okay looks pretty good. Just some more nits

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5041166

substrate/frame/scheduler/src/lib.rs
georgepisaltu: This is a bit more complicated. `schedule_retry` is called when a task fails and it has a retry config set. We don't know ahead of time if the task will fail, and we could know ahead of time if it has a retry config set, but it's expensive to do the DB read for each task, especially since the retry thing is niche and will probably not be used very often.

Right now the code tries to consume the weight for `schedule_retry` if it needs to run that function, but won't error out if there's not enough weight (which is probably wrong). Alternatively, what we could do is skip the retry logic if the necessary extra weight can't be consumed. Basically the code would run as if there was no retry config set. I think this would be fine if we also send an event (like `RetryFailed` or something).

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Yes. We could add 2 more extrinsics to remove a retry config, or we could make `period: Option<BlockNumberFor<T>>`, and unset it in the case of  `None`. I like it how it is now, but I don't feel strongly about this, WDYT?

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Renamed.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Yes. Initially I thought that it's not really an error if there's no space so I was hesitant about the `Result`, but I changed my mind reading the code again :grin: .

substrate/frame/scheduler/src/lib.rs
georgepisaltu: I agree, but I'd rather do this in another PR with it's own tests. This is pretty big as is.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Moved it to `ensure_privilege`.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: > Alternatively, what we could do is skip the retry logic if the necessary extra weight can't be consumed

I went with this for now, I'll work on adding a UT for it soon.

georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
ggwpez: Im also fine with both 🤷‍♂️

substrate/frame/scheduler/src/lib.rs
```diff
@@ -273,6 +284,15 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	pub type Retries<T: Config> = StorageMap<
```
ggwpez, line 288: Doc please.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -295,10 +315,19 @@ pub mod pallet {
 			id: Option<TaskName>,
 			result: DispatchResult,
 		},
+		/// Set a retry configuration for some task.
+		RetrySet {
+			task: TaskAddress<BlockNumberFor<T>>,
+			id: Option<TaskName>,
+			period: BlockNumberFor<T>,
```
ggwpez, line 322: It does not contain the `total_retries`?

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +469,82 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			id: TaskName,
+			retries: u8,
+			period: BlockNumberFor<T>,
+		) -> DispatchResult {
+			T::ScheduleOrigin::ensure_origin(origin.clone())?;
+			let origin = <T as Config>::RuntimeOrigin::from(origin);
+			let (when, agenda_index) = Lookup::<T>::get(&id).ok_or(Error::<T>::NotFound)?;
+			let agenda = Agenda::<T>::get(when);
+			let scheduled = agenda
```
ggwpez, line 532: Looking at our internal code this should probably be refactored into a `do_set_retry(name: Either<TaskName, TaskAddress>)` but i dont want to block this any further.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1311,92 @@ impl<T: Config> Pallet<T> {
...
+	/// the agenda was full, this function will return the task so it can be handled somewhere else.
+	fn try_schedule_retry(
+		weight: &mut WeightMeter,
+		now: BlockNumberFor<T>,
+		when: BlockNumberFor<T>,
+		agenda_index: u32,
+		task: ScheduledOf<T>,
+	) -> Result<(), ScheduledOf<T>> {
+		if let Err(()) =
+			weight.try_consume(T::WeightInfo::schedule_retry(T::MaxScheduledPerBlock::get()))
```
ggwpez, line 1329:
```suggestion
		if weight.try_consume(T::WeightInfo::schedule_retry(T::MaxScheduledPerBlock::get())).is_err()
```

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1311,92 @@ impl<T: Config> Pallet<T> {
...
+	fn ensure_privilege(
+		left: &<T as Config>::PalletsOrigin,
+		right: &<T as Config>::PalletsOrigin,
+	) -> Result<(), DispatchError> {
+		if matches!(T::OriginPrivilegeCmp::cmp_privilege(left, right), Some(Ordering::Less) | None)
+		{
+			return Err(BadOrigin.into());
+		}
+		Ok(())
+	}
```
ggwpez, line 1399: Is the `Retries` storage cleaned up when a task is cancelled?

ggwpez (approved):
LGTM, just some comments.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -273,6 +284,15 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	pub type Retries<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
```
georgepisaltu, line 290: Looking at it again, I think it's safe to have this be `Twox64Concat`. @ggwpez WDYT

substrate/frame/scheduler/src/lib.rs
georgepisaltu: I didn't include it because I tried to keep the event lean and I thought the period is the most important aspect of it.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: It wasn't, this just got lost in the impl, good catch! I also added a test for it.

substrate/frame/scheduler/src/lib.rs
ggwpez: Yea am also often not sure what all to include in events... but i think in this case its fine since it should only rarely be used.

substrate/frame/scheduler/src/lib.rs
ggwpez: Its only accessed through permissioned origins - so it is trusted already. In this case it should also not easily disbalance the trie so should be be fine to use twoX. But TBH the advantage is negligible as trade-off against a more balanced trie.

georgepisaltu: I'll just leave it as is then.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Added the retry count.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1137,18 +1248,31 @@ impl<T: Config> Pallet<T> {
 					}
 					let wake = now.saturating_add(period);
 					match Self::place_task(wake, task) {
-						Ok(_) => {},
+						Ok(new_address) => {
+							if let Some(RetryConfig { total_retries, remaining, period }) =
+								Retries::<T>::take((when, agenda_index))
```
muharem, line 1253: if current iteration for periodic failed and retry scheduled, is not this always `none`?

In description you mention that next periodic is placed on hold. What about just letting a user to decide? If a user does not want them overlap, a user controls it with a number of `retries`  and `period`. A user might not want their periodic tasks to be placed on hold, because of retry, to me retry has a lower priority. 
Also after all retries attempts, it can be too late to schedule next periodic iteration, I think in this case we loose the periodic task. 
Without the hold, the logic looks more straightforward to me. As I mentioned in the issue, I think users should assess the possibility of overlapping between a periodic task and retries on their own.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Will do.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: > if current iteration for periodic failed and retry scheduled, is not this always none?

If that happens, the function exits early here. But there is this [corner case](https://github.com/paritytech/polkadot-sdk/pull/3060/files#diff-88e61363c8fccbe1b7de6ee1c6dbc2afdb4e74e2aebab1f151953be8e2ec87ceR1326) where there isn't enough weight to do the retry logic, and in that case retries are ignored, test [here](https://github.com/paritytech/polkadot-sdk/pull/3060/files#diff-0d54d132118a2b638382f5259175174c3bbcc6919259fc5bc4262e7e1fff67efR1076).

georgepisaltu: Regarding the rest of your comment, I can see why you'd want the period and retry to work independently. I'm ok to make it do this:
- if a task fails, try to schedule a retry if a retry config is set
- if the task is periodic, try to schedule the next iteration

I think if we make this behavior configurable (i.e. let users decide which behaviour they want), it would get way too complicated and it's not worth it. I think it's better if it's simpler.

> I think users should assess the possibility of overlapping between a periodic task and retries on their own

I was trying to avoid this because I saw it as a footgun, but I think you're right. If users set the configurations correctly, it shouldn't be an issue.

georgepisaltu: Proposed changes in https://github.com/paritytech/polkadot-sdk/pull/3060/commits/9da58c6eb6eda067ce673f46485d2c3738293d7c

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +489,99 @@ pub mod pallet {
...
+		/// counter will be reset to the initial value. If a retried task runs out of retries, it
+		/// will be removed from the schedule.
+		///
+		/// Tasks which need to be scheduled for a retry are still subject to weight metering and
+		/// agenda space, same as a regular task. If a periodic task fails, it will be scheduled
+		/// normally while the task is retrying.
+		///
+		/// Tasks scheduled as a result of a named retry will have their own retry configuration
+		/// derived from the original task's configuration, but will have a lower value for
+		/// `remaining` than the original `total_retries`. Tasks scheduled as a result of a retry of
```
muharem, line 550: can you explain why you think it should be this way?
the way I think about it, is I want some call to be executed 5 times, and when any of it fail, I wanna retry it 3 times. I wanna be sure that every call will be given chance to be retried 3 times. If first exhausts all retries for some unexpected reason, I do not wanna leave the rest calls without retries. It is hard for me to model such situation.

muharem:
I do not know your motivation for a RP from you fork, but it makes it harder for me to pool you branch and review from my local machine.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1124,11 +1258,17 @@ impl<T: Config> Pallet<T> {
 			},
 			Err(()) => Err((Overweight, Some(task))),
```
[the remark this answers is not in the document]
georgepisaltu: > in this and the Err arm above you abandon the retry entry if there is any

The `Err` arms above deal with completely different problems:
- `Err(Unavailable)` is a task which may never be able to run and it may never even be attempted again
- `Err(Overweight)` is a task which couldn't run because the scheduler ran out of weight, but it will be run the next time the scheduler services agendas; the scheduler runs agendas from previous blocks if they didn't execute completely, in order of block number from earliest up until present

This means that an overweight task is not abandoned and will be run again soon, but also that there is absolutely nothing that we can do about it. There is no fallback for the current block if we run out of weight. This is the scheduler's behavior before this PR and I believe the retry mechanism has nothing to do with this.

> I would move the retry handling one level above, for this you just need to add for Ok varian an info if it failed or not.

Most of the logic is common for failed or successful tasks, there would be a lot of copied code. But if you just want to refactor this, I could move the common logic out of the `match`.

> or at least take it before `execute_dispatch`

Tasks are scheduled for retries if they actually run but fail in their execution. By definition, there is no way to handle this before running `execute_dispatch`.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1124,11 +1258,17 @@ impl<T: Config> Pallet<T> {
...
 			Ok(result) => {
+				let failed = result.is_err();
 				Self::deposit_event(Event::Dispatched {
 					task: (when, agenda_index),
 					id: task.maybe_id,
 					result,
 				});
+
+				if failed {
+					Self::try_schedule_retry(weight, now, when, agenda_index, &task);
```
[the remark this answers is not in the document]
georgepisaltu: Will do.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1137,18 +1277,31 @@ impl<T: Config> Pallet<T> {
 					}
 					let wake = now.saturating_add(period);
 					match Self::place_task(wake, task) {
-						Ok(_) => {},
+						Ok(new_address) => {
+							if let Some(RetryConfig { total_retries, remaining, period }) =
+								Retries::<T>::take((when, agenda_index))
+							{
+								// Reset the counter if the task ran successfully.
+								let remaining = if failed { remaining } else { total_retries };
```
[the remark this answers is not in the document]
georgepisaltu: Doc comment is outdated, I'll fix it. This is not needed anymore since the impl changed to retry clones. It had no effect on the current impl because for original tasks, `remaining == total_retries` anyway and retry clones don't get to this part of the code because they are not "periodic".

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1137,18 +1304,29 @@ impl<T: Config> Pallet<T> {
...
 					let wake = now.saturating_add(period);
 					match Self::place_task(wake, task) {
-						Ok(_) => {},
+						Ok(new_address) => {
+							if let Some(RetryConfig { total_retries, remaining, period }) =
+								Retries::<T>::take((when, agenda_index))
+							{
+								Retries::<T>::insert(
+									new_address,
+									RetryConfig { total_retries, remaining, period },
```
muharem, line 1313:
```suggestion
									RetryConfig { total_retries, total_retries, period },
```
it would be more clear what's happening

substrate/frame/scheduler/src/lib.rs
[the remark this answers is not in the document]
muharem: I thought that the task is removed if `Unavailable`, but now I see that we keep it in the storage. This is why I proposed to remove the retry as well. And also I see now that `Overweight` does not change it's address.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1124,11 +1285,17 @@ impl<T: Config> Pallet<T> {
 			},
 			Err(()) => Err((Overweight, Some(task))),
 			Ok(result) => {
+				let failed = result.is_err();
```
muharem, line 1288: But still if you `take` retry here, it would be easier to maintain, if new branches appears below. Now you need to remove it in different branches and make an additional insert in `schedule_retry`. I would just pass the reference to `retry` from this context for `schedule_retry`.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -1192,6 +1370,81 @@ impl<T: Config> Pallet<T> {
...
+		if weight
+			.try_consume(T::WeightInfo::schedule_retry(T::MaxScheduledPerBlock::get()))
+			.is_err()
+		{
+			Self::deposit_event(Event::RetryFailed {
+				task: (when, agenda_index),
+				id: task.maybe_id,
+			});
+			return;
+		}
```
muharem, line 1397: I meant, that it would be better to have this in `service_task` context. Now the `schedule_retry` is failable/no-op, which is not clear from `service_task` context.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -148,6 +159,24 @@ pub struct Scheduled<Name, Call, BlockNumber, PalletsOrigin, AccountId> {
 	_phantom: PhantomData<AccountId>,
 }
 
+impl<Name, Call, BlockNumber, PalletsOrigin, AccountId>
+	Scheduled<Name, Call, BlockNumber, PalletsOrigin, AccountId>
+where
+	Call: Clone,
+	PalletsOrigin: Clone,
+{
+	pub fn as_retry(&self) -> Self {
```
muharem, line 168: please provide some doc

substrate/frame/scheduler/src/lib.rs
```diff
@@ -148,6 +159,24 @@ pub struct Scheduled<Name, Call, BlockNumber, PalletsOrigin, AccountId> {
...
+		Self {
+			maybe_id: None,
+			priority: self.priority,
+			call: self.call.clone(),
+			maybe_periodic: None,
```
muharem, line 173: now I wonder what if retry will be scheduled as periodic task, would it simplify the solution

substrate/frame/scheduler/src/lib.rs
georgepisaltu: So, you mean if the task fails, check for a retry config and if necessary give that to `schedule_retry`? That sounds good.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: The downside is that you can't know in `service_task` ahead of time if the task will fail and that it has a retry config. You could check that you can consume the weight anyway, but you might be preventing users from running their tasks when it wasn't necessary.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Will do, forgot about this function.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: You would still need a way to tell if the task is a retry or a regular task, because a retry task needs to be dropped after it runs once successfully. I think the current solution is good enough for now after addressing all comments.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done. PTAL

substrate/frame/scheduler/src/lib.rs
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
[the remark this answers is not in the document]
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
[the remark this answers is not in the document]
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -854,6 +991,7 @@ impl<T: Config> Pallet<T> {
 			if let Some(id) = s.maybe_id {
 				Lookup::<T>::remove(id);
 			}
+			let _entry = Retries::<T>::take((when, index));
```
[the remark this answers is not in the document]
georgepisaltu: Done.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -481,8 +499,15 @@ pub mod pallet {
...
-		/// put on hold while the task is retrying.
+		/// agenda space, same as a regular task. If a periodic task fails, it will be scheduled
+		/// normally while the task is retrying.
+		///
+		/// Tasks scheduled as a result of a retry will have their own retry configuration derived
+		/// from the original task's configuration, but will have a lower value for `remaining` than
+		/// the original `total_retries`.
+		///
+		/// To cancel a retry configuration for a task, users should call this extrinsic with a
+		/// `retries` value equal to `0`.
```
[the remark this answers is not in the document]
georgepisaltu: Done.

NachoPal (approved):

NachoPal (approved):
LGTM. Just wondering if you considered extending `Scheduled` struct with `RetryConfig` fields instead of creating a new `Retries` storage item. In that way it would natively support retries when `schedule()` without the need of doing it in two steps (`schedule()` + `set_retry()`)

substrate/frame/scheduler/src/lib.rs
muharem: yeah, I thought about it over night, it does not make sense to make it as periodic

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +493,124 @@ pub mod pallet {
...
+		/// Set a retry configuration for a task so that, in case its scheduled run fails, it will
+		/// be retried after `period` blocks, for a total amount of `retries` retries or until it
+		/// succeeds.
+		///
+		/// Tasks which need to be scheduled for a retry are still subject to weight metering and
+		/// agenda space, same as a regular task. If a periodic task fails, it will be scheduled
+		/// normally while the task is retrying.
+		///
+		/// Tasks scheduled as a result of a retry for a periodic task are unnamed,
+		/// non-periodic clones of the original task and will have their own retry configuration
```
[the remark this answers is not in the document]
georgepisaltu: Rephrased.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +493,124 @@ pub mod pallet {
...
+		/// Set a retry configuration for a named task so that, in case its scheduled run fails, it
+		/// will be retried after `period` blocks, for a total amount of `retries` retries or until
+		/// it succeeds.
+		///
+		/// Tasks which need to be scheduled for a retry are still subject to weight metering and
+		/// agenda space, same as a regular task. If a periodic task fails, it will be scheduled
+		/// normally while the task is retrying.
+		///
+		/// Tasks scheduled as a result of a named retry for a periodic task are unnamed,
+		/// non-periodic clones of the original task and will have their own retry configuration
```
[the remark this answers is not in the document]
georgepisaltu: Rephrased.

substrate/frame/scheduler/src/lib.rs
```diff
@@ -440,6 +493,124 @@ pub mod pallet {
...
+			let (when, agenda_index) = Lookup::<T>::get(&id).ok_or(Error::<T>::NotFound)?;
+			let agenda = Agenda::<T>::get(when);
+			let scheduled = agenda
+				.get(agenda_index as usize)
+				.and_then(Option::as_ref)
+				.ok_or(Error::<T>::NotFound)?;
+			Self::ensure_privilege(origin.caller(), &scheduled.origin)?;
+			Retries::<T>::remove((when, agenda_index));
+			Self::deposit_event(Event::RetryCancelled { task: (when, agenda_index), id: Some(id) });
+			Ok(())
```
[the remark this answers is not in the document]
georgepisaltu: Done.

georgepisaltu:
> LGTM. Just wondering if you considered extending `Scheduled` struct with `RetryConfig` fields instead of creating a new `Retries` storage item. In that way it would natively support retries when `schedule()` without the need of doing it in two steps (`schedule()` + `set_retry()`)

I considered it but I decided against it. This is a niche mechanic which we don't expect to be widely used right now. Doing it in `Scheduled` would bloat existing `Scheduled` entries. It's not backwards compatible in terms of storage (would require a migration) or API. It makes more sense for the few users of this mechanic to pay extra than for everyone to pay more for something that is not a core feature. If, in the future, we find that a lot of people are doing retries, we can reconsider in order to optimize costs.

PR #3062, opened 2024-01-25
test-runtime: json_patch dep removed

michalkucharczyk:
We don't need dependency on `json_patch` there is implementation of needed function in `sc_chain_spec` crate.
This PR makes this function publicly available and removes the dep.

### DIFF
--- a/substrate/client/chain-spec/src/lib.rs
+++ b/substrate/client/chain-spec/src/lib.rs
@@ -338,6 +338,7 @@ pub use self::{
 		GenesisBlockBuilder,
 	},
 	genesis_config_builder::GenesisConfigBuilderRuntimeCaller,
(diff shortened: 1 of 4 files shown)


substrate/client/chain-spec/src/lib.rs
```diff
@@ -338,6 +338,7 @@ pub use self::{
 		GenesisBlockBuilder,
 	},
 	genesis_config_builder::GenesisConfigBuilderRuntimeCaller,
+	json_patch::merge as json_merge,
```
michalkucharczyk, line 341: Maybe not most elegant solution, but I don't want to add crate for this single tiny `merge` function. Let me know if it is ok.

bkchr (approved):

davxy (approved):

Issue #3058, opened 2024-01-25
Kusama parachains not producing blocks for a session


alexggh:
## Issue
After v1,001,000 upgrade https://kusama.subsquare.io/referenda/328, kusama parachains blocks were very rarely included in the relay chain until the next session.

## Investigation

v1,001,000 runtime upgrade brought with the async_backing_params API: https://github.com/polkadot-fellows/runtimes/blob/e115160682a87798cb6557db9c73e77d00ddb0ab/relay/kusama/src/lib.rs#L1973, that caused several new subsystems(prospective-parachains, statement-distribution-v2) to activate mid-session and that cause backing of candidate to slow down significantly see https://grafana.teleport.parity.io/goto/6jZXXmtSg?orgId=1

<img width="1271" alt="Screenshot 2024-01-25 at 12 46 11" src="https://github.com/paritytech/polkadot-sdk/assets/49718502/47cc5f74-cda3-4423-8fa5-2679703fed77">

## Reproducibility

The issue can be reproduce easily with zombienet if we hacked the runtime-api subsystem to fake the presence of `async_backign_params` at block after the session start, something like this:
```
-               Request::AsyncBackingParams(sender) => {
-                       query!(
-                               AsyncBackingParams,
-                               async_backing_params(),
-                               ver = Request::ASYNC_BACKING_STATE_RUNTIME_REQUIREMENT,
-                               sender
-                       )
+               Request::AsyncBackingParams(block_number, sender) => {
+                       let sender = sender;
+                       let version: u32 = (Request::ASYNC_BACKING_STATE_RUNTIME_REQUIREMENT);
+                       let runtime_version = client
+                               .api_version_parachain_host(relay_parent)
+                               .await
+                               .unwrap_or_else(|e| {
+                                       gum::warn!(target:LOG_TARGET,"cannot query the runtime API version: {}",e,);
+                                       Some(0)
+                               })
+                               .unwrap_or_else(|| {
+                                       gum::warn!(target:LOG_TARGET,"no runtime version is reported");
+                                       0
+                               });
+                       let res = if runtime_version >= version && block_number > 40 {
+                               client.async_backing_params(relay_parent).await.map_err(|e| {
+                                       RuntimeApiError::Execution {
+                                               runtime_api_name: stringify!(async_backing_params),
+                                               source: std::sync::Arc::new(e),
+                                       }
+                               })
+                       } else {
+                               Err(RuntimeApiError::NotSupported {
[...]

PR #3063, opened 2024-01-25
statement-distribution: fix parachains stalling on async_backing enablement

alexggh:
Topology is coming only at the beginning of each session, so we might lose it if prospective parachains was not enabled at the begining of the session, so cache it for later use.

Fixes: https://github.com/paritytech/polkadot-sdk/issues/3058

### DIFF
--- a/polkadot/node/network/statement-distribution/src/v2/mod.rs
+++ b/polkadot/node/network/statement-distribution/src/v2/mod.rs
@@ -36,8 +36,9 @@ use polkadot_node_primitives::{
 };
 use polkadot_node_subsystem::{
 	messages::{
-		CandidateBackingMessage, HypotheticalCandidate, HypotheticalFrontierRequest,
-		NetworkBridgeEvent, NetworkBridgeTxMessage, ProspectiveParachainsMessage,
+		network_bridge_event::NewGossipTopology, CandidateBackingMessage, HypotheticalCandidate,
+		HypotheticalFrontierRequest, NetworkBridgeEvent, NetworkBridgeTxMessage,
+		ProspectiveParachainsMessage,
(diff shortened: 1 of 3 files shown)


ordian (approved):
Nice catch! Worth adding a test?

sandreim (approved):

alexggh:
> Nice catch! Worth adding a test?

It is always worth adding a test :D, but I will do it `async` I don't think it makes sense to wait with this PR until I figure it out how to do that.

Note! This is tested with my hacked setup described here: https://github.com/paritytech/polkadot-sdk/issues/3058, but obviously I can't commit that :D.

rphmeier:
https://github.com/paritytech/polkadot-sdk/issues/761

Further discussion here, along with a more elegant proposed fix which pull the topology rather than having it pushed.

alexggh:
> #761
> 
> Further discussion here, along with a more elegant proposed fix which pull the topology rather than having it pushed.

Will have a look, but I guess that is further down the line, so I will go ahead and merge this fix since it is better to have it present rather than not.

> Just an idea, I imagine a unit test where we drive the overseer messages to do something similar with runtime api requests as you did in your setup. Then we'd just want to see that messages are sent according to new topology to start with.

Turns out it was easy to change the current tests to cover both cases.

eskimor (approved):

rphmeier (approved):

rphmeier:
> Will have a look, but I guess that is further down the line, so I will go ahead and merge this fix since it is better to have it present rather than not

Agreed! The fix is definitely necessary now but a new API for pulling topologies would prevent this footgun when building network protocol upgrades in the future.

PR #3065, opened 2024-01-25
bump reed-solomon-novelpoly version

alindima:
also remove some dead code and deduplicate some error handling

the new release brings performance improvements and support for systematic chunk recovery, needed in: https://github.com/paritytech/polkadot-sdk/pull/1644


ordian (approved):

prdoc/pr_3065.prdoc
```diff
@@ -0,0 +1,14 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
```
sandreim, line 1: I don't think we need a prdoc for bumping the version.

alexggh (approved):

sandreim (approved):

prdoc/pr_3065.prdoc
alindima: yeah, but is there a way to make the CI pass without it?

AndreiEres: Use `ro-silent` label

lexnv (approved):

PR #3066, opened 2024-01-25
 Enable asset Owner to mint tokens even if Asset (or account) is frozen

decentration:
This PR introduces changes to the `do_mint` function within the Assets pallet. 

The primary objective is to allow the owner of an asset to mint new tokens even when the asset (or specific accounts holding the asset), are frozen. 

Currently, when the **asset** is **frozen**:

- No one with any level of privilege can mint; 
- Or if an account is frozen no one with any level of privilege can mint to the account. 

This PR enables only the Owner of the asset, to be able to mint. A normal account, or issuer, admin or freezer, still would not be able to mint. 

### Why?
This is essential for assets that wish to mint tokens (such as a snapshot distribution) in its entirety, but keep it frozen, so that checks and verifications can be made before any account can move funds. This is also necessary for legal purposes, clearly delineating when an asset is _not live_ and _live_. 

### Real Use Case 
This issue showed up during the DED token distribution. 1.2m DED accounts needed to have asset minted, and had to be frozen. In order to fulfil this requirement, this costed extra fees by having to mint and freeze every account individually, and thereafter having to thaw every account individually, thereafter. 

## Changes Made:
**Modified `do_mint` Function only:** 

Updated the function to include checks that determine if the asset or the account is frozen. If so, the function now verifies whether the caller is the `owner` of the asset. If the caller is the `owner`, minting proceeds even if the asset/account is frozen._("owner of the asset" should not be confused with "owner of the account")._

**Added Tests:** 

Added new tests to ensure:

- Minting a token by the asset owner to a **frozen account**, (using freeze() is successful.
- Making sure that tokens minted by an address that is not the owner of the asset, to a **frozen account**, fails.

- Minting a token by the asset owner to **an entirely frozen asset** (using freeze_asset()), succeeds.
- Making sure that tokens minted to an address when the asset is entirely frozen, fails. 

These changes to `do_mint` should add no extra reads or writes to the pallet call.

### DIFF
--- a/substrate/frame/assets/src/functions.rs
+++ b/substrate/frame/assets/src/functions.rs
@@ -410,19 +410,37 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		amount: T::Balance,
 		maybe_check_issuer: Option<T::AccountId>,
 	) -> DispatchResult {
-		Self::increase_balance(id.clone(), beneficiary, amount, |details| -> DispatchResult {
+
+		// Retrieve asset details to check if the asset is frozen
+		let asset_details = Asset::<T, I>::get(&id).ok_or(Error::<T, I>::Unknown)?;
+
+		// Check if the asset is frozen
+		let is_asset_frozen = asset_details.status == AssetStatus::Frozen;
+	
+		// Check if the beneficiary account is frozen
+		let is_account_frozen = Account::<T, I>::get(&id, beneficiary)
+			.map_or(false, |account| account.status == AccountStatus::Frozen);
+	
+		// If the asset or account is frozen, ensure the caller is the owner
+		if is_asset_frozen || is_account_frozen {
+			ensure!(maybe_check_issuer == Some(asset_details.owner), Error::<T, I>::NoPermission);
+		} else {
+			// Perform the regular issuer check for non-frozen assets/accounts
 			if let Some(check_issuer) = maybe_check_issuer {
-				ensure!(check_issuer == details.issuer, Error::<T, I>::NoPermission);
+				ensure!(check_issuer == asset_details.issuer, Error::<T, I>::NoPermission);
 			}
+		}
+	
+		// Proceed with increasing the balance and updating the supply
+		Self::increase_balance(id.clone(), beneficiary, amount, |details| -> DispatchResult {
 			debug_assert!(details.supply.checked_add(&amount).is_some(), "checked in prep; qed");
-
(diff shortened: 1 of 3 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5030020

Captain1653:
Hi, I am reviewing old pull requests and came across this one. 

* **Violation of Core Invariants**:  Allowing minting on a frozen asset fundamentally contradicts this intended design. Please, see [this commit from Nov 15, 2022](https://github.com/paritytech/polkadot-sdk/commit/679d2dcd252a4b097e6a649aec34a44d1e9d3f24#diff-1e9f9fedeee1bdf9064888c5dfb715d7efe303c1fc839663640013a11f82f302R393). There was this check in the method `pub(super) fn increase_balance(`:
```rust
ensure!(details.status == AssetStatus::Live, Error::<T, I>::AssetNotLive);
```

* **Broader Architectural Impact**: Changing balance mutation rules cannot be done through a localized fix. Any adjustment of this magnitude requires an RFC and a comprehensive review of all related operations to ensure no edge cases or critical scenarios are missed.

* **Incomplete Execution Path**: The proposed changes only modify the upper-level business logic inside `do_mint`. The execution still delegates down to `increase_balance`, where the this AssetStatus check remains untouched.

Please, see this snippet on your branch:
```rust
pub(super) fn increase_balance(
		id: T::AssetId,
		beneficiary: &T::AccountId,
		amount: T::Balance,
		check: impl FnOnce(
			&mut AssetDetails<T::Balance, T::AccountId, DepositBalanceOf<T, I>>,
		) -> DispatchResult,
	) -> DispatchResult {
		if amount.is_zero() {
			return Ok(());
		}

		Self::can_increase(id.clone(), beneficiary, amount, true).into_result()?;
		Asset::<T, I>::try_mutate(&id, |maybe_details| -> DispatchResult {
			let details = maybe_details.as_mut().ok_or(Error::<T, I>::Unknown)?;
			ensure!(details.status == AssetStatus::Live, Error::<T, I>::AssetNotLive); // THIS CHECK REMAINS UNTOCHED
			check(details)?;
```

* **Root Cause of Test Failures**: This unhandled deep check is the  reason why the PR's own test `transferring_frozen_asset_should_not_work` fails. 

Given these factors, closing this PR is probably the most appropriate path forward for now.

Issue #3003, opened 2024-01-19
BEEFY can get stuck when a majority of validators is restarted


serban300:
If a majority of the validators are restarted while voting for a mandatory block and before a finality proof was created, it can happen for BEEFY to get stuck at that block since:
- there's a big chance that not all in-flight votes will reach the remaining online validators
- the validators memorize that they voted for block #x, so they won't vote again

Even though this shouldn't happen in production, we should address it since it can affect test nets

serban300:
Another issue is that in this situation, after restart some validators skipped mandatory blocks. We should investigate this too.

serban300:
Managed to reproduce the problem locally with a zombienet. Found 2 issues so far:

1. Finality notifications can be missed. For example:

```
[...]

Issue #2842, opened 2024-01-03
BEEFY validators skip finality proof when they are all closed at the same time


serban300:
The root cause of https://github.com/paritytech/polkadot-sdk/issues/2285 was that no validator had a BEEFY finality proof for block #7829924 . They just skipped the justification for #7829924 and moved on. More details in this comment: https://github.com/paritytech/polkadot-sdk/issues/2285#issuecomment-1815887673

Stumbled upon this issue also while working on https://github.com/paritytech/polkadot-sdk/issues/2787

Prerequisites:
1. A zombienet config:
```
[relaychain]
default_image = "docker.io/parity/polkadot:latest"
default_command = "/home/serban/workplace/sources/polkadot-sdk/target/release/polkadot"

# refer to ./README.md for more details on how to create snapshot and spec
chain = "rococo-local"
chain_spec_path = "/home/serban/workplace/snapshots/db-test-gen/rococo-local.json"

  [[relaychain.nodes]]
  name = "alice"
  validator = true

  [[relaychain.nodes]]
  name = "bob"
  validator = true
```
2. Adjust rococo-local epoch duration from:
```
	frame_support::parameter_types! {
		pub EpochDurationInBlocks: BlockNumber =
			prod_or_fast!(1 * HOURS, 1 * MINUTES, "ROCOCO_EPOCH_DURATION");
	}
```
to
```
	frame_support::parameter_types! {
		pub EpochDurationInBlocks: BlockNumber =
			prod_or_fast!(1 * HOURS, 1 * MINUTES / 6, "ROCOCO_EPOCH_DURATION");
	}
``` 

Reproduction steps: 
1. Start the zombienet. Let it run for ~60s.
2. Close the zombienet. I think it needs to be closed right when a finality proof is generated, but it can also be close randomly.
3. Start the zombienet again and also start a 3rd node separately:
```
./target/release/polkadot --chain=/home/serban/workplace/snapshots/db-test-gen/rococo-local.json --dave --tmp --enable-offchain-indexing=true --sync warp --reserved-nodes "/ip4/0.0.0.0/tcp/30310/p2p/12D3KooWQCkBm1BYtkHpocxCwMgR8yjitEeHGx8spzcDLGt2gkBm"
```
4. Check the logs of the 3rd node. See if it prints something like `🥩 ran out of peers to request justif #43 from`. If not, repeat steps 2-3

[...]

PR #3074, opened 2024-01-26
[BEEFY] Avoid missing voting sessions during node restart

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/3003 and https://github.com/paritytech/polkadot-sdk/issues/2842


substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -327,50 +329,172 @@ pub(crate) struct BeefyComms<B: Block> {
...
+			while *header.number() > state.best_beefy() {
+				if let Some(active) = find_authorities_change::<B>(&header) {
+					new_sessions.push((active, *header.number()));
+				}
+				header =
+					wait_for_parent_header(self.backend.blockchain(), header, HEADER_SYNC_DELAY)
+						.await?;
+			}
+
+			// Make sure we didn't miss any sessions during node restart.
```
svyatonik, line 484: DQ: that is the key change in that PR, right? :)

substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -327,50 +329,172 @@ pub(crate) struct BeefyComms<B: Block> {
...
+			for (validator_set, new_session_start) in new_sessions.drain(..).rev() {
+				info!(
+					target: LOG_TARGET,
+					"🥩 Handling missed BEEFY session after node restart: {:?}.",
+					new_session_start
+				);
+				self.init_session_at(&mut state, validator_set, new_session_start);
+			}
+			return Ok(state)
```
svyatonik, line 493: In previous function version, there was the `write_voter_state` call here, before `Ok(state)` is returned. Is that removed on purpose now?

substrate/client/consensus/beefy/src/worker.rs
serban300: Yes, exactly

substrate/client/consensus/beefy/src/worker.rs
serban300: Hmmm I don't see it. Do you have a link please ?

svyatonik: Nvm, was looking at different code piece :) sorry for the noise

svyatonik (approved):
Have found the only logic change in this PR and it looks ok to me

substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -236,12 +240,10 @@ impl<B: Block> VoterOracle<B> {
 	/// Return `Some(number)` if we should be voting on block `number`,
 	/// return `None` if there is no block we should vote on.
 	pub fn voting_target(&self) -> Option<NumberFor<B>> {
-		let rounds = if let Some(r) = self.sessions.front() {
-			r
-		} else {
+		let rounds = self.sessions.front().or_else(|| {
 			debug!(target: LOG_TARGET, "🥩 No voting round started");
 			return None
```
acatangiu, line 245: nit:
```suggestion
			None
```

substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -428,17 +553,72 @@ where
...
 		info!(
 			target: LOG_TARGET,
 			"🥩 New Rounds for validator set id: {:?} with session_start {:?}",
 			id,
 			new_session_start
 		);
 	}
+}
+
+/// A BEEFY worker plays the BEEFY protocol
```
acatangiu, line 569: nit:

```suggestion
/// A BEEFY worker/voter that follows the BEEFY protocol
```

substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -327,50 +329,172 @@ pub(crate) struct BeefyComms<B: Block> {
...
+
+			// Make sure we didn't miss any sessions during node restart.
+			for (validator_set, new_session_start) in new_sessions.drain(..).rev() {
+				info!(
+					target: LOG_TARGET,
+					"🥩 Handling missed BEEFY session after node restart: {:?}.",
+					new_session_start
+				);
+				self.init_session_at(&mut state, validator_set, new_session_start);
+			}
```
acatangiu, line 492: you need to check against accidental double queuing of some session(s).

imagine during normal operation consensus lags, voter starts queuing sessions as they are grandpa finalized (these are also persisted to aux db), then voter is restarted.

This walk back will double queue sessions already in queue coming from aux db.

substrate/client/consensus/beefy/src/worker.rs
```diff
@@ -327,50 +329,172 @@ pub(crate) struct BeefyComms<B: Block> {
...
+		let beefy_genesis = self
+			.runtime
+			.runtime_api()
+			.beefy_genesis(best_grandpa.hash())
+			.ok()
+			.flatten()
+			.filter(|genesis| {
+				// test
+				*genesis == beefy_genesis
+			})
```
acatangiu, line 373:
```suggestion
			.filter(|genesis| *genesis == beefy_genesis)
```

acatangiu (changes requested):

substrate/client/consensus/beefy/src/worker.rs
serban300: Good catch ! Done. Added a `can_add_session()` method. For the moment I just called it specifically for this new use cases.  But in the future we might want to use it in `init_session_at()` maybe for extra safety.

acatangiu: what about an integration/regression test?
I think it could be easy to load a custom aux db persisted state where you deliberately set it up with one or two missing sessions and verify initialization also picks up those.

acatangiu (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5091405

substrate/client/consensus/beefy/src/worker.rs
serban300: You're right. It's a good idea. Done. PTAL.

substrate/client/consensus/beefy/src/tests.rs
```diff
@@ -1285,6 +1305,104 @@ async fn should_initialize_voter_at_custom_genesis_when_state_unavailable() {
...
 
+#[tokio::test]
+async fn should_catch_up_when_loading_saved_voter_state() {
+	let keys = &[BeefyKeyring::Alice];
+	let validator_set = ValidatorSet::new(make_beefy_ids(keys), 0).unwrap();
+	let mut net = BeefyTestNet::new(1);
+	let backend = net.peer(0).client().as_backend();
+
+	// push 15 blocks with `AuthorityChange` digests every 10 blocks
+	let hashes = net.generate_blocks_and_sync(30, 10, &validator_set, false).await;
```
acatangiu, line 1316: nit:
```suggestion
	// push 30 blocks with `AuthorityChange` digests every 10 blocks
	let hashes = net.generate_blocks_and_sync(30, 10, &validator_set, false).await;
```

PR #3075, opened 2024-01-26
Fix delivery fees issue in Kusama

franciscoaguirre:
This fix aims to solve an issue in Kusama that resulted in failed reserve asset transfers.
It will be ported later to master as well.
The issue is that during multi-hop XCMs, like reserve asset transfers where the reserve is not the sender nor the destination, there are no assets available to pay for delivery fees.
This fix makes sure to deduct delivery fees from the assets being transferred.


polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -633,10 +661,26 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 				let mut message = vec![ReserveAssetDeposited(assets), ClearOrigin];
 				message.extend(xcm.0.into_iter());
+
+				// put back parked_fee in holding register to be charged by XcmSender
+				self.holding.subsume_assets(parked_fee);
 				self.send(dest, Xcm(message), FeeReason::DepositReserveAsset)?;
 				Ok(())
 			},
 			InitiateReserveWithdraw { assets, reserve, xcm } => {
+				// we need to do this take/put cycle to solve wildcards and get exact assets to be
```
KiChjang, line 671: This is less of a problem since it happens on the origin chain, and not the intermediary chain. We could thus use JIT withdrawal to resolve this.

However, we may want a consistent UX, which would require us to not use JIT withdrawal and remove the SetFeesMode instruction when executing the initial teleport or reserve asset transfer.

franciscoaguirre:
bot fmt

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: The thing with jit withdrawal is that it's not expected.
Users pay `assets` and expect execution fees to be deducted from that amount.
With jit withdrawal we are deducting delivery fees from the user's account itself, which might not have more assets after the transfer.
So user has to make sure they pay `assets` and also have enough for delivery fees.
This makes it so they only pay `assets` and both execution and delivery fees are deducted from that amount.

franciscoaguirre:
bot help

franciscoaguirre:
bot bench polkadot-pallet --runtime westend --subcommand xcm --pallet pallet_xcm
bot bench polkadot-pallet --runtime rococo --subcommand xcm --pallet pallet_xcm

franciscoaguirre:
bot bench polkadot-pallet --runtime westend --subcommand xcm --pallet pallet_xcm
bot bench polkadot-pallet --runtime rococo --subcommand xcm --pallet pallet_xcm

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -646,11 +675,25 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 				self.send(reserve, Xcm(message), FeeReason::InitiateReserveWithdraw)?;
 				Ok(())
 			},
 			InitiateTeleport { assets, dest, xcm } => {
 				// We must do this first in order to resolve wildcards.
+				let to_weigh = self.holding.saturating_take(assets.clone());
+				self.holding.subsume_assets(to_weigh.clone());
+
+				let mut message_to_weigh =
+					vec![ReceiveTeleportedAsset(to_weigh.clone().into()), ClearOrigin];
```
acatangiu, line 690:
```suggestion
					vec![ReceiveTeleportedAsset(to_weigh.into()), ClearOrigin];
```

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -633,10 +646,26 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 				Ok(())
 			},
 			InitiateReserveWithdraw { assets, reserve, xcm } => {
+				// we need to do this take/put cycle to solve wildcards and get exact assets to be
+				// weighed
+				let to_weigh = self.holding.saturating_take(assets.clone());
+				self.holding.subsume_assets(to_weigh.clone());
+
+				let mut message_to_weigh =
+					vec![WithdrawAsset(to_weigh.clone().into()), ClearOrigin];
```
acatangiu, line 662:
```suggestion
					vec![WithdrawAsset(to_weigh.into()), ClearOrigin];
```

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -624,6 +624,19 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+				// weighed
+				let to_weigh = self.holding.saturating_take(assets.clone());
+				self.holding.subsume_assets(to_weigh.clone());
+
+				let mut message_to_weigh =
+					vec![ReserveAssetDeposited(to_weigh.into()), ClearOrigin];
+				message_to_weigh.extend(xcm.0.clone().into_iter());
+				let (_, fee) = validate_send::<Config::XcmSender>(dest, Xcm(message_to_weigh))?;
+				// set aside fee to be charged by XcmSender
+				let parked_fee = self.holding.saturating_take(fee.into());
```
acatangiu, line 637: nit:

this can be extracted in some `fn extract_transport_fee()` or similar and reused

polkadot/runtime/rococo/src/weights/xcm/pallet_xcm.rs
```diff
@@ -0,0 +1,252 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Autogenerated weights for `pallet_xcm`
```
acatangiu: don't think this is the right place for these files

polkadot/xcm/xcm-executor/src/lib.rs
acatangiu: this is definitely needed for InitiateReserveWithdraw too but not for the current scenario where the intermediary chain is the reserve - I say we keep it, but add a check if jit_withdraw is set - if it is, then we can skip it and let it withdraw from origin's account instead of holding

(maybe we do this jit check for all of them, but honestly I would just deprecate JIT_WITHDRAW altogether, it just introduces extra complexity and error cases)

franciscoaguirre:
cargo fmt

franciscoaguirre:
bot fmt

Cargo.lock
```diff
@@ -1330,7 +1281,7 @@ dependencies = [
 
 [[package]]
 name = "binary-merkle-tree"
-version = "4.0.0-dev"
+version = "10.0.0"
```
acatangiu, line 1284: where are all these changes coming from?

for complete crate compatibility on this patch release we shouldn't bump any dependencies references

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: I removed the `SetFeesMode { jit_withdraw: true }` instruction from the teleport and reserve asset transfer extrinsics and re-run the benchmarks. It should only charge for the fees once now and in a more consistent way.

acatangiu (commented):
Oh, and forgot to ask, why did you remove it from `TransferReserveAsset`? I can see same need for it as `InitiateReserveWithdraw`

franciscoaguirre:
> Oh, and forgot to ask, why did you remove it from `TransferReserveAsset`? I can see same need for it as `InitiateReserveWithdraw`

`TransferReserveAssets` doesn't put assets into holding, it only calls the native transfer from one account to another. In this case, we can't set aside the delivery fee from the assets because they are not in the holding.

We can either:
- Put the assets in holding first to reuse the `saturating_take` implementation in `AssetsInHolding`
- Implement something similar to `saturating_take` in `MultiAssets`

Cargo.lock
franciscoaguirre: I don't really know, but the benchmark job needed it to not throw an error. All the changes are right here in this PR, so no idea. I can remove this file from here and hope it builds without the changes.

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -633,10 +646,26 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 			InitiateReserveWithdraw { assets, reserve, xcm } => {
+				// we need to do this take/put cycle to solve wildcards and get exact assets to be
+				// weighed
+				let to_weigh = self.holding.saturating_take(assets.clone());
+				self.holding.subsume_assets(to_weigh.clone());
+
+				let mut message_to_weigh =
+					vec![WithdrawAsset(to_weigh.into()), ClearOrigin];
+				message_to_weigh.extend(xcm.0.clone().into_iter());
+				let (_, fee) = validate_send::<Config::XcmSender>(reserve, Xcm(message_to_weigh))?;
```
ERussel, line 664: As I understand fee is calculated in the native asset (for example, KSM). How will it work for the cases when, for example, RMRK or USTD is transferred from the Kusama Asset Hub and no KSM in the holding? Will the extrinsic itself fail on the Asset Hub and the state will be rolled back?

polkadot/xcm/xcm-executor/src/lib.rs
acatangiu: that's true for `pallet-xcm`, but what about other xcm pallets (e.g. `xtokens` who surfaced this)?

we can't assume in executor that no one will use `SetFeesMode { jit_withdraw: true }` - as long as we have it, we have to assume it can/will be used

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: Until fees can be paid in another asset, there should always be enough KSM to pay for fees, not only for delivery, also for execution.

ERussel: Does this mean that KSM must be included into the xcm together with the sending token (USDT)?

polkadot/xcm/xcm-executor/src/lib.rs
albertov19: @franciscoaguirre, just to confirm; this means that we can't just send now (as an example) USDT to a para chain from AssetHub. We also have to include KSM and AT LEAST the amount of the XCM delivery fee? And then whatever we put over the min XCM Delivery Fee amount will be delivered to the target account in the target chain?

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: Yes. It should've always been the case in order to pay for the regular execution fees.

acatangiu (approved):

franciscoaguirre:
Closing in favor of https://github.com/paritytech/polkadot-sdk/pull/3085

rageruun:
can anyone help? Tokens stuck in parachain bridge when transmitted from the Kusama network to the Moonriver (xcKSM)
image
hash https://kusama.subscan.io/extrinsic/0x37b705c316c4b4b8d552964a130a552e06bad1f55171fca693e4b4e1917dd336?tab=event

bkchr:
@michalisFr can you help here please.

rageruun:
> can anyone help? Tokens stuck in parachain bridge when transmitted from the Kusama network to the Moonriver (xcKSM) image hash https://kusama.subscan.io/extrinsic/0x37b705c316c4b4b8d552964a130a552e06bad1f55171fca693e4b4e1917dd336?tab=event

rageruun:
can you help with a stuck transaction?

пн, 29 янв. 2024 г. в 21:11, Bastian Köcher ***@***.***>:

albertov19:
@rageruun please do not spam this issue.

The Moonbeam is working on indexing all the failed transactions and it will be submitting a governance proposal to help in the cases were assets are being held by the Moonriver owned account in other chains

rageruun:
то есть можно ожидать успешного завершения транзакции?

пн, 29 янв. 2024 г. в 23:20, albertov19 ***@***.***>:

michalisFr:
> @michalisFr can you help here please.

@gpestana What could I help with?

Issue #3039, opened 2024-01-24
[RPC-Spec-V2] transaction: Implement transaction_unstable_broadcast


lexnv:
Implement the newly added RPC method for broadcasting transactions [transaction_unstable_broadcast](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_broadcast.md).

The `transaction_unstable_broadcast` receives a hex-encoded scale-encoded transaction and returns an operation ID.
No information about the validity of the transaction is reported to users.
The `transaction_unstable_broadcast` broadcasts the transaction to peers until the user calls [transaction_unstable_stop](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_stop.md#transaction_unstable_stop)

PR #3079, opened 2024-01-26
rpc-v2/tx: Implement `transaction_unstable_broadcast` and `transaction_unstable_stop` 

lexnv:
This PR implements the [transaction_unstable_broadcast](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_broadcast.md) and [transaction_unstable_stop](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_stop.md).

The [transaction_unstable_broadcast](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_broadcast.md) submits the provided transaction at the best block of the chain.
If the transaction is dropped or declared invalid, the API tries to resubmit the transaction at the next available best block.

### Broadcasting 
The broadcasting operation continues until either:

- the user called `transaction_unstable_stop` with the operation ID that identifies the broadcasting operation
- the transaction state is one of the following: 
  - Finalized: the transaction is part of the chain
  - FinalizedTimeout: we have waited for 256 finalized blocks and timedout
  - Usurped the transaction has been replaced in the tx pool
  
The broadcasting retires to submit the transaction when the transaction state is:
- Invalid: the transaction might become valid at a later time
- Dropped: the transaction pool's capacity is full at the moment, but might clear when other transactions are finalized/dropped

### Stopping

The `transaction_unstable_broadcast` spawns an abortable future and tracks the abort handler.
When the [transaction_unstable_stop](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_stop.md) is called with a valid operation ID; the abort handler of the corresponding `transaction_unstable_broadcast` future is called. This behavior ensures the broadcast future is finishes on the next polling.
When the `transaction_unstable_stop` is called with an invalid operation ID, an invalid jsonrpc specific error object is returned. 

### Testing

This PR adds the testing harness of the transaction API and validates two basic scenarios:
- transaction enters and exits the transaction pool
- transaction stop returns appropriate values when called with valid and invalid operation IDs

Closes: https://github.com/paritytech/polkadot-sdk/issues/3039

Note that the API should be enabled after: https://github.com/paritytech/polkadot-sdk/issues/3084.

cc @paritytech/subxt-team


substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	/// Generate an unique operation ID for the `transaction_broadcast` RPC method.
+	pub fn generate_unique_id(&self) -> String {
+		let generate_operation_id = || {
+			// The lenght of the operation ID.
+			const OPERATION_ID_LEN: usize = 16;
+
+			let mut rng = rand::thread_rng();
+			(&mut rng)
```
jsdw, line 70: I can see that this is documented in `rand` because `sample_iter` takes `self`, but since you don't use `rng` again after this line, perhaps you don't need it?

lexnv: Thanks for the suggestion! Looks much cleaner now

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+			const OPERATION_ID_LEN: usize = 16;
+
+			rand::thread_rng()
+				.sample_iter(Alphanumeric)
+				.take(OPERATION_ID_LEN)
+				.map(char::from)
+				.collect::<String>()
+		};
+
+		let mut id = generate_operation_id();
```
jsdw, line 76: Would there be any downside to having an incrementing u64 or whatever for the operation ID (like I think we do with the chainHead operation IDs) instead of randomly generating them?

The one I can think of is that, unless the operation ID is per connected client, it would leak details of eg how many other operations were in progress. I guess with the chainHead one, it is scoped nicely to a single chainHead_follow subscription anyway.

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+				// The `now_or_never` returns `None` if the stream is `Pending`.
+				//
+				// If the stream contains `Hash0x1 Hash0x2 Hash0x3 Hash0x4`, we want only `Hash0x4`.
+				while let Some(next) = best_block_import_stream.next().now_or_never() {
+					let Some(next) = next else {
+						// Nothing to do if the best block stream terminated.
+						return
+					};
+					best_block_hash = next;
+				}
```
jsdw, line 139: Random thought: this "fetch the last available item from stream" could perhaps be wrapped into a little helper function and then that could be tested (though offhand I wouldn't know what to call it!)

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+						},
+
+						// The transaction is still in the pool, the ready or future queue.
+						TransactionStatus::Ready | TransactionStatus::Future |
+						// Transaction has been broadcasted as intended.
+						TransactionStatus::Broadcast(_) |
+						// Transaction has been included in a block, but the block is not finalized yet.
+						TransactionStatus::InBlock(_) |
+						// Transaction has been retracted, but it may be included in a block at a later time.
+						TransactionStatus::Retracted(_) => (),
```
jsdw, line 182: How confident are you that if we see a Retracted event, we'll see another event again for the tx in the future? :)

(In Subxt, we also keep waiting for further events after a Retracted, but I'm not 100% sure whether one will always be emitted!)

substrate/client/rpc-spec-v2/src/transaction/tests.rs
```diff
@@ -0,0 +1,166 @@
...
+		.unwrap_err();
+	assert_matches!(err,
+		Error::Call(err) if err.code() == super::error::json_rpc_spec::INVALID_PARAM_ERROR && err.message() == "Invalid operation id"
+	);
+
+	// Ensure stop can be called, the tx was decoded and the broadcast future terminated.
+	let _: () = tx_api
+		.call("transaction_unstable_stop", rpc_params![&operation_id])
+		.await
+		.unwrap();
```
jsdw, line 165: Would there be any benefit to waiting a bit to ensure that the tx pool has marked the tx as invalid etc by the time you try to stop it, to show that you can still stop a tx even if it's marked invalid?

substrate/client/rpc-spec-v2/src/transaction/tests.rs
```diff
@@ -0,0 +1,166 @@
...
+		tx_api.call("transaction_unstable_broadcast", rpc_params![&xt]).await.unwrap();
+
+	// Make an invalid stop call.
+	let err = tx_api
+		.call::<_, serde_json::Value>("transaction_unstable_stop", ["invalid_operation_id"])
+		.await
+		.unwrap_err();
+	assert_matches!(err,
+		Error::Call(err) if err.code() == super::error::json_rpc_spec::INVALID_PARAM_ERROR && err.message() == "Invalid operation id"
+	);
```
jsdw, line 159: Is this really a separate independent test, ie that calling unstable_stop with an invalid ID will result in an error?

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
jsdw: I'd just go for a single function that mutably borrows a stream and returns the last "next available thing" back, I reckon; no need for a trait :)

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
jsdw: Coolio; well, I'm happy as is but always a little scared that we won't get any future events because the "old" API doesn't make any promises really!

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+					};
+					best_block_hash = next;
+				}
+
+				let submit =
+					pool.submit_and_watch(best_block_hash, TX_SOURCE, decoded_extrinsic.clone());
+
+				// The transaction was not included to the pool, because it is invalid.
+				// However an invalid transaction can become valid at a later time.
+				let Ok(mut stream) = submit.await else { return };
```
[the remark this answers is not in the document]
lexnv: Ahh yea, this is wrong. Must be a left-over comment, will remove it : D

jsdw (approved):
Looked good to me (just a couple of tiny things really)! I'd feel reassured if somebody with more experience with the transaction pool stuff could check that we are handling the events properly (I'm never super confident in what tx pool events could be final versus which will def be followed by more events)

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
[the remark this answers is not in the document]
lexnv: I've had another look over this while writing some extra tests.

It seems indeed that the tx did not enter the pool because of an error.

There's only one that we might need to care about: 

https://github.com/paritytech/polkadot-sdk/blob/6ea472ad5a68018dc054ec0a5d866fc481d077e4/substrate/client/transaction-pool/api/src/error.rs#L45-L46

However, I still think we should return here and do nothing the tx further. If I'm looking at the right place in the code, the default time for marking a tx as "banned" is 30 mins:

https://github.com/paritytech/polkadot-sdk/blob/6ea472ad5a68018dc054ec0a5d866fc481d077e4/substrate/client/transaction-pool/src/graph/rotator.rs#L49

It might turn out to be beneficial to add to the spec a soft limit "tx broadcast ensures the tx is broadcasted for at least a number of blocks (16) if the server chooses to validate the tx" -- then we could safely return from here; and when we drop this future we remove `BroadcastState` from internal tracking

substrate/client/rpc-spec-v2/src/transaction/api.rs
```diff
@@ -28,10 +28,33 @@ pub trait TransactionApi<Hash: Clone> {
...
 }
+
+#[rpc(client, server)]
+pub trait TransactionBroadcastApi<Hash: Clone> {
+	/// Broadcast an extrinsic to the chain.
+	///
+	/// # Unstable
+	///
+	/// This method is unstable and subject to change in the future.
+	#[method(name = "transaction_unstable_broadcast", blocking)]
```
niklasad1, line 50: I don't understand why this is "blocking", AFAIU this should just be an ordinary method i.e, doesn't wait sync I/O such as the storage DB.

substrate/client/rpc-spec-v2/src/transaction/api.rs
```diff
@@ -28,10 +28,33 @@ pub trait TransactionApi<Hash: Clone> {
...
+	fn broadcast(&self, bytes: Bytes) -> RpcResult<Option<String>>;
+
+	/// Broadcast an extrinsic to the chain.
+	///
+	/// # Unstable
+	///
+	/// This method is unstable and subject to change in the future.
+	#[method(name = "transaction_unstable_stop", blocking)]
+	fn stop_broadcast(&self, operation_id: String) -> Result<(), ErrorBroadcast>;
```
niklasad1, line 59: same here, 

I don't understand why this is "blocking", AFAIU this should just be an ordinary method i.e, doesn't wait sync I/O such as the storage DB.

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+#[async_trait]
+impl<Pool, Client> TransactionBroadcastApiServer<BlockHash<Pool>>
+	for TransactionBroadcast<Pool, Client>
+where
+	Pool: TransactionPool + Sync + Send + 'static,
+	Pool::Hash: Unpin,
+	<Pool::Block as BlockT>::Hash: Unpin,
+	Client: HeaderBackend<Pool::Block> + BlockchainEvents<Pool::Block> + Send + Sync + 'static,
+{
+	fn broadcast(&self, bytes: Bytes) -> RpcResult<Option<String>> {
```
niklasad1, line 104: why is the return type `Option<String>` it looks like always an ID is returned?!

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,217 @@
...
+			while !is_done {
+				// Wait for the next block to become available.
+				let Some(mut best_block_hash) = best_block_import_stream.next().await else {
+					return
+				};
+				// We are effectively polling the stream for the last available item at this time.
+				// The `now_or_never` returns `None` if the stream is `Pending`.
+				//
+				// If the stream contains `Hash0x1 Hash0x2 Hash0x3 Hash0x4`, we want only `Hash0x4`.
+				while let Some(next) = best_block_import_stream.next().now_or_never() {
```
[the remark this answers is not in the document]
jsdw: I mentioned a helper fn too; I think that'd be quite a nice way to abstract (and be able to test) this behaviour :))

substrate/client/rpc-spec-v2/src/transaction/api.rs
jsdw: Good point; I had just thought "oh, it's not async so this is ok" but I reminded myself that blocking means it runs the task in `tokio::task::spawn_blocking`, which doesn't appear to be necessary here since not much work happens outside the future

substrate/client/rpc-spec-v2/src/transaction/tests.rs
lexnv: Yep good call! I have added a few pool checks for this. We could capture this more properly in my followup which adds a middleware -- ensuring that we drop the broadcast future without calling any pool fn

substrate/client/rpc-spec-v2/src/transaction/tests.rs
lexnv: Yep good call! I have added a few pool checks for this. We could capture this more properly in my followup which adds a middleware -- ensuring that we drop the broadcast future without calling any pool fn

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
lexnv: Yep, the downside would be that we'll be leaking the total number of transactions submitted by this API across all connections. We could replace this in the future when we add support for https://github.com/paritytech/polkadot-sdk/issues/3081

substrate/client/rpc-spec-v2/src/transaction/api.rs
lexnv: Yep that makes sense, I think I just copied the method attribute from somewhere else. Have made it non-blocking

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,292 @@
...
+				// Wait for the last block to become available.
+				let Some(best_block_hash) =
+					last_stream_element(&mut best_block_import_stream).await
+				else {
+					return;
+				};
+
+				let mut stream = match pool
+					.submit_and_watch(best_block_hash, TX_SOURCE, decoded_extrinsic.clone())
+					.await
```
lexnv, line 136: We are effectively trying to resubmit the transaction to the next best block here; if we cannot make progress at this time:
- `submit_and_watch` directly returned a recoverable error (ImmediatelyDroped etc)
- `submit_and_watch` produced an event that may be recoverable (Invalid / Dropped etc)

When we get the `Invalid` event, the transaction-pool will ban it for some time (default 30 mins).  In this case, the looping seems redundant unless chain developers manually pick another timeout.

However, I believe we can simplify this and try to submit the transaction only once. 
From the [transaction_unstable_broadcast](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_broadcast.md#transaction_unstable_broadcast), the phrasing states that the server tries to broadcast, but without any guarantees: 
> JSON-RPC server will try to propagate this transaction

This could be interpreted as we'll submit the transaction to the transaction pool now, and if it fails or is invalid we'll stop.

@skunert would love to get your thoughts on this 🙏

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,292 @@
...
+/// some unique transactions via RPC and have them included in the pool.
+const TX_SOURCE: TransactionSource = TransactionSource::External;
+
+#[async_trait]
+impl<Pool, Client> TransactionBroadcastApiServer<BlockHash<Pool>>
+	for TransactionBroadcast<Pool, Client>
+where
+	Pool: TransactionPool + Sync + Send + 'static,
+	Pool::Error: IntoPoolError,
+	Pool::Hash: Unpin,
```
skunert, line 102: I think these Unpin requirements could be removed too.

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
skunert: Yeah sounds like its not enforced by the spec, but IMO the current implementation makes sense. If we can recover and retry in the next block, why not.

substrate/client/rpc-spec-v2/src/transaction/api.rs
```diff
@@ -28,10 +28,33 @@ pub trait TransactionApi<Hash: Clone> {
...
 	#[subscription(
 		name = "transactionWatch_unstable_submitAndWatch" => "transactionWatch_unstable_watchEvent",
 		unsubscribe = "transactionWatch_unstable_unwatch",
 		item = TransactionEvent<Hash>,
 	)]
 	fn submit_and_watch(&self, bytes: Bytes);
 }
+
+#[rpc(client, server)]
+pub trait TransactionBroadcastApi<Hash: Clone> {
```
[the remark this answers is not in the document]
lexnv: Yep that's not used :D Thanks for spotting this!

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -0,0 +1,292 @@
...
+		// `transaction_stop` method.
+		let (fut, handle) = futures::future::abortable(broadcast_transaction_fut);
+		// The future expected by the executor must be `Future<Output = ()>` instead of
+		// `Future<Output = Result<(), Aborted>>`.
+		let fut = fut.map(drop);
+
+		// Keep track of this entry and the abortable handle.
+		{
+			let mut broadcast_ids = self.broadcast_ids.write();
+			broadcast_ids.insert(id.clone(), BroadcastState { handle });
```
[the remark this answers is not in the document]
lexnv: Good catch! I've removed the `operationId` from our internal mappings as soon as the broadcast future exits.

I believe this is in-line with the spec; since the [tx_stop](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_stop.md) mentions it works for active broadcasts:

> A JSON-RPC error is generated if the operationId doesn't correspond to any active transaction_unstable_broadcast operation.

As a side-effect have added a small task executor wrapper for our testing; just to make sure we don't race between tx_stop and broadcast future clearing the internal state

skunert (approved):
Looks nice!

PR #3082, opened 2024-01-26
Error on duplicated crates

ggwpez:
#2970 accidentally added a crate twice to the workspace. Now extending the workspace check to explicitly error then.  

I think the check should also be required now.


liamaharon (approved):

seadanda (approved):

bkchr (approved):
Nice :rocket:

PR #2405, opened 2023-11-20
pallet-xcm: ensure xcm outcome is always complete, revert effects otherwise

xlc:
On extrinsics/call, ensure local XCM execution is complete/successful. Otherwise, fail the extrinsic so that state changes don't get committed to the db.

Added regression tests that fail without the fix.

fixes #2237

PR #3086, opened 2024-01-26
Contracts: xcm host fn fixes

pgherveou:
## Xcm changes:
- Fix `pallet_xcm::execute`, move the logic into The `ExecuteController` so it can be shared with anything that implement that trait.
- Make `ExecuteController::execute` retursn `DispatchErrorWithPostInfo` instead of `DispatchError`, so that we don't charge the full `max_weight` provided if the execution is incomplete (useful for force_batch or contracts calls)
- Fix docstring for `pallet_xcm::execute`, to reflect the changes from #2405
- Update the signature for `ExecuteController::execute`, we don't need to return the `Outcome` anymore since we only care about `Outcome::Complete`

## Contracts changes:

- Update host fn `xcm_exexute`, we don't need to write the `Outcome` to the sandbox memory anymore. This was also not charged as well before so it if fixes this too.
- One of the issue was that the dry_run of a contract that call `xcm_execute` would exhaust the `gas_limit`.

This is because `XcmExecuteController::execute` takes a `max_weight` argument, and since we don't want the user to specify it manually we were passing everything left by pre-charghing `ctx.ext.gas_meter().gas_left()`

- To fix it I added a `fn influence_lowest_limit` on the `Token` trait and make it return false for `RuntimeCost::XcmExecute`.
- Got rid of the `RuntimeToken` indirection, we can just use `RuntimeCost` directly.


pgherveou:
bot fmt

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -2136,26 +2122,27 @@ pub mod env {
 		ctx.charge_gas(RuntimeCosts::CopyFromContract(msg_len))?;
 		let message: VersionedXcm<CallOf<E::T>> =
 			ctx.read_sandbox_memory_as_unbounded(memory, msg_ptr, msg_len)?;
+		ensure_executable::<E::T>(&message)?;
 
 		let execute_weight =
 			<<E::T as Config>::Xcm as ExecuteController<_, _>>::WeightInfo::execute();
 		let weight = ctx.ext.gas_meter().gas_left().max(execute_weight);
```
athei, line 2129: Why would we even pass in all the remaining weight if it is bigger than `execute_weight`? `execute_weight` should be  the upper bound for the weight than can be used.  We are not doing that either for `call_runtime`. This seems to be the root of the problem.

Why not just use the `execute_weight`?

athei (commented):
This fixes the problem but maybe we can side step it entirely. See my comment.

substrate/frame/contracts/src/wasm/runtime.rs
pgherveou: execute_weight is actually the base weight for executing the pallet_xcm dispatchable, then you top up with max_weight to pay for it.

So, I think my reasoning here was that we can't pass the function anything less than execute_weight. 

https://github.com/paritytech/polkadot-sdk/blob/a8ea8610c15a2426ea20067b52fb106ee6dc97f8/polkadot/xcm/pallet-xcm/src/lib.rs#L1021-L1036

athei: Ahh yes you are right. I thought this was the upper limit from a `Weigher`.

athei:
I thought about this change again. While this fixes the problem that the `gas_required` will be too high it will introduce the inverse problem: It might be too low. Since we **require** the gas to be there before calling into the XCM it has influence the lowest experienced gas. Otherwise it will become too optimistic. Reason is that the XCM execution itself is not metered. So we have to check that the gas is actually there before executing as we do.

So the current code (before the PR) is actually correct with regard to influencing the lowest experienced gas.

The actual problem here is that taking "all the gas" is a very vast overestimation of what the XCM needs. Especially  in case of dry run where we supply the whole block gas limit.

There are only two fixes for my problem I can thing of:
- Let the contract specify a limit
- Use a `Weigher` as `pallet-xcm` does in order to come up with a better upper bound

The problem of the second approach is that `WeightInfo::weight()` is fallible. I assume we cannot find out the pre-dispatch weight for all XCM programs. I hope there is some metering going on in the background in those causes. I assume there has to be since `pallet-xcm::execute` uses the same mechanism.

You have to investigate what is going on there? What if a program is not Weighable? How does pallet-xcm handle this situation? Does the executor about the execution some where?

pgherveou:
> It might be too low

why is that? We know that the pre-dispatch weight is `execute_weight` don't we get exactly what we need by letting the xcm run with a lot of weight and then using the refund to calculate a precise dry-run weight for the call.

athei:
You are right this should work because the weight we take is dependent on the remaining weight. As long as it is not a fixed number it should work.

athei (approved):

pgherveou:
@paritytech/xcm can you take a look at these minor changes in pallet-xcm when you have a chance 🙏

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5148520

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5148505

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5148504

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5148503

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5148521

substrate/frame/contracts/uapi/Cargo.toml
```diff
@@ -23,6 +23,9 @@ scale = { package = "parity-scale-codec", version = "3.6.1", default-features =
 [target.'cfg(target_arch = "riscv32")'.dependencies]
 polkavm-derive = '0.5.0'
 
+[package.metadata.docs.rs]
+targets = ["wasm32-unknown-unknown", "x86_64-unknown-linux-gnu"]
+
```
athei, line 28: Seems unrelated to this PR. Merge failure somehow?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5155531

substrate/frame/contracts/uapi/Cargo.toml
pgherveou: arg yes will remove thanks

pgherveou: fyi updated slightly the xcm dispatchable so that it returns DispatchErrorWithPostInfo instead of DispatchError so that we don't run out of gas if ever we want to recover from these errors

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5157124

athei (commented):
Yes the change makes sense to me that we are only return `Ok` for a complete execution.

pgherveou:
Ping @KiChjang or @franciscoaguirre

franciscoaguirre (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5254494

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5260479

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5262268

PR #3087, opened 2024-01-26
Link mermaid docs to crate that uses them

Morganamilo:
While including files cross crate works locally. When pushed to crates.io each crate is seperate so the file path no longer exists.

Instead change it to a symlink, which cargo will change to a read directory when published.


substrate/primitives/runtime/docs
```diff
@@ -0,0 +1 @@
+../../../docs
```
bkchr, line 1: I don't get what this file is doing?

ggwpez: It think its GitHub's way of rendering links.

ggwpez (approved):
Yes this is why we need to run `cargo publish --dry-run` in the CI to check that all crates are in publishable state.

bkchr (approved):

Morganamilo:
> Yes this is why we need to run `cargo publish --dry-run` in the CI to check that all crates are in publishable state.

It's not that simple as a dry run publish needs all of the dependencies published for real.

ggwpez:
> It's not that simple as a dry run publish needs all of the dependencies published for real.

I remember JP forked and modified the code of the crates-io server to simulate that... but it was a pain to maintain.

PR #3095, opened 2024-01-27
Fix misleading comment about RewardHandler in epm config

btwiuse:
In pallet_election_provider_multi_phase::Config, the effect of

       type RewardHandler = ()

is to mint rewards from the void, not "nothing to do upon rewards"


btwiuse:
Previously the gear team was investigating why the total supply was increasing when theoretically it shouldn't (https://github.com/gear-tech/gear/pull/3389)

It turned the newly minted tokens went to some staking-miner operator, contrary to our initial assumption that it would do "nothing upon rewards".

bkchr:
This pull request doesn't make any sense. This should be documented in the pallet. Like in the docs it says how this pallet works.

In general the logic of just creating the fees by increasing the total insurance looks weird to me.

bkchr:
Basically this changes the inflation.

@gpestana is this really what we want?

kianenigma:
For context, the decision to "mint" the rewards that are paid out to the staking miner was intentional at the time. It is a good idea to provide a way now configure this, such that one option is to mint, and another is to  (try to) transfer them from a known account, eg. The treasury.

kianenigma (approved):

kianenigma:
A part of the issue here is also that `traits::Currency` impl doesn't emit any events related to increasing the issunace.

bkchr (approved):

PR #3106, opened 2024-01-29
With many on-demand cores, performance in order processing is subpar

eskimor:
This is a first rudimentary fix. Actual implementation perf improvements are no their way.

This also includes a fix where we used max instead of min: Will become significant once we have >1 claimqueues.

### DIFF
--- a/polkadot/primitives/src/v6/mod.rs
+++ b/polkadot/primitives/src/v6/mod.rs
@@ -393,7 +393,7 @@ pub const MAX_POV_SIZE: u32 = 5 * 1024 * 1024;
 /// Default queue size we use for the on-demand order book.
 ///
 /// Can be adjusted in configuration.
-pub const ON_DEMAND_DEFAULT_QUEUE_MAX_SIZE: u32 = 10_000;
+pub const ON_DEMAND_DEFAULT_QUEUE_MAX_SIZE: u32 = 500;
 
 /// Backing votes threshold used from the host prior to runtime API version 6 and from the runtime
 /// prior to v9 configuration migration.
--- a/polkadot/runtime/parachains/src/scheduler.rs
+++ b/polkadot/runtime/parachains/src/scheduler.rs
@@ -584,7 +584,8 @@ impl<T: Config> Pallet<T> {
 			return
 		}
 		// If there exists a core, ensure we schedule at least one job onto it.
-		let n_lookahead = Self::claimqueue_lookahead().max(1);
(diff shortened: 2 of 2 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5056241

polkadot/primitives/src/v6/mod.rs
```diff
@@ -393,7 +393,7 @@ pub const MAX_POV_SIZE: u32 = 5 * 1024 * 1024;
 /// Default queue size we use for the on-demand order book.
 ///
 /// Can be adjusted in configuration.
-pub const ON_DEMAND_DEFAULT_QUEUE_MAX_SIZE: u32 = 10_000;
+pub const ON_DEMAND_DEFAULT_QUEUE_MAX_SIZE: u32 = 500;
```
sandreim, line 396: Is this sufficient to not build overweight blocks ?

polkadot/runtime/parachains/src/scheduler.rs
```diff
@@ -584,7 +584,8 @@ impl<T: Config> Pallet<T> {
 			return
 		}
 		// If there exists a core, ensure we schedule at least one job onto it.
-		let n_lookahead = Self::claimqueue_lookahead().max(1);
+		// TODO (now): Add a test for this. This wax `max` and nobody noticed.
+		let n_lookahead = Self::claimqueue_lookahead().min(1);
```
bkchr, line 588: This was max out of the reason to always have at least 1 lookahead.

bkchr: Because `0` makes no sense.

polkadot/primitives/src/v6/mod.rs
bkchr: This changes will not fix anything. This variable is only used for migrations or at genesis. If you want to lower this, you need to set this in the `HostConfiguration` on chain.

polkadot/runtime/parachains/src/scheduler.rs
eskimor: LOL. Just realizing I was after all holding it wrong. Thanks.

eskimor:
@antonva closing this one, as nothing is really left of this insubstantial PR anyways. Can you make sure queue size is reduced effectively. The default is still a good idea to also change, but this will already be part of my bigger PR anyways.

PR #1222, opened 2023-08-28
transactional processing for xcm

vstam1:
Moved from: https://github.com/paritytech/polkadot/pull/6951

closes https://github.com/paritytech/polkadot-sdk/issues/490

PR #3109, opened 2024-01-29
Transactional account modification

ggwpez:
Improvised replacement of #1222 for the current release.


bkchr:
We can not just yolo change some things to transactional. Can we not backport parts of the #1222 pr?

PR #3114, opened 2024-01-29
Add limits to XCMv4

franciscoaguirre:
Some more work regarding XCMv4. Two limits from v3 were not transferred over, those are:
- The instructions limit
- The number of assets limit
Both of these are now in v4.

For some reason `AssetInstance` increased in size, don't know why CI didn't catch that before.


paritytech-cicd-pr:
**Job name**: check-runtime-migration-westend
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5061516

KiChjang (approved):

polkadot/xcm/src/v4/mod.rs
```diff
@@ -68,6 +71,26 @@ pub struct Xcm<Call>(pub Vec<Instruction<Call>>);
...
+		instructions_count::using_once(&mut 0, || {
+			let number_of_instructions: u32 = <Compact<u32>>::decode(input)?.into();
+			instructions_count::with(|count| {
+				*count = count.saturating_add(number_of_instructions as u8);
+				if *count > MAX_INSTRUCTIONS_TO_DECODE {
+					return Err(CodecError::from("Max instructions exceeded"))
+				}
+				Ok(())
+			})
+			.unwrap_or(Ok(()))?;
```
bkchr, line 87:
```suggestion
			.expect("Called in `using` context and thus can not return `None`; qed")?;
```

acatangiu (approved):

PR #3135, opened 2024-01-30
Bump Bridge Hub Rococo spec version

claravanstaden:
To deploy https://github.com/paritytech/polkadot-sdk/pull/3029


svyatonik (approved):
Approving without knowing much details of how it gonna be actually deployed (from the `master` branch or cherry-picked to release branch or to some dedicated branch, forked off the release branch) - I hope it'll be discussed with devops/releng. If from master, then we should be careful to not to include something that will break the bridge hub - e.g. this https://github.com/paritytech/polkadot-sdk/pull/3117 (which reverts some doubtful change + adds some migrations).

Lederstrumpf (approved):

acatangiu (approved):

PR #3154, opened 2024-01-31
Contracts: Stabilize `caller_is_root` API

xermicus:
Can this API be marked stable? Implemented in [solang here](https://github.com/hyperledger/solang/pull/1620)

### DIFF
--- a/prdoc/pr_3154.prdoc
+++ b/prdoc/pr_3154.prdoc
@@ -1,0 +1,8 @@
+title: Contracts: Stabilize caller_is_root API
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Removed the `#[unstable]` attrribute on `caller_is_root` host function.
+
+crates: ["pallet-contracts"]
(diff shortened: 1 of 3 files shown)


pgherveou (commented):
This need to be added to ink! integration tests first

athei (approved):
IMHO an integration test in solang is fine, too. We were just saying we need an integration test in ink! because we just assumed they would implement it first.  If it works for Solidity it should definitely work for ink!. That said, if it were a more complex function (like XCM) we would probably want both implementations.

pgherveou (approved):

pgherveou:
> IMHO an integration test in solang is fine, too. We were just saying we need an integration test in ink! because we just assumed they would implement it first. If it works for Solidity it should definitely work for ink!. That said, if it were a more complex function (like XCM) we would probably want both implementations.

@xermicus maybe add a ticket to add a Test for the ink! team so we don't forget

PR #3159, opened 2024-01-31
Config migration for on demand max queue size

antonva:
Decreases the value from `10_000` to `500`.

### DIFF
--- a/polkadot/runtime/parachains/src/configuration/migration/v12.rs
+++ b/polkadot/runtime/parachains/src/configuration/migration/v12.rs
@@ -26,23 +26,8 @@ use sp_std::vec::Vec;
 
 use frame_support::traits::OnRuntimeUpgrade;
 
-use super::v10::V10HostConfiguration;
-pub type V11HostConfiguration<BlockNumber> = configuration::HostConfiguration<BlockNumber>;
-
-mod v10 {
-	use super::*;
-
-	#[frame_support::storage_alias]
-	pub(crate) type ActiveConfig<T: Config> =
-		StorageValue<Pallet<T>, V10HostConfiguration<BlockNumberFor<T>>, OptionQuery>;
-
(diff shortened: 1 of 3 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5089613

polkadot/runtime/parachains/src/configuration/migration/v12.rs
```diff
@@ -0,0 +1,328 @@
...
+	// Unusual formatting is justified:
+	// - make it easier to verify that fields assign what they supposed to assign.
+	// - this code is transient and will be removed after all migrations are done.
+	// - this code is important enough to optimize for legibility sacrificing consistency.
+	#[rustfmt::skip]
+	let translate =
+		|pre: V11HostConfiguration<BlockNumberFor<T>>| ->
+		V12HostConfiguration<BlockNumberFor<T>>
+	{
+		V12HostConfiguration {
```
eskimor, line 109: Creating a complete new `V12HostConfiguration` seems complete overkill, for just changing a value. What is the reason for adding a new type here?

eskimor:
> This "migration" just decreases an on-chain value? Then it should be a referendum of `ParaConfig::set_on_demand_queue_max_size(500)`.
> Otherwise it forces all relay chains to also apply this.

This value is not yet used on any production chain, but should be decreased in value no later than the launch of coretime as the current value is not safe. If this PR makes it into the release, this seems the easiest way to ensure this.

bkchr:
> If this PR makes it into the release, this seems the easiest way to ensure this.

What @ggwpez is the correct way forward to decrease these values on chain. Just go ahead and create the two proposals. No need to wait for any runtime release.

Issue #3129, opened 2024-01-30
prospective parachains API chain: Provide `N` candidates


eskimor:
Don't only provide one candidate, but N, where N is given in the message. Implementation: simple replacement of next() with take(N).

alindima:
I think it's more complicated than a simple replacement with `take(N)`. The current implementation always takes the first available node.
For multiple candidates, we need to return a chain of valid candidates, of the requested size. So we need to do a depth-first search through the tree. If we can't find a chain of N candidates, we should return the chain of the maximum size we could find instead.

eskimor:
Err on what is the cheapest for the validators in the worst case (non cooperating collators). We don't have to find an optimal solution, if the parachain is producing forks.

alindima:
I guess it's no harm in accepting some amount of forks coming from collators. After all, if that weren't the case, the entire fragment trees concept could have been replaced with fragment "chains".
I implemented a depth-first search as part of my PR, but it's bounded by the requested candidate count (which shouldn't be more than two or three). So I think it's a decent solution, which would still permit having forks in the parachains, let me know your thoughts

PR #3160, opened 2024-01-31
prospective-parachains: respond with multiple backable candidates

alindima:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3129


polkadot/node/core/prospective-parachains/src/fragment_tree.rs
```diff
@@ -763,43 +763,106 @@ impl FragmentTree {
...
+					.filter(|(_, hash)| self.scope.get_pending_availability(&hash).is_none())
+					.filter(|(_, hash)| pred(&hash))
+					.map(|(ptr, hash)| (*ptr, *hash))
+					.collect()
+			},
+		};
+
+		let mut best_result = accumulator.clone();
+		for (child_ptr, child_hash) in children {
+			// We could use a HashSet/BTreeSet for tracking the visited nodes but realistically,
```
eskimor, line 836: Why not just use a set right away?

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
```diff
@@ -763,43 +763,106 @@ impl FragmentTree {
...
 		// For plausibly unique parachains, this shouldn't matter much.
 		// figure out alternative selection criteria?
-		match base_node {
+		self.select_children_inner(base_node, count, count, &pred, &mut vec![])
+	}
+
+	// Try finding a candidate chain starting from `base_node` of length `expected_count`.
+	// If not possible, return the longest we could find.
+	// Does a depth-first search, since we're optimistic that there won't be more than one such
+	// chains. Assumes that the chain must be acyclic.
```
eskimor, line 797: What happens if they are cyclic? What is the worst case complexity here, assuming parachains misbehave as much as they possibly can?

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
```diff
@@ -763,43 +763,106 @@ impl FragmentTree {
...
+		};
+
+		let mut best_result = accumulator.clone();
+		for (child_ptr, child_hash) in children {
+			// We could use a HashSet/BTreeSet for tracking the visited nodes but realistically,
+			// accumulator will not hold more than 2-3 elements (the max number of cores for a
+			// parachain).
+			if accumulator.contains(&child_hash) {
+				// We've already visited this node. There's a cycle in the tree, ignore it.
+				continue
```
eskimor, line 841: Seems we handle cycles just fine?

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -146,12 +146,20 @@ async fn run_iteration<Context>(
 					handle_candidate_seconded(view, para, candidate_hash),
 				ProspectiveParachainsMessage::CandidateBacked(para, candidate_hash) =>
 					handle_candidate_backed(&mut *ctx, view, para, candidate_hash).await?,
-				ProspectiveParachainsMessage::GetBackableCandidate(
+				ProspectiveParachainsMessage::GetBackableCandidates(
 					relay_parent,
 					para,
+					n_cores,
```
eskimor, line 152: Is this name justified here? Does prospective parachains need to know about cores in this context? We just want n candidates, what for should be our business.

polkadot/roadmap/implementers-guide/src/node/utility/provisioner.md
```diff
@@ -187,16 +187,16 @@ this process is a vector of `CandidateHash`s, sorted in order of their core inde
...
-An empty required path indicates that the requested candidate should be a direct child of the most recently included
+An empty required path indicates that the requested candidate chain should start with the most recently included
 parablock for the given `para_id` as of the given relay parent.
 
 In contrast, a required path with one or more entries prompts [prospective
 parachains](../backing/prospective-parachains.md) to step forward through its fragment tree for the given `para_id` and
-relay parent until the desired parablock is reached. We then select a direct child of that parablock to pass to the
-provisioner.
+relay parent until the desired parablock is reached. We then select the chain starting with the direct child of that
+parablock to pass to the provisioner.
```
eskimor, line 199: Thanks for not forgetting the guide!

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
alindima: that's a good question. we stop the search on the respective branch when we see that node A has a child that we've already visited. maybe we should just return an error. I don't think it's a valid state transition to accept. Is it?

Even if we were to accept cycles, we'd still be bounded by a function of the requested number of candidates.

alindima: I modified the code in the latest revision to accept cycles in the fragment trees. This is actually a documented behaviour of prospective-parachains even with some tests for it. See: https://github.com/paritytech/polkadot-sdk/blob/74b597fcaf143d8dd7f8d40e59f51065514f21d7/polkadot/node/core/prospective-parachains/src/fragment_tree.rs#L43
I also documented the performance considerations. Cycles are not an issue because we are limited by the core count anyway so we won't loop indefinitely

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
alindima: removed this comment. no longer needed because we allow cycles

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
```diff
@@ -753,53 +753,127 @@ impl FragmentTree {
...
+	// sudo/governance, because:
+	// 1. `num_forks` at a given level is at most `max_candidate_depth * max_validators_per_core`
+	//    (because each validator in the assigned group can second `max_candidate_depth`
+	//    candidates). The prospective-parachains subsystem assumes that the number of para forks is
+	//    limited by collator-protocol and backing subsystems. In practice, this is a constant which
+	//    can only be altered by sudo or governance.
+	// 2. `expected_count` is equal to the number of cores a para is scheduled on (in an elastic
+	//    scaling scenario). For non-elastic-scaling, this is just 1. In practice, this should be a
+	//    small number (1-3), capped by the total number of available cores (a constant alterable
+	//    only via governance/sudo).
```
eskimor, line 816: So with realistic numbers (current and expected/exaggerated future numbers): How bad does it get? What would be limits to those parameters?

eskimor (commented):
Looks sensible! Algorithmic complexity still makes me uneasy: Pinged Rob on what is his stance on forks by now.

eskimor (approved):

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
rphmeier: Elastic scaling will eventually enable chains to be bounded by the rate of a single node's block production.

In Cumulus currently, block authorship is typically slower than block verification (in validators) due to the need to read from disk. Most of the time is spent in accessing the trie for storage, not in code execution. Access patterns for the trie are also very inefficient. So it's not feasible to use more than 3 cores until that bottleneck is cleared.

If this bottleneck is alleviated, then I could see parachains using many more cores.

alindima: thinking about this a bit more, I don't think this traversal is any different in terms of worst-case complexity from the routine used for populating the fragment tree (which populates the tree breadth-first). Is it?
In an acyclic tree, it'll visit each node once (in a worst case scenario). In a cyclic one, it'd be bounded by the requested count so it's the same complexity.

The complexity is linear in the number of nodes, but the number of nodes is an exponential function (because it's an n-ary tree, at least in theory, of arity `num_forks` and height of `max_candidate_depth`).

If we'll hit a bottleneck, this wouldn't be the first place where that'd surface.

I think we could limit the number of total forks allowed for a parachain (regardless of their level in the tree). So that the total number of branches in the tree would bring down the worst-case complexity massively. WDYT?

alindima: anyway, I think this is somewhat orthogonal to this PR.
to quote the docs from fragment_tree module:

```
//! The code in this module is not designed for speed or efficiency, but conceptual simplicity.
//! Our assumption is that the amount of candidates and parachains we consider will be reasonably
//! bounded and in practice will not exceed a few thousand at any time. This naive implementation
//! will still perform fairly well under these conditions, despite being somewhat wasteful of
//! memory.
```

we should revisit this assumption maybe, but it doesn't seem like it'll be a problem for the forseeable future

polkadot/node/core/prospective-parachains/src/tests.rs
```diff
@@ -961,6 +991,572 @@ fn check_backable_query() {
...
+			let candidate_hash = candidate.hash();
+			introduce_candidate(&mut $virtual_overseer, candidate.clone(), pvd).await;
+			second_candidate(&mut $virtual_overseer, candidate.clone()).await;
+			back_candidate(&mut $virtual_overseer, &candidate, candidate_hash).await;
+
+			(candidate, candidate_hash)
+		}};
+	}
+
+	// Parachain 1 looks like this:
```
sandreim, line 1018: 🥇

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
sandreim: `Prospective-parachains` seems to be a good subsystem benchmark target. That should give some clear numbers on CPU usage in worst case scenarios.

sandreim (approved):
🚀

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
alindima: opened a separate issue to track this security item and added it to the elastic scaling task: https://github.com/paritytech/polkadot-sdk/issues/3219

polkadot/node/core/prospective-parachains/src/fragment_tree.rs
```diff
@@ -756,53 +756,127 @@ impl FragmentTree {
...
+				} else {
+					return vec![]
+				};
 			}
 
 			node
 		};
 
-		// TODO [now]: taking the first selection might introduce bias
+		// TODO: taking the first best selection might introduce bias
```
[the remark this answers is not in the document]
alindima: no, this was here beforehand
It could be solved by randomising the order in which we visit a node's children, but that would make tests harder to write.
Since validators don't favour particular collators when requesting collations, the order of potential forks in the fragment tree should already be somewhat "random" based on network latency (unless collators find a way to censor/DOS other collators)

Issue #3165, opened 2024-02-01
async_backing: Candidate timeouts on group rotation boundary in versi


alexggh:
When running tick and glutton parachains on versi with async-backing ever 2-3 minutes candidate timeout on backing and that blocks the core for about 2 min.

Some logs regarding 1 candidate: https://grafana.teleport.parity.io/goto/Nke60BpSR?orgId=1.

## Root cause

It seems that at group rotation boundary there is a problem in the way group assignment work and we end up in a situation where the collator and the backing subsystem end up using the group assignment before the rotation, but since the candidate is backed in a block after rotation the availability will use a different group for fetching the chunk which results in the candidate timing out.

The main culprit for this problem seems to be the backed assumption in runtime function validator_groups https://github.com/paritytech/polkadot-sdk/blob/8a8f6f98775a50758123ae760bb020f833354e59/polkadot/runtime/parachains/src/runtime_api_impl/v7.rs#L46 that when are backing a candidate is going to be included on chain next block, which is not the case with async_backing. 
```
	let now = <frame_system::Pallet<T>>::block_number() + One::one();

	let groups = <scheduler::Pallet<T>>::validator_groups();
	let rotation_info = <scheduler::Pallet<T>>::group_rotation_info(now);

	(groups, rotation_info)
```

Hence the usage of `validator_groups` in backing subsystem and `group_responsible` https://github.com/paritytech/polkadot-sdk/blob/8a8f6f98775a50758123ae760bb020f833354e59/polkadot/runtime/parachains/src/runtime_api_impl/v7.rs#L108 from CoreState in `availability-distribution` will give us different groups, so the candidate never passes the availability part.

@rphmeier: Thoughts on how to fix this ?

eskimor:
The backing time should be irrelevant for this. As I described, I think on the parathreads ticket for determining the backing group we have to use the relay parent of the candidate. If this is not the case, we need to fix that. I don't see how anything else can work.

eskimor:
Consequences on [this](https://github.com/paritytech/polkadot-sdk/issues/828#issuecomment-1691888251) comment.

alexggh:
> If this is not the case, we need to fix that

I confirmed that is not the case, reproducibility steps:

1. Start any zombienet with a config where async backing is enabled and a group rotation frequency lower than the session lenght which is 10 blocks on zombienet. E.g:
```
[relaychain.genesis.runtimeGenesis.patch.configuration.config]
[...]

PR #3170, opened 2024-02-01
Fixup backing group computation time

alexggh:
Fixes: https://github.com/paritytech/polkadot-sdk/issues/3165


alexggh:
- Tested on versi and confirmed this gets us 6s block times on a small network, see the dramatic improvement.
Screenshot 2024-02-01 at 17 32 19


- Looked at other places where backing group is used and it seems this is the only place where it wasn't fixed.
- Added a zombienet test that would have caught this issue.

ordian (approved):
Great work! ❤️

polkadot/zombienet_tests/functional/0011-async-backing-6-seconds-rate.zndsl
```diff
@@ -0,0 +1,20 @@
...
+alice: parachain 2001 is registered within 60 seconds
+
+# Ensure parachains made progress.
+alice: reports substrate_block_height{status="finalized"} is at least 10 within 100 seconds
+
+# This parachains should produce blocks at 6s clip, let's assume an 8s rate, allowing for
+# some slots to be missed on slower machines
+alice: parachain 2000 block height is at least 30 within 240 seconds
+# This should already have produced the needed blocks
+alice: parachain 2001 block height is at least 30 within 6 seconds
```
bkchr, line 19: Hmm. I see this already breaking :P 

Can we not ensure that we never see any candidate timeout? I mean on these test networks we should be relative sure that there are no timeouts.

bkchr: https://github.com/paritytech/polkadot-sdk/blob/0c3f029c35ad70f7f35b3ab889fbbf29fe92a8db/polkadot/node/core/av-store/src/lib.rs#L766-L775

We handle there the events already, we could add there a `candidate_timeout` event. Not really the best place, but maybe we could add it somewhere. Or we just scan every block with some piece of JS code for the events.

bkchr: https://polkadot.js.org/docs/api/examples/promise/system-events

polkadot/runtime/parachains/src/runtime_api_impl/v7.rs
```diff
@@ -95,6 +97,11 @@ pub fn availability_cores<T: initializer::Config>() -> Vec<CoreState<T::Hash, Bl
 						.expect("Occupied core always has pending availability; qed");
 
 				let backed_in_number = *pending_availability.backed_in_number();
+
+				// Use the same block number for determining the responsible group as what the
+				// backing subsystem would use when it calls validator_groups api.
+				let backing_group_allocation_time =
+					pending_availability.relay_parent_number() + One::one();
```
rphmeier, line 104: Yes, this is the correct fix!

rphmeier (approved):

AndreiEres (approved):

polkadot/zombienet_tests/functional/0011-async-backing-6-seconds-rate.zndsl
alexggh: > Hmm. I see this already breaking :P

I missed some environment variables for the CI, it is passing now.


> Can we not ensure that we never see any candidate timeout? I mean on these test networks we should be relative sure that there are no timeouts.

I'll look into how to add that assertion as well, as a follow up., opened this to track it:
https://github.com/paritytech/polkadot-sdk/issues/3180

bkchr: > I missed some environment variables for the CI, it is passing now.

I had only seen that it fails after writing this comment :P  What I meant is that this will for sure be flaky :D

alexggh: I'll keep an eye on eye on it  2 out of 2 for now.

To be honest I don't understand why it should be flaky producing blocks at 6s rate is one of the property of our system, I understand it can be worse than that if you don't have enough hw or it is contended, but in this case the test is really small so the CI hw should be enough and if it is not we need to make sure we use reasonably proper hw.

I do think it is valuable to be able to test the outside properties of our system rather than just testing no CandidateTimeout happens, that will catch just one reason we don't produce blocks at 6s, but won't catch all the other reasons we can't think of yet.

PR #3108, opened 2024-01-29
revert paritytech/polkadot#6577 & related changes

Lederstrumpf:
Moves `pallet_mmr` back behind `pallet_session` to address polkadot-fellows/runtimes#160.

Opening draft for CI - should be merged or closed depending on outcome of w3f/polkadot-spec#718.

PR #3171, opened 2024-02-01
Bump Rococo Runtime Spec Version

alistair-singh:
To deploy fix: https://github.com/paritytech/polkadot-sdk/pull/3108 and unblock the re-initialization of Ethereum bridge.

### DIFF
--- a/polkadot/runtime/rococo/src/lib.rs
+++ b/polkadot/runtime/rococo/src/lib.rs
@@ -150,7 +150,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("rococo"),
 	impl_name: create_runtime_str!("parity-rococo-v2.0"),
 	authoring_version: 0,
-	spec_version: 1_006_002,
+	spec_version: 1_006_003,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 24,


ggwpez (approved):

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -150,7 +150,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("rococo"),
 	impl_name: create_runtime_str!("parity-rococo-v2.0"),
 	authoring_version: 0,
-	spec_version: 1_006_002,
+	spec_version: 1_006_003,
```
bkontur, line 153: @EgorPopelyaev is this ok? or change to 1_700_000?  1.7.0 release was branched off yesterday, and if we need  to release BridgeHubRococo from master

ggwpez: 1.7 would be on the release branch then, not master.

acatangiu:
@alistair-singh what do you say we just wait for 1.7.0 release to come out then just use that?

in the meantime I believe you can re-initialize the Ethereum bridge without said fix - it'll be a problem only if we end up stalling BEEFY for multiple sessions before the 1.7.0 upgrade, which is unlikely

- the current BEEFY light-client can just work as it did before, making progress from non-mandatory commits
- once we upgrade to 1.7.0, it will also correctly accept mandatory commits

wdyt?

alistair-singh:
I am happy to wait for 1.7. We can close this PR. If a stall happens before 1.7 is rolled out we can always re-initialize again.

PR #3179, opened 2024-02-02
Initial support for building RISC-V runtimes targeting PolkaVM

koute:
This PR adds initial support for building RISC-V runtimes targeting PolkaVM.

- Setting the `SUBSTRATE_RUNTIME_TARGET=riscv` environment variable will now build a RISC-V runtime instead of a WASM runtime.
- This only adds support for *building* runtimes; running them will need a PolkaVM-based executor, which I will add in a future PR.
- Only building the minimal runtime is supported (building the Polkadot runtime doesn't work *yet* due to one of the dependencies).
- The builder now sets a `substrate_runtime` cfg flag when building the runtimes, with the idea being that instead of doing `#[cfg(not(feature = "std"))]` or `#[cfg(target_arch = "wasm32")]` to detect that we're building a runtime you'll do `#[cfg(substrate_runtime)]`. (Switching the whole codebase to use this will be done in a future PR; I deliberately didn't do this here to keep this PR minimal and reviewable.)
- Further renaming of things (e.g. types, environment variables and proc macro attributes having "wasm" in their name) to be target-agnostic will also be done in a future refactoring PR (while keeping backwards compatibility where it makes sense; I don't intend to break anyone's workflow or create unnecessary churn).
- This PR also fixes two bugs in the `wasm-builder` crate:
    * The `RUSTC` environment variable is now removed when invoking the compiler. This prevents the toolchain version from being overridden when called from a `build.rs` script.
    * When parsing the `rustup toolchain list` output the `(default)` is now properly stripped and not treated as part of the version.
- I've also added a minimal CI job that makes sure this doesn't break in the future. (cc @paritytech/ci)

cc @athei

------

Also, just a fun little tidbit: quickly comparing the size of the built runtimes it seems that the PolkaVM runtime is slightly smaller than the WASM one. (`production` build, with the `names` section substracted from the WASM's size to keep things fair, since for the PolkaVM runtime we're currently stripping out everything)

- `.wasm`: 625505 bytes
- `.wasm` (after wasm-opt -O3): 563205 bytes
- `.wasm` (after wasm-opt -Os): 562987 bytes
- `.wasm` (after wasm-opt -Oz): 536852 bytes
- `.polkavm`: ~~580338 bytes~~ 550476 bytes (after enabling extra target features; I'll add those in another PR once we have an executor working)

### DIFF
--- a/.gitlab/pipeline/build.yml
+++ b/.gitlab/pipeline/build.yml
@@ -329,6 +329,14 @@ build-linux-substrate:
     # - printf '\n# building node-template\n\n'
     # - ./scripts/ci/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
 
+build-minimal-runtime-polkavm:
+  stage: build
+  extends:
+    - .docker-env
+    - .common-refs
+  script:
+    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-node
+
 .build-subkey:
   stage: build
   extends:
--- a/.gitlab/pipeline/build.yml
+++ b/.gitlab/pipeline/build.yml
@@ -335,7 +335,7 @@ build-minimal-runtime-polkavm:
     - .docker-env
     - .common-refs
   script:
-    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-node
+    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-runtime
 
 .build-subkey:
   stage: build
--- a/substrate/primitives/api/Cargo.toml
+++ b/substrate/primitives/api/Cargo.toml
@@ -48,6 +48,7 @@ std = [
 	"sp-externalities?/std",
 	"sp-metadata-ir?/std",
 	"sp-runtime/std",
+	"sp-runtime-interface/std",
 	"sp-state-machine/std",
 	"sp-std/std",
 	"sp-test-primitives/std",
--- a/substrate/primitives/api/Cargo.toml
+++ b/substrate/primitives/api/Cargo.toml
@@ -28,7 +28,9 @@ sp-state-machine = { path = "../state-machine", default-features = false, option
 sp-trie = { path = "../trie", default-features = false, optional = true }
 hash-db = { version = "0.16.0", optional = true }
 thiserror = { version = "1.0.48", optional = true }
-scale-info = { version = "2.10.0", default-features = false, features = ["derive"] }
+scale-info = { version = "2.10.0", default-features = false, features = [
+  "derive",
+] }
 sp-metadata-ir = { path = "../metadata-ir", default-features = false, optional = true }
 log = { version = "0.4.17", default-features = false }
 
@@ -38,23 +40,23 @@ sp-test-primitives = { path = "../test-primitives" }
 [features]
 default = ["std"]
 std = [
-	"codec/std",
-	"hash-db",
-	"log/std",
-	"scale-info/std",
-	"sp-api-proc-macro/std",
-	"sp-core/std",
-	"sp-externalities",
-	"sp-externalities?/std",
-	"sp-metadata-ir?/std",
-	"sp-runtime/std",
-	"sp-runtime-interface/std",
-	"sp-state-machine/std",
-	"sp-std/std",
-	"sp-test-primitives/std",
-	"sp-trie/std",
-	"sp-version/std",
-	"thiserror",
+  "codec/std",
+  "hash-db",
+  "log/std",
+  "scale-info/std",
+  "sp-api-proc-macro/std",
+  "sp-core/std",
+  "sp-externalities",
+  "sp-externalities?/std",
+  "sp-metadata-ir?/std",
+  "sp-runtime/std",
+  "sp-runtime-interface/std",
+  "sp-state-machine/std",
+  "sp-std/std",
+  "sp-test-primitives/std",
+  "sp-trie/std",
+  "sp-version/std",
+  "thiserror",
 ]
 # Special feature to disable logging completely.
 #
--- a/substrate/primitives/api/Cargo.toml
+++ b/substrate/primitives/api/Cargo.toml
@@ -21,6 +21,7 @@ sp-api-proc-macro = { path = "proc-macro", default-features = false }
 sp-core = { path = "../core", default-features = false }
 sp-std = { path = "../std", default-features = false }
 sp-runtime = { path = "../runtime", default-features = false }
+sp-runtime-interface = { path = "../runtime-interface", default-features = false }
 sp-externalities = { path = "../externalities", default-features = false, optional = true }
 sp-version = { path = "../version", default-features = false }
 sp-state-machine = { path = "../state-machine", default-features = false, optional = true }
--- a/substrate/primitives/api/Cargo.toml
+++ b/substrate/primitives/api/Cargo.toml
@@ -29,7 +29,7 @@ sp-trie = { path = "../trie", default-features = false, optional = true }
 hash-db = { version = "0.16.0", optional = true }
 thiserror = { version = "1.0.48", optional = true }
 scale-info = { version = "2.10.0", default-features = false, features = [
-  "derive",
+	"derive",
 ] }
 sp-metadata-ir = { path = "../metadata-ir", default-features = false, optional = true }
 log = { version = "0.4.17", default-features = false }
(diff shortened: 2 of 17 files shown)


.gitlab/pipeline/build.yml
```diff
@@ -329,6 +329,14 @@ build-linux-substrate:
     # - printf '\n# building node-template\n\n'
     # - ./scripts/ci/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
 
+build-minimal-runtime-polkavm:
+  stage: build
+  extends:
+    - .docker-env
+    - .common-refs
+  script:
+    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-node
```
bkchr, line 338:
```suggestion
    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-runtime
```
Not sure why you want to check the entire node :D

substrate/primitives/api/Cargo.toml
```diff
@@ -21,13 +21,16 @@ sp-api-proc-macro = { path = "proc-macro", default-features = false }
 sp-core = { path = "../core", default-features = false }
 sp-std = { path = "../std", default-features = false }
 sp-runtime = { path = "../runtime", default-features = false }
+sp-runtime-interface = { path = "../runtime-interface", default-features = false }
```
bkchr, line 24: I think we should directly import polkavm here. No need to pull in the entire `sp-runtime-interface`.

substrate/primitives/runtime-interface/src/polkavm.rs
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub use polkavm_derive::{polkavm_export, polkavm_import};
+
+#[polkavm_derive::polkavm_define_abi(allow_extra_input_registers)]
+pub mod polkavm_abi {}
```
bkchr, line 21: What does that do? :D

substrate/utils/wasm-builder/src/lib.rs
```diff
@@ -365,5 +405,51 @@ fn get_bool_environment_variable(name: &str) -> Option<bool> {
 
 /// Returns whether we need to also compile the standard library when compiling the runtime.
 fn build_std_required() -> bool {
-	crate::get_bool_environment_variable(crate::WASM_BUILD_STD).unwrap_or(true)
+	let default = match runtime_target() {
+		RuntimeTarget::Wasm => true,
+		RuntimeTarget::Riscv => false,
+	};
```
bkchr, line 411:
```suggestion
	let default = runtime_target() == RuntimeTarget::Wasm;
```

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -120,25 +121,71 @@ pub(crate) fn create_and_compile(
...
+			project_cargo_toml,
+			&project,
+			bloaty_blob_out_name_override,
+			check_for_runtime_version_section,
+			&build_config,
+		),
+		RuntimeTarget::Riscv => {
+			let out_name = bloaty_blob_out_name_override
+				.unwrap_or_else(|| get_blob_name(target, project_cargo_toml));
+			let out_path = project.join(format!("{}.polkavm", out_name));
```
bkchr, line 159:
```suggestion
			let out_path = project.join(format!("{out_name}.polkavm"));
```

bkchr (approved):
Nice work!

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -711,22 +764,34 @@ fn offline_build() -> bool {
...
-	);
+	let mut rustflags = String::new();
+	match target {
+		RuntimeTarget::Wasm => {
+			rustflags.push_str(
+				"-C target-cpu=mvp -C target-feature=-sign-ext -C link-arg=--export-table ",
+			);
+		},
+		RuntimeTarget::Riscv => {
+			rustflags.push_str("-C target-feature=+lui-addi-fusion -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique ");
```
athei, line 784: I was hitting duplicate symbol name when implementing PolkaVM support for ink!. Essentially the `call` export was having problems. Is the `--unique` a solution for that?

athei (approved):

substrate/primitives/runtime-interface/src/polkavm.rs
koute: It's a magic macro. (;

Basically, it defines the trait machinery required to automatically convert types on an FFI boundary. It essentially generates this:

```rust
pub mod polkavm_abi {
    pub trait IntoHost {
        type Regs;
        type Destructor;

        fn into_host(value: Self) -> (Self::Regs, Self::Destructor);
    }

    pub trait FromHost {
        type Regs;
        fn from_host(value: Self::Regs) -> Self;
    }

    #[doc(hidden)]
    pub mod private {
        /* ...more stuff... */
    }
}
```

The idea here is that you can define how the types are marshaled through the FFI boundary in one place, and then when you use the `polkavm_import` or `polkavm_export` macro you can give it a path to this module, and it will use the `IntoHost` and `FromHost` traits defined within to do the conversions. For example, something like this:

```rust
#[polkavm_derive::polkavm_define_abi]
pub mod my_abi {}

impl self::my_abi::FromHost for FancyType {
    type Regs = (u32, u32,); // It needs two registers to be marshaled through the FFI boundary.
    fn from_host(value: Self::Regs) -> Self { ... }
}

#[polkavm_derive::polkavm_import(abi = my_abi)]
extern {
    fn hostcall() -> FancyType;
}

fn get_fancy() {
    let fancy = unsafe { hostcall(); }
}
```

This is somewhat similar to the traits we already have in Substrate for converting into/from WASM FFI types, except it's part of the upstream instead of being Substrate-specific, and you can have multiple ABIs defined (e.g. you could have one for Substrate runtimes and another for smart contracts, and they won't conflict with each other).

substrate/utils/wasm-builder/src/wasm_project.rs
koute: No, the `--unique` just allows LLD to emit multiple sections, which allows our linker to better optimize the blob in some cases.

Your problems with duplicate symbols are probably fixed in the newest version of the PolkaVM derive crate. With each successive version I've been coming up with less hacky ways to do imports/exports, and the way they're implemented right now manages to exclusively use `rustc` to generate the symbol names for us so there should not be any dupes anymore.

substrate/primitives/api/Cargo.toml
koute: The problem here is that this needs access to the `polkavm_abi` module we've defined in `sp-runtime-interface`, so just directly including the PolkaVM crates is not sufficient.

So I can think of three options here:
1) Define the `polkavm_abi` module in `sp-runtime-interface` and depend on it here.
2) Define the `polkavm_abi` in `sp-wasm-interface` (which actually isn't WASM specific, since PolkaVM support also uses what's defined in there because it reuses parts of the existing machinery to avoid having to rewrite everything, *and* the `wasmtime` feature for that crate is optional anyway; a better name for that crate would be, uh, `sp-runtime-interface`, but that's already taken, so maybe merging these two crates would be the way to go :P)
3) Define the `polkavm_abi` in a new crate, say, `sp-polkavm-interface`, and depend on it here. (But that seems like an overkill to me to add a new crate for ~10 lines of code, and also because the stuff in `sp-wasm-interface` is also used it'd be weird to have WASM only use `sp-wasm-interface` but PolkaVM use both `sp-polkavm-interface` and `sp-wasm-interface`.)

What I would suggest we do here is:

1) Leave the `polkavm_abi` definition in `sp-runtime-interface` and depend on it here.
2) (In another PR) Remove the `sp-wasm-interface` crate and merge it with `sp-runtime-interface`. Then refactor `sp-runtime-interface` so that it's more modularized and you don't need to always pull in the whole thing.

Does this sound good to you? (Alternatively maybe you have a better idea?)

bkchr: Ahh okay, yeah then fine to leave it the way you have done it right now.

.gitlab/pipeline/build.yml
```diff
@@ -329,6 +329,14 @@ build-linux-substrate:
     # - printf '\n# building node-template\n\n'
     # - ./scripts/ci/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
 
+build-minimal-runtime-polkavm:
+  stage: build
+  extends:
+    - .docker-env
+    - .common-refs
+  script:
+    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-runtime
```
ggwpez, line 338: Btw why does it only run check and not build?  
I think there are possible errors that only arise on `build`.

bkchr: The builder runs as part of the `build.rs`. It can not detect if you are running `cargo check` and thus, it will always compile. So, everything works here as expected.

PR #3186, opened 2024-02-02
Removes Snowbridge parachain directory 

claravanstaden:
Removes the `bridges/snowbridge/parachain` directory and moves everything up to under `snowbridge` directly. We are cleaning up our local dev env after merging our crates into the polkadot-sdk.


bkontur (approved):

ggwpez (approved):
Maybe we can add review rules for this subfolder, since CODEOWNERS does not work for non-members of the same org.

acatangiu (approved):

PR #3188, opened 2024-02-02
Small fixes coretime related fixes

tdimitrov:
The PR contains small fixes for:
* A test `HostConfig v11` storage migration - v11  was compared with itself instead of with `v10`.
* Outdated comment for `ClaimQueue`
* Typos

### DIFF
--- a/polkadot/runtime/parachains/src/configuration/migration/v11.rs
+++ b/polkadot/runtime/parachains/src/configuration/migration/v11.rs
@@ -253,7 +253,7 @@ mod tests {
 
 		new_test_ext(Default::default()).execute_with(|| {
 			// Implant the v10 version in the state.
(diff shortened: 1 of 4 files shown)


eskimor (approved):

polkadot/runtime/parachains/src/configuration/migration/v11.rs
```diff
@@ -264,7 +264,7 @@ mod tests {
 			let mut configs_to_check = v11::PendingConfigs::<Test>::get().unwrap();
 			configs_to_check.push((0, v11.clone()));
 
-			for (_, v10) in configs_to_check {
+			for (_, v11) in configs_to_check {
```
eskimor, line 267: good catch. Maybe we should move the v11 above into its own scope {} with that assert.

bkchr (approved):

PR #3191, opened 2024-02-02
subsystem-bench: add regressions guard

AndreiEres:
Added a guard binary to check possible regression in subsystem benchmarks

Example run:
```
target/debug/subsystem-bench-guard --current=out.yaml --reference=ref.yaml --threshold=1

polkadot/node/subsystem-bench/examples/availability_write.yaml #1

Total network usage, s
Received from peers                 9930.000   +1.0%
Sent to peers                      47700.000   -0.2%

Total CPU usage, s
availability-distribution              0.045  -30.9%
bitfield-distribution                  0.102   +0.0%
availability-store                     0.305   +0.0%
Test environment                       3.302   +0.0%

Error:
   0: Found regressions more then 1%
```

### DIFF
--- a/polkadot/node/subsystem-bench/Cargo.toml
+++ b/polkadot/node/subsystem-bench/Cargo.toml
@@ -12,6 +12,10 @@ publish = false
 name = "subsystem-bench"
 path = "src/subsystem-bench.rs"
 
+[[bin]]
+name = "subsystem-bench-guard"
+path = "src/subsystem-bench-guard.rs"
+
 # Prevent rustdoc error. Already documented from top-level Cargo.toml.
 doc = false
 
--- a/polkadot/node/subsystem-bench/src/subsystem-bench-guard.rs
+++ b/polkadot/node/subsystem-bench/src/subsystem-bench-guard.rs
@@ -1,0 +1,160 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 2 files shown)


sandreim (approved):
I was hoping to have the regression tests written as separate test objectives. That would allow more flexibility when implementing a specific test. For example we might want to assert on some subsystem metrics as well.  Additionally this avoids the need to have reference files in git.

sandreim:
I accidentally approved while my intention was to have more discussion 😁

AndreiEres:
Here are my thoughts on the approaches. I think each of them has merit. My only concern is how often we need to update the reference files.

### Approach # 1
To store reference benchmark results in project files

**Pros**

- We can see the impact immediately, but... (see cons)
- We avoid possible fluctuations in the reference results: they are constant until we change them.
- We don't have to set up additional logic in CI, we get results just by running a benchmark on our PR.

**Cons**

- The immediate impact that we can see only makes sense on the reference hardware. To make it right for a local machine, we should replace the reference files with our own.
- We have to keep track of the latest changes in the repo and manually update the files every time something changes. I'm not sure how often this can be done, since the Polkadot's performance can be affected by work done on the Substrate etc.

### Approach # 2
To save last benchmarks run on the master branch as a reference file

**Pros**

- In the CI results, we always see an up-to-date comparison to the master branch.
- We don't have to bother about updates, they are done automatically

Cons
- We can't see the impact until a corresponding job finishes in CI
- Every run has slightly different results, which can affect the success of the CI job
- We need additional logic in CI (frankly it's not hard and only a one-time job)

alexggh:
> I was hoping to have the regression tests written as separate test objectives. That would allow more flexibility when implementing a specific test. For example we might want to assert on some subsystem metrics as well. Additionally this avoids the need to have reference files in git.

Do you mean create a separate objective in that case our configs would look like:
```
TestConfiguration:
# Test 1
- objective: !DataAvailabilityRead
   .....

# Test 2
- objective: !DataAvailabilityRead
   .....

# Test 3
- objective: !DataAvailabilityRead
  .....

- objective: !AssertMetrics {
    "metrics_name": value.
  }
```

Or add a separate field that lives alongside each objective something like:
```
# Test 1
- objective: !DataAvailabilityRead
    n_validators: 300
    n_cores: 20
    assertions: 
      - "metric_name": value.

```

I would go for the latter since I imagine each objective would have their own specific set of assertions.

Issue #3084, opened 2024-01-26
[RPC-Spec-V2] Extend testing for transaction_unstable_broadcast


lexnv:
The https://github.com/paritytech/polkadot-sdk/pull/3079 introduces a test harness for the transaction class.

Extend the tests with various scenarios to validate the behavior of transaction_unstable_broadcast.

- To easily correlate the transaction pool status with the status of the `transaction_unstable_broadcast`, introduce a zero-cost abstraction middleware to capture `TxStatus` events for the testing environment.
- Add the `CheckMortality` transaction extension to the substrate test extrinsics
- Construct a transaction that will become valid in a future block and ensure the broadcast function tries to resubmit the transaction at a later block
  - The state captured by the middleware should be (first block) `Invalid`, (from valid block:) `Ready`, `InBlock`, `Finalized`
- Add other testing scenarios
  

cc @paritytech/subxt-team

PR #3193, opened 2024-02-02
rpc-v2/tx/tests: Add transaction broadcast tests and check propagated tx status

lexnv:
This PR adds tests for the `transaction_broadcast` method.

The testing needs to coordinate the following components:
- The `TestApi` marks transactions as invalid and implements `ChainApi::validate_transaction`
   - this is what dictates if a transaction is valid or not and is called from within the `BasicPool`
- The `BasicPool` which maintains the transactions and implements `submit_and_watch` needed by the tx broadcast to submit the transaction
  - The status of the transaction pool is exposed by mocking the BasicPool
- The `ChainHeadMockClient` which mocks the `BlockchainEvents::import_notification_stream` needed by the tx broadcast to know to which blocks the transaction is submitted

The following changes have been added to the substrate testing to accommodate this:
- `TestApi` gets ` remove_invalid`, counterpart to `add_invalid` to ensure an invalid transaction can become valid again; as well as a priority setter for extrinsics
- `BasicPool` test constructor is extended with options for the `PoolRotator`
  - this mechanism is needed because transactions are banned for 30mins (default) after they are declared invalid
  - testing bypasses this by providing a `Duration::ZERO`

### Testing Scenarios

- Capture the status of the transaction as it is normally broadcasted
- `transaction_stop` is valid while the transaction is in progress
- A future transaction is handled when the dependencies are completed
- Try to resubmit the transaction at a later block (currently invalid)
  - An invalid transaction status is propagated; the transaction is marked as temporarily banned; then the ban expires and transaction is resubmitted
  
This builds on top of: https://github.com/paritytech/polkadot-sdk/pull/3079
Part of: https://github.com/paritytech/polkadot-sdk/issues/3084

cc @paritytech/subxt-team


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5112217

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5124629

substrate/client/rpc-spec-v2/src/transaction/tests/setup.rs
```diff
@@ -0,0 +1,120 @@
...
+		RpcTransactionBroadcast::new(client_mock.clone(), pool.clone(), Arc::new(task_executor))
+			.into_rpc();
+
+	(api, pool, client_mock, tx_api, executor_recv, pool_recv)
+}
+
+/// Get the next event from the provided middleware in at most 60 seconds.
+macro_rules! get_next_event {
+	($middleware:expr) => {
+		tokio::time::timeout(std::time::Duration::from_secs(60), $middleware.recv())
```
skunert, line 98: Do we expect to ever get close to 60 seconds here? Can't we fail much faster?

skunert (approved):

substrate/client/rpc-spec-v2/src/transaction/tests/setup.rs
lexnv: We should get the response almost immediately. I'll bring this value down a bit :D

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
```diff
@@ -0,0 +1,518 @@
...
+
+	// The future broadcast awaits for the finalized status to be reached.
+	// Force the future to exit by calling stop.
+	let _: () = tx_api
+		.call("transaction_unstable_stop", rpc_params![&operation_id])
+		.await
+		.unwrap();
+
+	// Ensure the broadcast future finishes.
+	let _ = get_next_event!(&mut exec_middleware);
```
jsdw, line 91: Does this ensure it finishes? Could there be more than 1 more event here?

I'm wondering whether asserting that the response is `None` or something might be more descriptive?

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
```diff
@@ -0,0 +1,518 @@
...
+		.call::<_, serde_json::Value>("transaction_unstable_stop", rpc_params![&operation_id])
+		.await
+		.unwrap_err();
+	assert_matches!(err,
+		Error::JsonRpc(err) if err.code() == json_rpc_spec::INVALID_PARAM_ERROR && err.message() == "Invalid operation id"
+	);
+}
+
+#[tokio::test]
+async fn tx_invalid_stop() {
```
jsdw, line 134:
```suggestion
async fn tx_stop_with_invalid_operation_id() {
```

Maybe?

substrate/test-utils/runtime/transaction-pool/src/lib.rs
```diff
@@ -214,6 +215,22 @@ impl TestApi {
 		self.chain.write().invalid_hashes.insert(Self::hash_and_length_inner(xts).0);
 	}
 
+	/// Remove a transaction that was prior declared as invalid via `[Self::add_invalid]`.
```
jsdw, line 218:
```suggestion
	/// Remove a transaction that was previously declared as invalid via `[Self::add_invalid]`.
```

jsdw (approved):
Just a couple of small nits!

I find a couple of the tests quite long and hard to follow (I wonder if it's possible to make them any smaller in general), but I don't know this code very well, and offhand they look useful!

substrate/client/rpc-spec-v2/src/transaction/tests/setup.rs
```diff
@@ -0,0 +1,120 @@
...
+			match event {
+				crate::transaction::tests::middleware_pool::MiddlewarePoolEvent::TransactionStatus { transaction, status } => {
+                    events.entry(transaction).or_insert_with(|| vec![]).push(status);
+				},
+				other => panic!("Expected TransactionStatus, received {:?}", other),
+			};
+		}
+		events
+	}};
+}
```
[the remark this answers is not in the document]
lexnv: This gives a more useful line number on which the panic happened. The alternative of having a function would always result in a panic at like 100 for example, inside the function :D

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
lexnv: I've added a way to check the active number of tasks, thanks!

PR #3209, opened 2024-02-05
Build more runtimes targeting PolkaVM

koute:
This PR improves compatibility with RISC-V and PolkaVM, allowing more runtimes to successfully compile.

In particular, it makes the following changes:

- The `sp-mmr-primitives` and `sp-consensus-beefy` crates unconditionally required an `std`-only dependency; now they only require those dependencies when the `std` feature is actually enabled. (Our RISC-V target is, unlike WASM, a true `no_std` target where you can't accidentally use stuff from `std` anymore.)
- One of our dependencies (the `bitvec` trace) uses a crate called `radium` which doesn't compile under RISC-V due to incomplete autodetection logic in their `build.rs` file. The good news is that this is already fixed in the newest upstream version of `radium`, and the newest version of `bitvec` uses it. The bad news is that the newest version of `bitvec` is not currently released on crates.io, so we can't use it. I've [created an issue](https://github.com/ferrilab/ferrilab/issues/5) asking for a new release, but in the meantime I forked the currently used `radium` 0.7, [fixed the faulty logic](https://github.com/paritytech/radium-0.7-fork/commit/ed66c8a294b138c67f93499644051d97d4c7fbda) and used cargo's patching capabilities to use it for the RISC-V runtime builds. This might be a little hacky, but it is the least intrusive way to fix the problem, doesn't affect WASM builds at all, and we can trivially remove it once a new `bitvec` is released.
- The new runtimes are added to the CI to make sure their compilation doesn't break.

### DIFF
--- a/.gitlab/pipeline/build.yml
+++ b/.gitlab/pipeline/build.yml
@@ -329,13 +329,16 @@ build-linux-substrate:
     # - printf '\n# building node-template\n\n'
     # - ./scripts/ci/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
 
-build-minimal-runtime-polkavm:
+build-runtimes-polkavm:
   stage: build
   extends:
     - .docker-env
     - .common-refs
   script:
     - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p minimal-runtime
+    - SUBSTRATE_RUNTIME_TARGET=riscv cargo check -p westend-runtime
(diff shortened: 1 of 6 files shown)


michalkucharczyk (approved):

davxy (approved):

andresilva (approved):
maybe wasm-builder should be renamed to runtime-builder, or something like that, at some point

koute:
> maybe wasm-builder should be renamed to runtime-builder, or something like that, at some point

Yes, I'm planning to rename a bunch of stuff in a future refactoring PR.

PR #3213, opened 2024-02-05
[pallet_broker] Remove leases that have already expired in rotate_sale

seadanda:
Leases can be force set, but since `Leases` is a `StorageValue`, if a lease misses its sale rotation in which it should expire, it can never be cleared.

This can happen if a lease is added with an `until` timeslice that lies in a region whose sale has already started or has passed, even if the timeslice itself hasn't passed.

This solves that issue in a minimal way, with all expired leases being cleaned up in each sale rotation, not just the ones that are expiring in the coming region.

TODO:
- [x] Write test

### DIFF
--- a/substrate/frame/broker/src/tests.rs
+++ b/substrate/frame/broker/src/tests.rs
@@ -864,6 +864,29 @@ fn cannot_set_expired_lease() {
 }
 
 #[test]
+fn short_leases_are_cleaned() {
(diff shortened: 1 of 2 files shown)


bkchr (approved):
Please add a test before merging.

georgepisaltu (approved):

PR #3215, opened 2024-02-05
transaction-pool: Improve transaction status documentation and add helpers

lexnv:
This PR improves the transaction status documentation.
- Added doc references for describing the main states
- Extra comment wrt pool ready / future queues
- `FinalityTimeout` no longer describes a lagging finality gadget, it signals that the maximum number of finality gadgets has been reached 

A few helper methods are added to indicate when:
- a final event is generated by the transaction pool for a given event
- a final event is provided, although the transaction might become valid at a later time and could be re-submitted

The helper methods are used and taken from https://github.com/paritytech/polkadot-sdk/pull/3079 to help us better keep it in sync.

cc @paritytech/subxt-team


bkchr (approved):

substrate/client/transaction-pool/api/src/lib.rs
```diff
@@ -88,19 +95,32 @@ impl PoolStatus {
...
+/// such case). An `Invalid` transaction may re-enter the pool only if it is resubmitted.
 /// 4. `Retracted` transactions might be included in some next block.
 ///
-/// The stream is considered finished only when either `Finalized` or `FinalityTimeout`
-/// event is triggered. You are however free to unsubscribe from notifications at any point.
-/// The first one will be emitted when the block, in which transaction was included gets
-/// finalized. The `FinalityTimeout` event will be emitted when the block did not reach finality
-/// within 512 blocks. This either indicates that finality is not available for your chain,
-/// or that finality gadget is lagging behind. If you choose to wait for finality longer, you can
-/// re-subscribe for a particular transaction hash manually again.
```
michalkucharczyk, line 103: I think description of FinalityTimeout is worth keeping. It reflects current implementation.

I am not sure if transaction is retriable when finality timeout event was reported. It could be the case, but it also mean that transaction was actually finalized. Maybe it is worth putting more info on this.

michalkucharczyk (approved):

skunert (approved):

PR #3222, opened 2024-02-06
Add try_state and integrity_test to XCM simulator fuzzer

louismerlin:
This adds `try_state()` and `integrity_test()` to the four runtimes of the XCM-simulator fuzzer.

With this, we are able to stress-test [message-queue's try_state](https://github.com/paritytech/polkadot-sdk/blob/7df1ae3b8111d534cce108b2b405b6a33fcdedc3/substrate/frame/message-queue/src/lib.rs#L1245-L1347).

This also adds the `Transact` block-listing from #2424 to avoid false-positives.

Thank you @ggwpez for the help with the runtime configurations.


polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
```diff
@@ -155,6 +158,34 @@ fn run_input(xcm_messages: [XcmMessage; 5]) {
 	println!();
 
 	for xcm_message in xcm_messages {
+		fn matches_blocklisted_messages(message: Instruction<()>) -> bool {
+			matches!(message, Transact { .. })
+		}
+		// We check XCM messages recursively for blocklisted messages
+		fn matches_recursive(message: &mut Instruction<()>) -> Vec<Instruction<()>> {
```
ggwpez, line 165: Can you pull these two functions out maybe?

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
```diff
@@ -155,6 +158,34 @@ fn run_input(xcm_messages: [XcmMessage; 5]) {
...
+			match message {
+				DepositReserveAsset { xcm, .. } |
+				ExportMessage { xcm, .. } |
+				InitiateReserveWithdraw { xcm, .. } |
+				InitiateTeleport { xcm, .. } |
+				TransferReserveAsset { xcm, .. } |
+				SetErrorHandler(xcm) |
+				SetAppendix(xcm) => Vec::from(xcm.inner()).iter_mut().flat_map(matches_recursive).collect(),
+				_ => vec![message.clone()],
```
ggwpez, line 174: Hm, this function is called `matches_recursive`, but it dies not call the `matches_blocklisted_messages`?  
To me it looks like it unpacks them.

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
```diff
@@ -197,8 +228,20 @@ fn run_input(xcm_messages: [XcmMessage; 5]) {
 		}
 		#[cfg(not(fuzzing))]
 		println!();
+		// We run integrity tests and try_runtime invariants
+		[ParaA::execute_with, ParaB::execute_with, ParaC::execute_with].iter().for_each(
+			|execute_with| {
+				execute_with(|| {
+					parachain::AllPalletsWithSystem::try_state(Default::default(), All).unwrap();
```
ggwpez, line 235: `All` also tries to decode the storage every time, it should be very slow once the state gets bigger.  
Maybe its enough to call this only every 100th time or so, depending on performance.

polkadot/xcm/xcm-simulator/fuzzer/src/parachain.rs
```diff
@@ -60,9 +84,9 @@ impl frame_system::Config for Runtime {
 	type RuntimeCall = RuntimeCall;
 	type Nonce = u64;
 	type Hash = H256;
-	type Hashing = ::sp_runtime::traits::BlakeTwo256;
+	type Hashing = BlakeTwo256;
 	type AccountId = AccountId;
-	type Lookup = IdentityLookup<Self::AccountId>;
+	type Lookup = AccountIdLookup<AccountId, ()>;
```
ggwpez, line 89: Should be possibly to only write:  
```rust
	type AccountId = AccountId;
	type Lookup = AccountIdLookup<AccountId, ()>;
	type Block = Block;
	type AccountData = pallet_balances::AccountData<Balance>;
```
Since the others get pulled in from the `derive_impl`. Probably need to set `BlockNumber` to `u64`.

polkadot/xcm/xcm-simulator/fuzzer/src/parachain.rs
```diff
@@ -367,3 +389,60 @@ construct_runtime!(
...
+#[sp_version::runtime_version]
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+	spec_name: create_runtime_str!("node-template"),
+	impl_name: create_runtime_str!("node-template"),
+	authoring_version: 1,
+	// The version of the runtime specification. A full node will not attempt to use its native
+	//   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
+	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
+	// This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
+	//   the compatible custom types.
```
ggwpez, line 404: I think you can delete these comments 😆 Maybe there is also a default or a mock for testing.

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
```diff
@@ -56,9 +83,9 @@ impl frame_system::Config for Runtime {
 	type RuntimeCall = RuntimeCall;
 	type Nonce = u64;
 	type Hash = H256;
-	type Hashing = ::sp_runtime::traits::BlakeTwo256;
+	type Hashing = BlakeTwo256;
 	type AccountId = AccountId;
-	type Lookup = IdentityLookup<Self::AccountId>;
+	type Lookup = sp_runtime::traits::AccountIdLookup<AccountId, ()>;
```
ggwpez, line 88: ditto

ggwpez (commented):
Thanks! Left some nits, and then afterwards please look for the CI.

ggwpez:
It looks like the `cargo test` does not work on MacOS, probably because of how the fuzzer works. Maybe need to mark these tests as `ignore`.

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
louismerlin: Should be addressed

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
louismerlin: Should be addressed

polkadot/xcm/xcm-simulator/fuzzer/src/parachain.rs
louismerlin: Good catch, done

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
louismerlin: Done

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
louismerlin: To have optimal results, a fuzzer needs stability (each input should yield the same code coverage every time they are run). It would probably be counter-productive to do some kind of modulo 100 on try_state.

The harness is fast and the coverage is good, I don't think this is an issue :)

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
```diff
@@ -148,13 +151,39 @@ pub fn relay_ext() -> sp_io::TestExternalities {
...
+fn recursively_matches_blocklisted_messages(message: &mut Instruction<()>) -> bool {
+	match message {
+		DepositReserveAsset { xcm, .. } |
+		ExportMessage { xcm, .. } |
+		InitiateReserveWithdraw { xcm, .. } |
+		InitiateTeleport { xcm, .. } |
+		TransferReserveAsset { xcm, .. } |
+		SetErrorHandler(xcm) |
+		SetAppendix(xcm) =>
+			Vec::from(xcm.inner()).iter_mut().any(recursively_matches_blocklisted_messages),
```
ggwpez, line 164: You could still use iterators to avoid any clones, but its fine 🙏

ggwpez (approved):

polkadot/xcm/xcm-simulator/fuzzer/src/fuzz.rs
louismerlin: Good point, addressed in 11302c9a139ac04c687dfce8cb5c4104d67f4bec

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
```diff
@@ -259,3 +265,53 @@ construct_runtime!(
 		MessageQueue: pallet_message_queue,
 	}
 );
+
+#[sp_version::runtime_version]
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+	spec_name: create_runtime_str!("node-template"),
```
kianenigma, line 271: this doesn't sound like the right name to me. 

```suggestion
	spec_name: create_runtime_str!("xcm-simulator-node"),
```

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
```diff
@@ -259,3 +265,53 @@ construct_runtime!(
 		MessageQueue: pallet_message_queue,
 	}
 );
+
+#[sp_version::runtime_version]
```
kianenigma, line 269: I don't see anywhere else in this PR that this runtime apis are needed. It seems like you just want to call into the methods exposed from `Executive`. Are you sure you need this?

kianenigma (changes requested):
I might be wrong because I don't know the context, but I think the `impl_runtime_apis` is not needed.

louismerlin:
> I might be wrong because I don't know the context, but I think the `impl_runtime_apis` is not needed.

Just verified with coverage, and the fuzzer is able to access `try_runtime` without `impl_runtime_apis`. Thank you very much for catching this!

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
louismerlin: This was removed in 3281a1f.

polkadot/xcm/xcm-simulator/fuzzer/src/relay_chain.rs
louismerlin: This was removed in 3281a1f.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5138091

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5138231

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5138266

kianenigma (approved):

polkadot/xcm/xcm-simulator/fuzzer/Cargo.toml
```diff
@@ -35,6 +37,18 @@ polkadot-runtime-parachains = { path = "../../../runtime/parachains" }
 polkadot-parachain-primitives = { path = "../../../parachain" }
 
 [features]
+default = ["try-runtime"]
```
ggwpez, line 40:
```suggestion
```

louismerlin: 4b1e9cf43646ec3943f495407463adf4ea5a5c4d

Issue #2346, opened 2023-11-15
statement-distribution: v2: Reduce reputation CostMinor("Unexpected Statement")


alexggh:
While investigating https://github.com/paritytech/polkadot-sdk/issues/1426 I observed that  we have the following log on versi:
```
Reduce reputation peer=PeerId("12D3KooWP2LGsamcPoNV25GsH3FVUzGmYsTtDkop7tbmvREppqyU") rep=CostMinor("Unexpected Statement")
```

## Root cause
The log is emitted from here https://github.com/paritytech/polkadot-sdk/blob/5b0622bc4d0cf5f0f57904874ea90eb08de79c52/polkadot/node/network/statement-distribution/src/v2/mod.rs#L1549 caused by the fact that there seems to be a race between two nodes that sent each other the same message around the same time.

So, in the moment  nodes process what they received `direct_statement_providers` won't return anything because it does not expect to receive from a node that it sent the message to.

## Severity

The issue is mostly benign because all we do is just a minor reputation update, which would definitely be offsetted by all the valid messages we receive from a given node, so I don't expect it creates any issue.

Tried some naive fixes here, but they didn't completely solve the issue: https://github.com/paritytech/polkadot-sdk/pull/2267, so more time would need to be invested for a proper handling of this edge case.

PR #3223, opened 2024-02-06
statement-distribution: Fix CostMinor("Unexpected Statement")

alexggh:
On grid distribution messages have two paths of reaching a node, so there is the possiblity of a race when two peers send each other the same statement around the same time. Statement local_knowledge will tell us that the peer should have not send the statement because we sent it to it.

Fix it by also keeping track only of the statement we received from a given peer and penalize it only if it sends it to us more than once.

Fixes: https://github.com/paritytech/polkadot-sdk/issues/2346

Additionally, also use different Cost labels for different paths to make it easier to debug things.


polkadot/node/network/statement-distribution/src/v2/grid.rs
```diff
@@ -908,6 +913,14 @@ struct MutualKnowledge {
 	/// `Some` only if we have advertised, acknowledged, or requested the candidate
 	/// from them.
 	local_knowledge: Option<StatementFilter>,
+	/// Knowledge peer sent to us, this is different from
+	/// `local_knowledge` and `remote_knowledge`, through the fact that includes only
```
sandreim, line 917: I think this is not clear enough. Just by reading these comments, `remote_knowledge` gets updated when remote peers have advertised, acknowledged, or requested the candidate, but  `received_knowledge` also does the same as per comment.

polkadot/node/network/statement-distribution/src/v2/grid.rs
```diff
@@ -908,6 +913,14 @@ struct MutualKnowledge {
...
+	/// statements that we received from peer while the other two include knowledge from
+	/// what we have sent to the peer and he sent back.
+	/// This is needed because sometimes we have races where peers send us statements that
```
sandreim, line 920: I might be missing something, but .... Is there any reason we cannot handle the race by accepting statements for candidates we have already sent, but just once. That can be as easy setting only `local_knowledge` to `Some` when have advertised, acknowledged, or requested the candidate from the peer, but not the `remote_knowledge` that we'd update once the they advertise it once.

alexggh: Let me describe the flow a bit better:
1. When we send the manifest to a peer we set `local_knowledge` and that represents all the statements that we already have and that we are going to send to peer.
2. When receive the manifest we set `remote_knowledge` and that represents all the statements that peer already has.


Now, after this step we can circulate new statements we received from peers and we do that and we call `sent_or_received_direct_statement` which set both the local knowledge and the remote knowledge see here: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/network/statement-distribution/src/v2/grid.rs#L1020, hence why local_knowledge will be set to true and we have no way to know if the peer sent us the message 0, 1 or 2 times.

Now a solution would be probably to try and break sent_or_received_direct_statement in 2 parts `sent` and `received` however that would be a more intrusive solution and would be easy to get things wrong while doing, which I didn't think was worth the effort for this fix which is minor anyway.

But, I can try to do that if you think this fix is a bit too ugly.

polkadot/node/network/statement-distribution/src/v2/grid.rs
alexggh: I'll fix the comment.

ordian (approved):

sandreim (approved):

rphmeier (approved):

alexggh:
Validated the fix on versi as well, merging now.

PR #3224, opened 2024-02-06
Prepare syncing for parallel sync strategies

dmitry-markin:
This PR should supersede https://github.com/paritytech/polkadot-sdk/pull/2814 and accomplish the same with less changes. It's needed to run sync strategies in parallel, like running `ChainSync` and `GapSync` as independent strategies, and running `ChainSync` and Sync 2.0 alongside each other.

The difference with https://github.com/paritytech/polkadot-sdk/pull/2814 is that we allow simultaneous requests to remote peers initiated by different strategies, as this is not tracked on the remote node in any way. Therefore, `PeerPool` is not needed.

CC @skunert


substrate/client/network/sync/src/engine.rs
```diff
@@ -675,21 +677,23 @@ where
...
 			match action {
-				SyncingAction::SendBlockRequest { peer_id, request } => {
+				SyncingAction::SendBlockRequest { peer_id, key, request } => {
 					// Sending block request implies dropping obsolete pending response as we are
 					// not interested in it anymore (see [`SyncingAction::SendBlockRequest`]).
-					// Furthermore, only one request at a time is allowed to any peer.
-					let removed = self.pending_responses.remove(&peer_id);
-					self.send_block_request(peer_id, request.clone());
+					let removed = self.pending_responses.remove(peer_id, key);
+					self.send_block_request(peer_id, key, request.clone());
```
altonen, line 696: Is there some code that triggers this behavior right now, i.e., that a block request is canceled because it's superseded by a new block request? I believe all strategies use `CancelRequest` to signal they're not interested in the block response anymore so maybe the `trace!()` below could be converted to a warning if a request is canceled without an explicit cancel action

substrate/client/network/sync/src/engine.rs
```diff
@@ -699,24 +703,24 @@ where
 						removed,
 					)
 				},
-				SyncingAction::CancelBlockRequest { peer_id } => {
-					let removed = self.pending_responses.remove(&peer_id);
+				SyncingAction::CancelRequest { peer_id, key } => {
+					let removed = self.pending_responses.remove(peer_id, key);
```
altonen, line 707: Looking at the code this should not be possible but just to make sure: what can't happen is that there is a ready response waiting which gets canceled and then a new request is sent, replacing the old request. It's not possible for this old, now-canceled response still to be returned to the strategy which it would incorrectly assume it to be a response to the new request?

substrate/client/network/sync/src/pending_responses.rs
```diff
@@ -87,8 +90,20 @@ impl<B: BlockT> PendingResponses<B> {
...
+	pub fn remove_all(&mut self, peer_id: &PeerId) {
+		let to_remove = self
+			.pending_responses
+			.keys()
+			.filter(|(peer, _key)| peer == peer_id)
+			.cloned()
+			.collect::<Vec<_>>();
+		to_remove.iter().for_each(|k| {
+			self.pending_responses.remove(k);
+		});
```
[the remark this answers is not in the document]
dmitry-markin: No, [`tokio_stream::StreamMap`](https://docs.rs/tokio-stream/latest/tokio_stream/struct.StreamMap.html) doesn't provide ways to manipulate the underlying `Vec` directly, so we have to remove entries one by one using available API.

substrate/client/network/sync/src/strategy.rs
```diff
@@ -92,15 +103,68 @@ pub enum SyncingAction<B: BlockT> {
 		number: NumberFor<B>,
 		justifications: Justifications,
 	},
-	/// Syncing strategy has finished.
+	/// Strategy finished. Nothing to do, this is handled by `SyncingStrategy`.
```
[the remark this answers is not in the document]
dmitry-markin: Having `SyncingAction::Finished` is needed to be able to convert specific actions using `Into::into()` here: https://github.com/paritytech/polkadot-sdk/blob/0d245f1fdecb52e3547fe63558622ea99b9af2d7/substrate/client/network/sync/src/strategy.rs#L483-L488

Otherwise we'd need to introduce match statements for every startegy in [`SyncingStartegy::actions()`](https://github.com/paritytech/polkadot-sdk/blob/0d245f1fdecb52e3547fe63558622ea99b9af2d7/substrate/client/network/sync/src/strategy.rs#L480).

substrate/client/network/sync/src/strategy.rs
```diff
@@ -183,77 +270,107 @@ where
...
+					.as_mut()
+					.map(|chain_sync| chain_sync.on_block_response(peer_id, request, blocks));
+			},
+			key => {
+				error!(
+					target: LOG_TARGET,
+					"`on_block_response()` called with unexpected key {key:?}",
+				);
+				debug_assert!(false);
+			},
```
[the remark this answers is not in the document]
dmitry-markin: Yes, this is possible. Added the check but we'll need to be careful to cancel all pending responses when switching between the strategies (not the case for warp & state now, but will be needed for fast sync).

substrate/client/network/sync/src/engine.rs
dmitry-markin: This should not be possible, because the order of actions is preserved: we first cancel the pending response (dropping the response future), and only after initiate a new request.

substrate/client/network/sync/src/strategy.rs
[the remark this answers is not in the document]
altonen: Similar check is missing for state/warp proof response functions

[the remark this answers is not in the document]
dmitry-markin: Thanks, fixed.

substrate/client/network/sync/src/strategy.rs
skunert, line 1: While reading I was asking myself if it was feasible to use a message based system for the individual strategies too. Instead of the syncing engine calling individual `on_XY` handlers of the strategy, it could pass a message down to the individual strategies, and each strategy decides if it wants to handle it.

Would get rid of strategy knowing which substrategy is interested in the individual results. But yeah this is more an educational question for me (and maybe there are cases were this does not even work), nothing to act upon.

skunert (approved):
The logic LGTM! Just left some questions and minor remarks.

substrate/client/network/sync/src/strategy.rs
dmitry-markin: Firstly, I'd like to comment on the feasibility of a fully async interface with strategies. When we implemented a bidirectional async interface between `ProtocolController` and `Notifications`, it turned out that it breaks the constraints of the originally half-synchronous interface (where we call `on_XY` handlers on `ProtocolController` and poll it as a stream for actions), requiring some tricks with handling of duplicate messages and discarding some "invalid" messages. We decided to just live with some smaller inconsistencies, because the proper implementation would require complex ACKing system with state machines in both `ProtocolController` and `Notifications` with lots of states. In syncing, our goal is to focus on a syncing state machines, and not on a message passing state machines — this is why we got rid of all the polling in the strategies.

On the other hand, what could work is introducing synchronous subscriber-like system, where we call a generic `on_event(event: Event)` handler that dispatches the events to specific strategies down the tree. Logically, this would be the same as calling specific `on_XY` handlers, but could replace the manual `match`es on active strategies with a subscriber-looking system, where a strategy instead registers itself for specific events. I'm not sure though if it's possible to have a "proper" Rust implementation of this without downcasting of abstract events when they reach specific strategies — otherwise, it looks like event matching would just move to strategies with a burden of them knowing about all the event types.

@altonen do you have something to add?

altonen: What Sebastian is proposing is similar to the trait approach I've been harping about. If we store the active strategies in `HashMap<StategyKey, Box<dyn Strategy>>`, `SyncingStrategy` could iterate over all active strategies when handling an event and each strategy can decide whether it wants to handle that particular event. Of course if a key is provided, e.g., when a response is received, then `SyncingStrategy` would only call the specified strategy. This would clean up much of the code in `SyncingStrategy` and would allow plugging custom syncing implementations.

Like you described in the second paragraph, I don't see how Sebastian's proposal necessarily implies any async code though. I think the code would still work the way it does now but instead of `SyncingStrategy` checking explicitly if a strategy could be interested in an event, it won't make any assumptions, passes the event to the strategy and if it's not interested, it will just ignore it.

skunert: Yes, the second paragraph captures pretty well what I meant. When we at some point add more strategies  with different response handlers we would just add another message instead of adding a new `on_XY` on `Strategy` and the concrete implementation.

dmitry-markin: If I got it right, what @altonen is proposing implies that all strategies handle all event types. I.e., should provide `on_XY` handlers for all `XY`, implementing some generic `Strategy` trait, even though they may not be interested in all `XY`.

PR #3227, opened 2024-02-06
subsystem-bench: Use only test files for subsystem benchmarks

AndreiEres:
This PR removes the configuration of subsystem benchmarks via CLI arguments. After this, we only keep configurations only in yaml files. It removes unnecessary code duplication


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5130136

polkadot/node/subsystem-bench/src/availability/mod.rs
```diff
@@ -13,71 +13,70 @@
 
 // You should have received a copy of the GNU General Public License
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
-use crate::{core::mock::ChainApiState, TestEnvironment};
+
```
alexggh, line 16: What's with this re-org of imports, different editor config ?

alexggh (approved):
Looks good to me.

This would need updating as well: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/subsystem-bench/README.md

polkadot/node/subsystem-bench/src/availability/mod.rs
AndreiEres: I think I just removed extra lines. Did they matter?

alexggh: I don't think you removed anything things seems to be just shuffled around.
Anyways, if it compiles it is fine.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5138717

Issue #3144, opened 2024-01-30
Fix (backing) relying on para ids being unique


eskimor:
[Here](https://github.com/paritytech/polkadot-sdk/blob/a190e0e9253562fdca9c1b6e9541a7ea0a50c018/polkadot/node/core/backing/src/lib.rs#L1076-L1084) we assign a `ParaId` to a validator group as if that were a unique mapping. This needs to be fixed. Any other occurrences of this assumption need to be fixed as well. Goal should be that a parachain being scheduled on multiple cores does not brick it.

Without full elastic scaling implemented only one core will be successfully occupied in the end and all other potentially produced blocks won't make it (wasted effort), but the parachain should still be able to get some blocks included.

sandreim:
Both node side and runtime make assumptions around a 1:1 mapping between parachain and backing group. 

### Problem
On the node side the `statement-table` crate implements a generic `Table` which is designed with the said assumption in mind.  The table is used to manage own and received statements of different candidates under a given relay chain block. To instantiate the table we need to provide[ all the backing groups,](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/core/backing/src/lib.rs#L1082) and in the case of multiple cores assigned to the same para, only the last core group would be retained. The statement table will fail to import statements and generate `BackedCandidates`. Statements for a given candidate  will be checked  based on the backing group info - [number of required votes](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/statement-table/src/generic.rs#L329), validator indices and [authority ids](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/statement-table/src/generic.rs#L392). One can easily see how this fails if candidate is build on a core for which we don't have the backing group info.

On the runtime side things are even worse, we cannot reason about the `CoreIndex`, hence know the backing group when [processing a backed candidate](https://github.com/paritytech/polkadot-sdk/blob/452f81366870436634f72e2a187179e5a9ad9894/polkadot/runtime/parachains/src/inclusion/mod.rs#L685) . So, even if only one core is occupied by a para with multiple cores assigned we wouldn't know the backing group, unless we test indices across all of them which raises a security issue as there is no way to ensure backers attest only their own backed candidate and not also other candidates on other cores.

### Proper Solution
[...]

Issue #3141, opened 2024-01-30
Fix provisioner issue for on-demand/coretime


eskimor:
Provisioner code does not seem to be fully on-demand ready.

https://github.com/paritytech/polkadot-sdk/blob/a02b53475b9fe2567ae05ad4e03c81f4f3d6b463/polkadot/node/core/provisioner/src/lib.rs#L662

antonva:
This should be fine and is generally the same as https://github.com/paritytech/polkadot-sdk/issues/664,
which is that `next_up_on_available()` used to use the `CoreIndex` to index into the `paras::parachains()` vec directly to map to a `ParaId`.

Ever since we refactored this out of the scheduler and into the `AssignmentProvider` interface we no longer have this assumption.

alindima:
Having a deeper look at the code and speaking with @antonva about this, we came to the following conclusion:

Assume core A is occupied by Para1, the `next_up_on_available` is Para2 and the bitfields for making the candidate of Para1 have been recorded on-chain.
The next block producer will call the provisioner, which will request from prospective-parachains the next backable candidate of Para2, which has the Para1 candidate as parent. There's obviously no such candidate so the core will transition to scheduled.
Para2 will be able to get a candidate in the next iteration.

So it's a slight inefficiency that is made less likely due to the core affinities.

I suggest we fix it by modifying the provisioner to check if the newly scheduled para is different than the previous one and if it is, request a backable candidate with the required_path of the new para instead (taking into account a possible parent candidate for Para 2 present on another core)

alindima:
I found a very similar problem in collation-generation: https://github.com/paritytech/polkadot-sdk/issues/3327

PR #3229, opened 2024-02-06
Elastic scaling: use an assumed `CoreIndex` in `candidate-backing`

sandreim:
First step in implementing https://github.com/paritytech/polkadot-sdk/issues/3144

### Summary of changes
- switch statement `Table` candidate mapping from `ParaId` to `CoreIndex`
- introduce experimental `InjectCoreIndex`  node feature.
- determine and assume a `CoreIndex` for a candidate based on statement validator index. If the signature is valid it means validator controls the validator that index and we can easily map it to a validator group/core.
- introduce a temporary provisioner fix until we fully enable elastic scaling in the subystem. The fix ensures we don't fetch the same backable candidate when calling `get_backable_candidate` for each core.

TODO:
- [ ] fix tests
- [ ] add new tests


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5184854

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5185421

alindima (commented):
Looks good! Nice work!

Cargo.lock
```diff
@@ -21234,9 +21236,9 @@ dependencies = [
 
 [[package]]
 name = "wasmi"
-version = "0.31.2"
+version = "0.31.0"
```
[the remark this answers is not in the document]
sandreim: yes, probably because of a master merge

polkadot/primitives/src/vstaging/mod.rs
```diff
@@ -68,5 +68,12 @@ pub mod node_features {
 		/// Every time a new feature flag is assigned it should take this value.
 		/// and this should be incremented.
 		FirstUnassigned = 1,
+		/// Experimental features start at bit 16. Note that experimental features pop in and out
+		/// of exsitence without warning.
+		///
+		/// This feature enables the extension of `BackedCandidate::validator_indices` by 8 bit.
+		/// The value stored there represents the assumed core index where the candidates
+		/// are backed.
+		InjectCoreIndex = 16,
```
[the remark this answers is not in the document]
sandreim: I think the description is not good. A better description is experimental features will not make it to production(kusama/polkadot)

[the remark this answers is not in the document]
sandreim: I'll fix it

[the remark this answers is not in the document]
alindima: > A better description is experimental features will not make it to production(kusama/polkadot).

you mean that they won't be enabled there, right?

I wouldn't give it a special index, is there any reason for that?

While we're at it, maybe also rename it to smth like `ElasticScalingExperimental` as we discussed yesterday in the sync.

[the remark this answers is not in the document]
sandreim: Yes, I'll assign an appropriate index, it seems that after all it is likely we will go with this feature in production as the candidate commitments changes will take much longer to go through

polkadot/primitives/src/v6/mod.rs
```diff
@@ -743,17 +744,35 @@ impl<H> BackedCandidate<H> {
...
-pub fn check_candidate_backing<H: AsRef<[u8]> + Clone + Encode>(
+pub fn check_candidate_backing<H: AsRef<[u8]> + Clone + Encode + core::fmt::Debug>(
 	backed: &BackedCandidate<H>,
 	signing_context: &SigningContext<H>,
 	group_len: usize,
 	validator_lookup: impl Fn(usize) -> Option<ValidatorId>,
 ) -> Result<usize, ()> {
+	log::debug!(
+		target: LOG_TARGET,
+		"checking candidate {:?}",
```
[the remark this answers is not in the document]
sandreim: what information is missing ?

polkadot/node/core/backing/src/lib.rs
```diff
@@ -981,6 +990,74 @@ async fn handle_active_leaves_update<Context>(
...
+{
+	let parent = relay_parent;
+
+	let (validator_groups, group_rotation_info) = request_validator_groups(parent, sender)
+		.await
+		.await
+		.map_err(Error::RuntimeApiUnavailable)?
+		.map_err(Error::FetchValidatorGroups)?;
+
+	let cores = request_availability_cores(parent, sender)
```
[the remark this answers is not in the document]
sandreim: done

polkadot/primitives/src/vstaging/mod.rs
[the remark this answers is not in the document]
sandreim: fixed

polkadot/primitives/src/v6/mod.rs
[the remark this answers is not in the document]
alindima: the log will look like:
```
runtime::primitives: checking candidate
```

we won't know what checking operation this is related to. I'd specify that we're checking the validity and number of backing votes.

alindima (approved):

polkadot/primitives/src/v6/mod.rs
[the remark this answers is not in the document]
sandreim: removed it, this isn't actually useful and has quite large output.

polkadot/statement-table/src/generic.rs
```diff
@@ -387,9 +385,10 @@ impl<Ctx: Context> Table<Ctx> {
 		authority: Ctx::AuthorityId,
 		candidate: Ctx::Candidate,
 		signature: Ctx::Signature,
+		group: Ctx::GroupId,
 	) -> ImportResult<Ctx> {
-		let group = Ctx::candidate_group(&candidate);
```
eskimor, line 391: I think we should add a check that the `ParaId` assigned to the core matches the paraid that got signed in the statement. Otherwise validators could sign garbage (not assigned paraid candidates), knowing they will be filtered out in the runtime, but DoSing the parachain as the legitimate statements are no longer accepted.

polkadot/node/core/backing/src/lib.rs
```diff
@@ -981,6 +990,74 @@ async fn handle_active_leaves_update<Context>(
...
+	validator_groups: &[Vec<ValidatorIndex>],
+	group_rotation_info: &GroupRotationInfo,
+	statement: &SignedFullStatementWithPVD,
+) -> Option<CoreIndex> {
+	let statement_validator_index = statement.validator_index();
+	for (group_index, group) in validator_groups.iter().enumerate() {
+		for validator_index in group {
+			if *validator_index == statement_validator_index {
+				return Some(
+					group_rotation_info.core_for_group(GroupIndex(group_index as u32), cores.len()),
```
eskimor: We should also check that the paraid of the core matches the candidate in the statement.

eskimor: With the [upcoming exposure of claimqueue](https://github.com/orgs/paritytech/projects/119/views/20?pane=issue&itemId=42349591) this will need to be changed to check for any occurrence in the claim queue of the core. @tdimitrov

polkadot/statement-table/src/generic.rs
eskimor: (better to do this right where we determine the core from the statement - see other comment)

polkadot/node/core/backing/src/tests/mod.rs
```diff
@@ -639,6 +680,84 @@ fn backing_works() {
...
+	let core_index_2 = core_index_from_statement_inner(
+		&test_state.availability_cores,
+		&test_state.validator_groups.0,
+		&test_state.validator_groups.1,
+		&signed_statement_2,
+	)
+	.unwrap();
+
+	assert_eq!(core_index_2, CoreIndex(1));
+}
```
eskimor, line 728: Let's add a test that this fails if the paraid of the statement does not match the core information of the test state.

polkadot/node/core/backing/src/lib.rs
sandreim: done

polkadot/node/core/backing/src/tests/mod.rs
sandreim: done

polkadot/statement-table/src/generic.rs
sandreim: done

polkadot/node/core/backing/src/lib.rs
```diff
@@ -224,6 +227,15 @@ struct PerRelayParentState {
...
 	minimum_backing_votes: u32,
+	/// If true, we're appendindg extra bits in the BackedCandidate validator indices bitfield,
+	/// which represent the assigned core index.
+	inject_core_index: bool,
+	/// The core state for all cores
+	cores: Vec<CoreState>,
+	/// The validator groups at this relay parent.
+	validator_groups: Vec<Vec<ValidatorIndex>>,
+	/// The associated group rotation information.
+	group_rotation_info: GroupRotationInfo,
```
eskimor, line 238: Those two are really per session. Storing them per relay parent is rather unnecessary. Seems fine for an interim solution though.

polkadot/node/core/backing/src/lib.rs
```diff
@@ -479,6 +488,12 @@ fn table_attested_to_backed(
 	}
 	vote_positions.sort_by_key(|(_orig, pos_in_group)| *pos_in_group);
 
+	if inject_core_index {
+		let core_index_to_inject: BitVec<u8, BitOrderLsb0> =
+			BitVec::from_vec(vec![core_index.0 as u8]);
+		validator_indices.extend(core_index_to_inject);
```
eskimor, line 494: Shouldn't we be using some nice `BackedCandidate` function to be introduced in the other PR for that?

polkadot/node/core/backing/src/lib.rs
```diff
@@ -981,6 +996,75 @@ async fn handle_active_leaves_update<Context>(
...
+				}
+
+				if let StatementWithPVD::Seconded(candidate, _pvd) = statement.payload() {
+					let candidate_para_id = candidate.descriptor.para_id;
+					let assigned_para_id = match &cores[core_index.0 as usize] {
+						CoreState::Free => {
+							gum::debug!(target: LOG_TARGET, ?candidate_hash, "Invalid CoreIndex, core is not assigned to any para_id");
+							return None
+						},
+						CoreState::Occupied(occupied) => occupied.candidate_descriptor.para_id,
```
eskimor, line 1049: This looks wrong. We should be checking [next_up_on_available](https://github.com/paritytech/polkadot-sdk/blob/fbe1ad516de65b20807c13b502051103547ad538/polkadot/primitives/src/v6/mod.rs#L1020). (Worth checking the collator protocol, maybe it is wrong there as well.)

alindima: You're right. I am discovering more and more places where async backing (and by extension elastic scaling) has not been optimised for on-demand.
See https://github.com/paritytech/polkadot-sdk/issues/3141, https://github.com/paritytech/polkadot-sdk/issues/3327, https://github.com/paritytech/polkadot-sdk/blob/d7862aa8c9b4f8be1d4330bc11c742bf48d407f6/polkadot/node/network/collator-protocol/src/validator_side/mod.rs#L515 and https://github.com/paritytech/polkadot-sdk/blob/d7862aa8c9b4f8be1d4330bc11c742bf48d407f6/polkadot/node/core/backing/src/lib.rs#L1065

All of these places (may be more) would work fine as long as the on-demand parachains maintain their core affinities. Whenever there's a switch, cores will be left unoccupied for a relay parent.

I think we should optimise this as a separate issue

polkadot/node/core/backing/src/lib.rs
alindima: that's a good idea, will do that in the follow-up PR

polkadot/node/core/backing/src/lib.rs
eskimor: Yes although do it right from the start in new changes.

alindima: I wonder whether we should also take into account next_up_on_timeout.
It adds quite a lot of complexity. I think we're better off keeping things simple and accepting than on rare timeout occasions, we'll see an empty core for a relay chain block duration

alindima: done

polkadot/node/core/backing/src/lib.rs
```diff
@@ -981,6 +996,81 @@ async fn handle_active_leaves_update<Context>(
...
+) -> Option<CoreIndex> {
+	let compact_statement = statement.as_unchecked();
+	let candidate_hash = CandidateHash(*compact_statement.unchecked_payload().candidate_hash());
+
+	let n_cores = cores.len();
+
+	gum::trace!(target: LOG_TARGET, ?group_rotation_info, ?statement, ?validator_groups, n_cores = ?cores.len() , ?candidate_hash, "Extracting core index from statement");
+
+	let statement_validator_index = statement.validator_index();
+	for (group_index, group) in validator_groups.iter().enumerate() {
```
eskimor, line 1030: Given that we are doing this for each and every statement, a table ValidatorIndex -> GroupIndex might make sense (also valid for an entire session), so we get O(1) complexity here.

polkadot/node/core/backing/src/lib.rs
```diff
@@ -1020,6 +1093,16 @@ async fn construct_per_relay_parent_state<Context>(
 	.map_err(Error::JoinMultiple)?;
 
 	let session_index = try_runtime_api!(session_index);
+
+	let inject_core_index = request_node_features(parent, session_index, ctx.sender())
```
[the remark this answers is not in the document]
alindima: correct. I'll open a separate issue to track this buffering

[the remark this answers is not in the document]
alindima: https://github.com/paritytech/polkadot-sdk/issues/3421

polkadot/node/core/backing/src/lib.rs
alindima: done, PTAL

polkadot/node/core/backing/src/lib.rs
```diff
@@ -232,8 +234,8 @@ struct PerRelayParentState {
 	inject_core_index: bool,
 	/// The core states for all cores.
 	cores: Vec<CoreState>,
-	/// The validator groups at this relay parent.
-	validator_groups: Vec<Vec<ValidatorIndex>>,
+	/// The validator index -> group mapping at this relay parent.
+	validator_to_group: IndexedVec<ValidatorIndex, Option<GroupIndex>>,
```
eskimor, line 238: Is it really necessary to copy this to the relay parent state? Can't we just keep the session index here and then fetch a reference from state directly where we need it?

eskimor (approved):

polkadot/node/core/backing/src/lib.rs
alindima: this being an LruMap, it makes it a bit harder to reason about the existence of the session index in the map.

And we can't store a mutable reference here and just populate the map when the session is not present because we'd need to hand out multiple mutable references.

Realistically, backing can only happen on the current session, right? So the LruMap doesn't need a capacity larger than 1 (so then, it doesn't even need to be a LruMap). Am I right?

alindima: with current async backing code, we cannot start backing a candidate in the next session in advance. But we may want to keep this possibility, making the LruMap neccessary.

What I can do is wrap it in an Arc, because they don't need to modify it. Which would achieve the same

alindima: done. I'll merge it once CI passes again

PR #3231, opened 2024-02-06
Runtime: allow backing multiple candidates of same parachain on different cores 

sandreim:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3144

Builds on top of https://github.com/paritytech/polkadot-sdk/pull/3229

### Summary
Some preparations for Runtime to support elastic scaling, guarded by config node features bit `FeatureIndex::ElasticScalingMVP`. This PR introduces a per-candidate `CoreIndex`  but does it in a hacky way to avoid changing `CandidateCommitments`, `CandidateReceipts` primitives and networking protocols.

#### Including `CoreIndex` in `BackedCandidate`
If the `ElasticScalingMVP` feature bit is enabled then `BackedCandidate::validator_indices` is extended by 8 bits. 
The value stored in these bits represents the assumed core index for the candidate.

It is temporary solution which works by creating a mapping from `BackedCandidate` to `CoreIndex` by assuming the `CoreIndex` can be discovered by checking in which validator group the validator that signed the statement is.

TODO:
- [x] fix tests
- [x] add new tests
- [x] Bump runtime API for Kusama, so we have that node features thing! -> https://github.com/polkadot-fellows/runtimes/pull/194


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5135138

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1095,12 +1102,16 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
 	// 1. Core index assigned to the parachain which has produced the candidate
 	// 2. The relay chain block number of the candidate
 	backed_candidates.retain_mut(|bc| {
-		// Get `core_idx` assigned to the `para_id` of the candidate
-		let core_idx = match scheduled.get(&bc.descriptor().para_id) {
-			Some(core_idx) => *core_idx,
-			None => {
-				log::debug!(target: LOG_TARGET, "Can't get core idx of a backed candidate for para id {:?}. Dropping the candidate.", bc.descriptor().para_id);
-				return false
+		let core_idx = if let Some(core_idx) = strip_candidate_core_index::<T>(bc) {
```
sandreim, line 1105: Seems that there is potential to strip twice ....

alindima:
Seeing this new `InjectCoreIndex` feature made me think about how are we going to enable elastic scaling (on the runtime side). On the node side, all of the changes that I can think of are backwards compatible.

I think we'll need another NodeFeature bit, `AllowMultipleCoresPerPara`, which will switch on the logic in the runtime for accepting multiple candidates for a parachain.

This way:
1. if `InjectCoreIndex` is enabled and `AllowMultipleCoresPerPara` is enabled, elastic scaling is enabled and working
1. if `InjectCoreIndex` is enabled and `AllowMultipleCoresPerPara` is disabled, parachains can acquire multiple cores but the runtime will only accept one candidate at a time. (what this PR aims to do)
1. if `InjectCoreIndex` is disabled and `AllowMultipleCoresPerPara` is enabled, this makes no sense. Elastic scaling won't work and parachains who acquire multiple cores will be stalled (the current status quo)
1. if `InjectCoreIndex` is disabled and `AllowMultipleCoresPerPara` is disabled. Identical to option 3 above.

Seeing this, I'm wondering if having this `InjectCoreIndex` makes any sense on its own after all. I guess it does make sense because it'll be disabled once we have the CoreIndex in the receipt.

CC @eskimor

alindima:
Maybe a `NodeFeature` is not the most suitable name for this, because in this case it enables changes in the runtime as well (not only on the node). Especially the `AllowMultipleCoresPerPara ` I mention above. But I think it's fine

sandreim:
@alindima 
I think `InjectCoreIndex` is sufficient to enable elastic scaling in Runtime as well as node side. Not sure why you want another feature. The soft switch which you seem to be looking for is actually having paras assigned to multiple cores.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5195725

alindima:
> @alindima I think `InjectCoreIndex` is sufficient to enable elastic scaling in Runtime as well as node side. Not sure why you want another feature. The soft switch which you seem to be looking for is actually having paras assigned to multiple cores.

I thought we wanted to maintain the possibility of not having elastic scaling but still not bricking a parachain which occupies more cores.
Now that I think about it again, maybe it isn't very helpful. Only for the possibility of disabling elastic scaling to mitigate a bug. If there were some bug with elastic scaling that stalled a parachain, disabling `InjectCoreIndex ` would not unbrick the parachain.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5195969

sandreim:
> I thought we wanted to maintain the possibility of not having elastic scaling but still not bricking a parachain which occupies more cores. 

This is included in this PR, they won't break if the `InjectCoreIndex` feature is enabled, even if parachains have multiple cores assigned and their collators are unupgraded.

alindima:
> > I thought we wanted to maintain the possibility of not having elastic scaling but still not bricking a parachain which occupies more cores.
> 
> This is included in this PR, they won't break if the `InjectCoreIndex` feature is enabled, even if parachains have multiple cores assigned and their collators are unupgraded.

I am talking about after we enable elastic scaling

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5198441

sandreim:
Anything we implement for elastic scaling must not break parachains that don't make use of this feature, while also not requiring them to upgrade (except of when we add CoreIndex to the commited candidate receipt).

We should prevent parachains (I'll cut a ticket) from acquiring multiple cores as a safety measure and as an enablement mechanism. We initially plan to launch on Rococo, while on other chains the possibility of assigning multiple cores is disabled - hence elastic scaling is disabled.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1089,18 +1094,27 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
-		// Get `core_idx` assigned to the `para_id` of the candidate
-		let core_idx = match scheduled.get(&bc.descriptor().para_id) {
-			Some(core_idx) => *core_idx,
-			None => {
-				log::debug!(target: LOG_TARGET, "Can't get core idx of a backed candidate for para id {:?}. Dropping the candidate.", bc.descriptor().para_id);
-				return false
+		let core_idx = if let Some(core_idx) = bc.assumed_core_index(core_index_enabled) {
+			core_idx
+		} else {
+			// Get `core_idx` assigned to the `para_id` of the candidate
```
alindima, line 1111: I think we should check here if we got multiple candidates on different cores for this para here.
If we did and the `InjectCoreIndex` is not enabled, we shouldn't back anything for this para here, because we can't tell if disabled validators voted or not (because we can't know the assigned group).

alindima: or just filter the votes from all disabled validators of all groups assigned to this para (invalid votes belonging to another group should not be recorded anyway, but we can't check that). with the current code, we'd just randomly use one of them

sandreim: Before this function is called, the `elastic_scaling_mvp_filter` will filter out any candidates that are missing core_index when the `ElasticScalingCoreIndex` feature is enabled.  If the feature is not enabled we assume that the para:core mapping is unique.

I don't think the logic here needs to be altered if a para has multiple cores assigned.

alindima: oh, right. If `elastic_scaling_mvp_filter` will do this in the future then it's fine

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5214682

polkadot/runtime/parachains/src/util.rs
```diff
@@ -98,13 +104,104 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+		} else {
+			// This shouldn't happen, but at this point the candidate without core index is fine
+			// since we know the para:core mapping is unique.
+			true
+		}
+	})
+}
+
+// Returns `true` if the candidate contains an injected `CoreIndex`.
+fn has_core_index<T: configuration::Config + scheduler::Config>(
```
eskimor, line 161: Shouldn't this be just calling `assumed_core_index` on `BackedCandidate`?

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -31,6 +31,7 @@ use crate::{
 	paras,
 	scheduler::{self, FreedReason},
 	shared::{self, AllowedRelayParentsTracker},
+	util::filter_elastic_scaling_candidates,
```
eskimor, line 34: We should structure this better in components that make sense. "util" tells nothing and suggests no separation of concerns at all.

eskimor:
With regards to node features: Guarding the node to not submit things to the runtime the runtime does not understand makes a lot of sense obviously. I am not sure about guarding features in the runtime itself. If we can do it, why wouldn't we? Seems like an unnecessary complication. I would also say that the node feature should be `ElasticScalingMVP`. Whether or not it is actually working already or just preventing parachains to be bricked can evolve without additional features, if need be. (No interfaces are changing)

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -591,6 +592,8 @@ impl<T: Config> Pallet<T> {
 
 		METRICS.on_candidates_processed_total(backed_candidates.len() as u64);
 
+		filter_elastic_scaling_candidates::<T>(&mut backed_candidates);
```
eskimor, line 595: Ok, we need to clean up this function more. We should ensure at the top-level at if [on enter](https://github.com/paritytech/polkadot-sdk/blob/a5ba15722208e0aeeeff3ca0dd161bedc31a8b1d/polkadot/runtime/parachains/src/paras_inherent/mod.rs#L111) anything is filtered, then we should err out like [here](https://github.com/paritytech/polkadot-sdk/blob/a5ba15722208e0aeeeff3ca0dd161bedc31a8b1d/polkadot/runtime/parachains/src/paras_inherent/mod.rs#L487), as the filtering should have happened on create_inherent already. Otherwise it is possible to produce blocks with lots of garbage (occupying storage) not having any effect.

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -601,21 +602,27 @@ impl<T: Config> Pallet<T> {
...
 		let now = <frame_system::Pallet<T>>::block_number();
 
-		ensure!(candidates.len() <= scheduled.len(), Error::<T>::UnscheduledCandidate);
+		ensure!(candidates.len() <= scheduled_by_core.len(), Error::<T>::UnscheduledCandidate);
 
 		if scheduled.is_empty() {
 			return Ok(ProcessedCandidates::default())
 		}
 
+		let core_index_enabled = configuration::Pallet::<T>::config()
```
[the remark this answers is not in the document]
alindima: I generally agree with you, but I think it's a chicken-and-egg situation. In order to know if we have the core index encoded in the validator_indices, we need to know the backing group size. In order to know the backing group, we need to know the core index

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -680,10 +687,20 @@ impl<T: Config> Pallet<T> {
 				};
 
 				let para_id = backed_candidate.descriptor().para_id;
-				let mut backers = bitvec::bitvec![u8, BitOrderLsb0; 0; validators.len()];
+				let core_idx = if let Some(core_idx) =
+					backed_candidate.assumed_core_index(core_index_enabled)
+				{
+					core_idx
```
alindima, line 693: let's also check scheduled_by_core here

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -738,16 +755,16 @@ impl<T: Config> Pallet<T> {
 
 					let mut backer_idx_and_attestation =
 						Vec::<(ValidatorIndex, ValidityAttestation)>::with_capacity(
-							backed_candidate.validator_indices.count_ones(),
+							backed_candidate.validator_indices(core_index_enabled).count_ones(),
```
alindima, line 758: we can only call validator_indices once

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: yeah, I think we'd better move the function here in the paras_inherent module. IMO It's not a util unless it's used by multiple modules

polkadot/runtime/parachains/src/util.rs
```diff
@@ -98,13 +104,104 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+	}
+
+	// A mapping from parachain to all assigned cores.
+	let mut cores_per_parachain: BTreeMap<ParaId, Vec<CoreIndex>> = BTreeMap::new();
+
+	for (core_index, para_id) in <scheduler::Pallet<T>>::scheduled_paras() {
+		cores_per_parachain.entry(para_id).or_default().push(core_index);
+	}
+
+	// We keep a candidate if the parachain has only one core assigned or if
```
alindima, line 133: shouldn't we do the same thing if ElasticScalingMVP feature is disabled? only keep one candidate

polkadot/runtime/parachains/src/util.rs
```diff
@@ -98,13 +104,104 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
 	subset
 }
 
+/// On first pass it filters out all candidates that have multiple cores assigned and no
+/// `CoreIndex` injected.
+///
+/// On second pass we filter out candidates with the same core index. This can happen if for example
+/// collators distribute collations to multiple backing groups.
```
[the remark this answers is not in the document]
alindima: > how can providing collations to multiple backing groups lead to multiple candidates having the same core index?

I don't get that either. I don't think it's possible.

What I think is instead possible is to have the same candidate multiple times in the vector. But this shouldn't be any different than the current runtime logic, we should only pick one as elastic scaling is not yet supported.

polkadot/runtime/parachains/src/util.rs
alindima: I think it should

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1137,13 +1151,17 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
 		let disabled_indices = BitVec::<u8, bitvec::order::Lsb0>::from_iter(validator_group.iter().map(|idx| disabled_validators.contains(idx)));
+		let mut validator_indices = bc.validator_indices(core_index_enabled);
 		// The indices of statements from disabled validators in `BackedCandidate`. We have to drop these.
-		let indices_to_drop = disabled_indices.clone() & &bc.validator_indices;
+		let indices_to_drop = disabled_indices.clone() & &validator_indices;
 		// Apply the bitmask to drop the disabled validator from `validator_indices`
-		bc.validator_indices &= !disabled_indices;
+		validator_indices &= !disabled_indices;
+		// Update the backed candidate
+		bc.set_validator_indices(validator_indices);
```
alindima, line 1160: This actually siltently removes the core index from the validator_indices bitfield, so the inclusion module won't have access to the core indices in process_candidates()

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5272280

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -648,7 +655,7 @@ impl<T: Config> Pallet<T> {
 			//
 			// In the meantime, we do certain sanity checks on the candidates and on the scheduled
 			// list.
-			for (candidate_idx, backed_candidate) in candidates.iter().enumerate() {
+			for (candidate_idx, backed_candidate) in candidates.iter_mut().enumerate() {
```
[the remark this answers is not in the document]
alindima: it is being used now. We remove the core index from the validator indices before passing on the BackedCandidate to the validation done in `verify_backed_candidate` and `check_candidate_backing`

polkadot/runtime/parachains/src/util.rs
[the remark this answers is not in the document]
alindima: removed this

polkadot/runtime/parachains/src/util.rs
```diff
@@ -98,13 +104,104 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+	let validator_indices_raw = candidate.validator_indices(core_index_enabled);
+	let (validator_indices_slice, core_idx_slice) = validator_indices_raw.split_at(core_idx_offset);
+	let core_idx: u8 = core_idx_slice.load();
+
+	let current_block = frame_system::Pallet::<T>::block_number();
+
+	// Get the backing group of the candidate backed at `core_idx`.
+	let group_idx = match <scheduler::Pallet<T>>::group_assigned_to_core(
+		CoreIndex(core_idx as u32),
+		current_block,
```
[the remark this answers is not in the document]
alindima: right, changed this

polkadot/runtime/parachains/src/util.rs
alindima: I added this

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: changed this

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5272574

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: Correct, I added this

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5274207

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5281342

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5282553

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5282536

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5282555

polkadot/node/core/backing/src/lib.rs
```diff
@@ -488,20 +488,20 @@ fn table_attested_to_backed(
...
-		validity_votes: vote_positions
+		vote_positions
 			.into_iter()
 			.map(|(pos_in_votes, _pos_in_group)| validity_votes[pos_in_votes].clone())
 			.collect(),
 		validator_indices,
-	})
+	);
+
+	if inject_core_index {
```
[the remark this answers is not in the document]
alindima: done, moved the optional `CoreIndex` to `BackedCandidate::new`

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -591,6 +591,25 @@ impl<T: Config> Pallet<T> {
 
 		METRICS.on_candidates_processed_total(backed_candidates.len() as u64);
 
+		let core_index_enabled = configuration::Pallet::<T>::config()
+			.node_features
+			.get(FeatureIndex::ElasticScalingMVP as usize)
+			.map(|b| *b)
+			.unwrap_or(false);
+
+		let dropped_elastic_scaling_candidates = filter_elastic_scaling_candidates::<T>(
```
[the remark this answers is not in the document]
alindima: agreed

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1168,3 +1202,84 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+			return false
+		},
+	};
+
+	let group_validators = match <scheduler::Pallet<T>>::group_validators(group_idx) {
+		Some(validators) => validators,
+		None => return false,
+	};
+
+	group_validators.len() == validator_indices.len()
```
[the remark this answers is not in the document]
alindima: yeah, we're already doing the check here: https://github.com/paritytech/polkadot-sdk/blob/fee810a5ea177f95ddf13ea51270b84f1e117bae/polkadot/primitives/src/v6/mod.rs#L755

polkadot/primitives/src/v6/mod.rs
```diff
@@ -748,44 +803,42 @@ impl<H> BackedCandidate<H> {
 /// Returns either an error, indicating that one of the signatures was invalid or that the index
 /// was out-of-bounds, or the number of signatures checked.
 pub fn check_candidate_backing<H: AsRef<[u8]> + Clone + Encode + core::fmt::Debug>(
-	backed: &BackedCandidate<H>,
```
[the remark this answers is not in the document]
alindima: I wanted to make it explicit that we should have stripped the core index from the validator indices before calling this function. The split is already being done before calling this function so we can avoid doing the split again by passing in the needed members

polkadot/primitives/src/v6/mod.rs
```diff
@@ -735,6 +759,37 @@ impl<H> BackedCandidate<H> {
...
+
+	/// Get a copy of the validator indices and the assumed core index, if any.
+	pub fn validator_indices_and_core_index(
+		&self,
+		core_index_enabled: bool,
+	) -> (&BitSlice<u8, bitvec::order::Lsb0>, Option<CoreIndex>) {
+		// This flag tells us if the block producers must enable Elastic Scaling MVP hack.
+		// It extends `BackedCandidate::validity_indices` to store a 8 bit core index.
+		if core_index_enabled {
+			let core_idx_offset = self.validator_indices.len().saturating_sub(8);
```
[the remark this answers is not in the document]
alindima: hmm, how is this related to core count? I think it would be a problem to assume that the core index is present if the length is greater than 8 for a backing group size larger than 8. Isn't that right?

Is it fair to assume that backing group sizes won't grow past 8? If it is, maybe we should add some runtime logic that forbids this

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5290104

polkadot/runtime/parachains/src/paras_inherent/mod.rs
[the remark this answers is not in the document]
alindima: > If this is not true, then the check is redundant here and should be removed.

I wonder if it's safe to remove it. After we call filter_elastic_scaling_candidates(), we filter out backing votes of disabled validators (before doing the check I linked above). So if ElasticScalingMVP is enabled but a validator was not updated, we'd maybe wrongfully filter disabled votes because we wouldn't realise that the core index was not actually encoded (only if the backing group size is larger than 8).

So we'd calculate a wrong core index for the candidate and possibly wrongfully remove the backing of the candidate which had not encoded the core index after ElasticScalingMVP was enabled.

Considering that this would require a backing group size larger than 8 (which we don't have on any network AFAIK) and that I think it only enables the auto-censorship of unupgraded nodes, I'd say it's fine and we could remove this check here. WDYT?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5292516

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5292787

polkadot/primitives/src/v6/mod.rs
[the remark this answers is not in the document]
eskimor: LOL. I was completely off here and confused two things. Right the limit is the backing group size - which we are more on the edge of reducing in size than increasing. Maybe someone has a testnetwork with large backing groups :thinking: Indeed adding a check to configuration that groups can't be bigger than 8 might make sense. Ok so options are:

1. Leave as is.
2. Do limit group sizes to <=8.
3. Add additional bits which indicate how the rest should be treated.

At this point I am also leaning towards (1). It is ugly, but it is also a hacky interim solution ... so, might as well look like it.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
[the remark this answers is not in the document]
eskimor: let's keep as is then.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5296698

polkadot/zombienet_tests/functional/0012-elastic-scaling-mvp.zndsl
```diff
@@ -0,0 +1,27 @@
...
+
+# Ensure parachain was able to make progress.
+validator: parachain 2000 block height is at least 10 within 200 seconds
+
+# Register the second core assigned to this parachain.
+alice: js-script ./0012-register-para.js return is 0 within 600 seconds
+
+validator: reports substrate_block_height{status="finalized"} is at least 35 within 100 seconds
+
+# Parachain will now be stalled
```
eskimor, line 18: Can we have a check for this actually happening? block height is at most?

eskimor (approved):
Nice work! Only one nit left: Do we have to copy to relay parent state? Also - actually checking that the parachain is stalled would be good.

polkadot/zombienet_tests/functional/0012-elastic-scaling-mvp.zndsl
alindima: yes, I can add that.

The test should act as a regression test even without this check though. Before this PR, `parachain 2000 block height is at least 30 within 200 seconds` would have not been satisfied because the parachain would never get unstalled and it'd be stuck at a height around 15.

alindima: done

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5309090

polkadot/zombienet_tests/functional/0012-elastic-scaling-mvp.zndsl
```diff
@@ -0,0 +1,28 @@
...
+validator: parachain 2000 block height is lower than 20 within 300 seconds
+
+# Enable the ElasticScalingMVP node feature.
+alice: js-script ./0012-enable-node-feature.js with "1" return is 0 within 600 seconds
+
+# Wait two sessions for the config to be updated.
+sleep 120 seconds
```
eskimor, line 25: It would be better to wait for the actual session change :grimacing:  - If that is not possible right now, we should add that feature.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1166,5 +1211,103 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+	// a core index is provided by block author and it's indeed scheduled.
+	for backed_candidate in candidates {
+		let maybe_injected_core_index = get_injected_core_index::<T>(
+			allowed_relay_parents,
+			&backed_candidate,
+			core_index_enabled,
+		);
+
+		let scheduled_cores = scheduled.get_mut(&backed_candidate.descriptor().para_id);
+		if let Some(scheduled_cores) = scheduled_cores {
```
tdimitrov, line 1240:
```suggestion
        // Candidates without scheduled cores are silently filtered out.
		if let Some(scheduled_cores) = scheduled_cores {
```

I'd make this filtering more explicit but this is a very personal preference.

But let's have a comment pointing out why there is no `else` here.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -626,15 +651,26 @@ impl<T: Config> Pallet<T> {
 			ensure!(!votes_from_disabled_were_dropped, Error::<T>::BackedByDisabled);
 		}
 
+		// In `Enter` context (invoked during execution) we shouldn't have filtered any candidates
+		// due to a para having multiple cores assigned and no injected core index. They have been
+		// filtered during inherent data preparation (`ProvideInherent` context). Abort in such
+		// cases.
+		if context == ProcessInherentDataContext::Enter {
+			ensure!(
+				!dropped_elastic_scaling_candidates,
```
tdimitrov, line 660: Unrelated to this PR but why we check if we have dropped 'elastic scaling candidates' here, but we are fine dropping candidates without assigned core (which happens [here](https://github.com/paritytech/polkadot-sdk/blob/cb41758fe9a38ee26821bff93e615eb10848f7ad/polkadot/runtime/parachains/src/paras_inherent/mod.rs#L658))?

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1166,5 +1211,103 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+			} else {
+				dropped_elastic_scaling_candidates = true;
+			}
+		}
+	}
+
+	(backed_candidates_with_core, dropped_elastic_scaling_candidates)
+}
+
+// Returns `true` if the candidate contains a valid injected `CoreIndex`.
```
tdimitrov, line 1258:
```suggestion
```

The comment is wrong, but I think we don't need it. The sigunature by itself is pretty self-explanatory.

tdimitrov (approved):
Looks very good @alindima. Left a few nits and a question.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: Thanks for pointing this out!

prior to this PR, we just assumed that one para maps to one core. So we were only checking if the candidate's paraid has one scheduled core.
if it didn't, we'd return an error. if multiple candidates with the same paraid were supplied, we'd only check for one scheduled core.

The logic I added is more comprehensive and will work for proper elastic scaling also. But I decided to be more lenient and just filter out unscheduled candidates instead of erroring.

However, the check you mentioned should be I think extended to all dropped candidates 👍️ I don't see a reason why it should only be for some of them
because they should be all filtered by the block author. the other block importers should not do this

alindima: done

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5329732

eskimor (approved):

PR #3232, opened 2024-02-06
Validate code when scheduling uprades from registrar

Szegoo:
Currently, anyone can registrar a code that exceeds the code size limit when performing the upgrade from the registrar. This PR fixes that and adds a new test to cover this.

cc @bkchr @eskimor


polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -681,6 +681,16 @@ impl<T: Config> Pallet<T> {
 		debug_assert!(res2.is_ok());
 		T::OnSwap::on_swap(to_upgrade, to_downgrade);
 	}
+
+	/// Checks whether the provided code is valid.
+	fn validate_code(code: ValidationCode) -> DispatchResult {
+		let config = configuration::Pallet::<T>::config();
+
+		ensure!(code.0.len() > 0, Error::<T>::EmptyCode);
+		ensure!(code.0.len() <= config.max_code_size as usize, Error::<T>::CodeTooLarge);
```
bkchr, line 690: Looks like this should be done inside of `schedule_code_upgrade`?

bkchr (commented):
I would move the check, otherwise it looks good.

seadanda (approved):
LGTM

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1230,6 +1234,10 @@ impl<T: Config> Pallet<T> {
 		// Check that we can schedule an upgrade at all.
 		ensure!(Self::can_upgrade_validation_code(id), Error::<T>::CannotUpgradeCode);
 		let config = configuration::Pallet::<T>::config();
+		// Validation code sanity checks:
+		ensure!(new_code.0.len() > 0, Error::<T>::EmptyCode);
```
s0me0ne-unkn0wn, line 1238: A compressed blob should contain at least 8 bytes of ZSTD signature and 1 byte of payload, thus 9 bytes.

If it's uncompressed, it should contain a 4-byte Wasm magic, a 4-byte version, and at least one section.

Thus, it's safe to assume the minimal possible length of valid code is 9 bytes.

But I'm also okay with keeping this a more general "just-non-zero" check.

s0me0ne-unkn0wn (approved):
Looks good! Left a nit suggestion, not a blocker.

polkadot/runtime/parachains/src/paras/mod.rs
ggwpez: Such assumptions should then probably be checked by the `ValidationCode` type itself, maybe with some `check_sanity` function.  
Accessing the `.0` here and making assumptions about it seems unclean to me.

s0me0ne-unkn0wn: Fair enough! I meant just some stricter checks, but from the design point of view, you have totally valid point here.

eskimor (approved):

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1230,6 +1232,10 @@ impl<T: Config> Pallet<T> {
 		// Check that we can schedule an upgrade at all.
 		ensure!(Self::can_upgrade_validation_code(id), Error::<T>::CannotUpgradeCode);
 		let config = configuration::Pallet::<T>::config();
+		// Validation code sanity checks:
+		ensure!(new_code.check_sanity(), Error::<T>::InvalidCode);
+		ensure!(new_code.0.len() <= config.max_code_size as usize, Error::<T>::InvalidCode);
```
bkchr, line 1237: I still think that this should be moved to `schedule_code_upgrade`.

bkchr (approved):
Ty! Still some benchmarks are failing.

Szegoo:
> Still some benchmarks are failing.

🙈Should be good now.

PR #3234, opened 2024-02-06
[ref-docs]: Improve ref-docs for signed extensions

jsdw:
The goals of this signed extension ref doc is to provide an overview of signed extensions, with links to various examples (eg current signed extensions and the trait itself) to learn more. This page is linked from the `extrinsic_encoding` docs so that, when from learning how to encode signed extensions as part of extrinsics there, you can come here for a bit of a deeper look into what they are and some examples showing how they are useful.

I did find myself struggling to decide what was worth writing here though as opposed to on the trait etc itself. I also am not an expert on writing signed extensions (being more of a consumer/encoder of them from client code), but nevertheless I thought I'd have a stab at this :)

Feedback welcome!

Closes https://github.com/paritytech/polkadot-sdk-docs/issues/42

### DIFF
--- a/docs/sdk/src/reference_docs/signed_extensions.rs
+++ b/docs/sdk/src/reference_docs/signed_extensions.rs
@@ -5,7 +5,6 @@
 //! extrinsic that is submitted.
 //!
 //! Signed extensions [implement the following trait][sp_runtime::traits::SignedExtension]:
-//!
 #![doc = docify::embed!("../../substrate/primitives/runtime/src/traits.rs", SignedExtension)]
 //!
 //! Tuples of signed extensions automatically implement this trait too, which allows them to be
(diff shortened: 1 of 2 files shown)


kianenigma (commented):
I don't mind this, but I was thinking maybe we should also wait for https://github.com/paritytech/polkadot-sdk/pull/2280, and then revamp this page? cc @georgepisaltu wdyt? Do you think is it worth teaching people about signed extensions at this point?

jsdw:
> I think waiting for TransactionExtension to drop and revamping this is the best course of action. A very good starting point is the PR description of https://github.com/paritytech/polkadot-sdk/pull/2280. I will also review the proposed docs here as I am sure there will be bits and pieces that will still be relevant.

Yup, I agree; I've looked through the new PR and related bits, and while I'm sure bits are still relevant, I'm happy to wait until that stuff merges and then have another look and update the details/examples/etc to line up with the new stuff :)

Issue #3178, opened 2024-02-01
Refactor staking test modules into own files


gpestana:
The staking `tests.rs` has about 7k LOC. Most of the tests are nicely organised into modules, consider refactor each mod into a new file for readability.

gpestana:
ftr, @vicenterocha is interested in picking up this issue 🚀

gpestana:
Note: this refactoring is probably better suited to be done as part of the multi-block revamp and once https://github.com/paritytech/polkadot-sdk/pull/1933 is merged.

rossbulat:
This is now stale as work is underway to tidy up the runtimes post AH migration (https://github.com/paritytech/polkadot-sdk/pull/11940)

@vicenterocha What do you think about closing this issue now? My concern is that it may otherwise give contributors the impression that it should be worked on. Checking `substrate/frame/staking-async/src/tests` also, looks like tests have already been modularised.

cc @Ank4n

sigurpol:
I fully agree with you @rossbulat  as per my comment [here](https://github.com/paritytech/polkadot-sdk/pull/11129#issuecomment-4649328153) and I am in favor of closing this issue.
Tests on staking-async can surely benefit of extra love and can be improved, but at least the modularization issue seems indeed gone.

PR #3235, opened 2024-02-06
Refactor staking test modules into their own files

vicenterocha:
This PR refactors substrate staking tests, as detailed in #3178.

The idea is to do this in 2 PRs. On this one:
1. Move all mod tests from `tests.rs` into their own test file under a new `tests/` folder
2. Move all non mod tests from `tests.rs` into `tests/mod.rs`

Even we this changes the `tests/mod.rs` is still quiet big (6k LOC). Next PR should address it by extrating `tests/mods.rs` into further submodules.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5136778

gpestana:
@vicenterocha thanks for the PR! One quick comment, we try to avoid rewriting history/force push on intermediate commits on open PRs -- especially on PRs ready to review, which is not the case here but it's still better to not squash int commits in general. The commits will be squashed before merging so no problem with adding a bunch of commits to the PR.

substrate/frame/staking/src/tests/election_data_provider.rs
```diff
@@ -0,0 +1,433 @@
+use super::*;
```
gpestana, line 1: All the new files are missing the Substrate header (see e.g. [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/staking/src/slashing.rs#L1-L15) or in any other rs file under `polkadot-sdk/substrate/`).

vicenterocha: actually had them before I started this all over again 🙈

Ank4n:
@vicenterocha thanks for the contribution. Quick question, when I run these tests (`cargo test -p pallet-staking`) I see one less test being ran than the previous count (Old: 156, New: 155). Was there a test you removed or?

vicenterocha:
@Ank4n thanks for having a look. Running tests on this branch actually gives me 156, whereas on master it gives me 155.

Screenshot 2024-03-02 at 22 20 49


Screenshot 2024-03-02 at 22 23 03


I've compare both runs and it seems like this is the the additional test:
`test mock::test_genesis_config_builds ... ok`


I can provide a full test output of both master and current branch.

Ank4n:
> @Ank4n thanks for having a look. Running tests on this branch actually gives me 156, whereas on master it gives me 155.
> 
> **refactor-staking-tests:** Screenshot 2024-03-02 at 22 20 49
> 
> **master:** Screenshot 2024-03-02 at 22 23 03
> 
> I've compare both runs and it seems like this is the the additional test: `test mock::test_genesis_config_builds ... ok`
> 
> I can provide a full test output of both master and current branch.
> 
> Not sure exactly how this test is defined though.

Thanks for investigating it. The test was [added](https://github.com/paritytech/polkadot-sdk/pull/3412) couple of weeks ago. My guess is I ran the test with your branch that was not updated with master which is why I missed one. And vice-versa for you, your master might be outdated.

Ank4n (approved):

vicenterocha:
Running on this branch

Screenshot 2024-04-14 at 17 36 43


Running on master

image

gpestana:
@vicenterocha thanks a lot for helping with this work, we have a bunch of changes coming up related to the pallet-staking with the multi-block election work and I think it's better to refactor all the tests during that work. In any case, I can point you some other issues to help with if you want. Thanks :)

PR #3236, opened 2024-02-07
Add forklift to remaining jobs

AndWeHaveAPlan:
Add [forklift caching](https://gitlab.parity.io/parity/infrastructure/ci_cd/forklift/forklift) to remainig jobs

by .sh and .py scripts:
- cargo-check-each-crate x6 (`.gitlab/check-each-crate.py`)
- build-linux-stable (`polkadot/scripts/build-only-wasm.sh`)

by before_script:
- build-linux-substrate
- build-subkey-linux (with `.build-subkey` job)
- cargo-check-benches x2

**To disable feature set FORKLIFT_BYPASS variable to true in [project settings in gitlab](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/settings/ci_cd)**
(forklift now handles FORKLIFT_BYPASS by itself)

### DIFF
--- a/.gitlab/check-each-crate.py
+++ b/.gitlab/check-each-crate.py
@@ -55,7 +55,7 @@ for i in range(0, crates_per_group + overflow_crates):
 
 	print(f"Checking {crates[crate][0]}", file=sys.stderr)
 
-	res = subprocess.run(["cargo", "check", "--locked"], cwd = crates[crate][1])
+	res = subprocess.run(["forklift", "cargo", "check", "--locked"], cwd = crates[crate][1])
 
 	if res.returncode != 0:
 		sys.exit(1)
--- a/polkadot/scripts/build-only-wasm.sh
+++ b/polkadot/scripts/build-only-wasm.sh
@@ -13,6 +13,14 @@ fi
 
 WASM_BUILDER_RUNNER="$PROJECT_ROOT/target/release/wbuild-runner/$1"
 
+fl_cargo () {
+    if command -v forklift >/dev/null 2>&1; then
+        forklift cargo "$@";
+    else
+        cargo "$@";
+    fi
+}
+
 if [ -z "$2" ]; then
   export WASM_TARGET_DIRECTORY=$(pwd)
 else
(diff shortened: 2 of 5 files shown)


.gitlab/check-each-crate.py
```diff
@@ -55,7 +55,7 @@
 
 	print(f"Checking {crates[crate][0]}", file=sys.stderr)
 
-	res = subprocess.run(["cargo", "check", "--locked"], cwd = crates[crate][1])
+	res = subprocess.run(["forklift", "cargo", "check", "--locked"], cwd = crates[crate][1])
```
ggwpez, line 58: Does the forklift have a `--quiet` option or something?  
I dont know if the noise here comes from forklift maybe https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5143188 but its difficult to see why the job failed.

AndWeHaveAPlan: errors start at line [3708](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5143188#L3708) with ```error[E0599]: the function or associated item `new` exists for struct `BenchmarkingState<Block<ExtrinsicWrapper<u64>>>`, but its trait bounds were not satisfied```

but yeah, that output seems redundant now, I'll fix it

FYI there are no flags as settings provided by config file. But it is possible to set vars by environment
`FORKLIFT_general_logLevel="warn"`

AndWeHaveAPlan: fixed, from now on forklift will only write simple startup message and status report at the end (and default logLevel is now `warn`)

rcny (approved):

alvicsam (approved):

ggwpez (approved):
Approving to unblock.

polkadot/scripts/build-only-wasm.sh
```diff
@@ -22,8 +22,8 @@ fi
 if [ -d $WASM_BUILDER_RUNNER ]; then
   export DEBUG=false
   export OUT_DIR="$PROJECT_ROOT/target/release/build"
-  cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" \
+  forklift cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" \
     | grep -vE "cargo:rerun-if-|Executing build command"
 else
-  cargo build --release -p $1
+  forklift cargo build --release -p $1
```
ggwpez, line 28: Actually i think this script is meant for endusers - not for CI.  
Can you maybe run it through the `RUSTC_WRAPPER` env variable like `RUSTC_WRAPPER=forklift ./build-only-wasm.sh`?

AndWeHaveAPlan: Makes sense. But wrappers need a "coordinator"  #process, and I don't like the idea of copy/paste script just for ci, so I added check for forklift

ggwpez (approved):

polkadot/scripts/build-only-wasm.sh
ggwpez: Yup sounds good.

PR #3241, opened 2024-02-07
Contracts update doc.rs metadata

pgherveou:
Adding Rust metadata for doc
see https://docs.rs/about/metadata


substrate/frame/contracts/uapi/Cargo.toml
```diff
@@ -23,6 +23,9 @@ scale = { package = "parity-scale-codec", version = "3.6.1", default-features =
 [target.'cfg(target_arch = "riscv32")'.dependencies]
 polkavm-derive = '0.5.0'
 
+[package.metadata.docs.rs]
+targets = ["wasm32-unknown-unknown", "x86_64-unknown-linux-gnu"]
```
athei, line 27: Is there any benefit of building it for multiple targets?

```suggestion
default-target = ["wasm32-unknown-unknown"]
```

pgherveou: yeah probably not. let's keep it simple

athei (approved):

PR #3243, opened 2024-02-07
contracts: Don't fail fast if the `Weight` limit of a cross contract call is too big

athei:
When doing a cross contract call you can supply an optional Weight limit for that call. If one doesn't specify the limit (setting it to 0) the sub call will have all the remaining gas available. If one does specify the limit we subtract that amount eagerly from the  Weight meter and fail fast if not enough `Weight` is available.

This is quite annoying because setting a fixed limit will set the `gas_required` in the gas estimation according to the specified limit. Even if in that dry-run the actual call didn't consume that whole amount. It effectively discards the more precise measurement it should have from the dry-run.

This PR changes the behaviour so that the supplied limit is an actual limit: We do the cross contract call even if the limit is higher than the remaining `Weight`. We then fail and roll back in the cub call in case there is not enough weight. 

This makes the weight estimation in the dry-run no longer dependent on the weight limit supplied when doing a cross contract call.

### DIFF
--- a/substrate/frame/contracts/src/tests.rs
+++ b/substrate/frame/contracts/src/tests.rs
@@ -3063,10 +3063,10 @@ fn gas_estimation_for_subcalls() {
 						Determinism::Enforced,
 					)
 					.result,
-					error.clone(),
+					error,
 				);
 
-				// Check that it fails with too litte proof_size
+				// Check that it fails with too little proof_size
 				assert_err!(
 					Contracts::bare_call(
 						ALICE,
--- a/substrate/frame/contracts/src/tests.rs
+++ b/substrate/frame/contracts/src/tests.rs
@@ -3049,7 +3049,7 @@ fn gas_estimation_for_subcalls() {
 					.result
 				);
 
-				// Check that it fails with too litte ref_time
+				// Check that it fails with too little ref_time
 				assert_err!(
 					Contracts::bare_call(
 						ALICE,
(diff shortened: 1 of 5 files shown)


substrate/frame/contracts/src/tests.rs
```diff
@@ -2938,91 +2937,159 @@ fn gas_estimation_nested_call_fixed_limit() {
...
+						Some(result.storage_deposit.charge_or_zero()),
+						input.clone(),
+						DebugInfo::Skip,
+						CollectEvents::Skip,
+						Determinism::Enforced,
+					)
+					.result
+				);
+
+				// Check that it fails with too litte ref_time
```
pgherveou, line 3052:
```suggestion
				// Check that it fails with too little ref_time
```

substrate/frame/contracts/src/tests.rs
```diff
@@ -2938,91 +2937,159 @@ fn gas_estimation_nested_call_fixed_limit() {
...
+						input.clone(),
+						DebugInfo::Skip,
+						CollectEvents::Skip,
+						Determinism::Enforced,
+					)
+					.result,
+					error.clone(),
+				);
+
+				// Check that it fails with too litte proof_size
```
pgherveou, line 3069:
```suggestion
				// Check that it fails with too little proof_size
```

pgherveou (approved):

substrate/frame/contracts/src/tests.rs
```diff
@@ -2938,91 +2937,159 @@ fn gas_estimation_nested_call_fixed_limit() {
...
+						0,
+						result.gas_required.sub_ref_time(1),
+						Some(result.storage_deposit.charge_or_zero()),
+						input.clone(),
+						DebugInfo::Skip,
+						CollectEvents::Skip,
+						Determinism::Enforced,
+					)
+					.result,
+					error.clone(),
```
pgherveou, line 3066:
```suggestion
					error,
```

PR #3250, opened 2024-02-08
Asset Conversion: Pool Account ID derivation with additional Pallet ID seed

muharem:
Introduce `PalletId` as an additional seed parameter for pool's account id derivation.

The PR also introduces the `pallet_asset_conversion_ops` pallet with a call to migrate a given pool to thew new account. Additionally `fungibles::lifetime::ResetTeam` and `fungible::lifetime::Refund` traits, to facilitate the migration of pools.

PR #3251, opened 2024-02-08
Asset Conversion: Pool Touch Call

muharem:
Introduce `touch` call designed to address operational prerequisites before providing liquidity to a pool. 

This function ensures that essential requirements, such as the presence of the pool's accounts, are fulfilled. It is particularly beneficial in scenarios where a pool creator removes the pool's accounts without providing liquidity.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5147335

muharem:
bot help

muharem:
bot bench substrate-pallet --pallet=pallet_asset_conversion

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
```diff
@@ -32,6 +32,18 @@ pub trait Create<AccountId>: Inspect<AccountId> {
 	) -> DispatchResult;
 }
 
+/// Trait for refunding the deposit of a target asset account.
+pub trait Refund<AccountId> {
```
georgepisaltu, line 36: Shouldn't this be `trait Refund<AccountId>: Inspect<AccountId>` to avoid defining `AssetId` and `Balance` again?

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
```diff
@@ -32,6 +32,18 @@ pub trait Create<AccountId>: Inspect<AccountId> {
...
+	/// Means of identifying one asset class from another.
+	type AssetId: AssetId;
+	/// Scalar type for representing balance of an account.
+	type Balance: Balance;
+	/// Returns the amount of account deposit and depositor address, if any.
+	fn deposit_held(id: Self::AssetId, who: AccountId) -> Option<(AccountId, Self::Balance)>;
```
georgepisaltu, line 42: I think this function is unnecessary in the `Refund` interface. Having `fn refund` should be enough, the other `fungibles` trait impls can be used to determine the deposits.

substrate/frame/assets/src/impl_fungibles.rs
```diff
@@ -308,3 +308,23 @@ impl<T: Config<I>, I: 'static> fungibles::InspectEnumerable<T::AccountId> for Pa
...
+	fn deposit_held(id: Self::AssetId, who: T::AccountId) -> Option<(T::AccountId, Self::Balance)> {
+		use ExistenceReason::*;
+		match Account::<T, I>::get(&id, &who).ok_or(Error::<T, I>::NoDeposit).ok()?.reason {
+			DepositHeld(b) => Some((who, b)),
+			DepositFrom(d, b) => Some((d, b)),
+			_ => None,
+		}
+	}
+	fn refund(id: Self::AssetId, who: T::AccountId) -> DispatchResult {
+		match Self::deposit_held(id.clone(), who.clone()) {
```
georgepisaltu, line 324: I think calling `deposit_held` here only complicates things, especially because you don't care about the actual value of the deposit at this level, only if it is provided by someone else. It would be simpler to move the logic from `deposit_held` above here.

georgepisaltu (commented):
Overall looks good, a couple of comments.

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
muharem: I think it is more flexible, refund type do not have to implement the inspect trait.

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
muharem: `fungibles` traits have no function to check the deposit held for an asset's account existence.

substrate/frame/assets/src/impl_fungibles.rs
muharem: I will duplicate that match and get from above. No strong opinion though

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
muharem: Here is the example where it is used 
https://github.com/paritytech/polkadot-sdk/pull/3250/files#diff-fbf88fab34bcf066e3c722cd1345365529c1c652df82604ba23aea4f047993d5R243
https://github.com/paritytech/polkadot-sdk/blob/05e639d4bcc3b4d6f99b3ff386155c7dc8115518/substrate/frame/asset-conversion/ops/src/lib.rs#L182

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
muharem: > Is there ever a usecase where `Refund` would be implemented on something other than the thing implementing `Inspect`.

Not likely in the context of the FRAME. I really see the point for super traits when some of the functions/methods used in a sub trait. But when need to decide convenience (no needs for type definition) vs flexibility (interface segregation), I will go for flexibility. 
In this case it's not that important, perhaps `Refund` wont need to be implemented on a type not implementing `Inspect`. I am fine with super trait.

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
georgepisaltu: I was indeed missing something because that's from another PR :grin: .

I still feel this functionality for querying the deposit shouldn't be in a `Refund` trait, but rather in some other parent interface to separate these concerns. In practical terms though, it doesn't matter too much. In the future, if we need to separate them, it will be very easy to do so.

substrate/frame/support/src/traits/tokens/fungibles/lifetime.rs
muharem: @georgepisaltu I had to set an additional bound to Currency type in Assets pallet for this. Please check and approve if you happy with it.
https://github.com/paritytech/polkadot-sdk/pull/3251/commits/290c0f1e84597374f6760e28aff731f96f2b4826

muharem: however I do not see the benefits. in most cases you still have to define what Balance and AssetId are. and not you can not have a separate type implementing only refund.

georgepisaltu (approved):

substrate/frame/assets/src/lib.rs
```diff
@@ -1697,7 +1697,9 @@ pub mod pallet {
 
 		fn should_touch(asset: T::AssetId, who: &T::AccountId) -> bool {
 			match Asset::<T, I>::get(&asset) {
+				// refer to the [`Self::new_account`] function for more details.
 				Some(info) if info.is_sufficient => false,
+				Some(_) if frame_system::Pallet::<T>::can_accrue_consumers(who, 2) => false,
```
ggwpez, line 1702: Why does it only touch when it can still accrue consumers? I assume someone could send an asset to this account to use up another consumer just after this was checked here?

muharem: Yes, someone can, and from that moment, a creation of non sufficient account would require a deposit. The result is only valid for a current transaction. 
Not sure I got the question right. Here the doc for the trait - https://github.com/paritytech/polkadot-sdk/blob/0d71753e0eb1c0fdbeb09af5db7a2e1b5f150b13/substrate/frame/support/src/traits/misc.rs#L1190

georgepisaltu (approved):

substrate/frame/support/src/traits/tokens/fungible/union_of.rs
```diff
@@ -925,3 +925,28 @@ impl<
 		}
 	}
 }
+
+impl<
+		Left: fungible::Inspect<AccountId>,
+		Right: fungibles::Inspect<AccountId> + fungibles::Refund<AccountId>,
```
ggwpez, line 931: Why is this adapter requiring heterogeneous types for Left and Right?  
Why is the left side just assumed to not be Refund?

ggwpez (approved):

substrate/frame/support/src/traits/tokens/fungible/union_of.rs
muharem: yes, we ether assume there is no refund for `fungible`, or we introduce a no-op impl for `fungible`, which is balance pallet,  with asset id `()`, or we have to introduce one more refund trait without asset id. Given this, I made a chose for no-op in adapter (no refund trait for fungible, no adapter for it). I am open to go for another solution, as I do not see an obvious winner here.

PR #3254, opened 2024-02-08
[Backport] Version bumps  from 1.7.0 release

EgorPopelyaev:
This PR backports version bumps from `1.7.0` release branch and moves related prdoc files to the appropriate folder.


joepetrowski (approved):

cumulus/polkadot-parachain/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "polkadot-parachain-bin"
-version = "4.0.0"
+version = "1.7.0"
```
liamaharon, line 3: why are these versions regressing?

liamaharon: we've already published 4.0 for this crate https://crates.io/crates/polkadot-parachain-bin/versions

EgorPopelyaev: We used to bump it like the node release ones. But if it is not needed during the node release, probably, it should be bumped during the crates.io release

EgorPopelyaev: I had the same question about the [workspace](https://github.com/paritytech/polkadot-sdk/blob/2556e33fb4c52ea8192de68df33c151768edc2a9/polkadot/Cargo.toml#L21) version for polkadot

Morganamilo (approved):

cumulus/polkadot-parachain/Cargo.toml
liamaharon: > We used to bump it like the node release ones. But if it is not needed during the node release, probably, it should be bumped during the crates.io release

I'm not sure what you mean, could you please elaborate?

EgorPopelyaev: I mean that we have some bumps during the node release like the `Node_Version` or `spec_version`. And there are also version changes when the crates.io release is done. Those two versions (cumulus one and the polkadot workspace) we used to bump during the node release. So should it be done that way or should I ignore those, and they will be bumped during the crates.io release which is done by Morgan.

liamaharon: what's the reason for not leaving the cargo.toml version consistent with crates.io?

ggwpez: I think this version means nothing. There is another version in the code to set the node version @skunert or someone from @paritytech/sdk-node should know, just ping them.

ggwpez:
Where are the version bumps for all other crates?

cumulus/polkadot-parachain/Cargo.toml
EgorPopelyaev: I guess there is no reason for it. It is built in the release scripts as we used to do it before. I'll revert it and adjust the scripts.

EgorPopelyaev:
@Morganamilo the creates release is not done yet, rigth?

ggwpez:
> @Morganamilo the creates release is not done yet, right?

Okay then lets do that in another MR.

ggwpez (approved):

PR #3275, opened 2024-02-09
Asset Conversion: doc guide to address incorrect exchange rate for non-withdrawable pool

muharem:
This documentation guide to address an incorrect exchange rate for a non-withdrawable pool.


substrate/frame/asset-conversion/src/lib.rs
```diff
@@ -413,6 +413,11 @@ pub mod pallet {
 		/// Params `amount1_min`/`amount2_min` represent that.
 		/// `mint_to` will be sent the liquidity tokens that represent this share of the pool.
 		///
+		/// NOTE: when encountering an incorrect exchange rate and non-withdrawable pool liquidity,
```
seadanda, line 416: This is totally subjective but it might be better to say:
```suggestion
		/// NOTE: to rectify an incorrect exchange rate and non-withdrawable pool liquidity,
```

Which I think makes it a bit clearer.

seadanda (approved):

joepetrowski (approved):

substrate/frame/asset-conversion/src/lib.rs
seadanda: Apologies, non-withdrawable pool liquidity can't be rectified as you pointed out. Wording is fine as is

PR #3286, opened 2024-02-12
Assets: can_decrease/increase for destroying asset is not successful

muharem:
Functions `can_decrease` and `can_increase` do not return successful consequence results for assets undergoing destruction; instead, they return the `UnknownAsset` consequence variant.

This update aligns their behavior with similar functions, such as `reducible_balance`, `increase_balance`, `decrease_balance`, and `burn`, which return an `AssetNotLive` error for assets in the process of being destroyed.


acatangiu (approved):

substrate/frame/assets/src/functions.rs
```diff
@@ -132,6 +132,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			Some(details) => details,
 			None => return DepositConsequence::UnknownAsset,
 		};
+		if details.status == AssetStatus::Destroying {
```
georgepisaltu, line 135: nit: I'd use `matches!` instead of `==`. Same for usage below.

georgepisaltu (approved):

joepetrowski (approved):

substrate/frame/assets/src/functions.rs
muharem: what's the benefit? with `==` I follow the style within the function

georgepisaltu: Pattern matching is the standard way of comparing `enum`s and it doesn't require `PartialEq`/`Eq`, but in this case it's not a big deal.

liamaharon (approved):

bkchr:
@muharem what is the status of this pr?

muharem:
> @muharem what is the status of this pr?

I did add it to the backlog for audit. But it is ready to be merged.

muharem:
bot merge

Polkadot-Forum:

https://forum.polkadot.network/t/discussion-treating-changes-in-dispatchable-function-errors-as-breaking-changes/11222/1

Issue #3269, opened 2024-02-08
SIMD removal from rust


driemworks:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

SIMD was removed from the latest rust nightly version https://github.com/rust-lang/rust/pull/117372 which causes ahash 0.7.7 to break: https://github.com/tkaitchuck/aHash/issues/200. This has been resolved in versions >= 0.8.X https://github.com/tkaitchuck/aHash/pull/183. 

- run `cargo +nightly tree -i ahash@0.7.7` to see the full dependency graph. 
- more info: https://users.rust-lang.org/t/error-e0635-unknown-feature-stdsimd/106445

``` sh
error[E0635]: unknown feature `stdsimd`
  --> /home/XXXX/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.7.7/src/lib.rs:33:42
   |
33 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
   |                                          ^^^^^^^

For more information about this error, try `rustc --explain E0635`.
error: could not compile `ahash` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

### Steps to reproduce

1. Clone the polkadot-sdk master branch
2. Update to the latest rust nightly version (as of this comment: `nightly-x86_64-unknown-linux-gnu - rustc 1.78.0-nightly (8ace7ea1f 2024-02-07)`)
3. Build the project

``` sh
git clone https://github.com/paritytech/polkadot-sdk.git
rustup update nightly
cargo +nightly build --release
```

ggwpez:
Hm, we can update criterion which seems to pull it in a lot and maybe some other deps.  
Note that we normally use stable and not nightly to build, but its indeed a bummer.

ggwpez:
Other ideas @bkchr ?

burdges:
https://github.com/rust-lang/rust/issues/27731#issuecomment-1937538034 

https://github.com/rust-lang/stdarch/pull/1486

https://github.com/rust-lang/rust/issues/120930

PR #3296, opened 2024-02-12
Fix compilation on latest nightly because of `stdsimd`

bkchr:
We need to bump `ahash` to make it compile again.

Closes: https://github.com/paritytech/polkadot-sdk/issues/3269


ggwpez (approved):
So they just fixed it in a patch version, although its a breaking change.  
Nice 😆

alexggh (approved):

bkchr:
> So they just fixed it in a patch version, although its a breaking change. Nice 😆

They used some nightly feature, which they should not be doing :see_no_evil:

sandreim (approved):

Issue #3028, opened 2024-01-23
rpc server: update jsonrpsee to v0.21 and provide rate limiting middleware


niklasad1:
Superseeds #47

The rate-limiting in jsonrpsee has been replaced by JSON-RPC specific middleware and backpressure. 

The way forward is that substrate should provide rate limiting and additional APIs for builders to provide their own custom middleware for similar things.

Thus, what I have in mind is that substrate/polkadot-sdk should provide the following:

- New CLI option to enable global rate limiting such `--rpc-rate-limit <CALLS_PER_MINUTE>`
- Rewrite RpcMetrics as JSON-RPC specific middleware

### Alternatives

https://github.com/AcalaNetwork/subway is a RPC proxy that does similar things but has more functionality but I reckon it may be nice for builders to have something simple to use in polkadot-sdk useful.

niklasad1:
This issue https://github.com/paritytech/jsonrpsee/issues/1264 is blocker to port over the prometheus metrics for `WS_SESSIONS`

niklasad1:
Closed by #3301

PR #3301, opened 2024-02-12
rpc server: add rate limiting middleware

niklasad1:
Add RPC server rate limiting which can be utilized by the CLI  `--rpc-rate-limit <calls/per minute>`

Resolves first part of https://github.com/paritytech/polkadot-sdk/issues/3028

//cc @PierreBesson @kogeler  you might be interested in this one


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5217870

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5220070

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5220217

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5220485

substrate/client/rpc-servers/src/middleware/mod.rs
```diff
@@ -16,8 +16,12 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-//! JSON-RPC specific middleware.
+//! RPC middleware to collect prometheus metrics on RPC calls.
```
niklasad1, line 19:
```suggestion
//! JSON-RPC specific middleware.
```

Cargo.toml
```diff
@@ -613,3 +613,4 @@ wasmi = { opt-level = 3 }
 x25519-dalek = { opt-level = 3 }
 yamux = { opt-level = 3 }
 zeroize = { opt-level = 3 }
+
```
niklasad1, line 616:
```suggestion
```

substrate/client/rpc-servers/src/middleware/rate_limit.rs
```diff
@@ -0,0 +1,80 @@
...
+pub struct RateLimitLayer(governor::Quota);
+
+impl RateLimitLayer {
+	/// Create new rate limit enforced per minute.
+	///
+	/// # Panics
+	///
+	/// Panics if n is zero.
+	pub fn per_minute(n: u32) -> Self {
+		Self(governor::Quota::per_minute(NonZeroU32::new(n).unwrap()))
```
niklasad1, line 45: This can only happen if  `--rpc-rate-limit 0` which doesn't make any sense, I could refactor this check that n > 0 in the code or just disable the rate-limit for that case.

Any suggestions?

prdoc/pr_3301.prdoc
```diff
@@ -0,0 +1,11 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: rpc server add rate limiting.
+
+doc:
+  - audience: Node Operator
+    description: |
+      Add rate limiting for RPC server which can be utilized by the CLI `--rpc-rate-limit <calls/per minute>`
```
jsdw, line 9:
```suggestion
      Add rate limiting for RPC server which can be utilized by the CLI `--rpc-rate-limit <calls per minute>`
```

substrate/client/rpc-servers/src/middleware/rate_limit.rs
jsdw: Could the CLI arg take in an `Option<NonZeroU32>` instead of `Option<u32>` to prevent this from being a possibility?

niklasad1: yepp, since no other CLI stuff is using NonZeroInteger types  I wanted to be consistent but worth breaking that.

substrate/client/rpc-servers/src/middleware/rate_limit.rs
```diff
@@ -0,0 +1,80 @@
...
+	S: Send + Sync + RpcServiceT<'a> + Clone + 'static,
+{
+	type Future = BoxFuture<'a, MethodResponse>;
+
+	fn call(&self, req: Request<'a>) -> Self::Future {
+		let rate_limit = self.rate_limit.clone();
+		let service = self.service.clone();
+
+		async move {
+			// Random delay between 0-50ms to poll waiting futures until completion.
```
jsdw, line 70: Maybe a bit more descriptive would be:

```suggestion
			// Wait until this connection is no longer being rate limited. This adds 
			// up to 50ms of jitter in an attempt to avoid thundering herds.
```

I guess this implementation will keep every pending request in memory at this point while it waits to be serviced. Should we just outright reject some requests though?

Looking at the source for `until_ready_with_jitter`, I can see that the error case tells you when the request might next be allowed through, and perhaps we should start rejecting requests that would have to wait for more than some certain amount of time (eg a minute) just to cap how many requests can ever be left waiting?

Or maybe we have another way to deal with such things already elsewhere?

jsdw (approved):
Looks good to me overall, and the governor crate looks pretty nice! Just a question on the rate limiting policy and whether we should be dropping requests past some limit :)

substrate/client/rpc-servers/src/middleware/rate_limit.rs
niklasad1: good point, let's keep it simple for now and just deny when the rate limit is exceeded.

jsdw: I think you'd do something like `tokio::time::sleep(jitter + rejected.wait_time_from(std::time::Instant::now()))` to copy their approach but with a tokio timer instead of the futures one they use.

I guess with the above code a request wakes up when it might _possibly_ be ready (plus some jitter) and then if it's not actually possible to continue yet (eg because other requests came in), you loop around and give up after 10 tries.

I think that works for me! I'm not sure what the number of attempts should be but at least this sort of thing means we eventually give up on things :)

niklasad1: yeah, what I did above is equivalent to what the governor does but when I tried it locally with rate-limit 10 and it ended rejecting the 11th call quite fast.

As the wakeup is order of milliseconds but the rate-limit period is per minute and I'm still unsure whether what's reasonable limit.

jsdw: Ok yup that sounds good; the way you did it originally is very simple at least, and still an improvement, so happy to start with that and iterate if needed :)

substrate/client/rpc-servers/src/lib.rs
```diff
@@ -153,7 +170,10 @@ pub async fn start_server<M: Send + Sync + 'static>(
 				let transport_label = if is_websocket { "ws" } else { "http" };
 
 				let metrics = metrics.map(|m| MetricsLayer::new(m, transport_label));
-				let rpc_middleware = RpcServiceBuilder::new().option_layer(metrics.clone());
+				let rate_limit = rate_limit.map(|r| RateLimitLayer::per_minute(r));
+				let rpc_middleware =
+					RpcServiceBuilder::new().option_layer(rate_limit).option_layer(metrics.clone());
```
lexnv, line 175: Do we want to add rate-limited methods to the call metrics? I think it might still be useful to report them, wdyt?

If so, we should register the `metrics` as an optional layer first.

lexnv (approved):
LGTM!

substrate/client/rpc-servers/src/middleware/rate_limit.rs
niklasad1: yeah, ok it works now with proper clock source

substrate/client/rpc-servers/src/lib.rs
niklasad1: yupp, good catch

PR #3310, opened 2024-02-13
Bridges: add test 0002 to CI

serban300:
Bridges: add test 0002 to CI

### DIFF
--- a/bridges/zombienet/tests/0002-mandatory-headers-synced-while-idle/run.sh
+++ b/bridges/zombienet/tests/0002-mandatory-headers-synced-while-idle/run.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+source "${BASH_SOURCE%/*}/../../utils/common.sh"
+source "${BASH_SOURCE%/*}/../../utils/zombienet.sh"
+
+# We use `--relayer-delay` in order to sleep some time before starting relayer.
+# We want to sleep for at least 1 session, which is expected to be 60 seconds for test environment.
+${BASH_SOURCE%/*}/../../environments/rococo-westend/spawn.sh &
+env_pid=$!
+
+ensure_process_file $env_pid $TEST_DIR/rococo.env 400
+rococo_dir=`cat $TEST_DIR/rococo.env`
+echo
+
+ensure_process_file $env_pid $TEST_DIR/westend.env 180
+westend_dir=`cat $TEST_DIR/westend.env`
+echo
+
+# Sleep for some time before starting the relayer. We want to sleep for at least 1 session,
+# which is expected to be 60 seconds for the test environment.
+echo -e "Sleeping 90s before starting relayer ...\n"
+sleep 90
+${BASH_SOURCE%/*}/../../environments/rococo-westend/start_relayer.sh $rococo_dir $westend_dir relayer_pid
+
+# Sometimes the relayer syncs 2 parachain heads in the begining leading to test failures.
+# We do this as a workaround.
+# TODO: investigate. Not sure if it's expected.
+echo -e "Sleeping 90s before runing the tests ...\n"
+sleep 90
+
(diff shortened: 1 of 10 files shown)


serban300:
The PR is ready for review. @paritytech/bridges-core PTAL !

bridges/zombienet/tests/0002-mandatory-headers-synced-while-idle/run.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Sleep for some time before starting the relayer. We want to sleep for at least 1 session,
+# which is expected to be 60 seconds for the test environment.
+echo -e "Sleeping 90s before starting relayer ...\n"
+sleep 90
+${BASH_SOURCE%/*}/../../environments/rococo-westend/start_relayer.sh $rococo_dir $westend_dir relayer_pid
+
+# Sometimes the relayer syncs 2 parachain heads in the begining leading to test failures.
+# We do this as a workaround.
+# TODO: investigate. Not sure if it's expected.
```
svyatonik, line 29: That is the key point of this test :) If we really do that, we should investigate - we won't be refunding costs of such transactions and if they're submitted on a regular basis, we definitely should investigate. Let's at least submit an issue (with as steps to reproduce - IIUC we need to revert this sleep + revert `oldParachainHeaders` change in the js script) to not forget that.

svyatonik (approved):
Won't block this PR from merge, but if there's really an issue - let's investigate it soonish

bridges/zombienet/tests/0002-mandatory-headers-synced-while-idle/run.sh
serban300: Opened issue https://github.com/paritytech/parity-bridges-common/issues/2838 for investigating this.

With the `oldParachainHeaders` I was trying to fix another issue. Namely the fact that sometimes we might miss the parachains head update (if there's only 1) and we'll get `throw new Error("No bridged parachain headers imported");`. But now that you mentioned it, I realized it's not the best fix. Changed the approach.

Also removed the sleep and just commented the test. Basically, not adding it in the CI. Added a TODO to uncomment it after the issue is fixed.

svyatonik: Ok, cool, thank you! It could be enabled on CI imo if it works in its current form - then we can avoid pinging CI team for another approval :)

svyatonik: > With the oldParachainHeaders I was trying to fix another issue. Namely the fact that sometimes we might miss the parachains head update (if there's only 1) and we'll get throw new Error("No bridged parachain headers imported");. But now that you mentioned it, I realized it's not the best fix. Changed the approach.

But we should start relayer right before calling that js script, right?

svyatonik: Ah, maybe it takes some time for zombienet to start because you're starting `zombienet` just to check it. This way we may also accidentally miss the duplicate submissions though. But eventually we should see it (given multiple runs on CI)

serban300: Ok, good point ! Enabled the the test in CI then. And increased the sleep to 180s to make sure we don't hit the issue.

serban300: > Ah, maybe it takes some time for zombienet to start because you're starting `zombienet` just to check it. This way we may also accidentally miss the duplicate submissions though. But eventually we should see it (given multiple runs on CI)

Yes. Also we run the `zndsl`s sequentially. So when we get to `run_zndsl ${BASH_SOURCE%/*}/westend-to-rococo.zndsl $rococo_dir` already 5 minutes have passed because of the first zndsl. We could run them in parallel, but I would do it in a separate PR.

acatangiu (approved):

PR #3313, opened 2024-02-13
chainHead: Error on duplicate unpin hashes

lexnv:
This PR addresses an issue where calling chainHead_unpin with duplicate hashes could lead to unintended side effects. 

This backports: https://github.com/paritytech/json-rpc-interface-spec/pull/135

While at it, have added a test to check that the global reference count is decremented only once on unpin.

cc @paritytech/subxt-team

### DIFF
--- a/substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
+++ b/substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
@@ -759,9 +759,6 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionsInner<Block, BE> {
 			return Err(SubscriptionManagementError::SubscriptionAbsent)
 		};
 
-		// Keep only unique hashes.
-		let hashes = hashes.into_iter().collect::<HashSet<_>>();
-
 		// Ensure that all blocks are part of the subscription before removing individual
 		// blocks.
 		for hash in &hashes {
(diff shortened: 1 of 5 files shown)


substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
```diff
@@ -759,10 +759,13 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionsInner<Block, BE> {
 			return Err(SubscriptionManagementError::SubscriptionAbsent)
 		};
 
+		// Keep only unique hashes.
+		let hashes = hashes.into_iter().collect::<HashSet<_>>();
```
bkchr, line 763: Maybe we should error on duplicated hashes instead?

lexnv: Yep that sounds good to me, I'll create a small PR in the spec to clarify this (part of https://github.com/paritytech/json-rpc-interface-spec/issues/133)

bkchr (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5209989

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5209987

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5209988

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5209970

tadeohepperle (approved):
Nice! :)

jsdw (approved):

davxy (approved):
Just curious. What is the unintended consequence if there are duplicate hashes?

I ask because the only thing you do after the check is to call `self.subscriptions.unpin_blocks()`. And if this is where the unintended behavior is triggered maybe you should add the check there (i.e. in the called function) as it may be called somewhere else

lexnv:
The side-effect before this PR was that duplicate hashes could decrement the global counter of the blocks multiple times:

https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs#L778-L781


Have moved the duplicate check in the `unpin_blocks` logic.

With this check enforced, we'll return an error before doing any other work. The downside of moving the check in the `unpin_blocks` is that we'll perform the check under a lock. However, considering that we don't expect a large number of blocks for the most common use-case; and that this closes the gap on potentially calling the `unpin_blocks` from other code paths; I'd say it is a reasonable thing to do.

Thanks for the review everyone! 🙏

PR #3318, opened 2024-02-14
Girazoki modify native asset to consider chain location

girazoki:
Modifies the behavior of the `NativeAsset` xcm filter to allow any asset that comes from the same origin. The motivation for this change is to be able to accept assets whose reserve chain is the one sending the message (origin):

Before this change:
```
- origin: (1, Parachain(100))
- asset: (1, Parachain(100))
- result: accepted
```

```
- origin: (1, Parachain(100))
- asset: (1, Parachain(100), GeneralIndex(1))
- result: **not accepted**
```

After this change:

```
- origin: (1, Parachain(100))
- asset: (1, Parachain(100))
- result: accepted
```

```
- origin: (1, Parachain(100))
- asset: (1, Parachain(100), GeneralIndex(1))
- result: **accepted**
```

I can this as a separate struct though if we feel like `NativeAsset` should just accept the asset if `origin === location`

### DIFF
--- a/polkadot/xcm/xcm-builder/src/filter_asset_location.rs
+++ b/polkadot/xcm/xcm-builder/src/filter_asset_location.rs
@@ -26,7 +26,7 @@ pub struct NativeAsset;
 impl ContainsPair<Asset, Location> for NativeAsset {
 	fn contains(asset: &Asset, origin: &Location) -> bool {
 		log::trace!(target: "xcm::contains", "NativeAsset asset: {:?}, origin: {:?}", asset, origin);
-		matches!(asset.id, AssetId(ref id) if id == origin)
+		matches!(asset.id, AssetId(ref id) if id.chain_location() == *origin)
 	}
 }
 
--- a/polkadot/xcm/xcm-builder/src/filter_asset_location.rs
+++ b/polkadot/xcm/xcm-builder/src/filter_asset_location.rs
@@ -206,4 +206,32 @@ mod tests {
 			)
 		}
 	}
+	#[test]
+	fn native_asset_filter_works() {
+		frame_support::parameter_types! {
+			pub ParaA: Location = Location::new(1, [Parachain(1001)]);
+			pub ParaB: Location = Location::new(1, [Parachain(1002)]);
+
+			pub AssetXLocation: Location = Location::new(1, [Parachain(1001), GeneralIndex(1111)]);
+			pub AssetYLocation: Location = Location::new(1, [Parachain(1001)]);
+			pub AssetZLocation: Location = Location::new(1, [Parachain(1002)]);
+		}
+
+		let test_data: Vec<(Location, Asset, bool)> = vec![
+			(ParaA::get(), (AssetXLocation::get(), 1).into(), true),
+			(ParaA::get(), (AssetYLocation::get(), 1).into(), true),
+			(ParaA::get(), (AssetZLocation::get(), 1).into(), false),
+			(ParaB::get(), (AssetXLocation::get(), 1).into(), false),
+			(ParaB::get(), (AssetYLocation::get(), 1).into(), false),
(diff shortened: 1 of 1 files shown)


franciscoaguirre (approved):

prdoc/pr_3318.prdoc
```diff
@@ -0,0 +1,24 @@
...
+doc:
+  - audience: Runtime Dev
+    description: |
+      Currently the `NativeAsset` filter from xcm-builder allows you to filter
+      assets whose location is coincident with the origin that sends the message.
+      This PR proposes to modify its behavior to accept any token whose location
+      has the origin as reserve chain.
+
+      ```Example
+      Current behavior:
```
acatangiu, line 12:
```suggestion
      Previous behavior:
```

prdoc/pr_3318.prdoc
```diff
@@ -0,0 +1,24 @@
+title: "Modify native asset to consider chain location"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Currently the `NativeAsset` filter from xcm-builder allows you to filter
+      assets whose location is coincident with the origin that sends the message.
+      This PR proposes to modify its behavior to accept any token whose location
+      has the origin as reserve chain.
```
acatangiu, line 9:
```suggestion
      The `NativeAsset` filter from xcm-builder allows you to filter assets whose
      location coincides with the origin that sends the message.
      This PR changes that behavior to now accept any token whose location has
      the origin as reserve chain.
```

polkadot/xcm/xcm-builder/src/filter_asset_location.rs
```diff
@@ -26,7 +26,7 @@ pub struct NativeAsset;
 impl ContainsPair<Asset, Location> for NativeAsset {
 	fn contains(asset: &Asset, origin: &Location) -> bool {
 		log::trace!(target: "xcm::contains", "NativeAsset asset: {:?}, origin: {:?}", asset, origin);
-		matches!(asset.id, AssetId(ref id) if id == origin)
+		matches!(asset.id, AssetId(ref id) if id.chain_location() == *origin)
 	}
 }
```
acatangiu, line 31: What is a "native" asset? We don't have a clear definition or complete consensus on any definition.

This PR changes the inferred meaning of `NativeAsset` of `origin` chain from:
```diff
- The one single asset whose location matches the `origin` chain location.
+ Any asset originating from sub-locations of the `origin` chain location,
+ i.e. originates from "somewhere" on that chain.
```

Whether this change is "correct" is subjective AFAICT.

But since this `NativeAsset` struct is not used by any Relay chain or any System Parachain, and is mostly used by ecosystem parachains, I can defer that decision to ecosystem engineers. Ergo, I'm approving this.

girazoki: I agree that the correctness of calling it "Native" is subjective, but I am also open to implement this as a separate struct. Although we would have to come up with a new terminology (ReserveAsset maybe?).

In any case I think some of the ecosystem projects (e.g., Moonbeam & Acala) represent their native token as a sub-location of their chain (appending a `PalletIndex` or `GeneralKey`)

albertov19: In terms of terminology, I think it is more common to refer to it as a reserve asset because of the concept of the "Reserve Chain".

I think we all agree that a reserve/native asset is an asset that originates in a reserve chain.

A broader question is, for example, Polkadot and AssetHub-Polkadot are both reserve chains for DOT; I guess this means that DOT is a reserve/native asset of both chains as well. How does System Parachain or the Relay Chain refer to its own asset? I think if it does not have (theoretically speaking) the concept of "NativeAsset," then the name "ReserveAsset" is more natural because there is a reserve chain for that asset.

PR #3319, opened 2024-02-14
Add Coretime to Westend

antonva:
Adds the coretime and on demand pallets to enable Coretime on Westend. 
In order for the migration to run successfully, we need the Broker/Coretime parachain to be live.

TODO:
- [ ] Broker parachain is live https://github.com/paritytech/polkadot-sdk/pull/3272


bkchr (approved):

polkadot/runtime/westend/src/lib.rs
```diff
@@ -1696,6 +1749,7 @@ mod benches {
 		[runtime_parachains::initializer, Initializer]
```
seadanda, line 1749: You'll need `[runtime_parachains::coretime, Coretime]` here and the corresponding file in weights to generate the weights

seadanda (approved):

PR #3321, opened 2024-02-14
rpc-v2/tx: Remove the broadcast event from `transaction_submitAndWatch`

lexnv:
This PR backports the changes from the rpc-v2 spec: https://github.com/paritytech/json-rpc-interface-spec/pull/134

The `Broadcasted` event has been removed:
- it is hard to enforce a `Dropped { broadcasted: bool }` event in cases of a load-balancer being placed in front of an RPC server
  - when the server exists, it is impossible to guarantee this field if the server did not previously send a `Broadcasted` event
- the number of peers reported by this event does not guarantee that peers are unique
  - the same peer can disconnect and reconnect, increasing this metric number
- the number of peers that receive this transaction offers no guarantee about the transaction being included in the chain at a later time 

cc @paritytech/subxt-team


substrate/client/rpc-spec-v2/src/transaction/event.rs
```diff
@@ -70,20 +50,17 @@ pub struct TransactionDropped {
 ///
 /// The status events can be grouped based on their kinds as:
 ///
-/// 1. Runtime validated the transaction:
+/// 1. Runtime validated the transaction and entered the pool:
```
jsdw, line 53:
```suggestion
/// 1. Runtime validated the transaction and it entered the pool:
```

jsdw: I can't speak to the accuracy of this addition, just the grammer :)

jsdw (approved):
Looks good to me!

niklasad1 (approved):

PR #3324, opened 2024-02-14
pallet-membership weights

Dinonard:
## Summary
* use benchamarked weights instead of hardcoded ones for `pallet-membership`
* rename benchmark to match extrinsic name
* remove unnecessary dependency from `clear_prime`


substrate/frame/membership/src/lib.rs
```diff
@@ -500,8 +516,7 @@ mod benchmark {
 		}
 
 		clear_prime {
-			let m in 1 .. T::MaxMembers::get();
```
ggwpez, line 503: I would not change this without reason since it will cause a breaking change in the `WeightInfo` trait and cause runtimes with it to not compile anymore.

substrate/frame/membership/src/migrations/v4.rs
```diff
@@ -47,7 +47,7 @@ pub fn migrate<T: frame_system::Config, P: GetStorageVersion + PalletInfoAccess,
 			target: LOG_TARGET,
 			"New pallet name is equal to the old prefix. No migration needs to be done.",
 		);
-		return Weight::zero()
+		return Weight::zero();
```
ggwpez, line 50: Please revert unrelated changes.

ggwpez (commented):
Good catch 😅

substrate/frame/membership/src/migrations/v4.rs
Dinonard: Oh, my bad, this was done by `cargo fmt` 😅

substrate/frame/membership/src/lib.rs
Dinonard: I understand that but it's redundant - there's no dependency on `m` related to the weight.
Seems wrong to leave it, especially since it's a minute change.

I will also update westend/rococo/example-node not to break anything.

If you strongly suggest to leave it as is, I'll do it like that.
But in that case I should also revert the rename of `reset_member` to keep it compatible.

Dinonard: Thanks for the fast review btw 🙂

bkchr (approved):

substrate/frame/membership/src/lib.rs
```diff
@@ -267,13 +276,19 @@ pub mod pallet {
...
+		pub fn change_key(
+			origin: OriginFor<T>,
+			new: AccountIdLookupOf<T>,
+		) -> DispatchResultWithPostInfo {
 			let remove = ensure_signed(origin)?;
 			let new = T::Lookup::lookup(new)?;
 
+			let mut members_length = T::MaxMembers::get();
+
 			if remove != new {
```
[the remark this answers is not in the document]
ggwpez: Yea. Maybe even return an error... but dont want to side-track this MR.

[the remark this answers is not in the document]
bkchr: Yeah would be nice! Ty!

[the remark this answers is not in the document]
Dinonard: Done!

I opted for the same approach as in swap_member to keep it consistent, and to minimize changes.

ggwpez (approved):

Dinonard:
Can this be merged or is there something else I should do?

PR #3325, opened 2024-02-14
Ensure referenda `TracksInfo` is sorted

ggwpez:
Changes:
- Add `integrity_check` to trait `TracksInfo` to assert the sorting 
- Use the check in `integrity_test`
- Rely upon sorted property in the `info` for speedup

Note that the pallet already relies upon this (so far) untested assumption [here](https://github.com/paritytech/polkadot-sdk/blob/44c7a5eb8c375d77f685abf585961421e5f8b228/substrate/frame/referenda/src/lib.rs#L1280).


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5216437

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5216441

substrate/frame/referenda/src/lib.rs
```diff
@@ -1304,6 +1309,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		}
 	}
 
+	#[cfg(any(feature = "std", test))]
+	fn do_integrity_test() {
+		T::Tracks::check_integrity().expect("Static tracks configuration is valid.");
```
bkchr, line 1314: The expect could use a better message to tell the user what the problem is.

ggwpez: Changed to return `&'static str` in the `check_integrity` for a proper error message.  
Since here we dont actually know what the issue was.

bkchr: Nice! I had the same idea while writing this comment, but just returning the error is a good solution :D

substrate/frame/referenda/src/types.rs
```diff
@@ -144,15 +144,34 @@ pub trait TracksInfo<Balance, Moment> {
 	/// The origin type from which a track is implied.
 	type RuntimeOrigin;
 
-	/// Return the array of known tracks and their information.
+	/// Sorted array of known tracks and their information.
+	///
+	/// The array MUST be sorted by `Id`. Consumers of this trait are advised to assert
+	/// [`check_integrity`] prior to any use.
```
bkchr, line 150:
```suggestion
	/// [`Self::check_integrity`] prior to any use.
```

joepetrowski (approved):

PR #3328, opened 2024-02-14
Add Paseo NetworkId for the new testnet to XCM

franciscoaguirre:
# NOTE

This needs an RFC discussion (TBD) that, if approved, should happen at minimum in v5.

## Added Paseo Network Id to XCMv4

Given the new [Paseo testnet](https://forum.polkadot.network/t/the-new-polkadot-community-testnet/4956) and how it will be used in the ecosystem, it sounded like a good idea for it to have its own `NetworkId`.

### Conversion from v4::NetworkId to v3::NetworkId is now fallible

Because all `NetworkId`s were the same in both v3 and v4, the conversions between them were infallible. Now, `v4::NetworkId` contains an extra variant, so the conversion from it to `v3::NetworkId` is fallible. The only case were there'll be an error is if the new "Paseo" `NetworkId` tries to be converted into v3.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5218166

franciscoaguirre:
bot fmt

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5219604

polkadot/xcm/src/v3/junction.rs
```diff
@@ -126,7 +128,8 @@ impl From<NewNetworkId> for NetworkId {
 			BitcoinCore => Self::BitcoinCore,
 			BitcoinCash => Self::BitcoinCash,
 			PolkadotBulletin => Self::PolkadotBulletin,
-		}
+			Paseo => return Err(()),
```
hbulgarini, line 131: @franciscoaguirre i'm struggling to understand the logic here: Why Paseo would return an error?

xlc:
umm… Gav told me XCM v4 is only rename and therefore does not require RFC. but I guess that’s not the case?

This is a breaking change is someway so https://forum.polkadot.network/t/improve-xcm-development-and-release-process/6497

franciscoaguirre:
We can postpone this to v5

bkchr:
Not needed anymore.

PR #3331, opened 2024-02-15
Feat/use relay blocknumber in pallet broker

### LINKED ISSUE #3201: Broker: Use relay chain blocks for `leadin_length` and `interlude_length`
We should probably use the relay chain block numbers for tracking `leadin_length` and `interlude_length`. Relay chain block numbers are almost deterministic and more future proof.

gitofdeepanshu:
Resolves #3201 

# ChangeLOG
Changing `sale_start`, `interlude_length` and `leading_length` in `pallet_broker` to use relay chain blocknumbers.

# Reasoning
Relay chain block numbers are almost deterministic and more future proof.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5223045

substrate/frame/broker/src/benchmarking.rs
```diff
@@ -771,7 +771,7 @@ mod benches {
 		let core_count = n.try_into().unwrap();
 		let config = new_config_record::<T>();
 
-		let now = frame_system::Pallet::<T>::block_number();
+		let now = <<<T as crate::Config>::Coretime as CoretimeInterface>::RealyChainBlockNumberProvider as BlockNumberProvider>::current_block_number();
```
seadanda, line 774: This typo has been fixed on master. You can merge master and fix `RealyChainBlockNumberProvider` -> `RelayChainBlockNumberProvider`

seadanda:
You'll need to add a [PRDoc](https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/prdoc.md) to describe the changes

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -80,7 +80,7 @@ impl<T: Config> Pallet<T> {
 			last_committed_timeslice: commit_timeslice.saturating_sub(1),
 			last_timeslice: Self::current_timeslice(),
 		};
-		let now = frame_system::Pallet::<T>::block_number();
+		let now = <<<T as crate::Config>::Coretime as CoretimeInterface>::RealyChainBlockNumberProvider as BlockNumberProvider>::current_block_number();
```
ggwpez, line 83: Please create a `BlockNumberProviderOf<T>` type alias for this, or a `now` function.

substrate/frame/broker/src/types.rs
```diff
@@ -206,7 +206,7 @@ pub struct PoolIoRecord {
 /// The status of a Bulk Coretime Sale.
 #[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
 pub struct SaleInfoRecord<Balance, BlockNumber> {
```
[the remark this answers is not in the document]
gitofdeepanshu: you are right, we should be more verbose to avoid confusion, thanks for suggestion.

prdoc/pr_3331.prdoc
```diff
@@ -0,0 +1,13 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Use Relay Blocknumber in Pallet Broker
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Changing `sale_start`, `interlude_length` and `leading_length` in `pallet_broker` to use relay chain block numbers instead of parachain block numbers.
+      Relay chain block numbers are almost deterministic and more future proof.
```
seadanda, line 10: I would add a note to warn people who are already running with a configured broker pallet that with this change they will potentially have different real-world leadin and interlude times. If they previously were timing these against 12s blocks on their parachain they will now be timed by 6s blocks, and therefore the interlude and leadin period will pass twice as quickly. These can be updated using the `configure` extrinsic on the broker pallet.

There is an argument to be made for a migration here to automatically fix the timings, but I don't have a strong opinion on it. Maybe @bkchr or @ggwpez would

seadanda: Actually, with this change we will definitely need a migration, as they are in general totally different types.
All the rest of the changes are looking great

gitofdeepanshu: AFAIK, the `BlocknumberFor<T>` and `RelayBlockNumberFor<T>` are both configured as `u32` ubiquitously in the ecosystem, so I don't think any migration is required.
Am I missing something?

seadanda: FRAME is a toolkit which is designed to be as generic as possible. Making changes in a pallet can make no assumptions about how it is configured except for those implied by any trait bounds and so in general these are different types and should be treated as such for any changes to the pallet.

gitofdeepanshu: Okay, make sense.

So the migration will be from the type `BlockNumberFor<T>` to  `RelayBlockNumberOf<T>` for storage variables 
`Configuration<T>` and `SaleInfo<T>`.

Both these blocknumber types implements trait `Atleast32BitUnsigned`.

My question is how are we supposed to handle cases where `BlockNumberFor<T>` is a larger type than `RelayBlockNumberOf<T>`?
For example converting u128 to u32 where the current `BlockNumberFor<T>` > u32::MAX

We can use `TryFrom` but how to handle the case where it fails?

seadanda: Remember that these configurations are defining the length of sections of a sale, if the types are `Atleast32BitUnsigned`, then values that are too big will be defining sales that are at least `4_294_967_295` blocks long, and we can assume this is a misconfiguration, as that would be a pretty long sale! I think in this case we can log it as an error and insert some default value (maybe u32::MAX, as zero is not meaningful). The config would then be fixed through governance. A similar argument goes for the sale start.

gitofdeepanshu:
I wanted to test the case where `BlockNumberFor<T>` and `RelayBlockNumberOf<T>`, but it turns out internally in the mocks, they are configured to be of the same type. 

What should I do about that?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5523027

substrate/frame/broker/src/utility_impls.rs
```diff
@@ -118,4 +117,9 @@ impl<T: Config> Pallet<T> {
 
 		Ok(Some((region_id, region)))
 	}
+
+	/// Returns the current block number of Relay Chain
+	pub(crate) fn now() -> RelayBlockNumberOf<T> {
```
Szegoo, line 122: Maybe being a bit more explicit here is a good idea. Something like: `relay_height`?

substrate/frame/broker/Cargo.toml
```diff
@@ -25,36 +29,38 @@ sp-runtime = { path = "../../primitives/runtime", default-features = false }
...
 frame-support = { path = "../support", default-features = false }
 frame-system = { path = "../system", default-features = false }
+log = { workspace = true }
 
 [dev-dependencies]
 sp-io = { path = "../../primitives/io" }
 
 [features]
-default = ["std"]
+default = ["std", "try-runtime"]
```
bkchr, line 38:
```suggestion
default = ["std"]
```

substrate/frame/broker/src/test_fungibles.rs
```diff
@@ -97,10 +97,10 @@ where
 		_provenance: Provenance,
 	) -> DepositConsequence {
 		if !Self::asset_exists(asset) {
-			return DepositConsequence::UnknownAsset
+			return DepositConsequence::UnknownAsset;
```
bkchr, line 100: Please revert these unrelated changes.

substrate/frame/broker/src/utility_impls.rs
bkchr: Yeah sounds like a good proposal to me.

substrate/frame/broker/src/migrations/v1.rs
```diff
@@ -0,0 +1,323 @@
...
+					match TryFrom::try_from(config_record.interlude_length) {
+						Ok(val) => val,
+						Err(_) => u32::MAX.into(),
+					};
+
+				let updated_leadin_length: RelayBlockNumberOf<T> =
+					match TryFrom::try_from(config_record.leadin_length) {
+						Ok(val) => val,
+						Err(_) => u32::MAX.into(),
+					};
```
bkchr, line 157: You should create some new `Config`:

```
trait MigrationConfig {
      fn convert_interlude_length(old: BlockNumber) -> RelayBlockNumber;
      fn update_leading_length(old: BlockNumber) -> RelayBlockNumber;
}
```

And then call this here.

substrate/frame/broker/src/migrations/v1.rs
```diff
@@ -0,0 +1,323 @@
...
+			weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
+
+			if let Some(sale_info) = v0::SaleInfo::<T>::take() {
+				log::info!(target: TARGET, "migrating SaleInfo record");
+
+				let updated_sale_start: RelayBlockNumberOf<T> =
+					match TryFrom::try_from(sale_info.sale_start) {
+						Ok(val) => val,
+						Err(_) => u32::MAX.into(),
+					};
```
bkchr, line 180: Same for this. Especially we can not just take the block number and convert this to the relay chain block number. The coretime chain got launched much later than the relay chain itself. I don't want to fuck around with the sales ;)

bkchr (commented):
Generally looking good, but still requires some changes.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5593633

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5593632

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5593613

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5593631

substrate/frame/broker/src/migrations/v1.rs
gitofdeepanshu: Thanks for the comments,
Do you mind explaining it a bit more, I don't think I understand it completely.

Do you want to implement the mentioned trait on Migration struct and use these trait functions inside instead of using tryFrom  directly inside `on_runtime_upgrade()`?

If yes, then we can write a trait like
```rust
trait MigrationConfig<T>
where
	T: Config,
	RelayBlockNumberOf<T>: TryFrom<BlockNumberFor<T>>,
{
	fn converter(old: BlockNumberFor<T>) -> RelayBlockNumberOf<T> {
		match TryFrom::try_from(old) {
			Ok(val) => val,
			Err(_) => u32::MAX.into(),
		}
	}
}
```
and use this everywhere.

substrate/frame/broker/src/migrations/v1.rs
gitofdeepanshu: > Especially we can not just take the block number and convert this to the relay chain block number.

Then how should we do it?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5595269

seadanda:
Hi @gitofdeepanshu it would be nice to get this finished and merged, are you still working on it?

Issue #3242, opened 2024-02-07
Build bridges testing framework


serban300:
The bridges test environments are a bit more complex because we need 2 networks plus a relayer

We should build a minimal framework that can start such an environment and then run the tests. We already have some tests here and some scripting around them: https://github.com/paritytech/polkadot-sdk/tree/master/bridges/zombienet . But it would be good to make everything more modular and more extensible.

ggwpez:
Just bridges? I think @joepetrowski wanted to have a general parachains testing thing. Dont know what you are aiming for here, but maybe it could both be considered.

serban300:
> Just bridges? I think @joepetrowski wanted to have a general parachains testing thing. Dont know what you are aiming for here, but maybe it could both be considered.

Not sure. For the moment just want to start with some simple scripting in order to be able to spawn a more complex zombienet environment and run some bridge tests. After that we can incrementally improve it and maybe reuse it for parachains.

joepetrowski:
If you just want Zombienet you should be able to declare a lot with https://github.com/paritytech/parachains-integration-tests. Talk with @NachoPal about it.

Yes for sure we want something more general/extensible too.

serban300:
> If you just want Zombienet you should be able to declare a lot with https://github.com/paritytech/parachains-integration-tests. Talk with @NachoPal about it.

Thanks ! I didn't know about this tool. Will check it out. This might be very useful !

xlc:
It should be relatively easy to modify Chopsticks to make it support bridges testing. It won't cover everything but should still be useful for high level runtime XCM testing.

bkontur:
Also it should be able to run those tests:
- locally from the scratch/genesis
- In the CI (Polkadot-sdk and Polkadot-fellows repo)

bkchr:
> * how do we define tests? we need to integrate with some DSL. (taking it even further, can we unify Zombienet and Chopsticks DSLs?)

We need the same for: https://github.com/polkadot-fellows/runtimes/issues/171

> * Tests written in yaml files.

Please no. Can we stop this right from the start? Looking at the examples gives me headaches.

pepoviola:
>  Will do that after we have some basic zombienet tests running in the CI

[...]

PR #3333, opened 2024-02-15
Bridge zombienet tests - move all test scripts to the same folder

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/3242


bridges/zombienet/environments/rococo-westend/bridges_rococo_westend.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
```
bkontur, line 1: @serban300 we reference this script in `polkadot-sdk/cumulus/parachains/runtimes/bridge-hubs/README.md` e.g. `./cumulus/scripts/bridges_rococo_westend.sh run-relay`, could you please also fix/check that README.md that it is still working?

bkontur:
I just thinking about the directory structure how actually looks like:
```
zombienet
- environments/rococo-westend
- helpers
- scripts
- tests
- utils
- run-tests.sh
- run-new-test.sh
```

so my thought:
- `scripts` and `utils` can be join together, they both contains bash utility scripts
- `helpers` they are also kind of scripts/utils, but contains just `js-helpers` actually
- maybe all this `helpers` / `scripts` / `utils` could be join and released as a `.tar` so that we could reuse that for `polkadot-fellows` or something like that
- `zombienet` what about renaming to `testing` (maybe we will change zombienet for something else in the future)? Also some stuff from `utils` is not related to the zombienet

serban300:
@bkontur addressed your comments. PTAL !

> * `scripts` and `utils` can be join together, they both contains bash utility scripts

Yes, I agree they both contain bash scripts, but `scripts` will be deleted after we migrate test `0003` to the new approach. So I would leave them as they are right now, and then just directly deleted them.

> * `helpers` they are also kind of scripts/utils, but contains just `js-helpers` actually

Renamed to `js-helpers`

> * maybe all this `helpers` / `scripts` / `utils` could be join and released as a `.tar` so that we could reuse that for `polkadot-fellows` or something like that

Yes, that's also what I'm thinking. Would be good to have some way of packaging these together and reuse them in other repos if needed. Not sure exactly how yet.



> * `zombienet` what about renaming to `testing` (maybe we will change zombienet for something else in the future)?

Yes, makes sense. Done

acatangiu (approved):

bkontur (approved):

Issue #3320, opened 2024-02-14
password-protected keystore with sc_keystore


Hugo-Trentesaux:
I tried to use a password-protected keystore with `LocalKeystore` from `sc_keystore` with this naive approach:

```rust
let suri = rpassword::prompt_password("SURI: ")?;
let keypair = pair_from_str(&suri)?;
let password = rpassword::prompt_password("Password: ")?;
let password = secrecy::Secret::new(password);
let keystore =
	sc_keystore::LocalKeystore::open(project_dir.data_dir(), Some(password))?;
keystore.insert(KeyTypeId(*b"user"), &suri, &keypair.public())?;
```

But when looking in the keystore file, I get the SURI in clear:

```sh
cat .local/share/mytool/7573657296723102a2e339276775717781cbf445953362f7cc65df7e9be5bce5ee21c570
"drill pledge public bunker involve nose vanish buddy day puppy coast patrol"%   
```

If this is the intended behavior, I think it's worth documenting in https://docs.rs/sc-keystore/latest/sc_keystore/struct.LocalKeystore.html#method.open

bkchr:
> But when looking in the keystore file, I get the SURI in clear:

If you load this SURI, you will find out that this is not the SURI of your private key. The actual SURI is `drill pledge public bunker involve nose vanish buddy day puppy coast patrol///YOUR_PASSWORD`.

https://github.com/paritytech/polkadot-sdk/blob/a17b49a44fd6471a3498df7e58be32768925015d/substrate/primitives/core/src/crypto.rs#L949-L950

Hugo-Trentesaux:
I still think that the doc of the `open` method of `LocalKeystore` could be more than one line (https://docs.rs/sc-keystore/latest/src/sc_keystore/local.rs.html#49). In particular, the `password` argument is not mentioned.

davxy:
@Hugo-Trentesaux what is the implementation of your `pair_from_str`?

Because if you generated the keypair **without** the password (as you suspect) you also end with a problem.
Ergo: the password you used to open the keystore doesn't affect that key you generated. 
If you try to sign using the pair you manually generated will not correspond to the one that your keystore will reference when using the same `public()`.

Example
```rust
		let suri = "drill pledge public bunker involve nose vanish buddy day puppy coast patrol";
		let (keypair, _): (sr25519::Pair, _) = Pair::from_phrase(&suri, None).unwrap(); // No password specified for key gen
		let password = String::from("pass");
[...]

PR #3336, opened 2024-02-15
Document LocalKeystore insert method

davxy:
Refer to https://github.com/paritytech/polkadot-sdk/issues/3320 discussion

### DIFF
--- a/substrate/client/keystore/src/local.rs
+++ b/substrate/client/keystore/src/local.rs
@@ -136,6 +136,13 @@ impl LocalKeystore {
 }
 
 impl Keystore for LocalKeystore {
+	/// Insert a new secret key.
+	///
+	/// WARNING: if the secret keypair has been manually generated using a password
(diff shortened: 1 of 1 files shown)


skunert (approved):

substrate/client/keystore/src/local.rs
```diff
@@ -136,6 +136,13 @@ impl LocalKeystore {
 }
 
 impl Keystore for LocalKeystore {
+	/// Insert a new secret key.
```
bkchr, line 139: Should be put above the method declaration in the trait.

davxy: I placed it here as the comments references the keystore construction using a password) which is a concept not referenced anywhere in the `Keystore` trait. I thought that talking about passwords in the trait could be misleading.

bkchr (approved):

PR #3337, opened 2024-02-15
fix(zombienet): docker `img` version to use in merge queues for bridges

pepoviola:
Fix image tag to use in merge queues tests, currently is failing because is using the wrong tag (e.g https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5224342).

Thx!

### DIFF
--- a/.gitlab/pipeline/zombienet/bridges.yml
+++ b/.gitlab/pipeline/zombienet/bridges.yml
@@ -7,12 +7,12 @@
     - .kubernetes-env
     - .zombienet-refs
   rules:
-    - !reference [.build-refs, rules]
-    - if: $CI_COMMIT_REF_NAME =~ /^gh-readonly-queue.*$/
-  before_script:
     # Docker images have different tag in merge queues
-    - if [[ $CI_COMMIT_REF_NAME == *"gh-readonly-queue"* ]]; then export DOCKER_IMAGES_VERSION="${CI_COMMIT_SHORT_SHA}"; fi
-    - export BRIDGES_ZOMBIENET_TESTS_IMAGE_TAG="${DOCKER_IMAGES_VERSION}"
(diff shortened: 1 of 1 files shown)


bkontur (approved):

.gitlab/pipeline/zombienet/bridges.yml
```diff
@@ -9,9 +9,9 @@
   rules:
     - !reference [.build-refs, rules]
     - if: $CI_COMMIT_REF_NAME =~ /^gh-readonly-queue.*$/
```
serban300, line 11: I think we should delete this if the variable assignment has been moved.

serban300 (commented):
Thanks for fixing this ! Sorry, I was thinking it wouldn't work in `before_script` since we use `DOCKER_IMAGES_VERSION` as part of the image name.

.gitlab/pipeline/zombienet/bridges.yml
pepoviola: I think we should keep, since this rule make these test _only_ run on merge queues.

pepoviola:
> Thanks for fixing this ! Sorry, I was thinking it wouldn't work in `before_script` since we use `DOCKER_IMAGES_VERSION` as part of the image name.

Ahh, that true. We should export the final var here. I will fix it.

alvicsam (approved):

.gitlab/pipeline/zombienet/bridges.yml
serban300: It should also run in the CI. And it seems to be running. If this rule makes it only run in merge queues it's by mistake.

serban300:
@pepoviola I see that there are still some errors. Are you working on them ? If not I can try to fix them since it's a bridge testing error. Please let me know how you prefer.

pepoviola:
Hi @serban300, thanks for the heads up. Let me check and fix it. 
Thx!

PR #3341, opened 2024-02-15
Fix parachain upgrade scheduling when done by the owner/root

bkchr:
When using `schedule_code_upgrade` to change the code of a parachain in the relay chain runtime, we had already fixed to not set the `GoAhead` signal. This was done to not brick any parachain after the upgrade, because they were seeing the signal without having any upgrade prepared. The remaining problem is that the parachain code is only upgraded after a parachain header was enacted, aka the parachain made some progress. However, this is quite complicated if the parachain is bricked (which is the most common scenario why to manually schedule a code upgrade). Thus, this pull request replaces `SetGoAhead` with `UpgradeStrategy` to signal to the logic kind of strategy want to use. The strategies are either `SetGoAheadSignal` or `ApplyAtExpectedBlock`. `SetGoAheadSignal` sets the go ahead signal as before and awaits a parachain block. `ApplyAtExpectedBlock` schedules the upgrade and applies it directly at the `expected_block` without waiting for the parachain to make any kind of progress.


polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -386,16 +386,19 @@ pub(crate) enum PvfCheckCause<BlockNumber> {
...
 }
 
-/// Should the `GoAhead` signal be set after a successful check of the new wasm binary?
+/// Should the upgrade of the validation code be enacted directly after the checks succeeded?
 #[derive(Debug, Copy, Clone, PartialEq, TypeInfo, Decode, Encode)]
-pub enum SetGoAhead {
-	Yes,
+pub enum EnactUpgradeDirectly {
 	No,
+	Yes,
```
bkchr, line 401: Change the order here to decode with the correct meaning.

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1621,43 +1612,45 @@ impl<T: Config> Pallet<T> {
...
+			//
+			// The first parablock that has a relay-parent higher or at the same height of
+			// `expected_at` will trigger the code upgrade. The parablock that comes after that will
+			// be validated against the new validation code.
+			//
+			// Here we are trying to choose the block number that will have
+			// `validation_upgrade_delay` blocks from the relay-parent of inclusion of the the block
+			// that scheduled code upgrade but no less than `minimum_validation_upgrade_delay`. We
+			// want this delay out of caution so that when the last vote for pre-checking comes the
+			// parachain will have some time until the upgrade finally takes place.
```
eskimor, line 1634: note that we pick the maximum here - the validation_upgrade_delay needs to be maintained. Reasoning see [here](https://github.com/paritytech/polkadot-sdk/blob/69434d9a32af6103d7cbde9798cf594027fb620e/polkadot/runtime/parachains/src/configuration.rs#L126). It seems this gets violated by these changes.

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1621,43 +1612,45 @@ impl<T: Config> Pallet<T> {
...
-		// but no less than `minimum_validation_upgrade_delay`. We want this delay out of caution
-		// so that when the last vote for pre-checking comes the parachain will have some time until
-		// the upgrade finally takes place.
-		let expected_at = cmp::max(
-			relay_parent_number + cfg.validation_upgrade_delay,
-			now + cfg.minimum_validation_upgrade_delay,
-		);
+		if enact_upgrade_directly == EnactUpgradeDirectly::Yes {
+			FutureCodeHash::<T>::remove(&id);
+			weight.saturating_add(Self::set_current_code(id, *code_hash, now));
```
eskimor, line 1621: This makes it available for use immediately, which is unsound in the case of chain reversions. See comment below.

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1612,32 +1669,37 @@ impl<T: Config> Pallet<T> {
...
+		// `validation_upgrade_delay` blocks from the relay-parent of inclusion of the the block
+		// that scheduled code upgrade but no less than `minimum_validation_upgrade_delay`. We
+		// want this delay out of caution so that when the last vote for pre-checking comes the
+		// parachain will have some time until the upgrade finally takes place.
+		let expected_at = cmp::max(
+			relay_parent_number + cfg.validation_upgrade_delay,
+			now + cfg.minimum_validation_upgrade_delay,
+		);
+
+		if upgrade_strategy == UpgradeStrategy::ApplyAtExpectedBlock {
```
eskimor, line 1693: Nit: Why is this not a `match`? Would be a tad more future proof in case someone introduces more variants.

eskimor (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5684618

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5684638

polkadot/runtime/parachains/src/paras/mod.rs
```diff
@@ -1621,38 +1669,50 @@ impl<T: Config> Pallet<T> {
...
+		// parachain will have some time until the upgrade finally takes place.
 		let expected_at = cmp::max(
 			relay_parent_number + cfg.validation_upgrade_delay,
 			now + cfg.minimum_validation_upgrade_delay,
 		);
 
-		weight += T::DbWeight::get().reads_writes(1, 4);
-		FutureCodeUpgrades::<T>::insert(&id, expected_at);
+		match upgrade_strategy {
+			UpgradeStrategy::SetGoAheadSignal => {
```
sandreim, line 1694: This match arm here seems to be doing what needs to be done by for `UpgradeStrategy::ApplyAtExpectedBlock`.

bkchr: LOL. I'm an idiot. Broken it with one of the last commits.

PR #3351, opened 2024-02-16
Update relayer image and remove extra sleep from bridges+zombienet test

svyatonik:
~Let's get 5 green checks before merging that~ Ok, I think 3 is enough - let's proceed with merge


svyatonik:
@paritytech/bridges-core PTAL

serban300 (approved):

serban300:
Updated the relayer as part of https://github.com/paritytech/polkadot-sdk/pull/4069 and also removed the sleep. It's waiting in the merge queue. Closing this PR.

PR #3361, opened 2024-02-16
contracts: Fix double charge of gas for host functions

athei:
This PR is fixing  a bug in the sync mechanism between wasmi and pallet-contracts. This bug leads to essentially double charging all the gas that was used during the execution of the host function. When the `call` host function is used for recursion this will lead to a quadratic amount of gas consumption with regard to the nesting depth.

We also took the chance to refactor the code in question and improve the rust docs.

### DIFF
--- a/substrate/frame/contracts/fixtures/contracts/recurse.rs
+++ b/substrate/frame/contracts/fixtures/contracts/recurse.rs
@@ -1,0 +1,51 @@
+// This file is part of Substrate.
+
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This fixture calls itself as many times as passed as argument.
+
+#![no_std]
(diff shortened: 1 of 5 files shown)


substrate/frame/contracts/fixtures/contracts/recurse.rs
```diff
@@ -0,0 +1,51 @@
...
+	input!(calls_left: u32, );
+
+	// own address
+	output!(addr, [0u8; 32], api::address,);
+
+	if calls_left == 0 {
+		return
+	}
+
+	api::call_v1(
```
deuszx, line 42: is `call_v2` covered implicitly? Or it doesn't matter?

athei: It doesn't matter. The newer version just allows for explicitly passing a deposit limit and a proof size limit. The older version sets some defaults for it. But they run the same code.

xermicus (approved):

substrate/frame/contracts/src/gas.rs
```diff
@@ -79,8 +97,13 @@ pub struct GasMeter<T: Config> {
 	gas_left: Weight,
 	/// Due to `adjust_gas` and `nested` the `gas_left` can temporarily dip below its final value.
 	gas_left_lowest: Weight,
-	/// Amount of fuel consumed by the engine from the last host function call.
-	engine_consumed: u64,
+	/// The amount of resources that was consumed by the execution engine.
+	///
+	/// This should be equivalent to `self.gas_consumed().ref_time()` but expressed in whatever
+	/// unit the execution engine uses to track resource consumption. We have to track it seperatly
```
pgherveou, line 103:
```suggestion
	/// unit the execution engine uses to track resource consumption. We have to track it separately
```

pgherveou (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5291247

Issue #3352, opened 2024-02-16
[ci] build-runtimes-polkavm job doesn't work


alvicsam:
The job is [broken](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5230166) for a few weeks. Should we disable it until it's fixed?

cc @koute @paritytech/ci

koute:
Unless I screwed/misunderstood something up this job was working when I merged it in https://github.com/paritytech/polkadot-sdk/pull/3209.

No; let's not disable it. I'll fix it.

While we're at it, do you know why this job doesn't show up on GitHub in the checks list above the "Add a comment"? Wouldn't this make possible to merge PRs which break the job? How can I make it show up?

alvicsam:
Hm, it's interesting. It seems that it runs only on master branch. I'll also take a look.

alvicsam:
@koute [this PR](https://github.com/paritytech/polkadot-sdk/pull/3354) is now merged and I can see the job in PR pipeline. As soon as you fix it we can mark it as `Required` in GitHub.

koute:
@alvicsam Okay, thanks for the help!

PR with a fix is up: https://github.com/paritytech/polkadot-sdk/pull/3363

PR #3363, opened 2024-02-16
Make `sp-keystore` `no_std`-compatible and fix the `build-runtimes-polkavm` CI job

koute:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3352

### DIFF
--- a/substrate/primitives/keystore/Cargo.toml
+++ b/substrate/primitives/keystore/Cargo.toml
@@ -17,8 +17,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "3.6.1", default-features = false, features = ["derive"] }
-parking_lot = { version = "0.12.1", default-features = false }
-thiserror = "1.0"
+parking_lot = { version = "0.12.1", default-features = false, optional = true }
 sp-core = { path = "../core", default-features = false }
 sp-externalities = { path = "../externalities", default-features = false }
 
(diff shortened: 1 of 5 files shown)


substrate/primitives/keystore/Cargo.toml
```diff
@@ -28,7 +27,7 @@ rand_chacha = "0.2.2"
 
 [features]
 default = ["std"]
-std = ["codec/std", "sp-core/std", "sp-externalities/std"]
+std = ["codec/std", "dep:parking_lot", "sp-core/std", "sp-externalities/std"]
```
bkchr, line 30:
```suggestion
std = ["codec/std", "parking_lot", "sp-core/std", "sp-externalities/std"]
```

bkchr (approved):

dmitry-markin (approved):

substrate/primitives/keystore/Cargo.toml
koute: @bkchr Hm... sure, I can do this, but just to confirm - do we want to never use the `dep:` syntax unless necessary?

I'm asking because this is not just a stylistic choice. The nice thing about the new (well, new-ish) `dep:` syntax it that it prevents an implicit feature from being defined, that is, if an `optional` dependency is *not* referred to with a `dep:` then cargo will implicitly define a new feature like this:

```
[features]
parking_lot = ["dep:parking_lot"]
```

And in consequence this now *technically* becomes part of the public API for the crate, so another crate which depends on the `sp-keystore` can then say:

```
[dependencies]
sp-keystore = { version = "...", features = ["parking_lot"] }
```

...which doesn't make much sense because the `parking_lot` dependency here is an *internal* dependency, so it makes *zero* sense for another crate to be able to enable it. And using it through `dep:` prevents this, because then `cargo` won't create this implicit feature.

bkchr: Okay good points, then let's ignore what I said. I didn't realized that this would then not be exposed :+1:

PR #3367, opened 2024-02-17
add ability to concurrently fetch child keys for remote externality tests

brunopgalvao:
Closes https://github.com/paritytech/polkadot-sdk/issues/2245

> Child-tree data is currently fetched sequentially.
> 
> Most chains to date don't have that many child-trees so this hasn't been an issue, but the Rococo Contracts chain has almost 3000 child tries making this serial approach unnecessarily add a really long time to scraping keys.

This PR makes it so requests for child keys are concurrent rather than sequential.

### DIFF
--- a/substrate/utils/frame/remote-externalities/src/lib.rs
+++ b/substrate/utils/frame/remote-externalities/src/lib.rs
@@ -51,6 +51,7 @@ use std::{
 };
 use substrate_rpc_client::{rpc_params, BatchRequestBuilder, ChainApi, ClientT, StateApi};
 use tokio_retry::{strategy::FixedInterval, Retry};
+use tokio::task;
 
 type KeyValue = (StorageKey, StorageData);
 type TopKeyValues = Vec<KeyValue>;
@@ -875,9 +876,8 @@ where
 		pending_ext: &mut TestExternalities<HashingFor<B>>,
 	) -> Result<ChildKeyValues, &'static str> {
 		let child_roots = top_kv
-			.iter()
-			.filter(|(k, _)| is_default_child_storage_key(k.as_ref()))
-			.map(|(k, _)| k.clone())
+			.into_iter()
+			.filter_map(|(k, _)| is_default_child_storage_key(k.as_ref()).then(|| k.clone()))
 			.collect::<Vec<_>>();
 
 		if child_roots.is_empty() {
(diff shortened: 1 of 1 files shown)


liamaharon (commented):
Exciting @brunopgalvao! 

I'd be curious to see a before/after benchmark of scraping the rococo contracts chain with this code.

Will do a proper review next week.

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -893,34 +893,70 @@ where
...
+		// Create an Arc to share ownership of the client.
+		let client = Arc::new(self.as_online().rpc_client().clone());
 		let mut child_kv = vec![];
-		for prefixed_top_key in child_roots {
-			let child_keys =
-				Self::rpc_child_get_keys(client, &prefixed_top_key, StorageKey(vec![]), at).await?;
 
-			let child_kv_inner =
-				Self::rpc_child_get_storage_paged(client, &prefixed_top_key, child_keys, at)
+		// Spawn tasks for each child root and collect their handles.
```
liamaharon, line 900: since this is network bound (not cpu bound), we can use use async/await on a single thread rather than spawning multiple threads. simpler, same performance, and has less issues with jsonrpsee. example: https://github.com/paritytech/polkadot-sdk/blob/afbb749b4cdc7a924da37b5f604f847a4791e8d8/substrate/utils/frame/remote-externalities/src/lib.rs#L729-L734

brunopgalvao: Okay. Updated the code accordingly.

`wss://rococo-contracts-rpc.polkadot.io:443` will not work because the node is not configured with the `--rpc-max-request-size 10000000 --rpc-max-response-size` flags.

Ideally we can test this with chopsticks. I ran:
```
npx @acala-network/chopsticks@latest --endpoint=wss://rococo-contracts-rpc.polkadot.io:443
```
Which gave me a running cloned version of the Rococo Contracts node.

However, when I ran:
```
TEST_WS=ws://127.0.0.1:8000 cargo test --features=remote-test -p frame-remote-externalities -- --nocapture
```
I got the following error:
```
⠋     0.002 s   Loading snapshot...Feb 20 14:58:50.169 ERROR remote-ext: Error = Transport(Request rejected `426`) 
```
Thoughts?

liamaharon: > wss://rococo-contracts-rpc.polkadot.io:443 will not work because the node is not configured with the --rpc-max-request-size 10000000 --rpc-max-response-size flags.

It should not need those flags. I just scraped from that rpc using latest version of try-runtime-cli locally, what error do you get?

Chopsticks will be unusable for scraping the entire state, afaik it makes a single request for each state entry..

brunopgalvao: Hey @liamaharon, kinda stumbled here, can you describe to me the steps required to run try-runtime-cli against `rococo-contracts-rpc.polkadot.io:443`? I would use `follow-chain` with the `substrate-contracts-node`?

liamaharon: Try `try-runtime create-snapshot --uri wss://rococo-contracts-rpc.polkadot.io:443 ../snaps/rococo-contracts.snap`

brunopgalvao: Looks like the error is coming from JSONrpsee:
```rust
/// Max number of request slots exceeded.
#[error("Max concurrent requests exceeded")]
MaxSlotsExceeded,
```
https://github.com/paritytech/jsonrpsee/blob/6da4608bee6e1023257b780bfe8d1778c0fad4e3/core/src/client/error.rs#L57-L59

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -893,34 +892,67 @@ where
 
 		let at = self.as_online().at_expected();
 
-		let client = self.as_online().rpc_client();
+		// Create an Arc to share ownership of the client.
+		let client = Arc::new(self.as_online().rpc_client().clone());
```
liamaharon, line 896: I don't think need Arc since this no longer splits across threads

substrate/utils/frame/remote-externalities/src/lib.rs
liamaharon: Try limit the concurrency to the `PARALLEL_REQUESTS` https://github.com/paritytech/polkadot-sdk/blob/9bf1a5e23884921498b381728bfddaae93f83744/substrate/utils/frame/remote-externalities/src/lib.rs#L354

brunopgalvao: Removed `Arc` and limited concurrency to `PARALLEL_REQUESTS`.

Issue #3326, opened 2024-02-14
[Deprecation] remove `pallet::getter` usage from all pallets


muraca:
This issue tracks the removal of `pallet::getter` in all pallets, as per #223. 
This approach avoids a huge PR and lots of merge conflicts, a mistake I already made when I tried to perform this task last year.

Bridges:
- [ ] `pallet-bridge-grandpa`
- [ ] `pallet-bridge-messages`
- [ ] `pallet-bridge-relayers`
- [ ] `pallet-xcm-bridge-hub-router`
- [ ] `snowbridge-pallet-ethereum-client`
- [ ] `snowbridge-pallet-inbound-queue`
- [ ] `snowbridge-pallet-outbound-queue`
- [ ] `snowbridge-pallet-system`

Cumulus:
- [ ] `cumulus-pallet-aura-ext`
- [ ] `pallet-collator-selection`
- [ ] `cumulus-pallet-parachain-system`
- [ ] `pallet-parachain-template`
- [ ] `staging-parachain-info`

Polkadot:
- [ ] `polkadot-runtime-common`
- [ ] `polkadot-runtime-parachains`
- [ ] `pallet-xcm`
- [ ] `xcm-simulator-example`
- [ ] `xcm-simulator-fuzzer`

FRAME:
- [ ] `pallet-template`
- [ ] `pallet-alliance`
- [ ] `pallet-aura`
- [ ] `pallet-authority-discovery`
- [ ] `pallet-babe`
- [ ] `pallet-balances`
- [ ] `pallet-beefy`
- [ ] `pallet-beefy-mmr`
- [ ] `pallet-bounties`
- [ ] `pallet-child-bounties`
- [ ] `pallet-collective`
- [ ] `pallet-contracts-mock-network`
- [ ] `pallet-democracy`
- [ ] `pallet-election-provider-multi-phase`
- [ ] `pallet-elections-phragmen`
- [ ] `pallet-example-basic`
- [ ] `pallet-example-kitchensink`
- [ ] `pallet-example-offchain-worker`
- [ ] `pallet-fast-unstake`
- [ ] `pallet-grandpa`
- [ ] `pallet-identity`
- [ ] `pallet-im-online`
- [ ] `pallet-insecure-randomness-collective-flip`
- [ ] `pallet-membership`
- [ ] `pallet-mmr`
- [ ] `pallet-nft-fractionalization`
- [ ] `pallet-node-authorization`
- [ ] `pallet-offences`
[...]

PR #3370, opened 2024-02-17
removed `pallet::getter` from `pallet-sudo`

muraca:
part of #3326 

@ggwpez @kianenigma @shawntabrizi


ggwpez (approved):

substrate/frame/sudo/src/extension.rs
```diff
@@ -86,7 +86,7 @@ where
 		info: &DispatchInfoOf<Self::Call>,
 		_len: usize,
 	) -> TransactionValidity {
-		let sudo_key: T::AccountId = <Pallet<T>>::key().ok_or(UnknownTransaction::CannotLookup)?;
+		let sudo_key: T::AccountId = <Key<T>>::get().ok_or(UnknownTransaction::CannotLookup)?;
```
gupnik, line 89: I will do `Key::<T>::get()` here as well.

gupnik (approved):

substrate/frame/sudo/src/extension.rs
kianenigma: Yes please 🙏🏻

kianenigma, line 89:
```suggestion
		let sudo_key: T::AccountId = Key::<T>::get().ok_or(UnknownTransaction::CannotLookup)?;
```

muraca:
@ggwpez the silent label was removed as I had to write a prdoc because these are breaking changes

ggwpez:
How is this a breaking change?

muraca:
All getters visibility is `pub`, while the storage items left behind might not have this visibility, for example here we have `pub(super)`. 
This is probably due to the fact that the key is managed by pallet logic and we don't want other runtime logic to modify it. 

`pallet_sudo::Key::<T>::get()` is not accessible from other pallets.

kianenigma:
Yeah, so I also first added `Silent` to these and added to the queue, but then removed them. 

If we are to be pedantic, they are all breaking changes.

muraca:
> `pallet_sudo::Key::<T>::get()` is not accessible from other pallets.

what should we do with this issue?

kianenigma:
One solution would be to just remove the usage of all getters in pallets one by one, but keep the getters. 

This will not be breaking change in any way, and do 99% of the work.

Then, in one breaking PR, we mark the getters as deprecated and/or remove them.

muraca:
> One solution would be to just remove the usage of all getters in pallets one by one, but keep the getters.

Keeping the getter would not solve the issue, if you want to not use it.
Say that a pallet is using the sudo key, for whatever reason. If not with a getter, how will this other pallet access such data?   

The solutions I see are either `pub` for all the storage items where we remove a getter, or we manually implement getters for items that must not be changed by other logic, for example the sudo key.

PR #3376, opened 2024-02-18
fix(zombienet): increase timeout in download artifacts

pepoviola:
Fix timeouts downloading the artifacts (e.g https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5246040)
Thx!


polkadot/zombienet_tests/misc/0002-upgrade-node.zndsl
```diff
@@ -10,9 +10,8 @@ dave: parachain 2001 block height is at least 10 within 200 seconds
 # POLKADOT_PR_ARTIFACTS_URL=https://gitlab.parity.io/parity/mirrors/polkadot/-/jobs/1842869/artifacts/raw/artifacts
 # with the version of polkadot you want to download.
 
-# avg 30s in our infra
-alice: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 60 seconds
-bob: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 60 seconds
+alice: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 90 seconds
+bob: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 90 seconds
```
bkchr, line 14:
```suggestion
alice: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 240 seconds
bob: run ./0002-download-polkadot-from-pr.sh with "{{POLKADOT_PR_ARTIFACTS_URL}}" within 240 seconds
```

Not sure how often I need to repeat myself, as timeouts are quite useless in CI because you don't know the load etc. Here the timeout is even more qestionable, because the timeout ensures that we start something in a given amount of time.

bkchr (approved):

polkadot/zombienet_tests/misc/0002-upgrade-node.zndsl
pepoviola: Agree, we should allow to run without specify the timeout.
Thx!

sandreim (approved):

Issue #3208, opened 2024-02-05
Create test case to confirm pallet indexes are correct


claravanstaden:
Stemming from this conversation: https://github.com/polkadot-fellows/runtimes/pull/137#discussion_r1477920049

Write a test that tests that the expected behaviour (that the MessageQueue pallet index is larger than the EthereumOutboundQueue pallet index).

kianenigma:
I am not sure what this issue is asking at the moment. The underlying issue seems important and something that should be solved as a part of CRV2 with a new and better standard. As for now, if you want this issue to remain do-able and mentor-able, please specify the exact test case that you want to be checked.

PR #3381, opened 2024-02-19
Snowbridge - Test pallet order

claravanstaden:
- Adds a test to check the correct digest for Snowbridge outbound messages. For the correct digest to be in the block, the the MessageQueue pallet should be configured after the EthereumOutbound queue pallet. The added test fails if the EthereumOutbound is configured after the MessageQueue pallet.
- Adds a helper method `run_to_block_with_finalize` to simulate the block finalizing. The existing `run_to_block` method does not finalize and so it cannot successfully test this condition.

Resolves: https://github.com/paritytech/polkadot-sdk/issues/3208


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5253265

claravanstaden:
@bkontur this is the test you requested for https://github.com/paritytech/polkadot-sdk/issues/3208.

bridges/snowbridge/runtime/test-common/src/lib.rs
```diff
@@ -193,12 +195,96 @@ pub fn send_transfer_token_message_success<Runtime, XcmConfig>(
...
+				assethub_parachain_id,
+				weth_contract_address,
+				destination_address,
+				fee_amount,
+			);
+
+			assert_ok!(outcome.ensure_complete());
+
+			// check events
+			let mut events = <frame_system::Pallet<Runtime>>::events()
```
bkontur, line 257: @claravanstaden if you add MessageQueue before EthereumOutboundQueue, what is the exact issue/problem/error? Does MessageQueue or anything else trigger any other Event or something? I mean, is it possible to add a check for that negative scenario? E.g. `System::events()` does not contain `MessageQueue::Processed` or whatever?

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/snowbridge.rs
```diff
@@ -135,6 +135,32 @@ fn ethereum_to_polkadot_message_extrinsics_work() {
 	);
 }
 
+/// Tests that the digest items are as expected when a Ethereum Outbound message is received.
+/// If the MessageQueue pallet is configured before (i.e. the MessageQueue pallet is listed before
+/// the EthereumOutboundQueue in the construct_runtime macro) the EthereumOutboundQueue, this test
+/// will fail.
+#[test]
+pub fn test_pallet_order_works() {
```
bkontur, line 143: @claravanstaden I would suggest to rename this `test_pallet_order_works` to something very specific - exactly what it does, e.g. `fn ethereum_outbound_queue_processes_messages_before_message_queue_works` or something more descriptive

bridges/snowbridge/runtime/test-common/src/lib.rs
claravanstaden: The MessageQueue::Processed event will still be there, but `Other` digest item will not be found in the DigestLogs. https://github.com/paritytech/polkadot-sdk/pull/3381/files/58729e09808a758c079a0dd5334718f1db45adbd#diff-6abc523c7c31eb7defa80d6c139fd2b2a7de5b299da9f9bfeebdb2748881886cR279

I thought about adding a negative test, but I thought it didn't really make sense because that wouldn't use the bridge hub runtime and wouldn't catch the actual runtime misconfiguration. I did however test that this test fails if the order is incorrect.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/tests/snowbridge.rs
claravanstaden: Agreed, done in https://github.com/paritytech/polkadot-sdk/pull/3381/commits/e7fc3a42b426cb87c0dd8a0f053fb1469de5ed3c.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -102,7 +102,7 @@ use polkadot_runtime_common::prod_or_fast;
 
 #[cfg(feature = "runtime-benchmarks")]
 use benchmark_helpers::DoNothingRouter;
-#[cfg(not(feature = "runtime-benchmarks"))]
+//#[cfg(not(feature = "runtime-benchmarks"))]
```
acatangiu, line 105: remove this

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -730,7 +726,7 @@ construct_runtime!(
 
 		// Message Queue. Importantly, is registered last so that messages are processed after
 		// the `on_initialize` hooks of bridging pallets.
-		MessageQueue: pallet_message_queue = 250,
+		MessageQueue: pallet_message_queue = 175,
```
acatangiu, line 729: @paritytech/frame-coders any concern about swapping this index on live Rococo?

acatangiu (approved):

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -367,10 +367,6 @@ parameter_types! {
 impl pallet_message_queue::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type WeightInfo = weights::pallet_message_queue::WeightInfo<Runtime>;
-	#[cfg(feature = "runtime-benchmarks")]
-	type MessageProcessor =
-		pallet_message_queue::mock_helpers::NoopMessageProcessor<AggregateMessageOrigin>;
-	#[cfg(not(feature = "runtime-benchmarks"))]
```
claravanstaden, line 373: @bkontur the problem with the benchmark tests lies here. If the `NoopMessageProcessor` is used, the messages aren't actually processed and so the nonces doesn't increment and all the post-message sending checks fail. It looks like the tests pass with the without the special `NoopMessageProcessor`, will double check CI.

bridges/snowbridge/runtime/test-common/src/lib.rs
bkontur: ok, got it, so there is nothing (Event, DigestItem,..) to check for case when MessageQueue processes first?

claravanstaden: Exactly.

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
gupnik: Seems that it is still the last pallet. So, shouldn't be a concern @acatangiu.

bkontur (approved):
you can revert that benchmark stuff: https://github.com/paritytech/polkadot-sdk/pull/3381#discussion_r1497123273
and fix it as a separate

claravanstaden:
Thanks @bkontur, I'll make a separate PR. Reverted in https://github.com/paritytech/polkadot-sdk/pull/3381/commits/c950af0bb0687f88ad9eb7f99149e48f9b301735

PR #1462, opened 2023-09-08
PoV Reclaim (Clawback) Node Side

skunert:
This PR provides the infrastructure for the pov-reclaim mechanism discussed in #209. The goal is to provide the current proof size to the runtime so it can be used to reclaim storage weight.

## New Host Function
- A new host function is provided [here](https://github.com/skunert/polkadot-sdk/blob/5b317fda3be205f4136f10d4490387ccd4f9765d/cumulus/primitives/pov-reclaim/src/lib.rs#L23). It returns the size of the current proof size to the runtime. If recording is not enabled, it returns 0.

## Implementation Overview
- Implement option to enable proof recording during import in the client. This is currently enabled for `polkadot-parachain`, `parachain-template` and the cumulus test node.
- Make the proof recorder ready for no-std. It was previously only enabled for std environments, but we need to record the proof size in `validate_block` too.
- Provide a recorder implementation that only the records the size of incoming nodes and does not store the nodes itself.
- Fix benchmarks that were broken by async backing changes
- Provide new externalities extension that is registered by default if proof recording is enabled.
- I think we should discuss the naming, pov-reclaim was more intuitive to me, but we could also go with clawback like in the issue.

## Impact of proof recording during import
With proof recording: 6.3058 Kelem/s
Without proof recording: 6.3427 Kelem/s

The measured impact on the importing performance is quite low on my machine using the block import benchmark. With proof recording I am seeing a performance hit of 0.585%.

PR #3386, opened 2024-02-19
Multi tree support

cheme:
This PR enable multi tree storage of paritydb.
It currently depends on https://github.com/paritytech/trie/pull/210

cc\ @MattHalpinParity @arkpar 

A new option is offered to client: paritydbmulti that should probably be consider experimental (it is not default). For polkadot parachain db it is not used (only impact state storage).

Code is based on https://github.com/paritytech/polkadot-sdk/tree/arkpar/new-trie . I am currently not really happy with the way I did merge the changes from https://github.com/paritytech/polkadot-sdk/pull/1462 into it (I kept the WithDatabase version, but I am not sure there is not a better way to run it (basically new code with 'wrap' did not run well with Box<Dyn NodeDb> change, so after a few attempt I just made the new RwLock of this branch not present under no_std).

Right now I did not move a lot of test to support multitree, neither did I test a lot, so I am opening the branch in draft.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5265104

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5265708

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5266004

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5266589

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5266881

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5267620

substrate/primitives/state-machine/src/backend.rs
```diff
@@ -248,9 +245,9 @@ pub trait Backend<H: Hasher>: sp_std::fmt::Debug {
 	/// Does not include child storage updates.
 	fn storage_root<'a>(
 		&self,
-		delta: impl Iterator<Item = (&'a [u8], Option<&'a [u8]>)>,
+		delta: impl Iterator<Item = (&'a [u8], Option<&'a [u8]>, Option<ChildChangeset<H::Out>>)>,
 		state_version: StateVersion,
-	) -> (H::Out, BackendTransaction<H>)
+	) -> TrieCommit<H::Out>
```
rphmeier, line 250: This is now tightly coupling every state machine used in Polkadot-SDK to the parity trie format.

I hope there is a way to avoid it.
I had planned to experiment with some next-gen trie-DBs that would be incompatible with this change.

cheme: Yes, different backend requires an abstraction on the payload return by the state machine.
That is something I tried to achieve in the past (abstract state backend, abstract payload and abstract db storage), and never put it to completion (big refactoring, quite impacting at the time, maybe more doable today).
The issue with the existing code is that it abstracted only on H (the hasher), so you will need to implement it anyway.
This change  (TrieCommit) just change the existing static backend to the new one needed for new trie design: switching from PrefixedMemoryDb<H> + H::Out to TrieCommit<H::Out>. (old BackendTransaction looks like an abstraction but is just a type alias).
If the next-gen trie-db it is what is describe in your recent blog post, PrefixedMemoryDb<H> + H::Out is likely not what you need, neither is TrieCommit .
For experiment I would probably just use enum variants (db , statemachine triebackend and db payload (Backendtransaction). Making substrate generic over state backend would really be wonderful, but quite a refactoring.

burdges: I could not find `trie_db::Changeset` anywhere, but anyways..

We'd want abstraction over data structures, not abstraction over tries.  We know optimal data structures should store data only at leaves, never at internal nodes, making tries a dead end.  We could envision several data structures worth exploring, but they're all binary trees with no data at internal nodes, aka no more tries.

cheme: yeah that's not merge https://github.com/paritytech/trie/pull/210

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5273481

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5277477

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5281505

substrate/primitives/state-machine/src/backend.rs
rphmeier: > Making substrate generic over state backend would really be wonderful, but quite a refactoring.

I am kind of surprised to hear this. At least, from the runtime perspective the interfaces look very general, which means that we just have to find the right boundary.
  
Maybe the `sp_state_machine` is the wrong place for that generalization, but it should be somewhere

arkpar: I guess we can use a new, substrate-defined type here.

The goal of this PR is to bring some optimizations to the existing trie format. Basically, for the DB-backed tries we want to be able to communicate the tree structure to the database level, as opposed to just pushing a set of nodes. 

For the next trie generation, I envision the connection between the trie and the DB will be even more integrated.

burdges: Tries were always an incorrect design choice though. 

Intuitively this is clear:  If you have internal nodes then you expect to double PoV size over storing data only at leaves.

There is experemental evidence by others that the possible optimizations from internal nodes do not really pan out in pracitce.  You could do that sort of optimization with leaves anyways.   Also that radix 2 is always the correct choice when using symmetric hashing.

Now if you want a quick hack that redices PoV size for tries by replacing blake2 by blake3 compression plus binary merkle proofs, then one could do this.  I'd rather see something like what Rob proposes, meaning avoid tries entirely. 

Anyways we should not bake tries in any deeper.  If anything, our goals should be to depricate tries entirely.

cheme: > Intuitively this is clear: If you have internal nodes then you expect to double PoV size over storing data only at leaves.

Not if the internal node should always be part of the pov, but yes not having internal node sounds like a good idea (even if I would prefer having single length key in this case (losing ordering is not really good: if ordering don't matter when key length differ, I am not sure there is a real use of having shorter keys).

I think everybody is thinking in the same direction here. (this PR is not related to it (it does not change anything about the trie, naming some branches new-trie is confusing I admit, it is really only a change for addressing merkle structure, which I think is not strictly incomaptible with a single state deterministic node address storage)).

burdges: Anyone looked at https://github.com/cosmos/iavl yet?

cheme:
> Anyone looked at https://github.com/cosmos/iavl yet?

I remember really liking it at first (and more generally the idea to use a custom balanced tree), but the fact that the structure depends on the modification order is a bit holding me back.

substrate/primitives/state-machine/src/backend.rs
rphmeier: > For the next trie generation, I envision the connection between the trie and the DB will be even more integrated.

Yes, this seems very reasonable for maximum performance. Substrate should support some generality over storage backend. There are a few good options we have seen:
  * Jellyfish hex (small witness size, compact disk representation)
  * IAVL+ (balanced, hard to DoS, lower PoV usage)
  * Verkle (compact proofs, high compute costs, ZK friendly)
  * Binary Merkle (can be SSD optimized)

I personally suggest we should not assume that there is one "next generation" of the trie. Different options will work for different use-cases and architectures.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5343475

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5344811

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5356288

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5356758

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5357805

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5361526

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5636817

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5636983

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5639238

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5639416

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5639879

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5640482

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5640945

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5641274

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5641760

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5648970

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5649139

cheme:
looks like derived branch is running ok on kusuma, so I am switching this from draft to review.
Note that from the branch running, there were a few change, https://github.com/paritytech/polkadot-sdk/pull/3386/commits/a769dcc2445fabbee4c911ae0f8cc97e25b78e76 is the one that may break things.

substrate/client/db/src/lib.rs
```diff
@@ -1880,66 +2143,85 @@ impl<Block: BlockT> Backend<Block> {
 		&self,
 		transaction: &mut Transaction<DbHash>,
 		displaced: &DisplacedLeavesAfterFinalization<Block>,
+		clean_body: bool,
+		clean_state: bool,
```
cheme, line 2147: probably something I wrote, but having two bool here is a bit hard to read, likely an enum would be nicer.

cheme: actually two different function would make more senses (I think it is either call for body or state, not both at the same time).

cheme (commented):
looked at recent change, did not see things wrong (most my comment are related to old code I think).

PR #3389, opened 2024-02-19
ProposerFactory impl Clone

tmpolaczyk:
In Tanssi, we need a way to stop the collator code and then start it again. This is to support rotating the same collator between different runtimes. Currently, this works very well, except for the proposer metrics, because they only get registered the first time they are started. Afterwards, we see this warning log:

> Failed to register proposer prometheus metrics: Duplicate metrics collector registration attempted

~~So this PR adds a method to set metrics, to allow us to register metrics manually before creating the `ProposerFactory`, and then clone the same metrics every time we need to start the collator.~~ Implemented Clone instead

### DIFF
--- a/substrate/client/basic-authorship/src/basic_authorship.rs
+++ b/substrate/client/basic-authorship/src/basic_authorship.rs
@@ -189,11 +189,6 @@ impl<A, C, PR> ProposerFactory<A, C, PR> {
 	pub fn set_soft_deadline(&mut self, percent: Percent) {
 		self.soft_deadline_percent = percent;
 	}
-
-	/// Set metrics.
-	pub fn set_metrics(&mut self, metrics: PrometheusMetrics) {
-		self.metrics = metrics;
-	}
 }
 
 impl<Block, C, A, PR> ProposerFactory<A, C, PR>
--- a/substrate/client/basic-authorship/src/basic_authorship.rs
+++ b/substrate/client/basic-authorship/src/basic_authorship.rs
@@ -173,6 +173,11 @@ impl<A, C, PR> ProposerFactory<A, C, PR> {
 	pub fn set_soft_deadline(&mut self, percent: Percent) {
 		self.soft_deadline_percent = percent;
 	}
+
+	/// Set metrics.
+	pub fn set_metrics(&mut self, metrics: PrometheusMetrics) {
+		self.metrics = metrics;
+	}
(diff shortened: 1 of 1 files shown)


bkchr:
> An alternative would be to `#[derive(Clone)]` for `ProposerFactory`, but that seemed more intrusive.

I would propose this way :D

tmpolaczyk:
> I would propose this way :D

Ok, I manually implemented `Clone`, because deriving it would add unnecessary bounds such as `C: Clone` because of generics. And I wasn't able to use `CloneNoBound` because `frame_support` is not a dependency here, not sure if I should just add it.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5261781

substrate/client/basic-authorship/src/basic_authorship.rs
```diff
@@ -87,6 +87,24 @@ pub struct ProposerFactory<A, C, PR> {
...
+			client: self.client.clone(),
+			transaction_pool: self.transaction_pool.clone(),
+			metrics: self.metrics.clone(),
+			default_block_size_limit: self.default_block_size_limit.clone(),
+			soft_deadline_percent: self.soft_deadline_percent.clone(),
+			telemetry: self.telemetry.clone(),
+			include_proof_in_block_size_estimation: self
+				.include_proof_in_block_size_estimation
+				.clone(),
+			_phantom: self._phantom.clone(),
```
bkchr, line 103:
```suggestion
			_phantom: self._phantom,
```

substrate/client/basic-authorship/src/basic_authorship.rs
```diff
@@ -173,6 +191,11 @@ impl<A, C, PR> ProposerFactory<A, C, PR> {
 	pub fn set_soft_deadline(&mut self, percent: Percent) {
 		self.soft_deadline_percent = percent;
 	}
+
+	/// Set metrics.
+	pub fn set_metrics(&mut self, metrics: PrometheusMetrics) {
+		self.metrics = metrics;
+	}
```
bkchr, line 198:
```suggestion
```

bkchr (approved):

skunert (approved):

PR #3391, opened 2024-02-19
contracts: Passing zero as `deposit_limit` to call_v2 correctly limits the deposit to nothing

athei:
Only affects unstable `v2` versions of `call` and `instantiate`. Also only when `0` was passed as `desposit_limit`. Old behaviour was allowing all the remaining balance. New correct behaviour is to not allow any deposit. Otherwise both passing the `SENTINEL` and encoding `0` would mean the same thing.


athei:
Thank you github rebase function.

PR #3424, opened 2024-02-21
Snowbridge benchmark tests fix

claravanstaden:
When running `cargo test -p bridge-hub-rococo-runtime --features runtime-benchmarks`, two of the Snowbridge benchmark tests fails. The reason is that when the runtime-benchmarks feature is enabled, the `NoopMessageProcessor` message processor is used. The Snowbridge tests rely on the outbound messages to be processed using the message queue, so that we can check the expected nonce and block digest logs.

This PR changes the conditional compilation to only use `NoopMessageProcessor` when compiling the executable to run benchmarks against, not when running tests.


cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs
```diff
@@ -367,11 +365,11 @@ parameter_types! {
 impl pallet_message_queue::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type WeightInfo = weights::pallet_message_queue::WeightInfo<Runtime>;
-	#[cfg(feature = "runtime-benchmarks")]
+	#[cfg(all(not(feature = "std"), feature = "runtime-benchmarks"))]
```
acatangiu, line 368: why introduce the additional filter on `feature = "std"`?

claravanstaden: When running `cargo test -p bridge-hub-rococo-runtime --features runtime-benchmarks`, the tests should rely on `BridgeHubMessageRouter` (because the tests check that the message is actually processed). However, when running the actual benchmarks, `BridgeHubMessageRouter` causes a wasm trap error. So it should only use `NoopMessageProcessor ` when running the benchmarks, not tests with the benchmark feature enabled. I agree this isn't an ideal fix though, I'll look into it further later

ggwpez: I think the question was why to use `feature = "std"` instead of `test`.

claravanstaden: The `test` feature doesn't work. We've had the same issue with the fork versions for a local net and sepolia: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L592

`std` is enabled when running tests, so `test` would have ideal but it doesn't seem to work. [We added it for clarity](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L592) but it doesn't actually trigger when running tests.

claravanstaden: Anyways, please let me know if this is acceptable or not. 😅 If not I'll try the BenchmarkHelper way, it will just take a bit more time.

acatangiu: To me it's acceptable if you add a code comment explaining above

acatangiu (approved):

serban300 (approved):

Issue #3400, opened 2024-02-20
Bridges: Add manual asset transfer test for Polkadot<>Kusama


serban300:
Similar to the Rococo<>Westend tests that we have, we should add a Polkadot<>Kusama asset transfer test that could be run manually as a minimal sanity check.

serban300:
Closed by https://github.com/polkadot-fellows/runtimes/pull/198

PR #3429, opened 2024-02-21
Bridge testing: Add manual test for P<>K asset transfer

serban300:
Closes: https://github.com/paritytech/polkadot-sdk/issues/3400

Bridge testing: Add manual test for P<>K asset transfer


bridges/testing/README.md
```diff
@@ -9,23 +9,37 @@ To start those tests, you need to:
...
 
-After that, you could run tests with the `./run-tests.sh` command. Hopefully, it'll show the
-"All tests have completed successfully" message in the end. Otherwise, it'll print paths to zombienet
-process logs, which, in turn, may be used to track locations of all spinned relay and parachain nodes.
+Extra steps for the Polkadot<>Kusama test:
+
+- clone the [`polkadot-fellows/runtimes`](https://github.com/polkadot-fellows/runtimes) locally and do the following
+  adaptations (as an example see:
+  https://github.com/serban300/runtimes/commit/4e5da529e8b3543e0400a6e524d5e883db1610ed):
+    - Add the `sudo` pallet to the Polkadot and Kusama runtimes and give sudo rights to Alice.
```
svyatonik, line 34: You could use a [`sudofi`](https://github.com/paritytech/parachain-utils/tree/main/tools/sudofi) to add pallet to runtimes. For chain specs, maybe: maybe we could patch `chain-spec-generator` for that? I.e. add `sudo` feature? Alternatively, we could prepare a chain spec patch and a `jq` call to patch chain spec, generated by `chain-spec-generator`

bridges/testing/README.md
```diff
@@ -9,23 +9,37 @@ To start those tests, you need to:
...
+    - Modify the Polkadot and Kusama runtimes such that the epoch time = 1 minute. The `chain-spec-generator` can't be
```
svyatonik, line 35: I'd also suggest to automate this step by etiher fixing the `chain-spec-generator` tool, or by preparing a json patch + calling `jq`

bridges/testing/README.md
```diff
@@ -9,23 +9,37 @@ To start those tests, you need to:
...
+      compiled with the `fast-runtime` feature.
+    - Modify the Polkadot BridgeHub and the Kusama BridgeHub genesis config in order to make Alice bridge owner.
```
svyatonik, line 37: Also a good candidate for patching with json

cumulus/polkadot-parachain/chain-specs/bridge-hub-kusama-local.json
svyatonik, line 1: Do you really need those chainspecs here? I thought you are using `chain-spec-generator` binary to generate chain specs for all involved chains?

svyatonik (commented):
Left a couple of questions, but my main concern is whether we really want to have it here, in `polkadot-sdk`? Why not in the `runtimes` or a separate repo (if zombienet tests are not accepted there)? I'd like to hear other opinions on that

bridges/testing/README.md
serban300: Thanks for the suggestion. I didn't know about `sudofi`. Will try.

cumulus/polkadot-parachain/chain-specs/bridge-hub-kusama-local.json
serban300: Yes, I'm using `chain-spec-generator`, but even so, AFAIU with the current approach for the `BridgeHub` initialization we need to add `PolkadotLocal` and `KusamaLocal` variants to the `BridgeHubRuntimeType` enum and we need a default config for them. Will look if it's possible to change the approach.

serban300:
> Left a couple of questions, but my main concern is whether we really want to have it here, in `polkadot-sdk`? Why not in the `runtimes` or a separate repo (if zombienet tests are not accepted there)? I'd like to hear other opinions on that

I agree it would be more appropriate to have it in `repos`. The main issue is that I'm not sure how to package these bash and js utils in order to reuse them there and just tried to add this test ASAP to have a sanity check. But I will look into this issue now.

cumulus/polkadot-parachain/chain-specs/bridge-hub-kusama-local.json
svyatonik: Sorry, I don't get it :) By initialization you mean calling the `bridges/testing/environments/polkadot-kusama/bridges_polkadot_kusama.sh` with e.g. `init-bridge-hub-kusama-local` param or initialization of the GRANDPA pallet? In former case - can you, please, point to the place where this enum is touched? I thought we are just using some RPCs there? And in the latter case (GRANDPA init) I'm also not sure where it is required

acatangiu:
> my main concern is whether we really want to have it here, in `polkadot-sdk`? Why not in the `runtimes` or a separate repo (if zombienet tests are not accepted there)? I'd like to hear other opinions on that

I agree. `polkadot-sdk` is not the best place for this.
Ideally it would live in https://github.com/polkadot-fellows/runtimes/ next to Polkadot and Kusama networks it tests.
If that can't be done, next option would be in https://github.com/paritytech/parity-bridges-common next to the Polkadot and Kusama relayers (which it also tests).

cumulus/polkadot-parachain/chain-specs/bridge-hub-kusama-local.json
serban300: Sorry, I mean when initializing the zombienet node. At some point it does:
```
./target/release/polkadot-parachain export-genesis-state -d /tmp/bridges-tests-run-lNYWQ/bridge_hub_polkadot_local_network/export-genesis-state/1002 --chain /tmp/bridges-tests-run-lNYWQ/bridge_hub_polkadot_local_network/bridge-hub-polkadot-local-1002-polkadot-local.json /tmp/bridges-tests-run-lNYWQ/bridge_hub_polkadot_local_network/cfg/genesis-state-1002
```

And here I'm getting `Thread 'main' panicked at 'Invalid value: "Value 'bridge-hub-polkadot-local' is not configured yet"', cumulus/polkadot-parachain/src/command.rs:116`

I just saw the enum and added the variants that I needed. But it might be easily fixable in a different way, not sure. I'll look.

svyatonik: Ah, ok I see. Thanks for explaining! Maybe we should do something similar to AH then - there's no `asset-hub-kusama-local.json`, just `asset-hub-kusama.json`. But I'm not sure as well

bridges/testing/README.md
serban300: Good idea. Opened a PR in the `runtimes` repo: https://github.com/polkadot-fellows/runtimes/pull/196

bridges/testing/README.md
serban300: Done

bridges/testing/README.md
serban300: `sudofi` doesn't seem to work correctly on the `runtimes` repo, and I'm not sure if we could patch the chain spec. For example I don't know if we could open the HRMP channels by patching the chain spec. I would keep this step manual for the moment and maybe improve it in the future.

serban300:
> > > Left a couple of questions, but my main concern is whether we really want to have it here, in `polkadot-sdk`? Why not in the `runtimes` or a separate repo (if zombienet tests are not accepted there)? I'd like to hear other opinions on that
> > 
> > 
> > I agree it would be more appropriate to have it in `runtimes`. The main issue is that I'm not sure how to package these bash and js utils in order to reuse them there and just tried to add this test ASAP to have a sanity check. But I will look into this issue now.
> 
> Maybe we could just download everything we need from the startup script using `wget` and github refs? We may bump refs when something changes like we'll be bumping e.g. `substrate-relay` docker image ref from this repo. In any case - it isn't a blocker if we all agree to keep it here, but I thought that the point is to remove Kusama/Polkadot references form this repo at all

Thanks for the suggestion ! `wget` sounds good. Will try it.

serban300:
As discussed, opened a PR in the `runtimes` repo: https://github.com/polkadot-fellows/runtimes/pull/198

Closing this one.

PR #3433, opened 2024-02-21
Check that the validation code matches the parachain code

bkchr:
This introduces a check to ensure that the parachain code matches the validation code stored in the relay chain state. If not, it will print a warning. This should be mainly useful for parachain builders to make sure they have setup everything correctly.

### DIFF
--- a/cumulus/client/consensus/aura/src/collators/mod.rs
+++ b/cumulus/client/consensus/aura/src/collators/mod.rs
@@ -20,5 +20,60 @@
 //! included parachain block, as well as the [`lookahead`] collator, which prospectively
 //! builds on parachain blocks which have not yet been included in the relay chain.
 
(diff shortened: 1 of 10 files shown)


skunert (approved):

cumulus/client/consensus/aura/src/collators/mod.rs
```diff
@@ -20,5 +20,60 @@
...
+				%error,
+				?relay_parent,
+				%para_id,
+				"Failed to fetch validation code hash",
+			);
+			return
+		},
+	};
+
+	match state_validation_code_hash {
```
koute, line 58: Could be turned into a `let Some(state) = state_validation_code_hash else { ... }`.

koute (approved):

davxy (approved):

PR #3439, opened 2024-02-22
Fix `wasm-builder` not exiting if compilation fails

koute:
This PR fixes a subtle bug in `wasm-builder` first introduced in https://github.com/paritytech/polkadot-sdk/pull/1851 (sorry, my bad! I should have caught this during review) where the status code of the `cargo` subprocess is not properly checked, which results in builds silently succeeding when they shouldn't (that is: if we successfully build a runtime blob, and then modify the code so that it won't compile, and recompile it again, then the build will succeed and silently use the *old* blob).

cc @athei This is the bug you were seeing.

[edit]Also fixes a similar PolkaVM-specific bug where I accidentally used the wrong comparison operator.[/edit]

### DIFF
--- a/substrate/utils/wasm-builder/src/wasm_project.rs
+++ b/substrate/utils/wasm-builder/src/wasm_project.rs
@@ -877,7 +877,7 @@ fn build_bloaty_blob(
 			if polkavm_path
 				.metadata()
 				.map(|polkavm_metadata| {
-					polkavm_metadata.modified().unwrap() >= elf_metadata.modified().unwrap()
+					polkavm_metadata.modified().unwrap() < elf_metadata.modified().unwrap()
 				})
 				.unwrap_or(true)
 			{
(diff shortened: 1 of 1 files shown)


michalkucharczyk (approved):

davxy (approved):

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -855,7 +855,7 @@ fn build_bloaty_blob(
 	println!("{} {}", colorize_info_message("Using rustc version:"), cargo_cmd.rustc_version());
 
 	// Use `process::exit(1)` to have a clean error output.
-	if build_cmd.status().map(|s| s.success()).is_err() {
+	if !matches!(build_cmd.status().map(|s| s.success()), Ok(true)) {
```
bkchr, line 858:
```suggestion
	if build_cmd.status().map_or(true, |s| !s.success()) {
```

bkchr (approved):

substrate/utils/wasm-builder/src/wasm_project.rs
koute: Sorry, I added this to the merge queue before I saw your comment. I can change this in my next PR.

bkchr: Not that important :)

PR #3441, opened 2024-02-22
[prdoc] Optional SemVer bumps and Docs

ggwpez:
Changes:
- Add an optional `bump` field to the crates in a prdoc.
- Explain the cargo semver interpretation for <1 versions in the release doc.

### DIFF
--- a/prdoc/schema_user.json
+++ b/prdoc/schema_user.json
@@ -3,8 +3,8 @@
     "$id": "https://raw.githubusercontent.com/paritytech/prdoc/master/prdoc_schema_user.json",
     "version": {
       "major": 1,
-      "minor": 0,
-      "patch": 1
+      "minor": 1,
+      "patch": 0
     },
     "title": "Polkadot SDK PRDoc Schema",
     "description": "JSON Schema definition for the Polkadot SDK PR documentation",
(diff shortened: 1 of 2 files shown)


bkchr (approved):

Morganamilo (approved):

prdoc/schema_user.json
```diff
@@ -125,10 +124,16 @@
...
+            "$ref": "#/$defs/semver_bump"
+          },
           "note": {
             "type": "string"
           }
-        }
+        },
+        "required": [
+          "name"
+        ]
```
liamaharon, line 136: Bump also needs to be required if we are to follow SemVer in our new release process with the stable/unstable branches.
```suggestion
        "required": [
          "name",
          "bump"
        ]
```

liamaharon (changes requested):

prdoc/schema_user.json
ggwpez: As first step i wanted to make it optional, so we can try it out on a few crates ourselves to ensure that all the tooling/CI checks works.  
If we start with it being required, then many devs will put wrong stuff in there accidentally and we dont notice.

liamaharon: Ok, as long as we remember to make it required before we cut the next release yah?

liamaharon (approved):

PR #3444, opened 2024-02-22
Statement Distribution Per Peer Rate Limit

Overkillus:
- [x] Drop requests from a PeerID that is already being served by us.
- [x] Don't sent requests to a PeerID if we already are requesting something from them at that moment (prioritise other requests or wait).
- [x] Tests
- [ ] ~~Add a small rep update for unsolicited requests (same peer request)~~ not included in original PR due to potential issues with nodes slowly updating
- [x] Add a metric to track the amount of dropped requests due to peer rate limiting
- [x] Add a metric to track how many time a node reaches the max parallel requests limit in v2+

Helps with but does not close yet: https://github.com/paritytech-secops/srlabs_findings/issues/303


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5312613

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5329413

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5332328

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3335,33 +3337,55 @@ pub(crate) async fn respond_task(
...
+				};
 
-		let req = match receiver.recv(|| vec![COST_INVALID_REQUEST]).await.into_nested() {
-			Ok(Ok(v)) => v,
-			Err(fatal) => {
-				gum::debug!(target: LOG_TARGET, error = ?fatal, "Shutting down request responder");
-				return
+				// If peer currently being served drop request
+				if active_peers.contains(&request.peer) {
+					continue
```
sandreim, line 3360: Shouldn't this be handled in the context of a candidate/relay parent ? Otherwise we would drop legit requests.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5352693

polkadot/node/network/statement-distribution/src/v2/mod.rs
Overkillus: This is definitely a fair point. Rob mentioned it https://github.com/paritytech-secops/srlabs_findings/issues/303#issuecomment-1588185410 as well.

It's a double-edged sword. If we make it 1 per context then the request/response protocol is more efficient, but with async backing enabled (which to my understanding broadens the scope of possible contexts) the mpsc channel (size 25) can still be easily overwhelmed. The problem of dropped requests gets somewhat a bit better after the recent commit `[Rate limit sending from the requester side](https://github.com/paritytech/polkadot-sdk/pull/3444/commits/58d828db45b7c023e8a8266d3180be4361723519)` which also introduces the rate limit on the requester side so effort is not wasted.

If we are already requesting from a peer and want to request something else as well we simply wait for the first one to finish.

What could be argued is the limit as a small constant 1-3 instead of 1 per peerID. I would like to do some testing and more reviews and changing it to a configurable constant instead of a hard limit to 1 is actually a trivial change. (I don't think it's necessary. Better have a single completely sent candidate than two half sent candidates.)

Truth be told this whole solution is far from ideal. It's a fix but the final solution is to manage DoS at a significantly lower level than in parachain consensus. It will do for now but DoS needs to be looked on in more detail in near future.

Overkillus: > I would rather put the constraint on the responder side rather than the requester. The reasoning is to avoid any bugs in other clients implementation due to the subtle requirement to wait for first request to finish.

I agree that constraint on the responder side is more important, but having it in both places seems even better.

> The reasoning is to avoid any bugs in other clients implementation due to the subtle requirement to wait for first request to finish.

Even without this PR the risk seems exactly the same. We already were checking for the number of parallel requests being handled in the requester so even if there would be bugs in other clients they could surface there with or without this PR.

The only really bad scenario would be if for some reason we keep a PeerID marked as still being active when it really isn't effectively blacklisting that peer. If that would happen it means that the future connected to that PeerID is still in the [`pending_responses`](https://github.com/paritytech/polkadot-sdk/blob/f5de090aa56d0801f043c2b13892191b3d0c9f50/polkadot/node/network/statement-distribution/src/v2/requests.rs#L424) which would eventually brick the requester anyway since we cannotgo over the max limit of [`MAX_PARALLEL_ATTESTED_CANDIDATE_REQUESTS`](https://github.com/paritytech/polkadot-sdk/blob/f5de090aa56d0801f043c2b13892191b3d0c9f50/polkadot/node/network/statement-distribution/src/v2/requests.rs#L327-L329). Both situations are just as fatal, but the new system at least has extra protections against DoS attempts so it's a security gain.

By adding the constraint on the requester side we limit the wasted effort and can potentially safely add reputation updates if we get those unsolicited requests which protects us from DoS even further. (Some honest requests might slip through so the rep update needs to be small.)

>  Maybe instead of relying on a hardcoded constant we can derive the value based on the async backing parameters.

We potentially could but also I don't why this value would need to change a lot. Even if we change some async backing params we should be fine. It seems to be more sensitive to [channel size](https://github.com/paritytech/polkadot-sdk/blob/f5de090aa56d0801f043c2b13892191b3d0c9f50/polkadot/node/network/protocol/src/request_response/mod.rs#L327) as we want to make it hard to dominate that queue as a malicious node and the `MAX_PARALLEL_ATTESTED_CANDIDATE_REQUESTS`.

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3335,33 +3337,55 @@ pub(crate) async fn respond_task(
...
+					},
+				};
 
-		let req = match receiver.recv(|| vec![COST_INVALID_REQUEST]).await.into_nested() {
-			Ok(Ok(v)) => v,
-			Err(fatal) => {
-				gum::debug!(target: LOG_TARGET, error = ?fatal, "Shutting down request responder");
-				return
+				// If peer currently being served drop request
+				if active_peers.contains(&request.peer) {
```
alexggh, line 3359: 1. Are we 100 percent sure peers won't have more than one request in-flight ? 
2. What would a peer whose requests gets dropped see ?
3. If we don't expect this too happen, I would warn if two frequent, see warn_if_frequent.

polkadot/node/network/statement-distribution/src/v2/requests.rs
```diff
@@ -462,10 +478,17 @@ fn find_request_target_with_update(
...
 	props: &RequestProperties,
 	peer_advertised: impl Fn(&CandidateIdentifier, &PeerId) -> Option<StatementFilter>,
+	response_manager: &ResponseManager,
 ) -> Option<PeerId> {
 	let mut prune = Vec::new();
 	let mut target = None;
 	for (i, p) in known_by.iter().enumerate() {
+
+		// If we are already sending to that peer, skip for now
+		if response_manager.is_sending_to(p) {
```
alexggh, line 488: I would investigate here, if this might create more Nones, than before, I'm under impression that we call this functions that aren't that many peers which advertised the candidates to us.

Also, if we don't expect this to  happen let's sprinkle with warnings, so that we know if we hit this conditions in the wild.

alexggh (commented):
Added some comments, what is not clear to me is why `can_request`  from `answer_request` is not considered fast enough, that already has the benefit that we punish peers that are misbehaving.

Just my2cents, if what we want is to prevent peers from exhausting the buget for the number of connections, I would think the proper fix needs to go deeper into the layer network, rather than here, since there is a lot time spent until the message reaches here.

polkadot/node/network/statement-distribution/src/v2/mod.rs
Overkillus: > Are we 100 percent sure peers won't have more than one request in-flight?

They can have more than 1 and this is the main reason why this change is getting implemented so we drop them when it happens. To reduce the wasted requester work there is also the rate limiter on the requester side so they don't request from the same peer.

Old nodes or malicious nodes can definitely send more requests. New nodes shouldn't have more than 1 in-flight, but even if something goes wrong and they have more than nothing should collapse.

> What would a peer whose requests gets dropped see?

First of all he shouldn't have sent it since he also should have known that he is requesting something else already. If it was an option he should have requested from a different peer that has the same data. 

If it happens anyway he would see the same thing as if the packet was lost along the way or if the packet was undecodable so AFAIK nothing. He might be potentially wasting our time so best to ignore it. Request will remain `in_flight` with no response and eventually will be reattempted `next_retry_time`. So the absolute worst case here as far as I can see is attempting again after the delay.

> If we don't expect this too happen, I would warn if two frequent, see warn_if_frequent.

We indeed don't expect this to happen ONCE people update. Old nodes can still send us too many requests since they will not have this rate limit so in this period this might happen more often that preferred.

Overkillus:
re: @alexggh 

> Added some comments, what is not clear to me is why can_request from answer_request is not considered fast enough, that already has the benefit that we punish peers that are misbehaving.

Technically fair point but I think there's a reason. It's not about speed but technically yeah `respond_task` is a bit quicker than `answer_request` so it seems like a minimal gain here. It also would logically make sense for this logic to be in something called `can_request`.

Even before this change we tracked the parallel requests up to the limit (`MAX_PARALLEL_ATTESTED_CANDIDATE_REQUESTS`) in `respond_task` and my guess why is simply that `respond_task` encapsulates the async component of the request filtering procedure while `can_request` focuses on the simpler sync filtering. To properly filter out the most current `PeerID` this logic here needed to be async as well hence it's also in the async `respond_task`.

The `can_request` is primarily doing the filtering based on the knowledge we have about what was sent/received previously while the async `respond_task` is very time sensitive as we are operating on stuff that is _currently_ being sent over.

> Just my2cents, if what we want is to prevent peers from exhausting the buget for the number of connections, I would think the proper fix needs to go deeper into the layer network, rather than here, since there is a lot time spent until the message reaches here.

I 100% agree with you here. I said this myself and me and Robert have already spoken about it and this approach will be pursued. That's why this solution is not ideal but it at least makes the attack vector much harder to pull off with minimal changes to the current code-base. Not ideal but worth the tiny effort before we tackle the wider more general issue at the lower level.

polkadot/node/network/statement-distribution/src/v2/requests.rs
Overkillus: > I would investigate here, if this might create more Nones, than before

It will definitely create more `None`s. How much more I'm not certain, but generally I don't think that those are bad because it means that we are already downloading other data from that same peer so might as well finish that before starting another. Additionally it means that we will prioritise another node with that data so it might even help with load balancing.

> Also, if we don't expect this to happen let's sprinkle with warnings, so that we know if we hit this conditions in the wild.

We do expect this to happen. When we iterate over the peers this is doing the filtering (removes peers we are already sending to). So in the wild that might happen somewhat frequently especially with async backing.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5601644

polkadot/node/network/statement-distribution/src/v2/mod.rs
burdges: You need not choose a uniform distribution over recent relay parents either.  Instead the full value on the first couple, then 1/2 and 1/4 for a few, and then 1 one for a while.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5724775

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5725972

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5727272

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5728130

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5728298

polkadot/node/malus/src/variants/spam_statement_requests.rs
```diff
@@ -0,0 +1,160 @@
...
+		&self,
+		_subsystem_sender: &mut Sender,
+		msg: FromOrchestra<Self::Message>,
+	) -> Option<FromOrchestra<Self::Message>> {
+		match msg {
+			FromOrchestra::Communication {
+				msg: NetworkBridgeTxMessage::SendRequests(mut requests, if_disconnected),
+			} => {
+				// AttestedCandidateV2 requests arrive 1 by 1
+				if requests.len() == 1 {
```
sandreim, line 71: This might be true for now, but can be easily broken in the future. Why not loop through all incoming requests ?

polkadot/node/malus/src/variants/spam_statement_requests.rs
```diff
@@ -0,0 +1,160 @@
...
+
+						// Duplicate the request spam_factor times and append to the list
+						for _ in 0..self.spam_factor - 1 {
+							let (new_outgoing_request, _) = OutgoingRequest::new(
+								peer_to_duplicate.clone(),
+								payload_to_duplicate.clone(),
+							);
+							let new_request = Requests::AttestedCandidateV2(new_outgoing_request);
+							requests.push(new_request);
+						}
```
sandreim, line 86: It should work without needing `peer_to_duplicate` and `payload_to_duplicate` clones:

```suggestion
  requests.extend(0..self.spam_factor - 1).into_iter().map(|_| 
                             Requests::AttestedCandidateV2(OutgoingRequest::new(
								peer_to_duplicate.clone(),
								payload_to_duplicate.clone(),
							)))
```

polkadot/node/malus/src/variants/spam_statement_requests.rs
```diff
@@ -0,0 +1,160 @@
...
+		Error,
+	>
+	where
+		RuntimeClient: 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block> + AuxStore,
+		RuntimeClient::Api: ParachainHost<Block> + BabeApi<Block> + AuthorityDiscoveryApi<Block>,
+		Spawner: 'static + SpawnNamed + Clone + Unpin,
+	{
+		gum::info!(
+			target: MALUS,
+			"😈 Started Malus node that sends {:?} statement distribution requests instead of 1.",
```
sandreim, line 143:
```suggestion
			"😈 Started Malus node that duplicates {:?} statement distribution requests by a custom factor.",
```

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3350,33 +3352,57 @@ pub(crate) async fn respond_task(
...
+					gum::debug!(target: LOG_TARGET, "Dropping repeated request from peer");
```
sandreim, line 3375:
```suggestion
					gum::debug!(target: LOG_TARGET, "Dropping additional request from peer");
```

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3350,33 +3352,57 @@ pub(crate) async fn respond_task(
...
+					continue
+				}
+
+				// If we are over parallel limit wait for one to finish
+				if pending_out.len() >= MAX_PARALLEL_ATTESTED_CANDIDATE_REQUESTS as usize {
+					gum::debug!(target: LOG_TARGET, "Over max parallel requests, waiting for one to finish");
+					let result = pending_out.select_next_some().await;
+					let (_, peer) = result;
```
sandreim, line 3383:
```suggestion
					let (_, peer) = pending_out.select_next_some().await;
```

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3350,33 +3352,57 @@ pub(crate) async fn respond_task(
...
+					active_peers.remove(&peer);
+				}
+
+				// Start serving the request
+				let (pending_sent_tx, pending_sent_rx) = oneshot::channel();
+				let peer = request.peer;
+				if let Err(err) = sender
+					.feed(ResponderMessage { request: request, sent_feedback: pending_sent_tx })
```
sandreim, line 3391:
```suggestion
					.feed(ResponderMessage { request, sent_feedback: pending_sent_tx })
```

polkadot/node/malus/src/variants/spam_statement_requests.rs
Overkillus: Handled multiple requests :+1:

polkadot/node/malus/src/variants/spam_statement_requests.rs
Overkillus: Borrowed the extend syntax but still need the clones to satisfy the borrow checker.

polkadot/node/malus/src/variants/spam_statement_requests.rs
Overkillus: reworded

polkadot/node/network/statement-distribution/src/v2/mod.rs
Overkillus: Reworded to `Peer already being served, dropping request` to be even clearer.

polkadot/node/malus/src/variants/spam_statement_requests.rs
```diff
@@ -0,0 +1,161 @@
...
+								let (new_outgoing_request, _) = OutgoingRequest::new(
+									peer_to_duplicate.clone(),
+									payload_to_duplicate.clone(),
+								);
+								Requests::AttestedCandidateV2(new_outgoing_request)
+							}));
+
+						}
+						_ => {
+							new_requests.push(request);
```
sandreim, line 95: We can dedup this line by moving it above the match.

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3350,33 +3352,56 @@ pub(crate) async fn respond_task(
...
-				return
+				// If peer currently being served drop request
+				if active_peers.contains(&request.peer) {
+					gum::debug!(target: LOG_TARGET, "Peer already being served, dropping request");
+					continue
+				}
+
+				// If we are over parallel limit wait for one to finish
+				if pending_out.len() >= MAX_PARALLEL_ATTESTED_CANDIDATE_REQUESTS as usize {
+					gum::debug!(target: LOG_TARGET, "Over max parallel requests, waiting for one to finish");
```
sandreim, line 3381: This one might get spammy on real network. We could downgrade to `trace` to preserve logs readability.

polkadot/node/network/statement-distribution/src/v2/requests.rs
```diff
@@ -422,28 +426,37 @@ impl RequestManager {
...
-		self.pending_responses
-			.next()
-			.await
-			.map(|response| UnhandledResponse { response })
+		if let Some(response) = self.pending_responses.next().await {
+			// Upon receiving a response, remove the peer from the active set.
+			self.active_peers.remove(&response.requested_peer);
+			return Some(UnhandledResponse { response });
+		}
+		None
```
sandreim, line 445: This looks it can be written like prev impl:
```suggestion
                self.pending_responses.next().await.map(|response| {
                          self.active_peers.remove(&response.requested_peer);
                          UnhandledResponse { response }
                })
```

sandreim (approved):
Nice work!

polkadot/node/malus/src/variants/spam_statement_requests.rs
Overkillus: This one is somewhat of a conscious trade-off but if you see a graceful way of doing it feel free to suggest it.

Problem is Request does not implement copy/clone. I could possibly track the index where I push it but this will make it even less readable imo.

alexggh (approved):
Good job!

AndWeHaveAPlan (approved):

PR #3447, opened 2024-02-22
Use generic hash for runtime wasm in resolve_state_version_from_wasm

tmpolaczyk:
Changes the runtime hash algorithm used in `resolve_state_version_from_wasm` from `DefaultHasher` to a caller-provided one (usually `HashingFor<Block>`), to match the one used elsewhere.

This fixes an issue where the runtime wasm is compiled 3 times when starting the `tanssi-node` with `--dev`. With this fix, the runtime wasm is only compiled 2 times. The other redundant compilation is caused by the `GenesisConfigBuilderRuntimeCaller` struct, which ignores the runtime cache.

### DIFF
--- a/substrate/client/chain-spec/src/genesis_block.rs
+++ b/substrate/client/chain-spec/src/genesis_block.rs
@@ -20,6 +20,7 @@
 
 use std::{marker::PhantomData, sync::Arc};
 
+use codec::Encode;
 use sc_client_api::{backend::Backend, BlockImportOperation};
 use sc_executor::RuntimeVersionOf;
 use sp_core::storage::{well_known_keys, StateVersion, Storage};
@@ -44,7 +45,7 @@ where
 		let runtime_code = sp_core::traits::RuntimeCode {
 			code_fetcher: &code_fetcher,
 			heap_pages: None,
-			hash: <H as HashT>::hash(wasm).as_ref().to_vec(),
+			hash: <H as HashT>::hash(wasm).encode(),
 		};
 		let runtime_version = RuntimeVersionOf::runtime_version(executor, &mut ext, &runtime_code)
 			.map_err(|e| sp_blockchain::Error::VersionInvalid(e.to_string()))?;
--- a/substrate/client/chain-spec/src/genesis_block.rs
+++ b/substrate/client/chain-spec/src/genesis_block.rs
@@ -18,7 +18,7 @@
 
 //! Tool for creating the genesis block.
 
-use std::{collections::hash_map::DefaultHasher, marker::PhantomData, sync::Arc};
+use std::{marker::PhantomData, sync::Arc};
 
 use sc_client_api::{backend::Backend, BlockImportOperation};
 use sc_executor::RuntimeVersionOf;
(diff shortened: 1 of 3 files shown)


koute (approved):

substrate/client/chain-spec/src/genesis_block.rs
```diff
@@ -43,12 +43,7 @@ where
 		let runtime_code = sp_core::traits::RuntimeCode {
 			code_fetcher: &code_fetcher,
 			heap_pages: None,
-			hash: {
-				use std::hash::{Hash, Hasher};
-				let mut state = DefaultHasher::new();
-				wasm.hash(&mut state);
-				state.finish().to_le_bytes().to_vec()
-			},
+			hash: sp_core::blake2_256(wasm).to_vec(),
```
bkchr, line 46: This is not correct, you need to get the hasher used by the `Block` type. For most chains this is blake256, but not for all.

bkchr (changes requested):

substrate/client/chain-spec/src/genesis_block.rs
bkchr: I mean "not correct" not being really the right term here, but if we want to fix this, we should fix it properly.

bkchr: https://github.com/paritytech/polkadot-sdk/blob/36b7037c6c0a587c11f9f096b2893a5755dd64ce/substrate/primitives/state-machine/src/backend.rs#L427-L433

bkchr: Change the function to take `H: Hash` and then you can pass `HashingOf<Block>` on the calling side of this function.

tmpolaczyk: Thanks, it was easier than I expected.

substrate/client/chain-spec/src/genesis_block.rs
```diff
@@ -43,12 +44,7 @@ where
 		let runtime_code = sp_core::traits::RuntimeCode {
 			code_fetcher: &code_fetcher,
 			heap_pages: None,
-			hash: {
-				use std::hash::{Hash, Hasher};
-				let mut state = DefaultHasher::new();
-				wasm.hash(&mut state);
-				state.finish().to_le_bytes().to_vec()
-			},
+			hash: <H as HashT>::hash(wasm).as_ref().to_vec(),
```
bkchr, line 47:
```suggestion
			hash: <H as HashT>::hash(wasm).encode(),
```

substrate/client/chain-spec/src/genesis_block.rs
```diff
@@ -18,23 +18,24 @@
 
 //! Tool for creating the genesis block.
 
-use std::{collections::hash_map::DefaultHasher, marker::PhantomData, sync::Arc};
+use std::{marker::PhantomData, sync::Arc};
 
 use sc_client_api::{backend::Backend, BlockImportOperation};
```
bkchr, line 23:
```suggestion
use codec::Encode;
use sc_client_api::{backend::Backend, BlockImportOperation};
```

bkchr (approved):
Ty!

bkchr:
@tmpolaczyk could you add some prdoc? Then accept my changes and we can merge it.

bkchr (approved):

Issue #2560, opened 2023-11-30
Skip `System::Events` in `TryDecodeEntireState`


liamaharon:
@joepetrowski encountered this failed try-runtime check on one of his PRs https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/4551392

This is the offending block https://rococo.subscan.io/block/0x565751cc2bfaf946216484bb34466c1e5a4f75c444235557d48f822974987048?tab=event 

This is likely due to events being emitted in that block using an old storage layout. We avoid this issue with other storage items by running migrations prior to checking if state is decodable. This doesn't work for `System::Events`.

We should modify our decode state checks to skip `System::Events`.

xlc:
can you just modify the try runtime check to call initialize first, which will remove events and bunch other storage or just ignore the events. I just don't get why we need macro magic for this.

liamaharon:
We want this to be a general solution that doesn't make the assumption that the storage has the same characteristics as System::Events. @bkchr pointed out it's possible for downstream users have similar storage, which we can't assume will be cleared out during on_initialize.

xlc:
Can you name an example? Otherwise I will say it is overengineering to build a solution to solve a problem that does not yet exist.

bkchr:
Given the amount of code written, I would not say it is overengineered. Theoretically you could have values set by inherents that do not require to be migrated because you change them in every block any way. Generally, if the macro that generates the code for checking that all types can be decoded wouldn't be the same that we now extend with this option, I would not have proposed to do it this way. 

Generally you could also say that having this entire machinery written in Rust is the wrong way, because you can also just write some piece of javascript that can decode the entire state using the metadata.

xlc:
> Given the amount of code written, I would not say it is overengineered. 

It may not be overengineered too much, nonetheless could still be poor code in the repo.

Let's define the problem. We have a tool try-runtime that's trying to decode all the storages and ensure all of them are decodable. Right now, it decodes the stroages at end of a block, using the metadata from next block? And because metadata could change, it have false positive.

Shouldn't the solution to be using the right metadata?

> Generally you could also say that having this entire machinery written in Rust is the wrong way, because you can also just write some piece of javascript that can decode the entire state using the metadata.

[...]

PR #3454, opened 2024-02-23
Add new storage attr macro `#[disable_try_decode_storage]` and set it on `System::Events`

liamaharon:
Closes https://github.com/paritytech/polkadot-sdk/issues/2560

Allows marking storage items with `#[disable_try_decode_storage]`, and uses it with `System::Events`. 

Question: what's the recommended way to write a test for this? I couldn't find a test for similar existing macro `#[whitelist_storage]`.

### DIFF
--- a/substrate/frame/support/procedural/src/lib.rs
+++ b/substrate/frame/support/procedural/src/lib.rs
@@ -1371,6 +1371,25 @@ pub fn whitelist_storage(_: TokenStream, _: TokenStream) -> TokenStream {
 	pallet_macro_stub()
 }
 
+/// The optional attribute `#[pallet::disable_try_decode_storage]` will declare the
+/// storage as whitelisted from decoding during try-runtime checks. This should only be
+/// attached to transient storage which cannot be migrated during runtime upgrades.
+///
+/// ### Example
+/// ```ignore
+/// #[pallet::storage]
+/// #[pallet::disable_try_decode_storage]
+/// pub(super) type Events<T: Config> = StorageValue<_, Vec<Box<EventRecord<T::RuntimeEvent, T::Hash>>>, ValueQuery>;
+/// ```
+///
+/// NOTE: As with all `pallet::*` attributes, this one _must_ be written as
+/// `#[pallet::disable_try_decode_storage]` and can only be placed inside a `pallet` module in order
+/// for it to work properly.
+#[proc_macro_attribute]
+pub fn disable_try_decode_storage(_: TokenStream, _: TokenStream) -> TokenStream {
+	pallet_macro_stub()
+}
+
 /// The `#[pallet::type_value]` attribute lets you define a struct implementing the `Get` trait
 /// to ease the use of storage types. This attribute is meant to be used alongside
 /// [`#[pallet::storage]`](`macro@storage`) to define a storage's default value. This attribute
--- a/substrate/frame/support/procedural/src/pallet/expand/storage.rs
+++ b/substrate/frame/support/procedural/src/pallet/expand/storage.rs
@@ -834,7 +834,7 @@ pub fn expand_storages(def: &mut Def) -> proc_macro2::TokenStream {
 			.storages
 			.iter()
 			.filter_map(|storage| {
-				if storage.cfg_attrs.is_empty() {
+				if storage.try_decode {
 					let ident = &storage.ident;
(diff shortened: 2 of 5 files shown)


substrate/frame/support/procedural/src/pallet/expand/storage.rs
```diff
@@ -834,7 +834,7 @@ pub fn expand_storages(def: &mut Def) -> proc_macro2::TokenStream {
 			.storages
 			.iter()
 			.filter_map(|storage| {
-				if storage.cfg_attrs.is_empty() {
```
liamaharon, line 837: @kianenigma do you remember why we check `if storage.cfg_attrs.is_empty()` here?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5317738

substrate/frame/support/procedural/src/pallet/expand/storage.rs
liamaharon: It seems it's to avoid this compile error https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5317738

I wonder if there's a better way to handle that.

substrate/frame/support/procedural/src/lib.rs
```diff
@@ -1371,6 +1371,25 @@ pub fn whitelist_storage(_: TokenStream, _: TokenStream) -> TokenStream {
...
+/// The optional attribute `#[pallet::disable_try_decode_storage]` will declare the
+/// storage as whitelisted from decoding during try-runtime checks. This should only be
+/// attached to transient storage which cannot be migrated during runtime upgrades.
+///
+/// ### Example
+/// ```ignore
+/// #[pallet::storage]
+/// #[pallet::disable_try_decode_storage]
+/// pub(super) type Events<T: Config> = StorageValue<_, Vec<Box<EventRecord<T::RuntimeEvent, T::Hash>>>, ValueQuery>;
+/// ```
```
kianenigma, line 1383: No more `ignore` docs please 🙈 a real pallet requires just a few extra lines of boilerplate.

kianenigma (approved):

kianenigma:
I thought an off-band discussion concluded that we don't want to expose this to the end-user facing API, and just hardcode somewhere that `System::events` is an exception.

This PR is not doing that, but it is neatly done.

I would like to see: 

- [ ] the docs for the new macro being fixed
- [ ] the docs for `#[pallet::storage]` updated with just one line that references this new attribute, namely [here](https://github.com/paritytech/polkadot-sdk/blob/e76b244853c52d523e1d5c7a28948573df60df46/substrate/frame/support/src/lib.rs#L2697).


Before merge tho

ggwpez (approved):

liamaharon:
> I thought an off-band discussion concluded that we don't want to expose this to the end-user facing API, and just hardcode somewhere that System::events is an exception.

Basti requested this to be done as an attribute macro rather than some hardcoded exception for `System::Events` [here](https://matrix.to/#/!wVSWmzIkrlPHxGPhqm:parity.io/$BEs5keBJs5Ba5wAjbGLdVNoqRsYzH-fnHjWbyTwvCnc?via=parity.io&via=web3.foundation&via=matrix.org).

xlc:
@liamaharon can you quote it instead of private link

bkchr:
My point basically was/is that the same thing can happen to downstream users. Like some storage item that is always deleted at the beginning of the next block. Thus, you don't really care if you can still decode it.

xlc:
In that case it is the try runtime tool should call initialize hook before validate storage entries.

substrate/frame/support/procedural/src/lib.rs
liamaharon: This seems to be impossible in the `frame-support-procedural` crate, because depending on anything `frame` related results in cyclic deps.

xlc (changes requested):
It will be great if you can help address my concerns before merge https://github.com/paritytech/polkadot-sdk/issues/2560#issuecomment-1963296839

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5367017

Issue #865, opened 2022-02-19
XCM v3: Integrate `NonFungibleAsset` traits


gavofyork:
Integrate `NonFungibleAsset` traits into the XCM assets system to make NFTs transferable between chains.

bkontur:
Some partial work was done: https://github.com/paritytech/polkadot-sdk/pull/2924 and https://github.com/paritytech/polkadot-sdk/pull/2796

And also there is older PR for `NonFungiblesV2Adapter` https://github.com/paritytech/polkadot-sdk/pull/1242

bkontur:
> Integrate NonFungibleAsset traits into the XCM assets system to make NFTs transferable between chains.

@Szegoo @liamaharon Are we missing anything else to accomplish this? Are there any open issues for XCM NFTs transfer?

Szegoo:
The only thing missing AFAIK is benchmarking XCM NFT transfers correctly, which is fixed here: https://github.com/paritytech/polkadot-sdk/pull/3455

PR #3455, opened 2024-02-23
XCM coretime region transfers

Szegoo:
This PR introduces changes enabling the transfer of coretime regions via XCM.

TL;DR: There are two primary issues that are resolved in this PR:

1. The `mint` and `burn` functions were not implemented for coretime regions. These operations are essential for moving assets to and from the XCM holding register.
2. The transfer of non-fungible assets through XCM was previously disallowed. This was due to incorrectly benchmarking non-fungible asset transfers via XCM, which led to assigning it a weight of `Weight::Max`, effectively preventing its execution.

### `mint_into` and `burn` implementation

This PR addresses the issue with cross-chain transferring regions back to the Coretime chain. Remote reserve transfers are performed by withdrawing and depositing the asset to and from the holding registry. This requires the asset to support burning and minting functionality. 

This PR adds burning and minting; however, they work a bit differently than usual so that the associated region record is not lost when burning. Instead of removing all the data, burning will set the owner of the region to `None`, and when minting it back, it will set it to an actual value. So, when cross-chain transferring, withdrawing into the registry will remove the region from its original owner, and when depositing it from the registry, it will set its owner to another account

This was originally implemented in this PR: #3455, however we decided to move all of it to this single PR (https://github.com/paritytech/polkadot-sdk/pull/3455#discussion_r1547324892)

### Fixes made in this PR

- Update the `XcmReserveTransferFilter` on coretime chain since it is meant as a reserve chain for coretime regions.
- Update the XCM benchmark to use `AssetTransactor` instead of assuming `pallet-balances` for fungible transfers. 
- Update the XCM benchmark to properly measure weight consumption for nonfungible reserve asset transfers. ATM reserve transfers via the extrinsic do not work since the weight for it is set to `Weight::max()`.

Closes: https://github.com/paritytech/polkadot-sdk/issues/865


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5330031

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5355832

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5355986

Szegoo:
The benchmark for the coretime chain should be re-run since the weight is currently set to `Weight::max()`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5466093

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -179,24 +166,27 @@ benchmarks! {
...
 		let recipient = [0u8; 32];
 		let versioned_dest: VersionedLocation = destination.into();
 		let versioned_beneficiary: VersionedLocation =
 			AccountId32 { network: None, id: recipient.into() }.into();
 		let versioned_assets: VersionedAssets = assets.into();
 	}: _<RuntimeOrigin<T>>(send_origin.into(), Box::new(versioned_dest), Box::new(versioned_beneficiary), Box::new(versioned_assets), 0)
-	verify {
-		// verify balance after transfer, decreased by transferred amount (+ maybe XCM delivery fees)
-		assert!(<pallet_balances::Pallet<T> as Inspect<_>>::balance(&caller) <= balance - transferred_amount);
-	}
```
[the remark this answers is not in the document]
Szegoo: Tested locally, we can actually withdraw  exactly `transferred_amount` since the fee is covered by with `ensure_successful_delivery`

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
```diff
@@ -740,8 +740,15 @@ impl_runtime_apis! {
...
 
 				fn reserve_transferable_asset_and_dest() -> Option<(Asset, Location)> {
-					// Reserve transfers are disabled
-					None
+					// Coretime chain can reserve transfer regions to some random parachain.
+					let raw_region_id = 42;
+					Some((
+						Asset {
+							fun: NonFungible(Index(raw_region_id)),
+							id: AssetId(Location::new(0, [PalletInstance(50)]))
```
Szegoo, line 748: This will still fail since we don't support minting regions, resulting in setting the weight to the max value.

I added 'region minting' here: https://github.com/paritytech/polkadot-sdk/pull/3553, however, that would still disallow this.

prdoc/pr_3455.prdoc
```diff
@@ -0,0 +1,18 @@
...
+      With this PR, the benchmarks for pallet-xcm no longer assume the use of 
+      `pallet-balances` for fungible implementations. Furthermore, with this PR 
+      non-fungible teleport and reserve transfers are properly benchmarked.
+
+crates: 
+  - name: pallet-xcm
+  - name: contracts-rococo-runtime
+  - name: coretime-rococo-runtime
+  - name: coretime-westend-runtime
+  - name: coretime-westend-runtime
```
franciscoaguirre, line 18: Duplicate

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -134,35 +122,34 @@ benchmarks! {
...
-		let balance = balance + transferred_amount;
-		assert_eq!(<pallet_balances::Pallet<T> as Inspect<_>>::balance(&caller), balance);
+		match &asset.fun {
+			Fungible(amount) => {
+				// Add transferred_amount to origin
+				<T::XcmExecutor as XcmAssetTransfers>::AssetTransactor::deposit_asset(
+					&Asset { fun: Fungible(*amount), id: asset.id },
+					&origin_location,
+					None,
+				).map_err(|_| BenchmarkError::Override(BenchmarkResult::from_weight(Weight::MAX)))?;
```
franciscoaguirre, line 132: Some error logging would be nice:

```suggestion
				).map_err(|error| {
				  log::error!("Asset couldn't be deposited, error: {:?}", error);
				  BenchmarkError::Override(BenchmarkResult::from_weight(Weight::MAX))
				})?;
```

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -134,35 +122,34 @@ benchmarks! {
...
+			},
+			NonFungible(instance) => {
+				<T::XcmExecutor as XcmAssetTransfers>::AssetTransactor::deposit_asset(&asset, &origin_location, None)
+					.map_err(|_| BenchmarkError::Override(BenchmarkResult::from_weight(Weight::MAX)))?;
```
franciscoaguirre, line 136: Same

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
Szegoo: @franciscoaguirre Any ideas on what we might do in this regard?

acatangiu: cc @seadanda

bkontur: @Szegoo is this still an issue? I think you need to create a region with `raw_region_id=42` first, so maybe some broker pallet magic, e.g.:
```
        let caller: T::AccountId = whitelisted_caller();
		T::Currency::set_balance(
			&caller.clone(),
			T::Currency::minimum_balance().saturating_add(20u32.into()),
		);

        /// TODO: maybe some more magic e.g. `setup_and_start_sale` in the pallet_broker's benchmarks
		let region = Broker::<T>::do_purchase(caller.clone(), 10u32.into())?;
```

or alternatively, simply inserting some data could do the trick:
```
let raw_region_id = 42;
let region = RegionRecord<AccountId, Balance> {
	end: ...,
	owner: ...,
	paid: ...,
};
Regions::<T>::insert(&raw_region_id, &region);
```

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -134,35 +122,40 @@ benchmarks! {
...
 		let recipient = [0u8; 32];
 		let versioned_dest: VersionedLocation = destination.into();
 		let versioned_beneficiary: VersionedLocation =
 			AccountId32 { network: None, id: recipient.into() }.into();
 		let versioned_assets: VersionedAssets = assets.into();
 	}: _<RuntimeOrigin<T>>(send_origin.into(), Box::new(versioned_dest), Box::new(versioned_beneficiary), Box::new(versioned_assets), 0)
-	verify {
-		// verify balance after transfer, decreased by transferred amount (+ maybe XCM delivery fees)
-		assert!(<pallet_balances::Pallet<T> as Inspect<_>>::balance(&caller) <= balance - transferred_amount);
-	}
```
bkontur, line 154: @Szegoo a similar check as for `reserve_transfer_assets` won't work here?

Szegoo: Yes, we cannot apply the same type of check for teleportation because the asset no longer exists on this chain (it was burned here and minted on the destination chain). It would have to be called on the destination chain.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
Szegoo: Inside `pallet-xcm` when running the benchmark for reserve transfers it will attempt to deposit the region to an account before transferring it.

This requires the `mint_into` to be implemented for regions.
 
A possible solution could be to do something like this: 
```rust
	#[cfg(features = "runtime-benchmarks")]
	fn mint_into(item: &Self::ItemId, who: &T::AccountId) -> DispatchResult {
		let region_id: RegionId = (*item).into();

		// a dummy value for benchmarking.
		let end = region_id.begin.saturating_add(1);
		Self::issue(region_id.core, region_id.begin, end, who.clone(), None);

		Ok(())
	}
```

However, this requires that the `runtime-benchmarks` feature propagates so that this implementation is used.

bkontur: @Szegoo I see, that `mint_into` seems kind of hacky :). I think there is simple solution without implementing `mint_into`. We already use this pattern for benchmarks elsewhere (I think I saw it elsewhere also), e.g. [`ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests`](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/pallets/parachain-system/src/lib.rs#L1460-L1530). For this particular benchmarks, you just need "a created region", you don't care how was created, so I would suggest to do the same here:
- create pallet function into the `pallet_broker`'s [`utility_impls.rs`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/broker/src/utility_impls.rs), which will simply just insert what is expected for success execution:
```
	pub fn create_region_for_benchmark(region_id: RegionId, ..) {
		/// OR this way
		let region = RegionRecord<AccountId, Balance> {
			end: ...,
			owner: ...,
			paid: ...,
		};
		Regions::<T>::insert(&raw_region_id, &region);

		/// OR reuse/utilize above `fn issue() -> RegionId`
		Self::issue(....)
	}      
```  
- and then in the runtime's `fn reserve_transferable_asset_and_dest` you just add:
```
   let raw_region_id = Broker::create_region_for_benchmark(...);
```

Szegoo: Yes, the problem is that `mint_into` is called from the `pallet-xcm` during benchmarking. Even if we create the region beforehand, it will still attempt to call `deposit_asset`: https://github.com/paritytech/polkadot-sdk/pull/3455/files#diff-299b81a18ba51e2c2fdaa3b6c7c7dd99ced49c9c7b8772fa1225ebecced332e0R188 (the `deposit_asset` function is calling `mint_into`). 

If not implemented it will fail and set the weight to max.

Maybe we don't have to set the weight to max if `deposit_asset` fails given that it would fail in the subsequent steps, however not sure if this is a better solution than what I proposed above.

Szegoo: I'm not sure if I understood this correctly. Are you proposing that we call `create_region_for_benchmark` before executing the benchmark code? However, in that case, we still face the issue of `deposit_asset` failing during the benchmark.

bkontur: yes, `create_region_for_benchmark` ensures that region is created correctly,
and I meant that we just use `mint_into` from [here](https://github.com/paritytech/polkadot-sdk/pull/3553/files#diff-40b92c4632f430a76d4f2c6a366582cbd2e4898fb24cd1a106a20c39872ea4d1R69-R80), there is no need to split it for `#[cfg(not(features = "runtime-benchmarks"))]` and `#[cfg(features = "runtime-benchmarks")]` as you suggested [here](https://github.com/paritytech/polkadot-sdk/pull/3455#discussion_r1541564254)

bkontur: so final solution:
1. call `Broker::create_region_for_benchmark` before benchmark in the `fn reserve_transferable_asset_and_dest`, it makes sure we have a correct region
2. `AssetTransactor::deposit_asset(` in the benchmark, ensures that region belongs to `origin_location`, that triggers transfer to destination
    - yes, we need here `mint_into` from [here](https://github.com/paritytech/polkadot-sdk/pull/3553/files#diff-40b92c4632f430a76d4f2c6a366582cbd2e4898fb24cd1a106a20c39872ea4d1R69-R80) - you can copy this `mint_into` to this branch and see that it is working

I think this is clean solution. I would not change or ignore MAX for failed `AssetTransactor::deposit_asset(`.

Szegoo: I think this will still fail:

1. call `Broker::create_region_for_benchmark` <- creates the region we will be reserve transferring
2. `AssetTransactor::deposit_asset` will fail since based on the code from #3553 we can't mint an already existing region: https://github.com/paritytech/polkadot-sdk/pull/3553/files#diff-40b92c4632f430a76d4f2c6a366582cbd2e4898fb24cd1a106a20c39872ea4d1R76

> I think this is clean solution. I would not change or ignore MAX for failed AssetTransactor::deposit_asset(.

Yeah, I agree, that isn't a good solution.

Maybe branching based on the feature flag might not actually be a bad solution

bkontur: What exactly will fail?

1. call `Broker::create_region_for_benchmark` will create region without owner (I see, that in [#3553](https://github.com/paritytech/polkadot-sdk/pull/3553/files#diff-a8dc0f7b2323a7a2b4fa8a038ab776271c156e96c0fc2f7d36c81be26889283dR87) is changed to Option):
```
let region = RegionRecord<AccountId, Balance> {
			end: ...,
			owner: None,
			paid: ...,
		};
Regions::<T>::insert(&raw_region_id, &region);
``` 

2. `AssetTransactor::deposit_asset` will call `mint_into` and that `ensure!(record.owner.is_none(), Error::<T>::NotAllowed);` will pass, because we set `owner: None`, so `mint_into` sets an owner with `origin_location` as AccountId, so it exactly does what we need.

3. `pallet_xcm::reserve_transfer_assets` executes `TransferAsset` XCM instruction which should end with calling `pallet_broker::do_transfer`, which just does some owner change.

so far so good, unless I am missing anything.

Maybe, I would suggest to add here: https://github.com/paritytech/polkadot-sdk/pull/3553 a xcm-emulator tests or xcm unit-tests for the coretime-rococo/westend to ensure that `pallet_xcm::reserve_transfer_assets` really works, because you need to probably handle some `Fungible` fees besides `NonFungible(regionId)`, because we need to handle/set `BuyExecution` on the destination, but that is not a problem, I could help, just tell me.

Szegoo: > call Broker::create_region_for_benchmark will create region without owner 

Ah, ok, I get it now. Yes, this is a good solution. I was assuming the region would be created with an owner.

Szegoo: I merged #3553 into this branch to test if it works, and it did, I managed to run the benchmark and generate weights :)

@bkontur Should we first merge #3553, or have both of these changes in this single PR?

bkontur: @Szegoo very cool, I see we don't even need `create_region_for_benchmark` and Pallet::issue` works, much better :)

> @bkontur Should we first merge https://github.com/paritytech/polkadot-sdk/pull/3553, or have both of these changes in this single PR?

I would go just with one PR

bkontur (approved):
lgtm, and when you have final PR, please run:
```
bot bench cumulus-coretime --runtime=coretime-rococo --pallet=pallet_xcm
bot bench cumulus-coretime --runtime=coretime-westend --pallet=pallet_xcm
```

Szegoo:
@seadanda, since you've already reviewed #3553, and the changes have now been moved into this PR, could you please review this one as well?

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -134,35 +122,40 @@ benchmarks! {
...
+					})?;
+			}
+		};
 
 		let recipient = [0u8; 32];
 		let versioned_dest: VersionedLocation = destination.into();
 		let versioned_beneficiary: VersionedLocation =
 			AccountId32 { network: None, id: recipient.into() }.into();
 		let versioned_assets: VersionedAssets = assets.into();
 	}: _<RuntimeOrigin<T>>(send_origin.into(), Box::new(versioned_dest), Box::new(versioned_beneficiary), Box::new(versioned_assets), 0)
```
seadanda, line 151: Is it worth still verifying the outcome of this in each case?

Szegoo: You mean placing a check here like we did for `reserve_transfer_assets`? In that case: https://github.com/paritytech/polkadot-sdk/pull/3455/files/3c714a5a239dbb6dc1182e27e335033b5d3d4080#r1541325546

cumulus/parachains/runtimes/contracts/contracts-rococo/src/xcm_config.rs
```diff
@@ -227,7 +227,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = Nothing;
```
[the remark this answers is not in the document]
Szegoo: Yeah, not really relevant. I will revert this since some assets may be represented by contracts and in that case, this could possibly be a reserve location.

substrate/frame/broker/src/lib.rs
```diff
@@ -61,7 +65,10 @@ pub mod pallet {
 	use sp_runtime::traits::{Convert, ConvertBack};
 	use sp_std::vec::Vec;
 
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
```
seadanda, line 68: Was the bump to 1 not enough?

Szegoo: Yeah, should be. Thanks for noticing this

Szegoo:
```suggestion
	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```

seadanda: No worries, I think I saw the try-runtime job flake at one stage a few days ago which I thought might have made you think you needed to bump again.

seadanda: Ah I've just seen it again - you need to add the broker migration to the runtimes' [migrations](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs#L109) (`coretime-rococo` migration CI job is failing)

Szegoo:
Would be great if we could get this merged soon :)

seadanda (approved):

substrate/frame/broker/src/lib.rs
seadanda: @Szegoo those two CI jobs are still failing, in case you didn't see this ^

Szegoo: I somehow missed that comment. Updated the migrations now.

seadanda (approved):
Thanks!

Szegoo:
Seems like we forgot to update the weights in this PR. I don't have the permission to run the bench bot so could someone run it and submit the changes in separate PR?

@seadanda @bkontur

bkontur:
> Seems like we forgot to update the weights in this PR. I don't have the permission to run the bench bot so could someone run it and submit the changes in separate PR?
> 
> @seadanda @bkontur

@Szegoo @seadanda waiting here: https://github.com/paritytech/polkadot-sdk/pull/4396#issuecomment-2097747580

Issue #3155, opened 2024-01-31
[Templates] Systematic way to synchronize the templates in this repo and `substrate-developer-hub`


kianenigma:
https://github.com/substrate-developer-hub/substrate-node-template and https://github.com/substrate-developer-hub/substrate-parachain-template are both in the `substrate-developer-hub` org, and as the commit dates suggest not very up to date. 

The existence of `substrate-developer-hub` in itself is questionable in my opinion. It is not well maintained, and I much rather everything to be in `/paritytech`, and ideally in `polkadot-sdk` repo. 

But, whatever we decide to do here, we should also do them in a backwards compatible way. `substrate-developer-hub` is used in many external resources and should be kept as an archive.

Options include: 

0. Rebuild a bot that tries to update deps in the other repos, and/or a human owner who makes sure they are kept up to date. 
1. Archive the external templates, only use the ones here. Downside is that they will not be easy to clone. 
2. Archive the old ones, make two new repos in this org, use git submodules to make sure the code remains up to date, but is clone-able separately. 

Moreover, this also gives the opportunity to re-consider our templates. I personally think we should scrap the `node-template` as it is mainly targeted towards "solo-chain development". The two main templates we want should be: 

1. A minimal template that has almost nothing, and is mainly there to help you test your custom FRAME pallet. 
2. A parachain template. 

cc @gupnik 

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/eng-automation/issues/6
```

# Outcome: 

- https://github.com/paritytech/polkadot-sdk-minimal-template
- https://github.com/paritytech/polkadot-sdk-solochain-template
- https://github.com/paritytech/polkadot-sdk-parachain-template

NukeManDan:
Related to https://github.com/paritytech/polkadot-sdk-docs/issues/67#issuecomment-1875964758 . I am an advocate to archive the who devhub org as soon as a MVP for the Rust Docs vision to replace it is in place (perhaps unarchive to update links/redirects when ready to promote the maintained version(s)).

IIRC (although I didn't deeply eval the tooling myself) the CI scripts to modify a few files and make a top level Cargo.toml was fairly simple, likely easy to update for this. One other path I would suggest in making a tool (perhaps something for `cargo-generate` or a `just` file w/ dual purposed CI scripts) that would let anyone derive a template for themselves at any commit/tag/release. CI running tool this and making a separate repo that is a template with tags to clone at a version would.

[...]

Issue #218, opened 2023-02-07
Create some compile-time notion of "runtime mode"


sam0x17:
This came out of a conversation with @kianenigma, @ggwpez and other members of the FRAME team so I'll do my best to reproduce the highlights of that conversation here. The short version is that right now in a substrate-based project, we don't really have a well-defined, standardized notion of what the build "intent" or "mode" is at runtime, nor do we have this information in an easily-consumable form at compile-time, and this is something we could improve upon. I'll explain what I mean:

### Some background:

Before I worked in Rust I was a long-time rubyist. Ruby on Rails (web framework used at a lot of startups) is known for having a very clear notion of what in the Rails world is called an "environment". When you run a Rails app, there are three default "environments" that come built-in: "test", "development" and "production". An app's environment is determined when it launches and cannot change while it is running, and this information can be accessed via globally accessible methods such as `Rails.env.test?`, `Rails.env.production?` etc., and is set by the environment variable `RAILS_ENV`.

When you run unit and integration tests, the app will automatically be in test mode (`RAILS_ENV=test`), and all sorts of configuration and things can then load and happen on an environment-specific basis (just like `cfg(test)` in Rust). It is quite common, for example, to have blocks of code that will do a switch statement that does different things for each of the defined environments.

The "development" environment is the default and is typically used for local development/tinkering. The "production" environment is used on deployed servers, and people also often add an additional "staging" environment that closely matches "production" but with subtle differences like different db credentials, etc. Secret management is also typically set up at the environment level.

Running development mode uses a lazy asset pipeline suitable for local development, while running the app in production mode will result in performance-optimized compiled assets, tighter security settings (CORS, etc), production-specific API keys and secrets, etc..

By standardizing these "environments" at the framework level, all tooling and plugins that integrate with Rails can make certain assumptions based on what environment we are in, and this is healthy for the ecosystem. Crucially, this allows gem (crate) authors to specify that certain features can or cannot be used in certain environments, and since all Rails apps are structured roughly the same way, this will _just work_ out of the box for arbitrary Rails apps.

It would be great if we could rely on something conceptually similar to this when building with substrate!

### My Suggestion

[...]

PR #3460, opened 2024-02-23
Repot all templates into a single directory

kianenigma:
The first step towards https://github.com/paritytech/polkadot-sdk/issues/3155

Brings all templates under the following structure

```
templates
|   parachain
|   |   polkadot-launch
|   |   runtime              --> parachain-template-runtime
|   |   pallets              --> pallet-parachain-template
|   |   node                 --> parachain-template-node
|   minimal
|   |   runtime              --> minimal-template-runtime
|   |   pallets              --> pallet-minimal-template
|   |   node                 --> minimal-template-node
|   solochain
|   |   runtime              --> solochain-template-runtime
|   |   pallets              --> pallet-template (the naming is not consistent here)
|   |   node                 --> solochain-template-node
```

The only note-worthy changes in this PR are: 

- More `Cargo.toml` fields are forwarded to use the one from the workspace. 
- parachain template now has weights and benchmarks
- adds a shell pallet to the minimal template
- remove a few unused deps 

A list of possible follow-ups: 

- [ ] Unify READMEs, create a parent README for all
- [ ] remove references to `docs.substrate.io` in templates
- [ ] make all templates use `#[derive_impl]`
- [ ] update and unify all licenses
- [ ] Remove polkadot launch, use https://github.com/paritytech/polkadot-sdk/blob/35349df993ea2e7c4769914ef5d199e787b23d4c/cumulus/zombienet/examples/small_network.toml instead.


templates/solochain/runtime/src/lib.rs
```diff
@@ -257,14 +253,15 @@ impl pallet_template::Config for Runtime {
 
 // Create the runtime by composing the FRAME pallets that were previously configured.
 construct_runtime!(
-	pub enum Runtime {
+	pub struct Runtime {
```
gupnik, line 256: This was done intentionally to `enum` everywhere as the syntax for instanced pallets doesn't gel with FMT otherwise. It doesn't matter here though as there are no instanced pallets, but probably still better to let it be? This would be solved with v2.

templates/solochain/pallets/template/src/benchmarking.rs
```diff
@@ -0,0 +1,35 @@
+//! Benchmarking setup for pallet-template
```
gupnik, line 1: Header missing?

templates/minimal/pallets/template/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! A shell pallet built with [`frame`].
```
gupnik, line 1: Same?

gupnik (approved):

kianenigma:
merge issues are due to https://github.com/paritytech/polkadot-sdk/pull/3002

EgorPopelyaev (approved):

templates/solochain/pallets/template/src/benchmarking.rs
kianenigma: These files didn't have a license header beforehand, not adding them here to minimize diff, but it is a follow-up.

templates/solochain/runtime/src/lib.rs
kianenigma: How is it that `enum` works but `struct` doesn't?

templates/parachain/node/src/cli.rs
skunert, line 1: Github does not let me suggest here, but these examples need to be renamed to `parachain-node`: https://github.com/paritytech/polkadot-sdk/blob/35349df993ea2e7c4769914ef5d199e787b23d4c/templates/parachain/node/src/cli.rs#L47-L58

skunert (approved):
This structure makes a lot of sense, nice!

templates/parachain/polkadot-launch/config.json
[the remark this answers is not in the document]
kianenigma: Thanks for pointing this out! This will be super useful for future documentation as well. I will list it as a follow-up.

templates/parachain/node/src/cli.rs
kianenigma: should be fixed in https://github.com/paritytech/polkadot-sdk/pull/3460/commits/26000e0ebbe964a7c81e9ee303f94b5b83ea8a86

davxy (approved):

templates/parachain/pallets/template/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
 name = "pallet-parachain-template"
-authors = ["Anonymous"]
+version = "4.0.0-dev"
```
ggwpez, line 3: If we dont publish it then it could also be `0.1.0`.

templates/parachain/runtime/src/lib.rs
```diff
@@ -489,6 +489,7 @@ impl pallet_collator_selection::Config for Runtime {
 /// Configure the pallet template in pallets/template.
 impl pallet_parachain_template::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = pallet_parachain_template::weights::SubstrateWeight<Runtime>;
```
ggwpez, line 492: Maybe put a comment here since this is using the Substrate default instead of custom benchmarked weights?  
I saw production parachains doing as well a few times already...

templates/solochain/node/Cargo.toml
```diff
@@ -0,0 +1,92 @@
...
+sp-runtime = { path = "../../../substrate/primitives/runtime" }
+sp-io = { path = "../../../substrate/primitives/io" }
+sp-timestamp = { path = "../../../substrate/primitives/timestamp" }
+sp-inherents = { path = "../../../substrate/primitives/inherents" }
+sp-keyring = { path = "../../../substrate/primitives/keyring" }
+sp-api = { path = "../../../substrate/primitives/api" }
+sp-blockchain = { path = "../../../substrate/primitives/blockchain" }
+sp-block-builder = { path = "../../../substrate/primitives/block-builder" }
+
+# frame and pallets
```
ggwpez, line 56: Did you group the dependencies or were they grouped already?  
Personally find stuff faster in an alphabetically sorted list, but not a strong opinion 🙈

templates/solochain/pallets/template/src/benchmarking.rs
ggwpez: Maybe worth adding something here for the tempaltes: https://github.com/paritytech/polkadot-sdk/blob/81156d03cec0b34c62faa853c3b38e3ba7c82ced/.github/workflows/check-licenses.yml#L40

ggwpez (approved):

michalkucharczyk (approved):

templates/solochain/runtime/src/lib.rs
joepetrowski: I don't know enough about `cargo fmt` or the config to say, but yeah when you try to instance a pallet with default imports (e.g. `BalancesTwo: pallet_balances::<Instance2> = 1`) then it tries to change it to `pallet_balances<Instance2>` within a `struct` but not within an `enum`, and it doesn't compile (which makes sense).

Also `pallet_balances::<Instance2>::{Config, Call, ..}` works just fine, only when you let it do the default imports.

kianenigma: Okay, I thought they are always equivalent.

templates/parachain/pallets/template/Cargo.toml
kianenigma: Everything in templates should be no publish and I will put all versions to `0.1.0` or `0.0.0`

templates/parachain/runtime/Cargo.toml
```diff
@@ -18,23 +19,29 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 smallvec = "1.11.0"
 
 # Local
 pallet-parachain-template = { path = "../pallets/template", default-features = false }
 
-# Substrate
+# Substrate / FRAME
 frame-benchmarking = { path = "../../../substrate/frame/benchmarking", default-features = false, optional = true }
 frame-executive = { path = "../../../substrate/frame/executive", default-features = false }
 frame-support = { path = "../../../substrate/frame/support", default-features = false }
```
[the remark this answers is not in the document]
kianenigma: Hopefully none

templates/solochain/node/Cargo.toml
ggwpez: Yea, not sure how its configured currently. CI should fail if its wrong.

templates/parachain/runtime/src/lib.rs
kianenigma: I still hope we'd had something like https://github.com/paritytech/polkadot-sdk/issues/218 and then we could ban the usage of these `SubstrateWeights` in production. 

We can still hack around it by just making them emit a warning in `no_std`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5437671

templates/parachain/runtime/src/lib.rs
ggwpez: >We can still hack around it by just making them emit a warning in no_std

Yea good idea actually.

PR #3461, opened 2024-02-23
REPLACED BY #6991 - allow ranked collective members below mininum required rank to vote but with zero weight

abdbee:
**Note, due to a small mishap that required me to rewrite this branch's history and force-push, this PR has now been replaced by #6991** 

This PR adjusts the `ranked_collective` pallet to allow ranked collective members below the minimum required rank to vote. But this vote doesn't affect the outcome of a poll because it carries zero weight (0 voting power).

Currently, a vote by members below the minimum rank results in a `RankTooLow` error. This PR removes this error and allows such members to vote successfully. This means `Tally.bare_ayes` increases by one for such members, but no changes are made to `Ayes`/`Nays`


substrate/frame/ranked-collective/src/lib.rs
```diff
@@ -625,7 +623,7 @@ pub mod pallet {
 								None => pays = Pays::No,
 							}
 							let min_rank = T::MinRankOfClass::convert(class);
-							let votes = Self::rank_to_votes(record.rank, min_rank)?;
+							let votes = Self::rank_to_votes(record.rank, min_rank).unwrap_or(0);
```
bkchr, line 626: I checked again the code and this would basically break [`support`](https://github.com/paritytech/polkadot-sdk/blob/0f4e849e0ac2de8c9880077c085985c5f656329c/substrate/frame/ranked-collective/src/lib.rs#L130).

We probably should extend `Tally` with some sort of `out_of_rank_ayes` or whatever. Then the UI can combine `bare_ayes` and `out_of_rank_ayes` to show some total number of ayes.

Maybe @ggwpez has a better idea.

abdbee: HI @bkchr , I extended `Tally` to include both `out_of_rank_ayes` and `out_of_rank_nays`. thought it'd be a good idea to include `out_of_rank_nays` too, so that it reflects both sides.

bkchr: There is another problem, as I just realized. This is based on the assumption that `VoteWeight::convert` never returns `0`. Which is not guaranteed. 

So, we probably need to change `Voting` as well to take an `Option<Votes>` or you have some better idea.

PR #3463, opened 2024-02-23
Add documentation around FRAME Offchain workers

kianenigma:
- deprecation companion: 
- pba-content companion: 
- inspired by https://substrate.stackexchange.com/questions/11058/how-can-i-create-ocw-that-wont-activates-every-block-but-will-activates-only-w/11060#11060

### DIFF
--- a/docs/sdk/src/reference_docs/frame_offchain_workers.rs
+++ b/docs/sdk/src/reference_docs/frame_offchain_workers.rs
@@ -1,0 +1,109 @@
+//! # Offchain Workers
+//!
+//! This reference document explains how offchain workers work in Substrate and FRAME. The main
+//! focus is upon FRAME's implementation of this functionality. Nonetheless, offchain workers are a
+//! Substrate-provided feature and can be used with possible alternatives to [`frame`] as well.
+//!
+//! Offchain workers are a commonly misunderstood topic, therefore we explain them bottom-up,
+//! starting at the fundamentals and then describing the developer interface.
+//!
+//! ## Context
+//!
+//! Recall from [`crate::reference_docs::wasm_meta_protocol`] that the node and the runtime
+//! communicate with one another via host functions and runtime APIs. Many of these interactions
+//! contribute to the actual state transition of the blockchain. For example [`sp_api::Core`] is the
+//! main runtime API that is called to execute new blocks.
+//!
+//! Offchain workers are in principle not different in any way: It is a runtime API exposed by the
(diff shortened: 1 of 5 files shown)


docs/sdk/src/reference_docs/frame_offchain_workers.rs
```diff
@@ -0,0 +1,115 @@
...
+//! offchain worker API.
+//!
+//! Substrate's CLI allows some degree of configuration about this, allowing node operators to
+//! specify when they want to run the offchain worker API. See
+//! [`sc_cli::RunCmd::offchain_worker_params`].
+//!
+//! ## Nondeterministic Execution
+//!
+//! Needless to say, given the above description, the code in your offchain worker API can be
+//! nondeterministic, as it is executed at unknown times, by unknown nodes, and has no impact on the
```
Szegoo, line 43:
```suggestion
//! nondeterministic, as it is not part of the blockchain's STF, so it can be executed at unknown times, by unknown nodes, and has no impact on the
```

docs/sdk/src/reference_docs/frame_offchain_workers.rs
```diff
@@ -0,0 +1,115 @@
...
+//!
+//! Some of the reasons why you might want to do the opposite, and actually embed an offchain worker
+//! API into the WASM blob are:
+//!
+//! * Accessing the state is easier within the `offchain_worker` function, as it is already a part
+//!   of the runtime, and [`frame::pallet_macros::storage`] provides all the tools needed to read
+//!   the state. Other client libraries might provide varying degrees of capability here.
+//! * It will be updated in synchrony with the runtime. A Substrate's offchain application is part
+//!   of the same WASM blob, and is therefore guaranteed to be up to date.
+//!
```
muharem, line 103: One of the notable properties, as you mentioned, is that OCW is part of the runtime API, making it more accessible. If your chain is operational, it implies that your OCW can be triggered by users. There should be use cases for this.

muharem (approved):
thank you!

docs/sdk/src/reference_docs/frame_offchain_workers.rs
kianenigma: Interestingly, this is exactly what the user in https://substrate.stackexchange.com/questions/11058/how-can-i-create-ocw-that-wont-activates-every-block-but-will-activates-only-w/11060#11060 asked for, but there is no easy way to do this yet, right? I suppose the best you can do is to link it to `state_call`, and trigger it.

PR #3468, opened 2024-02-23
[DNM] Decouple `pallet-staking` and `pallet-stake-tracker`

georgepisaltu:
This PR is a rough estimation of what implementing the changes suggested [here](https://github.com/paritytech/polkadot-sdk/pull/1933#issuecomment-1948221164) would look like.

Key ideas:
- `pallet-staking` is actively driving `pallet-stake-tracker`
- `pallet-stake-tracker` is given the necessary info in the "events" it handles, doesn't query `pallet-staking`
- `pallet-stake-tracker` doesn't question the validity of the ledgers or bonded amounts in `pallet-staking`, assumes the commands are always correct

The next step is to get rid of `pallet_stake_tracker::Pallet::<T>::should_remove_target`. That function is currently guarding the removal from `TargetList` of validators by ensuring their score in the list is `0` and they are not bonded. We can get rid of it by either:
- having a specific event that is emitted from `pallet_staking` when we are sure the conditions are met
- (*preferred*) injecting an `is_bonded: bool` argument into the event which is responsible for this removal (in my POC it seems to be only `on_validator_stake_update`)

I feel the effort is sufficiently fleshed out to get some input and decide whether it's something we want to pursue. The tests aren't fixed yet because I'd like us to settle on what the interfaces look like before fixing them.

### DIFF
--- a/substrate/frame/staking/src/ledger.rs
+++ b/substrate/frame/staking/src/ledger.rs
@@ -33,9 +33,12 @@
 
 use frame_support::{
 	defensive,
-	traits::{LockableCurrency, WithdrawReasons},
+	traits::{Currency, LockableCurrency, WithdrawReasons},
 };
-use sp_staking::{OnStakingUpdate, Stake, StakerStatus, StakingAccount, StakingInterface};
+use sp_staking::{
+	currency_to_vote::CurrencyToVote, OnStakingUpdate, Stake, StakerStatus, StakerStatusProvider,
+	StakingAccount, StakingInterface,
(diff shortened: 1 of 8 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5332742

kianenigma:
I suppose this is not relevant anymore? let's close it if so.

Issue #3202, opened 2024-02-05
Runtime API: make backing_state return multiple candidates pending availability


alindima:
This is being used by prospective-parachains when building fragment trees. Make the runtime API return an ordered collection of candidates pending availability for the given paraid.

alindima:
the interface already supports this, just need to alter the implementation

alindima:
The assumption that this returns an ordered collection of candidates is being made, for example here: https://github.com/paritytech/polkadot-sdk/blob/53f615de503c8936c2639f51b99019c64eab1018/polkadot/node/core/prospective-parachains/src/lib.rs#L354

alindima:
on a second thought, this can't be done without https://github.com/paritytech/polkadot-sdk/issues/3131

Issue #3132, opened 2024-01-30
Runtime: Enactment of candidates takes dependencies into account


eskimor:
Also dependency management, if there are dependencies not yet included - only enact once they are available as well.

Issue #3131, opened 2024-01-30
Runtime: Accept full chain for para on multiple cores (elastic scaling)


eskimor:
We accept N candidates for N scheduled cores, if those candidates form a chain back to the last included one.

sandreim:
It is not clear if this one refers to scheduler to assign multiple cores per para in same RCB

alindima:
or to accepting multiple parablocks in parachains_inherent?

eskimor:
This is about accepting multiple backed candidates in  paras_inherent, if there are multiple cores for that para.

eskimor:
Scheduler should already be fine.

alindima:
currently, the check for whether the backed candidate is a descendant of the current para head is done in: `<inclusion::Pallet<T>>::process_candidates`.

for this to work with multiple candidates, I think we'll need to change the format of the inherent data to contain the parent_head_hash, so that we're able to check the dependency chain and also check the PersistedValidationData hash against the one in the candidate receipt

eskimor:
> for this to work with multiple candidates, I think we'll need to change the format of the inherent data to contain the parent_head_hash, so that we're able to check the dependency chain and also check the PersistedValidationData hash against the one in the candidate receipt

I don't think this is necessary. It seems right now we check depedency chains via `PersistedValidationData` hash [here](https://github.com/paritytech/polkadot-sdk/blob/e1c033ebe15e6de5ee6a5a81eab2dd213f09d574/polkadot/runtime/parachains/src/inclusion/mod.rs#L1223). We should be able to make persisted validation data for each provided candidate similar to [here](https://github.com/paritytech/polkadot-sdk/blob/342d7205734b776f0b5d2b141dd730c51dfb2374/polkadot/runtime/parachains/src/util.rs#L30) and then put the candidates in a table based on their persisted validation data hashes. Then we just need to lookup the hash provided in the [descriptor](https://github.com/paritytech/polkadot-sdk/blob/a9b13f7679c122be233d282444b25e32af0a5a91/polkadot/primitives/src/v6/mod.rs#L486) to find the parent.

eskimor:
> I actually also need to change the ParachainsInherentData format due to a change of BackedCandidate primitive. Currently I am struggling to figure out how to implement this change only for Rococo runtime.

[...]

Issue #64, opened 2021-10-11
After runtime upgrade, Runtime API calls use new code with unmigrated storage


JoshOrndorff:
If the Runtime is upgraded in block N (by a `system::set_code` transaction for example), the new code is immediately written to storage. However the storage migration contained in that upgrade are not executed until the beginning of block `N + 1`. This is fine for extrinsics and other on-chain execution because there is no on-chain execution between the end of block `N` and the beginning of block `N + 1`.

However, runtime APIs can be called from off-chain contexts at any time. For example, maybe an RPC call allows users to fetch data that must be gathered from a Runtime API. Or maybe the node runs a maintenance task that calls a runtime API at a regular interval.

Any runtime API call that is made against the state associated with block `N` will use the new code with the old unmigrated storage. In general the storage schema between these two runtime versions will not match which leads to misinterpreting stored data.

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/2024-09-17-polkadot-finality-lag-slow-parachain-production-immediately-after-runtime-upgrade-post-mortem/10057/1

PR #3479, opened 2024-02-26
Elastic scaling: runtime dependency tracking and enactment

alindima:
Changes needed to implement the runtime part of elastic scaling: https://github.com/paritytech/polkadot-sdk/issues/3131, https://github.com/paritytech/polkadot-sdk/issues/3132, https://github.com/paritytech/polkadot-sdk/issues/3202

Also fixes https://github.com/paritytech/polkadot-sdk/issues/3675

TODOs:

- [x] storage migration
- [x] optimise process_candidates from O(N^2)
- [x] drop backable candidates which form cycles
- [x] fix unit tests
- [x] add more unit tests
- [x] check the runtime APIs which use the pending availability storage. We need to expose all of them, see https://github.com/paritytech/polkadot-sdk/issues/3576
- [x] optimise the candidate selection. we're currently picking randomly until we satisfy the weight limit. we need to be smart about not breaking candidate chains while being fair to all paras - https://github.com/paritytech/polkadot-sdk/pull/3573

Relies on the changes made in https://github.com/paritytech/polkadot-sdk/pull/3233 in terms of the inclusion policy and the candidate ordering


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5345809

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5348688

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5356143

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5373727

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5376896

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5379483

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1110,53 +1095,117 @@ fn limit_and_sanitize_disputes<
...
+}
+
+// Helper function for filtering candidates which don't pass the given predicate. When/if the first
+// candidate which failes the predicate is found, all the other candidates that follow are dropped.
+fn filter_candidates_with_core<
+	T: inclusion::Config + paras::Config + inclusion::Config,
+	F: FnMut(ParaId, &mut BackedCandidate<T::Hash>, CoreIndex) -> bool,
+>(
+	candidates: &mut BTreeMap<ParaId, Vec<(BackedCandidate<T::Hash>, CoreIndex)>>,
+	mut pred: F,
```
sandreim, line 1136: do we plan to reuse this function with diff predicate ? currently there is only one, so I'd vote for dropping it and renaming the function to a specialized name.

sandreim (commented):
Logic looks good at first pass, but readability can certainly be improved. 

I think we should also see if we can adjust `apply_weight_limit` candidate selection to account for elastic scaling.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -950,83 +932,86 @@ pub(crate) fn sanitize_bitfields<T: crate::inclusion::Config>(
...
-) -> SanitizedBackedCandidates<T::Hash> {
-	// Remove any candidates that were concluded invalid.
-	// This does not assume sorting.
-	backed_candidates.indexed_retain(move |candidate_idx, backed_candidate| {
-		!candidate_has_concluded_invalid_dispute_or_is_invalid(candidate_idx, backed_candidate)
+) -> BTreeMap<ParaId, Vec<(BackedCandidate<T::Hash>, CoreIndex)>> {
+	// Map the candidates to the right paraids, while making sure that the order between candidates
+	// of the same para is preserved.
+	let mut candidates_per_para: BTreeMap<ParaId, Vec<_>> = BTreeMap::new();
+	for candidate in backed_candidates {
```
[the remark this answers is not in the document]
alindima: it's being done in `filter_unchained_candidates()`. Duplicates will not form a chain unless they're cyclic

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1189,61 +1241,194 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+			})
+			.flatten()
+		{
+			Some(head_data) => Some(head_data),
+			None => <paras::Pallet<T>>::para_head(&para_id),
+		};
+		// this cannot be None
+		let latest_head_data = match maybe_latest_head_data {
+			None => {
+				log::warn!(target: LOG_TARGET, "Latest included head data for paraid {:?} is None", para_id);
```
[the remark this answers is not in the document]
alindima: yeah, as the comment above says, this cannot be None (unless there's a bug in the scheduler). it's just defensive programming

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1189,61 +1241,194 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+				continue
+			},
+			Some(latest_head_data) => latest_head_data,
+		};
+		para_latest_head_data.insert(*para_id, latest_head_data);
+	}
+
+	filter_candidates::<T, _>(candidates, |para_id, candidate| {
+		let Some(latest_head_data) = para_latest_head_data.get(&para_id) else { return false };
```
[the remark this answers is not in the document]
alindima: I think in this particular case it helps with code deduplication and I actually think it improves readability. It's useful because all the filtering done here has the particular meaning that we stop once we find a candidate which doesn't satisfy the predicate and drop all the next candidates of the paraid.
I'd much rather have this deduplicated.

The predicates are documented at the function level or when being called, with comments. I can't extract them to separate functions as they capture outer state

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: yeah, I think I can make filter_candidates more generic and remove filter_candidates_with_core

alindima: done

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5389421

polkadot/runtime/parachains/src/paras_inherent/mod.rs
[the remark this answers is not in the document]
sandreim: When reading the code I would expect to easily grasp what is being done at high level  and having the predicates separate helps with that.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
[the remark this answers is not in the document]
sandreim: Please mention that explicitly in that fn doc.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5390684

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5395740

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5403989

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5404991

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5412028

sandreim:
`candidate_pending_availability` runtime API is needed by collators. At first glance we might need to return all of them or maybe just the last one.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1189,61 +1211,205 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
-			core_index_enabled,
-		);
+	candidates: BTreeMap<ParaId, Vec<BackedCandidate<T::Hash>>>,
+) -> BTreeMap<ParaId, Vec<(BackedCandidate<T::Hash>, CoreIndex)>> {
+	let mut backed_candidates_with_core = BTreeMap::new();
 
-		let scheduled_cores = scheduled.get_mut(&backed_candidate.descriptor().para_id);
-		// Candidates without scheduled cores are silently filtered out.
+	for (para_id, backed_candidates) in candidates.into_iter() {
+		// Sanity check, should never be true.
```
ordian, line 1331: we could use defensive! macro here

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -1189,61 +1211,205 @@ fn filter_backed_statements_from_disabled_validators<T: shared::Config + schedul
...
+// In the process, duplicated candidates will also be dropped (even if they form a valid cycle;
+// cycles are not allowed if they entail backing duplicated candidates).
+fn filter_unchained_candidates<T: inclusion::Config + paras::Config + inclusion::Config>(
+	candidates: &mut BTreeMap<ParaId, Vec<BackedCandidate<T::Hash>>>,
+	allowed_relay_parents: &AllowedRelayParentsTracker<T::Hash, BlockNumberFor<T>>,
+) {
+	let mut para_latest_head_data: BTreeMap<ParaId, HeadData> = BTreeMap::new();
+	for para_id in candidates.keys() {
+		// this cannot be None
+		let latest_head_data = match <inclusion::Pallet<T>>::para_latest_head_data(&para_id) {
```
ordian, line 1245: use defensive_unwrap instead

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1098,40 +1100,52 @@ impl<T: Config> Pallet<T> {
...
-			let candidate =
-				CommittedCandidateReceipt { descriptor: pending.descriptor, commitments };
+		// This does not take elastic-scaling into account, it enacts the first candidate.
+		let enacted_candidate =
+			<PendingAvailability<T>>::mutate(&para, |candidates| match candidates {
+				Some(candidates) => candidates.pop_front(),
+				// TODO: this should also check the descendants, as they have been made available
+				// before their parent. Or just change the semantic of force_enact to enact all
+				// candidates of a para.
+				_ => None,
```
ordian, line 1110: is there an issue for this TODO?

ordian (commented):
Overall looking good, left a couple of questions. Happy to approve once this is tested/burned-in.

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1028,67 +1016,81 @@ impl<T: Config> Pallet<T> {
...
+
+		for (para_id, pending_candidates) in <PendingAvailability<T>>::iter() {
+			// We assume that pending candidates are stored in dependency order. So we need to store
+			// the earliest dropped candidate. All others that follow will get freed as well.
+			let mut earliest_dropped_idx = None;
+			for (index, candidate) in pending_candidates.iter().enumerate() {
+				if pred(candidate) {
+					earliest_dropped_idx = Some(index);
+					// Since we're looping the candidates in dependency order, we've found the
+					// earliest disputed index for this paraid.
```
eskimor, line 1077: Apparently you are assuming here to only be called for disputed candidates. That's not true, you also call this function for timed out.

In particular I would even reconsider the name for this function. What this function does is forcefully freeing cores - free cores that have _not_ become available. Which allows you to do the drain below, that would make little sense if we are freeing here based on availability. The name should reflect this, some documentation would also be good.

Maybe `cancel_cores` or `free_failed_cores`?

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1098,40 +1100,52 @@ impl<T: Config> Pallet<T> {
...
+				candidate.availability_votes,
+				candidate.core,
+				candidate.backing_group,
 			);
 		}
 	}
 
-	/// Returns the `CommittedCandidateReceipt` pending availability for the para provided, if any.
+	/// Returns the first `CommittedCandidateReceipt` pending availability for the para provided, if
+	/// any.
```
eskimor, line 1131: This does look wrong. I don't think returning just the first will be enough. We might need to do so for backwards compatibility for paras which only have one core, but we are going to need an additional call returning all.

eskimor (commented):
Not done yet, but publishing already existing remarks.

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -495,49 +503,55 @@ impl<T: Config> Pallet<T> {
...
+							for (candidate_idx, candidate) in candidates.iter_mut().enumerate() {
+								if candidate.core == core_index {
+									// defensive check - this is constructed by loading the
+									// availability bitfield record, which is always `Some` if
+									// the core is occupied - that's why we're here.
+									if let Some(mut bit) = candidate
+										.availability_votes
+										.get_mut(validator_index.0 as usize)
+									{
+										*bit = true;
```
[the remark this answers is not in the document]
sandreim: Yes, I agree, this will be very slow. While your suggestion makes sense, I'd argue that it might not be wise to allocate such large quantity of memory here. Maybe it would be better to put the commitments in separate storage.

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1028,67 +1016,81 @@ impl<T: Config> Pallet<T> {
...
+				}
 			}
-		}
 
-		for para_id in cleaned_up_ids {
-			let _ = <PendingAvailability<T>>::take(&para_id);
-			let _ = <PendingAvailabilityCommitments<T>>::take(&para_id);
+			if let Some(earliest_dropped_idx) = earliest_dropped_idx {
+				// Do cleanups and record the cleaned up cores
+				<PendingAvailability<T>>::mutate(&para_id, |record| {
```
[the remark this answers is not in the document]
sandreim: yes, indeed. I'll rework the inclusion module and figure out if we can do better on how to store the candidates pending availbility now that they can become so large.

polkadot/runtime/parachains/src/inclusion/mod.rs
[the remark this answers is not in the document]
sandreim: But, `PendingAvailabilityCommitments` is already a separate storage item, I guess in-memory should work fine.

polkadot/runtime/parachains/src/inclusion/mod.rs
alindima: I can document/comment this better. "earliest disputed index" was just an outdated comment

polkadot/runtime/parachains/src/inclusion/mod.rs
alindima: this should be tackled in this PR. I'm curious what you and other think about it. The easiest way would be to have force_enact enact all pending candidates of the para

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -605,55 +611,28 @@ impl<T: Config> Pallet<T> {
...
-
-		METRICS.on_candidates_sanitized(backed_candidates_with_core.len() as u64);
+		ensure!(count <= total_scheduled_cores, Error::<T>::UnscheduledCandidate);
 
-		// In `Enter` context (invoked during execution) there should be no backing votes from
-		// disabled validators because they should have been filtered out during inherent data
-		// preparation (`ProvideInherent` context). Abort in such cases.
-		if context == ProcessInherentDataContext::Enter {
-			ensure!(!votes_from_disabled_were_dropped, Error::<T>::BackedByDisabled);
-		}
```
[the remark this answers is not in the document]
alindima: that's right. I generally added debug logs to the filtering functions called in `sanitize_backed_candidates` whenever a candidate is filtered and the reason why it was dropped.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
alindima: 👍🏻  used defensive! macro instead

polkadot/runtime/parachains/src/inclusion/mod.rs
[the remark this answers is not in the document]
alindima: good catch. I added an intermediate btreemap from paraid to earliest_dropped_idx so that we don't run into undefined behaviour

polkadot/runtime/parachains/src/inclusion/mod.rs
alindima: done and renamed to `free_failed_cores `, have a look

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5486461

polkadot/runtime/parachains/src/inclusion/mod.rs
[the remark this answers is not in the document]
alindima: > I would suggest rewriting this into a small in-memory procedure which just translates per validator bitfields into votes per core and use that end result to write once into PendingAvailability.

I think this is a good suggestion

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -495,49 +503,55 @@ impl<T: Config> Pallet<T> {
...
+										*bit = true;
+									}
+
+									// In terms of candidate enactment, we only care if the first
+									// candidate of this para was made available. We don't enact
+									// candidates until their predecessors have been enacted.
+									if candidate_idx == 0 &&
+										candidate.availability_votes.count_ones() >= threshold
+									{
+										paras_made_available.insert(para_id);
```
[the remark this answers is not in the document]
alindima: I simplified this as much as I could. We're no longer doing this check twice `pending_availability.availability_votes.count_ones() >= threshold`

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -2127,19 +2235,44 @@ mod sanitizers {
...
+					all_backed_candidates_with_core
+						.get(&ParaId::from(2))
+						.unwrap()
+						.iter()
+						.next()
+						.unwrap()
+						.0,
+					untouched
+				);
+				assert_eq!(all_backed_candidates_with_core.get(&ParaId::from(1)), None);
```
eskimor, line 2275: We are still missing some actual elastic scaling tests here, right?

eskimor (commented):
Nice work @alindima ! Code is already much cleaner! :1st_place_medal:

polkadot/runtime/parachains/src/paras_inherent/tests.rs
sandreim: these should be much easier to implement with benchbuilder elastic scaling support like https://github.com/paritytech/polkadot-sdk/pull/3678

alindima: yes, I'm adding more tests here as we speak. Will also make use of the benchbuilder for some tests

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -547,281 +515,272 @@ impl<T: Config> Pallet<T> {
...
+									.entry(para_id)
+									.or_insert_with(|| BTreeSet::new())
+									.insert(index);
+							}
+						}
+					}
+				});
 			} else {
-				<PendingAvailability<T>>::insert(&para_id, &pending_availability);
+				// No parachain is occupying that core yet.
```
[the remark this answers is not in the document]
alindima: after some refactoring, this is no longer present

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1098,40 +1085,62 @@ impl<T: Config> Pallet<T> {
...
-		if let (Some(pending), Some(commitments)) = (pending, commitments) {
-			let candidate =
-				CommittedCandidateReceipt { descriptor: pending.descriptor, commitments };
+		// This does not take elastic-scaling into account, it enacts the first candidate.
+		let enacted_candidate =
+			<PendingAvailability<T>>::mutate(&para, |candidates| match candidates {
+				Some(candidates) => candidates.pop_front(),
+				// TODO: this should also check the descendants, as they may have been made
+				// available before their parent. Or just change the semantic of force_enact to
+				// enact all candidates of a para.
```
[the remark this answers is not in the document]
alindima: yeah, I modified force_enact to enact all cores of the para

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -495,50 +483,30 @@ impl<T: Config> Pallet<T> {
...
-					pending_availability.as_mut().and_then(|candidate_pending_availability| {
-						candidate_pending_availability.availability_votes.get_mut(validator_index)
-					}) {
-					*bit = true;
-				}
+				let core_index = CoreIndex(bit_idx as u32);
+				votes_per_core
+					.entry(core_index)
+					.or_insert_with(|| BTreeSet::new())
+					.insert(validator_index);
```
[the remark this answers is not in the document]
alindima: makes sense. will do

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -542,20 +539,25 @@ impl<T: Config> Pallet<T> {
...
 		// a core index that was freed due to a dispute.
 		//
 		// I.e. 010100 would indicate, the candidates on Core 1 and 3 would be disputed.
-		let disputed_bitfield = create_disputed_bitfield(expected_bits, freed_disputed.keys());
+		let disputed_bitfield =
+			create_disputed_bitfield(expected_bits, freed_disputed_cores.keys());
 
-		if !freed_disputed.is_empty() {
-			<scheduler::Pallet<T>>::free_cores_and_fill_claimqueue(freed_disputed.clone(), now);
+		if !freed_disputed_cores.is_empty() {
```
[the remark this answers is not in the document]
alindima: I think the reason why this was not done yet is because the disputed candidates are further being used for sanitizing the bitfields and candidates, so it's useful to have them returned separately.

However, refactored the code so that we merge the two collections into one and only call `free_cores_and_fill_claimqueue` once

polkadot/runtime/parachains/src/inclusion/migration.rs
```diff
@@ -0,0 +1,238 @@
...
+			);
+
+			weight
+		}
+
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade(_state: Vec<u8>) -> Result<(), sp_runtime::TryRuntimeError> {
+			log::trace!(target: crate::inclusion::LOG_TARGET, "Running post_upgrade() for inclusion MigrateToV1");
+			ensure!(
+				Pallet::<T>::on_chain_storage_version() >= StorageVersion::new(1),
```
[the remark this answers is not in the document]
alindima: good idea, done

polkadot/runtime/parachains/src/inclusion/mod.rs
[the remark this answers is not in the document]
alindima: done

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -515,59 +511,53 @@ impl<T: Config> Pallet<T> {
...
+							}
 						}
 					}
 
-					if let Some(stopped_at_index) = stopped_at_index {
-						let evicted_candidates = candidates.drain(0..=stopped_at_index);
+					// Trim the pending availability candidates storage and enact candidates of this
+					// para now.
+					if let Some(last_enacted_index) = last_enacted_index {
+						let evicted_candidates = candidates.drain(0..=last_enacted_index);
```
eskimor, line 540: Nice! Would be really interested in benchmark results :nerd_face:

polkadot/runtime/parachains/src/scheduler.rs
```diff
@@ -391,16 +391,6 @@ impl<T: Config> Pallet<T> {
 		});
 	}
 
-	/// Get the para (chain or thread) ID assigned to a particular core or index, if any. Core
-	/// indices out of bounds will return `None`, as will indices of unassigned cores.
-	pub(crate) fn core_para(core_index: CoreIndex) -> Option<ParaId> {
-		let cores = AvailabilityCores::<T>::get();
```
eskimor, line 397: And another performance gain. :sunglasses:

alindima:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent -v UPSTREAM_MERGE=n

alindima:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::paras_inherent -v UPSTREAM_MERGE=n

polkadot/runtime/parachains/src/paras_inherent/tests.rs
alindima: done, added more tests

alindima:
Fixed the runtime API panic caused by https://github.com/paritytech/polkadot-sdk/issues/64 and reran benchmarks for westend and rococo

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5567688

polkadot/runtime/parachains/src/runtime_api_impl/v7.rs
```diff
@@ -92,16 +93,41 @@ pub fn availability_cores<T: initializer::Config>() -> Vec<CoreState<T::Hash, Bl
 		.enumerate()
 		.map(|(i, core)| match core {
 			CoreOccupied::Paras(entry) => {
-				let pending_availability =
-					<inclusion::Pallet<T>>::pending_availability(entry.para_id())
-						.expect("Occupied core always has pending availability; qed");
-
-				let backed_in_number = *pending_availability.backed_in_number();
+				// Due to https://github.com/paritytech/polkadot-sdk/issues/64, using the new storage types would cause
```
sandreim, line 96: Can we extract this into a separate function that does the conversion ?

alindima: we could to that, but I'd rather keep all the logic for this storage migration hack in the same place as it's easier to comprehend and remove in the future IMO

sandreim: Yeah, but it makes this a longer lump of code which is not really ideal for readability, but to not forget to remove it later, we should add an issue to the board.

alindima: don't think we need an issue to track removal. When the migration will be removed, this bit of code will fail to compile, which should notify us to remove it

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -145,6 +183,122 @@ mod enter {
...
+	// ParaId 3 has three pending candidates on cores 2, 3 and 4.
+	// All of them are being made available in this block. Propose 5 more candidates (one for each
+	// core) and check that they're successfully backed and the old ones enacted.
+	fn include_backed_candidates_elastic_scaling() {
+		let config = default_config();
+		assert!(config.configuration.config.scheduler_params.lookahead > 0);
+
+		new_test_ext(config).execute_with(|| {
+			// Set the elastic scaling MVP feature.
+			<configuration::Pallet<Test>>::set_node_feature(
```
eskimor, line 218: Isn't this set by default already?

polkadot/runtime/rococo/src/weights/runtime_parachains_paras_inherent.rs
```diff
@@ -13,161 +13,322 @@
...
-		Weight::from_parts(352_590_000 as u64, 0)
-			// Standard Error: 13_000
-			.saturating_add(Weight::from_parts(49_254_000 as u64, 0).saturating_mul(v as u64))
-			.saturating_add(T::DbWeight::get().reads(24 as u64))
-			.saturating_add(T::DbWeight::get().writes(16 as u64))
+		// Proof Size summary in bytes:
+		//  Measured:  `67785`
+		//  Estimated: `73725 + v * (23 ±0)`
+		// Minimum execution time: 949_716_000 picoseconds.
+		Weight::from_parts(482_361_515, 0)
```
eskimor, line 115: Interesting. Disputes became more expensive - any idea why?

polkadot/runtime/rococo/src/weights/runtime_parachains_paras_inherent.rs
```diff
@@ -13,161 +13,322 @@
...
+	/// Proof: `Paras::MostRecentContext` (`max_values`: None, `max_size`: None, mode: `Measured`)
 	fn enter_bitfields() -> Weight {
-		Weight::from_parts(299_878_000 as u64, 0)
-			.saturating_add(T::DbWeight::get().reads(21 as u64))
-			.saturating_add(T::DbWeight::get().writes(15 as u64))
+		// Proof Size summary in bytes:
+		//  Measured:  `42757`
+		//  Estimated: `48697`
+		// Minimum execution time: 437_627_000 picoseconds.
+		Weight::from_parts(460_975_000, 0)
```
eskimor, line 182: Wow. Bitfields even more so.

polkadot/runtime/rococo/src/weights/runtime_parachains_paras_inherent.rs
```diff
@@ -13,161 +13,322 @@
...
+	/// Proof: `Paras::UpgradeGoAheadSignal` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	/// Storage: `Paras::MostRecentContext` (r:0 w:1)
+	/// Proof: `Paras::MostRecentContext` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	/// The range of component `v` is `[101, 200]`.
+	fn enter_backed_candidates_variable(v: u32, ) -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `42829`
+		//  Estimated: `48769`
+		// Minimum execution time: 1_305_254_000 picoseconds.
+		Weight::from_parts(1_347_160_667, 0)
```
eskimor, line 253: :scream:

polkadot/runtime/parachains/src/runtime_api_impl/v7.rs
```diff
@@ -92,18 +93,41 @@ pub fn availability_cores<T: initializer::Config>() -> Vec<CoreState<T::Hash, Bl
...
+						.expect("Occupied core always has pending availability; qed")
+				} else {
+					let candidate = <inclusion::Pallet<T>>::pending_availability_with_core(
+						entry.para_id(),
+						CoreIndex(i as u32),
+					)
+					.expect("Occupied core always has pending availability; qed");
+
+					// Translate to the old candidate format, as we don't need the commitments now.
+					inclusion::migration::v0::CandidatePendingAvailability {
```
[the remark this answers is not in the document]
alindima: when the migration code gets deleted we can safely delete this code too. This code is only needed during the migration

polkadot/runtime/parachains/src/paras_inherent/tests.rs
alindima: not in unit tests. it's being set here: https://github.com/paritytech/polkadot-sdk/blob/5fd72a1f5e7f4db02b4962fb146e4b8d154a485c/polkadot/node/service/src/chain_spec.rs#L161

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -239,7 +232,7 @@ pub mod pallet {
 	/// the given `freed_concluded`).
 	///
 	/// The parameter `freed_concluded` contains all core indicies that became
-	/// free due to candidate that became available.
+	/// free due to candidates that became available or due to candidates being disputed.
```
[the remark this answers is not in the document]
alindima: agree, I'll remove this function and move this bit of code to the higher level

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -145,6 +183,122 @@ mod enter {
...
+		});
+	}
+
+	#[test]
+	// ParaId 1 has one pending candidate on core 0.
+	// ParaId 2 has one pending candidate on core 1.
+	// ParaId 3 has three pending candidates on cores 2, 3 and 4.
+	// All of them are being made available in this block. Propose 5 more candidates (one for each
+	// core) and check that they're successfully backed and the old ones enacted.
+	fn include_backed_candidates_elastic_scaling() {
```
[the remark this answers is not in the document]
alindima: fair, I'll add that

alindima:
yeah, the weights for rococo are way off when comparing to the previous values. I'm betting that's because they were last updated in 2021. The difference for westend shows that there isn't a significant change

polkadot/runtime/parachains/src/runtime_api_impl/v7.rs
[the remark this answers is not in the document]
eskimor: Not sure this is true as this would break the API ... no? Also why would we want to return megabytes of data here, if it is not needed?

polkadot/runtime/parachains/src/paras_inherent/tests.rs
eskimor: Maybe it should.

alindima: If we add this, I'd have to fix a bunch of tests which make the assumption that it's not enabled by default. I wouldn't do that now

polkadot/runtime/parachains/src/paras_inherent/mod.rs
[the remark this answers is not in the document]
alindima: done

polkadot/runtime/parachains/src/paras_inherent/tests.rs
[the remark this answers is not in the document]
alindima: added, have a look

polkadot/runtime/parachains/src/runtime_api_impl/v7.rs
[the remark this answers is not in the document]
eskimor: Ok. All good then, can't read.

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -300,6 +304,117 @@ mod enter {
...
 			);
 		});
+
+		// ParaId 0 has one pending candidate on core 0.
+		// ParaId 1 has one pending candidate on core 1.
+		// ParaId 2 has 4 pending candidates on cores 2, 3, 4 and 5.
+		// Cores 1, 2 and 3 are being made available in this block. Propose 6 more candidates (one
+		// for each core) and check that the right ones are successfully backed and the old ones
+		// enacted.
+		let config = default_config();
```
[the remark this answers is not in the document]
alindima: to avoid long and dubious test names :D that's arguably a bad reason but I didn't think too much about it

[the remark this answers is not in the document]
eskimor: fair.

polkadot/runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -300,6 +304,117 @@ mod enter {
...
+					.collect::<Vec<_>>(),
+				vec![CoreIndex(1)]
+			);
+			assert_eq!(
+				inclusion::PendingAvailability::<Test>::get(ParaId::from(2))
+					.unwrap()
+					.into_iter()
+					.map(|c| c.core_occupied())
+					.collect::<Vec<_>>(),
+				vec![CoreIndex(4), CoreIndex(5), CoreIndex(2), CoreIndex(3)]
```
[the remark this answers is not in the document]
alindima: done

Issue #3207, opened 2024-02-05
[RPC-Spec-V2] Allow chainHead operations to be made from a single connection


lexnv:
The [chainHead-cli](https://github.com/lexnv/chainhead-cli/blob/master/README.md) tool proves that we can have:
- one connection to the `chainHead_folow` 
- a separate connection (on a separate process, on a different socket) that calls `chainHead_storage` with the subscription ID of the first connection; and produces the result in the first connection

To restrict this case, ensure that the subscription ID provided by the `chainHead_follow` is valid within the context of that connection only.

### Repro case
This issue can be reproduced if the user makes 2 separate connections to the same RPC server.

For the common scenario, the user would establish 2 connections via `rpc-polkadot.io`, which would most likely generate 2 connections on different servers.
However, if the user knows the IP address of the server (or targets a local-development node for testing) this could be reproduced.

cc @paritytech/subxt-team

jsdw:
> To restrict this case, ensure that the subscription ID provided by the chainHead_follow is valid within the context of that connection only.

Sounds good to me; I'd previously assumed that subscription IDs were scoped to the connection already, and it's def safer to ensure that they are :)

> For the production environment use-case, this requires the user to make 2 separate connections to the same RPC server, which is unlikely due to our reverse-proxies. However, this is a valid use-case for testing environments.

I don't understand this sentence really?

lexnv:
For this case to happen, the user needs 2 separate connections to the same server.

From the user perspective, this is unlikely to happen for our production nodes. For example, connecting 2 times to `rpc-polkadot.io` would most probably generate 2 connections on different nodes. 

However, if the user targets a localhost node / development node, or a node with a known IP address this behavior could be reproduced. Indeed the misleading of my issue is `testing environments` which is not entirely true; since the target node could still be in production :D will change the description a bit, thanks for pointing this out! 🙏

PR #3343, opened 2024-02-15
rpc-servers: Allow chainHead methods to be called from a single connection

lexnv:
This PR adds middleware for the JSON rpc servers to ensure that the chainHead methods are called from a single connection.

The middleware intercepts the subscription IDs generated by the `chainHead_follow`.
Any further chainHead function call (ie `chainHead_storage`) must provide a subscription ID that was generated by the middleware's connection.
A middleware per connection is created in this context.

Closes: https://github.com/paritytech/polkadot-sdk/issues/3207

cc @paritytech/subxt-team

PR #3481, opened 2024-02-26
chainHead: Allow methods to be called from within a single connection context and limit connections

lexnv:
This PR ensures that the chainHead RPC class can be called only from within the same connection context.

The chainHead methods are now registered as raw methods. 
- https://github.com/paritytech/jsonrpsee/pull/1297
The concept of raw methods is introduced in jsonrpsee, which is an async method that exposes the connection ID:
The raw method doesn't have the concept of a blocking method. Previously blocking methods are now spawning a blocking task to handle their blocking (ie DB) access. We spawn the same number of tasks as before, however we do that explicitly.

Another approach would be implementing a RPC middleware that captures and decodes the method parameters:
- https://github.com/paritytech/polkadot-sdk/pull/3343
However, that approach is prone to errors since the methods are hardcoded by name. Performace is affected by the double deserialization that needs to happen to extract the subscription ID we'd like to limit. Once from the middleware, and once from the methods itself.

This PR paves the way to implement the chainHead connection limiter:
- https://github.com/paritytech/polkadot-sdk/issues/1505
Registering tokens (subscription ID / operation ID) on the `RpcConnections` could be extended to return an error when the maximum number of operations is reached.

While at it, have added an integration-test to ensure that chainHead methods can be called from within the same connection context.

Before this is merged, a new JsonRPC release should be made to expose the `raw-methods`:
- [x] Use jsonrpsee from crates io (blocked by: https://github.com/paritytech/jsonrpsee/pull/1297)

Closes: https://github.com/paritytech/polkadot-sdk/issues/3207

cc @paritytech/subxt-team


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5348068

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -510,3 +569,26 @@ fn method_started_response(
...
+{
+	let (tx, rx) = oneshot::channel();
+
+	let blocking_fut = async move {
+		let result = fut.await;
+		// Send the result back on the channel.
+		let _ = tx.send(result);
+	};
+
+	executor.spawn_blocking("substrate-rpc-subscription", Some("rpc"), blocking_fut.boxed());
```
niklasad1, line 591: yeah right, `spawn_blocking` doesn't return a handle for the future that's why you need the extra oneshot channel

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,60 @@
...
+use parking_lot::Mutex;
+use std::{
+	collections::{HashMap, HashSet},
+	sync::Arc,
+};
+
+/// Limit the RPC functionality to a single connection.
+#[derive(Default, Clone)]
+pub struct RpcConnections {
+	data: Arc<Mutex<HashMap<ConnectionId, HashSet<String>>>>,
```
niklasad1, line 29: could also be `Arc<Mutex<HashSet<(ConnectionId, SubscriptionId)>>>`  I guess but perhaps you have plan for it to more generic

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5547011

substrate/client/rpc-spec-v2/src/common/connections.rs
jsdw: mmm since `ConnectionId` is just a small u64, I'd just combine to `Arc<Mutex<HashSet<(ConnectionId, SubscriptionId)>>>` and that hsould be more efficient and probably small or no real extra space cost :)

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -209,83 +215,110 @@ where
...
-				// Block is not part of the subscription.
-				return ResponsePayload::error(ChainHeadRpcError::InvalidBlock);
-			},
-			Err(_) => return ResponsePayload::error(ChainHeadRpcError::InvalidBlock),
-		};
+		if !self
+			.rpc_connections
+			.contains_token(connection_details.id(), &follow_subscription)
+		{
+			return ResponsePayload::success(MethodResponse::LimitReached);
```
jsdw, line 235: IS this the right error to return? If this `if` is true then it means a connection is trying to call `chainHead_body` but hasn't called `chainHead_follo` yet I think?

lexnv: Yep that makes sense! 

When this check is true, the current connection hasn't obtained the `follow_subscription` ID by making a call to the `chainHead_follow` first. This could mean that the user is guessing the follow subscription of another user, or that the follow subscription was leaked from another valid subscription.


> (I also can't offhand see where we do anything to limit the number of chainHead_follow subscriptions, which I assumed might exist for us to emit this error, but maybe I'm either missing something or it's not present yet :))

I'll add it in this PR since it shouldn't be that complicated, initially wanted to do a followup for this :D

substrate/client/rpc-spec-v2/Cargo.toml
```diff
@@ -44,6 +44,7 @@ futures-util = { version = "0.3.30", default-features = false }
 rand = "0.8.5"
 
 [dev-dependencies]
+jsonrpsee = { version = "0.22", features = ["client-core", "macros", "client", "server"] }
```
niklasad1, line 47: "client", will add lots of dependencies AFAIU "ws-client" and "server" should be sufficient

```suggestion
jsonrpsee = { version = "0.22", features = ["ws-client", "server"] }
```

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -110,6 +110,8 @@ pub struct ChainHead<BE: Backend<Block>, Block: BlockT, Client> {
 	/// The maximum number of items reported by the `chainHead_storage` before
 	/// pagination is required.
 	operation_max_storage_items: usize,
+	/// Limit the RPC functionality to a single connection.
```
niklasad1, line 113: I would re-phrase this, I think you are trying to say that each connection has its own state and not that the RPC functionality is limited to a single connection? 

This connection state is then used to determine whether a certain call is permitted on a particular connection such as "unstable_storage calls" or something like that...

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -296,19 +329,33 @@ where
...
 		follow_subscription: String,
 		hash: Block::Hash,
 		items: Vec<StorageQuery<String>>,
 		child_trie: Option<String>,
 	) -> ResponsePayload<'static, MethodResponse> {
+		if !self
+			.rpc_connections
+			.contains_token(connection_details.id(), &follow_subscription)
+		{
+			return ResponsePayload::success(MethodResponse::LimitReached);
```
niklasad1, line 356: > This return value indicates the call couldn't be started because the server is overloaded, or that the followSubscription is invalid or stale.

I think it's quite strange to use `LimitReached` here but according to the spec...

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -357,25 +404,26 @@ where
...
 		items.truncate(num_operations);
 
-		let (rp, rp_is_success) = method_started_response(operation_id, Some(discarded));
-
+		let (rp, rp_fut) = method_started_response(operation_id, Some(discarded));
 		let fut = async move {
-			// Events should only by generated
-			// if the response was successfully propagated.
-			if rp_is_success.await.is_err() {
+			// This leaves on a seperate task, because the `rp` needs to be acknowledged by the
```
niklasad1, line 409: Wait for the server to send out the response and if it produces an error no event should be generated

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -408,44 +463,53 @@ where
...
-			});
-
-		let (rp, rp_fut) = method_started_response(operation_id, None);
+		let client = self.client.clone();
 
+		let (rp, rp_fut) = method_started_response(operation_id.clone(), None);
 		let fut = async move {
-			// Events should only by generated
-			// if the response was successfully propagated.
+			let event = client
```
niklasad1, line 470: move this below `rp_fut.await`? 

Makes not sense to me to do this storage call if  `rp_fut.await` produces an error...

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -3297,3 +3330,120 @@ async fn storage_closest_merkle_value() {
...
+		.await
+		.unwrap();
+	// Calls cannot be made from a different connection context.
+	let second_client = jsonrpsee::ws_client::WsClientBuilder::default()
+		.build(&server_url)
+		.await
+		.unwrap();
+
+	let mut sub: RpcClientSubscription<FollowEvent<String>> = client
+		.subscribe("chainHead_unstable_follow", rpc_params![true], "chainHead_unstable_unfollow")
```
niklasad1, line 3349: pretty sure that you could use the type API that is generated by the jsonrpsee proc macro here when you are using a "real ws client" (not sure about name though)

```rust
client.chainhead_unstable_follow(true)
```

lexnv: I completely forgot about those! I've added them since they improve a lot the code readability, thanks!

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
lexnv: Yep, that makes sense! Thanks for catching this! 🙏

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
lexnv: Yep I do like that phrasing better, thanks!

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
lexnv: I've rephrased the documentation a bit, let me know if it makes more sense now 🙏

lexnv:
I've ended up with a slightly more generic approach to handle connections.

With this revision, we are able to:
- Allow methods to be called from as single context (chainHead methods cannot be called with a leaked subscription ID from a different connection)
- Limit the number of active `chainHead_follow` subscription (code can easily be reutilzied for transaction etc

The internal structure that handles the connections looks like:

```rust
data: Arc<Mutex<HashMap<ConnectionId, ConnectionData>>>

// Where
ConnectionData { num_tokens: usize, tokens: HashSet<String> }
```

This is because we'd need to know how many `connectionId`s are registered at any time.
The `num_tokens` counts:
- (1) tokens that are added to `tokens` (active tokens already registered)
- (2) tokens that will soon be added after the PendingSubscription is accepted

More on (2):
- We'd need to return a jsonrpsee error object when reaching the limits.
Extracted from [spec](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/chainHead_unstable_follow.md#possible-errors):
  > A JSON-RPC error with error code -32800 can be generated if the JSON-RPC client has already opened 2 or more chainHead_unstable_follow subscriptions.
- A `PendingSubscriptionSink` does not expose the subscription ID before being accepted (and nor should it). Therefore, this is used to "reserve in advance" some capacity. 
 - I've added an easier to use interface to abstract these concepts
 ```rust
   let Some(mut reserved: ReservedConnection) = RpcConnections.reserve_token(conenction_id) else { error };
   
   // After accepting the sink, we gain access to the subscription ID. 
   sink.accept();
   
   // Keep track of the subscription ID.
   reserved.register(follow_subscription);
   
   // RAII unregisters / unreserves on drop.
   ```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5572190

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5572155

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -180,12 +192,20 @@ where
 		let subscriptions = self.subscriptions.clone();
 		let backend = self.backend.clone();
 		let client = self.client.clone();
+		let rpc_connections = self.rpc_connections.clone();
 
 		let fut = async move {
+			// Ensure the current connection ID has enough space to accept a new subscription.
+			let connection_id = pending.connection_id();
+
+			let Some(reserved_token) = rpc_connections.reserve_token(connection_id) else {
```
niklasad1, line 201: ok, you must return a JSON-RPC error if the limit is exceeded and that's why you look up the state in the hashmap before "subscription accept is called"?

Too bad to lock the mutex twice for that...  but no way around that I guess

niklasad1: Can we merge `subscriptions` and `connections`? The subscription ID is stored in both and would be nice to merge them is possible

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -198,6 +218,8 @@ where
 			};
 			debug!(target: LOG_TARGET, "[follow][id={:?}] Subscription accepted", sub_id);
 
+			let _registered_token = reserved_token.register(sub_id.clone());
```
niklasad1, line 221: doesn't return anything?!

```suggestion
			reserved_token.register(sub_id.clone());
```

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,228 @@
...
+		let mut data = self.data.lock();
+
+		let entry = data.entry(connection_id).or_insert_with(ConnectionData::default);
+		entry.num_tokens = entry.num_tokens.saturating_sub(1);
+	}
+
+	/// Register a token for the given connection.
+	///
+	/// Users should call [`Self::reserve_token`] before calling this method.
+	fn register_token(&self, connection_id: ConnectionId, token: String) {
```
niklasad1, line 84: this should return RegisteredToken?

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
niklasad1, line 221: RAII, when it dropped it's removed from connections

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,228 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use jsonrpsee::ConnectionId;
+use parking_lot::Mutex;
+use std::{
+	collections::{HashMap, HashSet},
+	sync::Arc,
+};
+
+/// Limit the RPC functionality to a single connection.
```
niklasad1, line 26:
```suggestion
/// Connection state which keeps track whether a connection exist and
/// the number of concurrent operations.
```

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,228 @@
...
+	data: Arc<Mutex<HashMap<ConnectionId, ConnectionData>>>,
+}
+
+#[derive(Default)]
+struct ConnectionData {
+	/// The total number of tokens.
+	///
+	/// # Note
+	///
+	/// Because a pending subscription sink does not expose the future subscription ID,
```
niklasad1, line 41: I wonder if we can make it explicit that token == subscription ID?

It may be hard to understand this comment about the subscription sink otherwise

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,228 @@
...
+	fn register_token(&self, connection_id: ConnectionId, token: String) {
+		let mut data = self.data.lock();
+
+		let entry = data.entry(connection_id).or_insert_with(ConnectionData::default);
+		// Should be already checked `Self::reserve_token`.
+		if entry.tokens.len() >= self.capacity {
+			return;
+		}
+
+		entry.tokens.insert(token);
```
niklasad1, line 93: we assume that subscription ID is unique so should be unlikely for duplicate subscription IDs here

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl Drop for RegisteredConnectionToken {
+	fn drop(&mut self) {
+		self.rpc_connections.unregister_token(self.connection_id, &self.token);
+	}
+}
+
+#[cfg(test)]
+mod tests {
```
niklasad1, line 160: Would useful with another test "is_empty" that connection IDs are cleaned from the HashMaps etc.

The code looks go to me otherwise

niklasad1 (approved):

substrate/client/rpc-spec-v2/src/common/connections.rs
lexnv: Indeed the `token` name is a bit misleading. This is the subscription ID in this case, however it will be the transaction `OperatationId` for the tx API 🤔 

Maybe `identifier` sounds better? 🤔

substrate/client/rpc-spec-v2/src/common/connections.rs
lexnv: Added a few extra checks to ensure we clean-up everything now, thanks!

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
jsdw: One thing I didn't realise when I saw this first is that the spec actually says to return `limitReached` when the followSubscription is invalid/stale, so this all maekssense to me now (a nit here might help make clear why we return an error that feels unrelaetd at a glance eg "the spec says to return LimitReached if followSubscription is invalid or stale" :))

substrate/client/rpc-spec-v2/src/chain_head/subscription/mod.rs
```diff
@@ -43,30 +64,55 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionManagement<Block, BE> {
...
+	/// Create a new instance from the inner state.
 	///
-	/// If the subscription was not previously inserted, returns the receiver that is
-	/// triggered upon the "Stop" event. Otherwise, if the subscription ID was already
-	/// inserted returns none.
-	pub fn insert_subscription(
+	/// # Note
+	///
+	/// Used for testing.
+	#[doc(hidden)]
```
jsdw, line 86: If only used for testing, you could have `#[cfg(test)]` here instead to guarantee that (and make it super obvious) :)

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -3297,3 +3350,176 @@ async fn storage_closest_merkle_value() {
...
+	};
+
+	// Cannot make a call from a different connection context.
+	let _response = ChainHeadApiClient::<String>::chain_head_unstable_unpin(
+		&second_client,
+		first_sub_id.clone(),
+		crate::chain_head::api::ListOrValue::Value(finalized_hash.clone()),
+	)
+	.await
+	.unwrap();
```
jsdw, line 3398: This code produces no error I guess (we just unwrap it), so what do we learn from it?

jsdw: Ah I see; we are unpinning the block from the wrong subscription and that hsould have no effect!

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -3297,3 +3350,176 @@ async fn storage_closest_merkle_value() {
...
+
+	let first_sub_id = match sub.kind() {
+		jsonrpsee::core::client::SubscriptionKind::Subscription(id) => match id {
+			jsonrpsee::types::SubscriptionId::Num(num) => num.to_string(),
+			jsonrpsee::types::SubscriptionId::Str(s) => s.to_string(),
+		},
+		_ => panic!("Unexpected subscription ID"),
+	};
+
+	// Cannot make a call from a different connection context.
```
jsdw, line 3391: Maybe tweak the comment to "trying to unpin from a different conneciton will have no effect"?

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,258 @@
...
+	num_identifiers: usize,
+	/// Active registered identifiers for the given connection.
+	///
+	/// # Note
+	///
+	/// For chainHead, this represents the subscription ID.
+	/// For transactionBroadcast, this represents the operation ID.
+	/// For transaction, this is empty and the number of active calls is tracked by
+	/// [`Self::num_identifiers`].
+	identifiers: HashSet<String>,
```
jsdw, line 64: Am I right to assume that we are not worries about subscription IDs and operation IDs colliding at all in our implementation here?

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,258 @@
...
+	/// Gives back the reserved space before the connection identifier is registered.
+	///
+	/// # Note
+	///
+	/// This may happen if the pending subscription cannot be accepted (unlikely).
+	fn unreserve_space(&self, connection_id: ConnectionId) {
+		let mut data = self.data.lock();
+
+		let entry = data.entry(connection_id).or_insert_with(ConnectionData::default);
+		entry.num_identifiers = entry.num_identifiers.saturating_sub(1);
```
jsdw, line 98: Do we ever clear out entries completely when the number goes to 0?

substrate/client/rpc-spec-v2/src/common/connections.rs
```diff
@@ -0,0 +1,258 @@
...
+
+	/// Unregister an identifier for the given connection.
+	fn unregister_identifier(&self, connection_id: ConnectionId, identifier: &str) {
+		let mut data = self.data.lock();
+		if let Some(connection_data) = data.get_mut(&connection_id) {
+			connection_data.identifiers.remove(identifier);
+			connection_data.num_identifiers = connection_data.num_identifiers.saturating_sub(1);
+
+			if connection_data.num_identifiers == 0 {
+				data.remove(&connection_id);
```
jsdw, line 128: Ah, I guess this is whewre we clean up :)

jsdw (approved):
Looks good to me; only very small comments in the end :)

substrate/client/rpc-spec-v2/src/common/connections.rs
lexnv: Yep, IIRC, we'll either use this for chainHead with subscriptions ID only; or for the tx with operation ID only

Issue #3413, opened 2024-02-20
Broker pallet should expose `sale_price` through runtime api


Szegoo:
Frontends should not be required to reimplement the logic for determining the sale price of cores, as this approach is unreliable and changes in pricing logic could make things break. 
Instead, it should be possible to directly query the price from the runtime.

bkchr:
Generally we can add it, but it isn't really required to have this declared centrally. The runtime api can just be declared and implemented next to the runtime.

PR #3485, opened 2024-02-26
Broker: sale price runtime api

Szegoo:
Defines a runtime api for `pallet-broker` for getting the current price of a core if there is an ongoing sale.

Closes: #3413


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5350673

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5350708

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5350707

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5351017

substrate/frame/broker/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Runtime API definition for the FRAME Broker pallet.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::Codec;
+
+sp_api::decl_runtime_apis! {
```
bkchr, line 24: We don't need an extra crate for this runtime api. An extra crate would only be required if we need to import this in the node side, which isn't the case here.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
```diff
@@ -595,6 +595,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl pallet_broker_runtime_api::BrokerApi<Block, Balance> for Runtime {
+		fn sale_price() -> Option<Balance> {
```
bkchr, line 599: And do we maybe want to expose more stuff?

Szegoo: It would be useful to fetch the unclaimed revenue of a region contributed to the instantaneous pool. I can add that as well.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5388634

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5388905

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
Szegoo: > It would be useful to fetch the unclaimed revenue of a region contributed to the instantaneous pool.

After thinking about this a bit more not sure if adding this would make sense. Getting the unclaimed revenue can require quite a lot of reading(reading `InstaPoolHistory` per each timeslice). So having this exposed could be an easy spam vector.

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -437,4 +437,17 @@ impl<T: Config> Pallet<T> {
...
+	pub fn api_sale_price() -> Option<BalanceOf<T>> {
+		let status = Status::<T>::get()?;
+		let sale = SaleInfo::<T>::get()?;
+
+		if sale.first_core < status.core_count || sale.cores_sold < sale.cores_offered {
+			return None;
+		}
+		let now = frame_system::Pallet::<T>::block_number();
+		let price = Self::sale_price(&sale, now);
+		Some(price)
```
bkchr, line 451:
```suggestion
		Some(Self::sale_price(&sale, now))
```

bkchr (approved):

bkchr:
@Szegoo can you please add a prdoc Runtime Devs and Runtime Users

prdoc/pr_3485.prdoc
```diff
@@ -0,0 +1,18 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: "Broker: sale price runtime api"
+
+doc:
+  - audience: 
+    - Runtime Dev 
+    - Runtime User
```
Szegoo, line 9: @bkchr Are multiple audiences supported? Based on this I would assume not: https://github.com/paritytech/polkadot-sdk/blob/463ccb8f9028ebf5f7fc53d7e835c21f43172253/prdoc/schema_user.json#L101

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -437,4 +437,16 @@ impl<T: Config> Pallet<T> {
 		Self::deposit_event(Event::AllowedRenewalDropped { core, when });
 		Ok(())
 	}
+
+	/// If there is a sale ongoing returns the current price of a core.
+	pub fn api_sale_price() -> Option<BalanceOf<T>> {
+		let status = Status::<T>::get()?;
+		let sale = SaleInfo::<T>::get()?;
+
+		if sale.first_core < status.core_count || sale.cores_sold < sale.cores_offered {
```
eskimor, line 446: Can you explain the second part of the `or`, why don't we have a sale price if cores_sold < cores_offered? Also:

1. This looks like code duplication that could get out of sync. If we don't have a price for some reason, this should not be determined here, but e.g. in something like `sale_price`.
2. Wouldn't a reason for not having a sale price be of interest? E.g. would it make sense to make the return type more expressive?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5639528

substrate/frame/broker/src/dispatchable_impls.rs
Szegoo: > Can you explain the second part of the or

Yes, this was incorrect the comparison sign needs to be flipped for actually both parts of the `or`.

> this should not be determined here, but e.g. in something like sale_price.

I would rather leave `sale_price` as a low level function without much contextual information and checks. I extracted the checks in a new function.

> Wouldn't a reason for not having a sale price be of interest? E.g. would it make sense to make the return type more expressive?

We can return more information by returning an error instead of `None`. I didn't do it since it is a bit weird to return a `DispatchError` from a runtime api function, but it might be worth it for more expressiveness.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5639991

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -437,4 +436,25 @@ impl<T: Config> Pallet<T> {
 		Self::deposit_event(Event::AllowedRenewalDropped { core, when });
 		Ok(())
 	}
+
+	pub(crate) fn ensure_core_availability(
+		status: &StatusRecord,
+		sale: &SaleInfoRecordOf<T>,
+	) -> Result<(), DispatchError> {
+		ensure!(sale.first_core < status.core_count, Error::<T>::Unavailable);
+		ensure!(sale.cores_sold < sale.cores_offered, Error::<T>::SoldOut);
```
joepetrowski, line 445: What is the motivation to split this into another function?

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -437,4 +436,25 @@ impl<T: Config> Pallet<T> {
...
+
+		Ok(())
+	}
+
+	/// If there is an ongoing sale returns the current price of a core.
+	pub fn api_sale_price() -> Result<BalanceOf<T>, DispatchError> {
```
joepetrowski, line 451: Why not just `sale_price`? I don't think we normally prefix calls with `api_`, the fact that it's available means it's in the API :).

Szegoo: We already have a function called `sale_price`.

https://github.com/paritytech/polkadot-sdk/blob/cc1e6ac301ea88e3cb3253a84e4c6aa28f2d8f87/substrate/frame/nomination-pools/src/lib.rs#L3566

substrate/frame/broker/src/dispatchable_impls.rs
Szegoo: To avoid code duplication: https://github.com/paritytech/polkadot-sdk/pull/3485#discussion_r1537393976

But, yeah, I don't have a strong opinion here since it's only two lines anyway.

joepetrowski: Doh, I read that comment and completely missed it. I can see both sides, if something else wants to use it then it makes sense to split it out.

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -455,4 +455,25 @@ impl<T: Config> Pallet<T> {
 
 		Ok(())
 	}
+  
+  pub(crate) fn ensure_core_availability(
```
bkchr, line 459: Can you please add some comment? I don't really get from reading the name what this does, because you don't really pass any core. Or maybe call it `ensure_cores_for_sale`.

Szegoo: I think `ensure_cores_for_sale` is good.

bkchr:
Now if someone of the gentleman @joepetrowski @eskimor or @seadanda could approve this pr, we could merge it.

prdoc/pr_3485.prdoc
```diff
@@ -0,0 +1,20 @@
...
+      Defines a runtime api for `pallet-broker` for getting the current price 
+      of a core in the ongoing sale.
+
+  - audience: Runtime User
+    description: |
+      Defines a runtime api for `pallet-broker` for getting the current price 
+      of a core in the ongoing sale.
+
+crates:
+  - name: pallet-broker
```
seadanda, line 18: Minor bumps? Not up to date on this, @bkchr is minor the default if a change is listed here or should it be explicit?

substrate/frame/broker/src/dispatchable_impls.rs
seadanda: Maybe even `current_price` would be more descriptive, since in other places the sale refers to the whole sale period, whereas we want the price now. This would also allow the same name to be used for the runtime api function.

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -455,4 +454,25 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	/// If there is an ongoing sale returns the current price of a core.
+	pub fn api_sale_price() -> Result<BalanceOf<T>, DispatchError> {
+		let status = Status::<T>::get().ok_or(Error::<T>::Uninitialized)?;
+		let sale = SaleInfo::<T>::get().ok_or(Error::<T>::NoSales)?;
+
+		Self::ensure_cores_for_sale(&status, &sale)?;
+
+		let now = frame_system::Pallet::<T>::block_number();
```
seadanda, line 475: When we change to relay chain clocking this could give strange results if not updated. Maybe best to have a simple regression test here.

seadanda (approved):
Some nits and a regression test would be nice, but already happy to merge

substrate/frame/broker/src/dispatchable_impls.rs
Szegoo: The good thing is that a different type is used to define `now`, so if it isn't updated, it shouldn't even compile.

seadanda: Fair enough!

substrate/frame/broker/src/dispatchable_impls.rs
Szegoo: That makes sense, will update it now.

PR #3486, opened 2024-02-26
Skip awaiting for metric verification when booting up local zombienet networks

Overkillus:
When spinning larger zombienet networks locally during development awaiting for metrics to boot up and verify that nodes are up can take a lot of time:
```
Error 	 fetching metrics from: http://127.0.0.1:39555/metrics

Error 	 fetching metrics from: http://127.0.0.1:43957/metrics

Error 	 fetching metrics from: http://127.0.0.1:35039/metrics

Error 	 fetching metrics from: http://127.0.0.1:33173/metrics

Error 	 fetching metrics from: http://127.0.0.1:36597/metrics

Error 	 fetching metrics from: http://127.0.0.1:43837/metrics

```
Can take some time when playing with larger networks spinning the above error endlessly. 

PR uses the new setting [`node_verifier`](https://github.com/paritytech/zombienet/pull/1731) to disable the wait for metrics locally but keeps it on in CICD.

### DIFF
--- a/polkadot/zombienet_tests/functional/0001-parachains-pvf.toml
+++ b/polkadot/zombienet_tests/functional/0001-parachains-pvf.toml
@@ -1,5 +1,7 @@
 [settings]
 timeout = 1000
+# Don't wait for metrics to be available when run locally
+{% if not CI_JOB_ID %} node_verifier = "None" {% endif %}
 
 [relaychain]
 default_image = "{{ZOMBIENET_INTEGRATION_TEST_IMAGE}}"
--- a/polkadot/zombienet_tests/functional/0002-parachains-disputes.toml
+++ b/polkadot/zombienet_tests/functional/0002-parachains-disputes.toml
@@ -1,5 +1,7 @@
 [settings]
 timeout = 1000
+# Don't wait for metrics to be available when run locally
+{% if not CI_JOB_ID %} node_verifier = "None" {% endif %}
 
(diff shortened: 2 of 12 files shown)


pepoviola (approved):

sandreim:
I agree this is annoying, but the verification process is a good node startup sanity check. Removing it can actually backfire if not all nodes start properly. 

@pepoviola do we know why the node verifier process is so slow ?

pepoviola:
> I agree this is annoying, but the verification process is a good node startup sanity check. Removing it can actually backfire if not all nodes start properly.
> 
> @pepoviola do we know why the node verifier process is so slow ?

The verification itself is not slow, since is just query the `processStartTime` time and we are doing in batches of 10 nodes. _But_ the friction is caused when the nodes have an slow bootstrap (could be related to the load of the vm/infra/host) and the prometheus server take some time to response. Also, most of the times the test-suite doesn't query all the nodes as part of the assertions.

I'm working on a different approach that allow you to manage the verifier with this options:
- `None`, do not verify the nodes. I think this could be handy for local environments where you just want to spawn the network and make some manual checks.
- `Metric`, the current behavior of checking the prometheus server of the nodes
- `PID`, check that the process is alive (I think this will be ok for using in `CI`)

For v2, we want to add these options and also allow to set a `custom` verifier per node.

Thanks!!

Issue #3475, opened 2024-02-26
Bridge zombienet tests: check that the received amount on the destination AH is closed to the amount that was sent


serban300:
Part of https://github.com/paritytech/polkadot-sdk/issues/3176

In the bridge zombienet tests, for the moment we only check that some assets were transferred successfully and that the relayer was rewarded with > 0.

We should also check that the amount that is received on the destination is close to the amount that was sent (basically that the fees were reasonable)

serban300:
The adjustments for the Rococo<>Westend tests were merged via #3490 

Also adjusted the P<>K manua ltest PR: https://github.com/polkadot-fellows/runtimes/pull/198

Resolving the issue

PR #3490, opened 2024-02-27
Bridge zombienet tests: Check amount received at destination

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/3475


svyatonik (approved):
:+1: Thank you!

PR #3505, opened 2024-02-28
Remove `as frame_system::DefaultConfig` from the required syntax in `derive_impl`

gupnik:
Step in https://github.com/paritytech/polkadot-sdk/issues/171

This PR removes the need to specify `as [disambiguation_path]` for cases where the trait definition resides within the same scope as default impl path. 

For example, in the following macro invocation
```rust
#[derive_impl(frame_system::config_preludes::TestDefaultConfig as frame_system::DefaultConfig)]
impl frame_system::Config for Runtime {
   ...
}
```
the trait `DefaultConfig` lies within the `frame_system` scope and `TestDefaultConfig` impls the `DefaultConfig` trait. Using this information, we can compute the disambiguation path internally, thus removing the need of an explicit specification.

In cases where the trait lies outside this scope, we would still need to specify it explicitly, but this should take care of most (if not all) uses of `derive_impl` within FRAME's context.


liamaharon (approved):
Nice one! Can we do it for all cases in polkadot-sdk?

gupnik:
> Nice one! Can we do it for all cases in polkadot-sdk?

Yes, will do it in a separate PR to make the review easier.

substrate/frame/support/procedural/src/derive_impl.rs
```diff
@@ -172,6 +172,31 @@ fn combine_impls(
 	final_impl
 }
 
+fn compute_disambiguation_path(
```
ggwpez, line 175: Some internal docs please.

ggwpez (approved):
💯

substrate/frame/support/procedural/src/derive_impl.rs
gupnik: Added in the code but will add here as well

PR #3506, opened 2024-02-28
update development setup in sdk-docs

kianenigma:


docs/sdk/src/reference_docs/development_environment_advice.rs
```diff
@@ -111,3 +113,38 @@
...
+//! 		"check",
+//! 		"--workspace",
+//! 		"--message-format=json",
+//! 		"--all-targets",
+//! 		"--all-features",
+//! 		"--tests",
+//! 		"--target-dir=target/rust-analyzer"
+//! 	],
+//! }
+//! ```
```
seadanda, line 150: Since above we also provide the Lua, maybe it's worth doing the same here. WDYT?

```suggestion
//! ```

and the same in Lua for `neovim/nvim-lspconfig`:

//!
//! ```lua
//! ["rust-analyzer"] = {
//!   cargo = {
//!     buildScripts = {
//!       overrideCommand = {
//!         "cargo",
//!         "remote",
//!         "--build-env",
//!         "SKIP_WASM_BUILD=1",
//!         "--",
//!         "build",
//!         "--message-format=json",
//!         "--all-targets",
//!         "--all-features",
//!         "--tests",
//!         "--target-dir=target/rust-analyzer"
//!       }
//!     },
//!     check = {
//!       overrideCommand = {
//!         "cargo",
//!         "remote",
//!         "--build-env",
//!         "SKIP_WASM_BUILD=1",
//!         "--",
//!         "check",
//!         "--workspace",
//!         "--message-format=json",
//!         "--all-targets",
//!         "--all-features",
//!         "--tests",
//!         "--target-dir=target/rust-analyzer"
//!       },
//!     },
//!   },
//! },
//! ```
```

Disclaimer: I haven't checked this but this is a lua translation. @liamaharon please confirm.

seadanda (approved):
Just a Lua addition if you think it's worthwhile. It's a trivial difference but we've provided Lua for previous snippets.

docs/sdk/src/reference_docs/development_environment_advice.rs
```diff
@@ -39,6 +39,8 @@
 //!   // See the polkadot-sdk CI job that checks formatting for the current version used in
 //!   // polkadot-sdk.
 //!   "rust-analyzer.rustfmt.extraArgs": ["+nightly-2024-01-22"],
+//!   // useful when writing documentation
+//!   "rust-analyzer.semanticHighlighting.doc.comment.inject.enable": true,
```
liamaharon, line 43: Could you double check this does something? Should be `true` by default on latest RA version according to https://rust-analyzer.github.io/manual.html

docs/sdk/src/reference_docs/development_environment_advice.rs
```diff
@@ -111,3 +113,38 @@
...
+//! 		"remote",
+//! 		"--build-env",
+//! 		"SKIP_WASM_BUILD=1",
+//! 		"--",
+//! 		"check",
+//! 		"--workspace",
+//! 		"--message-format=json",
+//! 		"--all-targets",
+//! 		"--all-features",
+//! 		"--tests",
```
liamaharon, line 146: Could you check `--tests` is needed when already passing `--all-targets`?

docs/sdk/src/reference_docs/development_environment_advice.rs
seadanda: Done. I never even considered that the build task should really just run check. Thanks for all the extra minutes of my life that I will now have 😂

docs/sdk/src/reference_docs/development_environment_advice.rs
kianenigma: Yeah you are right, thanks for pointing it out.

kianenigma (commented):
Fix based in Liam's comments

liamaharon (approved):

PR #3513, opened 2024-02-28
Fix call enum's metadata regression

kianenigma:
This fixes an issue introduced in https://github.com/paritytech/substrate/pull/14101, in which I removed the `Call` enum's documentation and replaced it with a link to the `Pallet` struct, but this also removed any docs related to call from the metadata. 

I tried to add a regression test for this, but it seems to me that this is not possible, given that using `type-info` we only assert in type-ids for `Call`, `Event` and `Error`. I removed some doc comments from a test setup in `frame-support-test` to demonstrate the issue there. @jsdw do you have any comments on this? 

I also fixed a small issue in the custom html/css of `polkadot-sdk-doc` crate, making sure it does not affect the rust-doc page of all other crates. 

- [x] Investigate a regression test
- [x] prdoc


bkchr (approved):

bkchr:
> I tried to add a regression test for this, but it seems to me that this is not possible, given that using `type-info` we only assert in type-ids for `Call`, `Event` and `Error`

I don't get this comment? I mean you can find the documentation as well in the type info.

kianenigma:
> > I tried to add a regression test for this, but it seems to me that this is not possible, given that using `type-info` we only assert in type-ids for `Call`, `Event` and `Error`
> 
> I don't get this comment? I mean you can find the documentation as well in the type info.

You are right, thank you! I just needed to explore a bit more to find the right way to access it.

gupnik (approved):

Issue #2964, opened 2024-01-17
`reap_identity` should take batch of accounts


ggwpez:
While doing the identity migration on Rococo it became apparent that a `batch` is nice to use to speed it up, but the batch must be paid for.  
So in order to have free and fast migrations, it would need to accept a `Vec` of account IDs. cc @NachoPal

dharjeezy:
Hi @ggwpez are you talking of modifying the reap_identity function to take in vec of account ids or we introduce a new function that solely takes a vec?

ggwpez:
I think modifying `reap_identity` is fine. Its used as one-off anyway.

dharjeezy:
Circling back on this @ggwpez there is a `reap_identity` pallet impl function, are we modifying that too to take a vec of account_ids? or is it just the extrinsic we are modifying to take a vec?

Are we also modifying `on_reap_identity` handler to take a vec of account_ids too?

ggwpez:
>Circling back on this ggwpez there is a reap_identity pallet impl function, are we modifying that too to take a vec of account_ids? or is it just the extrinsic we are modifying to take a vec?

Dunno, just try it out and see what works best.

>Are we also modifying on_reap_identity handler to take a vec of account_ids too?

Probably okay to stay in singular.

ggwpez:
turns out we dont want this :(

PR #3516, opened 2024-02-29
reap_identity should take batch of accounts

dharjeezy:
closes https://github.com/paritytech/polkadot-sdk/issues/2964

Polkadot address: 12GyGD3QhT4i2JJpNzvMf96sxxBLWymz4RdGCxRH5Rj5agKW

### DIFF
--- a/polkadot/runtime/common/src/identity_migrator.rs
+++ b/polkadot/runtime/common/src/identity_migrator.rs
@@ -71,6 +71,7 @@ pub mod pallet {
 		traits::EnsureOrigin,
 	};
 	use frame_system::pallet_prelude::*;
+	use sp_std::vec::Vec;
 
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
--- a/polkadot/runtime/common/src/identity_migrator.rs
+++ b/polkadot/runtime/common/src/identity_migrator.rs
@@ -95,7 +95,7 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// The identity and all sub accounts were reaped for `who`.
-		IdentityReaped { who: T::AccountId },
+		IdentityReaped { who_vec: Vec<T::AccountId> },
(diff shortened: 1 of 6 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5386744

joepetrowski:
I am not sure about this PR at this stage. I see the issue discussion so sorry if we've misguided you. Also curious what @ggwpez thinks.

My issue is that we've already tested this on Rococo and Westend and it is in the 1.7.0 release that will go on Kusama. That means this change would most likely apply only to Polkadot. Of course bug fixes should go to Polkadot but I'm not as keen to throw untested optimizations at it.

This is certainly better, but it's code that's only used once and then will be removed from Polkadot. The main advantage is that it saves the sender a transaction fee when using `utility.batch`. Personally, I'd pay the transaction fee for the peace of mind that it's worked on two testnets and Kusama.

polkadot/runtime/common/src/identity_migrator.rs
```diff
@@ -112,20 +113,31 @@ pub mod pallet {
...
+
+			Self::deposit_event(Event::IdentityReaped { who_vec });
 			let post = PostDispatchInfo {
 				actual_weight: Some(<T as pallet::Config>::WeightInfo::reap_identity(
-					registrars, subs,
+					total_registrars,
+					total_subs,
 				)),
-				pays_fee: Pays::No,
+				pays_fee: Pays::Yes,
```
ggwpez, line 140: This is not really correct, as we want to omit fee payment in the success case.  
Normally we go for something like >=90% of good values justifies a free transaction.

ggwpez:
I supposed that it would be a trivial change (just one more for loop), but now i see that we would also need to calculate ratios again and reject empty vectors.  
But if we dont actually need it, then maybe not.

polkadot/runtime/common/src/identity_migrator.rs
dharjeezy: how do i identify good values to use as a justification of free transaction?

PR #3518, opened 2024-02-29
Fix compilation: allow unused imports

kapilsinha:
With the newest nightly compiler, there are lots of unused import warnings. Forbidding unused imports (and also denying warnings) breaks compilation. Other than the sassafras crate, no other crate forbids unused imports or denies warnings.


liamaharon:
You should use the latest stable version not nightly.

bkchr:
I mean these warnings are valid. Generally you should fix the warnings and not just remove `forbid(unused_imports)`

Issue #3496, opened 2024-02-27
Protocol command streams have been shut down


tugytur:
I've updated our westmint bootnode to release 1.8.0 and the logs get flooded with the following log

```
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
[Parachain] Protocol command streams have been shut down
```

bkchr:
Cc @paritytech/networking

altonen:
Are you able to capture 1-2 minutes of logs before this spam starts?

tugytur:
> Are you able to capture 1-2 minutes of logs before this spam starts?

It starts immediately.
```

17:04:31 [Parachain] 📦 Highest known block at #6447227
17:04:31 [Parachain] 〽️ Prometheus exporter started at 127.0.0.1:9615
17:04:31 [Parachain] Running JSON-RPC server: addr=127.0.0.1:9944, allowed origins=["*"]
17:04:31 [Relaychain] Sending fatal alert BadCertificate
17:04:31 [Relaychain] 🔍 Discovered new external address for our node: /ip4/213.167.226.186/tcp/30334/ws/p2p/12D3KooWSHLUgqsLVQUbPWPvDvhFSv89yPQDEZ7UaSjU19FYFKbq
17:04:32 [Parachain] 🔍 Discovered new external address for our node: /ip4/213.167.226.186/tcp/30339/ws/p2p/12D3KooWNjKeaANaeZxBAPctmx8jugSYzuw4vnSCJmEDPB5mtRd6
17:04:32 [Parachain] Report 12D3KooWBz8YQiSzZAfFh5U8jvTrAd7ga9JnEfNfskdk371dYcBW: -2147483648 to -2147483648. Reason: Genesis mismatch. Banned, disconnecting.
17:04:32 [Parachain] Report 12D3KooWSftdah4a5JuCiyi1BcgJVaiTmNHrvJwtGsuViVYejZZR: -2147483648 to -2147483648. Reason: Genesis mismatch. Banned, disconnecting.
[...]

PR #3523, opened 2024-02-29
Fix crash of synced parachain node run with `--sync=warp`

dmitry-markin:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3496.

### DIFF
--- a/substrate/client/network/sync/src/strategy.rs
+++ b/substrate/client/network/sync/src/strategy.rs
@@ -30,7 +30,7 @@ use crate::{
 };
 use chain_sync::{ChainSync, ChainSyncAction, ChainSyncMode};
 use libp2p::PeerId;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use prometheus_endpoint::Registry;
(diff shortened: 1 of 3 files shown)


altonen (approved):

substrate/client/network/sync/src/strategy.rs
```diff
@@ -587,3 +599,63 @@ where
...
+	use super::*;
+	use futures::executor::block_on;
+	use sc_block_builder::BlockBuilderBuilder;
+	use substrate_test_runtime_client::{
+		ClientBlockImportExt, ClientExt, DefaultTestClientBuilderExt, TestClientBuilder,
+		TestClientBuilderExt,
+	};
+
+	/// Regression test for crash when starting already synced parachain node with `--sync=warp`.
+	/// We must remove this after setting of warp sync target block is moved to initialization of
```
skunert, line 614: If this must later be removed, should we have an issue and refer to it here?

skunert (approved):

ggwpez:
Does this fix an end-user crash? Then it should probably go to the changelog via prdoc.

PR #3540, opened 2024-03-01
Contracts upload with Determinism::Enforced when possible

pgherveou:


substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -468,19 +468,32 @@ benchmarks! {
...
+	upload_code_determinism_relaxed {
+		let c in 0 .. T::MaxCodeLen::get();
+		let caller = whitelisted_caller();
+		T::Currency::set_balance(&caller, caller_funding::<T>());
+		let WasmModule { code, hash, .. } = WasmModule::<T>::sized(c, Location::Call, true);
+		let origin = RawOrigin::Signed(caller.clone());
+	}: upload_code(origin, code, None, Determinism::Relaxed)
+	verify {
+		assert!(T::Currency::total_balance_on_hold(&caller) > 0u32.into());
+		assert!(<Contract<T>>::code_exists(&hash));
```
athei, line 494: Can you add a check here the uploaded code actually has its determinism set to relaxed? I am worried that it might pass the enforced verification because it is just using a float local but no float instruction.

athei (approved):

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -468,19 +468,36 @@ benchmarks! {
...
+	#[pov_mode = Measured]
+	upload_code_determinism_relaxed {
+		let c in 0 .. T::MaxCodeLen::get();
+		let caller = whitelisted_caller();
+		T::Currency::set_balance(&caller, caller_funding::<T>());
+		let WasmModule { code, hash, .. } = WasmModule::<T>::sized(c, Location::Call, true);
+		let origin = RawOrigin::Signed(caller.clone());
+	}: upload_code(origin, code, None, Determinism::Relaxed)
+	verify {
+		assert!(T::Currency::total_balance_on_hold(&caller) > 0u32.into());
```
xermicus, line 495: Here I see assertion on two cases, however there's also the case of `Determinism::Relaxed` but no in-deterministic code, in which case there shouldn't be any penalty? Could be worthwhile to check unless this is already asserted on in some other place.

xermicus (commented):
I see

substrate/frame/contracts/src/benchmarking/mod.rs
pgherveou: this just assert here that we did indeed benchmark the most expensive code path.
since we will first try to load the module in deterministic mode, then fallback to non-deterministic.
The assertion here just validate that. All scenarios are tested in src/tests.rs below

pgherveou:
bot help

xermicus (approved):

athei:
Gotta love those random CI failures.

PR #3553, opened 2024-03-02
Fix: Region reserve transfers

Szegoo:
Reserve transferring from the Coretime chain to some other chain is fine. However, transferring back is problematic. Here is the explanation:

- Reserve transferring from the Coretime chain works since in this case the `TransferAsset` instruction is used which modifies the owner of the region.
- When transferring the region from some other chain to the Coretime chain the process is different. It consists of multiple steps:
  1. The derivative asset is withdrawn and burned on the source chain.
  2. The source chain sends an instruction to the Coretime chain to withdraw the asset from its sovereign account and place it into the holding registry. This process requires the region to be burned before putting it in the registry.
  3. Once the asset is in the registry, the XCM program clears its origin
  4. To transfer the region the `DepositAsset` instruction is used to move it from the registry to an actual account. This process involves 'minting' the asset back again.

This is problematic because it necessitates 'burning' and 'minting' the region when transferring it to and from the holding registry. If we were to actually burn the region when placing it into the holding registry we wouldn't know what to set the region record to when minting it back again.

For this reason, this PR proposes a solution that avoids actually minting and burning the region; it merely modifies the region's owner property.

- In the case of burning, it sets the owner to `None`.
- When minting the region, it assigns the owner to an actual value.

Setting the owner to `None` is a prerequisite for making it mintable.

https://github.com/paritytech/polkadot-sdk/blob/01ffbe01fe61ca4852ee7fd6504c7ee68a0f8d66/substrate/frame/broker/src/nonfungible_impl.rs#L55-L67

### TL;DR

This PR addresses the issue with cross-chain transferring regions back to the Coretime chain. Remote reserve transfers are performed by withdrawing and depositing the asset to and from the holding registry. This requires the asset to support burning and minting functionality. 

This PR adds burning and minting; however, they work a bit differently than usual so that the associated region record is not lost when burning. Instead of removing all the data, burning will set the owner of the region to `None`, and when minting it back, it will set it to an actual value. So, when cross-chain transferring, withdrawing into the registry will remove the region from its original owner, and when depositing it from the registry, it will set its owner to another account

### DIFF
--- a/prdoc/pr_3553.prdoc
+++ b/prdoc/pr_3553.prdoc
@@ -1,0 +1,28 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Region reserve transfers fix
+
+doc:
+  - audience: Runtime User
+    description: |
+      This PR addresses the issue with cross-chain transferring regions back to the
+      Coretime chain. TL;DR: Remote reserve transfers are performed by withdrawing
+      and depositing the asset to and from the holding registry. This requires the 
+      asset to support burning and minting functionality. 
+      This PR adds burning and minting; however, they work a bit differently than usual
+      so that the associated region record is not lost when burning. Instead of removing
+      all the data, burning will set the owner of the region to None, and when minting 
+      it back, it will set it to an actual value. So, when cross-chain transferring, 
+      withdrawing into the registry will remove the region from its original owner, and 
+      when depositing it from the registry, it will set its owner to another account
+
+migrations: 
+  db: []
+  runtime: 
+    - reference: pallet-regions
+      description: |
+        The region owner is optional.
+
+crates: 
+  - name: pallet-regions
--- a/prdoc/pr_3553.prdoc
+++ b/prdoc/pr_3553.prdoc
@@ -20,7 +20,7 @@ doc:
 migrations: 
   db: []
   runtime: 
-    - reference: pallet-regions
+    - reference: pallet-broker
       description: |
         The region owner is optional.
 
--- a/prdoc/pr_3553.prdoc
+++ b/prdoc/pr_3553.prdoc
@@ -25,4 +25,4 @@ migrations:
         The region owner is optional.
 
 crates: 
-  - name: pallet-regions
+  - name: pallet-broker
--- a/substrate/frame/broker/src/migration.rs
+++ b/substrate/frame/broker/src/migration.rs
@@ -21,10 +21,11 @@ use codec::{Decode, Encode};
 use core::marker::PhantomData;
 use frame_support::traits::{Get, OnRuntimeUpgrade};
 use sp_runtime::Saturating;
-use sp_std::vec::Vec;
 
 #[cfg(feature = "try-runtime")]
 use frame_support::ensure;
+#[cfg(feature = "try-runtime")]
+use sp_std::vec::Vec;
 
 #[derive(Encode, Decode)]
 pub struct RegionRecordV0<AccountId, Balance> {
--- a/substrate/frame/broker/src/migration.rs
+++ b/substrate/frame/broker/src/migration.rs
@@ -21,7 +21,7 @@ use codec::{Decode, Encode};
 use core::marker::PhantomData;
 use frame_support::traits::{Get, OnRuntimeUpgrade};
 use sp_runtime::Saturating;
-use sp_std::prelude::vec::Vec;
+use sp_std::vec::Vec;
 
 #[cfg(feature = "try-runtime")]
 use frame_support::ensure;
--- a/substrate/frame/broker/src/migration.rs
+++ b/substrate/frame/broker/src/migration.rs
@@ -22,6 +22,9 @@ use core::marker::PhantomData;
 use frame_support::traits::{Get, OnRuntimeUpgrade};
 use sp_runtime::Saturating;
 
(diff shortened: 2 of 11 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5415524

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5429462

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5430517

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5430516

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5430817

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5465803

seadanda (commented):
Looks great, couple of small things

substrate/frame/broker/src/nonfungible_impl.rs
```diff
@@ -46,11 +47,47 @@ impl<T: Config> Inspect<T::AccountId> for Pallet<T> {
...
+
+		// 'Minting' can only occur if the asset has previously been burned(i.e. moved to the
+		// holding registrar)
+		ensure!(record.owner.is_none(), Error::<T>::NotAllowed);
+		Self::issue(region_id.core, region_id.begin, record.end, who.clone(), record.paid);
+
+		Ok(())
+	}
+
+	fn burn(item: &Self::ItemId, maybe_check_owner: Option<&T::AccountId>) -> DispatchResult {
```
seadanda, line 81: When would you want to burn without checking the owner?

Szegoo: I implemented the `Mutate` trait, which contains the burn function with the optional owner check argument. We have a very similar implementation in the [`uniques`](https://github.com/paritytech/polkadot-sdk/blob/3c6ebd9e9bfda58f199cba6ec3023e0d12d6b506/substrate/frame/uniques/src/impl_nonfungibles.rs#L136) and [`nfts`](https://github.com/paritytech/polkadot-sdk/blob/3c6ebd9e9bfda58f199cba6ec3023e0d12d6b506/substrate/frame/nfts/src/impl_nonfungibles.rs#L252) pallet.

The `NonFungbile` asset transactor calls burn without checking the owner in the [`reduce_checked`](https://github.com/paritytech/polkadot-sdk/blob/3c6ebd9e9bfda58f199cba6ec3023e0d12d6b506/polkadot/xcm/xcm-builder/src/nonfungible_adapter.rs#L104C5-L104C19) function, which is only used for teleports, however, we won't be teleporting regions from the Coretime chain.

seadanda: I guess there's no harm in it being there as it is now, but I don't see a use for it since having regions being burnt without checking that it's the owner who instigated it, even as an admin call, would be strange

prdoc/pr_3553.prdoc
```diff
@@ -0,0 +1,28 @@
...
+      so that the associated region record is not lost when burning. Instead of removing
+      all the data, burning will set the owner of the region to None, and when minting 
+      it back, it will set it to an actual value. So, when cross-chain transferring, 
+      withdrawing into the registry will remove the region from its original owner, and 
+      when depositing it from the registry, it will set its owner to another account
+
+migrations: 
+  db: []
+  runtime: 
+    - reference: pallet-regions
```
seadanda, line 23:
```suggestion
    - reference: pallet-broker
```

prdoc/pr_3553.prdoc
```diff
@@ -0,0 +1,28 @@
...
+      description: |
+        The region owner is optional.
+
+crates: 
+  - name: pallet-regions
```
seadanda, line 28:
```suggestion
  - name: pallet-broker
```

seadanda (approved):

xlc:
The fact that XCM requires mint/burn to perform transfer just doesn't work well. e.g. incompatible with ERC20 / ERC721

Szegoo:
> The fact that XCM requires mint/burn to perform transfer just doesn't work well. e.g. incompatible with ERC20 / ERC721

Yeah, it doesn't really require minting and burning. It requires a way to move an asset from and to the holding registrar. I would guess that in the EVM world, this could be achieved by sending the asset to the precompile contract.

Szegoo:
@franciscoaguirre Could you review this?

substrate/frame/broker/src/migration.rs
```diff
@@ -0,0 +1,87 @@
...
+use sp_runtime::Saturating;
+
+#[cfg(feature = "try-runtime")]
+use frame_support::ensure;
+#[cfg(feature = "try-runtime")]
+use sp_std::vec::Vec;
+
+/// V0 region record.
+#[derive(Encode, Decode)]
+pub struct RegionRecordV0<AccountId, Balance> {
```
bkontur, line 32: I would move it to the `mod v1 {`, also maybe then `pub` is not needed

Szegoo:
Closed in favor of #3455

PR #3554, opened 2024-03-03
Update README.md Asset Conversion TX Payment

DrW3RK:
small docs fix.

### DIFF
--- a/substrate/frame/transaction-payment/asset-conversion-tx-payment/README.md
+++ b/substrate/frame/transaction-payment/asset-conversion-tx-payment/README.md
@@ -15,7 +15,7 @@ asset.
 
 ### Integration
 This pallet wraps FRAME's transaction payment pallet and functions as a replacement. This means
-you should include both pallets in your `construct_runtime` macro, but only include this
+you should not include both pallets in your `construct_runtime` macro, but only include this
 pallet's [`SignedExtension`] ([`ChargeAssetTxPayment`]).
 
 License: Apache-2.0


joepetrowski (changes requested):
Actually the docs are correct, you need both.

DrW3RK:
I see. It reads as -

"This means you should include both pallets in your `construct_runtime` macro, **but only include** this..."

which was a bit ambiguous.

joepetrowski:
It's talking about the `SignedExtension`. See AH Rococo for example:

- The runtime has [both pallets](https://github.com/paritytech/polkadot-sdk/blob/59b2661/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L905), `transaction_payment` and `asset_conversion_tx_payment` (ugh, "transaction"/"tx", style screwup 🤮), but
- The `SignedExtra` only has `ChargeAssetTxPayment`.
- While a "normal" one (e.g. [Bridge Hub Rococo](https://github.com/paritytech/polkadot-sdk/blob/59b2661444de2a25f2125a831bd786035a9fac4b/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L127)) only has `ChargeTransactionPayment`.

What the doc is trying to say where you are editing is that you only need `ChargeAssetTxPayment`, but you still need both pallets because it still needs to handle "regular" DOT transaction fees. The asset tx payment one will just also handle non-DOT assets.

the-right-joyce:
maybe we do want to explain it then a bit better or change language? I'm not a dev but it also read a bit confusing to me..

PR #3445, opened 2024-02-22
chainHead/follow: Provide multiple block hashes to the initialized event

lexnv:
This PR extends the Initialized event of the chainHead_follow subscription.

Now, the event provides multiple finalized block hashes. This information allows clients that are disconnected, and that want to reconnect, to not lose information about the state of the chain.

At the moment, the spec encourages servers to provide at least 1 minute of finalized blocks (~10 blocks). The users are responsible for unpinning these blocks at a later time. This PR tries to report at least 1 finalized block and at most 16 blocks, if they are available.

Closes: https://github.com/paritytech/polkadot-sdk/issues/3432
cc @paritytech/subxt-team

PR #3562, opened 2024-03-04
chainHead: Ensure reasonable distance between leaf and finalized block

lexnv:
This PR ensure that the distance between any leaf and the finalized block is within a reasonable distance.

 For a new subscription, the chainHead has to provide all blocks between the leaves of the chain and the finalized block.
 When the distance between a leaf and the finalized block is large:
 - The tree route is costly to compute
 - We could deliver an unbounded number of blocks (potentially millions)
 (For more details see https://github.com/paritytech/polkadot-sdk/pull/3445#discussion_r1507210283)

The configuration of the ChainHead is extended with:
 - suspend on lagging distance: When the distance between any leaf and the finalized block is greater than this number, the subscriptions are suspended for a given duration.
    - All active subscriptions are terminated with the `Stop` event, all blocks are unpinned and data discarded.
    - For incoming subscriptions, until the suspended period expires the subscriptions will immediately receive the `Stop` event.
    - Defaults to 128 blocks
 - suspended duration: The amount of time for which subscriptions are suspended
    - Defaults to 30 seconds
 
 
 cc @paritytech/subxt-team

### DIFF
--- a/substrate/client/rpc-spec-v2/src/chain_head/tests.rs
+++ b/substrate/client/rpc-spec-v2/src/chain_head/tests.rs
@@ -89,6 +89,7 @@ pub async fn run_server() -> std::net::SocketAddr {
 			subscription_max_ongoing_operations: MAX_OPERATIONS,
 			operation_max_storage_items: MAX_PAGINATION_LIMIT,
 			max_follow_subscriptions_per_connection: 1,
+			max_lagging_distance: MAX_LAGGING_DISTANCE,
 		},
 	)
 	.into_rpc();


substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
```diff
@@ -86,12 +96,26 @@ const MAX_ONGOING_OPERATIONS: usize = 16;
...
+/// block is larger than this value.
+///
+/// Subscriptions are suspended for the `suspended_duration`.
+const SUSPEND_ON_LAGGING_DISTANCE: usize = 128;
+
+/// The amount of time for which the subscriptions are suspended.
+///
+/// Subscriptions are suspended when the distance between any leaf
+/// and the finalized block is too large.
+const SUSPENDED_DURATION: Duration = Duration::from_secs(30);
```
lexnv, line 109: An alternative for this would be to simply recalculate the:

```rust
leaves = blockchain.leaves()?;
if leaf - finalized > 128 { return Err(); }
```

substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
```diff
@@ -623,6 +637,19 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionsInner<Block, BE> {
 		}
 	}
 
+	/// Suspends all subscriptions for the given duration.
```
skunert, line 640:
```suggestion
	/// Suspends all subscriptions for the configured duration.
```

skunert (approved):
Logic looks good to me!

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
jsdw: Is there much of a computation cost for doing the above? Personally I think fewer constants is a good thing, but if it's expensive to compute then having a delay makes sense :)

jsdw (approved):

substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs
lexnv: Yep that makes sense, I was also wondering if the `suspended duration` is overkill for this.

We might have a slight increase in reading the blockchain leaves numbers if they are placed on the disk, however I think we can live with that, since this is only an edge-case of an edge-case. 

Have removed that constant and we should allow subscriptions to resume normal work as soon as the distance is reasonable.

substrate/client/rpc-spec-v2/src/chain_head/subscription/inner.rs
```diff
@@ -599,11 +591,6 @@ impl<Block: BlockT, BE: Backend<Block>> SubscriptionsInner<Block, BE> {
 		sub_id: String,
 		with_runtime: bool,
 	) -> Option<InsertedSubscriptionData<Block>> {
-		if self.suspend.is_suspended() {
-			log::trace!(target: LOG_TARGET, "[id={:?}] Subscription already suspended", sub_id);
-			return None
-		}
```
jsdw: Just to confirm; is the point of removing this now that we will allow new subscriptions to be added any time, but stop them all regardless whenever `generate_events` says that the distance is too large? Whereas before we'd not allow new subscriptions to be added for the timeout period.

lexnv: Yep, that makes sense

jsdw: Thanks! The changes def simplify things a fair bit so I'm in favour, and the block distance being too large is def an edge case anyways :)

PR #2680, opened 2023-12-11
[Moved] Enable governance for Nomination Pool members

Ank4n:
closes https://github.com/paritytech/polkadot-sdk/issues/408 and https://github.com/paritytech/polkadot-sdk/issues/454.

## Update [April 1, 2024]
This PR has been broken down into three different PRs targeting individual pallets that are affected. Please refer to the following:
1) https://github.com/paritytech/polkadot-sdk/pull/3889.
2) https://github.com/paritytech/polkadot-sdk/pull/3904.
3) https://github.com/paritytech/polkadot-sdk/pull/3905.

## Problem
When a staker directly stakes on pallet-staking, their funds are locked in their account and stakers can use their locked funds for participating in governance. For delegators in Nomination Pool, their funds are transferred to a pool account (pallet sub account) instead. This means they lose access to their locked funds and cannot use it for governance.

We could fix this if NominationPool instead held funds in the user accounts itself and somehow conveyed to Staking Pallet that the pool account represents the locked amount of all its delegators.

##### How NP works currently:
1) transfer tokens from delegator to pool account, and
2) stake from pool account as a direct nominator. Tokens are locked in the pool account.

image

#####  How we want it to work:
1) delegate fund from delegator to pool account with tokens locked in delegator account itself.
2) stake from pool account as if all delegated funds to pool account are its own.

image

This is what we try to achieve with `pallet-delegated-staking`. It is a wrapper implementation on top of staking-pallet that adds a new primitive of `Delegation` to staking.

## Other design goals and challenges
- Both nomination pool and staking are already big codebases and we want to minimise changes to each of this pallet to avoid adding regression.
- Delegation based staking is configurable and runtimes can choose to enable or disable it.
- Avoid greedy slashing of delegators of a `Delegatee` account to keep slashing event bounded.

### Delegatee Account
We introduce a new primitive for staking, where an account can receive `delegation` from multiple `delegator` accounts. We call the account that accepts delegations as `Delegatee`. Their staking balance is made up of multiple child delegator accounts and for staking purposes their actual balance is ignored. Importantly, the balance is held in individual delegator accounts and not actually moved to the delegatee account.

image

### Staking pallet
Accepts a new type `DelegateeSupport` to support the above `Delegatee` accounts.

### Delegated Staking Pallet
Implements `DelegateeSupport` and `DelegatedStakingInterface`.

`DelegatedStakingInterface` is an extension of `StakingInterface` that also supports delegation of accounts.

image

### Nomination Pool
[...]

PR #3567, opened 2024-03-04
[DNM] Refactoring delegated staking interfaces

georgepisaltu:
Built on top of #2680 , but with #1933 in mind.

This PR roughly implements decoupling changes in `pallet-delegated-staking` and `pallet-staking`. The changes are most notably reflected in `trait StakingInterface`, `trait DelegateeSupport` and `trait DelegatedStakingInterface`.

The main difference from the original approach is:
- `trait DelegateeSupport` becomes an interface which extends a `trait StakingInterface` to allow registering bonds without locking the funds
- `trait DelegatedStakingInterface` no longer provides the full staking functionality, but instead it's meant to be a buffer which abstracts away any `StakingInterface` and lets users just delegate

NOTE: the traits might need renaming given the shift in functionality

Consequently, `pallet-staking` implements `DelegateeSupport` and `pallet-delegated-staking` implements `DelegatedStakingInterface`. Entities which use `DelegateeSupport` are inherently privileged because they are allowed to stake without being checked by the staking interface. This will need some origin validation in the near future.

What this means is that all of the responsibility for validating the state of delegators/delegatees falls on the callers of `DelegateeSupport`, which would be `DelegatedStakingInterface` implementors in our case. Most of the core staking logic remains unchanged.

Next step here would be to refactor the interface usage of  `pallet-nomination-pools`, but until then, feedback is welcome on the direction of the interfaces.

### DIFF
--- a/substrate/frame/staking/src/lib.rs
+++ b/substrate/frame/staking/src/lib.rs
@@ -328,7 +328,7 @@ pub use sp_staking::{Exposure, IndividualExposure, StakerStatus};
 use sp_std::{collections::btree_map::BTreeMap, prelude::*};
 pub use weights::WeightInfo;
 
-pub use pallet::{pallet::*, NoDelegation, UseNominatorsAndValidatorsMap, UseValidatorsMap};
+pub use pallet::{pallet::*, UseNominatorsAndValidatorsMap, UseValidatorsMap};
 
 pub(crate) const STAKING_ID: LockIdentifier = *b"staking ";
 pub(crate) const LOG_TARGET: &str = "runtime::staking";
@@ -413,21 +413,6 @@ pub enum RewardDestination<AccountId> {
 	None,
 }
 
-impl<AccountId: Clone> RewardDestination<AccountId> {
-	fn from(self, stash: &AccountId) -> Option<AccountId> {
-		match self {
-			// FIXME(ank4n): Figure out later how to handle Controller
-			RewardDestination::Staked | RewardDestination::Stash => Some(stash.clone()),
-			RewardDestination::Account(a) => Some(a),
-			#[allow(deprecated)]
-			_ => {
-				defensive!("reward destination not set or set as deprecated controller");
-				None
-			},
-		}
-	}
-}
-
 /// Preference of what happens regarding validation.
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo, Default, MaxEncodedLen)]
 pub struct ValidatorPrefs {
@@ -498,6 +483,9 @@ pub struct StakingLedger<T: Config> {
 	/// Refer to issue <https://github.com/paritytech/polkadot-sdk/issues/433>
 	pub legacy_claimed_rewards: BoundedVec<EraIndex, T::HistoryDepth>,
 
+	/// TODO: docs
+	pub locked_locally: bool,
+
 	/// The controller associated with this ledger's stash.
 	///
 	/// This is not stored on-chain, and is only bundled when the ledger is read from storage.
@@ -560,6 +548,7 @@ impl<T: Config> StakingLedger<T> {
 			active: self.active,
 			unlocking,
 			legacy_claimed_rewards: self.legacy_claimed_rewards,
+			locked_locally: self.locked_locally,
 			controller: self.controller,
 		}
 	}
--- a/substrate/frame/staking/src/lib.rs
+++ b/substrate/frame/staking/src/lib.rs
@@ -483,9 +483,6 @@ pub struct StakingLedger<T: Config> {
 	/// Refer to issue <https://github.com/paritytech/polkadot-sdk/issues/433>
 	pub legacy_claimed_rewards: BoundedVec<EraIndex, T::HistoryDepth>,
 
-	/// TODO: docs
-	pub locked_locally: bool,
-
 	/// The controller associated with this ledger's stash.
 	///
 	/// This is not stored on-chain, and is only bundled when the ledger is read from storage.
@@ -548,7 +545,6 @@ impl<T: Config> StakingLedger<T> {
 			active: self.active,
 			unlocking,
 			legacy_claimed_rewards: self.legacy_claimed_rewards,
-			locked_locally: self.locked_locally,
 			controller: self.controller,
 		}
 	}
--- a/substrate/frame/staking/src/pallet/impls.rs
+++ b/substrate/frame/staking/src/pallet/impls.rs
@@ -1084,6 +1084,35 @@ impl<T: Config> Pallet<T> {
 		outcome
 	}
 
+	/// TODO: docs
+	/// Basically another call for `bond`, but it's implied that the funds are already locked
+	/// and `pallet-staking` doesn't need to lock them here.
+	pub fn do_delegated_bond(
+		stash: T::AccountId,
+		value: BalanceOf<T>,
+		payee: RewardDestination<T::AccountId>,
+	) -> DispatchResult {
+		if StakingLedger::<T>::is_bonded(StakingAccount::Stash(stash.clone())) {
+			return Err(Error::<T>::AlreadyBonded.into())
+		}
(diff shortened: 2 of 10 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5431263

substrate/frame/staking/src/lib.rs
```diff
@@ -498,6 +483,9 @@ pub struct StakingLedger<T: Config> {
 	/// Refer to issue <https://github.com/paritytech/polkadot-sdk/issues/433>
 	pub legacy_claimed_rewards: BoundedVec<EraIndex, T::HistoryDepth>,
 
+	/// TODO: docs
+	pub locked_locally: bool,
```
Ank4n, line 487: This would be an expensive migration. We should try to avoid it.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1990,6 +1983,47 @@ pub mod pallet {
...
+		/// Basically another call for `bond`, but it's implied that the funds are already locked
+		/// and `pallet-staking` doesn't need to lock them here.
+		#[pallet::call_index(29)]
+		#[pallet::weight(T::WeightInfo::bond())]
+		pub fn delegated_bond(
+			origin: OriginFor<T>,
+			#[pallet::compact] value: BalanceOf<T>,
+			payee: RewardDestination<T::AccountId>,
+		) -> DispatchResult {
+			// To make this truely decoupled, we would need a `EnsureDelegatorOrigin` type on
```
Ank4n, line 1997: How would you do `EnsureDelegatorOrigin`? I am guessing we will use a signed extension that calls into Delegator Staking pallet? 
I don't see how it is better than decoupling it via a glue trait. And the worse thing about this is `SignedExtension` should be as computationally cheap as possible since it gets called for every dispatchable and doing a database read on it is a significant overhead.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5436105

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1990,6 +1992,23 @@ pub mod pallet {
...
+		#[pallet::weight(T::WeightInfo::bond())]
+		pub fn delegated_bond(
+			origin: OriginFor<T>,
+			stash: T::AccountId,
+			#[pallet::compact] value: BalanceOf<T>,
+			payee: RewardDestination<T::AccountId>,
+		) -> DispatchResult {
+			// `DelegateOrigin` type lets approved pallets on different parachains to call into this
+			// privileged extrinsic.
+			T::DelegateOrigin::ensure_origin(origin)?;
```
Ank4n, line 2009: How do we verify the `value` is available to bond (delegated and unbonded)?

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1084,6 +1084,35 @@ impl<T: Config> Pallet<T> {
...
+			return Err(Error::<T>::AlreadyBonded.into())
+		}
+
+		// Reject a bond which is considered to be _dust_.
+		if value < T::Currency::minimum_balance() {
+			return Err(Error::<T>::InsufficientBond.into())
+		}
+
+		frame_system::Pallet::<T>::inc_consumers(&stash).map_err(|_| Error::<T>::BadState)?;
+		let stash_balance = T::Currency::free_balance(&stash);
```
Ank4n, line 1105: We can't look at free balance of the stash. We need to somehow query how much balance is delegated to this `stash` (delegatee)

substrate/frame/staking/src/pallet/mod.rs
Ank4n, line 1: The way you are decoupling staking and delegate staking pallets is by injecting `DelegateOrigin` which would also come from and be checked in the delegate staking pallet. My solution was using a glue trait. If we only needed to ensure the account is a Delegatee, I agree this is a better solution. 

The issue though is, its not just ensuring its delegatee account but also checking if balance is valid. `DelegateeSupport` acts like a `Currency` provider in the original solution for these `delegatee` accounts. 

Another requirement we want to achieve is to somehow restrict delegatee to bond with `RewardDestination::Staked` that would be hard to achieve without this glue.

substrate/frame/staking/src/pallet/mod.rs
georgepisaltu: Why do we need to verify it if it's coming from a trusted delegated staking source? It should have already been verified by the caller.

But technically if we'll need it (though I think right now we don't), we can just make `DelegatorOrigin` return the available balance on the delegate side on success.

substrate/frame/staking/src/pallet/impls.rs
georgepisaltu: This check was here before for some other reason and now it's not even needed, so I removed it.

substrate/frame/staking/src/pallet/mod.rs
georgepisaltu: > The way you are decoupling staking and delegate staking pallets is by injecting `DelegateOrigin`

In this particular solution, the `DelegateOrigin` thing is necessary only if you want to use the delegated staking feature from a different chain, which is unnecessary right now as both `pallet-staking` and `pallet-delegated-staking` would be on the same runtime. This particular extrinsic could even be removed. It's only useful if we'd want to use it in some XCM `transact`s for example. So, I'd disregard it for the time being.

> The issue though is, its not just ensuring its delegatee account but also checking if balance is valid. DelegateeSupport acts like a Currency provider in the original solution for these delegatee accounts.

I am deliberately avoiding this by having the caller, `pallet-delegated-staking`, do all the checks and validation for me. I am essentially giving the caller of this interface root-like privileges because I am checking in `pallet-delegated-staking` that the funds are enough, that they are locked, that the origin is signed, etc.

> Another requirement we want to achieve is to somehow restrict delegatee to bond with `RewardDestination::Staked`

Why is it hard? It's not user facing functionality, it's runtime facing functionality. This isn't a problem if the caller (`pallet-delegated-staking`, which does provide the user facing functionality) only bonds using a pre-validated account and `<pallet_staking::Config as DelegateeSupport>::delegated_bond` always bonds with `RewardDestination::Staked` or something else that we want to do.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5462186

substrate/frame/staking/src/pallet/mod.rs
Ank4n: > In this particular solution, the DelegateOrigin thing is necessary only if you want to use the delegated staking feature from a different chain

We will expose it outside of runtime at some point.

> I am deliberately avoiding this by having the caller, pallet-delegated-staking, do all the checks and validation for me. I am essentially giving the caller of this interface root-like privileges because I am checking in pallet-delegated-staking that the funds are enough, that they are locked, that the origin is signed, etc.

This basically implies the caller can only be delegated staking pallet (in which case we don't need this extrinsic at all). How would the delegation + stake steps look like for an external delegatee (non-pool)?

> It's not user facing functionality, it's runtime facing functionality. 

It will be a runtime facing functionality eventually.

Ank4n: > How would the delegation + stake steps look like for an external delegatee (non-pool)?

I think this is where we are diverging. A foreign delegatee can register and get delegation via delegation pallet, and as long as they do all op via delegation pallet, we can make sure they are not doing anything funky. But this delegatee can call `staking::bond` or `bond_extra` directly as well. 

One way to handle it is by staking keeping `Delegatees` in its own storage as well. But then I would argue, an injected DelegateeSupport is better design than Staking having some half baked delegatee logic in its code.

PR #3574, opened 2024-03-05
fix(pallet-benchmarking): split test functions in v2

### LINKED ISSUE #376: bench syntax v2: run test in single functions
The v2 benchmark tests are running in one single function instead of one function per bench case.  

Current output:
```pre
test benchmark_tests::test_benchmarks ... ok
```

Output that we want:
```pre
test benchmarking::bench_sort_vector ... ok
test benchmarking::bench_accumulate_dummy ... ok
test benchmarking::bench_set_dummy_benchmark ... ok
```

It is probably using this wrong expansion arm here:  
https://github.com/paritytech/substrate/blob/2e535bac47f48d816f94b105715e09dc57ed52b0/frame/benchmarking/src/v1.rs#L541

instead of:  
https://github.com/paritytech/substrate/blob/2e535bac47f48d816f94b105715e09dc57ed52b0/frame/benchmarking/src/v1.rs#L508

pandres95:
Closes #376

### DIFF
--- a/substrate/frame/benchmarking/src/v1.rs
+++ b/substrate/frame/benchmarking/src/v1.rs
@@ -920,7 +920,7 @@ macro_rules! impl_bench_name_tests {
 // Every variant must implement [`BenchmarkingSetup`].
 //
 // ```nocompile
-//
+// 
 // struct Transfer;
 // impl BenchmarkingSetup for Transfer { ... }
 //
@@ -1866,7 +1866,7 @@ macro_rules! add_benchmark {
 						verify,
 						e,
 					);
-					return Err(e.into());
+					return Err(e.into())
 				},
 				Err($crate::BenchmarkError::Skip) => {
 					$crate::__private::log::error!(
--- a/substrate/frame/benchmarking/src/v1.rs
+++ b/substrate/frame/benchmarking/src/v1.rs
@@ -861,7 +861,7 @@ macro_rules! impl_bench_name_tests {
 					// Same per-case logic as when all cases are run in the
 					// same function.
 					match std::panic::catch_unwind(|| {
-						$module::<$test>::test_bench_by_name(stringify!($name).as_bytes())
+						$module::<$test>::[< test_benchmark_ $name >] ()
 					}) {
 						Err(err) => {
 							panic!("{}: {:?}", stringify!($name), err);
@@ -920,7 +920,7 @@ macro_rules! impl_bench_name_tests {
 // Every variant must implement [`BenchmarkingSetup`].
 //
 // ```nocompile
-// 
+//
 // struct Transfer;
 // impl BenchmarkingSetup for Transfer { ... }
 //
@@ -1866,7 +1866,7 @@ macro_rules! add_benchmark {
 						verify,
 						e,
 					);
-					return Err(e.into())
+					return Err(e.into());
 				},
 				Err($crate::BenchmarkError::Skip) => {
 					$crate::__private::log::error!(
--- a/substrate/frame/benchmarking/src/v1.rs
+++ b/substrate/frame/benchmarking/src/v1.rs
@@ -861,7 +861,7 @@ macro_rules! impl_bench_name_tests {
 					// Same per-case logic as when all cases are run in the
 					// same function.
 					match std::panic::catch_unwind(|| {
-						$module::<$test>::[< test_benchmark_ $name >] ()
+						$module::<$test>::test_bench_by_name(stringify!($name).as_bytes())
 					}) {
 						Err(err) => {
 							panic!("{}: {:?}", stringify!($name), err);
--- a/substrate/frame/support/procedural/src/benchmark.rs
+++ b/substrate/frame/support/procedural/src/benchmark.rs
@@ -441,6 +441,29 @@ pub fn benchmarks(
 		benchmarks_by_name_mappings.push(quote!(#name_str => Self::#test_ident()))
 	}
 
+	let impl_test_function = if let Some(item) = content.iter_mut().find(|item| matches!(item, Item::Macro(_))) {
+		let Item::Macro(item_macro) = item else {
+			return Err(syn::Error::new(item.span(), "Expected an item macro, as found such item"));
+		};
+
+		if item_macro
+			.mac
+			.path
+			.segments
+			.iter()
+			.any(|s| s.ident == "impl_benchmark_test_suite")
+		{
+			let tokens = item_macro.mac.tokens.clone();
+			*item = Item::Verbatim(quote! {});
+
+			quote! { impl_test_function!((#( {} #benchmark_names )*)()()#tokens); }
+		} else {
+			quote! {}
+		}
+	} else {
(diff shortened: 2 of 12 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5437291

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5445149

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -85,12 +85,12 @@ impl syn::parse::Parse for BenchmarkAttrKeyword {
...
 		if lookahead.peek(keywords::extra) {
 			let _extra: keywords::extra = input.parse()?;
-			return Ok(BenchmarkAttrKeyword::Extra)
+			return Ok(BenchmarkAttrKeyword::Extra);
 		} else if lookahead.peek(keywords::skip_meta) {
 			let _skip_meta: keywords::skip_meta = input.parse()?;
-			return Ok(BenchmarkAttrKeyword::SkipMeta)
+			return Ok(BenchmarkAttrKeyword::SkipMeta);
 		} else {
-			return Err(lookahead.error())
```
ggwpez, line 93: Please revert any unrelated format changes.

substrate/frame/benchmarking/src/v1.rs
```diff
@@ -861,7 +861,7 @@ macro_rules! impl_bench_name_tests {
 					// Same per-case logic as when all cases are run in the
 					// same function.
 					match std::panic::catch_unwind(|| {
-						$module::<$test>::[< test_benchmark_ $name >] ()
+						$module::<$test>::test_bench_by_name(stringify!($name).as_bytes())
```
ggwpez, line 864: Why was this changed?

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -441,6 +444,31 @@ pub fn benchmarks(
...
+			.mac
+			.path
+			.segments
+			.iter()
+			.any(|s| s.ident == "impl_benchmark_test_suite")
+		{
+			let tokens = item_macro.mac.tokens.clone();
+			*item = Item::Verbatim(quote! {});
+
+			quote! { impl_test_function!((#( {} #benchmark_names )*)(#( #extra_benchmark_names )*)(#( #skip_meta_benchmark_names )*)#tokens); }
```
ggwpez, line 464: Can you put a few new-lines in here please to make it easier to read?  
And why do we need the `{}`?

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -441,6 +444,31 @@ pub fn benchmarks(
 		benchmarks_by_name_mappings.push(quote!(#name_str => Self::#test_ident()))
 	}
 
+	let impl_test_function = if let Some(item) =
+		content.iter_mut().find(|item| matches!(item, Item::Macro(_)))
+	{
+		let Item::Macro(item_macro) = item else {
+			return Err(syn::Error::new(item.span(), "Expected an item macro, as found such item"));
+		};
+
```
pgherveou, line 453: maybe just use a find_map here

substrate/frame/benchmarking/src/v1.rs
pandres95: This is to ensure this generated `#[test]` calls can be executed for both `v1` and `v2` produced benchmarks, as `v1` produces calls named `Pallet::test_bench_$benchmark_name`, while `v2` produces calls named `test_{benchmark_name}`, but both macros share the `test_bench_by_name` call.

Another possibility is to achieve this, while ensuring retrocompatibility with `v1` is to change its analog benchmark generation in `v2` macro, like this:

> In `substrate/frame/support/procedural/src/benchmark.rs` line `434`:
```patch
- 		.map(|n| Ident::new(format!("test_{}", n).as_str(), Span::call_site()))
+ 		.map(|n| Ident::new(format!("test_benchmark_{}", n).as_str(), Span::call_site()))
```

substrate/frame/support/procedural/src/benchmark.rs
pandres95: Sure! I'll add those new lines.

The `quote! {}` in line [462](https://github.com/pandres95/polkadot-sdk/blob/bab21d25ab922f29114d3da956d88064edc3249c/substrate/frame/support/procedural/src/benchmark.rs#L462) is to suppress `impl_benchmark_test_suite` which otherwise left would generate the now existing `test_benchmarks` test function as well as the test for each benchmark function.

The `{}` behind `#benchmark_names` is necessary to match the input for the `impl_bench_case_tests` macro.

pgherveou (approved):

pgherveou:
LGTM, the only difference with v1 now is that the tests are nested into the benchmarks mod

pandres95:
Correct! That'd be the only difference.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5458546

ggwpez:
>LGTM, the only difference with v1 now is that the tests are nested into the benchmarks mod

I think that actually makes more sense then if they would teleport themselves outside.

substrate/frame/benchmarking/src/v1.rs
ggwpez: Yes i think that is better. I would like to keep the changes to the old V1 syntax as minimal as possible - since dozens f projects still use it and possibly depend on implementation specific behaviour.

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -441,6 +441,37 @@ pub fn benchmarks(
...
+				.path
+				.segments
+				.iter()
+				.any(|s| s.ident == "impl_benchmark_test_suite")
+			{
+				return None;
+			}
+
+			let tokens = item_macro.mac.tokens.clone();
+			*item = Item::Verbatim(quote! {});
```
ggwpez, line 462: Just as note: You could also directly replace the `item`with the new quote code, but it looks cleaner like this.

ggwpez (approved):
Nice!

ggwpez:
bot fmt

substrate/frame/support/procedural/src/benchmark.rs
pandres95: Actually, that's not possible, as it would be inserted before all the benchmarks structures and methods are defined, causing _undefined_ errors on checks (already tried that, actually 😅).

pandres95:
@ggwpez the prdoc is ready for reviewing.

Polkadot-Forum:

https://forum.polkadot.network/t/decoded-2024-sponsorship-for-active-community-members/7654/71

Issue #1797, opened 2023-10-05
Expose claim queue for asynchronous on-demand operability


eskimor:
In order for on-demand (and other forms of more exotic scheduling) to take advantage of asynchronous backing we need to expose the claim queue to the node side and have validators have accept collations for any `ParaId` that is in the claim queue, not just the next scheduled entry.

alindima:
This code in prospective-parachains will also need to be changed to use the claimqueue API: https://github.com/paritytech/polkadot-sdk/blob/2e4e65711233c6f3a1adc9ce49af8f4537de5439/polkadot/node/core/prospective-parachains/src/lib.rs#L843

I think it should lookahead according to allowed_ancestry_len or scheduling_lookahead, in order to allow collations coming from upcoming paras.

bkchr:
> * Make use of it in collators: Implement strategies for the right lookahead in collators (see comment below)

https://github.com/paritytech/polkadot-sdk/issues/3168 is included there. The collation generation subsystem will not be used for this anymore. Already the lookahead collator is not using the collation generation subsystem for triggering the block production.

Issue #793, opened 2022-08-16
Fix spurious runtime not supported errors


eskimor:
A couple of times we get messages on versi that certain runtime requests are not supported. Which is not expected, as the deployed runtime does support them:

```


2022-08-16 14:46:09 | 2022-08-16 12:46:09.709  WARN tokio-runtime-worker parachain::dispute-coordinator: error=Runtime(RuntimeRequest(NotSupported { runtime_api_name: "candidate_events" }))
-- | --
  |  

```

Full logs: https://grafana.parity-mgmt.parity.io/goto/7HD0ZPm4k?orgId=1

This should be investigated as working runtime calls are crucial for disputes to work correctly. Related issue paritytech/polkadot#5882 

TODOs:

- [x] Have https://github.com/paritytech/substrate/pull/12707 merged in substrate.
- [x] Show better error message instead of the misleading `parachain::dispute-coordinator: error=Runtime(RuntimeRequest(NotSupported))` https://github.com/paritytech/polkadot/pull/6288

tdimitrov:
I think we are quite familiar with the root cause for this issue and we can close it.

To summarise the problem. Each runtime call performed via runtime api subsystem consists of two calls:
1. ApiVersion call to determine the runtime version and if the desired runtime call is supported.
2. The desired runtime call.

The error occurs at step 1. For example (actual occurrence on versi):
```
WARN tokio-runtime-worker parachain::runtime-api: cannot query the runtime API version: Api called for an unknown Block: State already discarded for 0xf733059cf939d9c753f29b2a5d7cca05c334d9ced1151445f651702fe52f4d0a
```
Digging further what happened with the relay parent explains the error:
```
 WARN tokio-runtime-worker sync: 💔 Ignored block (#224794 -- 0xf733…4d0a) announcement from 12D3KooWD4RtZaY97mncXTFcerDmzuQBpeJjDmpUtEQLcFRLeJsw because all validation slots for this peer are occupied.
 ```
 
# Steps to investigate such issues
Usually this is what I do (based on the example I mentioned above):
1. You see a state discarded error. Or you look for it explicitly with something like `{level="WARN",subtarget="runtime-api"} |~ "State discarded`.
2. Pinpoint a single node for debugging and check the errors there - `{level="WARN",subtarget="runtime-api", node="versi-validator-a-node-10"}`
[...]

PR #3580, opened 2024-03-05
Expose `ClaimQueue` via a runtime api and use it in `collation-generation` 

tdimitrov:
The PR adds two things:
1. Runtime API exposing the whole claim queue
2. Consumes the API in `collation-generation` to fetch the next scheduled `ParaEntry` for an occupied core.

Related to https://github.com/paritytech/polkadot-sdk/issues/1797


polkadot/node/collation-generation/src/lib.rs
```diff
@@ -598,3 +617,39 @@ fn erasure_root(
...
+	// TODO: In the future we will return more than one scheduled para_id hence the weird processing
+	// below
+	claim_queue
+		.get(&core_idx)?
+		.into_iter()
+		.map(|paras_entry| paras_entry.para_id())
+		.take(1)
+		.next()
+		.map(|para_id| ScheduledCore { para_id, collator: None })
+}
```
tdimitrov, line 655: This is an overkill if we just need one `ScheduledCore`. I'll fix it if we decide not to fetch more than one entry.

polkadot/node/core/runtime-api/src/tests.rs
```diff
@@ -286,6 +287,13 @@ impl RuntimeApiSubsystemClient for MockSubsystemClient {
 	async fn disabled_validators(&self, _: Hash) -> Result<Vec<ValidatorIndex>, ApiError> {
 		todo!("Not required for tests")
 	}
+
+	async fn claim_queue(
+		&self,
+		_: Hash,
+	) -> Result<BTreeMap<CoreIndex, VecDeque<ParasEntry<BlockNumber>>>, ApiError> {
```
eskimor, line 294: We should not expose `ParasEntry` here. That is an internal type that should not matter much outside of the scheduler. A `Vec<ParaId>` should be good enough.

polkadot/node/collation-generation/src/tests.rs
```diff
@@ -634,3 +694,252 @@ fn submit_collation_leads_to_distribution() {
...
+		helpers::handle_runtime_calls_on_new_head_activation(
+			&mut virtual_overseer,
+			activated_hash,
+			AsyncBackingParams { max_candidate_depth: 0, allowed_ancestry_len: 0 },
+			cores,
+			runtime_version,
+			claim_queue,
+		)
+		.await;
+
```
eskimor, line 781: Shouldn't we add a read with timeout here to check that the subsystem is really not distributing a collation?

eskimor (approved):

polkadot/node/collation-generation/src/tests.rs
tdimitrov: I was thinking about the same but it is handled already in `test_harness`: https://github.com/paritytech/polkadot-sdk/blob/3c6ebd9e9bfda58f199cba6ec3023e0d12d6b506/polkadot/node/collation-generation/src/tests.rs#L56-L58

polkadot/node/collation-generation/src/lib.rs
tdimitrov: Not relevant anymore

polkadot/node/collation-generation/src/lib.rs
```diff
@@ -239,10 +244,25 @@ async fn handle_new_activations<Context>(
...
+								// read what's in the claim queue for this core
+								fetch_next_scheduled_on_core(
+									claim_queue,
+									CoreIndex(core_idx as u32),
+								)
+							},
+							None => {
+								// Runtime doesn't support claim queue runtime api. Fallback to
+								// `next_up_on_available`
+								occupied_core.next_up_on_available
```
alindima, line 258: next_up_on_available is being populated by the runtime with the same info as the claimqueue, right? So these two values are identical right now.

I assume we are explicitly using the new runtime API to pave the way for max_candidate_depths higher than 1. right?

alindima (commented):
Looks good

polkadot/node/collation-generation/src/lib.rs
```diff
@@ -598,3 +618,37 @@ fn erasure_root(
...
+async fn fetch_claim_queue(
+	sender: &mut impl overseer::CollationGenerationSenderTrait,
+	relay_parent: Hash,
+) -> crate::error::Result<Option<BTreeMap<CoreIndex, VecDeque<ParaId>>>> {
+	if has_required_runtime(
+		sender,
+		relay_parent,
+		RuntimeApiRequest::CLAIM_QUEUE_RUNTIME_REQUIREMENT,
+	)
+	.await
```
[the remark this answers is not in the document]
tdimitrov: I see your point but I think it is more or less the same.

Each runtime call has got a version check before it (done by the runtime api subsystem). So in the current case we have got:
Version check (by `has_required_runtime`) -> Version check (by runtime api subsystem) -> Runtime call

If we call the new api directly on an old runtime and handle the error we will have the same:
Version check (by runtime api subsystem for the 'old call') -> Version check (by runtime api subsystem for the 'new call') -> Runtime call

So in terms of runtime calls - it's the same when we have got old runtime.

For a new runtime we do save one version check but (1) we will remove the version check once everything is released on Polkadot and (2) the runtime calls are cached and never hit the runtime.

Considering these - I'd stick with the current version because it is more explicit.

Another drawback of relying on `NotSupported` for version check is that the former happens in other weird cases - e.g. [a runtime call being executed on pruned blocks](https://github.com/paritytech/polkadot-sdk/issues/793). Having `NotSupported` errors polluting the logs won't be ideal especially when we chase bugs.

polkadot/node/collation-generation/src/lib.rs
tdimitrov: Yes, both are correct.

polkadot/node/collation-generation/src/lib.rs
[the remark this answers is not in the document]
alindima: > For a new runtime we do save one version check but (1) we will remove the version check once everything is released on Polkadot and (2) the runtime calls are cached and never hit the runtime.

yes, that's exactly my point, saving one version check for a new runtime. We'd also save one round-trip to the runtime API subsystem, regardless of the runtime version.

anyway, I don't have very strong feelings about this

polkadot/node/collation-generation/src/lib.rs
```diff
@@ -239,10 +244,25 @@ async fn handle_new_activations<Context>(
 						// TODO [now]: this assumes that next up == current.
 						// in practice we should only set `OccupiedCoreAssumption::Included`
 						// when the candidate occupying the core is also of the same para.
```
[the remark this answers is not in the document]
tdimitrov: I guess you are right. But it is fixed thanks to the updated implementation of [`next_up_on_available`](https://github.com/paritytech/polkadot-sdk/blob/abd3f0c49a472fc4e4b52a814493224222921ffb/polkadot/runtime/parachains/src/scheduler.rs#L503-L509)?
Or I am missing something?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5569954

polkadot/node/collation-generation/src/lib.rs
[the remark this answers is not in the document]
alindima: oh, I think I was wrong actually. I think it should be fixed while implementing https://github.com/paritytech/polkadot-sdk/issues/3582

polkadot/node/collation-generation/src/lib.rs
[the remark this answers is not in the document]
tdimitrov: Your point is valid. I've created an issue for this - https://github.com/paritytech/polkadot-sdk/issues/3756

Issue #3528, opened 2024-02-29
Adjust configuration in subsystem regression tests to Kusama network


AndreiEres:
Subsystem regression tests for availability read and write introduced in https://github.com/paritytech/polkadot-sdk/pull/3311. 

However, their configuration was simply copied from benchmark examples. For better results, we should simulate the Kusama network.

AndreiEres:
Copied from @sandreim wrote here https://github.com/paritytech/polkadot-sdk/pull/3311#discussion_r1508876931

```
let mut config = TestConfiguration::default();
config.n_cores = 20;
```
> This value should be set to the number of block validators are expected to check per relay chain block. This depends on approval voting parameters. We should also factor in async backing, that is we'd expect to have 1 included candidate at every block.
>
> We can calibrate these in the followup PR.

PR #3583, opened 2024-03-05
subsystem-bench: adjust test config to Kusama

AndreiEres:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3528

```rust
latency:
    mean_latency_ms = 30 // common sense
    std_dev = 2.0 // common sense
n_validators = 300 // max number of validators, from chain config
n_cores = 60 // 300/5
max_validators_per_core = 5 // default
min_pov_size = 5120 // max
max_pov_size = 5120 // max
peer_bandwidth = 44040192 // from the Parity's kusama validators
bandwidth = 44040192 // from the Parity's kusama validators
connectivity = 90 // we need to be connected to 90-95% of peers
```

### DIFF
--- a/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
+++ b/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
@@ -39,14 +39,14 @@ fn main() -> Result<(), String> {
 	// TODO: Adjust the test configurations to Kusama values
 	let mut config = TestConfiguration::default();
 	config.latency = Some(PeerLatency { mean_latency_ms: 30, std_dev: 2.0 });
-	config.n_validators = 1000;
-	config.n_cores = 200;
-	config.max_validators_per_core = 5;
-	config.min_pov_size = 5120;
-	config.max_pov_size = 5120;
-	config.peer_bandwidth = 52428800;
-	config.bandwidth = 52428800;
-	config.connectivity = 75;
+	config.n_validators = 300; // max number of validators
+	config.n_cores = 50; // 42 paras producing blocks
+	config.max_validators_per_core = 3; // from config
+	config.min_pov_size = 5120; // https://grafana.teleport.parity.io/goto/Kr34vOASR?orgId=1
+	config.max_pov_size = 5120; // https://grafana.teleport.parity.io/goto/Kr34vOASR?orgId=1
+	config.peer_bandwidth = 52428800; // need to find out
+	config.bandwidth = 52428800; // need to find out
+	config.connectivity = 75; // we need to be connected to 90-95% of peers
 	config.num_blocks = 3;
 	config.generate_pov_sizes();
 
--- a/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
+++ b/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
@@ -16,7 +16,7 @@
 
 //! availability-read regression tests
 //!
-//! Simulated process of availability reading on Kusama
+//! Availability read benchmark based on Kusama parameters and scale.
 //!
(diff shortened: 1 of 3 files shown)


AndreiEres:
image
https://grafana.teleport.parity.io/goto/HQdBDO0Ig?orgId=1

We could've used these values assuming 90 and 99 percentiles
```rust
min_pov_size = 64 // 108 KiB, 90%
max_pov_size = 2024 // 1.78 MiB, 99%
```

AndreiEres:
Based on kusama telemetry:
- average bandwidth = 1555613
- median bandwidth = 752370

I'd use a value close to median:
```rust
peer_bandwidth = 819200 // median bandwidth is 752370
bandwidth = 819200 // median bandwidth is 752370
```

AndreiEres:
> Based on kusama telemetry:
> 
>     * average bandwidth = 1555613
> 
>     * median bandwidth = 752370

Another (and better) option is to use the reference hw recommendation: [65536000](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot#reference-hardware)

polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
```diff
@@ -16,7 +16,7 @@
 
 //! availability-read regression tests
 //!
-//! TODO: Explain the test case after configuration adjusted to Kusama
+//! Simulated process of availability reading on Kusama
```
sandreim, line 19: I'd go for something like `Availability read benchmark based on Kusama parameters and scale.`

polkadot/node/subsystem-bench/src/lib/configuration.rs
```diff
@@ -35,24 +35,39 @@ pub struct PeerLatency {
...
+
+// Based on number of kusama validators producing blocks
+fn default_n_cores() -> usize {
+	50
+}
+
+// PoV size, KiB. Based on empirical data from Kusama
 fn default_pov_size() -> usize {
-	5120
+	2 * 1024
```
sandreim, line 50: Let's use max value here such that any degradation has more impact.

polkadot/node/subsystem-bench/src/lib/configuration.rs
```diff
@@ -35,24 +35,39 @@ pub struct PeerLatency {
...
 fn default_connectivity() -> usize {
-	100
+	90
 }
 
-// Default backing group size
+// Based on Kusama `max_validators_per_core`
 fn default_backing_group_size() -> usize {
-	5
+	3
```
sandreim, line 70: This has to be kept in sync with the actual value from chain. Let's keep it 5 here and override in specific benchmarks.

polkadot/node/subsystem-bench/src/lib/configuration.rs
```diff
@@ -35,24 +35,39 @@ pub struct PeerLatency {
 	pub std_dev: f64,
 }
 
-// Default PoV size in KiB.
+// Based on Kusama `max_validators`
+fn default_n_validators() -> usize {
+	300
+}
+
+// Based on number of kusama validators producing blocks
```
sandreim, line 43: This is not related to the validators producing blocks. It's the total number of cores: cores assigned to specific paras plus on-demand (chain config value)

sandreim: But we don't want to fetch all cores. We'd want to represent a single node effort. Roughly this number has to be set to `n_cores * needed_approvals / n_validators.`

sandreim: `n_cores * needed_approvals / n_validators.`  only so for availability read, for write, we need to cover all cores as said above.

sandreim (approved):

PR #3595, opened 2024-03-06
sc-manual-seal: don't spawn threads in tests

andresilva:
Hopefully helps with test flakiness.

### DIFF
--- a/substrate/client/consensus/manual-seal/src/lib.rs
+++ b/substrate/client/consensus/manual-seal/src/lib.rs
@@ -429,7 +429,9 @@ mod tests {
 					sender,
 				}
 			});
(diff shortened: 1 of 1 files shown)


lexnv (approved):

michalkucharczyk (approved):

michalkucharczyk:
We could run this test 20x (by modyfying _gitlab-ci_ files) in this particular PR just to test if the fix works?
What was the failure-rate?

niklasad1:
@andresilva can you add timeouts the `rx.await` in this PR as well? Just to make it easier debug if it starts failing again? 

I'm not sure why it failed but could that too many threads were created and the task never was executed?

Issue #62, opened 2022-01-04
The road to the native runtime free world


bkchr:
# The Vision

We want to get rid off the native runtime. Initially the native runtime was added to overcome the performance "problems" of using wasmi for the runtime execution. However, with wasmtime we are around 2x slower than native. While this still isn't perfect, it is way faster than wasmi that is around 10x slower (all depending on the work load etc). While the performance plays an important role here there are also other things to consider, like ensuring that the native and wasm runtime are deterministic. We have already run into problems with [Polkadot](https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021/) when it comes to this determinism problems. While we can ensure to use the exact same compiler version for building the Polkadot binary in our release process, we can not ensure that everyone else is also using the same compiler version as we used for the wasm build. 

There is also a nice write up and some discussions here: https://github.com/paritytech/substrate/issues/7288

# The Plan

To remove the native runtime, we need to fix some things before:

1. Make the wasm memory growable. We have already run into a situation of a OOM. This happened because we only allow a fixed amount of memory per wasm instance. We solved this problem temporarily by running the native runtime which can just use the entire system memory and then afterwards we released a new Polkadot node with an increased memory budget. So, we were lucky because we had the native runtime that saved our ass, but if we remove it it would not be there anymore. So, we should support growing of the wasm memory. It can still in the end allocate in maximum 4GB, but that should be way enough for all our use cases and combined with some proper monitoring we should be able to be informed before we hit this upper limit.
2. Remove runtime api requirement of statically checking that the runtime api is implemented. Currently when trying to use a runtime api, it is required to tell the compiler that a certain type is required that needs to implement a runtime api. This was done on purpose to give people better hints at compile time that they don't have implemented a certain api, but this only works by checking the native runtime (because we can not tell the compiler to check the wasm file). So, if we want to remove the native runtime we also need to remove this requirement on having the runtime api being implemented for a type. This can be solved by implementing the runtime apis for a different trait that will then be able to call into the runtime. We will also need to change some stuff, like checking that a runtime api is implemented on startup instead of failing later when the node is running. https://github.com/paritytech/polkadot-sdk/issues/27
3. Genesis is currently being created using the native runtime. We will also need to fix this. Maybe with an extra tool or whatever. https://github.com/paritytech/polkadot-sdk/issues/25

# Open Questions

_If you want to help us out and contribute to this issue, in this section you can find open questions and tasks where we would appreciate any input._

[...]

Issue #5, opened 2023-05-31
Vision for cleaner substrate-node's developer-facing interface


kianenigma:
related to paritytech/substrate#13951 paritytech/substrate#14137 

## Vision 1

As it stands now, we take it for granted that that one builds a runtime, then pulls the substrate-node-template (or some similar repository of code), integrate it in a "blind" fashion (using probably nasty trial and error, without really knowing what they are doing) and never look at the node side code ever again. 

Taking the same mindset that I talked about for a "substrate-node-builder-cli" [here](https://github.com/paritytech/substrate/issues/13951#issuecomment-1564576741), Imagine an alternative like this: 

* You build your pallet and runtime using paritytech/substrate#14137 
* Then, you build a new project like: 

```
[dependencies]
my-runtime = "1.0.0"
substrate-node-builder = "1.0.0"
```

* a `main.rs` that looks like: 

```rust
use substrate_node_builder::{Builder, Consensus, Subcommand};
 
let node = Builder::default()
    // wire your node and runtime. All of the random types that the node needs from the runtime should not 
    // be raw re-exports anymore, and should be done via a `trait`.
    // this should also be helpful 
    .runtime(my_runtime::MyRuntime)
    // define your consensus engine. Can only be one of a few options. 
    .consensus(Consensus::ManualSeal)
    // Tweak the list of sub-commands from `Default`
    .add_subcommand(Subcommand::TryRuntime)
    .add_subcommand(Subcommand::Revert)
    .remove_subcommand(Subcommand::Benchmarks)
    // Tweak the list of RPCs endpoints
    .add_rpc(Rpc::Foo)
    .remove_rpc(Rpc::Bar)
    .build()
    .unwrap();

fn main() {
    node.run();
}
```

This is basically what a "creat-substrate-app" CLI would do for you, but done via code. Once we have this, the CLI would be rather trivial to build, and would merely be something that converts a YAML/JSON file t the above piece of code. 

## Vision 2

I have little clue about what are the blockers to reach the above. Historically I knew that code in `service.rs` has been quite a difficult one to play with. But, looking at a very simple node-template with manual seal, I _think_ I can wrap my head around it and conclude that the above is feasible. 

But, if the above is not possible, what I would be a good improvement to the existing quo, especially from a DevEx perspective is to replace the need to clone any _code_ for the node software with pure _binaries_. 

[...]

PR #3597, opened 2024-03-06
`parachain-omni-node` binary

kianenigma:
Related to #62, https://github.com/paritytech/polkadot-sdk/issues/5, https://github.com/paritytech/polkadot-sdk/issues/1984 and possibly more.

This node now should be able to take any runtime that has no consensus related expectations (ie. data being in header digests etc), and run it with its own manual consensus. It should be ideal for a scenario like a light tutorial. 

But moreover, if the approach here is good, it can already be used to build `solo-chain-omni-node` and `parachain-omni-node` etc.


templates/minimal/node/src/command.rs
```diff
@@ -51,12 +56,34 @@ impl SubstrateCli for Cli {
...
+		let code = std::fs::read(&self.runtime)
+			.map_err(|e| format!("Failed to runtime read {}: {}", &self.runtime, e))?;
+
+		Ok(Box::new(if maybe_path.is_empty() {
+			println!("Using development chain spec, no genesis state set.");
+			let mut properties = Properties::new();
+			properties.insert("tokenDecimals".to_string(), 0.into());
+			properties.insert("tokenSymbol".to_string(), "MINI".into());
+
+			// `.with_genesis_config(Default::default)` won't work, but should.
```
kianenigma, line 72: @michalkucharczyk this was my main chain-spec related issue so far, I would have expected it to work without this. Or perhaps the `Default::default` should do this for me.

templates/minimal/node/src/impl_fake_runtime_apis.rs
```diff
@@ -0,0 +1,98 @@
+#![allow(unused_variables)]
+
+pub struct FakeRuntime;
+use crate::types::{AccountId, Header, Nonce, OpaqueBlock as Block};
+
+sp_api::impl_runtime_apis! {
```
kianenigma, line 6: We can also manually implement this stuff without this macro, but this macro already does the job easily.

templates/minimal/node/src/command.rs
michalkucharczyk: Actually, `.with_genesis_config_patch(Default::default)` should do the job. (which means empty patch for default config).

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5596506

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5596541

paritytech-cicd-pr:
**Job name**: check-tracing
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5596542

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5621470

bkchr:
> parachain-omni-node

`polkadot-parachains` should be this binary. No need to have another special binary for this IMO.

kianenigma:
> > parachain-omni-node
> 
> `polkadot-parachains` should be this binary. No need to have another special binary for this IMO.

Agreed, I didn't know it is already a node capable of this.

Polkadot-Forum:

Polkadot-Forum:

https://forum.polkadot.network/t/rfc-should-we-launch-a-thousand-cores-program/7604/1

Polkadot-Forum:

Polkadot-Forum:

https://forum.polkadot.network/t/stabilizing-polkadot/7175/20

omni-nodes/common/src/fake_runtime.rs
```diff
@@ -0,0 +1,202 @@
...
+	pub(crate) type Nonce = u32;
+	pub(crate) type AuraId = sp_consensus_aura::sr25519::AuthorityId;
+	pub(crate) type BlockNumber = u32;
+	pub(crate) type Balance = u128;
+	pub(crate) type Weight = sp_weights::Weight;
+	pub(crate) type RuntimeCall = ();
+}
+use whatever::*;
+
+sp_api::impl_runtime_apis! {
```
kianenigma, line 30: This should ideally move to `cumulus_service` now. I think one set of fake runtime impls should be enough.

omni-nodes/parachain/src/chain_spec.rs
```diff
@@ -0,0 +1,201 @@
...
+		Extensions {
+			relay_chain: "rococo-local".into(),
+			// You MUST set this to the correct network!
+			para_id: 1000,
+		},
+	)
+	.with_name("Development")
+	.with_id("dev")
+	.with_chain_type(ChainType::Development)
+	.with_genesis_config_patch(testnet_genesis(
```
michalkucharczyk, line 82: Don't know if this is final or temporary proposal. I'd suggest fetching `"testnet"` genesis from runtime here as _omni-node_ shall make no assumptions about runtime.

If _omni-node_ is to provide two "hardcoded" chainspecs (`"development", "local_testnet"`) then we could require runtime's author to provided corresponding presets. If runtime does not provide those presets we could fail.

michalkucharczyk: oh, it is not actually used...

kianenigma:
> What is blocking this (assuming it is blocked)?
> 
> Upgrade the client to the new polkadot-sdk is always a big overhead and error-prune process and for no real benefits. Omni node is the solution to allow parachain teams only focus on the runtime. This is a super critical work and I would like to see it released ASAP.

There should be no blocker for syncing normal parachains as far as I can say. I have synced hydra, interlay and frequncy at https://github.com/paritytech/polkadot-sdk/pull/3597/commits/40fbf47f4c28cc7ea8520815d9b8df59cc3597a0 and it all works. Although, as additional tests, we should also test: 

1. If the DApp of the corresponding parachain can work with the omni-node (missing RPCs)
2. If the node can collate blocks (important, else this is not useful)

The code in this commit already provides some soft degree of configurability through a builder pattern as well. This can either remain a builder or become CLI flags. 

https://github.com/paritytech/polkadot-sdk/blob/40fbf47f4c28cc7ea8520815d9b8df59cc3597a0/omni-nodes/parachain/src/main.rs#L14-L20

xlc:
let’s get a simple version out that only supports the most common scenario and then we can iteratively make it compatible with more chains

kianenigma:
https://github.com/paritytech/polkadot-sdk/pull/4369/ is a pretty similar PR to what is being done here, and can be reviewed and merged as a first step.

kianenigma:
This branch will likely be closed in favor of s slimmer version brought to you by @serban300. I will leave open for visibility for now, but feel free to close when a replacemebt comes by.

PR #3602, opened 2024-03-07
Disable flaky test

athei:
Unfortunately, the flakiness wasn't fixed by https://github.com/paritytech/polkadot-sdk/pull/3595. Let's disable the test in the meanwhile since it is hanging on the CI a lot.


substrate/client/consensus/manual-seal/src/lib.rs
```diff
@@ -467,7 +467,9 @@ mod tests {
 		assert_eq!(client.header(created_block.hash).unwrap().unwrap().number, 1)
 	}
 
-	#[tokio::test]
+	// TODO: enable once the flakiness is fixed
```
liamaharon, line 470: Got a tracking issue to link to?

liamaharon (approved):

substrate/client/consensus/manual-seal/src/lib.rs
athei: https://github.com/paritytech/polkadot-sdk/issues/3603

liamaharon, line 470:
```suggestion
	// TODO: enable once the flakiness is fixed
	// See <https://github.com/paritytech/polkadot-sdk/issues/3603>
```

liamaharon (approved):

bkchr:
Instead of disabling the test, we should enable logging and see where it hangs/fails. Running this test locally doesn't lead to it stalling.

athei:
I think @andresilva is debugging it. The fix can enable the test again.

This makes merging really hard. You need to be lucky two times in a row (you get kicked out of the merge queue).

andresilva (approved):

Issue #690, opened 2023-03-15
Estimate Fees for XCM


xlc:
https://github.com/paritytech/polkadot/blob/543462890784d6cd17b6c6f58b73c94d3b32a65d/xcm/xcm-executor/src/lib.rs#L242-L251

Right now it withdraw the fee from origin account directly. This is just bad.

UI will have hard time to estimate the actual fee.
User have no way to limit the fee amount.

I don't think right now it is possible to estimate fee of a XCM.
We can estimate weight and therefore there is no reason we can't estimate fee. We need some traits and runtime calls to allow frontend to do estimation.

Also I would like to see that fees should only be withdrawn from a fee register, which are prefilled with an instruction to withdraw assets into fee register. This way, user can specify the max fee to avoid unexpected charge.

noandrea:
Hello, with the migration to the polkadot-sdk single project, are there any updates relative to the ability to estimate XCM fees?

xlc:
Don’t nudge me, I am the one asking question and hoping someone else will answer me as well.

bkchr:
CC @franciscoaguirre

acatangiu:
I’m also aware of this behaviour, but am unsure what the best solution would be. Can you elaborate on the desired/ideal UX, and we can work backwards from that to fix/improve fee handling?

gavofyork:
We can take a look at this in the next XCM retreat, but in the meantime if you have any specific proposal for improvement, perhaps you can put it in an RFC.

acatangiu:
What about multi-hop XCMs? For example, going through a reserve chain C for transferring `TokenC` from A to B, or going over a bridge.

> Runtime API to dry run XCM and returns outcome and fees and weights. If all the fees are deducted from fee register, then it is simply calculate how much is deducted from it.

Having this ^ would allow UIs to call into each runtime for every hop to query fees, but they'd be "guessing" what the intermediary XCMs look like. Probably we'd need to provide these intermediaries too as part of the dry-run. As in:
Dry-run this XCM, you get back execution cost/weight, delivery cost to first hop, first hop, and XCM to be sent to first hop, then you repeat for each hop to destination.

But I'm secretly hoping there's a better/smarter way to do it..

xlc:
[...]

PR #3607, opened 2024-03-07
XCM Fee Payment Runtime API

PraetorP:
The PR provides API for obtaining:
- the weight required to execute an XCM message,
- a list of acceptable `AssetId`s for message execution payment,
- the cost of the weight in the specified acceptable `AssetId`.

It is meant to address an issue where one has to guess how much fee to pay for execution. Also, at the moment, a client has to guess which assets are acceptable for fee execution payment.
See the related issue https://github.com/paritytech/polkadot-sdk/issues/690.
With this API, a client is supposed to query the list of the supported asset IDs (in the XCM version format the client understands), weigh the XCM program the client wants to execute and convert the weight into one of the acceptable assets. Note that the client is supposed to know what program will be executed on what chains. However, having a small companion JS library for the pallet-xcm and xtokens should be enough to determine what XCM programs will be executed and where (since these pallets compose a known small set of programs).
```Rust
pub trait XcmPaymentApi<Call>
	where
		Call: Codec,
	{
		/// Returns a list of acceptable payment assets.
		///
		/// # Arguments
		///
		/// * `xcm_version`: Version.
		fn query_acceptable_payment_assets(xcm_version: Version) -> Result<Vec<VersionedAssetId>, Error>;
		/// Returns a weight needed to execute a XCM.
		///
		/// # Arguments
		///
		/// * `message`: `VersionedXcm`.
		fn query_xcm_weight(message: VersionedXcm<Call>) -> Result<Weight, Error>;
		/// Converts a weight into a fee for the specified `AssetId`.
		///
		/// # Arguments
		///
		/// * `weight`: convertible `Weight`.
		/// * `asset`: `VersionedAssetId`.
		fn query_weight_to_asset_fee(weight: Weight, asset: VersionedAssetId) -> Result<u128, Error>;
		/// Get delivery fees for sending a specific `message` to a `destination`.
		/// These always come in a specific asset, defined by the chain.
		///
		/// # Arguments
		/// * `message`: The message that'll be sent, necessary because most delivery fees are based on the
		///   size of the message.
		/// * `destination`: The destination to send the message to. Different destinations may use
		///   different senders that charge different fees.
		fn query_delivery_fees(destination: VersionedLocation, message: VersionedXcm<()>) -> Result<VersionedAssets, Error>;
	}
```
An [example](https://gist.github.com/PraetorP/4bc323ff85401abe253897ba990ec29d) of a client side code.


mrshiposha:
Hi @franciscoaguirre, it is the PR we spoke of yesterday :)

franciscoaguirre:
@mrshiposha Awesome! I'd like to add [this comment](https://github.com/paritytech/polkadot-sdk/issues/690#issuecomment-1961381470) about how we should also have a way of estimating delivery fees.
To better integrate them, I think a single API call should provide given the XCM program, both the execution and delivery fees.
How that looks like depends, right now delivery fees are only paid in a specific asset (at least per router, like a bridge), while execution fees work as you describe, first weight and then acceptable assets to buy that weight.
Probably something like:

```rust
/// Given a message, returns the execution `Weight` and the delivery fees needed to be paid.
/// The execution weight can be converted to fees by using `query_acceptable_payment_assets` and `query_weight_to_asset_fee`.
fn query_xcm_fees(message: VersionedXcm<Call>) -> Result<(Weight, Assets), Error>;
```

Alternatively, we could have it return just `Assets` but that removes the ability to pay in a different fee.
Unless we return a vector of options, but that might be too much work for a single API call.

We could also keep this call for execution weight and have a separate API call for delivery fees.
The thing is, everyone would need to use both in order to estimate fees as accurately as possible, so we might as well just make them one call.

What do you think?

Also, I'm adding @xlc and @acatangiu to this conversation.

mrshiposha:
A reply to https://github.com/paritytech/polkadot-sdk/pull/3607#issuecomment-1983170660.

We haven't considered the delivery fee thing yet, but yeah, it should be covered somehow.
Give us some time to consider it so we can better understand the trade-offs related to the delivery fee.

We opened the PR in its current state to have a basis for further discussion, including the solution for the delivery fee

acatangiu:
> We opened the PR in its current state to have a basis for further discussion, including the solution for the delivery fee

This is great!

> We haven't considered the delivery fee thing yet, but yeah, it should be covered somehow.
Give us some time to consider it so we can better understand the trade-offs related to the delivery fee.

I think the currently proposed `XcmPaymentApi` runtime api is good, like Cisco says, should also have a fn to expose the configured delivery fee mechanism (e.g. [rococo config](https://github.com/paritytech/polkadot-sdk/blob/c16fcc47bc427dd01ed3d1c17d6483bd388bc92f/polkadot/runtime/rococo/src/xcm_config.rs#L106-L114)). IMO it can simply re-expose [fn price_for_delivery()](https://github.com/paritytech/polkadot-sdk/blob/c16fcc47bc427dd01ed3d1c17d6483bd388bc92f/polkadot/runtime/common/src/xcm_sender.rs#L87) trait implementor.

----

For the full fees story (like proposed in https://github.com/paritytech/polkadot-sdk/issues/690#issuecomment-1961381470), and providing users/builders/dapps with a nice UX, we will probably add specialized `query_*` runtime APIs for pallets like `pallet-xcm` and `xtokens`.
E.g. `fn query_fees_for_transfer_assets(assets, dest, beneficiary)` which will make use of underlying `XcmPaymentApi` to provide consolidated fees required, with single call.

franciscoaguirre:
> IMO it can simply re-expose [fn price_for_delivery()](https://github.com/paritytech/polkadot-sdk/blob/c16fcc47bc427dd01ed3d1c17d6483bd388bc92f/polkadot/runtime/common/src/xcm_sender.rs#L87) trait implementor.

I think this is a good first thing to add. But I think we could add an API that given a message, it'll give you the delivery fees needed to execute that message. So it dry-runs it to find all messages that need to be sent and to which destinations, and returns the resulting total fees.

franciscoaguirre:
Clarifying my previous comment. I think we should start simple. There are many complicated things we could do like the thing I described about dry-running, but those would be slow to review and merge. This API looks perfect for what we need right now, which is exposing runtime functions we don't expose currently. After we merge this we can build on top of it.

Basically what Adrian said 😄

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "xcm-payment-runtime-api"
```
franciscoaguirre, line 2: What do you think about `xcm-fee-payment-runtime-api`?

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/README.md
```diff
@@ -0,0 +1,2 @@
+
```
franciscoaguirre, line 1:
```suggestion
```

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
franciscoaguirre, line 2: Since it's its own crate, don't we want to move it out of the `xcm-builder` folder?

polkadot/runtime/westend/src/lib.rs
```diff
@@ -2255,6 +2258,37 @@ sp_api::impl_runtime_apis! {
...
+
+			Ok(WeightToFee::weight_to_fee(&weight))
+		}
+
+		fn query_xcm_weight(message: VersionedXcm<RuntimeCall>) -> Result<Weight, XcmPaymentApiError> {
+			let mut message = message
+				.try_into()
+				.map_err(|_| XcmPaymentApiError::VersionedConversionFailed)?;
+			<xcm_config::XcmConfig as xcm_executor::Config>::Weigher::weight(&mut message)
+				.map_err(|_| XcmPaymentApiError::WeightNotComputable)
```
franciscoaguirre, line 2288: I'd log the error here, just so we don't lose it

franciscoaguirre:
I added the new function exposing delivery fees in https://github.com/UniqueNetwork/polkadot-sdk/pull/3

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
PraetorP: I like the suggested name because it is more explicit. On the other hand, what kind of payments (in the context of XCM) besides fees for delivering/executing messages exists?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5577996

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5578142

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5578971

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5579118

polkadot/runtime/westend/src/lib.rs
PraetorP: logs added

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
franciscoaguirre: That's why I'm suggesting adding "fee" to the name, the only payments associated with XCMs are fees. Right now we have execution and delivery fees, but even if we wanted to add something else we would still call it a fee.

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+	#[codec(index = 0)]
+	Unimplemented,
+	/// Converting a versioned data structure from one version to another failed.
+	#[codec(index = 1)]
+	VersionedConversionFailed,
+	/// XCM message weight calculation failed.
+	#[codec(index = 2)]
+	WeightNotComputable,
+	/// XCM version not able to be handled.
+	UnhandledXcmVersion,
```
franciscoaguirre, line 88:
```suggestion
	/// XCM version not able to be handled.
	#[codec(index = 3)]
	UnhandledXcmVersion,
```

prdoc/pr_3607.prdoc
```diff
@@ -0,0 +1,19 @@
...
+title: "XCM payment API"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      A trait of XCM payment API was added.
+	    API provides functionality for obtaining
+      the weight required to execute an XCM message,
+	    a list of accepted `AssetId` for payment for its
+	    execution and the cost in the specified supported `AssetId`.
```
franciscoaguirre, line 13: I think this API is important enough for people to implement that we want to provide more information in this PRDoc, given it's the first channel we have to communicate these changes to people.
We'll also want to announce this on the forum.

```suggestion
      A runtime API was added for estimating the fees required for XCM execution and delivery.
      This is the basic building block needed for UIs to accurately estimate fees.
      An example implementation is shown in the PR. Ideally it's simple to implement, you only need to call existing parts of your XCM config.
      The API looks like so:
      ```rust
		fn query_acceptable_payment_assets(xcm_version: Version) -> Result<Vec<VersionedAssetId>, Error>;
		fn query_xcm_weight(message: VersionedXcm<Call>) -> Result<Weight, Error>;
		fn query_weight_to_asset_fee(weight: Weight, asset: VersionedAssetId) -> Result<u128, Error>;
		fn query_delivery_fees(destination: VersionedLocation, message: VersionedXcm<()>) -> Result<VersionedAssets, Error>;
      ```
      The first three relate to XCM execution fees, given an XCM, you can query its weight, then which assets are acceptable for buying weight and convert weight to a number of those assets.
      The last one takes in a destination and a message you want to send from the runtime you're executing this on, it will give you the delivery fees.
```

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
PraetorP: When forming the name, we were guided by brevity, so we decided to omit the word `fee` since (at the moment) `fee payment` == `payment`.
As I mentioned above, I think a more complete (explicit) name will still look better.

franciscoaguirre:
I think the PR is in a good state, after addressing all the comments I would merge it and then we can handle more things in new PRs:
- Implementing this API in all runtimes
- Adding emulator tests for this that use the estimated fees
- The high-level API for getting the fees for asset transfers

polkadot/xcm/xcm-builder/xcm-payment-runtime-api/Cargo.toml
PraetorP: Yes, we do. When choosing the location, we were guided by the fact that `pallet-transaction-payment-rpc-runtime-api` is located inside another crate, and decided to follow this style.
What do you say if we move it to a higher level (to the `xcm` folder)?

PraetorP: Crate renamed

franciscoaguirre (approved):
🚀

franciscoaguirre:
@PraetorP Please check the CI errors, [this](https://github.com/UniqueNetwork/polkadot-sdk/pull/5) is an attempt to fix at least one of them

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5638631

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -1799,6 +1804,54 @@ sp_api::impl_runtime_apis! {
...
+				.try_into()
+				.map_err(|_| XcmPaymentApiError::VersionedConversionFailed)?;
+			let message = message
+				.try_into()
+				.map_err(|_| XcmPaymentApiError::VersionedConversionFailed)?;
+			let (_, fees) = xcm_config::XcmRouter::validate(&mut Some(destination), &mut Some(message)).map_err(|error| {
+				log::error!("Error when querying delivery fees: {:?}", error);
+				XcmPaymentApiError::Unroutable
+			})?;
+			Ok(VersionedAssets::from(fees))
```
mrshiposha, line 1851: A client might not understand the returned XCM version. The returned `VersionedAssets` should follow the version of one of the inputs. For instance, if the `destination` and the `message` are both v3, the returned assets should also be in the v3 form.
The current code will always return the form of the latest XCM version, which might break some clients (no one uses this API at the moment, but IMHO, we need a proper way of handling this for future XCM versions).

For the same reason, we have the `xcm_version` parameter in the `query_acceptable_payment_assets` method.

Another question, though: what if the `destination` and the `message` are expressed via different XCM versions? What version should the returned assets follow?

The same goes for Westend.

mrshiposha (changes requested):

polkadot/runtime/rococo/src/lib.rs
acatangiu: > The returned VersionedAssets should follow the version of one of the inputs. For instance, if the destination and the message are both v3, the returned assets should also be in the v3 form.

👍 

> Another question, though: what if the destination and the message are expressed via different XCM versions? What version should the returned assets follow?

use the newest version possible

prdoc/pr_3607.prdoc
```diff
@@ -0,0 +1,26 @@
...
+      A runtime API was added for estimating the fees required for XCM execution and delivery.
+      This is the basic building block needed for UIs to accurately estimate fees.
+      An example implementation is shown in the PR. Ideally it's simple to implement, you only need to call existing parts of your XCM config.
+      The API looks like so:
+      ```rust
+		fn query_acceptable_payment_assets(xcm_version: Version) -> Result<Vec<VersionedAssetId>, Error>;
+		fn query_xcm_weight(message: VersionedXcm<Call>) -> Result<Weight, Error>;
+		fn query_weight_to_asset_fee(weight: Weight, asset: VersionedAssetId) -> Result<u128, Error>;
+		fn query_delivery_fees(destination: VersionedLocation, message: VersionedXcm<()>) -> Result<VersionedAssets, Error>;
+      ```
```
acatangiu, line 18:
```suggestion
      ```rust
            fn query_acceptable_payment_assets(xcm_version: Version) -> Result<Vec<VersionedAssetId>, Error>;
            fn query_xcm_weight(message: VersionedXcm<Call>) -> Result<Weight, Error>;
            fn query_weight_to_asset_fee(weight: Weight, asset: VersionedAssetId) -> Result<u128, Error>;
            fn query_delivery_fees(destination: VersionedLocation, message: VersionedXcm<()>) -> Result<VersionedAssets, Error>;
      ```
```

prdoc/pr_3607.prdoc
```diff
@@ -0,0 +1,26 @@
...
+      The first three relate to XCM execution fees, given an XCM, you can query its weight, then which assets are acceptable for buying weight and convert weight to a number of those assets.
+      The last one takes in a destination and a message you want to send from the runtime you're executing this on, it will give you the delivery fees.
+	
+crates:
+  - name: xcm-fee-payment-runtime-api
+  - name: rococo-runtime
+  - name: westend-runtime
+
```
acatangiu, line 26:
```suggestion

crates:
  - name: xcm-fee-payment-runtime-api
  - name: rococo-runtime
  - name: westend-runtime

```

acatangiu: fixing whitespaces to make prdoc happy

polkadot/runtime/westend/src/lib.rs
```diff
@@ -2255,6 +2258,54 @@ sp_api::impl_runtime_apis! {
 		}
 	}
 
+	impl xcm_fee_payment_runtime_api::XcmPaymentApi<Block, RuntimeCall> for Runtime {
```
franciscoaguirre, line 2261: I'd move all of these implementations to a public function in pallet-xcm.
Then each of these functions can just call that public function.
It simplifies implementation of this API for all future runtimes.

mrshiposha: I added the implementations of the `query_weight_to_asset_fee` and the `query_delivery_fees` as public functions to pallet-xcm, as these functions will most certainly be implemented the same way in most chains.

However, I left the other two methods as is since different chains use different approaches to handling fee assets. It seems sensible to leave them to be implemented manually by a chain's team.

polkadot/runtime/rococo/src/lib.rs
mrshiposha: Done. The implementation is located in pallet-xcm, as per the comment below.
I added the `IdentifyVersion` implementation for `VersionedXcm` to implement the "use the newest version between them" logic

acatangiu:
bot fmt

polkadot/xcm/xcm-fee-payment-runtime-api/README.md
```diff
@@ -0,0 +1 @@
+License: Apache-2.0
```
acatangiu, line 1: I think we could just drop this "empty" README.md file

all source files are already specifying GPL3 licence

acatangiu (approved):

polkadot/xcm/src/lib.rs
```diff
@@ -443,6 +443,16 @@ impl<C> IntoVersion for VersionedXcm<C> {
 	}
 }
 
+impl<C> IdentifyVersion for VersionedXcm<C> {
```
franciscoaguirre, line 446: Hmm, this is already implemented for types created with `versioned_type`. For some reason `VersionedXcm` is not. Making a note here to investigate why in a later issue/pr.

polkadot/node/service/Cargo.toml
```diff
@@ -206,6 +210,7 @@ runtime-benchmarks = [
 	"service/runtime-benchmarks",
 	"sp-runtime/runtime-benchmarks",
 	"westend-runtime?/runtime-benchmarks",
+	"xcm-builder/runtime-benchmarks",
```
franciscoaguirre, line 213: This feature doesn't exist in `xcm-builder`, It was my mistake I think 😅

acatangiu, line 213:
```suggestion
```

acatangiu (approved):

polkadot/node/service/Cargo.toml
```diff
@@ -142,6 +142,10 @@ polkadot-node-core-pvf-checker = { path = "../core/pvf-checker", optional = true
 polkadot-node-core-runtime-api = { path = "../core/runtime-api", optional = true }
 polkadot-statement-distribution = { path = "../network/statement-distribution", optional = true }
 
+xcm = { package = "staging-xcm", path = "../../xcm", default-features = false }
+xcm-builder = { package = "staging-xcm-builder", path = "../../xcm/xcm-builder", default-features = false }
+xcm-fee-payment-runtime-api = { path = "../../xcm/xcm-fee-payment-runtime-api" }
```
franciscoaguirre, line 147:
```suggestion
xcm = { package = "staging-xcm", path = "../../xcm" }
xcm-builder = { package = "staging-xcm-builder", path = "../../xcm/xcm-builder" }
xcm-fee-payment-runtime-api = { path = "../../xcm/xcm-fee-payment-runtime-api" }
```

We need the default `std` feature here

franciscoaguirre:
Please take a look at https://github.com/UniqueNetwork/polkadot-sdk/pull/6

mrshiposha:
@franciscoaguirre Oops, I saw the suggestions before I noticed your comment with the link to another PR :(
I accepted one of them, so the auto-merge is disabled

franciscoaguirre:
@mrshiposha No worries, accepting the other suggestion should pass CI, I'll run the auto-merge again

polkadot/node/service/Cargo.toml
```diff
@@ -219,10 +223,7 @@ try-runtime = [
 	"sp-runtime/try-runtime",
 	"westend-runtime?/try-runtime",
 ]
```
franciscoaguirre, line 225: We do need to add `"xcm-builder/runtime-benchmarks"` here 🤦 
I don't know why that was failing on my machine

mrshiposha: Let me examine the dependencies. I think we added some unneeded ones.

franciscoaguirre: @mrshiposha

mrshiposha: We don't need the `xcm-builder` here at all. We use only the `xcm` to get the XCM types.

franciscoaguirre: Even better then

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5664600

franciscoaguirre:
@mrshiposha you need to update the lockfile

Polkadot-Forum:

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-technical-summit-xcm/9384/1

PR #3608, opened 2024-03-07
Deprecate xcm::body::TREASURER_INDEX constant, use standard Treasury varian instead

muharem:
Deprecate the `xcm::body::TREASURER_INDEX` constant and use the standard Treasury variant from the `xcm::BodyId` type instead.

To align with the production runtimes: https://github.com/polkadot-fellows/runtimes/pull/149

### DIFF
--- a/polkadot/runtime/westend/constants/src/lib.rs
+++ b/polkadot/runtime/westend/constants/src/lib.rs
@@ -130,6 +130,8 @@ pub mod xcm {
 		const ROOT_INDEX: u32 = 0;
 		// The bodies corresponding to the Polkadot OpenGov Origins.
 		pub const FELLOWSHIP_ADMIN_INDEX: u32 = 1;
+		#[deprecated = "Will be removed after August 2024; Use `xcm::latest::BodyId::Treasury` \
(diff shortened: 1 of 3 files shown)


polkadot/runtime/westend/constants/src/lib.rs
```diff
@@ -130,6 +130,8 @@ pub mod xcm {
 		const ROOT_INDEX: u32 = 0;
 		// The bodies corresponding to the Polkadot OpenGov Origins.
 		pub const FELLOWSHIP_ADMIN_INDEX: u32 = 1;
+		#[deprecated = "Will be removed after August 2024; Use `xcm::latest::BodyId::Treasury` \
```
bkontur, line 133: why not just remove it?

bkontur (approved):

polkadot/runtime/westend/constants/src/lib.rs
muharem: to make sure none uses the same index for the time of deprecation

georgepisaltu (approved):

PR #3612, opened 2024-03-07
refund non-used weight in xcm with respect to require_weight_at_most

girazoki:
`Transact` was refunding only the difference between `weight` and `actual_weight` in the xcm-executor, which means it was only refunding if the extrinsic consumed less than the weight-hint. However a more fair approach would be that it  refunds the difference between `require_weight_at_most` and `actual_weight` since this is the difference between what the user provided as `max_weight` to spend in the Transact minus the `actual_weight` finally used

### DIFF
--- a/polkadot/xcm/xcm-builder/src/tests/transacting.rs
+++ b/polkadot/xcm/xcm-builder/src/tests/transacting.rs
@@ -323,3 +323,84 @@ fn clear_transact_status_should_work() {
 	let expected_hash = fake_message_hash(&expected_msg);
 	assert_eq!(sent_xcm(), vec![(Parent.into(), expected_msg, expected_hash)]);
 }
+
+#[test]
+fn paid_transacting_should_refund_payment_for_unused_weight_with_respect_to_require_weight_at_most()
+{
+	let one: Location = AccountIndex64 { index: 1, network: None }.into();
+	AllowPaidFrom::set(vec![one.clone()]);
+	add_asset(AccountIndex64 { index: 1, network: None }, (Parent, 300u128));
+	WeightPrice::set((Parent.into(), 1_000_000_000_000, 1024 * 1024));
+
+	let origin = one.clone();
+	let fees = (Parent, 300u128).into();
+	let message = Xcm::<TestCall>(vec![
+		WithdrawAsset((Parent, 300u128).into()), // enough for 150 units of weight.
+		BuyExecution { fees, weight_limit: Limited(Weight::from_parts(150, 150)) },
+		Transact {
+			origin_kind: OriginKind::Native,
+			// call should at most take 50, but we set 100 to overprice
+			require_weight_at_most: Weight::from_parts(100, 100),
+			// call estimated at 50 but only takes 10.
+			// 90 should be refunded
+			call: TestCall::Any(Weight::from_parts(50, 50), Some(Weight::from_parts(10, 10)))
+				.encode()
+				.into(),
+		},
+		RefundSurplus,
+		DepositAsset { assets: AllCounted(1).into(), beneficiary: one },
+	]);
+	let mut hash = fake_message_hash(&message);
+	let weight_limit = Weight::from_parts(150, 150);
+	let r = XcmExecutor::<TestConfig>::prepare_and_execute(
+		origin,
+		message,
+		&mut hash,
+		weight_limit,
+		Weight::zero(),
+	);
+	assert_eq!(r, Outcome::Complete { used: Weight::from_parts(60, 60) });
+	// The refunded amount should be equal to the 150
+	// provided minus the used 60. 90 in total
+	assert_eq!(
+		asset_list(AccountIndex64 { index: 1, network: None }),
+		vec![(Parent, 180u128).into()]
+	);
+}
+
+#[test]
+fn unpaid_transact_require_weight_at_most_refund_respected() {
+	AllowUnpaidFrom::set(vec![Parent.into()]);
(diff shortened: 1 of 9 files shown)


franciscoaguirre (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5468790

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5468789

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -755,7 +755,7 @@ impl<Config: config::Config> XcmExecutor<Config> {
 						},
 					};
 				let actual_weight = maybe_actual_weight.unwrap_or(weight);
-				let surplus = weight.saturating_sub(actual_weight);
+				let surplus = require_weight_at_most.saturating_sub(actual_weight);
```
bkontur, line 758: I'm not entirely sure about this, and I'm just thinking out loud. From what I understand, it depends on the implementation details of [`FixedWeightBounds`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/weight.rs#L62-L66) and/or [`WeightInfoBounds`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/weight.rs#L108-L112), where `require_weight_at_most` is added to the instruction's weighed weight in the `fn instr_weight_with_limit`. It might also depend on the fact that `AllowTopLevelPaidExecutionFrom` utilizes the total calculated weights for `BuyExecution` [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/barriers.rs#L100).

In other words, if the runtime utilizes `FixedWeightBounds / WeightInfoBounds` and `AllowTopLevelPaidExecutionFrom`, then this fix should probably be ok and works. However, what about other runtimes that use different implementations without incorporating `require_weight_at_most` to the total instruction weight for `Transact`? Will this fix still be effective? Will it potentially refund less or more than expected?

I can see several tests for transacting here: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/tests/transacting.rs, but I think we should ensure coverage for the scenario described in the prdoc: "a very large `require_weight_at_most` then this would be non-refunded", both for the PAID and UNPAID methods in those tests. Currently, they all use `require_weight_at_most: Weight::from_parts(50, 50)`, which seems expected and pretty fine.

girazoki: Runtimes that do not utilize `require_weight_at_most` for the calculation of the total weight of the message should probably be more concerned about  potential DOS no? how else are you going to make sure the weight being bought is sufficient to cover for the execution?

girazoki: With respect to the tests, yes I can add them.

franciscoaguirre: I agree if you don't count `require_weight_at_most` then here you'd be refunding more than was paid for.
However, the most common way is charging for `require_weight_at_most` and I think most people's mental model is you get refunded this difference (`require_weight_at_most` - `actual_weight`) instead of the one that was being actually used.
It's also the situation described a little further in the comment.

If we don't want to do this we can leave it as is, but we should change the comment, communicate that you might not be refunded if you input a very large value in `require_weight_at_most` and make sure users try to get as close to the extrinsic `weight` as possible.

polkadot/xcm/xcm-builder/src/tests/transacting.rs
```diff
@@ -323,3 +323,84 @@ fn clear_transact_status_should_work() {
...
+		&mut hash,
+		weight_limit,
+		Weight::zero(),
+	);
+	assert_eq!(r, Outcome::Complete { used: Weight::from_parts(60, 60) });
+	// The refunded amount should be equal to the 150
+	// provided minus the used 60. 90 in total
+	assert_eq!(
+		asset_list(AccountIndex64 { index: 1, network: None }),
+		vec![(Parent, 180u128).into()]
```
bkontur, line 367: I am a little bit confused, why 180 when comment above says 90? And also "150 provided" shouldn't be "300 provided" Are those numbers units or weights?
My guess is that 150 could related to the `TestConifg::Weigher::weight(message)` = `5instructions` x `UnitWeightCost(10, 10)` + `require_weight_at_most: Weight::from_parts(100, 100)`, that could be `Weight(150, 150)`.


The user starts with 300 units, then `WithdrawAsset(300)` and ends with 180? iiuc, it means that we withdrawn 300units and refunded/returned_back 180units, so 120units was charged for execution, right? Or am I missing something?

polkadot/xcm/xcm-builder/src/tests/transacting.rs
```diff
@@ -323,3 +323,84 @@ fn clear_transact_status_should_work() {
...
+		message,
+		&mut hash,
+		weight_limit,
+		Weight::zero(),
+	);
+	assert_eq!(r, Outcome::Complete { used: Weight::from_parts(40, 40) });
+	// The address is not refunded, even though we set a refund
+	// This is because we never set a BuyExecution for starters
+	// This means that there is no weight from which the refund can
+	// be substracted in the trader
```
bkontur, line 404: nice! :+1:

bkchr:
@girazoki do you still want to merge this?

PR #3616, opened 2024-03-07
Benchmarking: Add pov_mode to V2 syntax

ggwpez:
Changes:
- Port the `pov_mode` attribute from the V1 syntax to V2
- Update `pallet-whitelist` and `frame-benchmarking-pallet-pov`

Follow up: also allow this attribute on top-level benchmark modules.


ggwpez:
bot bench substrate-pallet --pallet=frame_benchmarking_pallet_pov

ggwpez:
bot bench substrate-pallet --pallet=pallet_whitelist

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -72,51 +76,158 @@ struct RangeArgs {
...
+			PovModeAttr::parse(input).map(BenchmarkAttr::PoV)
+		} else {
+			Err(lookahead.error())
+		}
+	}
+}
+
+/// A `#[pov_mode = .. { .. }]` attribute.
+#[derive(Debug, Clone)]
+struct PovModeAttr {
```
pgherveou, line 135: nit I would move the definition before the impl, I find it easier to scroll through the file this way

substrate/frame/support/procedural/src/benchmark.rs
```diff
@@ -72,51 +76,158 @@ struct RangeArgs {
...
+		let _eq: Token![=] = input.parse()?;
+		let root = PovEstimationMode::parse(input)?;
+
+		let mut maybe_content = None;
+		let _ = || -> Result<()> {
+			let content;
+			syn::braced!(content in input);
+			maybe_content = Some(content);
+			Ok(())
+		}();
```
pgherveou, line 102: why do you need to return a Result here?

pgherveou (approved):

substrate/frame/support/procedural/src/benchmark.rs
ggwpez: The `syn::braced!` macro just does a return 🙈  
Its internally using `parse_braces`, but that is private IIUC.

gupnik (approved):
Could we add some UI tests?

ggwpez:
bot update-ui

PR #3606, opened 2024-03-07
Contracts: Remove unstable on lock/unlock_delegate_dependency host fns

pgherveou:
Oversight from PR https://github.com/paritytech/polkadot-sdk/pull/3384.
These 2 functions should have been tagged as stable

PR #3619, opened 2024-03-08
Contracts Bump ApiVersion and add test

pgherveou:
ApiVersion should have been bumped with https://github.com/paritytech/polkadot-sdk/pull/3606
this does that and add a test so we don't forget to do that everytime

### DIFF
--- a/substrate/frame/contracts/proc-macro/src/lib.rs
+++ b/substrate/frame/contracts/proc-macro/src/lib.rs
@@ -497,13 +497,13 @@ fn expand_docs(def: &EnvDef) -> TokenStream2 {
 fn expand_env(def: &EnvDef, docs: bool) -> TokenStream2 {
 	let impls = expand_impls(def);
 	let docs = docs.then_some(expand_docs(def)).unwrap_or(TokenStream2::new());
-	let stable_count = def.host_funcs.iter().map(|f| if f.is_stable { 1 } else { 0 }).sum::<u16>();
+	let stable_api_count = def.host_funcs.iter().filter(|f| f.is_stable).count();
 
 	quote! {
 		pub struct Env;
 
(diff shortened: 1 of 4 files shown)


substrate/frame/contracts/src/lib.rs
```diff
@@ -222,10 +222,19 @@ pub struct Environment<T: Config> {
...
+		Self(2)
 	}
 }
 
+#[test]
+fn api_version_is_up_to_date() {
+	assert_eq!(
+		107,
+		crate::wasm::STABLE_API_COUNT,
+		"Stable API count has changed.Bump ApiVersion::default() and update the test."
```
pgherveou, line 234:
```suggestion
		"Stable API count has changed. Bump the returned value of ApiVersion::default() and update the test."
```

ascjones (approved):

substrate/frame/contracts/proc-macro/src/lib.rs
```diff
@@ -497,9 +497,14 @@ fn expand_docs(def: &EnvDef) -> TokenStream2 {
 fn expand_env(def: &EnvDef, docs: bool) -> TokenStream2 {
 	let impls = expand_impls(def);
 	let docs = docs.then_some(expand_docs(def)).unwrap_or(TokenStream2::new());
+	let stable_count = def.host_funcs.iter().map(|f| if f.is_stable { 1 } else { 0 }).sum::<u16>();
```
ascjones, line 500: Interesting way to do it instead of `filter(|f| f.is_stable()).count()` :grin:

pgherveou: ahah right that looks better

cmichi (approved):

PR #3623, opened 2024-03-08
`TransactionExtension` follow up refactoring and fixes

georgepisaltu:
This PR addresses various comments in #2280 

Key changes:
- outdated comments were updated
- metadata IR had `additonal_signed` renamed to `implicit`
- bumped extrinsic format version
- fixed `pre_dispatch_unsigned` default impl
- renamed `OriginTrait::as_system_signer` to `as_signer`
- removed `BenchmarkHelperTrait` from `claims`

### DIFF
--- a/docs/sdk/src/reference_docs/transaction_extensions.rs
+++ b/docs/sdk/src/reference_docs/transaction_extensions.rs
@@ -13,7 +13,7 @@ pub mod transaction_extensions_example {
 	use sp_runtime::{
 		impl_tx_ext_default,
 		traits::{Dispatchable, TransactionExtension, TransactionExtensionBase},
-		TransactionValidityError,
+		transaction_validity::TransactionValidityError,
 	};
 
 	// This doesn't actually check anything, but simply allows
--- a/polkadot/runtime/common/src/claims.rs
+++ b/polkadot/runtime/common/src/claims.rs
@@ -174,13 +174,6 @@ pub mod pallet {
 	#[pallet::without_storage_info]
 	pub struct Pallet<T>(_);
 
-	#[cfg(feature = "runtime-benchmarks")]
-	/// Helper trait to benchmark the `PrevalidateAttests` transaction extension.
-	pub trait BenchmarkHelperTrait<RuntimeCall, DispatchInfo> {
-		/// `Call` to be used when benchmarking the transaction extension.
-		fn default_call_and_info() -> (RuntimeCall, DispatchInfo);
-	}
-
 	/// Configuration trait.
 	#[pallet::config]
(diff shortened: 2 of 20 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5472632

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5473348

bkchr:
Thank you for doing this. However, I think we first need to revert your transaction extension pull request. 

This format version change needs to go through some RFC. I mean it isn't such a big change, but still should be done in a RFC and can be done quite quickly. 

Then we need to inform anyone in the ecosystem of the change of the extrinsic format. Pinging @TarikGul for polkadot-js and @carlos for ledger. Maybe @josepot should also be pinged. @SBalaguer please communicate with whoever I'm missing here.

xlc:
Glad to see I am not the only person to think an RFC is required for this https://github.com/paritytech/polkadot-sdk/pull/2280#issuecomment-1813867577.

I did not look into all the details of this TransactionExtension change so some of my comments may not be relevant:
- If possible, we should avoid unconditionally force a breaking change upon upgrade polkadot-sdk
- We should have some tests in the fellowship runtime to assert encoded extrensic and failure of such test should indicate a tx_version breaking change. Ideally downstream projects should also have such tests. Obviously we can't force them to write those tests but we can have some docs about it.
- This wouldn't be the only change that requires a communication to the whole ecosystem. There should have a process to trigger and complete it.

bkchr:
> * We should have some tests in the fellowship runtime to assert encoded extrensic and failure of such test should indicate a tx_version breaking change.

`tx_version` will get quite useless in the near future any way, when we have support for the merkleized metadata.

xlc:
maybe not exactly the tx_version, but whatever transaction encoding spec that we are required to confirm. ie ensure we are always spec confirming

bkchr:
> This format version change needs to go through some RFC. I mean it isn't such a big change, but still should be done in a RFC and can be done quite quickly.

I already said this in DM, I will say it here again. My actual sentiment is more that it would be nice to have a RFC. Generally the runtime side is nothing that we really spec. However, this is some external facing encoding that would make sense to have a RFC written for, but I could also be convinced from the opposite.

As said, the RFC should only be about changing the interpretation of the one bit and thus, the bump of the version. Thus, it should be a very straightforward RFC.

georgepisaltu:
Closing in favor of https://github.com/paritytech/polkadot-sdk/pull/3685

PR #3630, opened 2024-03-09
Enable mainnet system parachains to use async backing-enabled collator

s0me0ne-unkn0wn:
This is phase 2 of async backing enablement for the system parachains on the production networks. ~~It should be merged after polkadot-fellows/runtimes#228 is enacted. After it is released,~~ all the system parachain collators should be upgraded, and then we can proceed with phase 3, which will enable async backing in the runtimes.

UPDATE: Indeed, we don't need to wait for the runtime upgrade enactions. The lookahead collator handles the transition by itself, so we can upgrade ASAP.

## Scope of changes

Here, we eliminate the dichotomy of having "generic Aura collators" for the production system parachains and "lookahead Aura collators" for the testnet system parachains. Now, all the collators are started as lookahead ones, preserving the logic of transferring from the shell node to Aura-enabled collators for the asset hubs. So, indeed, it simplifies the parachain service logic, which cannot but rejoice.


cumulus/polkadot-parachain/src/service.rs
```diff
@@ -686,85 +683,6 @@ where
 	Ok(BasicQueue::new(verifier, Box::new(block_import), None, &spawner, registry))
 }
 
-/// Start an aura powered parachain node. Some system chains use this.
-pub async fn start_generic_aura_node(
```
bkontur, line 690: @s0me0ne-unkn0wn @seadanda I am not sure about removing this according to the Donal's [comment](https://github.com/paritytech/polkadot-sdk/issues/3944#issuecomment-2039431641):
> start_generic_aura_node and start_asset_hub_node are no longer needed for system parachains when they all shift to async backing in the next release, but if we're going for an omni-node approach then we'd still need start_generic_aura_node and a cli flag to enable async backing. I agree that this is a sensible default.

s0me0ne-unkn0wn: I don't see any sense in keeping the generic collator, considering that the lookahead collator behaves exactly as the generic collator if a runtime doesn't support async backing. Let me know if I'm missing something

seadanda: At the time of writing that I was under the impression that the lookahead collator expected some additional RuntimeApi functions, if that's not the case and parachains without the tooling for async backing can be run with the lookahead collator then it makes the node code even cleaner. Nice

s0me0ne-unkn0wn: @seadanda it took me quite some time as well to understand how [this guard](https://github.com/paritytech/polkadot-sdk/blob/c1063a530e46c4af0a934ca50422b69182255e60/cumulus/client/consensus/aura/src/collators/lookahead.rs#L461-L468) works indeed. Right now we're up to some live testing: paritytech/devops#3261

seadanda: OK that comment on `can_build_upon` makes sense.

alexggh, line 690: Isn't this part of the public API, `start_generic_aura_node` what if other parachains want to bump the crate version but they don't care about upgrading to the lookahead,  wouldn't they get errors  or am I missing something ?

s0me0ne-unkn0wn: Well, I've set a version bump to "major" in the prdoc, so the breaking change is okay. Here, we're effectively deprecating the generic collator in favor of the lookahead one. This makes sense, as the lookahead collator can do everything the generic collator can do, with some additional functionality.

I doubt if anyone really uses `polkadot-parachain` to run any non-Parity parachains, but if they do, it's not that big an effort to switch to the lookahead, basically, it's just calling another function.

So it seems sound to me, but thanks for bringing up the concern, maybe we'll get other opinions on that.

alexggh: I would go with some soft deprecation like this `#[deprecated = "why is deprecated ....."]`, it just makes easier for people to upgrade, it also gives us an entry point to notify them what to do to stop using this function. The alternative is they get a compilation error when they run `cargo  update`

s0me0ne-unkn0wn: But `cargo update` wouldn't update through a major version afaik?

alexggh: >But cargo update wouldn't update through a major version afaik?

That's true, I was trying to just give an example of unrelated mindless update people might do and they might give up because of compilation errors.

Anyways, I don't have data to back up that this might break other peoples flows, however I do think it is cheap to keep it around for a few releases as deprecated and that would be the less risky alternative since I do see how it would hurt us to keep it around.

s0me0ne-unkn0wn: Okay, now I found out that it's not a public interface indeed, it's declared `pub`, but it lives in a private module and is never reexported. Are you still in favor of keeping it for a while? :upside_down_face:

alexggh: If it is not exported outside, not need to keep it we can restore it from git if we need.

s0me0ne-unkn0wn:
A lot of code moved here and there because the litep2p stuff got merged, please have one more look.

bkchr (approved):

seadanda:
@s0me0ne-unkn0wn I was under the impression that there was something pending before this PR could be merged so I've been holding off on approval, but I've just seen in the chat that you're happy for it to be merged. What were you referring to in this comment: https://github.com/paritytech/devops/issues/3261#issuecomment-2047474907
LMK if it no longer applies and I'm happy to approve.

s0me0ne-unkn0wn:
@seadanda it's been addressed separately in #4036 but it's not been merged yet because of a glitching CI test :facepalm: 
Merging it into this one might me a really good idea!

s0me0ne-unkn0wn:
@seadanda done that!

seadanda (approved):
Yea I thought it was that PR. Didn't want this to merge out of order for some reason because of the CI issues. All LGTM 👍

PR #3636, opened 2024-03-10
[pallet_broker] Fix `adapt_price` behaviour at zero

seadanda:
This fixes the behaviour of `Linear` which is the default implementation of the `AdaptPrice` trait in the broker pallet. Previously if cores were offered but not sold in only one sale, the price would be set to zero and due to the logic being purely multiplicative, the price would stay at 0 indefinitely.

This could be further paired with a configurable minimum in the broker pallet itself, which will be a future PR.

This affects the Rococo and Westend Coretime chains, but Kusama has a different implementation so this isn't required for the Kusama launch. I actually thought I opened this a while ago.


prdoc/pr_3636.prdoc
```diff
@@ -0,0 +1,15 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: "[pallet_broker] Fix `adapt_price` behaviour at zero"
```
bkchr, line 4:
```suggestion
title: "[pallet_broker] Fix `Linear::adapt_price` behavior at zero"
```

substrate/frame/broker/src/adapt_price.rs
```diff
@@ -49,14 +50,24 @@ impl AdaptPrice for () {
...
-		FixedU64::from(2) - when
+		FixedU64::from(2).saturating_sub(when)
 	}
 	fn adapt_price(sold: CoreIndex, target: CoreIndex, limit: CoreIndex) -> FixedU64 {
 		if sold <= target {
-			FixedU64::from_rational(sold.into(), target.into())
+			// Range of [0.5, 1.0].
+			FixedU64::from_rational(1, 2).saturating_add(FixedU64::from_rational(
+				(sold).into(),
+				(target.saturating_mul(2)).into(),
```
bkchr, line 60:
```suggestion
				sold.into(),
				target.saturating_mul(2).into(),
```

substrate/frame/broker/src/adapt_price.rs
```diff
@@ -49,14 +50,24 @@ impl AdaptPrice for () {
...
+			))
 		} else {
-			FixedU64::one() +
-				FixedU64::from_rational((sold - target).into(), (limit - target).into())
+			// Range of (1.0, 2].
+
+			// Unchecked math: In this branch we know that sold < target. The limit must be >= sold
```
bkchr, line 65:
```suggestion
			// Unchecked math: In this branch we know that sold > target. The limit must be >= sold
```

substrate/frame/broker/src/adapt_price.rs
```diff
@@ -49,14 +50,24 @@ impl AdaptPrice for () {
...
+			// by construction, and thus target must be > limit.
```
bkchr, line 66:
```suggestion
			// by construction, and thus target must be < limit.
```

bkchr (approved):
Maybe some tests would be nice to cover some edge cases.

substrate/frame/broker/src/adapt_price.rs
seadanda: 🙈

eskimor (approved):

Issue #3245, opened 2024-02-07
[Staking] `check_payees` try-state check failing in Westend


gpestana:
The [`check_payees`](https://github.com/paritytech/polkadot-sdk/blob/9e6298e7f773c61d384677e6eebfa8d7eee9d2ac/substrate/frame/staking/src/pallet/impls.rs#L1835-L1850) try-state check in Staking is failing in Westend. Figure out what is the reason and a 

> [2024-02-07T16:13:20Z ERROR runtime::frame-support] ❌ "Staking" try_state checks failed: Other("number of entries in payee storage items does not match the number of bonded ledgers")

Example CI job error: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5142158#L2515

gpestana:
It seems that a staking ledger has been removed without clearing up the bonded and payee entry.

```
[2024-02-07T22:18:11Z INFO  runtime::staking] [19451704] 💸  count Ledger 72560, count Payee: 72561, count Bonded: 72561
```

gpestana:
### Solution

Reap the does not work, as it fails with `staking.NotController` since the staking ledger does not exist anymore in storage. This will require a small migration which I can work on.

- [ ] migration to remove the lingering Bonded and Payee entry for stash `5HHaaUvCwAb16KkKy7cpMnuzUgLWH94gEvMVXugc69ZEDfkj`

gpestana:
The try-state is failing again, now with 16 accounts that are faulty. A recent [`deprecate_controller_batch`](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestend-rpc.polkadot.io#/explorer/query/0xd03b07170d16bbcbda81959f3a5f6b47ca84361f17837b22807140be43ad379b) seems to have been the culprit here. The stashes affected seem to also have bonded in the same timespan as the last stash that has been fixed. 

Looking again into this and will check if this may happen to other bonded stashes (also in Kusama and Polkadot).

The solution should be the same as described above, for all the affected stashes.

PR #1484, opened 2023-09-11
Refactor staking ledger

gpestana:
Moves https://github.com/paritytech/substrate/pull/14582 into `polkadot-sdk`

---

This PR refactors the staking ledger logic to encapsulate all reads and mutations of `Ledger`, `Bonded`, `Payee` and stake locks within the `StakingLedger` struct implementation.

With these changes, all the reads and mutations to the `Ledger`, `Payee` and `Bonded` storage map should be done through the methods exposed by StakingLedger to ensure the data and lock consistency of the operations. The new introduced methods that mutate and read Ledger are:

- `ledger.update()`: inserts/updates a staking ledger in storage; updates staking locks accordingly (and ledger.bond(), which is synthatic sugar for ledger.update())
- `ledger.kill()`: removes all Bonded and StakingLedger related data for a given ledger; updates staking locks accordingly;
`StakingLedger::get(account)`: queries both the `Bonded` and `Ledger` storages and returns a `Option<StakingLedger>`. The pallet impl exposes fn ledger(account) as synthatic sugar for `StakingLedger::get(account)`.

Retrieving a ledger with `StakingLedger::get()` can be done by providing either a stash or controller account. The input must be wrapped in a `StakingAccount` variant (Stash or Controller) which is treated accordingly. This simplifies the caller API but will eventually be deprecated once we completely get rid of the controller account in staking. However, this refactor will help with the work necessary when completely removing the controller.

Other goals:

- No logical changes have been introduced in this PR;
- No breaking changes or updates in wallets required;
- No new storage items or need to perform storage migrations;
- Centralise the changes to bonds and ledger updates to simplify the OnStakingUpdate updates to the target list (related to https://github.com/paritytech/polkadot-sdk/issues/443)

Note: it would be great to prevent or at least raise a warning if `Ledger<T>`, `Payee<T>` and `Bonded<T>` storage types are accessed outside the `StakingLedger` implementation. This PR should not get blocked by that feature, but there's a tracking issue here https://github.com/paritytech/polkadot-sdk/issues/149

Related and step towards https://github.com/paritytech/polkadot-sdk/issues/443

PR #3639, opened 2024-03-11
Staking ledger bonding fixes

gpestana:
Currently, the staking logic does not prevent a controller from becoming a stash of *another* ledger (introduced by [removing this check](https://github.com/paritytech/polkadot-sdk/pull/1484/files#diff-3aa6ceab5aa4e0ab2ed73a7245e0f5b42e0832d8ca5b1ed85d7b2a52fb196524L850)). Given that the remaining of the code expects that never happens, bonding a ledger with a stash that is a controller of another ledger may lead to data inconsistencies and data losses in bonded ledgers. For more detailed explanation of this issue: https://hackmd.io/@gpestana/HJoBm2tqo/%2FTPdi28H7Qc2mNUqLSMn15w

In a nutshell, when fetching a ledger with a given controller, we may be end up getting the wrong ledger which can lead to unexpected ledger states. 

This PR also ensures that `set_controller` does not lead to data inconsistencies in the staking ledger and bonded storage in the case when a controller of a stash is a stash of *another* ledger. and improves the staking `try-runtime` checks to catch potential issues with the storage preemptively.

In summary, there are two important cases here:

1. **"Sane" double bonded ledger**

When a controller of a ledger is a stash of *another* ledger. In this case, we have:

```
> Bonded(stash, controller)
(A, B)  // stash A with controller B
(B, C) // B is also a stash of another ledger
(C, D)

> Ledger(controller)
Ledger(B) = L_a (stash = A)
Ledger(C) = L_b (stash = B)
Ledger(D) = L_c (stash = C)
```

In this case, the ledgers can be mutated and all operations are OK. However, we should not allow `set_controller` to be called if it means it results in a "corrupt" double bonded ledger (see below). 

3. **"Corrupt" double bonded ledger**

```
> Bonded(stash, controller)
(A, B)  // stash A with controller B
(B, B)
(C, D)
```
In this case, B is a stash and controller AND is corrupted, since B is responsible for 2 ledgers which is not correct and will lead to inconsistent states. Thus, in this case, in this PR we are preventing these ledgers from mutating (i.e. operations like bonding extra etc) until the ledger is brought back to a consistent state.

--- 

**Changes**:
- Checks if stash is already a controller when calling `Call::bond` (fixes the regression introduced by [removing this check](https://github.com/paritytech/polkadot-sdk/pull/1484/files#diff-3aa6ceab5aa4e0ab2ed73a7245e0f5b42e0832d8ca5b1ed85d7b2a52fb196524L850));
- Ensures that all fetching ledgers from storage are done through the `StakingLedger` API;
- Ensures that -- when fetching a ledger from storage using the `StakingLedger` API --, a `Error::BadState` is returned if the ledger bonding is in a bad state. This prevents bad ledgers from mutating (e.g. `bond_extra`, `set_controller`, etc) its state and avoid further data inconsistencies. 
- Prevents stashes which are controllers or another ledger from calling `set_controller`, since that may lead to a bad state.
- Adds further try-state runtime checks that check if there are ledgers in a bad state based on their bonded metadata.

Related to https://github.com/paritytech/polkadot-sdk/issues/3245


substrate/frame/staking/src/ledger.rs
```diff
@@ -201,6 +201,32 @@ impl<T: Config> StakingLedger<T> {
...
+            .ok_or(Error::<T>::NotController)?;
+
+		ensure!(self.stash != *controller, Error::<T>::AlreadyPaired);
+
+		match Ledger::<T>::get(&self.stash) {
+			Some(other_ledger) => {
+				if other_ledger.stash != self.stash {
+					// stash is a controller of another ledger, fail with `Error::<T>::DoubleBonded`
+					// to avoid data inconsistencies.
+					return Err(Error::<T>::DoubleBonded);
```
Ank4n, line 217: Would probably call something different or just a BadState error.

substrate/frame/staking/src/ledger.rs
```diff
@@ -201,6 +201,32 @@ impl<T: Config> StakingLedger<T> {
...
 
+	/// Sets the ledger controller to its stash.
+	pub(crate) fn set_controller_to_stash(self) -> Result<(), Error<T>> {
+		let controller = self.controller.as_ref()
+            .defensive_proof("Ledger's controller field didn't exist. The controller should have been fetched using StakingLedger.")
+            .ok_or(Error::<T>::NotController)?;
+
+		ensure!(self.stash != *controller, Error::<T>::AlreadyPaired);
+
+		match Ledger::<T>::get(&self.stash) {
```
Ank4n, line 212: Could use `if let Some()` syntax

Ank4n: or actually if the key (stash) just exists in ledger that is probably enough to error. The last check `self.stash != controller` would cover the below conditions.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1844,9 +1845,58 @@ impl<T: Config> Pallet<T> {
 		Self::check_count()
 	}
 
+	/// Invariants:
+	/// * A bonded (stash, controller) pair should have only one associated ledger. I.e. if the
```
Ank4n, line 1849: To me it would be more clear if it says something along the lines of: 
A stash should not be controller of another ledger and viceversa.

Ank4n (approved):
Pretty important fix 🙏 .

The whole stash/controller logic is not so great. Intuitively I would have not expected `Ledger` to be keyed by controller which seems the core issue behind this bug. Another observation is, it is not very clear why some extrinsics in staking pallet can be called via stash and others via controller.

substrate/frame/staking/src/ledger.rs
```diff
@@ -201,6 +201,32 @@ impl<T: Config> StakingLedger<T> {
...
+			Some(other_ledger) => {
+				if other_ledger.stash != self.stash {
+					// stash is a controller of another ledger, fail with `Error::<T>::DoubleBonded`
+					// to avoid data inconsistencies.
+					return Err(Error::<T>::DoubleBonded);
+				}
```
georgepisaltu, line 218:
```suggestion
				// stash must not be a controller of another ledger, otherwise fail with
				// `Error::<T>::DoubleBonded` to avoid data inconsistencies.
				ensure!(other_ledger.stash == self.stash, Error::<T>::DoubleBonded);
```

georgepisaltu (approved):

substrate/frame/staking/src/mock.rs
```diff
@@ -786,6 +786,58 @@ pub(crate) fn bond_controller_stash(controller: AccountId, stash: AccountId) ->
...
 
+pub(crate) fn setup_double_bonded_ledgers() {
+	assert_ok!(Staking::bond(RuntimeOrigin::signed(1), 10, RewardDestination::Staked));
+	assert_ok!(Staking::bond(RuntimeOrigin::signed(2), 20, RewardDestination::Staked));
+	assert_ok!(Staking::bond(RuntimeOrigin::signed(3), 20, RewardDestination::Staked));
+	// not relevant to the test case, but ensures try-runtime checks pass.
+	let _ = [1, 2, 3]
+		.iter()
+		.map(|s| Payee::<Test>::insert(s, RewardDestination::Staked))
+		.collect::<Vec<_>>();
```
kianenigma, line 797:
```suggestion
	[1, 2, 3]
		.iter()
		.for_each(|s| Payee::<Test>::insert(s, RewardDestination::Staked));
```

substrate/frame/staking/src/ledger.rs
gpestana: Done!

substrate/frame/staking/src/ledger.rs
gpestana: done!

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -975,9 +982,11 @@ pub mod pallet {
 			#[pallet::compact] max_additional: BalanceOf<T>,
 		) -> DispatchResult {
 			let stash = ensure_signed(origin)?;
-
 			let mut ledger = Self::ledger(StakingAccount::Stash(stash.clone()))?;
 
+			// return early if ledger is in a bad state (ledger's stash != expected).
+			ensure!(ledger.stash == stash, Error::<T>::BadState);
```
Ank4n, line 988: Ideally move this check inside `fn ledger`?

gpestana: Done!

EgorPopelyaev:
@gpestana just for info: `prdoc` file is missing here, which is needed to pass the CI check and also if you want this pr to be mentioned in the release notes

substrate/frame/staking/src/ledger.rs
```diff
@@ -106,11 +106,31 @@ impl<T: Config> StakingLedger<T> {
...
+	/// Returns [`Error::BadState`] when a bond is in "bad state". A bond is in a bad state when a
+	/// stash has a controller which is bonding a ledger associated with another stash.
 	pub(crate) fn get(account: StakingAccount<T::AccountId>) -> Result<StakingLedger<T>, Error<T>> {
-		let controller = match account {
-			StakingAccount::Stash(stash) => <Bonded<T>>::get(stash).ok_or(Error::<T>::NotStash),
-			StakingAccount::Controller(controller) => Ok(controller),
-		}?;
+		let (stash, controller) = match account.clone() {
+			StakingAccount::Stash(stash) =>
+				(stash.clone(), <Bonded<T>>::get(&stash).ok_or(Error::<T>::NotStash)?),
```
kianenigma, line 115:
```suggestion
				(stash.clone(), <Bonded<T>>::get(&stash).defensive_ok_or(Error::<T>::NotStash)?),
```

substrate/frame/staking/src/ledger.rs
```diff
@@ -106,11 +106,31 @@ impl<T: Config> StakingLedger<T> {
...
+			StakingAccount::Controller(controller) => (
+				Ledger::<T>::get(&controller)
+					.map(|l| l.stash)
+					.ok_or(Error::<T>::NotController)?,
```
kianenigma, line 119:
```suggestion
					.defensive_ok_or(Error::<T>::NotController)?,
```

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5523431

kianenigma (approved):

substrate/frame/staking/src/ledger.rs
```diff
@@ -106,11 +106,31 @@ impl<T: Config> StakingLedger<T> {
...
 	/// abstract the concept of controller/stash accounts from the caller.
+	///
+	/// Returns [`Error::BadState`] when a bond is in "bad state". A bond is in a bad state when a
+	/// stash has a controller which is bonding a ledger associated with another stash.
 	pub(crate) fn get(account: StakingAccount<T::AccountId>) -> Result<StakingLedger<T>, Error<T>> {
-		let controller = match account {
-			StakingAccount::Stash(stash) => <Bonded<T>>::get(stash).ok_or(Error::<T>::NotStash),
-			StakingAccount::Controller(controller) => Ok(controller),
-		}?;
+		let (stash, controller) = match account.clone() {
```
Ank4n, line 113: Don't really like that we are calling Ledger three times in the function. Can we refactor to something like this:
```
			pub(crate) fn get(account: StakingAccount<T::AccountId>) -> Result<StakingLedger<T>, Error<T>> {
		let controller = match account.clone() {
			StakingAccount::Stash(stash) => <Bonded<T>>::get(stash).ok_or(Error::<T>::NotStash),
			StakingAccount::Controller(controller) => Ok(controller),
		}?;

		let ledger = <Ledger<T>>::get(&controller)
			.map(|mut ledger| {
				ledger.controller = Some(controller.clone());
				ledger
			})
			.ok_or(Error::<T>::NotController)?;

		if let StakingAccount::Stash(expected_stash) = account {
			ensure!(ledger.stash == expected_stash, Error::<T>::BadState);
		}

		Ok(ledger)
	}
```

substrate/frame/staking/src/ledger.rs
```diff
@@ -201,6 +221,30 @@ impl<T: Config> StakingLedger<T> {
...
+		ensure!(self.stash != *controller, Error::<T>::AlreadyPaired);
+
+		// check if the ledger's stash is a controller of another ledger.
+		if let Some(bonded_ledger) = Ledger::<T>::get(&self.stash) {
+			// there is a ledger bonded by the stash. In this case, the stash of the bonded ledger
+			// should be the same as the ledger's stash. Otherwise fail to prevent data
+			// inconsistencies. See <https://github.com/paritytech/polkadot-sdk/pull/3639> for more
+			// details.
+			ensure!(bonded_ledger.stash == self.stash, Error::<T>::BadState);
+		}
```
Ank4n, line 239: Is this check redundant if ledger is read from our custom getter?

gpestana: No they are not the same. The custom getter checks for:

- given a `Bonded(stash, controller)`, the ledger of the controller is the correct one (i.e. `Ledger(controller).stash == stash`). So we allow double (sane) bonded ledgers to be fetched and mutated. This check is more lenient, since it's not necessarily wrong to mutate double bonded ledgers.

This check, on the other hand, is checking if the stash of the ledger is a controller of *another* ledger. In this case, we want to prevent the double bonded ledgers from running the `set_controller` otherwise it may get into a bad state. This check is stricter.

substrate/frame/staking/src/ledger.rs
gpestana: Ok, I simplified this further. lkwyt!

substrate/frame/staking/src/ledger.rs
gpestana: Should this be defensive? The `StakingLedger::get()` is also used in extrinsics like validate, nominate, etc so we may be throwing defensives on user calls, which are expected from a runtime perspective if the user enters the wrong account IDs.

substrate/frame/staking/src/ledger.rs
```diff
@@ -106,18 +106,36 @@ impl<T: Config> StakingLedger<T> {
...
 			})
-			.ok_or(Error::<T>::NotController)
+			.ok_or(Error::<T>::NotController)?;
+
+		// if ledger bond is in a bad state, return error to prevent applying operations that may
+		// further spoil the ledger's state. A bond is in bad state when the bonded controller is
+		// associted with a different ledger (i.e. a ledger with a different stash).
+		//
+		// See <https://github.com/paritytech/polkadot-sdk/issues/3245> for more details.
+		ensure!(Bonded::<T>::get(&stash).is_some() && ledger.stash == stash, Error::<T>::BadState);
```
Ank4n, line 136: Should we also check `Bonded::<T>::get(&stash)` is same as controller?

substrate/frame/staking/src/ledger.rs
```diff
@@ -106,18 +106,36 @@ impl<T: Config> StakingLedger<T> {
 	/// This getter can be called with either a controller or stash account, provided that the
 	/// account is properly wrapped in the respective [`StakingAccount`] variant. This is meant to
 	/// abstract the concept of controller/stash accounts from the caller.
+	///
+	/// Returns [`Error::BadState`] when a bond is in "bad state". A bond is in a bad state when a
+	/// stash has a controller which is bonding a ledger associated with another stash.
 	pub(crate) fn get(account: StakingAccount<T::AccountId>) -> Result<StakingLedger<T>, Error<T>> {
```
Ank4n, line 112: One issue I feel when reading this function, since it handles both reading by controller and stake, it becomes a little harder to reason about which checks we need for each. What do you think about just splitting these flows.

So internally two functions `get_by_stash` and `get_by_controller` that this function calls.

Leave that decision to you though. Feel free to merge if you are happy and we probably will do lot more cleanups and refactors once controllers are removed completely.

gpestana: > probably will do lot more cleanups and refactors once controllers are removed completely.

Yes, I agree with this statement. The idea of this fn was to keep the code refactors to a minimal once the controllers were deprecated and remove the whole `StakingAccount` machinery etc.

substrate/frame/staking/src/ledger.rs
gpestana: Yes, that's better. done.

substrate/frame/staking/src/ledger.rs
kianenigma: I suggested it to be defensive because if the caller of this function uses `StakingAccount::Stash(_)`, then `Bonded` storage item _MUST_ exist, and similarly for `StakingAccount::Controller(_)`.

But I see what you mean; if someone calls `staking.validate(alice)` and alice is not even a staker, then we will fail. Yeah you are right, defensive would produce too many false negatives here.

gpestana: On top of that, the defensives in testing are a bit too strict to work with. There are cases where we test that `validate(reaped_stash)` and ensure that we get a `NotStash` error, for example. And with defensive, we need to handle the panic instead of the returned error.

There are ways to handle this by breaking down and re-org the tests but I've come across this and wish that it would be easier to assert a defensive and runtime error at the same time and be able to proceed with the tests. I think this could be improved at the FRAME testing level, I'll think more about it.

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-digest-19-mar-2024/6889/1

Polkadot-Forum:

https://forum.polkadot.network/t/recover-corrupted-staking-ledgers-in-polkadot-and-kusama/9796/1

PR #3643, opened 2024-03-11
[Collator Selection] Fix weight refund for `set_candidacy_bond`

### LINKED ISSUE #3642: Incomplete weight refund for `set_candidacy_bond` in `pallet_collator_selection`
The current impl of `pallet_collator_selection::set_candidacy_bond` charges for both the length of the candidate list as well as the number of candidates kicked as a result of a higher bond being imposed on them and their existing bonds not meeting the new requirements.

However, when the candidacy bond decreases, there are no iterations on the list, but we still charge for the length of the candidate list as if we iterated over all the candidates but kicked none. In this particular case, we should refund weight as if the list was empty. This needs to be fixed in `pallet_collator_selection::set_candidacy_bond`.

georgepisaltu:
Fixes #3642 

This PR implements the weight refund of `pallet_collator_selection::set_candidacy_bond` to account for no iterations when the bond is decreased.


joepetrowski (approved):

bkchr (approved):

bkchr:
@georgepisaltu please add a prdoc.

PR #3644, opened 2024-03-11
process_inherent_data: use an in memory overlay to minimize reads and writes to storage

sandreim:
Changes on top of https://github.com/paritytech/polkadot-sdk/pull/3479

Introduce a generic`StorageMapOverlay` and use it to efficiently mutate `PendingAvailability` storage in memory. Early in `process_inherent_data` we create this overlay and only write back what was changed before returning from the function. 

Runtime APIs must still access the storage directly.

This changeset doesn't attempt to solve any logic issues that have been identified in the review of  https://github.com/paritytech/polkadot-sdk/pull/3479, but it does try to help with allowing to change values while iterating the in-memory map, currently at the cost of an extra clone.


alindima (commented):
Looks like an interesting idea. I wonder if we can see the improvement it adds

polkadot/runtime/parachains/src/util.rs
```diff
@@ -115,6 +118,61 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+}
+
+// An in memory overlay for large storage maps that we frequently iterate and want
+// to reading/writing to multiple times. A cache with write back capabilities.
+//
+// Maybe TODO(in the future): move this capability to FRAME and ensure we use blake256 hashing of
+// keys, to ensure attackers cannot control the keys and force us to create overweight blocks.
+pub(crate) struct StorageMapOverlay<K, V, P: PopulateKeys> {
+	data: hashbrown::HashMap<K, V>,
+	modified: hashbrown::HashSet<K>,
```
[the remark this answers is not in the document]
sandreim: It is mandatory there are no hash collisions that can be controlled by an attacker.

polkadot/runtime/parachains/src/util.rs
```diff
@@ -115,6 +118,61 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+	V: Clone,
+	P: PopulateKeys<Key = K, Value = V>,
+{
+	// Construct a new overlay instance.
+	pub fn new() -> Self {
+		let data = P::populate();
+		Self { data, modified: Default::default(), _phantom: Default::default() }
+	}
+
+	/// Get a value from cache.
```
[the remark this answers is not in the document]
sandreim: This is similar to what the storage would return. I could make it a reference that you can clone at caller scope.

sandreim:
> Looks like an interesting idea. I wonder if we can see the improvement it adds

I'll have to move the PR against master, but not sure it will be as relevant as that would not have all the complications introduced by https://github.com/paritytech/polkadot-sdk/pull/3479 .

polkadot/runtime/parachains/src/util.rs
```diff
@@ -115,6 +118,61 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+	pub fn get(&self, key: &K) -> Option<V> {
+		self.data.get(key).cloned()
+	}
+
+	/// Update a value and make key dirty.
+	pub fn set(&mut self, key: K, value: V) {
```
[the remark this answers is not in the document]
sandreim: I'd better add a remove function. The `into_iter()` is compatible to what the storage expects for deletion.

alindima:
> I'll have to move the PR against master, but not sure it will be as relevant as that would not have all the complications introduced by https://github.com/paritytech/polkadot-sdk/pull/3479 .

Maybe before doing that, we could compare performance against #3479

sandreim:
bot bench-all pallet --pallet "runtime_parachains::paras_inherent"

sandreim:
bot bench-all pallet --pallet=runtime_parachains

sandreim:
bot cancel

sandreim:
bot bench-all pallet --pallet=runtime_parachains -v UPSTREAM_MERGE=n

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5514561

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5515311

sandreim:
bot bench-all pallet --pallet=runtime_parachains -v UPSTREAM_MERGE=n

sandreim:
bot cancel

sandreim:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent

sandreim:
bot cancel

sandreim:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent

sandreim:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5516997

ggwpez:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5519777

sandreim:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5523172

polkadot/runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -667,6 +691,11 @@ impl<T: Config> Pallet<T> {
 			disputes,
 			parent_header,
 		};
+
+		// Write back to storage only keys that have been updated or deleted.
+		for (para_id, candidates) in pending_availability_overlay.into_iter() {
```
eskimor, line 696: What about using a `Drop` instance like [here](https://github.com/paritytech/polkadot-sdk/blob/ea458d0b95d819d31683a8a09ca7973ae10b49be/polkadot/runtime/parachains/src/paras/mod.rs#L2300)?

polkadot/runtime/parachains/src/util.rs
```diff
@@ -115,6 +118,68 @@ pub fn take_active_subset<T: Clone>(active: &[ValidatorIndex], set: &[T]) -> Vec
...
+
+impl<K, V, P> StorageMapOverlay<K, V, P>
+where
+	K: sp_std::hash::Hash + Eq + PartialEq + Clone + Copy,
+	V: Clone,
+	P: PopulateKeys<Key = K, Value = V>,
+{
+	// Construct a new overlay instance.
+	pub fn new() -> Self {
+		let data = P::populate();
```
eskimor, line 149: Given that we are doubling the memory footprint here, it would be really good if this improved performance by more than just a few percent.

polkadot/runtime/parachains/src/paras_inherent/mod.rs
sandreim: yes, good idea, I'll have to see how it should work in the case of this generic overlay.

polkadot/runtime/parachains/src/util.rs
sandreim: Right now it improves it by around 12% . I'd say it is not great, but that's what is currently possible. What kind of improvement are you expecting for the tradeoff to be worth it ?

alindima: I'll think about it a bit. In the first mutate, we update the availability votes. But when doing the enactments, we may enact more candidates than the ones that have been just made available in this block.
If for example we have A B C backed. B and C are already available. A is just being made available now. So we have to enact all candidates (A B and C).

That's why I chose to iterate twice there. I'll think more about it to see if we can call mutate only once for each para

sandreim:
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::paras_inherent  -v UPSTREAM_MERGE=n

PR #2657, opened 2023-12-07
[frame] `#[pallet::composite_enum]` improved variant count handling + removed `pallet_balances`'s `MaxHolds` config

bkontur:
I started this investigation/issue based on @liamaharon question [here](https://github.com/paritytech/polkadot-sdk/pull/1801#discussion_r1410452499).

## Problem

The `pallet_balances` integrity test should correctly detect that the runtime has correct distinct `HoldReasons` variant count. I assume the same situation exists for RuntimeFreezeReason.

It is not a critical problem, if we set `MaxHolds` with a sufficiently large value, everything should be ok. However, in this case, the integrity_test check becomes less useful.

**Situation for "any" runtime:**
- `HoldReason` enums from different pallets:
```rust
        /// from pallet_nis
        #[pallet::composite_enum]
	pub enum HoldReason {
		NftReceipt,
	}

        /// from pallet_preimage
        #[pallet::composite_enum]
	pub enum HoldReason {
		Preimage,
	}

        // from pallet_state-trie-migration
        #[pallet::composite_enum]
	pub enum HoldReason {
		SlashForContinueMigrate,
		SlashForMigrateCustomTop,
		SlashForMigrateCustomChild,
	}
```

- generated `RuntimeHoldReason` enum looks like:
```rust
pub enum RuntimeHoldReason {

    #[codec(index = 32u8)]
    Preimage(pallet_preimage::HoldReason),

    #[codec(index = 38u8)]
    Nis(pallet_nis::HoldReason),

    #[codec(index = 42u8)]
    StateTrieMigration(pallet_state_trie_migration::HoldReason),
}
```

- composite enum `RuntimeHoldReason` variant count is detected as `3`
- we set `type MaxHolds = ConstU32<3>`
- `pallet_balances::integrity_test` is ok with `3`(at least 3)

However, the real problem can occur in a live runtime where some functionality might stop working. This is due to a total of 5 distinct hold reasons (for pallets with multi-instance support, it is even more), and not all of them can be used because of an incorrect `MaxHolds`, which is deemed acceptable according to the `integrity_test`:
  ```
  // pseudo-code - if we try to call all of these:

  T::Currency::hold(&pallet_nis::HoldReason::NftReceipt.into(), &nft_owner, deposit)?;
  T::Currency::hold(&pallet_preimage::HoldReason::Preimage.into(), &nft_owner, deposit)?;
[...]

PR #3645, opened 2024-03-11
move integrity tests feature of insecure-ed-zero

girazoki:
Simple PR just moving the `insecure-ed-zero` only to its appropriate assert


girazoki:
Also is there a reason why we are only testing maxFreezes and not maxHolds?

ggwpez:
> Also is there a reason why we are only testing `maxFreezes` and not `maxHolds`?

Hm i dont think so... please add it then.

bkchr (approved):

girazoki:
> > Also is there a reason why we are only testing `maxFreezes` and not `maxHolds`?
> 
> Hm i dont think so... please add it then.

Will do yes

girazoki:
> > Also is there a reason why we are only testing `maxFreezes` and not `maxHolds`?
> 
> Hm i dont think so... please add it then.

Ah sorry, it seems there is no need for that according to https://github.com/paritytech/polkadot-sdk/pull/2657:

`In addition, MaxHolds is removed (as suggested https://github.com/paritytech/polkadot-sdk/pull/2657#discussion_r1443324573) from pallet_balances, and its Holds are now bounded to RuntimeHoldReason::VARIANT_COUNT. Therefore, there is no need to let the runtime specify MaxHolds.`

ggwpez:
Looks like we already did this and this MR became a No-OP. Closing.   
Screenshot 2024-07-17 at 23 12 37

PR #3646, opened 2024-03-11
westend: `SignedPhase` is a constant

bkchr:
In preparation for the merkleized metadata, we need to ensure that constants are actually constant. If we want to test the unsigned phase we could for example just disable signed voter. Or we add some extra mechanism to the pallet to disable the signed phase from time to time.


kianenigma (commented):
Given that we don't have a workaround in this PR yet, 
and given that I think this value is actually kinda useless in the metadata,
and given that I like the testing property of this constant being such, ensuring our OCW code path is being regularly, 


I suggest removing this from the metadata instead of this. 

@niklasad1 if the staking-miner doesn't break, then we are good. I think it is literally the only application that might read this from the metadata.

kianenigma:
> We will figure out another way to achieve this.

If not resolved within the context of this PR, please make an issue about it. If a good workaround exists, it might be a good one for mentoring.

kianenigma (approved):

niklasad1:
The staking miner is only using the `signed_phase constant` from the metadata and AFAIU this only removes `unsigned_phase constant`, correct?

This won't break the staking miner and I tested it right now as well.

polkadot/runtime/westend/src/lib.rs
```diff
@@ -584,7 +584,7 @@ impl pallet_election_provider_multi_phase::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type Currency = Balances;
 	type EstimateCallFee = TransactionPayment;
-	type SignedPhase = MaybeSignedPhase;
+	type SignedPhase = SignedPhase;
```
bkchr, line 587:
```suggestion
	type SignedPhase = MaybeSignedPhase;
```

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -586,7 +586,6 @@ pub mod pallet {
 		type EstimateCallFee: EstimateCallFee<Call<Self>, BalanceOf<Self>>;
 
 		/// Duration of the unsigned phase.
-		#[pallet::constant]
 		type UnsignedPhase: Get<BlockNumberFor<Self>>;
 		/// Duration of the signed phase.
 		#[pallet::constant]
```
bkchr, line 591:
```suggestion
```

niklasad1:
@bkchr why did you remove the `signed_phase`? It was still present when I had a look at this PR IIRC but after [this commit](https://github.com/paritytech/polkadot-sdk/pull/3646/commits/b7a9fac6bf89cf364024b8e27d8fc5532e4fcb5b) it broke the staking miner :)

I don't think it should be that hard to workaround though....

PR #3656, opened 2024-03-12
Make concurrent dialing more aggressive

dmitry-markin:
This PR changes the default value for concurrent dials from 8 to 16, and sets the TCP dial timeout to 20 secs. This should help connecting to nodes that have a lot of stale addresses in the DHT.

This is a remediation for https://github.com/paritytech/polkadot-sdk/issues/3519 (collator side).

### DIFF
--- a/substrate/client/network/src/service.rs
+++ b/substrate/client/network/src/service.rs
@@ -87,7 +87,7 @@ use std::{
 	collections::{HashMap, HashSet},
 	fs, iter,
 	marker::PhantomData,
-	num::NonZeroUsize,
+	num::{NonZeroU8, NonZeroUsize},
 	pin::Pin,
 	str,
 	sync::{
@@ -472,7 +472,10 @@ where
 				// NOTE: 24 is somewhat arbitrary and should be tuned in the future if necessary.
 				// See <https://github.com/paritytech/substrate/pull/6080>
 				.per_connection_event_buffer_size(24)
-				.max_negotiating_inbound_streams(2048);
+				.max_negotiating_inbound_streams(2048)
+				// Increase the default dial concurrency factor 8 to 16 to help with cases where DHT
+				// has plenty of stale peer addresses.
+				.dial_concurrency_factor(NonZeroU8::new(16).expect("0 < 16 < 256; qed"));
 
 			(builder.build(), bandwidth)
 		};
--- a/substrate/client/network/src/transport.rs
+++ b/substrate/client/network/src/transport.rs
@@ -22,7 +22,7 @@ use either::Either;
 use libp2p::{
 	core::{
 		muxing::StreamMuxerBox,
-		transport::{Boxed, OptionalTransport},
+		transport::{timeout::TransportTimeout, Boxed, OptionalTransport},
 		upgrade,
 	},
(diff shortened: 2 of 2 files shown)


substrate/client/network/src/transport.rs
```diff
@@ -31,6 +31,9 @@ use std::{sync::Arc, time::Duration};
 
 pub use libp2p::bandwidth::BandwidthSinks;
 
+/// Timeout after which a TCP connection attempt is considered failed.
+const TCP_DIAL_TIMEOUT: Duration = Duration::from_secs(20);
```
lexnv, line 35: dq: The default timeout is in the order of minutes?

substrate/client/network/src/service.rs
```diff
@@ -472,7 +472,10 @@ where
 				// NOTE: 24 is somewhat arbitrary and should be tuned in the future if necessary.
 				// See <https://github.com/paritytech/substrate/pull/6080>
 				.per_connection_event_buffer_size(24)
-				.max_negotiating_inbound_streams(2048);
+				.max_negotiating_inbound_streams(2048)
+				// Increase the default dial concurrency factor 8 to 16 to help with cases where DHT
+				// has plenty of stale peer addresses.
+				.dial_concurrency_factor(NonZeroU8::new(16).expect("0 < 16 < 256; qed"));
```
lexnv, line 478: dq: I think this is fine, but do you think there is a chance we might run into issues with the total number of file descriptors opened for the concurrent requests? 🤔

lexnv (approved):

altonen (approved):
LGTM but I think @bkchr's suggestion of pruning addresses more aggressively would be worth exploring as well. The number of known addresses shown in https://github.com/paritytech/polkadot-sdk/issues/3519#issuecomment-1988182237 looks like a bug to me.

substrate/client/network/src/transport.rs
dmitry-markin: On my linux box it's 2 min 10 secs.

substrate/client/network/src/service.rs
dmitry-markin: Yes, this is theoretically possible, especially for nodes with high out peer count. All my systems show `ulimit -n` as 1024. So, it means maximum 125 simultaneous connections to peers with bloated DHT records before the change, and 60 connections after the change.

Don't think the practical probability of hitting a lot of peers with bloated DHT records is high, but at least something to keep in mind.

dmitry-markin:
> LGTM but I think @bkchr's suggestion of pruning addresses more aggressively would be worth exploring as well. The number of known addresses shown in [#3519 (comment)](https://github.com/paritytech/polkadot-sdk/issues/3519#issuecomment-1988182237) looks like a bug to me.

Yes, hopefully that one is fixed by https://github.com/paritytech/polkadot-sdk/pull/3657 you already reviewed.

dmitry-markin:
As this doesn't help with reaching validators with stale DHT records, I don't think we should increase the default concurrency factor. As for the TCP dial timeout, I'm not that sure, but unless we see any benefits from setting it to 20 secs, I would keep everything as is.

So, I'm inclined to close this PR.

PR #3657, opened 2024-03-12
Do not report listen addresses via Identify protocol

dmitry-markin:


substrate/client/network/src/peer_info.rs
```diff
@@ -414,7 +414,8 @@ impl NetworkBehaviour for PeerInfoBehaviour {
 			},
 			FromSwarm::NewListenAddr(e) => {
 				self.ping.on_swarm_event(FromSwarm::NewListenAddr(e));
-				self.identify.on_swarm_event(FromSwarm::NewListenAddr(e));
+				// Do not report listen addresses to `Identify`, because remote nodes should only
+				// know about our external addresses.
```
altonen, line 418: Not sure if this is a good idea. Probably needs a burn-in in a real network to see if it's safe to remove because AFAIU many of our nodes rely on external address discovery via Identify and the issues we've experienced are limited Rococo. If this is necessary to fix it, then it probably should be behind a CLI flag to enable this behavior explicitly.

dmitry-markin: Yes, let's do a burn-in. This should not affect the external address discovery via Identify, as external addresses are handled separately via [`observed_addr`](https://github.com/libp2p/rust-libp2p/blob/v0.51/protocols/identify/src/handler.rs#L251) filed of Identify message.

altonen: Yeah I was mixing up listen and external addresses

dmitry-markin: The confusion is no surprise, as there is only one field `listen_addrs` in Identify message, where both listen and external addresses are added.

dmitry-markin:
Putting more thought into it, I think this is not enough to clean DHT from temporary spot instance addresses, as they will be discovered via Identify and published as external addresses anyway. Unless connections from temporary addresses are blocked by a firewall.

I think we need to explicitly filter all addresses except the explicitly set `--public-addr` from being added to DHT, even as external addresses. I would not introduce a new cli flag for this and enable this behavior always when `--public-addr` flag is passed, which seems reasonable as this is the intended dial address set by the operator.

@lexnv @altonen what do you think?

dmitry-markin:
The same applies to authority discovery, as a spot instance temporary addresses will end up in the authority discovery DHT records as external addresses.

dmitry-markin:
> I would not introduce a new cli flag for this and enable this behavior always when --public-addr flag is passed, which seems reasonable as this is the intended dial address set by the operator.

Another option is to add one more flag `--no-external-addr-discovery`, but the operator would need to pass all three: `--public-addr --hide-listen-addr --no-external-addr-discovery` while this behavior seems intended when `--public-addr` is passed.

I'm going to update the PR to cover both Identify and authority discovery once we decide on the flag usage.

altonen:
Out of caution, I'd go with an extra flag instead of modifying behavior of the current flags, simply because this is trying to address an issue we've encountered in Rococo, not on all networks, and because we don't know if there is someone relying on the current behavior of the flags. Your call though

lexnv:
I do like the `--public-addr xxx --public-addr-only`, this keeps things simple for us and node operations (as opposed of having 2 extra CLI options to hide the listen address and no external address).
Then, Rococo maintainers could add the extra flag to enable this behavior :D

dmitry-markin:
@lexnv could you have a look again, please? The logic with Identify was reworked + the action of the flag was extended on authority discovery DHT records.

substrate/client/network/src/peer_info.rs
```diff
@@ -71,6 +72,8 @@ pub struct PeerInfoBehaviour {
 	garbage_collect: Pin<Box<dyn Stream<Item = ()> + Send>>,
 	/// Record keeping of external addresses. Data is queried by the `NetworkService`.
 	external_addresses: ExternalAddresses,
+	/// Whether only these addresses should be advertised to remote peers in [`Identify`] messages.
+	public_addresses_only: Option<Vec<Multiaddr>>,
```
lexnv, line 76: Do we need to store the public addresses in this object?

When we construct the object we trigger an `Identify::NewExternalAddr` for each of those addresses.
Then, we only need to know if further external addresses from the swarm need propagation to the `identify`.
Could we use a flag here instead?

substrate/client/network/src/peer_info.rs
```diff
@@ -378,23 +404,41 @@ impl NetworkBehaviour for PeerInfoBehaviour {
...
-				self.identify.on_swarm_event(FromSwarm::NewExternalAddr(e));
-				self.external_addresses.add(e.addr.clone());
+				// Only notify [`Identify`] and add an external address if we do not have the
+				// explicit list of advertised addresses.
+				if self.public_addresses_only.is_none() {
+					self.identify.on_swarm_event(FromSwarm::NewExternalAddr(e));
+					self.external_addresses.add(e.addr.clone());
+				}
+			},
+			FromSwarm::ExpiredExternalAddr(e) => {
```
lexnv, line 435: What would happen if the external address that expired was part of the initial `public_addresses_only`?

We'd keep advertising the same addresses, although we'd know that they expired?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -224,6 +226,29 @@ where
...
+				addresses
+					.into_iter()
+					.map(|mut address| {
+						if let Some(multiaddr::Protocol::P2p(peer_id)) = address.iter().last() {
+							if peer_id != local_peer_id {
+								error!(
+									target: LOG_TARGET,
+									"Discarding invalid local peer ID in public address {address}.",
+								);
+								address.pop();
```
lexnv, line 243: I wonder what would be the best way to handle this. 

Since these are passed by the node-operator as a CLI param, we could probably return an error high level (where we receive the params)?

Node operators could abouse this CLI param to publish other addreses in DHT, would it be safer to return an error high level and here to `filter_map` and ignore the addresses with different peer IDs (in case we missed anything along the way)?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -306,30 +332,50 @@ where
...
+						a.with(multiaddr::Protocol::P2p(peer_id))
+					}
+				})
+				.collect()
+		};
 
-		debug!(target: LOG_TARGET, "Authority DHT record peer_id='{:?}' addresses='{:?}'", peer_id, addresses.clone().collect::<Vec<_>>());
+		debug!(
+			target: LOG_TARGET,
+			"Authority DHT record peer_id='{peer_id:?}' addresses='{addresses:?}'",
```
lexnv, line 375: nit: I think we could `Display` the peerID instead of `Debug`, I've seen some strange logs in grafana (my mistake here sorry)

substrate/client/network/src/peer_info.rs
dmitry-markin: Explicitly set external addresses never expire, as they are added with [`AddressScore::Infinite`](https://docs.rs/libp2p/0.51.3/libp2p/swarm/enum.AddressScore.html):
https://github.com/paritytech/polkadot-sdk/blob/4987d7982461e2e5ffe219cdf71ec697284cea7c/substrate/client/network/src/service.rs#L500-L506

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -306,30 +332,50 @@ where
...
+						a
+					} else {
+						a.with(multiaddr::Protocol::P2p(peer_id))
+					}
+				})
+				.collect()
+		};
 
-		debug!(target: LOG_TARGET, "Authority DHT record peer_id='{:?}' addresses='{:?}'", peer_id, addresses.clone().collect::<Vec<_>>());
+		debug!(
```
lexnv, line 373: dq:
I think moving the debug log here changes a bit the details we expose.
I was curious to know the addresses we publish in DHT prior to adding the `/p2p/local_peer_id`.

Having the log line here might hint us about `/p2p/other_peer` entries (because they will remain untouched).
Could you see any value in distinguishing between `rogue_ip_address` and `rogue_ip_address/p2p/local_peer_id`?

substrate/client/network/src/peer_info.rs
lexnv: Cool! Thanks for the info! 🙏

substrate/client/authority-discovery/src/worker.rs
dmitry-markin: `Vec<Multiaddr>` doesn't implement a `Display` trait, so we are left with `Debug` here, unless we want to implement custom formatting.

What was wrong with the logs in grafana?

substrate/client/authority-discovery/src/worker.rs
dmitry-markin: I moved the log here to see what addresses exactly are published to DHT. In case we want pretty addresses without `/p2p/...` part, we would also need to drop the `/p2p/...` part if it was provided by the operator.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -126,7 +126,9 @@ pub struct Worker<Client, Network, Block, DhtEventStream> {
 	/// List of keys onto which addresses have been published at the latest publication.
 	/// Used to check whether they have changed.
 	latest_published_keys: HashSet<AuthorityId>,
-	/// Same value as in the configuration.
+	/// Should we publish only these explicitly set addresses?
+	public_addresses_only: Option<Vec<Multiaddr>>,
```
lexnv, line 130: nit: We could prob have an enum for the mutual exclusive flags. Thinking outloud here, totally up to you

```rust
enum PublishedAddresses {
    List(Vec<Multiaddr>),
    External { non_global: bool }
}
```

substrate/client/network/src/peer_info.rs
dmitry-markin: Done

substrate/client/authority-discovery/src/worker.rs
lexnv: The logs where displaying the digest bytes of the peerID and not the peerID string format, which made this hard to follow

dmitry-markin: Aaah, got it wrong thinking you are suggesting to use `Display` for `Multiaddr`. Totally fine to `Display` peer id, will fix.

dmitry-markin: Done

lexnv (approved):
LGTM! Nice one 👍

substrate/client/authority-discovery/src/worker.rs
dmitry-markin: Seems reasonable, done.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -226,27 +238,32 @@ where
...
-								address.pop();
+		let published_addresses = match config.public_addresses_only {
+			Some(addresses) => {
+				PublishedAddresses::List({
+					// Make sure we have only valid peer ids in public addresses.
+					let local_peer_id: Multihash = network.local_peer_id().into();
+
+					addresses
+						.into_iter()
+						.map(|mut address| {
```
lexnv, line 249: small nit: If we `filter_map` these addresses and also add `multiaddr::Protocol::P2p(local_peer_id)` we could avoid a double allocation from the `.collect()` in `addresses_to_publish` for external addresses

dmitry-markin: Not sure that I understand. Could you elaborate?

substrate/client/authority-discovery/src/worker.rs
dmitry-markin: Probable makes sense returning an error high-level and terminating the node. I'll look if it's possible to implement it in a clean fashion.

As for abusing, don't think there is an abuse vector: even if `/p2p/...` protocol is not filtered (is it allowed to be passed as not the last protocol of the address?), the authority discovery records are signed with the publishing node's network key and the signatures a checked during `get_record` DHT call to match a peer id in the record.

What do you think?

dmitry-markin: > Probably makes sense returning an error high-level and terminating the node. I'll look if it's possible to implement it in a clean fashion.

It doesn't play well with the current design of the nodes startup. It would require introducing a new error enum for node services in polkadot, cumulus, substrate and pulling a libp2p dependency for `Multiaddr` introspection if we want to filter it in a higher level code. We could also make the authority-discovery constructor fallible and return `Result` from there, but this would require new error types anyway.

So, generally, I don't think we should implement this as an invalid peer id in the public address poses no real threat, and can just be discarded with an error in the log.

lexnv: That makes sense, thanks for looking into this 🙏

substrate/client/authority-discovery/src/worker.rs
lexnv: We could group [ the section](https://github.com/paritytech/polkadot-sdk/pull/3657/files#diff-6ce157fc193b9691b2c38c118ee81244da354c65166edad5fb4b93beaa7b0f85R376-R385) that adds `/p2p/peer_id` with the [iteration](https://github.com/paritytech/polkadot-sdk/pull/3657/files#diff-6ce157fc193b9691b2c38c118ee81244da354c65166edad5fb4b93beaa7b0f85R372) through external addresses and do a `collect()` only once. 

And for the `PublishedAddresses::List` return them directly in `addresses_to_publish` if we make sure to add `/p2p/peer_id` sufix here

dmitry-markin: Makes sense, done.

bkchr:
I'm not convinced by this pull request. You are trying to add some duct tape. This change will require manual intervention of node operators. Aka this will not fix anything. We have almost all the data that we need. Other nodes report us out addresses. We just need to take the data and put it together in a better way.

> This PR also fixes a subtle bug where expired external addresses were never removed from the list in `PeerInfo` when handling `FromSwarm::ExpiredExternalAddr`.

This is a correct fix. However, I'm not sure if you went down to libp2p checking when this is being generated. (I have done this and I think it is generated almost never because they have a address cache for 200 automatically).

As I already said in the issue or somewhere else, we need to start track the "lifeness" of these addresses manually in our networking code and prune them if they are too old. 

Authority discovery should be changed to report the addresses in the following order:
- Our own listen addresses/public addresses (This should already be good enough for proper validators as they are reachable from the outside)
- The rest you fill up with addresses reported from other nodes.

dmitry-markin:
> As I already said in the issue or somewhere else, we need to start track the "lifeness" of these addresses manually in our networking code and prune them if they are too old.

For this, we'd need to introduce the address probing protocol, something like [AutoNAT v2](https://github.com/libp2p/specs/issues/503) proposed in libp2p. Otherwise we don't have any source of truth for what addresses are real and what are incorrectly reported by remote peers with faulty NAT / port forwarding configurations or [incorrectly translated by libp2p](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/swarm/src/lib.rs#L1137-L1143) (this assigns a local listen port to observed external address and breaks things when a public address has a different port than a listen address).

> Authority discovery should be changed to report the addresses in the following order:
> 
>     * Our own listen addresses/public addresses (This should already be good enough for proper validators as they are reachable from the outside)
> 
>     * The rest you fill up with addresses reported from other nodes.

Currently, upon fetching an authority-discovery DHT record, we just insert the addresses into `Discovery` behavior via different forms of `add_known_address`, which all boil down to calling [`Discovery::add_know_address`](https://github.com/paritytech/polkadot-sdk/blob/4987d7982461e2e5ffe219cdf71ec697284cea7c/substrate/client/network/src/discovery.rs#L334), effectively just adding the addresses to the DHT routing table. When we dial a peer, we lookup addresses in libp2p Kad behavior's [`handle_pending_outbound_connection()`](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/protocols/kad/src/behaviour.rs#L2194) where they are fetched from a k-bucket's entry in the order they were inserted there. As the validator have already participated in DHT, there are chances there are already addresses in the routing table reported via Identify protocol, including all the listen addresses and incorrectly discovered external addresses.

TLDR; we can order addresses to make the address provided by the operator to go first, but we'll need to maintain the order in both (1) authority-discovery records and (2) DHT routing table (via providing the needed order in Identify).

@bkchr does this make sense?

bkchr:
> For this, we'd need to introduce the address probing protocol, something like [AutoNAT v2](https://github.com/libp2p/specs/issues/503) proposed in libp2p. Otherwise we don't have any source of truth for what addresses are real and what are incorrectly reported by remote peers with faulty NAT / port forwarding configurations or [incorrectly translated by libp2p](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/swarm/src/lib.rs#L1137-L1143) (it assigns a local listen port to an observed external address and breaks things when the external address has a different port than the listen address).

Generally we are not supporting authorities behind NATs. Authorities should be reachable directly. Otherwise we would need support for hole punching etc.

 

> Currently, upon fetching an authority-discovery DHT record, we just insert the addresses into `Discovery` behavior via different forms of `add_known_address`, which all boil down to calling [`Discovery::add_know_address`](https://github.com/paritytech/polkadot-sdk/blob/4987d7982461e2e5ffe219cdf71ec697284cea7c/substrate/client/network/src/discovery.rs#L334), effectively just adding the addresses to the DHT routing table.

Generally I was talking about the sending site. AKA what we put into the DHT for our node. 

> When we dial a peer, we lookup addresses in libp2p Kad behavior's [`handle_pending_outbound_connection()`](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/protocols/kad/src/behaviour.rs#L2194) where they are fetched from a k-bucket's entry in the order they were inserted there. As the validator have already participated in DHT, there are chances there are already addresses in the routing table reported via Identify protocol, including all the listen addresses and incorrectly discovered external addresses.

But we call this code in Kademlia, from [here](https://github.com/paritytech/polkadot-sdk/blob/4987d7982461e2e5ffe219cdf71ec697284cea7c/substrate/client/network/src/discovery.rs#L564). And before we already use this ephemeral list. So, we could give addresses in the ephemeral list a higher "weight" when they come from authority discovery.

dmitry-markin:
> Generally we are not supporting authorities behind NATs. Authorities should be reachable directly. Otherwise we would need support for hole punching etc.

Unfortunately, this is exactly what we are doing at Parity — running rococo validators on spot instances with hole punching.

> Generally I was talking about the sending site. AKA what we put into the DHT for our node.

:+1:

> But we call this code in Kademlia, from [here](https://github.com/paritytech/polkadot-sdk/blob/4987d7982461e2e5ffe219cdf71ec697284cea7c/substrate/client/network/src/discovery.rs#L564). And before we already use this ephemeral list. So, we could give addresses in the ephemeral list a higher "weight" when they come from authority discovery.

Yes, makes sense, missed this. Should be already working as needed if we publish the authority-discovery DHT records with the right order of addresses.

bkchr:
> Unfortunately, this is exactly what we are doing at Parity — running rococo validators on spot instances with hole punching.

Then we need to change the infrastructure and not the code ;)

bkchr:
> Should be already working as needed if we publish the authority-discovery DHT records with the right order of addresses.

Yes, but we could also optimize the function to not return duplicated addresses.

dmitry-markin:
Closing in favor of https://github.com/paritytech/polkadot-sdk/pull/3757.

Issue #1130, opened 2023-07-19
[asset-hubs] Investigate usage of MultiLocation vs VersionedMultiLocation


bkontur:
We have to usages of `MultiLocation` with storage:

**1. ForeignAssets:** asset_id as MultiLocation
```
pub type MultiLocationForAssetId = xcm::prelude::MultiLocation;

// ForeignAssets
pub type ForeignAssetsInstance = pallet_assets::Instance2;
impl pallet_assets::Config<ForeignAssetsInstance> for Runtime {
  ...
	type AssetId = MultiLocationForAssetId;
  ...
}

// Used in `pallet_assets` storage 
pub(super) type Asset<T: Config<I>, I: 'static = ()> = StorageMap<
		_,
		Blake2_128Concat,
		T::AssetId,
		...
	>;
```

**2. asset-conversion:** `Box<MultiLocation>` in storage
```
// Pallet
impl pallet_asset_conversion::Config for Runtime {
...
	type MultiAssetId = Box<MultiLocation>;
...
}

pub(super) type PoolIdOf<T> = (<T as Config>::MultiAssetId, <T as Config>::MultiAssetId);

// Pallet's storage of: PoolIdOf<Box<MultiLocation>, Box<MultiLocation>>
pub type Pools<T: Config> =
		StorageMap<_, Blake2_128Concat, PoolIdOf<T>, PoolInfo<T::PoolAssetId>, OptionQuery>;
```

**3. asset-conversion:** `Box<MultiLocation>` in runtime api see more here: https://github.com/paritytech/cumulus/issues/2915

---
TODOs/Questions:
- [ ] should we change `xcm::prelude::MultiLocation` to `xcm::VersionedMultiLocation` ?
  - [ ] prepare some migration for that?
- [ ] add some test that ensures compatibility of actually used `xcm::v3::MultiLocation` against `xcm::prelude::MultiLocation` ?

franciscoaguirre:
I also wrote in https://github.com/paritytech/polkadot-sdk/pull/1230#discussion_r1338731874 but we should definitely use `Versioned*` types for storage items

bkontur:
@franciscoaguirre 
ok, thank you, I will continue this direction with PoC for `VersionedLocation`,

We just need to think that this two locations below are not the same - they point to the same but have different encodings because of `V3/V4`:
```
let location1 = VersionedMultiLocation::V3(v3::MultiLocation::new(1, v3::Junctions::Here));
[...]

PR #3659, opened 2024-03-12
FRAME: Unity Balance Conversion for Different IDs of Native Asset

muharem:
Introduce types to define 1:1 balance conversion for different relative asset ids/locations of native asset.

Examples:
native asset on Asset Hub presented as `VersionedLocatableAsset` type in the context of Relay Chain is 
```
{
  `location`: (0, Parachain(1000)),
  `asset_id`: (1, Here),
}
```
and it's balance should be converted 1:1 by implementations of `ConversionToAssetBalance` trait.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5502008

polkadot/runtime/common/src/impls.rs
```diff
@@ -150,6 +151,27 @@ impl TryConvert<&VersionedLocation, xcm::latest::Location> for VersionedLocation
 	}
 }
 
+/// Adapter for [`Contains`] trait to match [`VersionedLocatableAsset`] type converted to the latest
+/// version of itself where it's location matched by `L` and it's asset id by `A` parameter types.
+pub struct ContainsLatest<L, A>(PhantomData<(L, A)>);
+impl<L, A> Contains<VersionedLocatableAsset> for ContainsLatest<L, A>
+where
+	L: Contains<xcm::v4::Location>,
+	A: Contains<xcm::v4::Location>,
```
bkontur, line 160: as name `ContainsLatest` says :)
```suggestion
	L: Contains<xcm::latest::prelude::Location>,
	A: Contains<xcm::latest::prelude::Location>,
```

polkadot/runtime/common/src/impls.rs
```diff
@@ -150,6 +151,27 @@ impl TryConvert<&VersionedLocation, xcm::latest::Location> for VersionedLocation
 	}
 }
 
+/// Adapter for [`Contains`] trait to match [`VersionedLocatableAsset`] type converted to the latest
+/// version of itself where it's location matched by `L` and it's asset id by `A` parameter types.
+pub struct ContainsLatest<L, A>(PhantomData<(L, A)>);
+impl<L, A> Contains<VersionedLocatableAsset> for ContainsLatest<L, A>
```
bkontur, line 157: I would move it to the `xcm-builder`, maybe here: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/matches_location.rs ?

bkontur (approved):
lgmt, just two nits and I didn't check integration tests, just runtimes and xcm stuff

polkadot/runtime/common/src/impls.rs
muharem: it's implemented over `VersionedLocatableAsset` type, which is part of common.

bkontur: I see, I thought `VersionedLocatableAsset` was part of the XCM stuff, so maybe we could move everything related to `VersionedLocatableAsset` and `ContainsLatest` to one place in the XCM builder., e.g.: 
https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/locatable_asset.rs.
_(I think I saw some complaints about adding too much stuff to the `polkadot-runtime-common`.)_

As I see, we already have `LocatableAssetId` in `xcm-builder` here: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/pay.rs#L191-L198, which could possibly also go to the `locatable_asset.rs`

And maybe `VersionedLocatableAsset` could also wrap that `LocatableAssetId`, but I think this can be just from V5, e.g.:
```
pub enum VersionedLocatableAsset {
...
	#[codec(index = 4)]
-        V4 { location: xcm::v4::Location, asset_id: xcm::v4::AssetId },
+	V4(LocatableAssetId),
}
```

bkontur: > Since we should be more conservative with xcm-builder...

Where is this coming from? Can you elaborate more?

> Every time someone see these types, ( `VersionedLocatableAsset` and `LocatableAssetId`), they suggest to remove it and use only `Location`, refactor it somehow, etc.

Well, perhaps it's worth considering refactoring it then.


I think keeping `VersionedLocatableAsset` outside of `xcm` / `xcm-builder`, where we currently have all `Versioned*` stuff and also `LocatableAssetId`, seems kind of strange. As far as I can see, it is used by relay runtimes and collectives runtime, so from this perspective, `polkadot-runtime-common` is not the best place (`xcm-builder` seems better):
```
[package]
name = "polkadot-runtime-common"
description = "Pallets and constants used in Relay Chain networks."
```

`VersionedLocatableAsset` contains v3 and v4. How do you propose keeping this up-to-date? Do you expect that when we add `xcm::v5`, someone will also update this? If so, there's no way to know from the xcm module's point of view (considering that one day we might move xcm out of polkadot-sdk) that some versioned stuff is used elsewhere. If not, for example, when we add `xcm::v6`, we remove `xcm:v3`. Is that okay? I think the check-migrations jobs should detect this and fail.

Another thing to consider, I see it is used for `pallet_collective` as the key for `ConversionRateToNative` or `pallet_treasury` as `AssetKind`. I think it's a similar situation to what's discussed here: https://github.com/paritytech/polkadot-sdk/issues/1130. And if I recall correctly, the last time we spoke with Joe about `pallet_assets` and `Location` as `AssetId`, everything pointed to the solution of using `xcm::latest::prelude::Location` as `AssetId` (with migrations) and `VersionedLocation` as `AssetIdParameter` (which is used for extrinsics).

I think `VersionedLocatableAsset` presents a very similar situation. Perhaps it's worth considering splitting `AssetKind` into `AssetKind` / `AssetKindParameter`. For example, this could help avoid using `Box` for extrinsic parameters like `asset_kind: Box<T::AssetKind>`, though I'm not sure if this is good or bad, it can be left for the runtime to decide. My point is that ideally, we should have one method/pattern for storing XCM-related data.

muharem: > Since we should be more conservative with xcm-builder

I meant between `polkadot-runtime-common` and `xcm-builder`, we should be more conservative with the last. It's better if we keep this as it is, and move later. Then moving it now, and removing it if we decide we do not need it. 

I do not see a problem with having it under this common crate, and using it in system chain, I think we have more cases like this. 
If we add new xcm version, and do not update this type, it wont compile because of the match. But we should not rely on the types location and match, there should be tests catching the issue. And in many cases if this types does not have the latest version, it is not a breaking issue. 

With this PR I would not move anything. I would try to explore first what's the alternatives to this type, since it was discussed few times. If we have another alternative, we introduce it and remove this type. If we keep this type, we can moving it.

franciscoaguirre: I agree there's no need to do the refactor here, but I would very much look into the refactor. I've previously had a go at simplifying this in https://github.com/paritytech/polkadot-sdk/pull/2006. This `LocatableAsset` was introduced to know when an asset was outside of the local chain to do a `spend` from the treasury and know to which location to send it. I argue it's enough with just an asset id for the asset, and then we use the location of the beneficiary (or sender) to know where to send it.

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/treasury.rs
```diff
@@ -0,0 +1,270 @@
...
+			Box::new(Location::new(0, Into::<[u8; 32]>::into(alice.clone())).into()),
+			None,
+		));
+		// claim the spend.
+		assert_ok!(Treasury::payout(bob_signed.clone(), 0));
+
+		assert_expected_events!(
+			Rococo,
+			vec![
+				RuntimeEvent::Treasury(pallet_treasury::Event::Paid { .. }) => {},
```
franciscoaguirre, line 227: Probably the beneficiary of this can be asserted as well

polkadot/runtime/common/src/impls.rs
muharem: I will check that draft PR, and I am also looking for a solution where that PayOverXcm can also detect that the asset is located on local chan, but beneficiary is on Asset Hub. We have this use case when treasury wants to fund it's own account on Asset Hub. I will initiate a discussion around it, and open an RFC.
@franciscoaguirre please approve this PR if you happy with it.

cumulus/parachains/integration-tests/emulated/tests/collectives/collectives-westend/src/tests/fellowship_treasury.rs
```diff
@@ -0,0 +1,236 @@
...
+			})),
+		});
+
+		// Dispatched from Root to `dispatch_as` `Signed(treasury_account)`.
+		assert_ok!(teleport_call.dispatch(root));
+
+		assert_expected_events!(
+			Westend,
+			vec![
+				RuntimeEvent::XcmPallet(pallet_xcm::Event::Sent { .. }) => {},
```
franciscoaguirre, line 85: Could assert the destination

cumulus/parachains/integration-tests/emulated/tests/collectives/collectives-westend/src/tests/fellowship_treasury.rs
```diff
@@ -0,0 +1,236 @@
...
+		type RuntimeOrigin = <Westend as Chain>::RuntimeOrigin;
+		type Runtime = <Westend as Chain>::Runtime;
+		type Treasury = <Westend as WestendPallet>::Treasury;
+
+		// Fund Fellowship Treasury from Westend Treasury.
+
+		let treasury_origin: RuntimeOrigin =
+			westend_runtime::governance::pallet_custom_origins::Origin::Treasurer.into();
+		let fellowship_treasury_location: Location =
+			Location::new(1, [Parachain(1001), PalletInstance(65)]);
```
franciscoaguirre, line 102: This is the sovereign account in Asset Hub of the fellowship treasury?

polkadot/xcm/xcm-builder/src/barriers.rs
```diff
@@ -322,6 +322,29 @@ impl<ParaId: IsSystem + From<u32>> Contains<Location> for IsChildSystemParachain
...
+
+/// Matches if the given location contains only the specified amount of parents and no interior
+/// junctions.
+pub struct IsParentsOnly<Count>(PhantomData<Count>);
+impl<Count: Get<u8>> Contains<Location> for IsParentsOnly<Count> {
+	fn contains(t: &Location) -> bool {
+		t.contains_parents_only(Count::get())
+	}
+}
+
```
franciscoaguirre, line 347: This doesn't look like it should go in `barriers.rs`

substrate/frame/support/src/traits/tokens/misc.rs
```diff
@@ -299,6 +300,33 @@ where
 	fn ensure_successful(_: AssetId) {}
 }
 
+/// Implements [`ConversionFromAssetBalance`], allowing for a 1:1 balance conversion of the asset
+/// when it meets the conditions specified by `C`. If the conditions are not met, the conversion is
+/// delegated to `O`.
+pub struct UnityOrOuterConversion<C, O>(core::marker::PhantomData<(C, O)>);
```
franciscoaguirre, line 306:
```suggestion
pub struct OneToOneIfElse<C, O>(core::marker::PhantomData<(C, O)>);
```

or

```suggestion
pub struct OneToOneOrOuterConversion<C, O>(core::marker::PhantomData<(C, O)>);
```

I think it expresses what you want better

prdoc/pr_3659.prdoc
```diff
@@ -0,0 +1,12 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Unity Balance Conversion for Different IDs of Native Asset 
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Introduce types to define 1:1 balance conversion for different relative asset ids/locations
```
franciscoaguirre, line 9: What is the purpose of this change? Should probably mention here to the users what has now changed in how the system is used.

polkadot/runtime/common/src/impls.rs
franciscoaguirre: I only want to understand the purpose of this change. Is it because DOT was being passed on to the asset rate converter? Or is it to allow things like USDT to be used 1:1 when they come from system parachains?

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/treasury.rs
```diff
@@ -0,0 +1,270 @@
...
+		let alice_location: Location =
+			[Junction::AccountId32 { network: None, id: Rococo::account_id_of(ALICE).into() }]
+				.into();
+		let asset_hub_location: Location = [Parachain(1000)].into();
+		let native_asset = Location::parent();
+
+		let treasury_spend_call = RuntimeCall::Treasury(pallet_treasury::Call::<Runtime>::spend {
+			asset_kind: bx!(VersionedLocatableAsset::V4 {
+				location: asset_hub_location.clone(),
+				asset_id: native_asset.into(),
```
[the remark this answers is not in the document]
muharem: yes, this is correct. DOT is parent on asset hub

polkadot/runtime/common/src/impls.rs
franciscoaguirre: Hmm, with sovereign accounts it definitely becomes harder. My draft PR was of the idea that we could just use the beneficiary location to know to which chain to send it to. It wouldn't work when we want to send it somewhere different than the location to fund a sovereign account.

We might need then a location for the chain and another for the beneficiary, like on the `send` extrinsic in pallet-xcm. There's no need to name it locatable asset though, just a regular asset, a regular destination, and a regular beneficiary.

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/treasury.rs
```diff
@@ -0,0 +1,270 @@
...
+		);
+
+		// Assert events triggered by xcm pay program:
+		// 1. treasury asset transferred to spend beneficiary;
+		// 2. response to Relay Chain Treasury pallet instance sent back;
+		// 3. XCM program completed;
+		assert_expected_events!(
+			AssetHubRococo,
+			vec![
+				RuntimeEvent::Balances(pallet_balances::Event::Transfer { .. }) => {},
```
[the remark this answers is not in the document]
muharem: this assert is broken, I cannot really match beneficiary here. but we are fine here, I check the balance above

cumulus/parachains/integration-tests/emulated/tests/collectives/collectives-westend/src/tests/fellowship_treasury.rs
muharem: yes

polkadot/xcm/xcm-builder/src/barriers.rs
muharem: its consistent to what we have in this file, `IsChildSystemParachain` just above for example

prdoc/pr_3659.prdoc
muharem: not sure why you mean, this does not changes anything for users, only introduces few new types. 
it does changes configuration for westend and rococo chains, but it changes it a way it is expected to work

substrate/frame/support/src/traits/tokens/misc.rs
muharem: we have type above with a name - `UnityAssetBalanceConversion`, which also means 1:1. its not wrong and I would keep it consistent.

cumulus/parachains/integration-tests/emulated/tests/collectives/collectives-westend/src/tests/fellowship_treasury.rs
muharem: It does not work right now with this macro

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/treasury.rs
muharem: It does not work right now with this macro

polkadot/runtime/common/src/impls.rs
muharem: > I only want to understand the purpose of this change. Is it because DOT was being passed on to the asset rate converter? Or is it to allow things like USDT to be used 1:1 when they come from system parachains?

The asset rate does not know there are different locations of native asset (as in the example in the PR description). And will require a rate to be set for it. With there types we make asset conversion understand different locations of native asset, and do not require a rate to be set for it.

franciscoaguirre: I guess that's one of the problems of marking the difference between the same asset on different chains

franciscoaguirre (approved):

polkadot/runtime/common/src/impls.rs
```diff
@@ -156,6 +156,27 @@ impl TryConvert<&VersionedLocation, xcm::latest::Location> for VersionedLocation
...
+		let (location, asset_id) = match asset.clone() {
+			V3 { location, asset_id } => match (location.try_into(), asset_id.try_into()) {
+				(Ok(l), Ok(a)) => (l, a),
+				_ => return false,
+			},
+			V4 { location, asset_id } => (location, asset_id),
+		};
+		L::contains(&location) && A::contains(&asset_id.0)
+	}
+}
```
bkontur, line 178: I think we should rename `ContainsLatest` to something specific, because it is not that clear if `latest` is related to the used XCM types or `lastest` version of `VersionedLocatableAsset` see also doc `type converted to the latest version of itself`. `VersionedLocatableAsset` has just V3/V4. When we add `xcm:V5` it would be even more confusing.

I would suggest to rename `ContainsLatest` to the `ContainsAssetFrom` and change order of L/A types, e.g.:

```suggestion
pub struct ContainsAssetFrom<A, L>(core::marker::PhantomData<(A, L)>);
impl<A, L> Contains<VersionedLocatableAsset> for ContainsAssetFrom<A, L>
where
	A: Contains<xcm::latest::prelude::Location>,
	L: Contains<xcm::latest::prelude::Location>,
{
	fn contains(asset: &VersionedLocatableAsset) -> bool {
		use VersionedLocatableAsset::*;
		let (location, asset_id) = match asset.clone() {
			V3 { location, asset_id } => match (location.try_into(), asset_id.try_into()) {
				(Ok(l), Ok(a)) => (l, a),
				_ => return false,
			},
			V4 { location, asset_id } => (location, asset_id),
		};
		L::contains(&location) && A::contains(&asset_id.0)
	}
}
```

bkontur: also maybe types:
```
        A: Contains<xcm::latest::prelude::Location>,
	L: Contains<xcm::latest::prelude::Location>,
```
could be better replaced with: `ContainsPair<Location, Location>` or `ContainsPair<AssetId, Location>`

bkontur (approved):
one nit, I suggest renaming `ContainsLatest` to the `ContainsAssetFrom`

Issue #3658, opened 2024-03-12
[RPC-Spec-V2] chainHead: Ensure unique reproted pruned blocks


lexnv:
Ensure the `Finalized` event contains unique `pruned` block hashes.

In the following topology, the finalized event could return duplicate pruned hashes:

```
finalized -> block 1
|
| ----> block 2 -----> block 3
            |
            | -----> block 4
```

This is an edge case, where the stale blocks reported as pruned are placed on a fork that contains forks.

PR #3667, opened 2024-03-12
chainHead: Report unique hashes for pruned blocks

lexnv:
This PR ensures that the reported pruned blocks are unique.

While at it, ensure that the best block event is properly generated when the last best block is a fork that will be pruned in the future.

To achieve this, the chainHead keeps a LRU set of reported pruned blocks to ensure the following are not reported twice:

```bash
	 finalized -> block 1 -> block 2 -> block 3
	
	                      -> block 2 -> block 4 -> block 5
	
	           -> block 1 -> block 2_f -> block 6 -> block 7 -> block 8
```

When block 7 is finalized the branch [block 2; block 3] is reported as pruned.
When block 8 is finalized the branch [block 2; block 4; block 5] should be reported as pruned, however block 2 was already reported as pruned at the previous step.

This is a side-effect of the pruned blocks being reported at level N - 1. For example, if all pruned forks would be reported with the first encounter (when block 6 is finalized we know that block 3 and block 5 are stale), we would not need the LRU cache.

cc @paritytech/subxt-team  

Closes https://github.com/paritytech/polkadot-sdk/issues/3658


jsdw:
I had assumed that ensuring pruned hashes were unique was as simple as using a `HashSet` or something when building the pruned list. That would ensure the blocks in a single pruned list were unique.

Am I right in my understanding that the extra bits in this PR are to also ensure that we don't see the same block hash mentioned again across different pruned lists, too?

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -434,10 +450,9 @@ where
 	///
 	/// The result does not include hashes from `to_ignore`.
```
skunert, line 451:
```suggestion
```

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -495,15 +511,64 @@ where
 
 		match self.best_block_cache {
 			Some(block_cache) => {
-				// If the best block wasn't pruned, we are done here.
-				if !pruned_block_hashes.iter().any(|hash| *hash == block_cache) {
+				// We need to generate a `NewBlock` event for the finalized block when:
```
skunert, line 514: BestBlock instead of NewBlock?

skunert (commented):
Logic looks correct, only some nits in the comments 👍

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -495,15 +509,64 @@ where
...
```
lexnv, line 512:
```suggestion
				// We need to generate a `BestBlock` event for the finalized block when:
```

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
lexnv: Best block :D

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -67,6 +73,8 @@ pub struct ChainHeadFollower<BE: Backend<Block>, Block: BlockT, Client> {
 	sub_id: String,
 	/// The best reported block by this subscription.
 	best_block_cache: Option<Block::Hash>,
```
jsdw, line 75: Nit: when I read `best_block_cache` below I thoguht it was a cache of best blocks. `current_best_block` would be clearer to me :)

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -50,6 +51,11 @@ use std::{
 /// The maximum number of finalized blocks provided by the
 /// `Initialized` event.
 const MAX_FINALIZED_BLOCKS: usize = 16;
+/// The size of the LRU cache for pruned blocks.
+///
+/// This is the exact value of the total number of pinned blocks, and ensures
+/// that all active pruned block hashes (if any) are kept in memory.
+const LRU_CACHE_SIZE: u32 = 512;
```
jsdw, line 58: So to make sure I understand; this number must equal the number of pinned blocks we keep?

Could we instead use `chain_head::MAX_PINNED_BLOCKS` or alias this to be that, so that there's no chance these numbers can get out of sync in the future?

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
lexnv: That makes sense, thanks!

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -486,24 +497,73 @@ where
...
+				// - (i) the last reported best block was pruned
+				// - (ii) the last reported best block is on a fork that will be pruned in the
+				//   future.
+				// Note: pruning happens on level n - 1.
+
+				// Best block already generated.
+				if block_cache == last_finalized {
 					events.push(finalized_event);
+					return Ok(events);
+				}
```
jsdw, line 519: To check my understanding, this is all abotu addressing this bit of the spec:

> The current best block, in other words the last block reported through a bestBlockChanged event, is guaranteed to either be the last item in finalizedBlockHashes, or to not be present in either finalizedBlockHashes or prunedBlockHashes.

So, when a new last finalized block is seen, we:
- Return happily if the best block was finalized.
- Else, return a new BestBlockChanged event if the current best block has been pruned.
- Else, return a new BestBlockChanged event if the current best block _will_ be pruned (ie is not a descendant of the new finalized chain)

To satisfy the spec, could the whole thing be simplified to pseudocode like this?:

```rust
if let Some(current_best_block) = self.best_block_cache {
    // current best block is in pruned list, so need to emit new BestBlock
    let is_in_pruned_list = pruned_block_hashes.iter().any(|hash| *hash == current_best_block);
    // current best block is not the last one we finalized
    let is_not_last_finalized = current_best_block != last_finalized;

    if is_in_pruned_list || is_not_last_finalized {
        let new_best_block = self.client.info().best_hash;
        events.push(BestBlockChanged { best_block_hash })
    }
}

events.push(finalized_event);
Ok(events)
```

In any case, I wonder whether we could avoid some of the early returns and such, and separate it so that we first write the logic to decide whether we need to push a new BestBlockChanged event and then do the push, with a single return at the end to hand back the events?

lexnv: Yep, that makes sense!

I think I was more worried about cases where we've already generated a `BestBlockChanged` event for a descendant for the finalized block. 

However, I think this is safe to regenerate a `BestBlockChanged` with an older block number, as long as it is the last reported finalized, from this spec statement:
> The current best block, in other words the last block reported through a bestBlockChanged event, is guaranteed to either be the last item in finalizedBlockHashes

The suggestion simplifies the code quite a bit, thanks! 🙏

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5753830

substrate/client/rpc-spec-v2/src/chain_head/chain_head_follow.rs
```diff
@@ -702,7 +707,10 @@ where
 		let merged = tokio_stream::StreamExt::merge(merged, stream_responses);
 		let stream = stream::once(futures::future::ready(initial)).chain(merged);
 
-		self.submit_events(&startup_point, stream.boxed(), pruned_forks, sink, sub_data.rx_stop)
-			.await
+		// These are the pruned blocks that we should not report again.
+		for pruned in pruned_forks {
```
skunert, line 711: Just to be sure I understand this correctly. When we generate the initial events we consider all blocks that are part of a fork that started below the finalized one as pruned, even if they are technically not pruned yet. So these will never be reported?

lexnv: Yep. I can't remember exactly, this was introduced quite a while ago to handle the case where the `Finalized` event would contain a `pruned block` not reported by a new block event. 

This abuses the purpose of `self.pruned_blocks`, which now holds:
- (previously stored in to_ignore)  forks that we did not report previously by the Initialized event
- (the intended purpose of this PR) pruned blocks already reported once

substrate/client/rpc-spec-v2/src/chain_head/tests.rs
```diff
@@ -3644,3 +3644,364 @@ async fn chain_head_limit_reached() {
 	// Initialized must always be reported first.
 	let _event: FollowEvent<String> = get_next_event(&mut sub).await;
 }
+
+#[tokio::test]
+async fn follow_unique_pruned_blocks() {
```
jsdw, line 3649: Nit: This test is quite long, and looking at it, I wonder whether we can extract a couple of helper fns or something so that it's easier to see what's actually beign tested vs all of the staet we're creating.

Like maybe (and could be at the top of this fn) something like:

```rust
async fn import_block(client, parent_hash, parent_num) -> block {
        let block = BlockBuilderBuilder::new(&*client)
		.on_parent_block(client.chain_info().genesis_hash)
		.with_parent_block_number(0)
		.build()
		.unwrap()
		.build()
		.unwrap()
		.block;
	let block_1_hash = block_1.hash();
	client.import(BlockOrigin::Own, block_1.clone()).await.unwrap();
        block
}
```

jsdw (approved):
This looks good to me; good job Alex!

PR #3674, opened 2024-03-13
fix(paseo-spec): New Paseo Bootnodes

al3mart:
This change includes new bootnodes for Paseo. 

_This is meant to be a silent PR_


ggwpez:
>This is meant to be a silent PR

So you dont want to have this announced in the CHANGELOG for visibility?

ggwpez (approved):

michalkucharczyk:
The state was changed, is that intended?

hbulgarini:
> So you dont want to have this announced in the CHANGELOG for visibility?

I think it's not necessary. 

> The state was changed, is that intended (PR description is not mentioning it)?

Thanks for noting it. Yes, the state was changed as we restarted the network weeks ago. This is intended.

michalkucharczyk (approved):

PR #3677, opened 2024-03-13
Always print connectivity report

alexggh:
This is printed every 10 minutes, I see no reason why it shouldn't be in all the logs, it would give us valuable information about what is going on with node connectivity when validators come-back to us to report issues.

### DIFF
--- a/polkadot/node/network/gossip-support/src/lib.rs
+++ b/polkadot/node/network/gossip-support/src/lib.rs
@@ -508,7 +508,7 @@ where
 			);
 		}
 		let pretty = PrettyAuthorities(unconnected_authorities);
-		gum::debug!(
+		gum::info!(
 			target: LOG_TARGET,
 			?connected_ratio,
 			?absolute_connected,


sandreim:
This can get pretty big if there are many unconnected authorities. Are we interested in exactly these? Because otherwise the metrics are always available per protocol - `polkadot_parachain_peer_count`

eskimor (approved):

eskimor:
> This can get pretty big if there are many unconnected authorities. Are we interested in exactly these? Because otherwise the metrics are always available per protocol - polkadot_parachain_peer_count

Right, but then there is also some network issue which would justify a bit of verbosity every 10 minutes. If we assume that these logs would be 1Meg in the worst case, then it would still take us 7 days to fill a GB. Sounds sensible for today's hardware.

sandreim:
> > This can get pretty big if there are many unconnected authorities. Are we interested in exactly these? Because otherwise the metrics are always available per protocol - polkadot_parachain_peer_count
> 
> Right, but then there is also some network issue which would justify a bit of verbosity every 10 minutes. If we assume that these logs would be 1Meg in the worst case, then it would still take us 7 days to fill a GB. Sounds sensible for today's hardware.

Your estimations might be right, but my question is mostly if we do this specifically to know the unconnected authorities or we are interested mainly in the connectivity ratio.

sandreim (approved):
In the end it shouldn't hurt.

alexggh:
> This can get pretty big if there are many unconnected authorities. Are we interested in exactly these? Because otherwise the metrics are always available per protocol

Those metrics are good to tell you something went wrong, but they will fail to tell which nodes are not connecting or some other stuff.

The flow I'm actually trying to optimise here, is people coming to us observing high level stuff is not working and providing us the log files they have, but there is literally nothing in there to tell you what went wrong.  So at minimum we need to understand how well connected that node was by default.  


> Right, but then there is also some network issue which would justify a bit of verbosity every 10 minutes. If we assume that these logs would be 1Meg in the worst case, then it would still take us 7 days to fill a GB. Sounds sensible for today's hardware.

My thoughts as well given how rarely we print this, I don't thing there is any danger there.

alexggh:
> In the end it shouldn't hurt.

famous last words :D

burdges:
Always possible to make a runlength encoded bitfield and then 7-bit or hex encode that, if youre worried about size.

bkchr:
https://github.com/paritytech/polkadot-sdk/pull/3913

Issue #3600, opened 2024-03-06
subsystem-bench: Extract prometheus endpoint launching from regression test cases to use it only with cli runner


AndreiEres:
We use an awkward `false` argument to not initialize a prometheus endpoint inside regression tests. Otherwise, each test will start an unconfigurable endpoint, causing errors because they will use the same port 9999.
https://github.com/paritytech/polkadot-sdk/blob/475e7a147676a4e7a9d255ddc7969dd35ea22882/polkadot/node/network/availability-recovery/tests/availability-recovery-regression-bench.rs#L100

It's better to extract the endpoint launching outside the test, as we already do with `valgrind` and `pyroscope`. But we still use prometheus inside the tests.
https://github.com/paritytech/polkadot-sdk/blob/475e7a147676a4e7a9d255ddc7969dd35ea22882/polkadot/node/subsystem-bench/src/cli/subsystem-bench.rs#L108-L121

gitofdeepanshu:
Can I take this?

Krayt78:
Is this still relevant to be worked on ? @AndreiEres

AndreiEres:
> Is this still relevant to be worked on ? @AndreiEres

Yes, please

Krayt78:
Hi @AndreiEres, i have some questions regarding testing these with prometheus/pyroscope/grafana.

I am following the readme and got the docker compose running, only caviat is that i am on windows using wsl on vscode and using docker desktop to create the container directly inside wsl.

I managed to get graphana working locally (i got both the parachain liveliness dashboard and kusama validator overview) and added the two data sources from prometheus and pyroscope and to run tests using the example tests. But for some reason i dont get any data in my graphana.

Is there something that i forgot in the setup to have the data shown inside graphana during the tests ?
Is there anything else that needs to be done besides making sure that the code compiles, the tests pass and the subsystem-bench  tests pass too ?

Krayt78:
On the console side i did : 

```
cargo build --profile=testnet --bin subsystem-bench -p polkadot-subsystem-bench

cd polkadot/node/subsystem-bench/docker
docker compose up
```

And once the compose was done : 

`target/testnet/subsystem-bench polkadot/node/subsystem-bench/examples/availability_read.yaml`

And on the graphana side, i logged in locally, added prometheus and pyroscope as data sources with these urls :
```
- Prometheus `http://prometheus:9090/`
- Pyroscope `http://pyroscope:4040/`
```
[...]

PR #3681, opened 2024-03-13
Extract prometheus launch

gitofdeepanshu:
Related to #3600 

extract the prometheus endpoint launching outside the test.

### DIFF
--- a/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
+++ b/polkadot/node/network/availability-distribution/tests/availability-distribution-regression-bench.rs
@@ -98,7 +98,7 @@ fn benchmark(config: TestConfiguration) -> BenchmarkUsage {
 fn run(config: TestConfiguration) -> BenchmarkUsage {
 	let mut state = TestState::new(&config);
 	let (mut env, _protocol_config) =
-		prepare_test(config.clone(), &mut state, TestDataAvailability::Write, false);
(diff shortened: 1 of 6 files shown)


gitofdeepanshu:
@AndreiEres 
this is my first time working with node and I don't think I have a good understanding of the code. 
are these acceptable changes?
btw how do you test a node?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5520539

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5522878

AndreiEres:
Thanks for the PR!

I'd suggest following these steps:
- Merge master branch and fix conflicts. 
- Seems like we have a lot of duplicated code now. It would be great to move the initialization up like we do with the Pyroscope agent.
- The Registry could be a problem, but we can also initialize it outside our tests and then pass it to testcases as an argument. 
- The endpoint initialization should be optional, as should the endpoint port. Look at how we do it with the Pyroscope agent. 
- It might be better to extract it into a `prometheus` module, look at the valgrind one. 
- Try running some benchmarks locally, using the prometheus endpoint, to see how it works. See the `subsystem-bench` README for instructions. If it's not clear how to run benchmarks, don't hesitate to ask, it would be a good reason to update unclear points in the instruction.

PR #3684, opened 2024-03-13
Implement crypto byte array newtypes in term of a shared type

davxy:
Introduces `CryptoBytes` type defined as:

```rust
pub struct CryptoBytes<const N: usize, Tag = ()>(pub [u8; N], PhantomData<fn() -> Tag>);
```

The type implements a bunch of methods and traits which are typically expected from a byte array newtype
(NOTE: some of the methods and trait implementations IMO are a bit redundant, but I decided to maintain them all to not change too much stuff in this PR)

It also introduces two (generic) typical consumers of `CryptoBytes`: `PublicBytes` and `SignatureBytes`.

```rust
pub struct PublicTag;
pub PublicBytes<const N: usize, CryptoTag> = CryptoBytes<N, (PublicTag, CryptoTag)>;

pub struct SignatureTag;
pub SignatureBytes<const N: usize, CryptoTag> = CryptoBytes<N, (SignatureTag, CryptoTag)>;
```

Both of them use a tag to differentiate the two types at a higher level. Downstream specializations will further specialize using a dedicated crypto tag. For example in ECDSA:

```rust
pub struct EcdsaTag;

pub type Public = PublicBytes<PUBLIC_KEY_SERIALIZED_SIZE, EcdsaTag>;
pub type Signature = PublicBytes<PUBLIC_KEY_SERIALIZED_SIZE, EcdsaTag>;
```

Overall we have a cleaner and most importantly **consistent** code for all the types involved

All these details are opaque to the end user which can use `Public` and `Signature` for the cryptos as before


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5523939

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5524096

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5524098

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5536282

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5536247

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5536836

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5540626

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5541121

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5542908

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5542907

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5548995

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -0,0 +1,233 @@
...
+		if data.len() != N {
+			return Err(())
+		}
+		let mut r = [0u8; N];
+		r.copy_from_slice(data);
+		Ok(Self::from_raw(r))
+	}
+}
+
+impl<const N: usize, T> UncheckedFrom<[u8; N]> for CryptoBytes<N, T> {
```
michalkucharczyk, line 148: nit: do we need unchecked_from?

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -0,0 +1,233 @@
...
+pub struct PublicTag;
+
+/// Generic encoded public key.
+pub type PublicBytes<const N: usize, SubTag> = CryptoBytes<N, (PublicTag, SubTag)>;
+
+/// Tag used for generic signature bytes.
+pub struct SignatureTag;
+
+/// Generic encoded signature.
+pub type SignatureBytes<const N: usize, SubTag> = CryptoBytes<N, (SignatureTag, SubTag)>;
```
michalkucharczyk, line 233: nit: it is all mostly wrappers, but maybe you could add some tests? However I am not 100% they are actually necessary. Feel free to ignore.

michalkucharczyk (approved):
lgtm!

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -0,0 +1,233 @@
...
+use core::marker::PhantomData;
+use scale_info::TypeInfo;
+use sp_runtime_interface::pass_by::{self, PassBy, PassByInner};
+
+/// Generic byte array.
+///
+/// The type is generic over a constant length `N` and a "tag" `T` which
+/// can be used to specialize the byte array without using newtypes.
+#[derive(Encode, Decode, MaxEncodedLen)]
+pub struct CryptoBytes<const N: usize, T = ()>(pub [u8; N], PhantomData<fn() -> T>);
```
koute, line 35:
```suggestion
#[derive(Encode, Decode, MaxEncodedLen)]
#[repr(transparent)]
pub struct CryptoBytes<const N: usize, T = ()>(pub [u8; N], PhantomData<fn() -> T>);
```

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -0,0 +1,233 @@
...
+	}
+}
+
+impl<const N: usize, T> From<CryptoBytes<N, T>> for [u8; N] {
+	fn from(v: CryptoBytes<N, T>) -> [u8; N] {
+		v.0
+	}
+}
+
+impl<const N: usize, T> AsRef<[u8; N]> for CryptoBytes<N, T> {
```
koute, line 123: You could also add `AsMut<[u8; N]>`.

substrate/primitives/core/src/crypto_bytes.rs
davxy: I think that the original intent of `UncheckedFrom` was to make explicit the fact that the construction of the object is performed without actually validating the content (e.g. in case of EC based crypto that the value represents a valid point on the used curve).

While this may have sense, should be said that the same types were exposing constructors like `CryptoBytes::from_raw([u8; N])` or `ByteArray::from_raw([u8; N])` where the docs explicitly states that the check is not performed.

IMHO may be sufficient to document that the constructors from a raw byte array doesn't ensure validity of  the bytes when these are then tried to be deserialized into the higher level object.

What could go wrong if you use a bad byte array:I've not removed it yet (as other traits and redundant stuff) to keep this PR as minimal as possible
- Signatures: nothing. The verification simply fails.
- Public keys: if used to verify something. It will fail.

So, in both cases, the failure is postponed to the usage stage instead of the construction phase.

May instead have sense to remove the `From<[u8; 32]>`? And keep `UncheckedFrom` only?

michalkucharczyk: >May instead have sense to remove the From<[u8; 32]>? And keep UncheckedFrom only?

I see your point, it makes sense. Being explicit here is better. On the other hand `from` is tempting as it is shorter.

But still it is a nit, cleanup could be done in some follow-up.

koute: > May instead have sense to remove the `From<[u8; 32]>`? And keep `UncheckedFrom` only?

Hmmm... it could make sense, but I guess maybe only if we remove all of the other easy ways to make a `T` in an unchecked fashion?

Also, [in my refactoring PR](https://github.com/paritytech/polkadot-sdk/pull/3689) I'm using these `From` impls to convert between raw arrays when passing these types through the FFI boundary between the runtime and the host. So if we remove `UncheckedFrom` I'll have to add another wrapper strategy type just for passing these using the `UncheckedFrom`. Not a big deal, but it is extra code that will have to be added.

davxy: I'd also like to add that the wrapped type is public. So maybe having this "UncheckedFrom" doesn't really provide a lot of safety?

skunert (approved):

substrate/primitives/core/src/sr25519.rs
```diff
@@ -54,22 +51,17 @@ const SIGNING_CTX: &[u8] = b"substrate";
...
 pub const CRYPTO_ID: CryptoTypeId = CryptoTypeId(*b"sr25");
 
+/// The byte length of public key
+pub const PUBLIC_KEY_SERIALIZED_SIZE: usize = 32;
+
+/// The byte length of signature
+pub const SIGNATURE_SERIALIZED_SIZE: usize = 64;
+
+#[allow(missing_docs)]
+pub struct Sr25519Tag;
```
bkchr, line 61: Missing docs? Or make it doc hidden. But currently it is "confusing"

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -0,0 +1,240 @@
...
+use scale_info::TypeInfo;
+use sp_runtime_interface::pass_by::{self, PassBy, PassByInner};
+
+/// Generic byte array.
+///
+/// The type is generic over a constant length `N` and a "tag" `T` which
+/// can be used to specialize the byte array without using newtypes.
+#[derive(Encode, Decode, MaxEncodedLen)]
+#[repr(transparent)]
+pub struct CryptoBytes<const N: usize, T = ()>(pub [u8; N], PhantomData<fn() -> T>);
```
bkchr, line 36: Some code comment explaining the PhantomData and why it looks like this. I forgot the actual name of this and also how it actually unfolds. I only remember that it changes if it is sync or send etc.

There exists some doc page in rust, maybe worthwhile to link this one.

bkchr (approved):

substrate/primitives/core/src/crypto_bytes.rs
davxy: https://github.com/paritytech/polkadot-sdk/pull/3684/commits/cd089eabd554fc85853dc0a39b659f6a60b61f11

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -27,10 +27,14 @@ use core::marker::PhantomData;
...
-/// Generic byte array.
+/// Generic byte array holding some crypto-related raw data.
 ///
 /// The type is generic over a constant length `N` and a "tag" `T` which
-/// can be used to specialize the byte array without using newtypes.
+/// can be used to specialize the byte array without requiring newtypes.
+///
+/// The tag `T` is held in a `PhantomData<fn() ->T>`, a trick allowing
+/// `CryptoBytes` to be `Send` and `Sync` regardless of `T` properties
+/// ([ref](https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns)).
```
michalkucharczyk, line 37: nice :)

PR #3689, opened 2024-03-14
Refactor the host <-> runtime interface machinery

koute:
This PR refactors the way the host functions are defined.

Previously the way a given type was marshalled through the host <-> runtime boundary was hardcoded for every type by the virtue of it implementing the relevant conversion traits.

This had two major consequences:
   * It was not obvious how a given type is going to be passed just by looking at its type alone, masking potentially expensive marshalling strategies. (For example, returning `Option<u32>` was done through the SCALE codec and involved extra memory allocations!)
   * It was not possible to use multiple marshalling strategies for a single type, making some of the future improvements we'd like to do (e.g. move the runtime memory allocator into the runtime) very hard to do.

So this PR disentangles this mess and makes the marshalling strategies explicit. This makes it much more clear how a given type in a given host function is marshalled, and also makes it possible to use different marshalling strategies for the same type in different host functions.

Before this PR you'd define a host function like this:

```rust
    #[runtime_interface]
    trait MyInterface {
        fn say_hello_world(name: &str) {
            println!("Hello {name}!");
        }
    }
```

and after this PR you'll define it like this:

```rust
    #[runtime_interface]
    trait MyInterface {
        fn say_hello_world(name: PassFatPointerAndRead<&str>) {
            println!("Hello {name}!", name);
        }
    }
```

In this case the strategy for passing the `&str` is now explicitly specified (`PassFatPointerAndRead`). Note that the *actual* API generated by this macro and the way arguments are accessed is completely unchanged! The `#[runtime_interface]` machinery automatically "strips" away the marshalling strategy wrappers, so neither the body of the `say_hello_world` function here nor its callers need to be changed.

This is a breaking change only for people who use `#[runtime_interface]` to define their own custom host functions.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5526091

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5526239

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5526830

koute:
(Sorry, had to force push since the initial commit wasn't properly signed for some reason so I had to resign in; no extra changes to the PR.)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5526980

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5527408

substrate/primitives/statement-store/src/runtime_api.rs
```diff
@@ -179,7 +219,7 @@ pub trait StatementStore {
 	}
 
 	/// Remove a statement from the store by hash.
-	fn remove(&mut self, hash: &Hash) {
+	fn remove(&mut self, hash: PassPointerAndRead<&Hash, 32>) {
```
michalkucharczyk, line 222: nit: maybe `std::mem::size_of::<Hash>()` instead of hard-coded 32?

substrate/primitives/state-machine/src/in_memory_backend.rs
```diff
@@ -21,11 +21,17 @@ use crate::{
...
 	TrieBackendBuilder,
 };
+use alloc::{collections::BTreeMap, vec::Vec};
 use codec::Codec;
 use hash_db::Hasher;
 use sp_core::storage::{ChildInfo, StateVersion, Storage};
 use sp_trie::{empty_trie_root, LayoutV1, PrefixedMemoryDB};
-use std::collections::{BTreeMap, HashMap};
+
+#[cfg(feature = "std")]
```
michalkucharczyk, line 30: is it intentional to not use `substrate_runtime`?

substrate/primitives/statement-store/src/runtime_api.rs
koute: This was deliberately done to make it explicit how much data is actually being passed.

Normally, yes, I agree, you'd use `size_of` to make sure that if the size of the type changes then the code doesn't need to be updated, but notice here that *we never want this to change*! If the size of `Hash` changes then this would be a backwards incompatible change, and any historical block which uses this host function would explode.

So making this explicit is not only there to make it immediately obvious how much data is being passed, but also as an additional safeguard to protect against unintended breakage.

substrate/primitives/state-machine/src/in_memory_backend.rs
koute: Yes, because `HashMap` is only available under `std`.

In general this is still quite messy because I only converted a subset of `feature = "std"` into `substrate_runtime`. I had to do some extra changes related to this (and this includes changes to the file you've commented on here) in this PR to make the `full_crypto` feature work again. I will be continuing with more of this refactoring in future PRs.

(Also, a side note, I think we might be able to actually remove the `full_crypto` feature altogether. The whole point of it was to make the code use real crypto libraries as-is instead of calling the host functions when compiled in a `no_std` environment. But once we switch to `substrate_runtime` to detect whether we're compiling a runtime then compiling in `no_std` will by default act like if the `full_crypto` is enabled.)

koute:
FYI, I'm currently running a full sync Polkadot bult-in on my VPS with this PR to see if anything breaks; it's currently at:

```
2024-03-19 10:32:55 ⚙️  Syncing  8.8 bps, target=#19966048 (24 peers), best: #18581007 (0x3b3c…e39e), finalized #18580992 (0xdbe6…c99b), ⬇ 510.5kiB/s ⬆ 0.7kiB/s
```

So it looks like there's no breakage as it's *almost* finished. I'll report back again once it finishes syncing completely.

michalkucharczyk (approved):

substrate/primitives/runtime-interface/test-wasm/src/lib.rs
```diff
@@ -56,76 +63,76 @@ pub trait TestApi {
...
-	fn invalid_utf8_data(_: &str) {}
+	fn invalid_utf8_data(_: PassFatPointerAndRead<&str>) {}
 
 	/// Overwrite the native implementation in wasm. The native implementation always returns
 	/// `false` and the replacement function will return always `true`.
 	fn overwrite_native_function_implementation() -> bool {
 		false
 	}
 
-	/// Gets an `u128` and returns this value
```
michalkucharczyk, line 119: nit: maybe it is worth keeping some test on how to marshal 128 bit integers in backward compatible way - just for reference purposes?

koute: Hmm... well, I got rid of this because marshaling `u128`s is not supported anymore, and it's not supported because a) it used the allocator (which was totally silly, as it's a small, constant length type), and b) nothing used it anyway**. (:

If anyone complains I'll add it back in a way that doesn't use the allocator.

** - besides one host function in the benchmarking machinery, which I modified in this PR.

substrate/primitives/io/src/lib.rs
```diff
@@ -284,10 +302,10 @@ pub trait Storage {
 	#[version(3, register_only)]
 	fn clear_prefix(
 		&mut self,
-		maybe_prefix: &[u8],
-		maybe_limit: Option<u32>,
-		maybe_cursor: Option<Vec<u8>>, //< TODO Make work or just Option<Vec<u8>>?
-	) -> MultiRemovalResults {
+		maybe_prefix: PassFatPointerAndRead<&[u8]>,
+		maybe_limit: PassByCodec<Option<u32>>,
+		maybe_cursor: PassByCodec<Option<Vec<u8>>>, //< TODO Make work or just Option<Vec<u8>>?
```
skunert, line 307: I know this TODO comment has been here before, but is it clear what it means and should it stay?

substrate/primitives/runtime-interface/src/impls.rs
```diff
@@ -57,37 +45,54 @@ macro_rules! impl_traits_for_primitives {
 			/// The type is passed directly.
 			impl RIType for $rty {
 				type FFIType = $fty;
+				type Inner = Self;
+			}
+
+			impl<'a> RIType for &'a $rty {
```
skunert, line 51: Is it intentional that this is still around? I think you said that it should be removed.

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
+}
 
-	fn from_ffi_value(_: &mut dyn FunctionContext, arg: Self::FFIType) -> Result<T> {
-		T::try_from(arg as u8).map_err(|_| format!("Invalid enum discriminant: {}", arg))
+#[cfg(not(substrate_runtime))]
+impl<T> IntoFFIValue for AllocateAndReturnFatPointer<T>
+where
+	T: AsRef<[u8]>,
+{
+	fn into_ffi_value(value: Self::Inner, context: &mut dyn FunctionContext) -> Result<u64> {
```
skunert, line 607: Nit: Just to be consistent with the other impls.
```suggestion
	fn into_ffi_value(value: Self::Inner, context: &mut dyn FunctionContext) -> Result<Self::FFIType> {
```

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
+	type FFIType = u32;
+	type Inner = T;
+}
+
+#[cfg(not(substrate_runtime))]
+impl<T, const N: usize> IntoFFIValue for AllocateAndReturnPointer<T, N>
+where
+	T: AsRef<[u8]>,
+{
+	fn into_ffi_value(value: Self::Inner, context: &mut dyn FunctionContext) -> Result<u32> {
```
skunert, line 566: Nit: Just to be consistent with the other impls. 
```suggestion
	fn into_ffi_value(value: Self::Inner, context: &mut dyn FunctionContext) -> Result<Self::FFIType> {
```

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
+		let value = value.as_ref();
+		let addr = context.allocate_memory(value.len() as u32)?;
```
skunert, line 568: Here we assume that `value.len() == N`, correct?

What if someone messes up and defines `AllocateAndReturnPointer<H256, 64>` as return type. Does that matter? We would allocate 32 bytes of memory here and try to read 64 bytes on the runtime side?

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
 ///
-/// The `u64` value is build by `length 32bit << 32 | pointer 32bit`
+/// This encodes a `T` into a `Vec<u8>` using the SCALE codec and then
+/// passes a pointer to that byte blob and its length to the host,
+/// which then reads it and decodes back into `T`.
 ///
-/// `Self` is encoded and the length and the pointer are taken from the encoded vector.
-impl<T: codec::Codec> RIType for Codec<T> {
+/// Raw FFI type: `u64` (a fat pointer; upper 32 bits is the size, lower 32 bits is the pointer)
+pub struct PassByCodec<T>(PhantomData<T>);
```
skunert, line 342: Nit: Maybe `PassFatPointerEncoded`?

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
+	type Destructor = Vec<u8>;
+
+	fn into_ffi_value(value: &mut Self::Inner) -> (Self::FFIType, Self::Destructor) {
+		let data = codec::Encode::encode(value);
+		(pack_ptr_and_len(data.as_ptr() as u32, data.len() as u32), data)
+	}
+}
+
+/// A trait signifying a primitive type.
+pub unsafe trait Primitive: Copy {}
```
skunert, line 427: DQ: Why does this need to be unsafe?

substrate/primitives/runtime-interface/src/pass_by.rs
```diff
@@ -15,419 +15,667 @@
...
 ///
-///     fn try_from(val: u8) -> Result<Test, ()> {
-///         match val {
-///             0 => Ok(Test::Test1),
-///             1 => Ok(Test::Test2),
-///             _ => Err(()),
-///         }
-///     }
-/// }
+/// THIS STRATEGY IS DEPRECATED; DO NOT USE FOR NEW HOST FUNCTIONS!
```
skunert, line 592: Should we point towards what potential users should use instead of this?

koute:
FYI, the burn-in with this PR applied successfully finished syncing Polkadot and is still running.

substrate/primitives/runtime-interface/src/pass_by.rs
koute: Yes.

It's possible, although unlikely as they would then get a compile time failure when compiling the runtime (since that requires `T: From<[u8; N]>`). But I guess I can add an `assert`?

substrate/primitives/runtime-interface/src/impls.rs
koute: Yep. Removed it!

substrate/primitives/runtime-interface/src/pass_by.rs
koute: I renamed `PassSliceRefByCodec` -> `PassFatPointerAndDecodeSlice` and `PassByCodec` -> `PassFatPointerAndDecode`.

These are might be a little too long, but at least they're consistent with the others.

substrate/primitives/runtime-interface/src/pass_by.rs
koute: I added a short note.

substrate/primitives/runtime-interface/src/pass_by.rs
koute: Hmm... good point, I think in one of the previous versions of the PR this made sense, but now I guess it doesn't. (:

I made it safe.

substrate/primitives/io/src/lib.rs
koute: Hmm.... no, I have no idea what it means. :P

This host function is `register_only` which means that it's not used. I tried to search around, but I couldn't find any docs as to whether there are any *concrete* plans to actually use it. It's [not in the spec](https://spec.polkadot.network/chap-host-api#id-ext_storage_clear_prefix) and there's no open RFC for it.

It was originally added in [this PR](https://github.com/paritytech/substrate/pull/11490), and there's this [TODO here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/src/storage/unhashed.rs#L145) to eventually enable it.

So either someone actually pushes it forward, or we delete it.

cc @ggwpez @KiChjang Do you guys perhaps know if there are any plans to do anything (= write an RFC) with this host function?

skunert (approved):

michalkucharczyk (approved):

skunert (approved):

s0me0ne-unkn0wn:
Superceded by #7375

PR #3690, opened 2024-03-14
Add elastic scaling support in ParaInherent BenchBuilder 

sandreim:
Extracted Benchbuilder enhancements used in https://github.com/paritytech/polkadot-sdk/pull/3644 . Might still require some work to fully support all scenarios when disputing elastic scaling parachains, but it should be useful in writing elastic scaling runtime tests.

### DIFF
--- a/polkadot/runtime/parachains/src/builder.rs
+++ b/polkadot/runtime/parachains/src/builder.rs
@@ -329,7 +329,7 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 
 	/// Create an `AvailabilityBitfield` where `concluding` is a map where each key is a core index
 	/// that is concluding and `cores` is the total number of cores in the system.
-	fn availability_bitvec(concluding: &BTreeMap<u32, u32>, cores: u32) -> AvailabilityBitfield {
+	fn availability_bitvec(concluding: &BTreeMap<u32, u32>, cores: usize) -> AvailabilityBitfield {
 		let mut bitfields = bitvec::bitvec![u8, bitvec::order::Lsb0; 0; 0];
 		for i in 0..cores {
 			if concluding.get(&(i as u32)).is_some() {
@@ -360,7 +360,7 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 	///
 	/// Note that this must be called at least 2 sessions before the target session as there is a
 	/// n+2 session delay for the scheduled actions to take effect.
-	fn setup_para_ids(cores: u32) {
+	fn setup_para_ids(cores: usize) {
 		// make sure parachains exist prior to session change.
 		for i in 0..cores {
 			let para_id = ParaId::from(i as u32);
@@ -415,6 +415,10 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 		mut self,
 		target_session: SessionIndex,
 		validators: Vec<(T::AccountId, ValidatorId)>,
+		// Total cores used in the scenario
+		total_cores: usize,
+		// Additional cores for elastic parachains
+		extra_cores: usize,
 	) -> Self {
 		let mut block = 1;
 		for session in 0..=target_session {
@@ -449,7 +453,8 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 		self.validators = Some(validators_shuffled);
 		self.block_number = block_number;
 		self.session = target_session;
-		// assert_eq!(paras::Pallet::<T>::parachains().len(), total_cores as usize);
+
+		assert_eq!(paras::Pallet::<T>::parachains().len(), total_cores - extra_cores);
 
 		self
 	}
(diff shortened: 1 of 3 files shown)


polkadot/runtime/parachains/src/builder.rs
```diff
@@ -442,7 +449,7 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 		self.validators = Some(validators_shuffled);
 		self.block_number = block_number;
 		self.session = target_session;
-		assert_eq!(paras::Pallet::<T>::parachains().len(), total_cores as usize);
+		// assert_eq!(paras::Pallet::<T>::parachains().len(), total_cores as usize);
```
alindima, line 452: this needs to be removed, right?

polkadot/runtime/parachains/src/builder.rs
```diff
@@ -476,16 +484,27 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 			})
 			.collect();
 
+		let mut current_core_idx = 0u32;
+
 		for (seed, _) in concluding_cores.iter() {
```
alindima, line 489: concluding_cores here is really concluding_paras

polkadot/runtime/parachains/src/builder.rs
```diff
@@ -494,112 +513,144 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
...
-					.map(|val_idx| {
-						let public = validators.get(*val_idx).unwrap();
-						let sig = UncheckedSigned::<CompactStatement>::benchmark_sign(
-							public,
-							CompactStatement::Valid(candidate_hash),
-							&self.signing_context(),
-							*val_idx,
+
+				let para_id = ParaId::from(*seed);
+				prev_head = None;
```
alindima, line 536: I think the `prev_head` declaration should be moved here, to make it clear that the value doesn't need to persist across different paras

polkadot/runtime/parachains/src/builder.rs
```diff
@@ -624,7 +677,14 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 					.cloned()
 					.unwrap_or(self.target_session);
 
-				let (para_id, core_idx, group_idx) = self.create_indexes(seed);
+				let para_id = ParaId::from(seed);
+				let core_idx = CoreIndex::from(current_core_idx);
+				current_core_idx +=1;
```
alindima, line 682: the behaviour here is weird. `start..last` are now paraids. It'd be better if we had them represent cores and have this logic dispute specific cores (regardless of their paraids)

alindima (commented):
I suggest we also add the benchmark from https://github.com/paritytech/polkadot-sdk/pull/3644. The current code on master should already allow supplying multiple candidates if elastic scaling mvp feature is enabled (but only one will be backed). We have a test that does that: https://github.com/paritytech/polkadot-sdk/blob/d55d4f64b50c858bce6fa1a429f50d486cc67d49/polkadot/runtime/parachains/src/inclusion/tests.rs#L1671

polkadot/runtime/parachains/src/builder.rs
sandreim: yeah, but maybe it can still be checked if there are no elastic paras.

polkadot/runtime/parachains/src/builder.rs
sandreim: There is more work to be done to support disputes on cores of elastic paras. Assuming elastic scaling not used, these are 1:1 mapped to cores. so it should be fine.

polkadot/runtime/parachains/src/builder.rs
sandreim: added back the correct check accounting for extra cores from elastic paras.

polkadot/runtime/parachains/src/builder.rs
sandreim: yes, that was moved around and forgot to remove this line.

sandreim: fixed

alindima:
> Not sure which benchmark you are refering to. If you re talking about the big inherent, it just doesn't make sense, that is for testing purposes, we don't need that weight.

A variation of the big inherent thingy. With realistic values, which exercises multiple candidates for a para (this is already allowed by the runtime). Otherwise, I'll add it in https://github.com/paritytech/polkadot-sdk/pull/3479 (but that PR is getting quite big and I could use the help :D )

polkadot/runtime/parachains/src/builder.rs
alindima: yeah, I was suggesting have this PR include these changes if useful.

Anyway, I don't think this'd be useful for benchmarking. And for tests, I don't think I can make a lot of use of the BenchBuilder (besides for a test or two, which is still useful).

I already added a lot of tests for paras_inherent (for candidate sanitization mostly) which exercise corner cases that are not easily expressable in such a high-level interface.

So it's fine to keep as is

alindima (approved):

eskimor (approved):

PR #3693, opened 2024-03-14
Adding listing to Coretime Chain, Broker pallet

poppyseedDev:
These PR changes add a listing feature to the Broker pallet allowing the purchased cores to be listed and re-purchased.


substrate/frame/broker/src/lib.rs
```diff
@@ -410,6 +420,24 @@ pub mod pallet {
...
 			core: CoreIndex,
 		},
+		/// A new listing has been created.
+		ListingCreated {
+			region_id: RegionId,
+			owner: T::AccountId,
+			price: BalanceOf<T>,
+		},
+		/// A region has been successfully sold.
+		RegionSold {
```
ggwpez, line 430: A secondary market can already be built by trading CoreTime NFTs, or? Could be traded on KodaDot for example?  
Why do we need additional functionality in here?

poppyseedDev: For the sake of simplicity.

Szegoo (changes requested):
I don't believe the market should be embedded within a system chain. What you proposed here could make sense if we wanted parachains to rely on a set of individuals for Coretime procurement. 

However, if we want parachains to be truly decentralized systems, that cannot be the case. So another alternative would be for the parachain to collectively(via its governance) purchase Coretime. However, that isn't ideal either. Given that parachains are fully transparent systems, this could make front-running attacks very common.

I have written a potential solution to this problem [here](https://forum.polkadot.network/t/building-a-coretime-marketplace/6661).

bkchr:
Yeah given the pallets and the work already being done by @Szegoo I don't see that this pull request is required. If there is functionality missing when it comes to transferring the regions etc, we should invest time on this.

Nevertheless, thank you for the pr!

PR #3695, opened 2024-03-14
pallet-xcm: add new extrinsic for asset transfers using explicit reserve

acatangiu:
## Description

Add `transfer_assets_using_reserve()` for transferring assets from local chain to destination chain using an explicit reserve location (typically local Asset Hub).

By default, an asset's reserve is its origin chain. But sometimes we may want to explicitly use another chain as reserve (as long as allowed by runtime `IsReserve` filter).

This is very helpful for transferring assets with multiple configured reserves (such as Asset Hub ForeignAssets), when the transfer strictly depends on the used reserve.

E.g. For transferring Foreign Assets over a bridge, Asset Hub must be used as the reserve location.

## Tests

Added integration tests for showcasing transferring assets from parachain over bridge using local Asset Hub reserve.

## TODOs

- [ ] benchmarks for new extrinsic
- [ ] add same tests on Westend side
- [ ] optional: deduplicate test code

### DIFF
--- a/cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
+++ b/cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
@@ -56,10 +56,16 @@ parameter_types! {
 	pub const RelayLocation: Location = Location::parent();
 	// Local native currency which is stored in `pallet_balances``
 	pub const PenpalNativeCurrency: Location = Location::here();
-	pub const RelayNetwork: Option<NetworkId> = None;
+	// The Penpal runtime is utilized for testing with various environment setups.
+	// This storage item allows us to customize the `NetworkId` where Penpal is deployed.
+	// By default, it is set to `NetworkId::Rococo` and can be changed using `System::set_storage`.
+	pub storage RelayNetworkId: NetworkId = NetworkId::Rococo;
+	pub RelayNetwork: Option<NetworkId> = Some(RelayNetworkId::get());
 	pub RelayChainOrigin: RuntimeOrigin = cumulus_pallet_xcm::Origin::Relay.into();
-	// FIXME: should also work for Westend
-	pub UniversalLocation: InteriorLocation = [GlobalConsensus(NetworkId::Rococo), Parachain(ParachainInfo::parachain_id().into())].into();
+	pub UniversalLocation: InteriorLocation = [
+		GlobalConsensus(RelayNetworkId::get()),
+		Parachain(ParachainInfo::parachain_id().into())
+	].into();
 }
 
 /// Type for specifying how a `Location` can be converted into an `AccountId`. This is used
--- a/cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
+++ b/cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
@@ -58,7 +58,8 @@ parameter_types! {
 	pub const PenpalNativeCurrency: Location = Location::here();
 	pub const RelayNetwork: Option<NetworkId> = None;
 	pub RelayChainOrigin: RuntimeOrigin = cumulus_pallet_xcm::Origin::Relay.into();
-	pub UniversalLocation: InteriorLocation = [Parachain(ParachainInfo::parachain_id().into())].into();
+	// FIXME: should also work for Westend
+	pub UniversalLocation: InteriorLocation = [GlobalConsensus(NetworkId::Rococo), Parachain(ParachainInfo::parachain_id().into())].into();
 }
 
 /// Type for specifying how a `Location` can be converted into an `AccountId`. This is used
--- a/polkadot/xcm/pallet-xcm/src/lib.rs
+++ b/polkadot/xcm/pallet-xcm/src/lib.rs
@@ -86,6 +86,7 @@ pub trait WeightInfo {
 	fn new_query() -> Weight;
 	fn take_response() -> Weight;
 	fn claim_assets() -> Weight;
+	fn transfer_assets_using_reserve() -> Weight;
 }
 
 /// fallback implementation
@@ -168,6 +169,10 @@ impl WeightInfo for TestWeightInfo {
 	}
 
 	fn claim_assets() -> Weight {
+		Weight::from_parts(100_000_000, 0)
+	}
+
+	fn transfer_assets_using_reserve() -> Weight {
 		Weight::from_parts(100_000_000, 0)
 	}
 }
@@ -1297,15 +1302,16 @@ pub mod pallet {
 				FeesHandling::Separate { local_xcm, remote_xcm }
 			};
 
-			Self::build_and_execute_xcm_transfer_type(
-				origin,
-				dest,
+			let (local_xcm, remote_xcm) = Self::build_xcm_transfer_type(
+				origin.clone(),
+				dest.clone(),
 				beneficiary,
 				assets,
 				assets_transfer_type,
 				fees,
 				weight_limit,
-			)
+			)?;
+			Self::execute_xcm_transfer(origin, dest, local_xcm, remote_xcm)
 		}
 
 		/// Claims assets trapped on this pallet because of leftover assets during XCM execution.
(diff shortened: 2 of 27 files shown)


franciscoaguirre:
Just to double check. This two reserves issue happens because a token is teleported between two locations, right? It's not because a token is reserved transferred somewhere and then reserved transferred again from that main reserve into some sort of secondary reserve.

acatangiu:
> Just to double check. This two reserves issue happens because a token is teleported between two locations, right? It's not because a token is reserved transferred somewhere and then reserved transferred again from that main reserve into some sort of secondary reserve.

It is the former. Teleporting assets between chains requires the same level of trust as allowing those chains to act as reserves for said asset. With teleports, it is conceptually the same asset on both chains, unlike reserve-based transfers where the real asset stays put within its borders while outside the borders (i.e. destination chain) a derivative asset is used to represent it.

----

This PR adds utility function for **_reserve-based transfers_** (not teleports) of a given asset, but using one of the assets secondary reserve locations (any other chain where it can be teleported to/from).

I added some example scenarios to the PR description for more clarity.

acatangiu:
cc @vgeddes @alistair-singh

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5625935

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -930,23 +934,10 @@ pub mod pallet {
...
-						SetFeesMode { jit_withdraw: true },
-						InitiateTeleport { assets: Wild(AllCounted(count)), dest, xcm: Xcm(vec![]) },
-					]);
-					T::Weigher::weight(&mut message).map_or(Weight::MAX, |w| T::WeightInfo::teleport_assets().saturating_add(w))
-				}
-				_ => Weight::MAX,
-			}
-		})]
+		#[allow(deprecated)]
+		#[deprecated(
```
franciscoaguirre, line 938: I'd also add the deprecation notice to the doc comment, since that's sent as metadata to UIs.

Polkadot-Forum:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5779940

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5781298

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5782834

acatangiu:
@paritytech/xcm this is ready for review now

franciscoaguirre (approved):

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1648,33 +1692,101 @@ impl<T: Config> Pallet<T> {
...
+			ensure!(
+				!matches!(assets_transfer_type, TransferType::RemoteReserve(_)),
+				Error::<T>::InvalidAssetUnsupportedReserve
+			);
+			let weight_limit = weight_limit.clone();
+			// remove `fees` from `assets` and build separate fees transfer instructions to be
+			// added to assets transfers XCM programs
+			let fee_asset_index =
+				assets.iter().position(|a| a.id == fees.id).ok_or(Error::<T>::FeesNotMet)?;
+			let fees = assets.remove(fee_asset_index);
```
bkontur, line 1736: I am not sure here, if this is a problem, but input `fees: Asset` amount is ignored in this branch, just the `assets.remove(fees.id);`, 

I am just thinking if we shouldn't just first remove `fees: Asset` amount from `assets` and when `0`, then remove  the whole asset by index.

Something like this pseudo-code:
```
let fees = {
    // find `fees.id` in `assets` -> found_fees_asset
    let mut found_fees_asset = assets.get_mut(fees.id)
   
    // just remove amount of `fees` from `found_fees_asset`
    found_fees_asset.fun = found_fees_asset.fun - fees.fun;

    // if we removed all amount, then also remove `fees.id` from `assets`
    if found_fees_asset.fun == 0 {
        assets.remove(fees.id);
    }

    // return untouched `input `fees: Asset``
    fees
} 
```

E.g. `AssetsInHolding` has similar function for removing Asset [`fn checked_sub(mut self, asset: Asset)`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-executor/src/assets.rs#L414)

---
For example, iiuc, I want to transfer this:
```
transfer_assets_using_type(
assets: [assets1(1000 tokens), assets2(5000 tokens)]
fees: [assets1(200 tokens)]
)
```

Just removing the index results in transferring:
```
assets: [assets2(5000 tokens)]
fees: [assets1(200 tokens)]
```

With my suggestion, it would result in transferring:
```
assets: [assets1(800 tokens), assets2(5000 tokens)]
fees: [assets1(200 tokens)]
```

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1443,6 +1400,92 @@ pub mod pallet {
...
+		/// - `weight_limit`: The remote-side weight limit, if any, for the XCM fee purchase.
+		#[pallet::call_index(15)]
+		#[pallet::weight(T::WeightInfo::transfer_assets())]
+		pub fn transfer_assets_using_type(
+			origin: OriginFor<T>,
+			dest: Box<VersionedLocation>,
+			beneficiary: Box<VersionedLocation>,
+			assets: Box<VersionedAssets>,
+			assets_transfer_type: Box<TransferType>,
+			fees: Box<VersionedAsset>,
```
bkontur, line 1451: Just thinking, what about changing this to AssetId?
```
- fees: Box<VersionedAsset>,
+ fees: Box<VersionedAssetId>,
```

It could maybe simplify and make clear my previous comment, so the all amounts would be contained on the one place - `assets`.

cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
```diff
@@ -58,9 +58,16 @@ parameter_types! {
 	pub const RelayLocation: Location = Location::parent();
 	// Local native currency which is stored in `pallet_balances``
 	pub const PenpalNativeCurrency: Location = Location::here();
-	pub const RelayNetwork: Option<NetworkId> = None;
+	// The Penpal runtime is utilized for testing with various environment setups.
+	// This storage item allows us to customize the `NetworkId` where Penpal is deployed.
+	// By default, it is set to `NetworkId::Rococo` and can be changed using `System::set_storage`.
+	pub storage RelayNetworkId: NetworkId = NetworkId::Westend;
```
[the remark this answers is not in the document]
acatangiu: Seeing how Rococo will be closed down soon in favor of https://github.com/paseo-network, I thought to default to Westend

cumulus/parachains/integration-tests/emulated/chains/parachains/testing/penpal/src/genesis.rs
```diff
@@ -80,15 +80,14 @@ pub fn genesis(para_id: u32) -> Storage {
 			assets: vec![
 				// Relay Native asset representation
 				(
-					Location::try_from(RelayLocation::get()).expect("conversion works"),
+					Location::try_from(RelayLocation::get()).unwrap(),
```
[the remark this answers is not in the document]
acatangiu: either way someone will be fixing conflicts so I don't see it as an advantage to do 4037 before 😆

polkadot/xcm/pallet-xcm/src/lib.rs
acatangiu: right, good catch!

polkadot/xcm/pallet-xcm/src/lib.rs
acatangiu: No, you are right that it's better to allow sending a particular asset that can also be used for fees but be able to specify _how much of it_ should be used for fees

acatangiu: Actually, I changed it to `AssetId` so it has same behavior as the existing extrinsics, only now we correctly identify asset to be used for fees using id instead of unreliable index in vec.

polkadot/xcm/pallet-xcm/src/lib.rs
acatangiu: actually did your other suggestion: https://github.com/paritytech/polkadot-sdk/pull/3695#discussion_r1561256820

acatangiu:
@bkontur addressed all comments, PTAL

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1443,6 +1399,86 @@ pub mod pallet {
...
+		///   relay to parachain, or `(parents: 2, (GlobalConsensus(..), ..))` to send from
+		///   parachain across a bridge to another ecosystem destination.
+		/// - `beneficiary`: A beneficiary location for the assets in the context of `dest`. Will
+		///   generally be an `AccountId32` value.
+		/// - `assets`: The assets to be withdrawn. This should include the assets used to pay the
+		///   fee on the `dest` (and possibly reserve) chains.
+		/// - `assets_transfer_type`: The XCM `TransferType` used to transfer the `assets`.
+		/// - `fees`: One of the included `assets` to be be used to pay fees.
+		/// - `fees_transfer_type`: The XCM `TransferType` used to transfer the `fees` assets.
+		/// - `weight_limit`: The remote-side weight limit, if any, for the XCM fee purchase.
```
bkontur, line 1441:
```suggestion
		/// Transfer assets from the local chain to the destination chain using explicit transfer
		/// types for assets and fees.
		///
		/// `assets` must have same reserve location or may be teleportable to `dest`. Caller must
		/// provide the `assets_transfer_type` to be used for `assets`:
		///  - `TransferType::LocalReserve`: transfer assets to sovereign account of destination
		///    chain and forward a notification XCM to `dest` to mint and deposit reserve-based
		///    assets to `beneficiary`.
		///  - `TransferType::DestinationReserve`: burn local assets and forward a notification to
		///    `dest` chain to withdraw the reserve assets from this chain's sovereign account and
		///    deposit them to `beneficiary`.
		///  - `TransferType::RemoteReserve(reserve)`: burn local assets, forward XCM to `reserve`
		///    chain to move reserves from this chain's SA to `dest` chain's SA, and forward another
		///    XCM to `dest` to mint and deposit reserve-based assets to `beneficiary`. Typically
		///    the remote `reserve` is Asset Hub.
		///  - `TransferType::Teleport`: burn local assets and forward XCM to `dest` chain to
		///    mint/teleport assets and deposit them to `beneficiary`.
		///
		/// Fee payment on the source, destination and all intermediary hops, is specified through
		/// `fees_id`, but make sure enough of the specified `fees_id` asset is included in the given list
		/// of `assets`. `fees_id` should be enough to pay for `weight_limit`. If more weight is needed
		/// than `weight_limit`, then the operation will fail and the sent assets may be at risk.
		///
		/// `fees_id` may use different transfer type than rest of `assets` and can be specified
		/// through `fees_transfer_type`.
		///
		/// - `origin`: Must be capable of withdrawing the `assets` and executing XCM.
		/// - `dest`: Destination context for the assets. Will typically be `[Parent,
		///   Parachain(..)]` to send from parachain to parachain, or `[Parachain(..)]` to send from
		///   relay to parachain, or `(parents: 2, (GlobalConsensus(..), ..))` to send from
		///   parachain across a bridge to another ecosystem destination.
		/// - `beneficiary`: A beneficiary location for the assets in the context of `dest`. Will
		///   generally be an `AccountId32` value.
		/// - `assets`: The assets to be withdrawn. This should include the assets used to pay the
		///   fee on the `dest` (and possibly reserve) chains.
		/// - `assets_transfer_type`: The XCM `TransferType` used to transfer the `assets`.
		/// - `fees_id`: One of the included `assets` to be be used to pay fees.
		/// - `fees_transfer_type`: The XCM `TransferType` used to transfer the `fees` assets.
		/// - `weight_limit`: The remote-side weight limit, if any, for the XCM fee purchase.
```

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/foreign_assets_transfers.rs
```diff
@@ -0,0 +1,608 @@
...
+	let test_args = TestContext {
+		sender: sender.clone(),
+		receiver: receiver.clone(),
+		args: TestArgs::new_para(
+			destination.clone(),
+			receiver.clone(),
+			native_amount_to_send,
+			assets.into(),
+			None,
+			fee_asset_item,
```
bkontur, line 348: Just thinking, do we want to change also this `TestArgs` to take just `fee_asset_id` instead of `fee_asset_item` index? Probably, it will cause lots of change.

acatangiu: not in this PR..

acatangiu:
bot fmt

bkontur (approved):
lgtm and lots of tests are added for new cases, cool :)

albertov19:
Hey @acatangiu, one scenario that is currently not covered by the `PolkadotXCM` pallet is when you try to transfer a token from Moonbeam to AH (reserve chain) that is not an XCM-fee token (not sufficient in AH). Moonbeam wants to pay that XCM execution with, for example, DOT.

Moonbeam only has the relay chain as DOT reserve, and we can't have AH because then we would have 2 reserve chains for 1 asset, which could create many new problems with SA's balances.

So, ideally, we could have a way in which we do something like:

```
Moonbeam (InitiateReserveWithdraw) -> Polkadot (WithdrawAsset, ClearOrigin, ....)

Polkadot  -> AssetHub (XCM Instructions to take out the Token and send it to address
```

The main problem with the approach above is the `ClearOrigin,` which deletes all information that the message is being routed from Moonbeam. Is there an alternate scenario that this can be done?

Mainly because we have new assets being created in AH, and they are not sufficient assets, meaning that to send them back, users need to pay with either USDC/USDT.

XTokens allowed for such an approach in which 1 XCM is sent to the Relay Chain to send DOT to Parachains SA account in AH while another XCM was sent to AH to withdraw DOT to buy execution to transfer the token. However, this approach relied on the first XCM message executing successfully, so teams would usually have a DOT buffer in AH SA.

albertov19:
> The silver lining here is that the problem is very generic and asset/chain agnostic, and we can decide on some mechanism to auto-balance these SAs, that every chain can use.
> E.g. a new `pallet_sa_management` where root/admin can register several SA locations per asset along with some % distribution for each, that periodically polls the remote balances of those SAs and moves liquidity around to stay within configured distribution %.

The main issue is SA management. It would then require users to go through undesirable routes to get their assets where they want to (for example, SA in Polkadot has 100 DOT, and in AH, it has 20 DOT, and a user wants to get 40 DOT in Polkadot). This creates bad UX.

What about creating a call in which you can route XCM and assets through another chain? I guess the application of this is too specific. And could we do something similar to what xTokens is doing, but just from a UI point of view?

Polkadot-Forum:

https://forum.polkadot.network/t/managing-sas-on-multiple-reserve-chains-for-same-asset/7538/1

acatangiu:
I created a forum topic to get more visibility and ideas on the problem: https://forum.polkadot.network/t/managing-sas-on-multiple-reserve-chains-for-same-asset/7538

xlc:
I raised this issue a year ago https://github.com/paritytech/cumulus/issues/2398

acatangiu:
> I raised this issue a year ago [paritytech/cumulus#2398](https://github.com/paritytech/cumulus/issues/2398)

Apparently it takes more than raising an issue to solve an ecosystem-wide problem :)

Please join the discussion on the forum, help me engage all the parachains, and collectively decide on a solution.

I can commit to implement whatever we, the ecosystem, decide is necessary, if help is required from Parity.

Polkadot-Forum:

https://forum.polkadot.network/t/xcm-user-and-developer-experience-improvements/4511/21

Polkadot-Forum:

https://forum.polkadot.network/t/rfc-xcm-asset-transfer-program-builder/8528/1

PR #3701, opened 2024-03-14
rpc: add option to `whitelist ips` in rate limiting

niklasad1:
This PR adds two new CLI options to disable rate limiting for certain ip addresses and whether to trust "proxy header".
After going back in forth I decided to use ip addr instead host because we don't want rely on the host header which can be spoofed but another solution is to resolve the ip addr from the socket to host name.

Example:

```bash
$ polkadot --rpc-rate-limit 10 --rpc-rate-limit-whitelisted-ips 127.0.0.1/8 --rpc-rate-limit-trust-proxy-headers
```

The ip addr is read from the HTTP proxy headers `Forwarded`, `X-Forwarded-For` `X-Real-IP` if `--rpc-rate-limit-trust-proxy-headers` is enabled if that is not enabled or the headers are not found then the ip address is read from the socket.

//cc @BulatSaif can you test this and give some feedback on it?


substrate/client/service/src/config.rs
```diff
@@ -108,6 +108,8 @@ pub struct Configuration {
 	pub rpc_batch_config: RpcBatchRequestConfig,
 	/// RPC rate limit per minute.
 	pub rpc_rate_limit: Option<NonZeroU32>,
+	/// RPC rate limit whitelisted hosts,
```
niklasad1, line 111:
```suggestion
	/// RPC rate limit whitelisted hosts.
```

substrate/client/cli/src/config.rs
```diff
@@ -349,6 +349,11 @@ pub trait CliConfiguration<DCV: DefaultConfigurationValues = ()>: Sized {
 		Ok(None)
 	}
 
+	/// RPC rate limit configuration.
```
niklasad1, line 352:
```suggestion
	/// RPC rate limit whitelisted hosts.
```

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5536089

substrate/client/rpc-servers/src/utils.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Helper to read the ip address of the client `X_FORWARDED_FOR` header.
+pub(crate) fn read_ip_from_proxy<B>(req: &Request<B>) -> Option<IpAddr> {
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+	//
+	// "X-Forwarded-For" returns a list of ip addresses
+	//
+	// X-Forwarded-For: 203.0.113.195,198.51.100.178
+	if let Some(ips) = req.headers().get(&X_FORWARDED_FOR).and_then(|v| v.to_str().ok()) {
+		if let Some(proxy_ip) = ips.split_once(',').and_then(|(v, _)| IpAddr::from_str(v).ok()) {
+			// NOTE: we assume that ip addr is global
```
niklasad1, line 45: checking if the ip address is global is a PITA, https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_global but shouldn't be an issue in practice. 

Enabling rate-limiting for a local rpc server doesn't make sense to me :)

BulatSaif:
Thank you, @niklasad1.

It seems we had different interpretations of "whitelisted hosts". I was expecting to see a filter to check "origins", similar to `--rpc-cors`, or a filter to check "Host" or `X-Forwarded-Host` headers. Since the RPC node is behind a proxy, I can manipulate the Host headers or origins on the proxy side, for example, adding authentication for one domain. However, based on the code implementation, it seems you implemented the IPs filter, which will also work for us. I will test it on Monody.

Given that we have an IP-based filter, can we rename the flag from:
```
polkadot --rpc-rate-limit-whitelisted-hosts localhost polkadot.js.org
```
to:
```
polkadot --rpc-rate-limit-excluded-ips 10.0.0.0/8 1.2.3.4/32
```
`hosts_to_ip_addrs` function can be removed.

niklasad1:
ah right, checking the `X-Forwarded-Host` header is slightly nicer and easier to implement. I'll take of it on Monday

niklasad1:
I'm a bit unsure about this now because it's easy for someone to just set the `X-Forwarded-For` or `X-Forwarded-Host` and we can't really check if whether it's spoofed or not which is quite easy to "bypass rate-limiting" if it's configured on server.

Thus, I think we need another CLI configuration that the node operation can set whether to trust certain proxies/headers or not.
If the node is not not behind a load balancer/reverse proxy then it doesn't make sense to trust `X-Forwarded-headers`.

Using the `X-Forwarded-Host` and `Origin header` makes me less sure about it and perhaps it's better to use ip anyway.

BulatSaif:
> I'm a bit unsure about this now because it's easy for someone to just set the X-Forwarded-For or X-Forwarded-Host and we can't really check if whether it's spoofed or not which is quite easy to do to "bypass rate-limiting" if it's configured on server.

Most proxies are set to drop X-Forwarded-For or X-Forwarded-Host by default. If a client tries to send `X-Forwarded-*`, it will be ignored.

My use case was a little bit different. I was planning to have two domains: `public-rpc.polkadot.io` with rate limiting and a proxy, and `internal-rpc.local` without rate limiting. Since the second domain is in the local zone, you cannot route outside traffic there.

niklasad1:
Got it but my only concern with the current code is that if a public node enables `--rpc-rate-limit-allow-host trusted.com` and is not behind a reverse proxy or anything.

It would trivial for a client to just set `X-Forwarded-Host: trusted.com` and fool the server to disable the rate-limiting for that connection.

So, I still think I want another flag for that to make it explicit but probably most likely that the server will be behind reverse proxy when rate-limiting is enabled. Lemme think about it a little more and see whether I can come up with something reasonable

niklasad1:
> @niklasad1 do you have any update?

oops, forgot about it, I'll take another iteration tomorrow

niklasad1:
> You can fool the server but not the proxy. If node is behind proxy (and it is always behind proxy since https certs are not supported) the X-Forwarded-Host will be set to correct host.

Yes, my point was that if one doesn't run the rpc node behind a reverse proxy that could be dangerous to trust the proxy headers because they could be spoofed.

I have now added another flag `---rpc-rate-limit-trust-proxy-headers` which users that know the node is behind a reverse proxy can set and without this flag only the remote addr of the socket is used.

jsdw (approved):
Codewise, this looks good to me so long as whitelisting based on IP is acceptable!

I do agree that the `rpc_rate_limit_trust_proxy_headers` flag is necessary; this prevents users who spin up nodes that are not in front of a load balancer from accidentally falling foul of malicious actors who realise they can pass those flags to bypass rate limiting :)

lexnv (approved):
LGTM! Nice one!

BulatSaif:
> > You can fool the server but not the proxy. If node is behind proxy (and it is always behind proxy since https certs are not supported) the X-Forwarded-Host will be set to correct host.
> 
> Yes, my point was that if one doesn't run the rpc node behind a reverse proxy that could be dangerous to trust the proxy headers because they could be spoofed.
> 
> I have now added another flag `--rpc-rate-limit-trust-proxy-headers` which users that know the node is behind a reverse proxy can set and without this flag only the remote addr of the socket is used.
> 
> Are you okay with these changes @BulatSaif?


Sorry, I was ooo last week, just saw this.

Yes, it is a good idea to add a flag to disable/enable trust to headers. I will test the PR on Monday and provide feedback.

bkchr:
Why are we adding all this stuff deeply into the node? What again was the excuse for this? Why isn't this all just being done by some proxy in front of the RPC port?

niklasad1:
> Why are we adding all this stuff deeply into the node? What again was the excuse for this? Why isn't this all just being done by some proxy in front of the RPC port?

I agree that ideally this should be done by a proxy but we decided to add a simple rpc-rate-limit middleware into the node because it's a tricky because the most proxies works on the HTTP layer. Thus, when the connection is upgraded to websocket connection the rate-limit etc from HTTP proxy doesn't restrict each RPC call/WebSocket message.

Thus, this PR just adds another CLI to disable such middleware for certain peers.

I'm aware that https://github.com/AcalaNetwork/subway would work but I don't know how stable it's but would definitely be great to remove this from the node if possible.

bkchr:
> I'm aware that https://github.com/AcalaNetwork/subway would work but I don't know how stable it's but would definitely be great to remove this from the node if possible.

Then we should start trying it.

BulatSaif:
@niklasad1, I tested it locally. Is it possible to accept IP addresses with a mask (10.0.0.0/8) or how to set a range of IPs like `10.0.0.0-10.255.255.255`?

```
docker run -it --rm paritypr/polkadot-debug:3701-b3787db0 --rpc-rate-limit-whitelisted-ips  10.0.0.0/8
error: invalid value '10.0.0.0/8' for '--rpc-rate-limit-whitelisted-ips <RPC_RATE_LIMIT_WHITELISTED_IPS>...': invalid IP address syntax
```

The remaining functions are working as expected. Thank you

niklasad1:
> I tested it locally. Is it possible to accept IP addresses with a mask (10.0.0.0/8) or how to set a range of IPs like 10.0.0.0-10.255.255.255?

I have changed it now and it requires the IP address to be in correct CIDR notation to support ranges good catch

BulatSaif (approved):

PR #3706, opened 2024-03-14
Extrinsic to restore corrupt staking ledgers

gpestana:
This PR adds a new extrinsic `Call::restore_ledger ` gated by `StakingAdmin` origin that restores a corrupted staking ledger. This extrinsic will be used to recover ledgers that were affected by the issue discussed in https://github.com/paritytech/polkadot-sdk/issues/3245.

The extrinsic will re-write the storage items associated with a stash account provided as input parameter. The data used to reset the ledger can be either i) fetched on-chain or ii) partially/totally set by the input parameters of the call.

In order to use on-chain data to restore the staking locks, we need a way to read the current lock in the balances pallet. This PR adds a `InspectLockableCurrency` trait and implements it in the pallet balances. An alternative would be to tightly couple staking with the pallet balances but that's inelegant (an example of how it would look like in [this branch](https://github.com/paritytech/polkadot-sdk/tree/gpestana/ledger-badstate-clean_tightly)).

More details on the type of corruptions and corresponding fixes https://hackmd.io/DLb5jEYWSmmvqXC9ae4yRg?view#/

We verified that the `Call::restore_ledger` does fix all current corrupted ledgers in Polkadot and Kusama. You can verify it here https://hackmd.io/v-XNrEoGRpe7APR-EZGhOA.

**Changes introduced**
- Adds `Call::restore_ledger ` extrinsic to recover a corrupted ledger;
- Adds trait `frame_support::traits::currency::InspectLockableCurrency` to allow external pallets to read current locks given an account and lock ID;
- Implements the `InspectLockableCurrency` in the pallet-balances.
- Adds staking locks try-runtime checks (https://github.com/paritytech/polkadot-sdk/issues/3751)

**Todo**
- [x] benchmark `Call::restore_ledger`
- [x] throughout testing of all ledger recovering cases
- [x] consider adding the staking locks try-runtime checks to this PR (https://github.com/paritytech/polkadot-sdk/issues/3751)
- [x] simulate restoring all ledgers (https://hackmd.io/Dsa2tvhISNSs7zcqriTaxQ?view) in Polkadot and Kusama using chopsticks -- https://hackmd.io/v-XNrEoGRpe7APR-EZGhOA

Related to https://github.com/paritytech/polkadot-sdk/issues/3245
Closes https://github.com/paritytech/polkadot-sdk/issues/3751


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537157

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537329

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537331

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537311

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537933

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537949

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537934

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5537932

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5556251

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5576589

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5590008

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5593770

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5594091

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5594092

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5594286

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5594320

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5597118

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5598028

gpestana:
bot bench substrate-pallet --pallet=pallet_staking

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5599068

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5603959

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5610582

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5611192

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5611340

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1982,85 @@ pub mod pallet {
...
+		#[pallet::weight(T::WeightInfo::reset_ledger())]
+		pub fn reset_ledger(
+			origin: OriginFor<T>,
+			stash: T::AccountId,
+			maybe_controller: Option<T::AccountId>,
+			maybe_total: Option<BalanceOf<T>>,
+		) -> DispatchResultWithPostInfo {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			let controller = Bonded::<T>::get(&stash).ok_or(Error::<T>::NotStash)?;
```
Ank4n, line 2009: if this controller is same as stash, we can return early as well (I think)?

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1982,85 @@ pub mod pallet {
...
+
+			let new_total = if let Some(new_total) = maybe_total {
+				match maybe_other_ledger.clone() {
+					// if there is a ledger associated with this stash, it is the *wrong* ledger.
+					// Ensure that the "other" ledger has their locks/total updated so that the
+					// imbalance in the locks between ledgers is 0 (i.e. no more locks are
+					// created/deleted).
+					Some(mut other_ledger) => {
+						// note: ledger.total == ledger's staking lock amount.
+						other_ledger.total = new_total
```
Ank4n, line 2034: shouldn't we check what is the locked amount for the other ledger? And only modify ledger amount based on the lock. Current code might lead to other ledger not in sync anymore.

Ank4n: Also it would probably be easier to read the code if we named all vars something like ledger_one, ledger_two, ledger_one_locked, ledger_one_stash and so on.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1982,85 @@ pub mod pallet {
...
+		///
+		/// The `maybe_*` input parameters will overwrite the corresponding data and metadata of the
+		/// ledger associated with the stash. If the input parameters are not set, the ledger will
+		/// be reset with the current values.
+		///
+		/// Upon successful execution, this extrinsic will *recreate* the ledger based on its past
+		/// state and/or the `maybe_controller` and `maybe_total` input parameters.
+		#[pallet::call_index(29)]
+		#[pallet::weight(T::WeightInfo::reset_ledger())]
+		pub fn reset_ledger(
```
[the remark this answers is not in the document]
gpestana: `recover_ledger` is better I agree.

substrate/frame/staking/src/pallet/mod.rs
gpestana: I think we could, but that's basically the same as the check in the line below. The check below is more comprehensive though.

substrate/frame/staking/src/pallet/mod.rs
gpestana: > Also it would probably be easier to read the code if we named all vars something like ledger_one, ledger_two, ledger_one_locked, ledger_one_stash and so on.

I will use the following convention: `other_ledger` and `ledger`, `other_controller`, `controller`.

gpestana: > shouldn't we check what is the locked amount for the other ledger? And only modify ledger amount based on the lock. Current code might lead to other ledger not in sync anymore.

The code is assuming that the `maybe_total` input will set the correct amount to the current ledger (`StakingAdmin` has the responsibility to define what correct amount is). This code makes sure that the correct amount is set and that there's an imbalance of 0 between locks, ie, if other_ledger exists, decrease it's total and lock (in `other_ledger.update()`.

If the `maybe_total > other.ledger.lock` (note that `lock == ledger.amount`), then the extrinsic fails since we can't guarantee the imbalance and the call was wrong.

It seems to me that we don't need to check the other ledger lock since the other ledger is not corrupted and we thus we guarantee that lock == ledger.amount.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5626232

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5632673

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5632829

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5632993

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5633009

substrate/frame/staking/src/pallet/mod.rs
gpestana: Closing this thread as the code has considerably changed too.

substrate/frame/staking/src/lib.rs
```diff
@@ -491,6 +491,16 @@ pub struct StakingLedger<T: Config> {
 	controller: Option<T::AccountId>,
 }
 
+/// State of a ledger with regards with its data and metadata integrity.
+#[derive(PartialEq, Debug)]
+enum LedgerIntegrityState {
+	/// Ledger and corresponding staking lock is OK.
+	Ok,
+	Corrupted,
```
gpestana, line 499: Add comments to all enum variants.

substrate/frame/staking/src/lib.rs
```diff
@@ -719,6 +729,12 @@ pub struct Nominations<T: Config> {
 	pub suppressed: bool,
 }
 
+impl<T: Config> Default for Nominations<T> {
```
gpestana, line 732: todo: remove Default, not used anymore.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -855,6 +856,8 @@ pub mod pallet {
 		BoundNotMet,
 		/// Used when attempting to use deprecated controller account logic.
 		ControllerDeprecated,
+		/// Cannot reset a ledger.
+		CannotResetLedger,
```
gpestana, line 860: todo: change name to `CannotRestoreLedger`.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
 
 			Ok(Some(T::WeightInfo::deprecate_controller_batch(controllers.len() as u32)).into())
 		}
+
+		/// Restores the state of a ledger which is in an inconsistent state.
+		///
+		/// The requirements to restore a ledger are the following:
+		/// * The stash is bonded;
```
gpestana, line 1990:
```suggestion
		/// * The stash is bonded;
		/// * The stash is not bonded but it has a staking lock left behind;
```

substrate/frame/balances/src/impl_currency.rs
```diff
@@ -918,3 +918,15 @@ where
...
 }
+
+impl<T: Config<I>, I: 'static> InspectLockableCurrency<T::AccountId> for Pallet<T, I> {
+	fn balance_locked(id: LockIdentifier, who: &T::AccountId) -> Self::Balance {
+		Self::locks(who)
+			.into_iter()
+			.filter(|l| l.id == id)
+			.collect::<Vec<_>>()
+			.pop()
+			.map(|l| l.amount)
```
Ank4n, line 929: could use reduce? It also (unsafely) assumes there is always one lock with an ID.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+		pub fn restore_ledger(
+			origin: OriginFor<T>,
+			stash: T::AccountId,
+			maybe_controller: Option<T::AccountId>,
+			maybe_total: Option<BalanceOf<T>>,
+		) -> DispatchResultWithPostInfo {
+			T::AdminOrigin::ensure_origin(origin)?;
+
+			// closure to call before return from extrinsic to ensure the restore was successful.
+			let integrity_checks = || {
```
Ank4n, line 2011: isn't `inspect_bond_state` doing something similar?

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -84,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+				Error::<T>::NotStash
+			} else {
+				Error::<T>::BadState
+			}
+		})?;
+
+		match Ledger::<T>::get(controller) {
+			Some(ledger) =>
+				if ledger.stash != *stash {
+					Ok(LedgerIntegrityState::Corrupted)
```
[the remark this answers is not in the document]
Ank4n: When this happens, another associated ledger is always killed. Is that correct? Then is it correct to say restore ledger should be called via the killed ledger which should also fix the lock the other ledger in a way that lock is always balanced (lock reduced in new ledger equal to lock increased in old/resurrected ledger)?

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+		///
+		/// The `maybe_*` input parameters will overwrite the corresponding data and metadata of the
+		/// ledger associated with the stash. If the input parameters are not set, the ledger will
+		/// be reset with the current values.
+		///
+		/// Upon successful execution, this extrinsic will *recreate* the ledger based on its past
+		/// state and/or the `maybe_controller` and `maybe_total` input parameters.
+		#[pallet::call_index(29)]
+		#[pallet::weight(T::WeightInfo::restore_ledger())]
+		pub fn restore_ledger(
```
Ank4n, line 2002: So this is like a throw away function right once the issues are fixed?

Ank4n: I think you earlier tried the approach where staking admin provides some values for the ledger, and this function applies them if it could verify its integrity. Would that approach not work? Seems a simpler and more general approach and potentially useful for future as well.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1843,14 +1875,44 @@ impl<T: Config> Pallet<T> {
...
 	}
 
+	/// Invariants:
+	///
+	/// * Staking locked funds for every bonded stash should be the same as its ledger's total.
+	///
+	/// NOTE: some of these checks result in warnings only. Once
+	/// <https://github.com/paritytech/polkadot-sdk/issues/3245> is resolved, turn warns into check
+	/// failures.
+	fn check_locks() -> Result<(), TryRuntimeError> {
```
kianenigma, line 1889: Does this pass now in all networks, excluding the few corrupted ledgers? it should

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1843,14 +1875,44 @@ impl<T: Config> Pallet<T> {
...
+		for (_, controller) in <Bonded<T>>::iter() {
+			if let Some(ledger) = <Ledger<T>>::get(&controller) {
+				let lock = <T::Currency as InspectLockableCurrency<T::AccountId>>::balance_locked(
+					crate::STAKING_ID,
+					&ledger.stash,
+				);
+
+				if lock != ledger.total {
+					// TODO before merging: use warn
```
kianenigma, line 1898: We don't catch warnings quite well yet, Error is actually safer, why switch to warn?

substrate/frame/staking/src/pallet/impls.rs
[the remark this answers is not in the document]
Ank4n: Ignore above, seems wrong. We always call via the original stash, and the overwritten ledger might be killed.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+				}
+
+				Ok::<_, Error<T>>(())
+			};
+
+			let current_lock = T::Currency::balance_locked(crate::STAKING_ID, &stash);
+
+			let (new_controller, new_total) = match Self::inspect_bond_state(&stash) {
+				Ok(LedgerIntegrityState::Corrupted) => {
+					let new_controller = maybe_controller.unwrap_or(stash.clone());
```
Ank4n, line 2029: This controller should not be same as the last controller right? Practically we cannot just insert a new controller here. I believe the only option here is to make controller same as stash?

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -84,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+			}
+		})?;
+
+		match Ledger::<T>::get(controller) {
+			Some(ledger) =>
+				if ledger.stash != *stash {
+					Ok(LedgerIntegrityState::Corrupted)
+				} else {
+					if lock != ledger.total {
+						Ok(LedgerIntegrityState::LockCorrupted)
```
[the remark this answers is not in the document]
kianenigma: if the stash is a mismatch, it means it is someone else's ledger, so I think it is kinda meaningless to even check `LockCorrupted`. `Corrupted` always also implies `LockCorrupted` as well. 

And I think we only have `Corrupted` now, or do we have any `LockCorrupted` atm as well?

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1843,14 +1875,44 @@ impl<T: Config> Pallet<T> {
...
 
+	/// Invariants:
+	///
+	/// * Staking locked funds for every bonded stash should be the same as its ledger's total.
+	///
+	/// NOTE: some of these checks result in warnings only. Once
+	/// <https://github.com/paritytech/polkadot-sdk/issues/3245> is resolved, turn warns into check
+	/// failures.
+	fn check_locks() -> Result<(), TryRuntimeError> {
+		for (_, controller) in <Bonded<T>>::iter() {
```
kianenigma, line 1890: why are you not using your `inspect_bond_state` here?

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+						);
+						new_total
+					} else {
+						current_lock
+					};
+
+					Ok((new_controller, new_total))
+				},
+				Ok(LedgerIntegrityState::CorruptedKilled) => {
+					if current_lock == Zero::zero() {
```
Ank4n, line 2047: Is there any case like this currently? If lock is zero and ledger is 0 why do we need to restore? Seems good state?

Affected user can just do a new bond. Otherwise this will allow staking admin to bond any non bonded account.

substrate/frame/staking/src/pallet/impls.rs
kianenigma:
```
Bonded::<T>::iter().all(|(stash, _)| Self::inspect_bond_state(stash).is_ok())
```

substrate/frame/staking/src/pallet/mod.rs
kianenigma: Yeah I also think there should be one function that defines whether a stash/ledger/controller triplet is correct, and it should be used everywhere, not re-defining these conditions.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+					return Ok(Pays::No.into());
+				},
+				Ok(LedgerIntegrityState::Ok) | Err(_) => Err(Error::<T>::CannotResetLedger),
+			}?;
+
+			// re-bond stash and controller tuple.
+			Bonded::<T>::insert(&stash, &new_controller);
+
+			// reset ledger state.
+			let mut ledger = StakingLedger::<T>::new(stash.clone(), new_total);
```
kianenigma, line 2087: it is worth noting that we erase any unlocking data for the recovered account.

substrate/frame/staking/src/pallet/mod.rs
Ank4n: Also would avoid bug if we simply this fn (seems too many code branches). OTOH it is gated by StakingAdmin, so safe-ish.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+					T::Currency::set_lock(
+						crate::STAKING_ID,
+						&stash,
+						new_total,
+						WithdrawReasons::all(),
+					);
+
+					Ok((stash.clone(), new_total))
+				},
+				Err(Error::<T>::BadState) => {
```
kianenigma, line 2073: If `stash` is a random account in staking, we return `BadState` and come to this arm right, and this arm provides free execution, which means I can endlessly spam the chain?

kianenigma: oh well, `AdminOrigin`, nvmd. But good for you to be aware of this detail.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+			Bonded::<T>::insert(&stash, &new_controller);
+
+			// reset ledger state.
+			let mut ledger = StakingLedger::<T>::new(stash.clone(), new_total);
+			ledger.controller = Some(new_controller);
+			ledger.update()?;
+
+			integrity_checks()?;
+
+			Ok(Pays::No.into())
```
kianenigma, line 2093: fwiw all this will do is to just refund the weight, as `AdminOrigin` is not "paying" anything anyhow.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+		/// be reset with the current values.
+		///
+		/// Upon successful execution, this extrinsic will *recreate* the ledger based on its past
+		/// state and/or the `maybe_controller` and `maybe_total` input parameters.
+		#[pallet::call_index(29)]
+		#[pallet::weight(T::WeightInfo::restore_ledger())]
+		pub fn restore_ledger(
+			origin: OriginFor<T>,
+			stash: T::AccountId,
+			maybe_controller: Option<T::AccountId>,
```
kianenigma, line 2005: I would for the sake of simplicity make the caller always provide all arguments. 

I am assuming that any stash that is eligible to `restore`-ed via this extrinsic is already in a "locked" state and they can change nothing, so everything can come in as an argument.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5665453

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5665452

substrate/frame/staking/src/pallet/impls.rs
gpestana: I refactored a bit the try state checks and added the `inspect_bond_state` for each bonded stash.

substrate/frame/staking/src/pallet/mod.rs
gpestana: Do you mean to also add an option to set the restored ledger's unlocking chunks? If so, I'm adding this now so we can set the ledger's chunks when restoring a ledger, if the caller wishes.

substrate/frame/staking/src/pallet/mod.rs
gpestana: I added a `maybe_unlocking` so that the called may decide to re-write the unlocking chunks instead of completely clearing them.

substrate/frame/staking/src/pallet/impls.rs
gpestana: The rationale is to throw warns until all the try state checks pass in Polkadot and Kusama to avoid raising alarms on a problem we already know about. And then revert to errs once the state is fixed https://github.com/paritytech/polkadot-sdk/issues/3728.

substrate/frame/staking/src/pallet/impls.rs
gpestana: On my list for a quick check tomorrow.

substrate/frame/staking/src/pallet/mod.rs
gpestana: Good point, I replaced the integrity check with a call to `inspect_bond_state` and also called it in the try-state checks. This should be our to-go method to ensure a bond/ledger is consistent.

substrate/frame/staking/src/pallet/mod.rs
gpestana: > Would that approach not work? Seems a simpler and more general approach and potentially useful for future as well.

It would, but since that's more loosely defined and just overwrites the ledger/bond/lock values, it's less safe (even when behind staking admin). So I see the options to reset the ledgers as one of (or a spectrum in-between of):

1. Have a `reset_ledger` or `set_ledger` call that basically re-writes any value in storage related to staking. This is a glorified `set_storage` for (some) staking storage items.
2. Have an extrinsic that will restore the stash/ledger state relying as much as possible on the on-chain state to learn what's the corruption case and its fix and to use as much of the current on-chain as possible (e.g. using the current locks of a stash to reset its ledger).

After mapping all the potential corruption cases, the second option seemed better and safer to me. I believe that it would be also easier to make the case for governance to review/accept the proposal (it relies more on current on-chain data, we won't need to present proofs for every single value re-written per ledger).

The option 2. loses in generality. But I think that's OK and perhaps we should rely on `set_code` or implement a `force_unbond/clean_ledger` (which does not rebond, only cleans) in the future.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -84,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+	pub(crate) fn inspect_bond_state(
+		stash: &T::AccountId,
+	) -> Result<LedgerIntegrityState, Error<T>> {
+		let lock = T::Currency::balance_locked(crate::STAKING_ID, &stash);
+
+		let controller = <Bonded<T>>::get(stash).ok_or_else(|| {
+			if lock == Zero::zero() {
+				Error::<T>::NotStash
+			} else {
+				Error::<T>::BadState
```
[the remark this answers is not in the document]
gpestana: > and this branch should be defensive as in to the best of our knowledge, no ledger should be in this error state right?

We actually may get into this state when the stash associated with a corrupted ledger calls kill on the ledger, this specific case is explained here: https://hackmd.io/DLb5jEYWSmmvqXC9ae4yRg#/5

substrate/frame/staking/src/pallet/impls.rs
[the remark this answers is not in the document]
gpestana: The reason to differentiate between `Corrupted` and `LockCorrupted` is that the `LockCorrupted` restore *requires* the `maybe_total` to be Some, i.e. in this case, we want to ensure that the lock and the ledger's total is re-written. On the other hand, `Corrupted` we will use the lock's value to reset the ledger's total (or, optionally, re-write it with `maybe_total`).

We could also collapse both and return an error if the total to reset the ledger is 0, ie.

```
let new_total = maybe_total.unwrap_or(lock);
ensure!(new_total != 0, Error::<T>::CannotResetLedger);
```

substrate/frame/staking/src/pallet/impls.rs
gpestana: However, I turned this check into an `ensure!`, so it will throw an error.

substrate/frame/staking/src/pallet/mod.rs
gpestana: >  If lock is zero and ledger is 0 why do we need to restore? Seems good state?

Not really. In this case, there is a lingering bonded entry for the stash. So basically:

```
Bonded(stash) = C;
Lock(stash) = 0;
Ledger(C) = None;
```

This may happen in the case when a corrupted ledger has been killed, e.g. [like in this case]https://hackmd.io/DLb5jEYWSmmvqXC9ae4yRg#/6).

substrate/frame/balances/src/impl_currency.rs
gpestana: I refactored it to use fold instead, much simpler.

substrate/frame/staking/src/pallet/mod.rs
gpestana: This is true, and in the [recovery cases I tested](https://hackmd.io/v-XNrEoGRpe7APR-EZGhOA?view), we don't need to set the controller for any of the ledgers. I kept the convention that we may be able to explicitly set all the items of the ledger through this extrinsic (and lock). This argument is not the strongest and I'm happy to reconsider it and remove the `maybe_controller`.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5667072

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,115 @@ pub mod pallet {
...
+
+			let current_lock = T::Currency::balance_locked(crate::STAKING_ID, &stash);
+
+			let (new_controller, new_total) = match Self::inspect_bond_state(&stash) {
+				Ok(LedgerIntegrityState::Corrupted) => {
+					let new_controller = maybe_controller.unwrap_or(stash.clone());
+
+					let new_total = if let Some(new_total) = maybe_total {
+						// enforce lock == ledger.amount.
+						T::Currency::set_lock(
```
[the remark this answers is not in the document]
gpestana: Yes, good idea. Done!

gpestana:
bot bench substrate-pallet --pallet=pallet_staking

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1945,19 +1978,18 @@ impl<T: Config> Pallet<T> {
...
 
 	/// Invariants:
-	/// * `ledger.controller` is not stored in the storage (but populated at retrieval).
 	/// * Stake consistency: ledger.total == ledger.active + sum(ledger.unlocking).
-	/// * The controller keying the ledger and the ledger stash matches the state of the `Bonded`
-	/// storage.
+	/// * The ledger's controller and stash matches the associated `Bonded` tuple.
+	/// * Staking locked funds for every bonded stash should be the same as its ledger's total.
+	/// * Staking ledger and bond are not corrupted.
 	fn check_ledgers() -> Result<(), TryRuntimeError> {
```
Ank4n, line 1985: How is the running time for this try state check? Don't mind putting this in release for now if it is not super long but want to point out that you are reading some storages multiple times and it can be optimized to be much faster. 
Ideally we read any storage only once and keep it in memory for further checks.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1980,6 +1983,106 @@ pub mod pallet {
...
+							crate::STAKING_ID,
+							&stash,
+							new_total.min(stash_balance),
+							WithdrawReasons::all(),
+						);
+						new_total
```
Ank4n, line 2024: Shouldn't this also return new_total.min(stash_balance)

substrate/frame/staking/src/pallet/mod.rs
Ank4n: In [this case](https://hackmd.io/DLb5jEYWSmmvqXC9ae4yRg#/6) `Lock(stash)` is equal to previous `ledger.total`.

kianenigma (approved):

substrate/frame/staking/src/pallet/mod.rs
gpestana: @Ank4n here I think it's more clear why we should not clear the bond in this case bur rather recover it https://github.com/paritytech/polkadot-sdk/pull/3866/files#diff-24281cfd6db327807105722f46ef26850905bc77759978a2fff784cb9379817cR7578

Ank4n (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/recover-corrupted-staking-ledgers-in-polkadot-and-kusama/9796/1

PR #3707, opened 2024-03-14
Crowdload burn remaining funds

ggwpez:
Crowdloan account should burn all funds after a crowd loan got dissolved to ensure that the account is reaped correctly.


liamaharon (approved):

polkadot/runtime/common/src/crowdloan/mod.rs
```diff
@@ -1740,6 +1746,41 @@ mod tests {
 		});
 	}
 
+	// Regression test to check that a pot account with just one provider can be dissolved.
+	#[test]
+	fn dissolve_provider_refs_ti_works() {
```
franciscoaguirre, line 1751:
```suggestion
	fn dissolve_provider_refs_total_issuance_works() {
```

polkadot/runtime/common/src/crowdloan/mod.rs
```diff
@@ -1629,9 +1634,10 @@ mod tests {
 
 			// Some funds are left over
 			assert_eq!(Balances::free_balance(&account_id), 10);
-			// They wil be left in the account at the end
+			// Remaining funds will be burned
 			assert_ok!(Crowdloan::dissolve(RuntimeOrigin::signed(1), para));
-			assert_eq!(Balances::free_balance(&account_id), 10);
+			assert_eq!(Balances::free_balance(&account_id), 0);
+			assert_eq!(Balances::total_issuance(), issuance - 10);
```
franciscoaguirre, line 1640: So sending tokens to the crowdloan address basically burns them?

franciscoaguirre (approved):
Makes sense

polkadot/runtime/common/src/crowdloan/mod.rs
ggwpez: Yep. It already did in the past, but the currency migration messed up the references which caused :(  
Users should never directly send funds but always use the pallet functions.

polkadot/runtime/common/src/crowdloan/mod.rs
```diff
@@ -576,7 +577,10 @@ pub mod pallet {
 			// can take care of that.
 			debug_assert!(Self::contribution_iterator(fund.fund_index).count().is_zero());
 
-			frame_system::Pallet::<T>::dec_providers(&Self::fund_account_id(fund.fund_index))?;
+			// Crowdloan over, burn all funds.
+			let _imba = CurrencyOf::<T>::make_free_balance_be(&pot, Zero::zero());
```
kianenigma, line 581: dq but I don't get why this is needed; a CL is dissolve-able only when all funds are already returned, no? if not, I think it should be like this.

seems to be checked as a part of `let can_dissolve = permitted && fund.raised.is_zero();`

So the invariant is `fund.raised` should always be equal to `CurrencyOf::<T>::free_balance(&pot)`?

kianenigma: Ahh okay, reading the test I realize that this is catering for the cases where a manual transfer happens to the fund account.

PR #3711, opened 2024-03-15
network:bridge: fix peer_count metric

alexggh:
The metric records the current protocol_version of the validator that just connected with the peer_map.len(), which contains all peers that connected, that has the effect the metric will be wrong since it won't tell us how many peers we have connected per version because it will always record the total number of peers

Fix this by counting by version inside peer_map, additionally because that might be a bit heavier than len(), publish it only on-active leaves.


s0me0ne-unkn0wn (approved):

ordian (approved):
This will delay updating the metric until the next block is produced, but it seems reasonable. If that is proven to be heavy/causing issues, we could put the `by_version_count` into the state and do incremental updates as before.

sandreim (approved):

Issue #266, opened 2022-08-08
Vision: Meta Transaction


xlc:
We need a way to support meta transaction to enable many useful scenarios. However meta transaction is can be very hard to implement correctly and securely.

TLDR; I don't know how to store nonce for accounts without any tokens without introducing new attack vector.

## Scope

Firstly we need to define a scope. Here are a list of the potential use cases that I can think of and we need to decide what do we want to support and what is out of the scope. Support all of the use cases may result the solution be too complicated.

### 1. Pay for transaction fee

Allow relayer to pay transaction fee for a user. This allows user account does not hold any fee token can still able to interact with the chain.

#### Use cases

- dApp onboard new users by allowing them to interact with the dApp contracts without paying transaction fee
- Create a better proxy account patten. Main account create limited proxy account A, send small fund on fee payment account B. Account A never holds any funds and can be rotated easily without need to move tx fee funds. All the transaction from A are broadcasted by B and B pays all the tx fee.
- Support any token to be tx fee token. User can craft `batch([sendCustomToken(relayerAddr, amount), doTheActualWork()])`.  And if it is profitable to pay tx fee for getting `amount` of custom token, relayer submit the tx to make the trade.  It is up to the relayer to ensure the sendCustomToken step are expected to be successful.

#### Flow

There are two possible flow to achieve the desired goal.

- User sign tx
- Send tx payload and signature to relayer
- Relayer sign the transaction and broadcast it
- Blockchain charges relayer for tx fee, validates the user signature, and dispatch the payload tx from user origin

OR

- User generate tx payload
- Relayer sign the tx payload
- User sign the tx payload with relayer signature and broadcast it
- Blockchain validates relayer signature, charges relayer for tx fee, and dispatch payload from user origin

### 2. Atomic multi origin tx

Allow atomically dispatch a transaction mixed from different origins. It can only be successfully executed IFF all the origins signs the whole payload.

#### Use cases

- Atomic exchange of assets between multiple users
  - e.g. `batch_all([as_user(alice, transfer(10 DOT, bob)), as_user(bob, transfer(100 aUSD, alice))])`
- Perform a sequence of actions from multiple users at once without revealing it to avoid front running
  - e.g. `batch([create_swap_pool(aUSD, DOT), as_user(alice, add_liquidity(100 aUSD, 10 DOT)), as_user(bob, add_liquidity(100 aUSD, 10 DOT))])` so that no one can make a trade between Alice add liquidity and Bob add liquidity

[...]

PR #3712, opened 2024-03-15
FRAME: Meta Transaction (extension based version)

muharem:
Experiment with a new Transaction Extension [model](https://github.com/paritytech/polkadot-sdk/pull/2280) that utilizes tx extensions to enable transaction fee sponsorship by an agent.

To see an example, refer to the `meta_tx_works` test case in `substrate/frame/transaction-payment/src/meta_tx.rs` file.

This implementation serves to demonstrate the concept and not production ready.

RFC: https://github.com/paritytech/polkadot-sdk/issues/4123


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5545227

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5545245

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5545246

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5545391

polkadot/runtime/westend/src/lib.rs
```diff
@@ -1563,6 +1569,144 @@ pub type TxExtension = (
...
+		use keyring::AccountKeyring;
+		use sp_io::hashing::blake2_256;
+		use sp_runtime::{
+			traits::{DispatchTransaction, TransactionExtensionBase},
+			MultiSignature,
+		};
+
+		// The part of `TxExtension` that has to be provided and signed by user who wants
+		// the transaciton fee to be sponsored by someone else.
+		type BaseTxExtension = (
```
kianenigma, line 1588: FYI there is https://paritytech.github.io/polkadot-sdk/master/frame/runtime/types_common/type.SystemSignedExtensionsOf.html exactly to simplify such setups :)

kianenigma:
This is a pretty good #3901 type of work, glad to see a PoC.

shawntabrizi:
Am I correct that this closes: https://github.com/paritytech/polkadot-sdk/issues/266

cc @xlc

substrate/frame/transaction-payment/src/lib.rs
```diff
@@ -943,3 +959,129 @@ where
...
+/// extensions. If signature is not provided, this extension is no-op.
+// TODO better doc.
+#[derive(
+	CloneNoBound, EqNoBound, PartialEqNoBound, Encode, Decode, RuntimeDebugNoBound, TypeInfo,
+)]
+#[codec(encode_bound())]
+#[codec(decode_bound())]
+pub struct SetFeeAgent<V: Verify>
+where
+	V: Codec + Debug + Sync + Send + Clone + Eq + PartialEq + StaticTypeInfo,
```
xlc, line 1008: better to avoid all those duplicated `Codec + Debug + Sync + Send + Clone + Eq + PartialEq + StaticTypeInfo`

xlc:
Does this solves some challenges I listed in #266?

> The common way to implement replay attack prevention mechanism is by having a sequence number / nonce that increments for each new message.
>
> However, this is incompatible on the default storage model of Substrate. The nonce have to be stored onchain and all the onchain storages must be covered with a deposit (ED). For some use cases, the signing account will not hold any balances at all and therefore shouldn't really be able to cover the storage deposit for nonce storage.

i.e. how is the nonce and ED handled?

acatangiu:
Possibly dumb question:

Why do this through transaction extensions? Even the tx ext set we have now is quite complicated and is seen as some "black magic" by many developers on Polkadot 😆.

Would this not be cleaner, clearer, less complicated to understand if it were implemented in a pallet?
Even the use-case itself seems to me a bit exotic to warrant "implementing it for everyone" through transaction extensions. I think beneficiaries of this feature would be perfectly satisfied going through some dedicated utility extrinsic, while the rest of Polkadot would appreciate less rather than even more complexity..

wdyt?

xlc:
@acatangiu I don't think that's possible but I hope I am wrong

muharem:
@xlc no, it does not. but I read the thread now, and I like what @shawntabrizi proposed. Basically implementation wont have any workaround for non-registered users, but will ensure there is a way to setup this feature, where for such users the provider reference can be used to register an account with a nonce.

@acatangiu yes, it's possible, and I am working on such option. Here I wanted to try what we can do with new Transaction Extensions and reuse some of them. With a solution based on custom pallet, I think we can still reuse them. I should have answers with examples soon, if we can reuse them and if it worths it.

georgepisaltu:
> Why do this through transaction extensions? Even the tx ext set we have now is quite complicated and is seen as some "black magic" by many developers on Polkadot 😆.

It's about to get even more complicated as we shift the signature checking logic from the `Checkable` -> `Applyable` logic into its own (new-school) transaction extension (see [POC impl](https://github.com/paritytech/polkadot-sdk/pull/3898/files#diff-ceaa392256ad79b68baabe78711d993a96a076db532bc554dd2cc2712ac8affaR104) in #3898 , which is a follow up to this PR). This will happen as we phase out (old-school) signed transactions in favor of (new-school) general transactions.

> Would this not be cleaner, clearer, less complicated to understand if it were implemented in a pallet?

It would, for sure. As @muharem said, this is more of an experiment of what it would look like with this new interface. We haven't compared the extension vs pallet approaches yet, but I am confident it would be simpler if implemented in a pallet. The origin "swap" that the transaction extension does can happen in a particular pallet's extrinsic which would then call into the actual extrinsic you want to run with the updated origin.

Moreover, this extension we're building here (and any extension we end up using) would obviously be part of the pipeline for all transactions, but would only do useful work when the feature is used (sponsoring a transaction in this case). An extension like `CheckTxVersion` is perfect as a transaction extension because all transactions need to run this check, while something like this sponsored fee extension would run very rarely. Such use cases that don't serve all transactions are IMO not suited to the transaction extension model and should have dedicated pallet logic. This pallet logic can then be abstracted away by clients.

The closest example to this idea would be what `pallet_utility` does in `batch` or `dispatch_as` wrt wrapping calls and altering the origin: these extrinsics offer such low-level functionality that it's bad UX if end users ever need to manually construct a `dispatch_as` IMO, a client should do it for them where appropriate.

If the argument for an extension is that it would be easier to use, I'd counter with the fact that it would probably be even worse if we had an extension for `dispatch_as` that clients need to understand and deal with it every time they construct a transaction. Right now, they don't even worry about its existence because they don't have to, only people concerned are the ones that use it.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5801113

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5801094

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5801114

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5801112

substrate/frame/transaction-payment/src/lib.rs
```diff
@@ -830,6 +774,32 @@ where
...
+/// # Transaction Validity
+///
+/// This extension sets the `priority` field of `TransactionValidity` depending on the amount
+/// of tip being paid per weight unit.
+///
+/// Operational transactions will receive an additional priority bump, so that they are normally
+/// considered before regular transactions.
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct ChargeTransactionPayment<T: Config>(#[codec(compact)] BalanceOf<T>);
```
muharem, line 789: no changes to the API of this type. `get_priority` method decoupled to the separate `Priority` type to be reused. the `withdraw_fee` and `can_withdraw_fee` just dropped.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5845679

muharem:
I have posted an RFC issue with the current solution and a solution based on a pallet, with some considerations. Please check it out and share you thoughts.
https://github.com/paritytech/polkadot-sdk/issues/4123
cc: @kianenigma @xlc @acatangiu @georgepisaltu

Polkadot-Forum:

https://forum.polkadot.network/t/2024-04-23-technical-fellowship-opendev-call/7592/1

muharem:
in favor of https://github.com/paritytech/polkadot-sdk/pull/6428

Issue #3704, opened 2024-03-14
Availability recovery bench hangs test runner


ggwpez:
Trying to run our tests now does not even start anymore with `creating test list failed`:

```sh
$ cargo nextest run --workspace --no-fail-fast

Finished test [unoptimized + debuginfo] target(s) in 4m 59s
error: creating test list failed

Caused by:
  for `polkadot-availability-recovery::availability-recovery-regression-bench`, command `/home/oliver/remote-builds/2722749496389579917/target/debug/deps/availability_recovery_regression_bench-b1fdf52926626227 --list --format terse` exited with code 1
--- stdout:
Warming up...
Took 0 runs to warm up

data_availability_read

Network usage, KiB                     total   per block
Sent to peers                          5.000       1.667
Received from peers               921609.000  307203.000

CPU usage, seconds                     total   per block
Test environment                       1.007       0.336
availability-recovery                 50.853      16.951


--- stderr:
data_availability_read: The resource `availability-recovery` is expected to be equal to 11.5 with a precision 0.05, but the current value is 16.950935916999995
Error: "Regressions found"
```

AndreiEres:
@ggwpez I'm very sorry for that. `cargo test` doesn't run it because these test require `subsystem-benchmarks` feature. Unfortunately, `nextext` ignores required features, so in CI we add this `--filter-expr 'not deps(/polkadot-subsystem-bench/)'`.

ggwpez:
Oh, so it is feature gated, okay.  
I think nextest tries to call `--list` on the bench since it is marked as test, and that fails.  
Maybe you can make it a no-op if the feature is not enabled?

bkchr:
> Unfortunately, `nextext` ignores required features, so in CI we add this `--filter-expr 'not deps(/polkadot-subsystem-bench/)'`.

I don't think that this is the issue. The issue is that you depend on `polkadot-subsystem-bench` in your dev dependencies and that depends on `availability-recovery` with the `subsystem-benchmarks` feature. Which basically means that the feature is always enabled when you run `cargo test`.

AndreiEres:
> The issue is that you depend on `polkadot-subsystem-bench` in your dev dependencies and that depends on `availability-recovery` with the `subsystem-benchmarks` feature. 

Yes, Oliver already found it and I prepared a fix here https://github.com/paritytech/polkadot-sdk/pull/3716

PR #3716, opened 2024-03-15
Avoid running subsystem-benchmarks with nextest

AndreiEres:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3704

### DIFF
--- a/polkadot/node/network/availability-recovery/Cargo.toml
+++ b/polkadot/node/network/availability-recovery/Cargo.toml
@@ -51,3 +51,4 @@ required-features = ["subsystem-benchmarks"]
 
 [features]
 subsystem-benchmarks = []
+subsystem-benchmarks-helpers = []
--- a/polkadot/node/network/availability-recovery/src/lib.rs
+++ b/polkadot/node/network/availability-recovery/src/lib.rs
@@ -857,7 +857,7 @@ async fn erasure_task_thread(
 
 		// In benchmarks this is a very hot loop not yielding at all.
 		// To update CPU metrics for the task we need to yield.
-		#[cfg(feature = "subsystem-benchmarks")]
+		#[cfg(feature = "subsystem-benchmarks-helpers")]
 		tokio::task::yield_now().await;
 	}
 }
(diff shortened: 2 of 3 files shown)


ggwpez (approved):
(no opinion on the code but it fixes my issue)

bkchr:
I don't get how this fix fixes the issue?

I also have the fundamental question, why do we have a benchmark registered as `test`?  Registering the benchmark with `harness = false` should be achieve the same result? 

Generally not sure why this is also not just wrapped into Criterion. I mean you have there logic for warmup etc that would all be done by criterion automatically.

AndreiEres:
@bkchr 
Tests require `subsystem-benchmarks` feature, but they also depend on `polkadot-availability-recovery` with the same feature, so `nextest` runs them. That was the problem.

About why it's done by tests: they are a bit specific, so we need to implement them in that way without criterion.

bkchr:
> About why it's done by tests: they are a bit specific, so we need to implement them in that way without criterion.

Okay fine, but you don't need to use criterion. If you use `harness = false` you basically have to write your main on your own, like you have done.

AndreiEres:
> If you use `harness = false` you basically have to write your main on your own, like you have done.

Yes, but what do you mean?

polkadot/node/network/availability-recovery/Cargo.toml
```diff
@@ -51,3 +51,4 @@ required-features = ["subsystem-benchmarks"]
 
 [features]
 subsystem-benchmarks = []
+subsystem-benchmarks-helpers = []
```
sandreim, line 54: `subsystem-benchmarks` feature seems to no longer be used at all.

sandreim (commented):
It's kind of odd to have 2 features in use now. I also don't really get why this fixes the issue.

polkadot/node/network/availability-recovery/Cargo.toml
AndreiEres: It's in use. We require that feature to run regression tests. That way we don't mix normal and regression tests.

sandreim:
Can't the test just be skipped if feature not defined ?

bkchr:
> Yes, but what do you mean?

I mean this: https://github.com/paritytech/polkadot-sdk/pull/3741

bkchr:
Closing in favor of: https://github.com/paritytech/polkadot-sdk/pull/3741

Issue #2240, opened 2023-11-09
More specific errors for `decrease_balance`


liamaharon:
should we have an issue for more specific errors for `decrease_balance`?

_Originally posted by @muharem in https://github.com/paritytech/polkadot-sdk/issues/1296#issuecomment-1801903527_
            

Currently `decrease_balance` errors don't differentiate between failing due to ED or frozen constraints. Could improve the API by emitting a different error, depending on why it failed.

chungquantin:
@liamaharon I have created a PR to work on this issue. Please give a check and let me know if there is another case need to be covered.

PR #3720, opened 2024-03-17
Handle existential deposit `BelowMinimum` error

chungquantin:
Issue: https://github.com/paritytech/polkadot-sdk/issues/2240

### DIFF
--- a/substrate/frame/support/src/traits/tokens/fungible/regular.rs
+++ b/substrate/frame/support/src/traits/tokens/fungible/regular.rs
@@ -187,7 +187,16 @@ pub trait Unbalanced<AccountId>: Inspect<AccountId> {
 			Exact => ensure!(reducible >= amount, TokenError::FundsUnavailable),
 		}
 
-		let new_balance = old_balance.checked_sub(&amount).ok_or(TokenError::FundsUnavailable)?;
+		let new_balance = match old_balance.checked_sub(&amount) {
+			Some(balance) => balance,
+			None => {
+				if matches!(preservation, Preservation::Protect | Preservation::Preserve) {
+					return Err(TokenError::BelowMinimum.into());
+				}
+				return Err(TokenError::FundsUnavailable.into());
+			},
+		};
+
 		if let Some(dust) = Self::write_balance(who, new_balance)? {
 			Self::handle_dust(Dust(dust));
 		}


substrate/frame/support/src/traits/tokens/fungible/regular.rs
```diff
@@ -187,7 +187,16 @@ pub trait Unbalanced<AccountId>: Inspect<AccountId> {
 			Exact => ensure!(reducible >= amount, TokenError::FundsUnavailable),
 		}
 
-		let new_balance = old_balance.checked_sub(&amount).ok_or(TokenError::FundsUnavailable)?;
+		let new_balance = match old_balance.checked_sub(&amount) {
+			Some(balance) => balance,
+			None => {
+				if matches!(preservation, Preservation::Protect | Preservation::Preserve) {
```
chungquantin, line 193: Explanation: If the `preservation` is `Protect` or `Preserve` which means the account must be alive after the method call. Hence, if it is failed to `checked_sub` the `old_balance` with the provided `amount` => `BelowMinimum` error thrown

substrate/frame/support/src/traits/tokens/fungible/conformance_tests/regular/unbalanced.rs
```diff
@@ -146,7 +146,7 @@ where
 			Preservation::Preserve,
 			Fortitude::Polite,
 		),
-		Err(TokenError::FundsUnavailable.into()),
+		Err(TokenError::BelowMinimum.into()),
```
liamaharon, line 149: Can we also add a conformance test for FundsUnavaliable?

substrate/frame/support/src/traits/tokens/fungible/regular.rs
```diff
@@ -184,7 +184,13 @@ pub trait Unbalanced<AccountId>: Inspect<AccountId> {
 		let reducible = Self::reducible_balance(who, preservation, force);
 		match precision {
 			BestEffort => amount = amount.min(reducible),
-			Exact => ensure!(reducible >= amount, TokenError::FundsUnavailable),
+			Exact =>
+				if reducible < amount {
+					if matches!(preservation, Preservation::Protect | Preservation::Preserve) {
+						return Err(TokenError::BelowMinimum.into());
+					}
```
liamaharon, line 191: I don't understand why to always return BelowMinimum here.

liamaharon (commented):
Could you please write in the pr comment each way the method could fail, and the corresponding err you're returning in each case? 

Would be nice to also have a conformance test covering each of these cases with a comment explaining the rationale.

substrate/frame/support/src/traits/tokens/fungible/conformance_tests/regular/unbalanced.rs
chungquantin: Yes I will add one

substrate/frame/balances/src/tests/fungible_tests.rs
chungquantin, line 1:
```rust
	/// Hold some funds in an account. If a hold for `reason` is already in place, then this
	/// will increase it.
	fn hold(reason: &Self::Reason, who: &AccountId, amount: Self::Balance) -> DispatchResult {
		// NOTE: This doesn't change the total balance of the account so there's no need to
		// check liquidity.

		Self::ensure_can_hold(reason, who, amount)?;
		// Should be infallible now, but we proceed softly anyway.
>>>>		Self::decrease_balance(who, amount, Exact, Protect, Force)?;
		Self::increase_balance_on_hold(reason, who, amount, BestEffort)?;
		Self::done_hold(reason, who, amount);
		Ok(())
	}

```

Because method `hold()` has the method `decrease_balance` called with `Preservation::Protect`, if it fails, it must throw `TokenError::BelowMinimum` instead

PR #3727, opened 2024-03-18
Snowbridge Beacon header age check

claravanstaden:
## Bug Explanation
Adds a check that prevents finalized headers with a gap larger than the sync committee period being imported, which could cause execution headers in the gap being unprovable. The current version of the Ethereum client checks that there is a header at least every sync committee, but it doesn't check that the headers are within a sync period of each other. For example:

Header 100 (sync committee period 1)
Header 9000 (sync committee period 2)
(8900 blocks apart)

These headers are in adjacent sync committees, but more than the sync committee period (8192 blocks) apart.

The reason we need a header every 8192 slots at least, is the header is used to prove messages within the last 8192 blocks. If we import header 9000, and we receive a message to be verified at header 200, the `block_roots` field of header 9000 won't contain the header in order to do the ancestry check.

## Environment
While running in Rococo, this edge case was discovered after the relayer was offline for a few days. It is unlikely, but not impossible, to happen again and so it should  be backported to polkadot-sdk 1.7.0 (so that [polkadot-fellows/runtimes](https://github.com/polkadot-fellows/runtimes) can be updated with the fix). 

Our Ethereum client has been operational on Rococo for the past few months, and this been the only major issue discovered so far.

### Unrelated Change
An unrelated nit: Removes a left over file that should have been deleted when the `parachain` directory was removed.


bridges/snowbridge/pallets/ethereum-client/src/lib.rs
```diff
@@ -15,8 +15,6 @@
 //! ## Consensus Updates
 //!
 //! * [`Call::submit`]: Submit a finalized beacon header with an optional sync committee update
-//! * [`Call::submit_execution_header`]: Submit an execution header together with an ancestry proof
```
vgeddes, line 18: This does not seem relevant?

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs
```diff
@@ -286,8 +286,8 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 		match call {
 			RuntimeCall::System(frame_system::Call::set_storage { items })
 				if items.iter().all(|(k, _)| {
-					k.eq(&bridging::XcmBridgeHubRouterByteFee::key()) |
```
vgeddes, line 289: This is certainly a bug, its still not relevant to the issue in the beacon light client. I suggest updating the PR description,  saying you added some miscellaneous unrelated fixes.

bridges/snowbridge/pallets/ethereum-client/src/lib.rs
claravanstaden: Correct, reverted in https://github.com/paritytech/polkadot-sdk/pull/3727/commits/27ecafd7698295cf6b8e0f67a556bc1f9a23afbb.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs
claravanstaden: I reverted this change to keep the PR as focused as possible. I added the change in this PR: https://github.com/paritytech/polkadot-sdk/pull/3761

acatangiu (approved):

bkontur (approved):

PR #3741, opened 2024-03-18
Make `availability-recovery-regression-bench` a benchmark

bkchr:
Closes: https://github.com/paritytech/polkadot-sdk/issues/3704


ggwpez (approved):
Yea this is more proper.

rcny (approved):

sandreim (approved):
A clean cut indeed. I assume this still fails the pipeline if benchmarks fails assuming we `allow_failure: false`, right ?

AndreiEres (approved):
Thank you

bkchr:
> . I assume this still fails the pipeline if benchmarks fails assuming we `allow_failure: false`, right ?

I tested it, `cargo bench` propagates the error. However, just setting `allow_failure` isn't enough as github is not respecting this.

PR #3757, opened 2024-03-20
Make public addresses go first in authority discovery DHT records

dmitry-markin:
Make sure explicitly set by the operator public addresses go first in the authority discovery DHT records.

Also update `Discovery` behavior to eliminate duplicates in the returned addresses.

This PR should improve situation with https://github.com/paritytech/polkadot-sdk/issues/3519.

Obsoletes https://github.com/paritytech/polkadot-sdk/pull/3657.

### DIFF
--- a/substrate/client/authority-discovery/src/worker.rs
+++ b/substrate/client/authority-discovery/src/worker.rs
@@ -35,6 +35,7 @@ use addr_cache::AddrCache;
 use codec::{Decode, Encode};
 use ip_network::IpNetwork;
 use libp2p::{core::multiaddr, identity::PublicKey, multihash::Multihash, Multiaddr, PeerId};
+use linked_hash_set::LinkedHashSet;
 use multihash_codetable::{Code, MultihashDigest};
 
 use log::{debug, error, log_enabled};
@@ -120,14 +121,22 @@ pub struct Worker<Client, Network, Block, DhtEventStream> {
 
 	/// Interval to be proactive, publishing own addresses.
 	publish_interval: ExpIncInterval,
+
 	/// Pro-actively publish our own addresses at this interval, if the keys in the keystore
 	/// have changed.
 	publish_if_changed_interval: ExpIncInterval,
+
 	/// List of keys onto which addresses have been published at the latest publication.
 	/// Used to check whether they have changed.
 	latest_published_keys: HashSet<AuthorityId>,
+
 	/// Same value as in the configuration.
 	publish_non_global_ips: bool,
+
+	/// Public addresses set by the node operator to always publish first in the authority
+	/// discovery DHT record.
+	public_addresses: LinkedHashSet<Multiaddr>,
+
 	/// Same value as in the configuration.
 	strict_record_validation: bool,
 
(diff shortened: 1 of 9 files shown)


substrate/client/authority-discovery/src/worker.rs
```diff
@@ -304,32 +338,48 @@ where
...
-				multiaddr::Protocol::Ip6(ip) if !IpNetwork::from(ip).is_global() => false,
-				_ => true,
+				if self.public_addresses.contains(&address) {
+					// Already added above.
+					None
+				} else {
+					Some(address)
+				}
+			}))
+			.filter(move |address| {
```
dmitry-markin, line 359: Not sure, may be we should always publish public addresses even if they are non-global.

lexnv (approved):
LGTM! Nice one 👍

substrate/client/authority-discovery/src/worker.rs
dmitry-markin: I was thinking about a testnet in a private network case. But we can also use `public_non_global_ips` flag for this.

bkchr: Yeah private network. When you run it locally on your computer, you also need authority discovery to work ;)

bkchr (approved):
Ty!

One thing I'm not sure if we are already doing it, could we ensure that `external_addresses` contains the listen addresses? Maybe we could even have an extra function for getting the listen addresses to have in the end, [public addresses, listen addresses, external addresses].

dmitry-markin:
> One thing I'm not sure if we are already doing it, could we ensure that `external_addresses` contains the listen addresses? 

`external_addresses` do not contain listen addresses, and I thought it's a feature and not a bug.

> Maybe we could even have an extra function for getting the listen addresses to have in the end, [public addresses, listen addresses, external addresses].

Do you mean [public addresses, external addresses, listen addresses]?

bkchr:
> `external_addresses` do not contain listen addresses, and I thought it's a feature and not a bug.

IDK :D I mean this was more some proposal by me. Also should not block this pr BTW.



> Do you mean [public addresses, external addresses, listen addresses]?

Listen addresses are more "trusted" than external addresses as you can just query your local network devices to get their ip addresses etc. For validators which should be exposed to the internet, the listen addresses should also give you the correct ip addresses directly. If the operator is giving public addresses, I assume that listen and public will probably be the same.

dmitry-markin:
> > `external_addresses` do not contain listen addresses, and I thought it's a feature and not a bug.
> 
> IDK :D I mean this was more some proposal by me. Also should not block this pr BTW.
> 
> > Do you mean [public addresses, external addresses, listen addresses]?
> 
> Listen addresses are more "trusted" than external addresses as you can just query your local network devices to get their ip addresses etc. For validators which should be exposed to the internet, the listen addresses should also give you the correct ip addresses directly. If the operator is giving public addresses, I assume that listen and public will probably be the same.

We can address it in a follow-up PR. Personally, I would refrain from adding listen addresses as they contain all the interfaces (including 127.0.0.1 and private networks), and it should be enough to have just one valid Identify response from a remote peer to learn a valid external address. We should take into account address translation done by libp2p (it assign a listen address port to discovered external addresses) to make an informed decision though.

PR #3763, opened 2024-03-20
XCM remove extra QueryId types from traits

pgherveou:
We do not need to make these traits generic over QueryId type, we can
just use the QueryId alias everywhere


pgherveou:
@franciscoaguirre these extra `type QueryId` on these traits seems unnecessary to me, we should just use the QueryId type alias, can you take a quick look at this refactoring?

franciscoaguirre (approved):
Good point. It made sense to have them be generic, but XCM already exposes the QueryId type and it's always assumed to be an integer. Probably not an issue to fix it as an integer 👍

acatangiu (approved):

PR #3764, opened 2024-03-20
authority-discovery: Set intervals to start when authority keys changes

lexnv:
The authority-discovery mechanism has implemented a few exponential timers for:
- publishing the authority records
  - goes from 2 seconds (when freshly booted) to 1 hour if the node is long-running
  - set to 1 hour after successfully publishing the authority record
- discovering other authority records
  - goes from 2 seconds (when freshly booted) to 10 minutes if the node is long-running

This PR resets the exponential publishing and discovery interval to defaults ensuring that long-running nodes:
- will retry publishing the authority records as aggressively as freshly booted nodes
  - Currently, if a long-running node fails to publish the DHT record when the keys change (ie DhtEvent::ValuePutFailed), it will only retry after 1 hour
- will rediscover other authorities faster (since there is a chance that other authority keys changed)

The subp2p-explorer has difficulties discovering the authorities when the authority set changes in the first few hours. This might be entirely due to the recursive nature of the DHT and the needed time to propagate the records. However, there is a small chance that the authority publishing failed and is only retried in 1h.

Let me know if this makes sense 🙏 

cc @paritytech/networking


substrate/client/authority-discovery/src/interval.rs
```diff
@@ -37,14 +38,20 @@ impl ExpIncInterval {
...
+		Self { start, max, next: start * 2, delay }
 	}
 
 	/// Fast forward the exponentially increasing interval to the configured maximum.
 	pub fn set_to_max(&mut self) {
 		self.next = self.max;
 		self.delay = Delay::new(self.next);
 	}
+
+	/// Fast forward the exponentially increasing interval to the configured start.
```
dmitry-markin, line 50: nit: Fast forward -> rewind?
```suggestion
	/// Rewind the exponentially increasing interval to the configured start.
```

dmitry-markin:
> publishing the authority records
> 
>     * goes from 2 seconds (when freshly booted) to 1 hour if the node is long-running

I wonder if this is working as intended. On rococo, the authority-dicovery DHT record was published only once during last two days: [logs](https://grafana.teleport.parity.io/explore?schemaVersion=1&panes=%7B%22mhu%22:%7B%22datasource%22:%22P3572579027B246FE%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bchain%3D%5C%22rococo%5C%22,%20node%3D%5C%22rococo-validator-a-node-1%5C%22%7D%20%7C~%20%60Authority%20DHT%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P3572579027B246FE%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221710720000000%22,%22to%22:%221710979199000%22%7D%7D%7D&orgId=1).

dmitry-markin (approved):

dmitry-markin:
The PR makes sense to me, but I don't understand why on rococo 1h-interval is not working properly.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -347,8 +347,17 @@ where
...
+		if only_if_changed {
+			// If the authority keys did not change and the `publish_if_changed_interval` was
+			// triggered then do nothing.
+			if keys == self.latest_published_keys {
+				return Ok(())
+			}
+
+			// We have detected a change in the authority keys, reset the timers to
+			// publish and gather data faster.
+			self.publish_interval.set_to_start();
```
alexggh, line 359: This would only detect that our keys have change, like when we rotate them and since we get here we are going to publish them, so what is the point of publishing them 2 seconds later ? 

Same for `query_interval` that is for the current node discovering other nodes, why we would try to discover more nodes just because our keys changed, the other nodes should actually try more often, not us.

alexggh: But I guess it can't hurt :)

alexggh (approved):

alexggh:
> > > publishing the authority records
> > > ```
> > > * goes from 2 seconds (when freshly booted) to 1 hour if the node is long-running
> > > ```
> > 
> > 
> > I wonder if this is working as intended. On rococo, the authority-dicovery DHT record was published only once during last two days: [logs](https://grafana.teleport.parity.io/explore?schemaVersion=1&panes=%7B%22mhu%22:%7B%22datasource%22:%22P3572579027B246FE%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bchain%3D%5C%22rococo%5C%22,%20node%3D%5C%22rococo-validator-a-node-1%5C%22%7D%20%7C~%20%60Authority%20DHT%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P3572579027B246FE%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221710720000000%22,%22to%22:%221710979199000%22%7D%7D%7D&orgId=1).
> 
> I did see it working properly in versi e.g [versi](https://grafana.teleport.parity.io/goto/E9ZGI1JSg?orgId=1), but I do agree the rococo logs don't make any sense and are worrisome.

Actually, it is explained by this:
```
			DhtEvent::ValuePut(hash) => {
				// Fast forward the exponentially increasing interval to the configured maximum. In
				// case this was the first successful address publishing there is no need for a
				// timely retry.
				self.publish_interval.set_to_max();
```

set_to_max resets the timer and we seem to call that more often the 1h window, https://grafana.teleport.parity.io/goto/NakJI1JSR?orgId=1, probably we want to fix that as well @lexnv

lexnv:
Thanks a lot guys for having a look at this 🙏 

Indeed, there seems to be an issue with this at least on Rococo.

We are setting the `publish_interval` to "1 hour from now" every time we receive a `DhtEvent::ValuePut`.
However, we are only interested in events that are correlated with our put records, and not all the network events.

I believe the DHT event stream is simply giving us whatever the network state is, not limited to our authority records.

Here we construct a new event stream for authority discovery:
https://github.com/paritytech/polkadot-sdk/blob/41a5776c765ebf7765cf70c6db2fca318a69b75a/substrate/bin/node/cli/src/service.rs#L602-L608

And here the stream is implemented, with the name used only for metrics:
https://github.com/paritytech/polkadot-sdk/blob/41a5776c765ebf7765cf70c6db2fca318a69b75a/substrate/client/network/src/service.rs#L1006-L1016

I'll have a look to associate only events for our published authority records and reset the timer only once.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -126,6 +126,9 @@ pub struct Worker<Client, Network, Block, DhtEventStream> {
 	/// List of keys onto which addresses have been published at the latest publication.
 	/// Used to check whether they have changed.
 	latest_published_keys: HashSet<AuthorityId>,
+
```
bkchr, line 129: Some docs would be nice.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -485,6 +505,11 @@ where
 				debug!(target: LOG_TARGET, "Successfully put hash '{:?}' on Dht.", hash)
 			},
 			DhtEvent::ValuePutFailed(hash) => {
+				if !self.latest_published_kad_keys.remove(&hash) {
+					// Not a value we have published or received multiple times.
+					return;
+				}
+
```
bkchr, line 512: Don't you want to reset the timer here? I mean given what you write in the pr description, I got the impression.

bkchr (approved):

substrate/client/authority-discovery/src/worker.rs
lexnv: Yep, that makes sense!

I reset the `publish_interval` here. I'm a bit concerned about being too aggressive when we constantly fail to put the DHT records. We'd reset and resubmit the DHT records at 2 seconds intervals until all records succeed.

This would mean that for freshly-started node (where failure might be high) we'd resubmit at 2 seconds internvals instead of exponentially increasing ones 🤔

dmitry-markin: Just speculating: may be introduce another "ephemeral" (i.e., `Option`) exponential-backoff timer just to handle failed publishings?

lexnv:
After collecting more data in our test nets [here](https://grafana.teleport.parity.io/explore?schemaVersion=1&panes=%7B%22l5y%22:%7B%22datasource%22:%22P28A19C7766384B81%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bchain%3D%5C%22versi%5C%22,%20level%3D%5C%22DEBUG%5C%22,%20target%3D%5C%22sub-authority-discovery%5C%22,%20node%3D~%5C%22versi-validator-auth-.%2A%5C%22%7D%20%7C~%20%60Authority%7CSuc%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P28A19C7766384B81%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%221711453797447%22,%22to%22:%221711475397447%22%7D%7D%7D&orgId=1):
- The publish interval timer is triggered exponentially from 2 sec to 1h until the ValuePut is received
- When the ValuePut is received the timer is fast-forwarded to 1h (if not already fast-forwarded): this ensures that we'll publish again records at 1h interval
  - Prior to this PR we'd only publish at the boot or when the keys changed, since the publish interval was constantly shifted forward by `ValuePut` events

Issue #3762, opened 2024-03-20
Westend not being able to process coretime messages


girazoki:
I realized coretime-messages are not being processed by the Westend relay chain: See https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestend-rpc.polkadot.io#/explorer/query/0xdbe4b90e36089012396b8aa78f192fbb96cce04381828b8e059d2050918123dd as an example.

I think it is due to the barrier, as westend is not allowing unpaid execution from Coretime: https://github.com/paritytech/polkadot-sdk/blob/3c3d6fceb82372a3019b37117aa453d564b212de/polkadot/runtime/westend/src/xcm_config.rs#L176, as opposed to Rococo.

And the messages sent by Coretime always start with `"UnpaidExecution"`: https://github.com/paritytech/polkadot-sdk/blob/7241a8db7b3496816503c6058dae67f66c666b00/cumulus/parachains/runtimes/coretime/coretime-westend/src/coretime.rs#L196

I am not sure if the issue was reported already, but I could not find anything related

PR #3768, opened 2024-03-20
Fix xcm config for coretime.

eskimor:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3762 .

### DIFF
--- a/polkadot/runtime/westend/src/xcm_config.rs
+++ b/polkadot/runtime/westend/src/xcm_config.rs
@@ -143,15 +143,13 @@ impl Contains<Location> for OnlyParachains {
 	}
 }
 
-pub struct SystemOrFellows;
-impl Contains<Location> for SystemOrFellows {
+pub struct Fellows;
+impl Contains<Location> for Fellows {
 	fn contains(location: &Location) -> bool {
-		let is_system = IsChildSystemParachain::<ParaId>::contains(location);
-		let is_fellows = matches!(
+		matches!(
 			location.unpack(),
 			(0, [Parachain(COLLECTIVES_ID), Plurality { id: BodyId::Technical, .. }])
-		);
-		is_system || is_fellows
+		)
 	}
 }
 
@@ -175,7 +173,7 @@ pub type Barrier = TrailingSetTopicAsId<(
 			// Subscriptions for version tracking are OK.
 			AllowSubscriptionsFrom<OnlyParachains>,
 			// Messages from system parachains or the Fellows plurality need not pay for execution.
-			AllowExplicitUnpaidExecutionFrom<SystemOrFellows>,
+			AllowExplicitUnpaidExecutionFrom<(IsChildSystemParachain<ParaId>, Fellows)>,
 		),
 		UniversalLocation,
 		ConstU32<8>,
--- a/polkadot/runtime/westend/src/xcm_config.rs
+++ b/polkadot/runtime/westend/src/xcm_config.rs
@@ -143,14 +143,15 @@ impl Contains<Location> for OnlyParachains {
 	}
 }
 
-pub struct CollectivesOrFellows;
-impl Contains<Location> for CollectivesOrFellows {
+pub struct SystemOrFellows<ParaId>(PhantomData<ParaId>);
+impl<ParaId: IsSystem + From<u32>> Contains<Location> for SystemOrFellows<ParaId> {
 	fn contains(location: &Location) -> bool {
-		matches!(
+		let is_system = IsChildSystemParachain::contains(location);
+		let is_fellows = matches!(
 			location.unpack(),
-			(0, [Parachain(COLLECTIVES_ID)]) |
-				(0, [Parachain(COLLECTIVES_ID), Plurality { id: BodyId::Technical, .. }])
-		)
+			(0, [Parachain(COLLECTIVES_ID), Plurality { id: BodyId::Technical, .. }])
+		);
+		is_system || is_fellows
 	}
 }
 
@@ -174,7 +175,7 @@ pub type Barrier = TrailingSetTopicAsId<(
 			// Subscriptions for version tracking are OK.
 			AllowSubscriptionsFrom<OnlyParachains>,
 			// Messages coming from system parachains need not pay for execution.
-			AllowExplicitUnpaidExecutionFrom<IsChildSystemParachain<ParaId>>,
+			AllowExplicitUnpaidExecutionFrom<SystemOrFellows<ParaId>>,
 		),
 		UniversalLocation,
 		ConstU32<8>,
--- a/polkadot/runtime/westend/src/xcm_config.rs
+++ b/polkadot/runtime/westend/src/xcm_config.rs
@@ -28,13 +28,11 @@ use frame_support::{
 };
 use frame_system::EnsureRoot;
 use pallet_xcm::XcmPassthrough;
-use polkadot_parachain_primitives::primitives::IsSystem;
 use runtime_common::{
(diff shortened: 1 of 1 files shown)


polkadot/runtime/westend/src/xcm_config.rs
```diff
@@ -172,8 +173,8 @@ pub type Barrier = TrailingSetTopicAsId<(
 			AllowTopLevelPaidExecutionFrom<Everything>,
 			// Subscriptions for version tracking are OK.
 			AllowSubscriptionsFrom<OnlyParachains>,
-			// Collectives and Fellows plurality get free execution.
-			AllowExplicitUnpaidExecutionFrom<CollectivesOrFellows>,
+			// Messages coming from system parachains need not pay for execution.
+			AllowExplicitUnpaidExecutionFrom<IsChildSystemParachain<ParaId>>,
```
acatangiu, line 177: does this change not take away unpaid execution from `Fellows`?

acatangiu: looks like so, [IsChildSystemParachain matches only on exact parachain origin and not its children](https://github.com/paritytech/polkadot-sdk/blob/8428f678fe5179399f6854e194f1c8b30a9102f9/polkadot/xcm/xcm-builder/src/barriers.rs#L317C3-L321C4);

whereas [Fellows is a child location of Collectives system chain](https://github.com/paritytech/polkadot-sdk/blob/d55d4f64b50c858bce6fa1a429f50d486cc67d49/polkadot/runtime/westend/src/xcm_config.rs#L151).

acatangiu (changes requested):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5588972

polkadot/runtime/westend/src/xcm_config.rs
```diff
@@ -172,8 +176,8 @@ pub type Barrier = TrailingSetTopicAsId<(
 			AllowTopLevelPaidExecutionFrom<Everything>,
 			// Subscriptions for version tracking are OK.
 			AllowSubscriptionsFrom<OnlyParachains>,
-			// Collectives and Fellows plurality get free execution.
-			AllowExplicitUnpaidExecutionFrom<CollectivesOrFellows>,
+			// Messages coming from system parachains need not pay for execution.
```
acatangiu, line 179: nit:

```suggestion
			// Messages from system parachains or the Fellows plurality need not pay for execution.
```

acatangiu (approved):

polkadot/runtime/westend/src/xcm_config.rs
```diff
@@ -142,14 +145,15 @@ impl Contains<Location> for OnlyParachains {
 	}
 }
 
-pub struct CollectivesOrFellows;
-impl Contains<Location> for CollectivesOrFellows {
+pub struct SystemOrFellows<ParaId>(PhantomData<ParaId>);
+impl<ParaId: IsSystem + From<u32>> Contains<Location> for SystemOrFellows<ParaId> {
```
bkchr, line 149: Why is this generic over `ParaId`?

polkadot/runtime/westend/src/xcm_config.rs
```diff
@@ -142,14 +145,15 @@ impl Contains<Location> for OnlyParachains {
...
 	fn contains(location: &Location) -> bool {
-		matches!(
+		let is_system = IsChildSystemParachain::<ParaId>::contains(location);
```
bkchr, line 151: Doesn't this allow any XCM from a system chain to use `UnpaidExecution`?

eskimor: From my understanding no, otherwise the code explicitly allowing the child "fellows" to also not pay would not make much sense (and the could would have been broken before already).

eskimor: https://github.com/paritytech/polkadot-sdk/blob/8428f678fe5179399f6854e194f1c8b30a9102f9/polkadot/xcm/xcm-builder/src/barriers.rs#L320

bkchr: The linked barrier is about "is this a system chain", but here in this context it is about allowing unpaid execution.

acatangiu: Yes it does, but only when the origin is a system chain “root”, and not any child sublocations. IMO it is ok to allow it

bkchr: > origin is a system chain “root”

Where is this checked?

franciscoaguirre: This means if the origin is any of the system chains or [collectives, fellows] then it can do unpaid execution.
It doesn't allow any account in a system chain to use unpaid execution, if that's what you're worried about.

franciscoaguirre: Only "root" can send XCMs with the location of the chain itself, users would appear as sublocations

polkadot/runtime/westend/src/xcm_config.rs
```diff
@@ -172,8 +176,8 @@ pub type Barrier = TrailingSetTopicAsId<(
 			AllowTopLevelPaidExecutionFrom<Everything>,
 			// Subscriptions for version tracking are OK.
 			AllowSubscriptionsFrom<OnlyParachains>,
-			// Collectives and Fellows plurality get free execution.
-			AllowExplicitUnpaidExecutionFrom<CollectivesOrFellows>,
+			// Messages from system parachains or the Fellows plurality need not pay for execution.
+			AllowExplicitUnpaidExecutionFrom<SystemOrFellows<ParaId>>,
```
bkontur, line 180: very nit, it can be simplified also e.g.:
```suggestion
			AllowExplicitUnpaidExecutionFrom<(
						IsChildSystemParachain<ParaId>>,
						Fellows,   // <- SystemOrFellows -> Fellows can just to match collectives stuff
			)>,						
```

polkadot/runtime/westend/src/xcm_config.rs
bkchr: I think now I found out what confused me.

`(0, [Parachain(SYSTEM_ID), AccountId32])`

This would be for example a location if send by a user or?

franciscoaguirre: Yes

bkchr: Okay thank you ;) This explains it. I thought we would just allow anything as long as it comes from the system chain. But we actually match explicit, so it doesn't make sense what I was worried about.

acatangiu: It was a good question and thought experiment, better safe than sorry 👍🏻

bkontur: > I think now I found out what confused me.
> 
> `(0, [Parachain(SYSTEM_ID), AccountId32])`
> 
> This would be for example a location if send by a user or?

yes, you just need to add `DescendOrigin(AccountId32`) to the XCM program

polkadot/runtime/westend/src/xcm_config.rs
bkontur: if we do it this way: https://github.com/paritytech/polkadot-sdk/pull/3768/files#r1533610770
we dont need any generics here

franciscoaguirre (approved):

Issue #3081, opened 2024-01-26
[RPC-Spec-V2] Limit transaction_unstable_broadcast operations


lexnv:
> The JSON-RPC server must allow at least 4 transactions being broadcasted at the same time per JSON-RPC client. Any attempt to broadcast more than 4 transactions simultaneously might result in null being returned.

- ensure that at least 4 operations per RPC client are accepted by the server (extracted from [spec](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/transaction_unstable_broadcast.md))

cc @paritytech/subxt-team

PR #3772, opened 2024-03-20
rpc-v2: Limit transactionBroadcast calls to 16

lexnv:
This PR limits the number of active calls to the transactionBroadcast APIs to 16.

cc @paritytech/subxt-team 

Closes: https://github.com/paritytech/polkadot-sdk/issues/3081


substrate/client/rpc-spec-v2/src/transaction/transaction.rs
```diff
@@ -81,8 +84,20 @@ where
...
 		let fut = async move {
+			let Some(_reserved_connection) = rpc_connections.reserve_space(pending.connection_id())
+			else {
+				let err = ErrorObject::owned(
+					BAD_FORMAT,
+					format!("Reached maximum number of connections"),
+					None::<()>,
+				);
+				let _ = pending.reject(err).await;
+				return
```
lexnv, line 98: TODO: We need a proper error in the spec for this API

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5590652

substrate/client/rpc-spec-v2/src/transaction/transaction.rs
lexnv: Have created: https://github.com/paritytech/json-rpc-interface-spec/pull/146

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5842139

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
```diff
@@ -26,6 +26,8 @@ use std::sync::Arc;
 use substrate_test_runtime_client::AccountKeyring::*;
 use substrate_test_runtime_transaction_pool::uxt;
 
+const MAX_TX_PER_CONNECTION: usize = 4;
```
jsdw, line 29: As a nit, the spec only says "at least 4 transactions can be broadcast at the same time". Do we have any reason to also use 4 as the upper bound here?

(One reason I can think of is just that it nudges clients to comply with that bit of the spec, ie if they try to send more than 4, which _might_ not be supported per the spec, then an error will come back)

substrate/client/service/src/builder.rs
```diff
@@ -655,10 +655,13 @@ where
 		(chain, state, child_state)
 	};
 
+	const MAX_TRANSACTION_PER_CONNECTION: usize = 4;
```
jsdw, line 658: Nit: the spec only says "at least 4 transactions can be broadcast at the same time". Do we have any reason to also use 4 as the upper bound here?

(One reason I can think of is just that it nudges clients to comply with that bit of the spec, ie if they try to send more than 4, which _might_ not be supported per the spec, then an error will come back)

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -172,6 +209,8 @@ where
 		// The future expected by the executor must be `Future<Output = ()>` instead of
 		// `Future<Output = Result<(), Aborted>>`.
 		let fut = fut.map(move |_| {
+			// Connection space is cleaned when this object is dropped.
+			let _reserved_identifier = reserved_identifier;
```
jsdw, line 213: Nit: I'd just call `drop(reserved_identifier)` here to be clearer.

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -187,7 +226,16 @@ where
 		Ok(Some(id))
 	}
 
-	fn stop_broadcast(&self, operation_id: String) -> Result<(), ErrorBroadcast> {
+	async fn stop_broadcast(
+		&self,
+		connection_details: ConnectionDetails,
+		operation_id: String,
+	) -> Result<(), ErrorBroadcast> {
+		// The operation ID must correlate to the same connectio ID.
```
jsdw, line 234:
```suggestion
		// The operation ID must correlate to the same connection ID.
```

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -100,13 +119,31 @@ where
...
+		let Some(reserved_identifier) = reserved_connection.register(id.clone()) else {
+			// This can only happen if the generated operation ID is not unique.
+			return Ok(None)
+		};
+
+		// The compiler can no longer deduce the type of the stream and complains
+		// about `one type is more general than the other`.
+		let mut best_block_import_stream: std::pin::Pin<
+			Box<dyn Stream<Item = <Pool::Block as BlockT>::Hash> + Send>,
+		> =
```
jsdw, line 146: I wonder why this type annotation needed adding!

jsdw (approved):
A couple of small nits but all makes sense to me!

I don't mind the limit being 4 just to exercise clients and nudge them to be able to handle a max of 4, but for most other cases I wonder what an actually sensible limit would be here

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
lexnv: This is an edge-case in the rust compiler, where the rustc cannot deduce the type of this stream.

I'm not entirely sure what is happening behind the scenes, I expect the async block cannot be properly deduced here.

The error I get is:
```rust
error[E0308]: mismatched types
   --> substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs:222:3
    |
146 |                 |notification| async move { notification.is_new_best.then_some(notification.hash) },
    |                                --------------------------------------------------------------------
    |                                |
    |                                the expected `async` block
    |                                the found `async` block
...
222 |         sc_rpc::utils::spawn_subscription_task(&self.executor, fut);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected `async` block `{async block@substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs:146:20: 146:88}`
               found `async` block `{async block@substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs:146:20: 146:88}`
note: the lifetime requirement is introduced here
   --> /home/lexnv/workspace/polkadot-sdk/substrate/client/rpc/src/utils.rs:178:34
    |
178 |     fut: impl Future<Output = ()> + Send + 'static,
    |                                     ^^^^

For more information about this error, try `rustc --explain E0308`.
```

substrate/client/service/src/builder.rs
lexnv: Yep, have changed it to 16

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
lexnv: This makes sense, thanks!

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
niklasad1: Try bumping jsonrpsee to v0.22.4 it has fixed it for me locally

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
```diff
@@ -521,3 +525,53 @@ async fn tx_broadcast_resubmits_dropped_tx() {
...
+	assert_eq!(
+		event,
+		MiddlewarePoolEvent::TransactionStatus {
+			transaction: xt.clone(),
+			status: TxStatusTypeTest::Ready
+		}
+	);
+	assert_eq!(1, exec_middleware.num_tasks());
+
+	let limit_reached: Option<String> =
```
niklasad1, line 557: I found this code hard to read and I would prefer something like:

```rust
	let operation_id: Option<String> =  tx_api.call("transaction_unstable_broadcast", rpc_params![&xt]).await.unwrap();
	assert!(operation_id.is_none(), "No operation ID => tx was rejected");
```

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -100,13 +119,31 @@ where
 	<Pool::Block as BlockT>::Hash: Unpin,
 	Client: HeaderBackend<Pool::Block> + BlockchainEvents<Pool::Block> + Send + Sync + 'static,
 {
-	fn broadcast(&self, bytes: Bytes) -> RpcResult<Option<String>> {
+	async fn broadcast(
+		&self,
+		connection_details: ConnectionDetails,
+		bytes: Bytes,
+	) -> RpcResult<Option<String>> {
```
niklasad1, line 126: we could probably change the return type `Option<String>` because this error is never used by the impl

however,  https://github.com/paritytech/jsonrpsee/blob/master/core/src/server/mod.rs#L153-#L171 IntoResponse is not implemented for Option<String> so never mind.

niklasad1 (approved):

substrate/client/rpc-spec-v2/src/transaction/tests/transaction_broadcast_tests.rs
lexnv: I've renamed this `operation_id_limit_reached` since we still need the `operation_id` for below, thanks!

Issue #3192, opened 2024-02-02
UnknownOpcode(192) again


nazar-pc:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

Yes, that same famous issue again.

Using Substrate derived from 127b9bec15f8a83964ef6049209df75da47b1479 and `nightly-2024-01-19` version of Rust I'm still getting "UnknownOpcode(192)" with modern version of LLVM.

As such I believe the issue is still very much present despite https://github.com/paritytech/polkadot-sdk/issues/1755 being resolved by building standard library from source.

I believe some long-term solution needs to be found for this issue at last.

cc @koute

bkchr:
Are you sure that you use the latest `wasm-builder`? Because it should recompile `std` to prevent this, but maybe there is some new unknown op code.

nazar-pc:
It is the same opcode 192 from my understanding. And yes, I am sure `wasm-builder` mentioned commit is what I use and it does contain mentioned fix with std recompilation.

bkchr:
Okay, thank you for confirming!

s0me0ne-unkn0wn:
@koute maybe it's time to consider enabling sign-ext... Again?
It's pretty much the de-facto standard throughout the Wasm ecosystem nowadays.
We could even do that deterministically, the same way it's implemented for bulk memory proposal now, although that would require more time to introduce a new executor parameter, make a release, wait for the nodes to update, etc.
I understand the whole Wasm thing is in a "just support it" state now, but that's something really annoying.

koute:
Yeah, this shouldn't happen anymore.

@nazar-pc Could you please post exact reproduction steps?

> I believe some long-term solution needs to be found for this issue at last.

Yes; the long-term solution is most likely getting rid of WASM altogether. (:

> @koute maybe it's time to consider enabling sign-ext... Again?

Well, we could do that; if you want to, be my guest! We'd need to:

1) enable the signext feature in `parity-wasm`,
2) to preserve current behavior add an explicit check when loading `.wasm` blobs that would reject blobs containing those instructions,
3) expose a flag in the executor crate to allow it to be enabled, 
4) add an on-chain knob that could be atomically enabled and link it to the flag,
5) write a Polkadot RFC and get it accepted (unless we make it parachain only - which technically doesn't need RFCs - but it probably doesn't make sense to make it parachain-only),
[...]

PR #3777, opened 2024-03-21
wasm-builder: Also set `mcpu` for c deps

bkchr:
Closes: https://github.com/paritytech/polkadot-sdk/issues/3192


michalkucharczyk (approved):

dmitry-markin (approved):

koute (approved):
Note that this could in theory break builds if someone has non-Rust dependencies in their build deps, in which case it would use this flag to compile them too, and since `mvp` is only valid for WASM it might break them**.

I don't know of a better way of doing this though.

** - interestingly, if I pass `-mcpu=mvp` to clang I get this and it compiles just fine (it completely ignores the flags, which, AFAIK, it shouldn't?):

```
$ clang -mcpu=mvp --target=x86_64-unknown-linux-gnu hello.c 
clang-16: warning: argument unused during compilation: '-mcpu=mvp' [-Wunused-command-line-argument]
```

but if I change the target to something else then I get this:

```
$ clang -mcpu=mvp --target=aarch64-unknown-linux-gnu hello.c 
clang-16: error: unsupported argument 'mvp' to option '-mcpu='
```

...which is a little strange that it's specifically hardcoded to ignore this only on AMD64.

s0me0ne-unkn0wn (approved):

PR #3780, opened 2024-03-21
Add a linear fee multiplier to ensure safety margins

claravanstaden:
Bridging fees are calculated using a static ETH/DOT exchange rate that can deviate significantly from the real-world exchange rate. We therefore need to add a safety margin to the fee so that users almost aways cover the cost of relaying.

# FAQ
> Why introduce a multiplier parameter instead of configuring an exchange rate which already has a safety factor applied?

When converting from ETH to DOT, we need to divide the multiplier by the exchange rate, and to convert from DOT to ETH we need to multiply the multiplier by the exchange rate.

> Other input parameters to the fee calculation can also deviate from real-world values. These include substrate weights, gas prices, and so on. Why does the multiplier introduced here not adjust those?

A single scalar multiplier won't be able to accommodate the different volatilities efficiently. For example, gas prices are much more volatile than exchange rates, and substrate weights hardly ever change.

So the pricing config relating to weights and gas prices should already have some appropriate safety margin pre-applied.

# Detailed Changes:
- Added multiplier field to PricingParameters
- Outbound-queue fee is multiplied by multiplier
- This multiplier is synced to the Ethereum side
- Improved Runtime API for calculating outbound-queue fees. This API makes it much easier to for configure parts of the system in preparation for launch.
- Improve and clarify code documentation
cc @vgeddes

### DIFF
--- a/bridges/snowbridge/pallets/inbound-queue/src/mock.rs
+++ b/bridges/snowbridge/pallets/inbound-queue/src/mock.rs
@@ -173,7 +173,8 @@ parameter_types! {
 	pub Parameters: PricingParameters<u128> = PricingParameters {
 		exchange_rate: FixedU128::from_rational(1, 400),
 		fee_per_gas: gwei(20),
-		rewards: Rewards { local: DOT, remote: meth(1) }
+		rewards: Rewards { local: DOT, remote: meth(1) },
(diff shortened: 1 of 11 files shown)


claravanstaden:
Closing, @vgeddes will open a PR.

Issue #3083, opened 2024-01-26
[RPC-Spec-V2] Add tests for `transactionWatch_unstable_submitAndWatch`


lexnv:
The https://github.com/paritytech/polkadot-sdk/pull/3079 introduces a test harness for the transaction class.

Extend the tests with various scenarios to validate the behavior of `transactionWatch_unstable_submitAndWatch`.
The status of the transaction should be correlated with the state of the transaction pool. 
A few scenarios ideas:
- invalid transaction is not included to the chain
- a valid transaction is included when the transaction pool encounters a test driven `ChainEvent::Finalized`
- transaction status advances through `Ready`, `InBlock`, `Finalized` with events driven from the transaction pool

cc @paritytech/subxt-team

PR #3784, opened 2024-03-21
rpc-v2/transaction: Generate `Invalid` events and add tests

lexnv:
This PR ensures that the transaction API generates an `Invalid` events for transaction bytes that fail to decode. 

The spec mentioned the `Invalid` event at the jsonrpc error section, however this spec PR makes things clearer:
- https://github.com/paritytech/json-rpc-interface-spec/pull/146

While at it have discovered an inconsistency with the generated events. The drop event from the transaction pool was incorrectly mapped to the `invalid` event.

Added tests for the API stabilize the API soon:
- https://github.com/paritytech/json-rpc-interface-spec/pull/144

Closes: https://github.com/paritytech/polkadot-sdk/issues/3083

cc @paritytech/subxt-team


substrate/client/rpc-spec-v2/src/transaction/transaction.rs
```diff
@@ -83,17 +78,21 @@ where
...
 			let decoded_extrinsic = match TransactionFor::<Pool>::decode(&mut &xt[..]) {
 				Ok(decoded_extrinsic) => decoded_extrinsic,
 				Err(e) => {
-					let err = ErrorObject::owned(
-						BAD_FORMAT,
-						format!("Extrinsic has invalid format: {}", e),
-						None::<()>,
+					log::debug!(target: LOG_TARGET, "Extrinsic bytes cannot be decoded: {:?}", e);
+
+					// The transaction is invalid.
```
niklasad1, line 86: I would move this comment down 2 lines above `let msg = to_sub_message`

bkchr (approved):

niklasad1 (approved):

Issue #3673, opened 2024-03-13
Make polkadot behave correctly if node changes peer-id at restart


alexggh:
During the investigation for https://github.com/paritytech/polkadot-sdk/issues/3314 we discovered there is a number of validators that change their PeerId at every restart, that seems not work properly right now. 

## Work-arrounds

1. Don't change PeerId at every restart, the peer-id is saved on disk next to the databse in the `network` folder, so you can simply reuse that by always providing `--base-path` argument and making sure the network folder is present where you restart the node.
2. If you restarted already, key rotations would fix the issue.

## [WIP] Things that need fixing

In general all the assumptions where we think we have just one PeerID per AuthorithyId are breaking this on a quick scan by me and @bkchr:

With kusama size nodes won't be able to connect to nodes if they aren't in the reserved set.
https://github.com/paritytech/polkadot-sdk/blob/809d46657f588ff938ca2e9fa1208e7ff21f33fe/polkadot/node/network/protocol/src/peer_set.rs#L95-L96

Some places where the assumption that just one peerID per authority exist.

https://github.com/paritytech/polkadot-sdk/blob/809d46657f588ff938ca2e9fa1208e7ff21f33fe/polkadot/node/network/gossip-support/src/lib.rs#L408

https://github.com/paritytech/polkadot-sdk/blob/a756baf3b211efc57ca4a0089b7ceba4c20b9f12/polkadot/node/network/statement-distribution/src/v2/mod.rs#L451

Another thing that I observed that is probably off while testing on versi is that `ConnectToResolvedValidators` gets called sometime with the new address and sometimes with old-address.

bkchr:
The first link is not really a problem. Just more the explanation why these people could not connect to all the other validators.

https://github.com/paritytech/polkadot-sdk/blob/809d46657f588ff938ca2e9fa1208e7ff21f33fe/polkadot/node/network/gossip-support/src/lib.rs#L327-L352

This function needs to be called multiple times per session. Currently we don't call this once and if a validator address was not known at this point, we will never resolve it in this session.

alexggh:
> The first link is not really a problem

Messed up the links, fixed it in the description.

alexggh:
Looked a bit more closer at `Kademlia` DHT and the biggest problem for this use-case seem to be is that both records will continue living in the DHT on different nodes, so on any given node any of the following might be true:

1.  When querying the DHT a node discovers only the old record.
2. When querying the DHT a node discovers only the new record.
[...]

PR #3786, opened 2024-03-21
authorithy-discovery: Make changing of peer-id while active a bit more robust

alexggh:
In the case when nodes don't persist their node-key or they want to generate a new one while being in the active set, things go wrong because both the old addresses and the new ones will still be present in DHT, so because of the distributed nature of the DHT both will survive in the network untill the old ones expires which is 36 hours. Nodes in the network will randomly resolve the authorithy-id to the old address or the new one.

More details in: https://github.com/paritytech/polkadot-sdk/issues/3673

This PR proposes we mitigate this problem, by:

1. Let the query for a DHT key retrieve more than one results(4), that is also bounded by the replication factor which is 20, currently we interrupt the querry on the first result.
~2. Modify the authority-discovery service to keep all the discovered addresses around for 24h since they last seen an address.~
~3. Plumb through other subsystems where the assumption was that an authorithy-id will resolve only to one PeerId. Currently, the authorithy-discovery keeps just the last record it received from DHT and queries the DHT every 10 minutes. But they could always receive only the old address, only the new address or a flip-flop between them depending on what node wins the race to provide the record~

2. Extend the `SignedAuthorityRecord` with a signed creation_time.
3. Modify authority discovery to keep track of nodes that sent us old record and once we are made aware of a new record update the nodes we know about with the new record.
4. Update gossip-support to try resolve authorities more often than every session.

~This would gives us a lot more chances for the nodes in the networks to also discover not only the old address of the node but also the new one and should improve the time it takes for a node to be properly connected in the network. The behaviour won't be deterministic because there is no guarantee the all nodes will see the new record at least once, since they could query only nodes that have the old one.~

## TODO
- [x] Add unittests for the new paths.
- [x] Make sure the implementation is backwards compatible
- [x]  Evaluate if there are any bad consequence of letting the query continue rather than finish it at first record found.
- [x] Bake in versi the new changes.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5605781

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5605800

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5605799

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5612017

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5612019

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -139,6 +139,11 @@ pub struct Worker<Client, Network, Block, DhtEventStream> {
 	/// Set of in-flight lookups.
 	in_flight_lookups: HashMap<KademliaKey, AuthorityId>,
 
+	/// Set of lookups we can still received records.
+	/// These are the entries in the `in_flight_lookups` for which
+	/// we got at least on successfull result.
```
lexnv, line 144: nit: one

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -550,10 +555,16 @@ where
...
 
-		let authority_id: AuthorityId = self
-			.in_flight_lookups
-			.remove(&remote_key)
-			.ok_or(Error::ReceivingUnexpectedRecord)?;
+		let authority_id: AuthorityId =
+			if let Some(authority_id) = self.in_flight_lookups.remove(&remote_key) {
+				authority_id
+			} else if let Some(authority_id) = self.known_lookups.get(&remote_key) {
+				authority_id.clone()
```
dmitry-markin, line 562: It seems this is going to never be called, as in `Discovery` behavior we terminate DHT queries after receiving the first result:
https://github.com/paritytech/polkadot-sdk/blob/e88d1cb79315792a3dbccb6bdef2543093ecaf5b/substrate/client/network/src/discovery.rs#L793-L799

dmitry-markin: Aah, sorry, missed that you also updated the `Discovery` :see_no_evil:

substrate/client/network/src/discovery.rs
```diff
@@ -785,16 +784,23 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
-									"Libp2p => Found record ({:?}) with value: {:?}",
+									"Libp2p => Found record ({:?}) with value: {:?} id {:?} stats {:?}",
 									r.record.key,
 									r.record.value,
+									id,
+									stats,
 								);
 
 								// Let's directly finish the query, as we are only interested in a
 								// quorum of 1.
```
lexnv, line 795: We'd probably need to update this comment as well.

substrate/client/network/src/discovery.rs
```diff
@@ -785,16 +784,23 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
 								if let Some(kad) = self.kademlia.as_mut() {
-									if let Some(mut query) = kad.query_mut(&id) {
-										query.finish();
+									if let Some(_query) = kad.query_mut(&id) {
+										// Let the query continue, to increase the chances we
+										// discover all possible addresses, for the cases where more
+										// addresses might exist in DHT, for example when the node
+										// changes its PeerId.
+
+										// query.finish();
```
lexnv, line 803: Should we still impose a limit on received results? Just to end the query a bit early, say 16 results?

dmitry-markin: Probably makes sense to double check that our code works as expected when a record is returned more than once — e.g., the authority-discovery would have generated an error in this case before this PR.

dmitry-markin: By the way, libp2p kad spec defines a mechanism for selecting the most up to date record among the returned (and even updating the outdated records):
https://github.com/libp2p/specs/blob/master/kad-dht/README.md#value-retrieval

But I don't think we can select the most recent PeerId at this point, can't we?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -550,10 +555,16 @@ where
...
+			} else {
+				return Err(Error::ReceivingUnexpectedRecord);
+			};
+
+		self.known_lookups.insert(remote_key.clone(), authority_id.clone());
 
 		let local_peer_id = self.network.local_peer_id();
```
dmitry-markin, line 570: On line 593 below (can't add comment there) we generate an error if the records contain different peer ids. Probably should also be updated.

polkadot/node/network/gossip-support/src/lib.rs
```diff
@@ -350,6 +374,57 @@ where
...
+	) where
+		Sender: overseer::GossipSupportSenderTrait,
+	{
+		let (_, resolved, _) = self.resolve_authorities(authorities).await;
+
+		let mut changed = Vec::new();
+
+		for (authority, new_addresses) in &resolved {
+			match self.resolved_authorities.get(authority) {
+				Some(old_address) if !old_address.is_superset(new_addresses) =>
```
dmitry-markin, line 393:
```suggestion
				Some(old_addresses) if !old_addresses.is_superset(new_addresses) =>
```

polkadot/node/network/gossip-support/src/lib.rs
```diff
@@ -259,7 +281,12 @@ where
 							// to clean up all connections.
 							Vec::new()
 						};
-					self.issue_connection_request(sender, connections).await;
+
+					if force_request || is_new_session {
+						self.issue_connection_request(sender, connections).await;
+					} else if re_resolve_authorities {
+						self.issue_connection_request_to_changed(sender, connections).await;
+					}
```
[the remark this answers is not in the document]
alexggh: Yeah, I'm more concerned that the other way around we may end-up calling re_resolve path instead of issue_connection_request at the first block in the session. But, I agree this is not obvious we don't affect the previous path I'll make it cleaner.

substrate/client/network/src/discovery.rs
alexggh: > Should we still impose a limit on received results? Just to end the query a bit early, say 16 results?

That seems to be already bound by the replication factor which is 20.

> Probably makes sense to double check that our code works as expected when a record is returned more than once.

Any particular path that you think might be troublesome ?

dmitry-markin: > > Probably makes sense to double check that our code works as expected when a record is returned more than once.
> 
> Any particular path that you think might be troublesome ?

Nothing specific, just thinking out loud.

alexggh:
Ended it up going in a slightly different direction on this PR and decided that the right fix here is to add a  signed creation_time to `SignedAuthorityRecord` with that we have a reliable why to decide which record is the newest, so the new algorithm is like that:
1. When publishing a new record nodes add the creation_time and sign it.
2. When querying the DHT, nodes query more than one node and keep track of the nodes that have the latest record.
3. When receive newer records we update the AddressCache and send the new record to nodes that we know have old records.

This is basically a variant of the suggestions from here: https://github.com/libp2p/specs/blob/master/kad-dht/README.md#value-retrieval

This changes should make network converge way faster to the new record than in the present.

@bkchr @dmitry-markin @lexnv  Let me know if you think this approach is reasonable and if you foresee any problems with it.

dmitry-markin:
> Ended it up going in a slightly different direction on this PR and decided that the right fix here is to add a signed creation_time to `SignedAuthorityRecord`

This seems reasonable to me, provided care is taken to preserve backward compatibility of the authority-discovery records.

alexggh:
> This seems reasonable to me, provided care is taken to preserve backward compatibility of the authority-discovery records.

Yeah, definitely it needs to be backwards compatible, the good part is that the records are serialized as protobuffs  and that makes it really easy to add new fields in a backwards compatible way.

alexggh:
This is now ready for review, I tested the following scenarios on versi with 30 nodes:
1. Split the network in two halfs master and this PR, all nodes are correctly detected.
2. Run this PR and changed the PeerID for half of the network, the network quickly(minutes) converges to the new PeerID, and properly connect to all authorities.

Let me know me if you think there are other scenarios that are worth checking.

substrate/client/authority-discovery/src/worker/schema/dht-v3.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package authority_discovery_v3;
+
+// First we need to serialize the addresses in order to be able to sign them.
+message AuthorityRecord {
+	// Possibly multiple `MultiAddress`es through which the node can be
```
dmitry-markin, line 7:
```suggestion
	// Possibly multiple `MultiAddress`es through which the node can be reached.
```

substrate/client/authority-discovery/src/worker/schema/dht-v3.proto
```diff
@@ -0,0 +1,33 @@
...
+	// A signature of the creation time.
+	bytes signature = 2;
+}
+
+// Then we need to serialize the authority record and signature to send them over the wire.
+message SignedAuthorityRecord {
+	bytes record = 1;
+	bytes auth_signature = 2;
+	// Even if there are multiple `record.addresses`, all of them have the same peer id.
+    // Old versions are missing this field. It is optional in order to provide compatibility both ways.
```
dmitry-markin, line 29: spaces -> tab
```suggestion
	// Old versions are missing this field. It is optional in order to provide compatibility both ways.
```

substrate/client/network/src/discovery.rs
```diff
@@ -409,6 +408,25 @@ impl DiscoveryBehaviour {
 		}
 	}
 
+	/// Puts a record into the DHT on the provided `peers`
+	///
+	/// If `update_local_storage` is true, the local storage is update as well.
+	pub fn put_record_to(
+		&mut self,
+		record: PeerRecord,
```
dmitry-markin, line 416: As only `record.record` is ever used, maybe pass `record: Record` directly?

substrate/client/network/src/discovery.rs
```diff
@@ -409,6 +408,25 @@ impl DiscoveryBehaviour {
...
+		peers: HashSet<PeerId>,
+		update_local_storage: bool,
+	) {
+		if let Some(kad) = self.kademlia.as_mut() {
+			if update_local_storage {
+				if let Err(_e) = kad.store_mut().put(record.record.clone()) {
+					warn!(target: "sub-libp2p", "Failed to update local starage");
+				}
+			}
+			kad.put_record_to(record.record, peers.into_iter(), Quorum::One);
```
dmitry-markin, line 426: Why are we not passing `Quorum::All`?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -668,6 +752,60 @@ where
 		Ok(())
 	}
 
+	fn handle_new_record(
+		&mut self,
+		authority_id: &AuthorityId,
+		kadmelia_key: KademliaKey,
```
dmitry-markin, line 758: `kadmelia_key` -> `kademlia_key`

substrate/client/authority-discovery/src/worker/schema/dht-v3.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package authority_discovery_v3;
```
lexnv, line 3: dq: Do we need to change a spec for this? Just to avoid silent changes for others (like lightclient).

substrate/client/authority-discovery/src/worker/schema/dht-v3.proto
```diff
@@ -0,0 +1,33 @@
...
+	bytes signature = 2;
+}
+
+// Then we need to serialize the authority record and signature to send them over the wire.
+message SignedAuthorityRecord {
+	bytes record = 1;
+	bytes auth_signature = 2;
+	// Even if there are multiple `record.addresses`, all of them have the same peer id.
+    // Old versions are missing this field. It is optional in order to provide compatibility both ways.
+	PeerSignature peer_signature = 3;
```
lexnv, line 30: I think we can make the peer signature adn timestamp non-optional and also close: https://github.com/paritytech/polkadot-sdk/issues/3824.

substrate/client/authority-discovery/src/worker/schema/tests.rs
```diff
@@ -74,6 +74,7 @@ fn v1_decodes_v2() {
 		record: vec_record_v2.clone(),
 		auth_signature: vec_auth_signature.clone(),
 		peer_signature: Some(peer_signature_v2.clone()),
+		creation_time: None,
```
lexnv, line 77: Would be good to have some extra tests here: v2<->v3 and v1<->v3

substrate/client/authority-discovery/src/worker/tests.rs
```diff
@@ -592,17 +640,250 @@ fn strict_accept_address_with_peer_signature() {
...
 
 	assert_eq!(
-		Some(&HashSet::from([addr])),
+		Some(HashSet::from([addr])),
 		cached_remote_addresses,
 		"Expect worker to only cache `Multiaddr`s with `PeerId`s.",
 	);
 }
 
+#[test]
```
lexnv, line 655: Nice!

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -753,6 +891,7 @@ fn sign_record_with_authority_ids(
 	peer_signature: Option<schema::PeerSignature>,
 	key_store: &dyn Keystore,
 	keys: Vec<AuthorityId>,
+	include_creation_time: bool,
```
dmitry-markin, line 894: Is there a reason to not assume it's always `true`?

substrate/client/network/src/discovery.rs
```diff
@@ -785,16 +803,24 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
-								if let Some(kad) = self.kademlia.as_mut() {
-									if let Some(mut query) = kad.query_mut(&id) {
-										query.finish();
+								// Let's directly finish the query if we are above 4.
+								// This number should small enough to make sure we don't
+								// unnecessarily flood the network with queries, but high
+								// enough to make sure we also touch peers which might have
+								// old record, so that we can update them once we notice
+								// they have old records.
+								if stats.num_successes() > 4 {
```
dmitry-markin, line 819: Would be nice to make this configurable per `get_record`/`get_value` request, and only pass > 1 for authority-discovery at the caller side. At least something to consider as a TODO for a follow-up PR. In this PR it's worth extracting 4 as something like `GET_RECORD_REDUNDANCY_FACTOR` to the beginning of `discovery.rs`.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -668,6 +752,60 @@ where
...
+	fn handle_new_record(
+		&mut self,
+		authority_id: &AuthorityId,
+		kadmelia_key: KademliaKey,
+		new_record: RecordInfo,
+	) -> bool {
+		let current_record_info = self.last_known_record.entry(kadmelia_key.clone()).or_default();
+		if new_record.creation_time > current_record_info.creation_time {
+			let peers_that_need_updating = current_record_info.peers_with_record.clone();
+			for record in new_record.record.iter() {
```
dmitry-markin, line 764: It isn't actually a loop, is it? May be something like:
```
new_record.record.as_ref().map(|record| { ... });
```
?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -668,6 +752,60 @@ where
...
+				self.network.put_record_to(
```
dmitry-markin, line 765: Is it possible to hit a race condition where a 3rd node has already updated the peers from `current_record_info.peers_with_record` with a more up-to-date data after we last received `current_record_info`?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -668,6 +752,60 @@ where
...
+			);
+			for record in current_record_info.record.iter() {
+				self.network.put_record_to(
+					record.clone(),
+					new_record.peers_with_record.clone(),
+					// If this is empty it means we received the answer from our node local
+					// storage, so we need to update that as well.
+					new_record.peers_with_record.is_empty(),
+				);
+			}
```
dmitry-markin, line 804:
```suggestion
			current_record_info.record.as_ref().map(|record| {
				self.network.put_record_to(
					record.clone(),
					new_record.peers_with_record.clone(),
					// If this is empty it means we received the answer from our node local
					// storage, so we need to update that as well.
					new_record.peers_with_record.is_empty(),
				);
			});
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -651,14 +711,38 @@ where
...
 			.flatten()
 			.take(MAX_ADDRESSES_PER_AUTHORITY)
 			.collect();
 
-		if !remote_addresses.is_empty() {
-			self.addr_cache.insert(authority_id, remote_addresses);
+		let mut addr_cache_needs_update = false;
+
+		let answering_peer_id = single(values.iter().map(|record| record.peer))
+			.map_err(|_| Error::ReceivingDhtValueFoundFromDifferentPeerIds)?
```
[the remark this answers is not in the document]
alexggh: Yeah, I was thinking about that.

substrate/client/authority-discovery/src/worker.rs
alexggh: It is false in some of the tests, to check we accept records without creation time.

substrate/client/network/src/discovery.rs
alexggh: Will do.

substrate/client/authority-discovery/src/worker.rs
alexggh: Maybe let me think about it.

substrate/client/network/src/discovery.rs
alexggh: Good point, it should be all.

substrate/client/authority-discovery/src/worker.rs
alexggh: Thought about it for a bit I don't think this should be a problem, because we alway push for the newer record and records get published periodically, all nodes all eventually converge to the newest record.

alexggh (commented):
@dmitry-markin  I think I addressed all the review feedback I got, unfortunately, it seems now I have another problem litep2p is missing some of the behaviours I was using in this PR, to be more specific:

1. Litep2p does not provide the PeerID where we found the record.
2. Litep2p does not implement the put_record_to(peer_id) API.

So, what is the maintenance  policy have we got, I assume, my only option is to go in litep2p and implement the missing behaviour, is that right ?

substrate/client/authority-discovery/src/worker/schema/dht-v3.proto
alexggh: I would keep that separate.

dmitry-markin:
@alexggh I see @lexnv has already addressed 1) in https://github.com/paritytech/litep2p/pull/76. As for 2), it should be not that difficult to implement `put_record_to` as we need to only contact one specific peer and not execute a full-fledged query. One thing that might complicate things is that litep2p kad implements a backpressure mechanism, so `put_record_to` would be an async function. As a fast & dirty way of fixing things you might want to introduce `try_put_record_to` that fails if the command stream in litep2p kad is clogged, and create an issue to follow up on that.

I don't know if @lexnv is already looking into the issue. If not, I can look into adding `put_record_to`.

lexnv:
@alexggh We have merged the required PRs to unblock this functionality. I think this needs a cargo update for the litep2p crate, let us know if there's anything missing 🙏

alexggh:
> @alexggh We have merged the required PRs to unblock this functionality. I think this needs a cargo update for the litep2p crate, let us know if there's anything missing 🙏

Thank you, appreciated the really quick turnaround, will update litep2p and run some tests in the following days, I'll keep you updated!

Cargo.toml
```diff
@@ -646,3 +646,6 @@ wasmi = { opt-level = 3 }
 x25519-dalek = { opt-level = 3 }
 yamux = { opt-level = 3 }
 zeroize = { opt-level = 3 }
+
+[patch."https://github.com/paritytech/litep2p"]
```
alexggh, line 650: Will be remove before merging once: https://github.com/paritytech/litep2p/pull/96, gets merged.

alexggh:
Ran the pull request on Versi, things work as expected.

polkadot/node/network/gossip-support/src/tests.rs
```diff
@@ -807,6 +815,313 @@ fn issues_update_authorities_after_session() {
 	);
 }
 
+// Test we connect to authorities that changed their address `TRY_RERESOLVE_AUTHORITIES` rate
+// and that is is no-op if no authority changed.
+#[test]
+fn test_quickly_connect_to_authorithies_that_changed_address() {
```
dmitry-markin, line 821: nit
```suggestion
fn test_quickly_connect_to_authorities_that_changed_address() {
```

substrate/client/authority-discovery/src/worker/tests.rs
```diff
@@ -611,15 +674,240 @@ fn strict_accept_address_with_peer_signature() {
...
+	let kademlia_key = hash_authority_id(tester.remote_authority_public.as_slice());
+	for key in kv_pairs.iter_mut() {
+		key.0 = kademlia_key.clone();
+	}
+	let cached_remote_addresses = tester.process_value_found(true, kv_pairs).0;
+
+	assert_eq!(
+		Some(HashSet::from([addr])),
+		cached_remote_addresses,
+		"Expect worker to only cache `Multiaddr`s with `PeerId`s.",
```
dmitry-markin, line 803:
```suggestion
		"Expect `Multiaddr` to remain the same",
```
?

substrate/client/authority-discovery/src/worker/tests.rs
```diff
@@ -611,15 +674,240 @@ fn strict_accept_address_with_peer_signature() {
...
+		Some(&TestSigner { keypair: &tester.remote_node_key }),
+		false,
+	);
+
+	let (cached_remote_addresses, network) = tester.process_value_found(true, kv_pairs);
+
+	assert_eq!(
+		Some(HashSet::from([addr])),
+		cached_remote_addresses,
+		"Expect worker to only cache `Multiaddr`s with `PeerId`s.",
```
dmitry-markin, line 727: This text description seems copied across the tests. Is it relevant in all the places?

dmitry-markin (approved):
LGTM :+1:

substrate/client/network/src/discovery.rs
```diff
@@ -92,8 +92,12 @@ const MAX_KNOWN_EXTERNAL_ADDRESSES: usize = 32;
 /// record is replicated to.
 pub const DEFAULT_KADEMLIA_REPLICATION_FACTOR: usize = 20;
 
+// The minimum number of peers we expect an answer before we terminate the request.
+const GET_RECORD_REDUNDANCY_FACTOR: u32 = 4;
```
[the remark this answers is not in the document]
alexggh: Will do.

substrate/client/authority-discovery/src/worker/tests.rs
alexggh: Updated with proper messages

substrate/client/network/src/litep2p/mod.rs
```diff
@@ -795,31 +802,70 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkBackend<B, H> for Litep2pNetworkBac
...
+													key: record.key.to_vec().into(),
+													value: record.value,
+													publisher: record.publisher.map(|peer_id| {
+														let peer_id: sc_network_types::PeerId = peer_id.into();
+														peer_id.into()
+													}),
+													expires: record.expires,
+												},
+												peer: None,
+											}
```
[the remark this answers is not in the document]
alexggh: Done.

bkchr (changes requested):
Changing the network protocol should be done via a RFC.

alexggh:
> https://github.com/paritytech/polkadot-sdk/pull/3786#discussion_r1601609034
> Yes this should go through an RFC. While this being compatible, it is still a change to the protocol.a

@bkchr: That's unfortunate, I was hoping to fix this scenario sooner rather than later, somehow I missed that this change would warrant an RFC,  I'll prepare one these days.

bkchr:
The RFC should be quite forward. This changes the protocol and while it stays compatible it should go through a RFC.

@lexnv also had this thought. Sorry for taking that long to look at this!

alexggh:
> The RFC should be quite forward. This changes the protocol and while it stays compatible it should go through a RFC.
> 
> @lexnv also had this thought. Sorry for taking that long to look at this!

Posted the RFC here: https://github.com/polkadot-fellows/RFCs/pull/91

alexggh:
> > The RFC should be quite forward. This changes the protocol and while it stays compatible it should go through a RFC.
> > @lexnv also had this thought. Sorry for taking that long to look at this!
> 
> Posted the RFC here: [polkadot-fellows/RFCs#91](https://github.com/polkadot-fellows/RFCs/pull/91)

RFC is now merged, this should be ready to be merged as well, @dmitry-markin, @bkchr, @lexnv when you've got time can you have another go at it.  Thank you!

substrate/client/network/src/litep2p/mod.rs
```diff
@@ -816,26 +822,23 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkBackend<B, H> for Litep2pNetworkBac
...
+									);
+								}
 
 								if let Some(ref metrics) = self.metrics {
 									metrics
 										.kademlia_query_duration
 										.with_label_values(&["value-get"])
 										.observe(started.elapsed().as_secs_f64());
 								}
+
```
dmitry-markin, line 841: nit: blank line
```suggestion
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -159,6 +160,16 @@ pub struct Worker<Client, Block: BlockT, DhtEventStream> {
 	/// Set of in-flight lookups.
 	in_flight_lookups: HashMap<KademliaKey, AuthorityId>,
 
+	/// Set of lookups we can still received records.
```
dmitry-markin, line 163:
```suggestion
	/// Set of lookups we can still receive records.
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -772,6 +846,59 @@ where
 		Ok(())
 	}
 
+	fn handle_new_record(
```
dmitry-markin, line 849: Adding the description for the return value would be nice.

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -772,6 +846,59 @@ where
...
+			);
+			debug!(
+					target: LOG_TARGET,
+					"Found a newer record for {:?} new record creation time {:?} old record creation time {:?}",
+					authority_id, new_record.creation_time, current_record_info.creation_time
+			);
+			self.last_known_records.insert(kademlia_key, new_record);
+			true
+		} else if new_record.creation_time == current_record_info.creation_time {
+			// Same record just update in case this is a record from old nods that don't have
```
dmitry-markin, line 876:
```suggestion
			// Same record just update in case this is a record from old nodes that don't have
```

dmitry-markin (approved):
Looks good!

substrate/client/authority-discovery/src/worker/tests.rs
```diff
@@ -640,17 +698,242 @@ fn strict_accept_address_with_peer_signature() {
...
+		"Expect `Multiaddr` to remain the same",
+	);
+	assert!(network
+		.as_ref()
+		.map(|network| network.put_value_to_call.lock().unwrap().is_empty())
+		.unwrap_or_default());
+}
+
+#[test]
+fn newer_records_overwrite_older_ones() {
```
lexnv, line 836: Nice!

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -168,6 +179,16 @@ pub struct Worker<Client, Block: BlockT, DhtEventStream> {
...
 }
 
+#[derive(Debug, Clone)]
+struct RecordInfo {
+	// Time since UNIX_EPOCH in nanoseconds.
+	creation_time: u128,
+	// Peers that we know have this record.
+	peers_with_record: HashSet<PeerId>,
+	// The record itself.
+	record: Record,
```
lexnv, line 189:
```suggestion
	/// Time since UNIX_EPOCH in nanoseconds.
	creation_time: u128,
	/// Peers that we know have this record.
	peers_with_record: HashSet<PeerId>,
	/// The record itself.
	record: Record,
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -539,15 +567,13 @@ where
 				}
 
 				if log_enabled!(log::Level::Debug) {
-					let hashes: Vec<_> = v.iter().map(|(hash, _value)| hash.clone()).collect();
-					debug!(target: LOG_TARGET, "Value for hash '{:?}' found on Dht.", hashes);
+					debug!(target: LOG_TARGET, "Value for hash '{:?}' found on Dht.", v.record.key);
 				}
```
lexnv, line 571: Since we don't allocate anymore here:

```suggestion

				debug!(target: LOG_TARGET, "Value for hash '{:?}' found on Dht.", v.record.key);

```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -701,67 +752,90 @@ where
...
+				authority_id.clone()
+			} else {
+				return Err(Error::ReceivingUnexpectedRecord);
+			};
 
-		let authority_id: AuthorityId = self
-			.in_flight_lookups
-			.remove(&remote_key)
-			.ok_or(Error::ReceivingUnexpectedRecord)?;
+		self.known_lookups.insert(remote_key.clone(), authority_id.clone());
```
lexnv, line 769: Let's remove the double insert into known_lookups :D

```suggestion
		let authority_id: AuthorityId =
			if let Some(authority_id) = self.in_flight_lookups.remove(&remote_key) {
                self.known_lookups.insert(remote_key.clone(), authority_id.clone());
				authority_id
			} else if let Some(authority_id) = self.known_lookups.get(&remote_key) {
				authority_id.clone()
			} else {
				return Err(Error::ReceivingUnexpectedRecord);
			};
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -701,67 +752,90 @@ where
...
-					&authority_id,
-				)?;
-				Ok(addresses)
+		let schema::SignedAuthorityRecord { record, peer_signature, .. } =
+			Self::check_record_signed_with_authority_id(
+				peer_record.record.value.as_slice(),
+				&authority_id,
+			)?;
+
+		let authority_record = schema::AuthorityRecord::decode(record.as_slice());
```
lexnv, line 778: nit: Should we `.map_err(Error::DecodingProto)?` to simplify a bit below?

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -772,6 +846,61 @@ where
...
+			self.network.put_record_to(
+				new_record.record.clone(),
+				peers_that_need_updating.clone(),
+				// If this is empty it means we received the answer from our node local
+				// storage, so we need to update that as well.
+				current_record_info.peers_with_record.is_empty(),
+			);
+			debug!(
+					target: LOG_TARGET,
+					"Found a newer record for {:?} new record creation time {:?} old record creation time {:?}",
```
lexnv, line 872: nit: We can return here and get rid of indentation

```
if new_record.creation_time > current_record_info.creation_time {
    ..
    return true
}

if new_record.creation_time == current_record_info.creation_time {
    ..
    return true
}
..
```

substrate/client/authority-discovery/src/worker.rs
```diff
@@ -168,6 +179,16 @@ pub struct Worker<Client, Block: BlockT, DhtEventStream> {
 	phantom: PhantomData<Block>,
 }
 
+#[derive(Debug, Clone)]
+struct RecordInfo {
+	// Time since UNIX_EPOCH in nanoseconds.
+	creation_time: u128,
+	// Peers that we know have this record.
+	peers_with_record: HashSet<PeerId>,
```
lexnv, line 187: Should we bound this to a decent number? We could have authority_keys * num_peers total?

lexnv:
LGTM! Maybe could bound the total number of peers in the record info using an LRU with a decent amount. In theory to not update every outdated peer (whole discoverable network) from this node, but only a limited number of peers?

lexnv (approved):
LGTM! Maybe could bound the total number of peers in the record info using an LRU with a decent amount. In theory to not update every outdated peer (whole discoverable network) from this node, but only a limited number of peers?

substrate/client/authority-discovery/src/worker.rs
alexggh: We shouldn't get more than 20(K_FACTOR) here, but you are right there is no hurt in explicitly bound it, so I'll do that.

substrate/client/authority-discovery/src/worker.rs
alexggh: Done.

substrate/client/authority-discovery/src/worker.rs
alexggh: Bounded it to DEFAULT_KADEMLIA_REPLICATION_FACTOR, there is no need to over-complicate this with a LRU.

prdoc/pr_3786.prdoc
```diff
@@ -0,0 +1,22 @@
...
+- name: sc-authority-discovery
+  bump: major
+- name: polkadot-gossip-support
+  bump: major
+- name: polkadot-network-bridge
+  bump: major
+- name: polkadot-node-subsystem-types
+  bump: major
+- name: sc-network
+  bump: minor
```
lexnv, line 22: nit; new line to pass prdoc? 🤔

alexggh:
Validate this PR in versi for the past 2 days with both litep2p and libp2p backends, things seems to work as expected, merging it now.

PR #3804, opened 2024-03-23
[parachain-template] refactor pallet configurations for enhanced modularity

asiniscalchi:
This series of patches refactors the configuration of various pallets into their own separate modules. The primary goal of this refactor is to improve the modularity and readability of the runtime configuration. 

- Moves the configuration for each pallet into its own module under `runtime/src/configs`. This includes configurations for `aura`, `authorship`, `balances`, and several others. 
- Each module contains the configuration parameter


bkchr:
I'm not convinced that this improves the readability. You split up 250 lines across multiple files. IMO this makes it much harder to find all the configs. For bigger runtimes this probably makes sense, but not for such small runtimes.

asiniscalchi:
An 800-line `lib.rs` could suggest a refactor is due for clarity and maintainability, perfectly aligning with the `parachain-template-runtime` module's goal to exemplify best practices.

bkchr:
Then maybe add all of them to the same `config.rs`. I honestly see no justification for splitting it up into that many files.

asiniscalchi:
Placing all configurations in a single config.rs file is a more cautious approach. This reflects an interest in modularization, albeit not as extensively as the current refactor proposes.
I think it's a good middle ground: https://github.com/paritytech/polkadot-sdk/pull/3809

PR #3806, opened 2024-03-23
Refactor crypto-related traits implementations in term of Public/Signature Bytes

davxy:
Another simple refactory to prune some duplicate code

Follow up of: https://github.com/paritytech/polkadot-sdk/pull/3684

### DIFF
--- a/substrate/primitives/core/src/crypto_bytes.rs
+++ b/substrate/primitives/core/src/crypto_bytes.rs
@@ -36,6 +36,9 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 #[cfg(all(not(feature = "std"), feature = "serde"))]
 use sp_std::alloc::{format, string::String};
 
+pub use public_bytes::*;
+pub use signature_bytes::*;
+
 /// Generic byte array holding some crypto-related raw data.
 ///
 /// The type is generic over a constant length `N` and a "tag" `T` which
@@ -374,6 +377,3 @@ mod signature_bytes {
 		}
 	}
 }
-
-pub use public_bytes::*;
-pub use signature_bytes::*;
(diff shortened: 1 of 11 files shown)


bkchr (approved):

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -231,14 +240,140 @@ impl<T> CryptoBytes<64, T> {
...
+
+	/// Generic encoded signature.
+	pub type SignatureBytes<const N: usize, SubTag> = CryptoBytes<N, (SignatureTag, SubTag)>;
+
+	impl<const N: usize, SubTag> Signature for SignatureBytes<N, SubTag> where Self: CryptoType {}
 
-/// Generic encoded public key.
-pub type PublicBytes<const N: usize, SubTag> = CryptoBytes<N, (PublicTag, SubTag)>;
+	#[cfg(feature = "serde")]
+	impl<const N: usize, SubTag> Serialize for SignatureBytes<N, SubTag>
```
michalkucharczyk, line 334: nit: maybe for signature we could use something like this:
https://github.com/paritytech/polkadot-sdk/blob/463ccb8f9028ebf5f7fc53d7e835c21f43172253/substrate/primitives/storage/src/lib.rs#L43 (alreay used widely in a codebase, sometimes also as `use impl_serde::serialize as bytes`.)
The code is here: https://github.com/paritytech/parity-common/blob/master/primitive-types/impls/serde/src/serialize.rs

michalkucharczyk (approved):

substrate/primitives/core/src/crypto_bytes.rs
```diff
@@ -231,14 +240,140 @@ impl<T> CryptoBytes<64, T> {
...
+		fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+			Ok(())
+		}
+	}
+}
 
-/// Generic encoded signature.
-pub type SignatureBytes<const N: usize, SubTag> = CryptoBytes<N, (SignatureTag, SubTag)>;
+pub use public_bytes::*;
+pub use signature_bytes::*;
```
michalkucharczyk, line 379: nit: move to top?

PR #3816, opened 2024-03-25
[prdoc] Require SemVer bump level

liamaharon:
A prerequisite for adding a stable branch and respecting SemVer on new stable releases is including SemVer bump levels in our PRDocs. 

Next release is scheduled for April 3rd, so it would be great to get this merged before then.

Also added "None" as a valid bump option, to support test/benchmark changes and CI to ensure changed crates have an entry.


ggwpez (approved):

kianenigma (approved):
We will never be fully ready for this until we merge it and force ourselves to be ready, nos vamos 🤠

PR #3749, opened 2024-03-19
pallet-xcm: Deprecate `execute` and `send` in favor of `execute_blob` and `send_blob`

franciscoaguirre:
`execute` and `send` try to decode the xcm in the parameters before reaching the filter line.
The new extrinsics decode only after the filter line.
These should be used instead of the old ones.

## TODO
- [x] Tests
- [x] Generate weights
- [x] Deprecation issue -> https://github.com/paritytech/polkadot-sdk/issues/3771
- [x] PRDoc
- [x] Handle error in pallet-contracts

This would make writing XCMs in PJS Apps more difficult, but here's the fix for that: https://github.com/polkadot-js/apps/pull/10350.
Already deployed! https://polkadot.js.org/apps/#/utilities/xcm

Supersedes https://github.com/paritytech/polkadot-sdk/pull/1798/

PR #3821, opened 2024-03-25
Patches for #3749

pgherveou:


pgherveou:
@franciscoaguirre feel free to tweak and merge in #3749  
That should fix the contracts stuff.

- I now assume that contracts author will pass an encoded_xcm message, so we don't have to do the encode / decode dance inside pallet_contracts, we simply forward the encoded message to the XCM Controller
- I removed some checks we were doing to ensure that no Transact were part of the XCM, this can be done directly in the XCM configuration,

franciscoaguirre (approved):
I think the only weird thing about this is the `CallFilter`. We'll have to communicate it thoroughly

PR #3825, opened 2024-03-25
[CI] Check that prdoc and Cargo.toml version bump align

ggwpez:


ggwpez:
@liamaharon going to close this for now, as your tool should be able to handle it.  
Maybe in the future we can still use both in parallel to check, but for now it can be enough just for inspiration.

PR #3820, opened 2024-03-25
Remove deprecated treasury pallet calls

chungquantin:
# ISSUE
- Link to the issue: https://github.com/paritytech/polkadot-sdk/issues/3800
# Deliverables
- [x] remove deprecated calls; (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] set explicit coded indexes for Error and Event enums, remove unused variants and keep the same indexes for the rest; (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] remove unused Config's type parameters; (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] remove irrelevant tests and adopt relevant using old api; (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] remove benchmarks for removed calls; (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/1a3d5f1f96756555ddebd1b898c03464ffffdb25)
- [x] prdoc (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] remove deprecated methods from the `treasury/README.md` and add up-to-date dispatchable functions documentation (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/d579b673672454d0dc7b5049e5cbbe6077da520b)
- [x] remove deprecated weight functions (https://github.com/paritytech/polkadot-sdk/pull/3820/commits/8f74134b82df9a6df2824bbbe1555667223f1a83)
> ### Separated to other issues
> - [ ] remove storage items like Proposals and ProposalCount, that are not used anymore

Adjust all treasury pallet instances within polkadot-sdk
- [x] `pallet_bounty`, `tip`, `child_bounties`: https://github.com/openguild-labs/polkadot-sdk/pull/3
- [x] Remove deprecated treasury weight functions used in Westend and Rococo runtime `collective-westend`, `collective-rococo`

Add migration for westend and rococo to clean the data from removed storage items
- [ ] https://github.com/paritytech/polkadot-sdk/pull/3828
# Test Outcomes
Successful tests by running `cargo test --features runtime-benchmarks`
```
running 38 tests
[...]

PR #3828, opened 2024-03-25
[FRAME] Remove storage migration type

muharem:
Introduce migration type to remove data associated with a specific storage of a pallet.

Based on existing `RemovePallet` migration type.

Required for https://github.com/paritytech/polkadot-sdk/pull/3820


substrate/frame/support/src/migrations.rs
```diff
@@ -369,6 +369,118 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+/// ```
+///
+/// WARNING: `RemoveStorage` has no guard rails preventing it from bricking the chain if the
+/// operation of removing storage for the given pallet would exceed the block weight limit.
+///
+/// If your storage has too many keys to be removed in a single block, it is advised to wait for
+/// a multi-block scheduler currently under development which will allow for removal of storage
+/// items (and performing other heavy migrations) over multiple blocks
+/// (see <https://github.com/paritytech/substrate/issues/13690>).
+pub struct RemoveStorage<P: Get<&'static str>, S: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>>(
```
muharem, line 428: it could be more general, something like `ClearPrefix`. And `RemovePallet` and `RemoveStorage` based on it. But it will require the introduction of few more types for resolving the prefix and having human readable prefix for logs. 
Perhaps copy/paste is better for simplicity.

substrate/frame/support/src/migrations.rs
```diff
@@ -369,6 +369,118 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+/// - `DbWeight` is a type that implements the [`Get`] trait for [`RuntimeDbWeight`], providing the
+///   weight for database operations.
+///
+/// On runtime upgrade, the `on_runtime_upgrade` function will clear the storage from the specified
+/// storage, logging the number of keys removed. If the `try-runtime` feature is enabled, the
+/// `pre_upgrade` and `post_upgrade` functions can be used to verify the storage removal before and
+/// after the upgrade.
+///
+/// # Examples:
+/// ```ignore
```
kianenigma, line 388: I think with a few lines of change this can be made non-ignore, can you try?

substrate/frame/support/src/migrations.rs
```diff
@@ -369,6 +369,118 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+/// pub type Executive = frame_executive::Executive<
+/// 	Runtime,
+/// 	Block,
+/// 	frame_system::ChainContext<Runtime>,
+/// 	Runtime,
+/// 	Migrations
+/// >;
+/// ```
+///
+/// WARNING: `RemoveStorage` has no guard rails preventing it from bricking the chain if the
```
kianenigma, line 421: This is good, please add it to other similar types like `RemovePallet` as well :)

substrate/frame/support/src/migrations.rs
```diff
@@ -369,6 +369,118 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+		let hashed_prefix = storage_prefix(P::get().as_bytes(), S::get().as_bytes());
+		match contains_prefixed_key(&hashed_prefix) {
+			true => log::info!("Found `{}` `{}` keys pre-removal 👀", P::get(), S::get()),
+			false => log::warn!(
+				"Migration RemoveStorage<{}, {}> can be removed (no keys found pre-removal).",
+				P::get(),
+				S::get()
+			),
+		};
+		Ok(sp_std::vec::Vec::new())
```
kianenigma, line 465:
```suggestion
		Ok(Default::default())
```

kianenigma (approved):

prdoc/pr_3828.prdoc
```diff
@@ -0,0 +1,12 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: "[FREAME] Remove storage migration type"
```
liamaharon, line 4:
```suggestion
title: "[FRAME] Remove storage migration type"
```

liamaharon (approved):

substrate/frame/support/src/migrations.rs
muharem: I checked, there is no more general migration solution like this. `RemovePallet` have it, it is where I copied this from, all credits to @liamaharon

substrate/frame/support/src/migrations.rs
muharem: I did try, requires additional deps for this pallet, example becomes very big and does not really give any assertions

substrate/frame/support/src/migrations.rs
ggwpez, line 421: Maybe a specific warning for para-chains? I think they are often not aware of such limitations.

ggwpez (approved):

substrate/frame/support/src/migrations.rs
```diff
@@ -369,6 +369,118 @@ impl<P: Get<&'static str>, DbWeight: Get<RuntimeDbWeight>> frame_support::traits
...
+					"`clear_prefix` failed to remove all keys for storage `{}` from pallet `{}`. THIS SHOULD NEVER HAPPEN! 🚨",
+					S::get(), P::get()
+				);
+				value
+			},
+		} as u64;
+
+		log::info!("Removed `{}` `{}` `{}` keys 🧹", keys_removed, P::get(), S::get());
+
+		DbWeight::get().reads_writes(keys_removed + 1, keys_removed)
```
ggwpez, line 449: The weight annotation makes it unusable for parachains anyway, since it does not report proper weight.

muharem: Not sure I understand, can you explain please?

muharem: I see now

liamaharon: I don't see :P Can you please explain?

kianenigma: The fact that `RuntimeDbWeight` only has `u64` aka. `ref_time` and no `proof_size`. 

The only warning against it is: 

https://github.com/paritytech/polkadot-sdk/blob/8eeacff484cf7c7ec6081495dcaffef5a0867bc5/substrate/primitives/weights/src/lib.rs#L79-L80

I can't find a related tracking issue for this matter, @ggwpez do you know any?

ggwpez: No there is no issue. This is legacy code and `reads_writes` should probably be deprecated...

bkchr:
@muharem what is the status of this pr?

muharem:
@bkchr I was not sure about merging it because it cannot be safely applied for a parachain migration. But now I think we will need it at least for relay chains. Also since there is a warning in the doc about this concern. I think we can merge this. If fine with you, I will merge it after CIs pass

bkchr (approved):

Issue #504, opened 2023-06-13
Remove support for legacy `ProtocolId`-based protocol names


dmitry-markin:
It's been quite a while since the genesis-hash based protocol naming convention was introduced (see issue https://github.com/paritytech/substrate/issues/7746 & PRs https://github.com/paritytech/substrate/pull/11938, https://github.com/paritytech/polkadot/pull/5870, https://github.com/paritytech/polkadot/pull/5876).

It should be safe to remove old-style legacy `ProtocolId`-based protocol names like `"/dot/sync/1"` now.

BulatSaif:
It is worth mentioning that this may be causing issues on parachains. Each `/genesis-hash/<protocol>` is duplicated by `/sub/<protocol>`. For example `/sup/kad` is present in most of the parachains causing all chains to join in a single DHT table (testnets and production network).
When I run DHT crawler for parachains I see more than 16k nodes even for small parachains which should have only 10 nodes.

Here list of the protocols reported by  parachains:  
### Kusama statemine parachain
```
INFO[0000] Agent count=1 value="Polkadot parachain/v0.9.430-9cb14fe3cee (statemine-connect-ew1-0)"
INFO[0000] 
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/block-announces/1
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/kad
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/light/2
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/state/2
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/sync/2
INFO[0000] Protocol count=1 value=/48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a/transactions/1
INFO[0000] Protocol count=1 value=/ipfs/id/1.0.0
INFO[0000] Protocol count=1 value=/ipfs/id/push/1.0.0
INFO[0000] Protocol count=1 value=/ipfs/ping/1.0.0
INFO[0000] Protocol count=1 value=/sup/block-announces/1
[...]

PR #3833, opened 2024-03-25
network/discovery: Add to DHT only peers that support genesis-based protocol

lexnv:
This PR adds to the DHT only the peers that support the genesis/fork/kad protocol.
Before this PR, any peer that supported the legacy `/kad/[id]` protocol was added to the DHT.

This is the first step in removing the support for the legacy kad protocols.

While I have adjusted unit tests to validate the appropriate behavior, this still needs proper testing in our stack.

Part of https://github.com/paritytech/polkadot-sdk/issues/504.

cc @paritytech/networking


substrate/client/network/src/discovery.rs
```diff
@@ -1181,12 +1202,33 @@ mod tests {
 			&[kademlia_protocol_name(supported_genesis_hash, None)],
 			remote_addr.clone(),
 		);
+		// Note: legacy protocol is not supported without genesis hash and fork ID,
+		// if the legacy is the only protocol supported, then the peer will not be added.
 		discovery.add_self_reported_address(
 			&another_peer_id,
 			&[legacy_kademlia_protocol_name(&supported_protocol_id)],
 			another_addr.clone(),
 		);
```
dmitry-markin, line 1211: May be make it clear by renaming `another_peer_id` -> `unsupported_peer_id` and checking before and after the call that it's `remote_peer_id` that's added at this stage? We can still use `another_peer_id` below for the peer with both protocols.

dmitry-markin (approved):

dmitry-markin:
Not sure if we should update the polkadot spec before merging this PR.

bkchr (approved):

bkchr:
@lexnv can probably merged?

PR #3835, opened 2024-03-25
migrations: prevent accidentally using unversioned migrations instead of `VersionedMigration`

### LINKED ISSUE #1324: Update VersionedMigration to ensure it is not possible to use the unchecked migration
We should have a new `UncheckedOnRuntimeUpgrade` trait, and have the unchecked migration implements it instead of `OnRuntimeUpgrade`. In this way, it will not be possible to accidental use the unchecked migration in the actual migration.

dastansam:
closes #1324 

#### Problem
Currently, it is possible to accidentally use inner unversioned migration instead of `VersionedMigration` since both implement `OnRuntimeUpgrade`. 

#### Solution

With this change, we make it clear that value of `Inner` is not intended to be used directly. It is achieved by bounding `Inner` to new trait `UncheckedOnRuntimeUpgrade`, which has the same interface (except `unchecked_` prefix) as `OnRuntimeUpgrade`.

#### `try-runtime` functions

Since developers can implement `try-runtime` for `Inner` value in `VersionedMigration` and have custom logic for it, I added the same `try-runtime` functions to `UncheckedOnRuntimeUpgrade`. I looked for a ways to not duplicate functions, but couldn't find anything that doesn't significantly change the codebase. So I would appreciate If you have any suggestions to improve this

cc @liamaharon @xlc 


dastansam:
`check-runtime-migration-westend` is failing but it seems it's been failing for several runs and error message didn't change
https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5651070

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -227,6 +227,57 @@ pub trait OnRuntimeUpgrade {
...
+	///
+	/// Internally, the default implementation
+	/// - Handles passing data from `pre_upgrade` to `post_upgrade`
+	/// - Ensure storage is not modified in `pre_upgrade` and `post_upgrade` hooks.
+	///
+	/// Combining the `pre_upgrade` -> `on_runtime_upgrade` -> `post_upgrade` logic flow into a
+	/// single method call is helpful for scenarios like testing a tuple of migrations, where the
+	/// tuple contains order-dependent migrations.
+	#[cfg(feature = "try-runtime")]
+	fn try_on_runtime_upgrade(checks: bool) -> Result<Weight, TryRuntimeError> {
```
xlc, line 260: is this really needed? I think we shouldn't invoke try runtime on `UncheckedOnRuntimeUpgrade` directly. instead, they should be called on the checked one

dastansam: yeah, I couldn't find a way to not duplicate it.
if we want to pass a custom logic for `try-runtime` we need `UncheckedOnRuntimeUpgrade` to have these functions, since we can't control impl of `OnRuntimeUpgrade` in `VersionedMigration`. I was thinking of maybe passing a new type with `try-runtime` interface to `VersionedMigration`, i.e:

```rs
pub struct VersionedMigration<const FROM: u16, const TO: u16, Inner, TryRuntimeHooks, Pallet, Weight>

// inside `OnRuntimeUpgrade::pre_upgrade()` impl for `VersionedMigration`
fn pre_upgrade() -> Result<sp_std::vec::Vec<u8>, sp_runtime::TryRuntimeError> {
let on_chain_version = Pallet::on_chain_storage_version();
if on_chain_version == FROM {
	Ok(VersionedPostUpgradeData::MigrationExecuted(TryRuntimeHooks::pre_upgrade()?).encode())
} else {
	Ok(VersionedPostUpgradeData::Noop.encode())
}
}

```

xlc: I don't really get it.
`UncheckedOnRuntimeUpgrade` should only be used by `VersionedMigration`
`VersionedMigration` already implements `OnRuntimeUpgrade`

I don't see any reason to have `try_on_runtime_upgrade` here

dastansam: ah yeah, got it. I thought you meant all functions under try-runtime. Will remove try-on-runtime-upgrade

xlc (approved):

substrate/frame/support/src/migrations.rs
```diff
@@ -71,8 +71,8 @@ use sp_std::{marker::PhantomData, vec::Vec};
 /// /// - https://internals.rust-lang.org/t/lang-team-minutes-private-in-public-rules/4504/40
 /// mod version_unchecked {
 /// 	use super::*;
-/// 	pub struct MigrateV5ToV6<T>(sp_std::marker::PhantomData<T>);
-/// 	impl<T: Config> OnRuntimeUpgrade for  VersionUncheckedMigrateV5ToV6<T> {
+/// 	pub struct VersionUncheckedMigrateV5ToV6<T>(sp_std::marker::PhantomData<T>);
+/// 	impl<T: Config> UncheckedOnRuntimeUpgrade for  VersionUncheckedMigrateV5ToV6<T> {
 /// 		// OnRuntimeUpgrade implementation...
```
liamaharon, line 76:
```suggestion
/// 		// UncheckedOnRuntimeUpgrade implementation...
```

substrate/frame/support/src/migrations.rs
```diff
@@ -71,8 +71,8 @@ use sp_std::{marker::PhantomData, vec::Vec};
 /// /// - https://internals.rust-lang.org/t/lang-team-minutes-private-in-public-rules/4504/40
 /// mod version_unchecked {
 /// 	use super::*;
-/// 	pub struct MigrateV5ToV6<T>(sp_std::marker::PhantomData<T>);
-/// 	impl<T: Config> OnRuntimeUpgrade for  VersionUncheckedMigrateV5ToV6<T> {
+/// 	pub struct VersionUncheckedMigrateV5ToV6<T>(sp_std::marker::PhantomData<T>);
+/// 	impl<T: Config> UncheckedOnRuntimeUpgrade for  VersionUncheckedMigrateV5ToV6<T> {
```
liamaharon, line 75:
```suggestion
/// 	impl<T: Config> UncheckedOnRuntimeUpgrade for VersionUncheckedMigrateV5ToV6<T> {
```

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -227,6 +227,26 @@ pub trait OnRuntimeUpgrade {
 	}
 }
 
+/// See [`Hooks::on_runtime_upgrade`].
```
liamaharon, line 230: Some better docs here would be nice, and also updating the docs here https://paritytech.github.io/polkadot-sdk/master/frame/traits/trait.Hooks.html#implementation-note-standalone-migrations

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -227,6 +227,26 @@ pub trait OnRuntimeUpgrade {
...
+pub trait UncheckedOnRuntimeUpgrade {
+	/// See [`Hooks::on_runtime_upgrade`].
+	fn unchecked_on_runtime_upgrade() -> Weight {
```
liamaharon, line 233: I *slightly* prefer this just being 

```suggestion
	fn on_runtime_upgrade() -> Weight {
```

Since it's already clear from the trait name that it's unchecked.

prdoc/pr_3835.prdoc
```diff
@@ -0,0 +1,24 @@
...
+
+title: "migrations: prevent accidentally using inner unversioned migration instead of `VersionedMigration`"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Currently, it is possible to accidentally use inner unversioned migration instead of `VersionedMigration` 
+      since both implement `OnRuntimeUpgrade`. With this change, we make it clear that `Inner` is not intended 
+      to be used directly. It is achieved by bounding `Inner` to new trait `UncheckedOnRuntimeUpgrade`, which 
+      has the same interface as `OnRuntimeUpgrade`, but can not be used directly for runtime upgrade migrations.
```
liamaharon, line 12: Can we write a short "Upgrade Guide" here for devs, explaining the steps to update their code to accomodate these changes?

prdoc/pr_3835.prdoc
```diff
@@ -0,0 +1,24 @@
...
+
+crates:
+  - name: "pallet-example-single-block-migrations"
```
liamaharon, line 15: Need to specify a `bump` for these crates, e.g.

```suggestion
crates:
  - name: "pallet-example-single-block-migrations"
    bump: "major"
```

I think most of these will be major, since the public APIs have breaking changes.

liamaharon (commented):
Awesome, thanks very much for this! Just some minor comments.

acatangiu (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5706519

liamaharon (approved):

liamaharon (changes requested):
Ah actually could you please make one more change, to the `single-block-migration` example?

We no longer need to wrap the Migration in a `mod version_unchecked`.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5730616

ggwpez (approved):

liamaharon:
Here too please @dastansam https://github.com/paritytech/polkadot-sdk/blob/5eff3f94beb768a7ba56d9091f6137fa6dff6bb1/substrate/frame/examples/single-block-migrations/src/migrations/v1.rs#L37-L48

liamaharon (approved):

PR #3850, opened 2024-03-27
pallet-referenda: Detect incorrect pre-image length

bkchr:
There has been a case that a referenda failed because the length given to `submit` was incorrect. The pallet can actually check the length if the pre-image already exists to ensure that these kind of issues are not happening again.


liamaharon (approved):

joepetrowski (approved):

substrate/frame/referenda/src/lib.rs
```diff
@@ -462,6 +464,18 @@ pub mod pallet {
 			let proposal_origin = *proposal_origin;
 			let who = T::SubmitOrigin::ensure_origin(origin, &proposal_origin)?;
 
+			// If the pre-image is already stored, ensure that it has the same length as given in
+			// `proposal`.
+			if let Some(hash) = proposal.lookup_hash() {
+				if let Some((l, r)) =
```
franciscoaguirre, line 470:
```suggestion
				if let Some((preimage_length, proposal_length)) =
```

😅

ggwpez (approved):

substrate/frame/referenda/src/lib.rs
kianenigma:
```
			ensure!(
				proposal.lookup_hash()
					.and_then(|hash| T::Preimages::len(&hash))
					.map(|hash_len| hash_len != proposal.lookup_len().unwrap_or(u32::MAX))
					.unwrap_or(true),
				Error::<T, I>::PreimageStoredWithDifferentLength
			)
```

just playing some `Option` API games :D

PR #3852, opened 2024-03-27
Prevent accidental change of network-key for active authorities 

alexggh:
As discovered during investigation of https://github.com/paritytech/polkadot-sdk/issues/3314 and https://github.com/paritytech/polkadot-sdk/issues/3673 there are active validators which accidentally might change their network key during restart, that's not a safe operation when you are in the active set because of distributed nature of DHT, so the old records would still exist in the network until they expire 36h, so unless they have a good reason validators should avoid changing their key when they restart their nodes. 

There is an effort in parallel to improve this situation https://github.com/paritytech/polkadot-sdk/pull/3786, but those changes are way more intrusive and will need more rigorous testing, additionally they will reduce the time to less than 36h, but the propagation won't be instant anyway, so not changing your network during restart should be the safest way to run your node, unless you have a really good reason to change it.

## Proposal
1. Do not auto-generate the network if the network file does not exist in the provided path. Nodes where the key file does not exist will get the following error:
```
Error: 
   0: Starting an authorithy without network key in /home/alexggh/.local/share/polkadot/chains/ksmcc3/network/secret_ed25519.
      
       This is not a safe operation because the old identity still lives in the dht for 36 hours.
      
       Because of it your node might suffer from not being properly connected to other nodes for validation purposes.
      
       If it is the first time running your node you could use one of the following methods.
      
       1. Pass --unsafe-force-node-key-generation and make sure you remove it for subsequent node restarts
      
       2. Separetly generate the key with: polkadot key generate-node-key --file <YOUR_PATH_TO_NODE_KEY>
```

2. Add an explicit parameters for nodes that do want to change their network despite the warnings or if they run the node for the first time. `--unsafe-force-node-key-generation` 

3. For `polkadot key generate-node-key` add two new mutually exclusive parameters `base_path` and `default_base_path` to help with the key generation in the same path the polkadot main command would expect it.
 
4. Modify the installation scripts to auto-generate a key in default path if one was not present already there, this should help with making the executable work out of the box after an instalation.

## Notes

Nodes that do not have already the key persisted will fail to start after this change, however I do consider that better than the current situation where they start but they silently hide that they might not be properly connected to their peers.

## TODO
- [x]  Make sure only nodes that are authorities on producation chains will be affected by this restrictions.
- [x] Proper PRDOC, to make sure node operators are aware this is coming.


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5670995

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5676836

substrate/client/cli/src/params/node_key_params.rs
```diff
@@ -79,22 +80,47 @@ pub struct NodeKeyParams {
...
+	///
+	/// This is an unsafe feature for producation networks because the previous
+	/// node address would still live in the DHT for 36h and the node would suffer
+	/// from bad connectivity to other nodes untill the old record expires.
+	/// For minimal node downtime if no custom `node-key-file` argument is provided
+	/// the network-key is usually persisted accross nodes restarts,
+	/// in the `network` folder from directory provided in `--base-path`
+	///
+	/// This is safe to be used the first time you run your node or if you already
+	/// lost your node keys or if you don't care about the above problem.
```
dmitry-markin, line 94: I wouldn't add a comment that it's safe, even under specific circumstances, — people might use it and forget they passed this argument.

alexggh:
> As it makes the life of first-time node starters harder, could we include the node key generation logic in the post-install pipeline of the .deb package and Docker image?

Done, although, I'm still concerned if it is wise to generate automatically keys at install, I've seen pipelines where part of producing your docker image for deployment you install all the required tooling, so in this case when operators would produce their v2 of the image they might get on the filesystem and if they use the same image to run multiple nodes they might get into trouble with using the same key.

substrate/client/cli/src/commands/generate_node_key.rs
```diff
@@ -58,9 +69,32 @@ impl GenerateNodeKeyCmd {
...
-			Some(file) => fs::write(file, file_data)?,
-			None => io::stdout().lock().write_all(&file_data)?,
+		match (&self.file, &self.base_path, self.default_base_path) {
+			(Some(file), None, false) => fs::write(file, file_data)?,
+			(None, Some(_), false) | (None, None, true) => {
+				let network_path = self
+					.base_path
+					.clone()
+					.unwrap_or_else(|| {
+						BasePath::from_project("", "", &C::executable_name()).path().to_path_buf()
```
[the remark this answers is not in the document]
alexggh: Yes, it makes sense.

substrate/client/cli/src/commands/generate_node_key.rs
```diff
@@ -58,9 +69,32 @@ impl GenerateNodeKeyCmd {
...
+				let key_path = network_path.join(NODE_KEY_ED25519_FILE);
+				if key_path.exists() {
+					eprintln!("Skip generation, a key already exists in {:?}", key_path);
+					return Ok(());
+				} else {
+					eprintln!("Generating key in {:?}", key_path);
+					fs::write(key_path, file_data)?
+				}
+			},
+			(None, None, false) => io::stdout().lock().write_all(&file_data)?,
```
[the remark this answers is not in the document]
alexggh: This was already there, I did not want to modify it since, I expect people might use this output to pass it as the --node-key arguments.

substrate/client/cli/src/params/node_key_params.rs
alexggh: It does says you need to remove it, I just wanted people to be able to bootstrap the first time they run a node without much inconvenience.

substrate/client/cli/src/params/node_key_params.rs
```diff
@@ -79,22 +80,47 @@ pub struct NodeKeyParams {
 	/// the chosen type.
 	#[arg(long, value_name = "FILE")]
```
dmitry-markin, line 81: While you are at it, may be make `--node-key` & `--node-key-file` conflicting arguments?

substrate/client/cli/src/commands/generate_node_key.rs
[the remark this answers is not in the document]
dmitry-markin: Sorry, missed it.

substrate/client/cli/src/params/node_key_params.rs
dmitry-markin: IMO it's better to recommend a user to use an explicit command, than something that can be misused.

dmitry-markin (approved):

polkadot/.rpm/polkadot.spec
```diff
@@ -37,6 +37,7 @@ getent passwd polkadot >/dev/null || \
 if [ ! -e "$config_file" ]; then
     echo 'POLKADOT_CLI_ARGS=""' > /etc/default/polkadot
 fi
+polkadot key generate-node-key --default-base-path
```
s0me0ne-unkn0wn, line 40:
```suggestion
if [ $1 -eq 1 ]; then
    polkadot key generate-node-key --default-base-path
fi
```

Only do that when installing, not when upgrading the package

polkadot/scripts/packaging/deb-maintainer-scripts/postinst
```diff
@@ -13,5 +13,6 @@ if [ "$action" = "configure" ]; then
     --ingroup polkadot polkadot
   if [ ! -e "$config_file" ]; then
     echo 'POLKADOT_CLI_ARGS=""' > /etc/default/polkadot
+    /usr/bin/polkadot key generate-node-key --default-base-path
   fi
```
s0me0ne-unkn0wn, line 17:
```suggestion
  if [ ! -e "$config_file" ]; then
    echo 'POLKADOT_CLI_ARGS=""' > /etc/default/polkadot
  fi
  if [ -z "$2" ]; then
    /usr/bin/polkadot key generate-node-key --default-base-path
  fi
```

The same. Both untested but I hope the CI will do that for us.

s0me0ne-unkn0wn (changes requested):

s0me0ne-unkn0wn (approved):

lexnv (approved):

alexggh:
@paritytech/release-engineering could you help me test the generated Deb and Rpm files after this PR. Thank you!

substrate/client/cli/src/params/node_key_params.rs
alexggh: I ended-up keeping this as well.

polkadot/scripts/packaging/polkadot.service
```diff
@@ -5,6 +5,7 @@ Documentation=https://github.com/paritytech/polkadot
 
 [Service]
 EnvironmentFile=-/etc/default/polkadot
+ExecStartPre=/usr/bin/polkadot key generate-node-key --default-base-path
```
s0me0ne-unkn0wn, line 8: It is executed on every restart. If the key is already there, wouldn't it re-generate the key on every restart then?

substrate/client/cli/src/params/node_key_params.rs
dmitry-markin: It doesn't seem right to me to start a phrase with words `This is safe ...` for an argument that we explicitly marked as `--unsafe-...`. If we want to discourage people using it, let's be consistent. Especially as we have a safe one-off alternative. First of all, the purpose of this PR is the opposite to running the node with this argument.

alexggh: Will update the comment you are right.

polkadot/scripts/packaging/polkadot.service
s0me0ne-unkn0wn: Okay, I'm not sure I follow then. You execute the key generation procedure on every node restart, and it generates a new key if it's not there already. How is it different from the current behavior we're trying to avoid?

s0me0ne-unkn0wn:
> In the end I went for the option where we generate the key in the systemd ExecStartPre, I definitely did not feel comfortable generating a key at installation time way too many things that could go wrong and end up with the same key on two different machines.

That looks totally sound to me. Many packages generate something on the fly during the installation. The point is that we want to do it just a single time, not if the key is just not there. If we want to regenerate in case the key is not there, we can just keep the current behavior, as it is exactly how it works right now :slightly_smiling_face:

polkadot/scripts/packaging/polkadot.service
alexggh: Yeah, I think you are right here I'm creating the same problem I'm trying to solve, good catch!

alexggh:
> > In the end I went for the option where we generate the key in the systemd ExecStartPre, I definitely did not feel comfortable generating a key at installation time way too many things that could go wrong and end up with the same key on two different machines.
> 
> That looks totally sound to me. Many packages generate something on the fly during the installation. The point is that we want to do it just a single time, not if the key is just not there. If we want to regenerate in case the key is not there, we can just keep the current behavior, as it is exactly how it works right now 🙂


Yeah, you are right, `ExecStartPre` is not the right place for key generation, I'm starting to think that the safest way would be to actually not generate automatically the key at installation either and just let our users explicitly decide when they want the key to be generated.

s0me0ne-unkn0wn:
> I'm starting to think that the safest way would be to actually not generate automatically the key at installation either and just let our users explicitly decide when they want the key to be generated.

Maybe, and anyway, that's the only way the node operators who deploy from raw binaries can do it. But when installing a package, you usually expect to get things set up and configured, that's the purpose of packages. What tells against this approach?

alexggh:
> > I'm starting to think that the safest way would be to actually not generate automatically the key at installation either and just let our users explicitly decide when they want the key to be generated.
> 
> Maybe, and anyway, that's the only way the node operators who deploy from raw binaries can do it. But when installing a package, you usually expect to get things set up and configured, that's the purpose of packages. What tells against this approach?

In the cloud era after instalation images get snapshoted in one form or the other and you might end up with the same peer-id on two different machines.
E.g:
1. You install the package as part of building your container image, the if you run the same docker image with two different kubernetes pods you end up with two nodes using the same key.
2. You install the package on an Amazon AMI, you snapshot it, then you spawn two different instances from it and run the polkadot executable, same thing you end up with two nodes using the same key.

All this failure modes, does not exist in current implementation because the key is generated at first run of the polkadot, but that has other failure modes which is what I'm trying to fix with this MR :D.

I agree, it is not ideal that after package install you might still need to do some explicit steps, but I still find that better than the alternative where you silently run your node with the wrong network-key.

substrate/client/cli/src/error.rs
```diff
@@ -78,6 +80,16 @@ pub enum Error {
...
 	#[error(transparent)]
 	GlobalLoggerError(#[from] sc_tracing::logging::Error),
+
+	#[error(
+		"Starting an authorithy without network key in {0}.
+		\n This is not a safe operation because the old identity still lives in the dht for 36 hours.
+		\n Because of it your node might suffer from not being properly connected to other nodes for validation purposes.
+		\n If it is the first time running your node you could use one of the following methods.
+		\n 1. Pass --unsafe-force-node-key-generation and make sure you remove it for subsequent node restarts
+		\n 2. Separately generate the key with: polkadot key generate-node-key --file <YOUR_PATH_TO_NODE_KEY>"
```
dmitry-markin, line 90: May be instead of `--file` argument recommend using `--base-path` & `--default-base-path`? This seems to be less error-prone and won't require first time users to know the exact network key location.

dmitry-markin (approved):

substrate/client/cli/src/error.rs
alexggh: Should be better now!

substrate/client/cli/src/params/node_key_params.rs
alexggh: Should be better now.

paritytech-cicd-pr:
**Job name**: build-runtime-starters
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5695702

substrate/client/cli/src/params/node_key_params.rs
dmitry-markin: Looks good, thanks!

substrate/client/cli/src/error.rs
dmitry-markin: Thanks!

substrate/client/cli/src/commands/generate_node_key.rs
```diff
@@ -36,18 +37,33 @@ use std::{
...
+	/// Save the key in the default directory. If a key already
+	/// exists in the directory, it won't be overwritten.
+	#[arg(long, conflicts_with_all = ["base_path", "file"])]
+	default_base_path: bool,
 }
 
 impl GenerateNodeKeyCmd {
 	/// Run the command
-	pub fn run(&self) -> Result<(), Error> {
+	pub fn run<C: SubstrateCli>(&self, cli: &C) -> Result<(), Error> {
```
bkchr, line 66:
```suggestion
	pub fn run(&self, chain_spec_id: Option<String>, executable_name: &str) -> Result<(), Error> {
```

substrate/client/cli/src/commands/generate_node_key.rs
```diff
@@ -36,18 +37,33 @@ use std::{
...
+	pub chain: Option<String>,
+	/// A directory where the key should be saved. If a key already
+	/// exists in the directory, it won't be overwritten.
+	#[arg(long, conflicts_with_all = ["file", "default_base_path"])]
+	base_path: Option<PathBuf>,
+
+	/// Save the key in the default directory. If a key already
+	/// exists in the directory, it won't be overwritten.
+	#[arg(long, conflicts_with_all = ["base_path", "file"])]
+	default_base_path: bool,
```
bkchr, line 61: Maybe we should just remove this and `chain` completely. IMO, if you run a validator you should not use the default base path and even if you use it, you could still use the `unsafe-force-generate-node-key` CLI command.

substrate/client/cli/src/commands/generate_node_key.rs
```diff
@@ -58,9 +74,33 @@ impl GenerateNodeKeyCmd {
...
+					&chain_spec,
+					cli,
+				);
+
+				fs::create_dir_all(network_path.as_path())?;
+
+				let key_path = network_path.join(NODE_KEY_ED25519_FILE);
+				if key_path.exists() {
+					eprintln!("Skip generation, a key already exists in {:?}", key_path);
+					return Ok(());
```
bkchr, line 93: IMO should return an error to let the program return an error to make people "aware".

substrate/client/cli/src/error.rs
```diff
@@ -78,6 +80,19 @@ pub enum Error {
...
 	GlobalLoggerError(#[from] sc_tracing::logging::Error),
+
+	#[error(
+		"Starting an authorithy without network key in {0}.
+		\n This is not a safe operation because the old identity still lives in the dht for 36 hours.
+		\n Because of it your node might suffer from not being properly connected to other nodes for validation purposes.
+		\n If it is the first time running your node you could use one of the following methods:
+		\n 1. [Preferred] Separately generate the key with: polkadot key generate-node-key --default-base-path
+		\n 2. [Preferred] Separately generate the key with: polkadot key generate-node-key --base-path <YOUR_BASE_PATH>
+		\n 3. [Preferred] Separately generate the key with: polkadot key generate-node-key --file <YOUR_PATH_TO_NODE_KEY>
```
bkchr, line 91:
```suggestion
		\n 1. [Preferred] Separately generate the key with: node key generate-node-key --default-base-path
		\n 2. [Preferred] Separately generate the key with: node key generate-node-key --base-path <YOUR_BASE_PATH>
		\n 3. [Preferred] Separately generate the key with: node key generate-node-key --file <YOUR_PATH_TO_NODE_KEY>
```

This is not only used by `polkadot`.

substrate/client/cli/src/error.rs
```diff
@@ -78,6 +80,19 @@ pub enum Error {
 
 	#[error(transparent)]
 	GlobalLoggerError(#[from] sc_tracing::logging::Error),
+
+	#[error(
+		"Starting an authorithy without network key in {0}.
+		\n This is not a safe operation because the old identity still lives in the dht for 36 hours.
+		\n Because of it your node might suffer from not being properly connected to other nodes for validation purposes.
```
bkchr, line 87:
```suggestion
		\n This is not a safe operation because other authorities in the network may depend on your node having a stable identity.
		\n Otherwise these other authorities may not being able to reach you.
```

substrate/client/cli/src/params/node_key_params.rs
```diff
@@ -79,22 +80,47 @@ pub struct NodeKeyParams {
...
 	pub node_key_file: Option<PathBuf>,
+
+	/// Forces key generation if node-key-file file does not exist.
+	///
+	/// This is an unsafe feature for producation networks because the previous
+	/// node address would still live in the DHT for 36h and the node would suffer
+	/// from bad connectivity to other nodes untill the old record expires.
```
bkchr, line 88:
```suggestion
```

Would not mention the DHT, because this is not being true for collators for example. Generally the doc could be improved a little bit.

substrate/client/cli/src/commands/generate_node_key.rs
alexggh: Care to develop here? Why shouldn't a default validator use the default-base-path what issue can arise from that  ? I thought most  of them actually do it

substrate/client/cli/src/commands/generate_node_key.rs
alexggh: Done.

substrate/client/cli/src/params/node_key_params.rs
alexggh: >Generally the doc could be improved a little bit

Improved it a little bit, let me know if it is good now.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5773063

substrate/client/cli/src/commands/generate_node_key.rs
bkchr: IDK :P These paths are "magic" and you will first need to do some digging to find them. Especially if you want to mount this folder in docker or whatever. I would personally find it easier to pass some explicit path, but maybe I'm weird xD

Nevertheless, if you want to keep this, we should introduce a new `GenerateNodeKeyCmd` that should be used for "nodes". If you use the dedicated `subkey` binary it doesn't make sense to have the `chain` parameter etc. Does this makes sense?

alexggh: Yes, I agree `chain` parameter doesn't make sense for the subkey binary, however it needs to be present for `node` because even when you run polkadot with a specific `base-path` we append to it the chain_id to obtain the configuration directory.

Thinking of  going the route of introducing a new GenerateNodeKeyCmd that should be used for nodes, would that work for you ?

alexggh: Done, let me know if it is better now.

substrate/client/cli/src/error.rs
```diff
@@ -78,6 +80,21 @@ pub enum Error {
...
+	#[error(
+		"Starting an authorithy without network key in {0}.
+		\n This is not a safe operation because other authorities in the network may depend on your node having a stable identity.
+		\n Otherwise these other authorities may not being able to reach you.
+		\n If it is the first time running your node you could use one of the following methods:
+		\n 1. [Preferred] Separately generate the key with: <NODE_BINARY> key generate-node-key --base-path <YOUR_BASE_PATH>
+		\n 2. [Preferred] Separately generate the key with: <NODE_BINARY> key generate-node-key --file <YOUR_PATH_TO_NODE_KEY>
+		\n 3. [Preferred] Separately generate the key with: <NODE_BINARY> key generate-node-key --default-base-path
+		\n 4. [Unsafe] Pass --unsafe-force-node-key-generation and make sure you remove it for subsequent node restarts"
+
```
bkchr, line 93:
```suggestion
```

bkchr (approved):
Just misses a prdoc for `Node operators`, otherwise looks good.

Ty!

alexggh:
> Just misses a prdoc for `Node operators`, otherwise looks good.
> 
> Ty!

Done, merging ...

Issue #3748, opened 2024-03-19
Add subsystem benchmark for `statement-distribution`


sandreim:
As we scale up the number of validators and cores, having this support would provide an easy way to optimize the subsystem's performance and also a means to profile/debug issues like https://github.com/paritytech/polkadot-sdk/issues/3744

PR #3863, opened 2024-03-27
[subsystem-benchmarks] Add statement-distribution benchmarks

AndreiEres:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3748

Adds a subsystem benchmark for statements-distribution subsystem.

Results in CI (reference hw):
```
$ cargo bench -p polkadot-statement-distribution --bench statement-distribution-regression-bench --features subsystem-benchmarks

[Sent to peers] standart_deviation 0.07%
[Received from peers] standart_deviation 0.00%
[statement-distribution] standart_deviation 0.97%
[test-environment] standart_deviation 1.03%

Network usage, KiB                     total   per block
Received from peers                1088.0000    108.8000
Sent to peers                      1238.1800    123.8180

CPU usage, seconds                     total   per block
statement-distribution                0.3897      0.0390
test-environment                      0.4715      0.0472
```


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5892414

alvicsam (approved):

polkadot/node/network/statement-distribution/benches/statement-distribution-regression-bench.rs
```diff
@@ -0,0 +1,78 @@
...
+	.map_err(|e| e.to_string())?;
+	println!("{}", average_usage);
+
+	// We expect no variance for received and sent
+	// but use 0.001 because we operate with floats
+	messages.extend(average_usage.check_network_usage(&[
+		("Received from peers", 108.8000, 0.001),
+		("Sent to peers", 123.8180, 0.001),
+	]));
+	messages.extend(average_usage.check_cpu_usage(&[("statement-distribution", 0.0390, 0.1)]));
```
alexggh, line 70: So, unless we are missing something here, can we concur that statement-distribution is not a subsystem with high resource(CPU + networking) demands ?

polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -728,8 +746,41 @@ pub struct NetworkEmulatorHandle {
...
+				AllMessages::StatementDistribution(
+					StatementDistributionMessage::NetworkBridgeUpdate(network),
+				)
+			})
+			.collect_vec()
+	}
+
 	/// Generates peer_connected messages for all peers in `test_authorities`
-	pub fn generate_peer_connected(&self) -> Vec<AllMessages> {
+	pub fn generate_approval_distribution_peer_connected(&self) -> Vec<AllMessages> {
```
alexggh, line 783: Nit: To not repeat these 2 functions you can name it something like

generate_subsystem_peer_connected(&self, lambda_to_transform_peer_connected_subsystem_specific_message)

and then call the lambda function in the map.

polkadot/node/subsystem-bench/src/lib/mock/candidate_backing.rs
```diff
@@ -0,0 +1,157 @@
...
+use polkadot_node_subsystem_types::OverseerSignal;
+use polkadot_primitives::{
+	CandidateHash, PersistedValidationData, SigningContext, ValidatorIndex, ValidatorPair,
+};
+use sp_core::Pair;
+use std::collections::HashMap;
+
+const LOG_TARGET: &str = "subsystem-bench::candidate-backing-mock";
+
+pub struct MockCandidateBacking {
```
alexggh, line 34: Nit: I would go with  keeping the current convention where clearly define a CandidateBackingState structure, that gets included here.

polkadot/node/subsystem-bench/src/lib/statement/mod.rs
```diff
@@ -0,0 +1,445 @@
...
+			})
+			.collect_vec();
+
+		for (seconding_peer_id, group_index) in one_hop_peers_and_groups
+			.into_iter()
+			.chain(two_hop_x_peers_and_groups)
+			.chain(two_hop_y_peers_and_groups)
+		{
+			let messages_sent_count = messages_tracker.get_mut(group_index).unwrap();
+			if *messages_sent_count {
```
alexggh, line 380: I'm not too sure about this, my understanding is that a node could receive a statement multiple times from different nodes in the group, so I think we should let this work.

alexggh (approved):
Overall, looks good to me, good job!

Left you some nit comments.

polkadot/node/network/statement-distribution/benches/statement-distribution-regression-bench.rs
AndreiEres: At least now we know about it

alexggh: > Yeah, it feels so, at least the signature checks seem damn fast compared to approval vote signatures :)

It is actually way less signature checks. For each candidate only the backing group would generate a statement so for polkadot you get 5 signature per candidate and with 100 candidates you check just 500 signatures in total.

In approval voting you need at least 30 votes(in practice more than that). per candidate so 30 signature per candidate and with 100 candidates you actually have to check 100 * 30 = 3000 signatures, that's around 6 times more signatures.

> At least now we know about it

Yes, it was meant as a good news :D.

polkadot/node/subsystem-bench/src/lib/mock/runtime_api.rs
```diff
@@ -223,6 +245,43 @@ impl MockRuntimeApi {
 								.clone()
 								.expect("Babe epoch unpopulated")));
 						},
+						RuntimeApiMessage::Request(
+							_block_hash,
+							RuntimeApiRequest::AsyncBackingParams(sender),
+						) => {
+							let _ = sender.send(Ok(AsyncBackingParams {
```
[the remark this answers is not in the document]
AndreiEres: Yeah, but better to set the production ones

polkadot/node/subsystem-bench/src/lib/mock/runtime_api.rs
```diff
@@ -223,6 +245,43 @@ impl MockRuntimeApi {
...
+								max_candidate_depth: 3,
+								allowed_ancestry_len: 2,
+							}));
+						},
+						RuntimeApiMessage::Request(_parent, RuntimeApiRequest::Version(tx)) => {
+							tx.send(Ok(RuntimeApiRequest::DISABLED_VALIDATORS_RUNTIME_REQUIREMENT))
```
alindima, line 258: This needs to return the latest version, right?

AndreiEres: At least a version that supports `DisabledValidators`

sandreim (approved):
Nice work! Thank you @AndreiEres

polkadot/node/subsystem-bench/src/lib/statement/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+				.values()
+				.filter(|v| v.load(Ordering::SeqCst))
+				.collect::<Vec<_>>()
+				.len();
+			gum::debug!(target: LOG_TARGET, "{}/{} manifest exchanges", manifests_count, candidates_advertised);
+
+			if manifests_count == candidates_advertised {
+				break;
+			}
+			tokio::time::sleep(Duration::from_millis(50)).await;
```
alindima, line 437: If I'm reading the code correctly, the manifests we are advertising already contain statements from all members of the backing group.
Therefore the only gossip that is emulated is the manifest/acknowledgement exchange. The statement gossip is not really stressing the subsystem.
In reality every backing group could continue sending Statement messages even after the backed candidate is fetched. (and we care to record these so that they can be punished in case of a dispute).
Maybe a manifest is advertised with only 2 backing votes but the backing group keeps issuing more votes even after that.

However, the code is already quite complicated and we can see that the CPU consumption of the subsystem is tiny, so I don't think implementing this will yield any benefit right now

alindima (approved):
Left an idea but overall looks good, nice work!

polkadot/node/subsystem-bench/src/lib/statement/mod.rs
sandreim: Thanks @alindima ! A ticket describing this improvement should do justice!

AndreiEres: Decided with @alindima, that in this specific case it doesn't make a big difference

PR #3865, opened 2024-03-27
[balances] Safeguard against consumer ref underflow

ggwpez:
There are some accounts that do not have a consumer ref while having a reserve.  
This adds a fail-safe mechanism to trigger in the case that `does_consume` is true, but the assumption of `consumer>0` is not.  

This should prevent those accounts from loosing balance and the TI from getting messed up even more, but is not an "ideal" fix. TBH an ideal fix is not possible, since on-chain data is in an invalid state.


substrate/frame/balances/src/tests/general_tests.rs
```diff
@@ -0,0 +1,111 @@
...
+			Balances::transfer_allow_death(Some(alice).into(), bob, 90),
+			DispatchError::ConsumerRemaining
+		);
+		assert_noop!(
+			Balances::transfer_all(Some(alice).into(), bob, false),
+			DispatchError::ConsumerRemaining
+		);
+
+		// normal transfers still work:
+		hypothetically!({
```
kianenigma, line 81: oh this is a great macro, I always manually used a transactional layer!

substrate/frame/balances/src/tests/mod.rs
```diff
@@ -278,6 +279,23 @@ pub fn info_from_weight(w: Weight) -> DispatchInfo {
 	DispatchInfo { weight: w, ..Default::default() }
 }
 
+/// Check that the total-issuance matches the sum of all accounts' total balances.
+pub fn ensure_ti_valid() {
```
kianenigma, line 283: Can we easily inject this at the end of all tests in this crate?

kianenigma (approved):

substrate/frame/balances/src/tests/mod.rs
ggwpez: Ah yes. I though about adding a `fn register_cleanup_hook` to the `TestExternalities`.  
Then we could set this in the externalities as hook and have it run at the end without any refactoring. Otherwise i think we need to refactor to not directly use the `execute_with` but wrap it again.

liamaharon: I would be curious to also know if it's feasible as a try-state hook

liamaharon (approved):

prdoc/pr_3865.prdoc
```diff
@@ -0,0 +1,11 @@
...
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Pallet balances now handles the case that historic accounts violate a invariant that they should have a consumer ref on `reserved > 0` balance.  
+      This disallows such accounts from reaping and should prevent TI from getting messed up even more.
+
+crates:
+  - name: pallet-balances
+    bump: minor
```
liamaharon, line 11: Not a `patch`?

ggwpez, line 11:
```suggestion
    bump: patch
```

ggwpez: Thats a good start, thought i wrote patch 😅

substrate/frame/balances/src/tests/mod.rs
kianenigma: > Otherwise i think we need to refactor to not directly use the execute_with but wrap it again.

Yeah this is what most pallets do, and I generally find it easier. 

> I would be curious to also know if it's feasible as a try-state hook

Ideally it should be, but yeah iterating all accounts will ruin everything else 🙈 We need to think of a system to separate try-state hooks that we always run in a place like CI vs. those that we want to run every month etc.

liamaharon: I would be really interested to bench it. We have a lot of really heavy staking hooks today without much issue, maybe as long as it's O(n) it's OK.

kianenigma: looking forward to adding it :)

ggwpez: I ran these checks in remote-externalities to fine some issues with this: https://github.com/ggwpez/wtfwt  
takes like 3-5 secs for one run on Polkadot.

liamaharon: 3-5s is acceptable imo

PR #3866, opened 2024-03-27
Clear ledger instead of recovering (bonded, no ledger or lock)

gpestana:


substrate/frame/staking/src/tests.rs
```diff
@@ -7563,18 +7563,19 @@ mod ledger_recovery {
...
+			//);
+
+			//assert_ok!(Staking::restore_ledger(
+			//	RuntimeOrigin::root(),
+			//	444,
+			//	None,
+			//	Some(total_444_before_corruption),
+			//	None,
+			//));
+			assert_ok!(Staking::restore_ledger(RuntimeOrigin::root(), 444, None, None, None),);
```
gpestana, line 7578: The problem with this is that this call will remove unbond 444 completely. However, you can see in the comments above and the test case that at no place the stash 444 wanted to unbond (it got to this state by a "kill(333)" by proxy).

So in sum, we should aim at:
- 333 killed and unbonded completely
- 444 recovered with a state pre-corruption


wdyt?

Ank4n: Yes this is true. In my [approach](https://github.com/paritytech/polkadot-sdk/pull/3868), I would just clear out both storage and locks since on-chain connection between 333 and 444 is broken. We could make a modification though where we squash these two together so it works as in your case but then we trust staking admin on this connection.

kianenigma:
Will this be needed, or we settled to keep the existing fix?

kianenigma:
ping?

PR #3873, opened 2024-03-28
Decorate mpsc-notification-to-protocol with the protocol name

alexggh:
Currently, all protocols use the same metric name for `mpsc-notification-to-protocol` this is bad because we can't actually tell which protocol might cause problems.

This patch proposes we derive the name of the metric from the protocol name, so that we have separate metrics for each protocol and properly detect which one is having problem processing its messages.


substrate/client/utils/src/mpsc.rs
```diff
@@ -55,7 +55,7 @@ impl<T> Clone for TracingUnboundedSender<T> {
 	fn clone(&self) -> Self {
 		Self {
 			inner: self.inner.clone(),
-			name: self.name,
+			name: self.name.clone(),
```
acatangiu, line 58: from a quick search it looks like `TracingUnboundedSender` isn't being cloned anywhere in the hotpath, but leaving this here in case someone knows different

acatangiu (approved):

substrate/client/network/src/protocol/notifications/service/mod.rs
```diff
@@ -624,11 +626,24 @@ pub fn notification_service(
...
 
 	(
 		ProtocolHandlePair::new(protocol.clone(), subscribers.clone(), cmd_rx),
 		Box::new(NotificationHandle::new(protocol.clone(), cmd_tx, event_rx, subscribers)),
 	)
 }
+
+// Decorates the mpsc-notification-to-protocol metric with the name of the protocol,
+// to be able to distiguish between different protocols in dashboards.
+fn metric_name_for_protocol(protocol: &ProtocolName) -> String {
```
sandreim, line 642:
```suggestion
fn metric_label_for_protocol(protocol: &ProtocolName) -> String {
```

substrate/client/utils/src/mpsc.rs
```diff
@@ -68,25 +68,25 @@ impl<T> Clone for TracingUnboundedSender<T> {
...
 	inner: Receiver<T>,
-	name: &'static str,
+	name: String,
 }
 
 /// Wrapper around [`async_channel::unbounded`] that tracks the in- and outflow via
 /// `UNBOUNDED_CHANNELS_COUNTER` and warns if the message queue grows
 /// above the warning threshold.
 pub fn tracing_unbounded<T>(
-	name: &'static str,
```
sandreim, line 78: Since 1.72 you can use `String::leak()` https://doc.rust-lang.org/nightly/alloc/string/struct.String.html#method.leak

substrate/client/utils/src/mpsc.rs
alexggh: No more cloning.

sandreim (approved):

PR #3874, opened 2024-03-28
bugfix: request fragment tree membership for all candidates

alindima:


polkadot/node/core/backing/src/lib.rs
```diff
@@ -911,7 +911,7 @@ async fn handle_active_leaves_update<Context>(
 			}
 
 			let mut seconded_at_depth = HashMap::new();
-			if let Some(response) = membership_answers.next().await {
+			while let Some(response) = membership_answers.next().await {
```
alexggh, line 914: Good catch! Any consequences for the current production chains ?

alexggh (approved):

polkadot/node/core/backing/src/lib.rs
alindima: AFAICT, the only consequence of this bug is that when getting a new leaf, this validator may already have a valid seconded candidate to propose for backing, but it will not use it. It will accept a new collation. Collator protocol does not guard against this because it ends up calling the backing subsystem for this info.

So the only potential impact is that we would accept more collations than neccessary. The DOS potential is quite low considering that the chance of the previous candidate remains valid at this leaf is quite low (because allowed_ancestry_len will be 2).

s0me0ne-unkn0wn (approved):

Issue #2977, opened 2024-01-18
doc: mermaid diagram not rendering


liamaharon:
Mermaid diagram is not rendering on this page, I can't spot any immediate issue https://paritytech.github.io/polkadot-sdk/master/frame/traits/trait.Hooks.html#summary

It renders fine on other pages like https://paritytech.github.io/polkadot-sdk/master/pallet_bags_list/example/index.html

@kianenigma any ideas?

PR #3875, opened 2024-03-28
Fix Mermaid diagram rendering

liamaharon:
Closes https://github.com/paritytech/polkadot-sdk/issues/2977

The issue appears to stem from the `aquamarine` crate failing to render diagrams in re-exported crates.

e.g. as raised [here](https://github.com/paritytech/polkadot-sdk/issues/2977), diagrams would render at `frame_support::traits::Hooks` but not the re-exported doc `frame::traits::Hooks`, even if I added `aquamarine` as a `frame` crate dependency.

To resolve this, I followed advice in https://github.com/mersinvald/aquamarine/issues/20 to instead render mermaid diagrams directly using JS by adding an `after-content.js`.

---

Also fixes compile warnings, enables `--all-features` and disallows future warnings in CI.


alvicsam (approved):

substrate/frame/support/src/traits/hooks.rs
```diff
@@ -318,7 +318,6 @@ pub trait IntegrityTest {
 	fn integrity_test() {}
 }
 
-#[cfg_attr(doc, aquamarine::aquamarine)]
```
kianenigma, line 321: Nice, can we also remove `simple_mermaid` as well, if it is still used anywhere?

kianenigma (approved):

substrate/frame/support/src/traits/hooks.rs
liamaharon: We're still using `simple_mermaid` in a few places for embedding `.mmd` files, so I think need to keep it

bkchr:
But this will still fail on docs.rs?

We should think on how to fix this there as well. Setting RUSTDOCFLAGS is also possible: https://github.com/rust-lang/docs.rs/blob/8b19841b449244d79976c29ae24534af3d7e8c82/templates/core/Cargo.toml.example#L42

liamaharon:
> But this will still fail on docs.rs?
> 
> We should think on how to fix this there as well. Setting RUSTDOCFLAGS is also possible: https://github.com/rust-lang/docs.rs/blob/8b19841b449244d79976c29ae24534af3d7e8c82/templates/core/Cargo.toml.example#L42

Would we need to set RUSTDOCFLAGS and include the after script in every crate? Maybe it's easier to open an issue downstream in the `aquamarine` crate and try get it fixed there..

liamaharon:
So this pr doesn't break crates.io and we can merge it I'm going to restore the `aquamarine` crate and use it alongside the new after-content script.

kianenigma:
> docs.rs

fwiw we have generally broader issues with docs.rs, especially with publishing `polkadot-sdk-docs` crate, see: https://github.com/sam0x17/docify/issues/22

It might then therefore be a better tradeoff to say for now we only focus on self-hosted rust docs.

ggwpez (approved):

liamaharon:
Confirmed fixed in prod 🥳 https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/traits/trait.Hooks.html

Issue #3775, opened 2024-03-21
statement-distribution: elastic scaling support


sandreim:
The current implementation assumes a 1:1 mapping between a parachain and backing group/core in quite a few places:
- https://github.com/paritytech/polkadot-sdk/blob/93b1abb280cf9d6c64f00b66c3593c2b04d50a9e/polkadot/node/network/statement-distribution/src/v2/mod.rs#L2129
- https://github.com/paritytech/polkadot-sdk/blob/93b1abb280cf9d6c64f00b66c3593c2b04d50a9e/polkadot/node/network/statement-distribution/src/v2/mod.rs#L2195
- https://github.com/paritytech/polkadot-sdk/blob/93b1abb280cf9d6c64f00b66c3593c2b04d50a9e/polkadot/node/network/statement-distribution/src/v2/mod.rs#L2314
- https://github.com/paritytech/polkadot-sdk/blob/93b1abb280cf9d6c64f00b66c3593c2b04d50a9e/polkadot/node/network/statement-distribution/src/v2/mod.rs#L3040

To fix this we need to implement something similar to the [assumed core index](https://github.com/paritytech/polkadot-sdk/blob/93b1abb280cf9d6c64f00b66c3593c2b04d50a9e/polkadot/node/core/backing/src/lib.rs#L1024) in the backing subsystem. This would then enable the subsystem to properly distribute/filter manifests based on the validator index / expected group considering there can be more than 1 backing group for a certain parachain.

PR #3879, opened 2024-03-28
statement-distribution: fix filtering of statements for elastic parachains

sandreim:
fixes https://github.com/paritytech/polkadot-sdk/issues/3775

Additionally moves the claim queue fetch utilities into `subsystem-util`.

TODO:
- [x] fix tests
- [x] add elastic scaling tests


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5687708

polkadot/node/subsystem-util/src/vstaging.rs
```diff
@@ -19,14 +19,19 @@
...
 use polkadot_overseer::SubsystemSender;
-use polkadot_primitives::{Hash, ValidatorIndex};
+use polkadot_primitives::{CoreIndex, Hash, Id as ParaId, ScheduledCore, ValidatorIndex};
 
-use crate::{has_required_runtime, request_disabled_validators, runtime};
+use crate::{has_required_runtime, request_claim_queue, request_disabled_validators, runtime};
 
 const LOG_TARGET: &'static str = "parachain::subsystem-util-vstaging";
 
+/// A snapshot of the runtime claim queue at an arbitrare relay chain block.
```
sandreim, line 32:
```suggestion
/// A snapshot of the runtime claim queue at an arbitrary relay chain block.
```

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2192,18 +2238,14 @@ async fn fragment_tree_update_inner<Context>(
...
 			if let (Some(confirmed), Some(prs)) = (confirmed_candidate, prs) {
-				let group_index = group_for_para(
-					&prs.availability_cores,
-					&prs.group_rotation_info,
-					receipt.descriptor().para_id,
-				);
-
 				let per_session = state.per_session.get(&prs.session);
-				if let (Some(per_session), Some(group_index)) = (per_session, group_index) {
+				// TODO(maybe for sanity): perform an extra check on the candidate backing group
```
alindima, line 2242: let's do this

polkadot/node/network/statement-distribution/src/v2/tests/cluster.rs
```diff
@@ -312,6 +312,66 @@ fn useful_cluster_statement_from_non_cluster_peer_rejected() {
...
+
+	let relay_parent = Hash::repeat_byte(1);
+	let peer_a = PeerId::random();
+
+	test_harness(config, |state, mut overseer| async move {
+		let candidate_hash = CandidateHash(Hash::repeat_byte(42));
+
+		let test_leaf = state.make_dummy_leaf_with_multiple_cores_per_para(relay_parent, 3);
+
+		// Peer A is not in our group, but it's group is assigned to same para as we are.
```
alindima, line 333:
```suggestion
		// Peer A is not in our group, but its group is assigned to same para as we are.
```

alindima (approved):
Looks good! Can we add the zombienet test for proper elastic scaling? or do you want to do that as a follow-up?

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -693,15 +693,24 @@ pub(crate) async fn handle_active_leaves_update<Context>(
...
 
+		let maybe_claim_queue = fetch_claim_queue(ctx.sender(), new_relay_parent)
+			.await
+			.map_err(JfyiError::FetchClaimQueue)?;
+
+		let groups_per_para = determine_groups_per_para(
+			&availability_cores,
+			&group_rotation_info,
+			&maybe_claim_queue,
+			seconding_limit - 1,
```
alexggh, line 704: Nit: I would make it explicit this is actually max_candidate_depth

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2126,17 +2135,54 @@ async fn provide_candidate_to_grid<Context>(
...
+						},
+						None => {
+							// Runtime doesn't support claim queue runtime api. Fallback to
+							// `next_up_on_available`
+							occupied_core.next_up_on_available.clone()
+						},
+					};
+
+					maybe_scheduled_core
+						.filter(|scheduled_core| scheduled_core.para_id == occupied_core.para_id())
```
alexggh, line 2168: Feels like this doesn't work if we change cores for paras for example:
Block1:
Core0, Occupied(Para1)
Core1, Occupied(Para2)
Core2, Occupied(Para1)


ClaimQueue: Core0 -> Para2, Core1 -> Para1, Core2 -> Para1

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2126,17 +2135,54 @@ async fn provide_candidate_to_grid<Context>(
...
+			CoreState::Free => None,
+		}
+	});
 
-	core_index
-		.map(|c| group_rotation_info.group_for_core(CoreIndex(c as _), availability_cores.len()))
+	let mut groups_per_para = HashMap::new();
+	// Map from `CoreIndex` to `GroupIndex` and collect as `HashMap`.
+	for (para, core_index) in para_core_indices {
+		let group_index = group_rotation_info.group_for_core(core_index, availability_cores.len());
```
alexggh, line 2181: Group rotation info needs the block to be able to compute the index with the usage of ClaimQueue which I guess is from the future, does this gives us the correct result at rotation boundaries ?

alexggh (commented):
The approach seems sane to me, left you some comments with some unclarities that I have.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: This filter does exactly that, it checks that the upcoming assignment is of the same para, otherwise the para / core tuple is filtered out.

sandreim: Do you refer strictly to Core1, where we should consider it as well ? If so I have to think about that a bit.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: Here we are interested in the group mapping for a core at the relay parent which is supposed to back candidates built on this relay parent. The claim queue surely refers to something in the future, but that backing group is always determined by the relay parent used to build the candidate.

sandreim: So yeah, it should give the correct result when we pass the rotation boundary, but that would be at the next relay parent.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: done!

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: Well, since we are interested in just the backing groups for this relay parent only, I don't think we should bother about the case I mentioned above.

sandreim:
> Looks good! Can we add the zombienet test for proper elastic scaling? or do you want to do that as a follow-up?

I'll craft a new PR with the test.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: This is not perfect, since for example in your case, para1 could still have assigned some other cores like Core10, and the backing group for core0 would still be used to back it and it would still be backed at Core10, but that also assumes some candidate pending availability there is included, which complicates things quite a lot

sandreim: I removed the filter, based on your example we now get backing groups for Core1, Core2 backing groups for Para1.

alexggh (approved):

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -3037,13 +3081,13 @@ pub(crate) async fn handle_response<Context>(
...
-					&relay_parent_state.availability_cores,
-					&relay_parent_state.group_rotation_info,
-					para,
-				);
+				let expected_groups = relay_parent_state.groups_per_para.get(&para);
 
-				Some(g_index) == expected_group
+				expected_groups.is_some() &&
+					expected_groups
+						.expect("checked is_some(); qed")
```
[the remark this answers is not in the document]
sandreim: fixed

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2126,17 +2137,53 @@ async fn provide_candidate_to_grid<Context>(
...
+	// on-demand parachains we also consider the core indices at next block if core has a candidate
+	// pending availability.
+	let para_core_indices = availability_cores.iter().enumerate().filter_map(|(index, core)| {
+		match core {
+			CoreState::Scheduled(scheduled_core) =>
+				Some((scheduled_core.para_id, CoreIndex(index as u32))),
+			CoreState::Occupied(occupied_core) => {
+				if max_candidate_depth >= 1 {
+					// Use claim queue if available, or fallback to `next_up_on_available`
+					let maybe_scheduled_core = match maybe_claim_queue {
```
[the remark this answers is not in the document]
sandreim: done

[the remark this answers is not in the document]
eskimor: > up to a certain point, right? given by allowed_ancestry_len AFAICT

Yes, but that limit should be enforced by the claim queue size itself (it should always we smaller or equal than that limit).

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2139,47 +2139,52 @@ async fn provide_candidate_to_grid<Context>(
...
-				}
-			},
-			CoreState::Free => None,
-		}
-	});
+	let para_core_indices: Vec<_> = if let Some(claim_queue) = maybe_claim_queue {
+		claim_queue
+			.keys()
+			.filter_map(|core_index| {
+				let Some(scheduled_core) = fetch_next_scheduled_on_core(&claim_queue, *core_index)
```
eskimor, line 2156: Ok, first (not introduced here): I find the name of this function vastly misleading as it does not fetch anything.

Second, why are we iterating only keys of the claim queue to fetch the actual element for each key, instead of iterating key/values to begin with?

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2139,47 +2139,52 @@ async fn provide_candidate_to_grid<Context>(
 
 // Utility function to populate per relay parent `ParaId` to `GroupIndex` mappings.
 fn determine_groups_per_para(
-	availability_cores: &[CoreState],
-	group_rotation_info: &GroupRotationInfo,
-	maybe_claim_queue: &Option<ClaimQueueSnapshot>,
+	availability_cores: Vec<CoreState>,
+	group_rotation_info: GroupRotationInfo,
+	maybe_claim_queue: Option<ClaimQueueSnapshot>,
```
eskimor, line 2144: Why are references no longer working?

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -693,15 +695,24 @@ pub(crate) async fn handle_active_leaves_update<Context>(
 			}
 		});
 
+		let maybe_claim_queue = fetch_claim_queue(ctx.sender(), new_relay_parent)
```
eskimor, line 698: Given that we pass per value (can't re-use the result), might make sense to move that call into `determine_groups_per_para`.

The result of request_availability_cores is used in `find_active_validator_state`, but wrongly. It uses `para_id` a function we should remove, as it makes no sense to either return the `ParaId` of the occupying para or the scheduled one.*)

Regardless, I would move the fetching code into the function that need the data. `handle_active_leaves_update` is already pages of code.

*) Note: This is obviously unrelated, so can be a separate PR, but it should be fixed for Coretime asap. @tdimitrov

polkadot/node/network/statement-distribution/src/v2/mod.rs
eskimor: See above, I think it would be best to call the runtime APIs from this function directly.

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2311,13 +2359,14 @@ async fn handle_incoming_manifest_common<'a, Context>(
...
-		&relay_parent_state.availability_cores,
-		&relay_parent_state.group_rotation_info,
-		para_id,
-	);
+	let expected_groups = relay_parent_state.groups_per_para.get(&para_id);
 
-	if expected_group != Some(manifest_summary.claimed_group_index) {
+	if expected_groups.is_none() ||
+		!expected_groups
+			.expect("checked is_some(); qed")
```
eskimor, line 2366: Missed that one. Please also use a pattern match.

polkadot/node/network/statement-distribution/src/v2/tests/cluster.rs
```diff
@@ -312,6 +312,66 @@ fn useful_cluster_statement_from_non_cluster_peer_rejected() {
 	});
 }
 
+// Both validators in the test are part of backing groups assigned to same parachain
+#[test]
+fn elastic_scaling_useful_cluster_statement_from_non_cluster_peer_rejected() {
```
eskimor, line 317: Nice!

eskimor:
Looks good, nevertheless had to leave a few nits.

polkadot/node/network/statement-distribution/src/v2/mod.rs
eskimor: https://github.com/paritytech/polkadot-sdk/issues/3948

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: > Why are references no longer working?

they work smoothly but no longer needed. I'll move the runtime api inside this fn.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: I wanted to reuse the fn, but you are right, it is actually useless in this case. A wrapper for this BTreeMap with a method like `iter_claims(depth: BlockNumber)` returning an iterator over `(core_index, para)` so we can lookup as far into the future as the lookahead param/ async backing settings allow.

sandreim: @tdimitrov wdyt ?

sandreim: https://github.com/paritytech/polkadot-sdk/pull/3950

alindima (approved):

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2138,28 +2137,34 @@ async fn provide_candidate_to_grid<Context>(
...
-fn determine_groups_per_para(
+async fn determine_groups_per_para(
+	sender: &mut impl overseer::StatementDistributionSenderTrait,
+	relay_parent: Hash,
 	availability_cores: Vec<CoreState>,
 	group_rotation_info: GroupRotationInfo,
-	maybe_claim_queue: Option<ClaimQueueSnapshot>,
 	max_candidate_depth: usize,
 ) -> HashMap<ParaId, Vec<GroupIndex>> {
+	let maybe_claim_queue = fetch_claim_queue(sender, relay_parent)
```
eskimor, line 2147: Ok, my thinking was to move all fetches here, including `availability_cores` for consistency. It's weird if we pass in one thing, but fetch the other our selves. I know this is used in one other place as well (wrongly), but given that we pass per value already, there is not much gain in avoiding the double fetch.

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -2359,14 +2373,12 @@ async fn handle_incoming_manifest_common<'a, Context>(
...
+		return None
+	};
 
-	if expected_groups.is_none() ||
-		!expected_groups
-			.expect("checked is_some(); qed")
-			.iter()
-			.any(|g| g == &manifest_summary.claimed_group_index)
-	{
+	if !expected_groups.iter().any(|g| g == &manifest_summary.claimed_group_index) {
```
eskimor, line 2381: Nice!

eskimor (approved):

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: LocalValidatorState is built outside this fn and requires `availability_cores`

Issue #3880, opened 2024-03-28
Make ProvingTrie Generic and Reusable


shawntabrizi:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Motivation

I was looking at ways to implement simple merkle proofs within a pallet, and it seems that `pallet_session::historical` has this ability with the `ProvingTrie`.

However, the full definition for this type is within the pallet crate, and thus the simple idea of creating and using a merkle trie and proof is not available to other pallets without fully reimplementing the type.

### Request

Create a generic and reusable version of the `ProvingTrie` which is made available by the Polkadot-SDK

### Solution

Take the existing `ProvingTrie`, and make it generic over the data being proved.

### Are you willing to help with this request?

Yes!

shawntabrizi:
going to try this quickly

PR #3881, opened 2024-03-28
Create a Basic Proving Trie for the Runtime

shawntabrizi:
This PR will introduce a `BasicProvingTrie` type, which makes it easy to construct and prove data in a base-16 merkle trie within the runtime.

Data into the merkle trie only require that they implement `Encode` / `Decode`.

A FRAME compatible `TrieError` was created and added to `DispatchError`.

Expected usage is to construct the merkle trie with all data offline, and then place only the merkle root of that trie on-chain.

Also offchain, a user is given a compact merkle proof of some data they want to prove exists on the blockchain.

Then in the runtime, you can call `verify_single_value_proof` or `verify_proof` with the root, proof, and the keys and values you want to verify exists in the merkle trie.

Closes https://github.com/paritytech/polkadot-sdk/issues/3880

Contributes to #5400


shawntabrizi:
For a full end to end usage in the runtime, I would like to create a test which uses this object to verify some data in a child trie.

However, I don't think the child trie currently supports creating a proof, and not sure how to get that working.

burdges:
I'd kinda expect merkle proofs should always be batched, so you prove for many keys simultaniously.  I've no worked out how much more efficent this makes them of course, but we so some dumb things like compressing, when the proofs should've never contained duplicates in the first palce. 

In principle, one could make this work in the runtme by using unsafe code which writes the queries into a buffer, and then some function in on_finalize checks them all.

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+use crate::{Decode, DispatchError, Encode};
+
+use sp_std::vec::Vec;
+use sp_trie::{
+	trie_types::{TrieDBBuilder, TrieDBMutBuilderV0},
+	LayoutV0, MemoryDB, Recorder, Trie, TrieMut, EMPTY_PREFIX,
+};
+
+/// A trait for creating a merkle trie for checking and generating merkle proofs.
+pub trait ProvingTrie<Hashing, Hash, Key, Value>
```
ggwpez, line 29: >Note: Decided not to update the session historical code to use this trait to avoid changing their code too much. Their implementation uses two keys rather than one to construct the trie, which seems specific to their scenario.

Can you create an issue for this please 😄 ? Sounds like a good mentor issue.

ggwpez, line 29: The `Hash` here should be implicetly known by `HashOf<Hashing>`, so this is for convenience?

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+{
+	fn generate_for<I>(items: I) -> Result<Self, DispatchError>
+	where
+		I: IntoIterator<Item = (Key, Value)>,
+	{
+		let mut db = MemoryDB::default();
+		let mut root = Default::default();
+
+		{
+			let mut trie = TrieDBMutBuilderV0::new(&mut db, &mut root).build();
```
ggwpez, line 76: @cheme why do we use a specific version here? Should this not be V1?

ggwpez (commented):
For airdrops? Yea i guess its useful. Maybe AAS (airdrop as a service) would also help.

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: im just not sure it is really needed that someone updates that object to use this new trait.

substrate/primitives/runtime/src/proving_trie.rs
cheme: In general the version does not matter much for proof as V0 can be use on a V1 trie and V1 can be use on a V0, and since it is a proof we mainly care for access node (both version should have same access patterns).
In this case though, because we return the root, we need to choose the right version (proof content will be the same but root will differ).
In the case of historical, I think the value where all 32 bytes so V0 and V1 would work the same, but I kept V0 to be conservative.
For a generic toolling, the version should be pass in parameter (even if in most case there is no reason to use V0).

cheme:
> I'd kinda expect merkle proofs should always be batched, so you prove for many keys simultaniously. I've no worked out how much more efficent this makes them of course, but we so some dumb things like compressing, when the proofs should've never contained duplicates in the first palce.
> 
> In principle, one could make this work in the runtme by using unsafe code which writes the queries into a buffer, and then some function in on_finalize checks them all.

in this case all insert are done at the same time and trie keep all change in memory before calculating the root.
There is a possible micro optimization that would be to sort with a btree then calculate new root without storing the content in triedbmut (but it will be store anyway so not really better memory wise (maybe currently we clone on instert though)). I remember writing this code (basically what trie-root does but with possibly some existin content), it is faster but by a low margin (pbbly ~20%) and the code being a bit involved I did not push it.

If considering batching between different extrinsic it is not doable currently (as the mem is not shared between two extrinsic (vm design I think)), I hope it will be doable with polkavm.

cheme:
Thinking of it, the best would be to use this inline root update calculation over ordered key value payload and store the stack of pending change (max size the max depth of the trie time max branch size and a few temp variables). This way it could be processed in multiple blocks or extrinsic.
I guess it can also be done in none order changeset but then the size of the pending change to store will be unbounded.

substrate/primitives/runtime/src/lib.rs
```diff
@@ -85,6 +85,7 @@ pub mod generic;
 pub mod legacy;
 mod multiaddress;
 pub mod offchain;
+pub mod proving_trie;
```
cheme, line 88: I am not too sure if it should be part of sp-runtime crate.
Thing is the choice of using this trie should not be a default for any runtime (I think it is not the best tree for most use case, and I would encourage more binary trie, anyway it is very use case specific).

substrate/primitives/runtime/src/proving_trie.rs
cheme, line 29: I am not sure I understand the need for a trait here, I would find it more usefull to have it only when needed (I mean there is only one implementation of this trait I think)

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	fn query(&self, key: Key) -> Option<Value> {
+		let trie = TrieDBBuilder::new(&self.db, &self.root).build();
+		key.using_encoded(|s| trie.get(s))
+			.ok()?
+			.and_then(|raw| Value::decode(&mut &*raw).ok())
+	}
+
+	fn create_proof(&self, key: Key) -> Option<Vec<Vec<u8>>> {
```
cheme, line 97: maybe better passing Vec<Key> and allow call multiple query here.
For now we can simply rename to create_single_value_access_proof I think.

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+		let mut recorder = Recorder::<LayoutV0<Hashing>>::new();
```
cheme, line 98: here version doesn't matter, can switch to V1 though (just to avoid confusion).

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+
+			key.using_encoded(|k| {
+				trie.get(k).ok()?.and_then(|raw| Value::decode(&mut &*raw).ok())
+			})?;
+		}
+
+		Some(recorder.drain().into_iter().map(|r| r.data).collect())
+	}
+
+	fn from_nodes(root: Hash, nodes: &[Vec<u8>]) -> Self {
```
cheme, line 112: this function need a warning in documentation stating that this only work for small set of content (proving on big set like air drop, this would be plain terrible: we should need to map nodes of state individually or by group into the state and only access when needed the only node need by the proof).

cheme (commented):
Code existing is fine, and I think extracting it can be a good move, I am just thinking it is pretty limited in its usages (~~small trie size,~~ single query at this time, suboptimal merkle trie) so I would prefer if it was not part of sp-runtime but in its own crate.
Would make sense to also replace the existing code to use this one.

substrate/primitives/runtime/src/proving_trie.rs
thiolliere: This is for checking proof isn't it? like by ensuring `from_nodes(proof).hash()` is equal to `root`

cheme: Oh yes I am dumb  :+1:

substrate/primitives/runtime/src/lib.rs
shawntabrizi: Do we have a binary trie available in Substrate?

Does this trie work along side child tries?

cheme: If the proof is to be run on child trie, there is no choice but to use this trie indeed.
(if using child trie root to prove, please be careful the chain will not be switching from trie V0 to trie V1 (should not matter for crowdloan as iirc the values are <= 32 byte), as a migration may change child trie root.
About this trie version, the simpliest would probably be to just pass the version in parameter.
When I read the PR I did not have this use case in mind but it is true that if what we are proving is from substrate state, then it could make sense to have this in runtime.

kianenigma:
@Ank4n @gpestana this primitive can possibly be used to migrate large chunks of data/snapshot to parachains as well, PTAL :)

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,178 @@
...
+	fn generate_for<I>(items: I) -> Result<Self, DispatchError>
+	where
+		I: IntoIterator<Item = (Key, Value)>;
+	/// Access the underlying trie root.
+	fn root(&self) -> &Hash;
+	/// Check a proof contained within the current memory-db. Returns `None` if the
+	/// nodes within the current `MemoryDB` are insufficient to query the item.
+	fn query(&self, key: Key) -> Option<Value>;
+	/// Create the full verification data needed to prove a key and its value in the trie. Returns
+	/// `None` if the nodes nodes within the current `MemoryDB` are insufficient to create a proof.
```
Ank4n, line 43:
```suggestion
	/// `None` if the nodes within the current `MemoryDB` are insufficient to create a proof.
```

Ank4n:
> @Ank4n @gpestana this primitive can possibly be used to migrate large chunks of data/snapshot to parachains as well, PTAL :)

Yeah we might end up doing something very similar to this to move staked balances to the staking parachain.

kianenigma:
> end up doing something very similar to this

I hope the outcome is similar and reusable enough that we push this PR to completion and use that. Do you think this is the case?

Polkadot-Forum:

https://forum.polkadot.network/t/disposable-parachains-for-airdrops-and-other-ideas/5769/20

Aideepakchaudhary:
Hi @shawntabrizi ,
I'm trying to add this code in a simple rust program and trying to create the proof outside the blockchain but I'm getting some error while running `cargo build`. Is it possible to create the proof outside the chain with the same code or not?

main.rs
``` rust
use csv;
mod proving_trie;
use anyhow::Error;
use sp_core::*;
use proving_trie::{BasicProvingTrie, ProvingTrie};
use codec::alloc::collections::BTreeMap;
use sp_runtime::traits::BlakeTwo256;
use sp_runtime::DispatchError;
// use sp_io::*;


fn read_from_file(path: &str) -> Result<Vec<(String,u128)>, anyhow::Error> {

    let mut account_details: Vec<(String,u128)> = Vec::new();
    let mut reader = csv::Reader::from_path(path)?;

    for result in reader.records() {
        let record = result?;

        let address = &record[0].trim();
        let balance = &record[1].trim().parse::<u128>()?;

        account_details.push( (address.to_string(), *balance));
    }
    Ok(account_details)
}

fn main() {
    let account_details  = read_from_file("./mockAccounts.csv").expect("Error while reading the file!");

    type BalanceTrie = BasicProvingTrie<BlakeTwo256, H256, String, u128>;

    let mut map = BTreeMap::<String, u128>::new();
    for i in 0..account_details.len() {
        map.insert(account_details[i].0.clone(), account_details[i.clone()].1.clone());
    }

    let trie = BalanceTrie::generate_for(map).expect("Root not generated");
    
}

```

facing this issue: 

```
error: linking with `cc` failed: exit status: 1
  |
  = note: env -u IPHONEOS_DEPLOYMENT_TARGET -u TVOS_DEPLOYMENT_TARGET -u XROS_DEPLOYMENT_TARGET LC_ALL="C" PATH="/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin:/opt/homebrew/opt/llvm/bin:/Users/deepak/.nvm/versions/node/v22.2.0/bin:/Users/deepak/.yarn/bin:/Users/deepak/.config/yarn/global/node_modules/.bin:/Users/deepak/.local/share/solana/install/active_release/bin:/Library/Frameworks/Python.framework/Versions/3.11/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/Apple/usr/bin:/usr/local/go/bin:/Users/deepak/.cargo/bin" VSLANG="1033" ZERO_AR_DATE="1" "cc" "-arch" "arm64" "/var/folders/n_/vrm0lsd95j3fr1jxkkq3phym0000gn/T/rustcd4pbNR/symbols.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.00.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.01.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.02.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.03.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.04.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.05.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.06.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.07.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.08.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.09.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.10.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.11.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.12.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.13.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.14.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.merkle_trie_cli.61e1fdef53eb68d6-cgu.15.rcgu.o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4.2pjibw1m09yi8ruo.rcgu.o" "-L" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps" "-L" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/build/secp256k1-sys-2121cb74afa609f0/out" "-L" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_runtime-e9705827fb6f398d.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_weights-74c2aace79a38f10.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libhash256_std_hasher-54b54b73512028ac.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libeither-013d9c063e5c360c.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_application_crypto-fa5b2c91430e9fb4.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_io-3a0abebed44e2cc3.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_arithmetic-92fd022bdd999734.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libinteger_sqrt-02a1cdc2c475ed91.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libnum_traits-fc96149d9e7ed281.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libanyhow-eb37837f0a0ab740.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_trie-520d4eef15885ccd.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_core-4f9d8083efa7c92b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/liblibsecp256k1-46203ce465467f08.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/librand-3ca1b5c200b0d819.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/liblibsecp256k1_core-b542267fe41546a9.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbitflags-c5795c861ae0760a.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbounded_collections-9094c065b6bba5f5.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/liblog-3cfe4ce4dd5a2b71.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libschnorrkel-c1a066cf1584727b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libmerlin-7f195bc87d4a17a6.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsha2-2df6dd9f3ecfa877.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libopaque_debug-8eff4400edb164b1.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libfake_simd-f8223c13de698977.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libblock_buffer-9d4e35314a4e46f2.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libblock_padding-a18a851fe8e40c5b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbyte_tools-98b7426342210149.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcurve25519_dalek-70fdfc6d5dcd04a4.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libdigest-91eea25c40663ddf.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libgeneric_array-3c3c86558dc3eb0f.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libed25519_zebra-d754470cce200782.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsha2-6afc99ccb3d83310.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libopaque_debug-9d2646870a17b872.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libblock_buffer-4c8e88c8b63bb507.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/librand_core-a333f70e66680358.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libhashbrown-136f8563444f2cdf.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libahash-eb17b6f57de58f82.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libgetrandom-078dbd448d8ce578.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/liblibc-94666428b53ffcf1.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcurve25519_dalek-4aeddee8a7a19511.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/librand_core-59c905e6eb8e3695.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libdigest-1b97bf9ec1562b6a.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1-a06fffab3eff7bb4.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1_sys-76438ad26072a6d2.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libss58_registry-76a87d3b6a79348b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libzeroize-ea6d5bdce8a8eb09.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_core_hashing-b6d1bab3dbac9a0b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsha2-72b2ecc22520833d.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsha3-cfea2b197c45ce36.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libkeccak-0c3d926a924cc05f.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtwox_hash-012aed9e2cc63090.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libblake2b_simd-e3cb6df99456f896.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libconstant_time_eq-b1afcc795bf7188e.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libarrayref-23cb54cc8f2da81b.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libdigest-645edfda8e19e9cd.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsubtle-84e66cd0cc18962d.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libblock_buffer-36e923480643e64f.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcrypto_common-464945110f8807cd.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libgeneric_array-3f3cda384c1ff5b8.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtypenum-29d10208e4bf8ba5.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_runtime_interface-b144b0592c9f3143.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_externalities-8f980631d059d6da.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libenvironmental-e8a2694ee5e532ba.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_storage-54ac4a35e4b234d6.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libref_cast-2e2c385847616e50.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_wasm_interface-ab8650bb4a375cdb.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libprimitive_types-55a6271bd55492b1.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libimpl_codec-2eb232dbde6d1723.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libuint-8450243421ba032e.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcrunchy-160912b463962e25.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libhex-d75db9f3d18cd6e4.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbyteorder-49c628e853923d46.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libfixed_hash-5ec372025142ffb1.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libstatic_assertions-bb3435cf1688aa17.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_tracing-b6d02bb9d81eb0ae.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtracing-4f2d475327e40dfc.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libpin_project_lite-64fb84ace3799e87.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtracing_core-ffaace9d10475155.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libmemory_db-e1ed6d7075cc7e01.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtrie_root-dcfa022b0cb1261f.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libscale_info-f92ff22af71c0ff9.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libtrie_db-83b0f7cbc69ed5b9.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsmallvec-7e69410f8976cd9e.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libhashbrown-336e5d6442bec1be.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libahash-6eb43eeacb59c315.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libonce_cell-30bdc62a1f28b8e4.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcfg_if-237ca97c10329bbb.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libzerocopy-137786a4bcf8a936.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libhash_db-cfcf443a4397e672.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libparity_scale_codec-f98397f26eae00dd.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libarrayvec-0162638febd04212.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbytes-d8b6c969ef2db873.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libbyte_slice_cast-fb4734530808ea6d.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsp_std-81e4d7aa4a73abaf.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcsv-c12a828e56298563.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libryu-593dffdec51dbc6c.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libitoa-938b23dfd7edab47.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libcsv_core-276d2fb8f04f7dd6.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libmemchr-6e0389edcacd1ee9.rlib" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libserde-9d80e5a85fd27070.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libstd-a7cb6d736aafff9f.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libpanic_unwind-93bc7737b38f7178.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libobject-cf8d39441fdd966a.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libmemchr-79c8c8150cb92bd5.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libaddr2line-3eddc7b4e472f56d.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libgimli-9d76ad8219925a7c.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/librustc_demangle-91d4d96a1f5d2185.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libstd_detect-7c84f7ef4e985efd.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libhashbrown-d38cf6c4f6ee8e43.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/librustc_std_workspace_alloc-e160f965e54ddbbc.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libminiz_oxide-4af12eecfecf4e51.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libadler-8fffc5a23f5aa4e8.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libunwind-ebd9a5041183cc77.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libcfg_if-978bae832c5c7d0a.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/liblibc-e49f559b9aa8fc23.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/liballoc-9dced93467c9110e.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/librustc_std_workspace_core-a4ecc63307a2b2e4.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libcore-4b44a1965098f9de.rlib" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libcompiler_builtins-bc98379c8a911237.rlib" "-liconv" "-lSystem" "-lc" "-lm" "-L" "/Users/deepak/.rustup/toolchains/nightly-2024-04-14-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib" "-o" "/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/merkle_trie_cli-9629d8edc3333fd4" "-Wl,-dead_strip" "-nodefaultlibs"
  = note: ld: warning: object file (/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1_sys-76438ad26072a6d2.rlib[4](84f2453a56d059fc-lax_der_parsing.o)) was built for newer 'macOS' version (14.4) than being linked (14.0)
          ld: warning: object file (/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1_sys-76438ad26072a6d2.rlib[5](c928380082b86401-precomputed_ecmult_gen.o)) was built for newer 'macOS' version (14.4) than being linked (14.0)
          ld: warning: object file (/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1_sys-76438ad26072a6d2.rlib[6](c928380082b86401-precomputed_ecmult.o)) was built for newer 'macOS' version (14.4) than being linked (14.0)
          ld: warning: object file (/Users/deepak/Documents/Disposable_chain/merkle-trie-cli/target/release/deps/libsecp256k1_sys-76438ad26072a6d2.rlib[7](c928380082b86401-secp256k1.o)) was built for newer 'macOS' version (14.4) than being linked (14.0)
          Undefined symbols for architecture arm64:
            "_ext_hashing_blake2_256_version_1", referenced from:
                sp_io::hashing::extern_host_function_impls::blake2_256::h03a3aee0751b044a in libsp_io-3a0abebed44e2cc3.rlib[6](sp_io-3a0abebed44e2cc3.sp_io.bc8cfb1b37e70159-cgu.3.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: removed it

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: updated name. would you be wiling to write code for the multi-access query?

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: Can i remove the layout alltogether? seems I must include this wrapper to satisfy traits

shawntabrizi:
@cheme @ggwpez sorry for super delay in updating this PR.

Can you take a second look and let me know what else is needed to be mergeable?

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: So you want me to make this a parameter we pass in? or make it V1, or leave it untouched?

substrate/primitives/runtime/src/proving_trie.rs
shawntabrizi: Updated name and also added the multi-value API

substrate/primitives/runtime/src/proving_trie.rs
cheme: either V1 only (would not be compatible in some case) or pass it as a parameter.

substrate/primitives/runtime/src/lib.rs
cheme: I am not too sure anymore what was my point :facepalm: 
Probably at first reading I was thinking it should be an external crate, then realizing it is being tightly linked with substrate state could actually be fine.
`ProvingTrieBase16` may not be a good idea as long as substrate does have a single trie backend.
Maybe it could simple be a module in sp-trie, but guess I am ok with things being this way.

substrate/primitives/runtime/src/proving_trie.rs
```diff
@@ -0,0 +1,235 @@
...
+	pub fn query(&self, key: Key) -> Option<Value> {
+		let trie = TrieDBBuilder::new(&self.db, &self.root).build();
+		key.using_encoded(|s| trie.get(s))
+			.ok()?
+			.and_then(|raw| Value::decode(&mut &*raw).ok())
+	}
+
+	/// Create the full verification data needed to prove all `keys` and their values in the trie.
+	/// Returns `None` if the nodes within the current `MemoryDB` are insufficient to create a
+	/// proof.
```
cheme, line 81: Since we do not pass the trie version in parameter and use v1 as default, it seems important to me to state it in the doc.
``` Proof are created with latest substrate trie format, runtime using non latest should not use it (eg trie_version = 0).```

**Review by cheme — APPROVED** [review]

File: substrate/primitives/runtime/src/proving_trie.rs (hunk 8 above)
Comment (shawntabrizi) [reply substrate/primitives/runtime/src/proving_trie.rs:179]: done

**Review by ggwpez — APPROVED** [review]

File: prdoc/pr_3881.prdoc [hunk 9]
```diff
@@ -0,0 +1,15 @@
...
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR introduces a Proving Trie object which can be used inside the runtime. This can allow
+      for things like airdrops where a single hash is stored on chain representing the whole airdrop
+      and individuals present a proof of their inclusion in the airdrop.
+
+crates:
+  - name: sp-runtime
+    bump: minor
```
ggwpez, line 15:
```suggestion
    bump: major
```
Probably from adding the variant to `DispatchError`.

Issue #3888, opened 2024-03-29
Break StakingInterface into smaller traits


Ank4n:
StakingInterface is too big and bloated.

It could be broken down into something like this
- trait StakingConfig {} // can be merged with StakingInspect
- trait StakingInspect {}
- trait StakingMutate: StakingInspect {}
- trait StakingUnsafe: StakingMutate {}

Other pallets that only need to query Staking configuration or data about a nominator/validator can use the right level of traits. 

`StakingMutate` allows all the ways to mutate the staking ledger. 

`StakingUnsafe` gives access to low level apis that allow other pallets to manipulate ledger overriding some general integrity checks. These are to be used by `DelegationStakingPallet` that is being worked on https://github.com/paritytech/polkadot-sdk/pull/2680.

PR #3889, opened 2024-03-29
Allow privileged virtual bond in Staking pallet

Ank4n:
This is the first PR in preparation for https://github.com/paritytech/polkadot-sdk/issues/454.

## Follow ups:
- https://github.com/paritytech/polkadot-sdk/pull/3904.
- https://github.com/paritytech/polkadot-sdk/pull/3905.

[Maybe followup](https://github.com/paritytech/polkadot-sdk/issues/4217) with migration of storage item `VirtualStakers` as a bool or enum in `Ledger`. 

## Context
We want to achieve a way for a user (`Delegator`) to delegate their funds to another account (`Agent`). Delegate implies the funds are locked in delegator account itself. Agent can act on behalf of delegator to stake directly on Staking pallet.

The delegation feature is added to Staking via another pallet `delegated-staking` worked on [here](https://github.com/paritytech/polkadot-sdk/pull/3904). 

### StakingUnchecked Trait
As the name implies, this trait allows unchecked (non-locked) mutation of staking ledger. These apis are only meant to be used by other pallets in the runtime and should not be exposed directly to user code path. Also related: https://github.com/paritytech/polkadot-sdk/issues/3888.

### Virtual Bond
Allows other pallets to stake via staking pallet while managing the locks on these accounts themselves. Introduces another storage `VirtualStakers` that whitelist these accounts. 

We also restrict virtual stakers to set reward account as themselves. Since the account has no locks, we cannot support compounding of rewards. Conservatively, we require them to set a separate account different from the staker. Since these are code managed, it should be easy for another pallet to redistribute reward and rebond them.

### Slashes
Since there is no actual lock maintained by staking-pallet for virtual stakers, this pallet does not apply any slashes. It is then important for pallets managing virtual stakers to listen to slashing events and apply necessary slashes.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5693970

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1315,9 +1307,7 @@ pub mod pallet {
 				Error::<T>::ControllerDeprecated
 			);
 
-			let _ = ledger
-				.set_payee(payee)
-				.defensive_proof("ledger was retrieved from storage, thus its bonded; qed.")?;
+			ledger.set_payee(payee)?;
```
Ank4n, line 1310: It can fail for other reasons than unbonded (not exists in storage).

substrate/primitives/staking/src/lib.rs
```diff
@@ -304,6 +310,27 @@ pub trait StakingInterface {
 	fn set_current_era(era: EraIndex);
 }
 
+/// Set of low level apis to manipulate staking ledger.
+///
+/// These apis bypass some or all safety checks and should only be used if you know what you are
+/// doing.
+pub trait StakingUnsafe: StakingInterface {
```
Ank4n, line 317: Move force_unstake also here.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1132,6 +1167,49 @@ impl<T: Config> Pallet<T> {
...
 	}
+
+	/// Whether the passed reward destination is restricted for the given account.
+	///
+	/// virtual stakers are not allowed to compound their rewards as this pallet does not manage
+	/// locks for them. For external pallets that manage the virtual bond, it is their
+	/// responsibility to distribute the reward and re-bond them.
+	///
+	/// Conservatively, we expect them to always set the reward destination to a non stash account.
+	pub(crate) fn restrict_reward_destination(
```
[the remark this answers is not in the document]
georgepisaltu: not nit: About the usage of this function - this is currently used in `Ledger::set_payee` and `Ledger::bond`, please correct me if I missed anything.

Call instances for this function seem to be as follows:
- for `set_payee`, it's called in the `set_payee` and `update_payee` extrinsics
- for `bond`, it's called in the `bond` extrinsic and in the `virtual_bond` implementation

It seems to me that we could move this check to the callers. Functionally it would be the same, but it's probably the responsibility of the callers to ensure a correct reward destination rather than the ledger, which doesn't even know why the destination is restricted.

substrate/frame/staking/src/tests.rs
```diff
@@ -6849,6 +6849,102 @@ mod staking_interface {
 	}
 }
 
+mod staking_unsafe {
```
georgepisaltu, line 6852: We should have some tests to ensure `StakingInterface` impl consistency:
- existing getter-like functions work normally with virtual stakers (e.g. `stash_by_ctrl`, `total_stake`, etc.); `withdraw_unbonded` and `unbond` are already partially covered on the happy path, but it might be worth it to add a test case for invalid calls after a virtual bond as well, just to be safe
- functions which shouldn't work fail as expected (e.g. calling `StakingInterface::bond` after calling `<StakingInterface as StakingUnsafe>::virtual_bond`)

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -379,6 +379,16 @@ pub mod pallet {
...
 
+	/// Stakers whose funds are managed by other pallets.
+	///
+	/// This pallet does not apply any locks on them, therefore they are only virtually bonded. They
+	/// are expected to be keyless accounts and hence should not be allowed to mutate their ledger
+	/// directly via this pallet. Instead, these accounts are managed by other pallets and accessed
+	/// via low level apis. We keep track of them to do minimal integrity checks.
+	// TODO(ank4n): Can we keep this entry in `Ledger`? Worth a migration?
+	#[pallet::storage]
+	pub type VirtualStakers<T: Config> = CountedStorageMap<_, Twox64Concat, T::AccountId, ()>;
```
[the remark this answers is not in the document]
Ank4n: Ideally I would like to move this to `StakingLedger` since it is always read in conjunction with that. What we can probably do is, have it like this in the initial version and migrate it as a bool/enum field in ledger along with Currency-Fungible MBM.

cc: @gpestana @kianenigma  thoughts?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5776270

substrate/primitives/staking/src/lib.rs
```diff
@@ -254,6 +254,9 @@ pub trait StakingInterface {
 	/// schedules have reached their unlocking era should allow more calls to this function.
 	fn unbond(stash: &Self::AccountId, value: Self::Balance) -> DispatchResult;
 
+	/// Update the reward destination for the ledger associated with the stash.
+	fn update_payee(stash: &Self::AccountId, reward_acc: &Self::AccountId) -> DispatchResult;
```
gpestana, line 258: We want to keep this trait as general as possible, but at the same time it seems that the `update_payee` is too simple in the context of the staking pallet. An accountId is not powerful enough to express the `RewardDestination` type used in our staking system. Maybe we could parameterise this method with a generic that can be used to express the reward destination in the staking pallet?

substrate/primitives/staking/src/lib.rs
gpestana: How about `StakingUnchecked`? "unsafe" makes me feel unsafe :)

substrate/frame/staking/src/pallet/mod.rs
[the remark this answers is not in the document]
gpestana: I would add a new field in the `StakingLedger` and save the extra reads associated with checking if a ledger is virtual if we keep this info in a separate storage. The downside is the migration, but I hope we can get to a place where migrations are not blockers or make us think twice about refactors and design decisions, especially as the MB migrations are used more and more.

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -858,6 +869,10 @@ pub mod pallet {
 		ControllerDeprecated,
 		/// Cannot reset a ledger.
 		CannotRestoreLedger,
+		/// Provided reward destination is not allowed.
+		RewardDestinationRestricted,
+		/// Not enough funds available to withdraw
```
gpestana, line 874:
```suggestion
		/// Not enough funds available to withdraw.
```

substrate/frame/staking/src/pallet/mod.rs
gpestana: The current assumptions are:
1. If ledger is bonded, then a new ledger is returned from `Self::ledger`;
2. If ledger is bonded, then it has a Payee entry 

What changed in these assumptions with this PR? `ledger.set_payee(payee)` should always be Ok in this context (provided there's no underlying issues with the storage state, as we've seen recently. But in this case, it's a system bug and should not happen).

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -149,6 +151,39 @@ impl<T: Config> Pallet<T> {
 		Self::slashable_balance_of_vote_weight(who, issuance)
 	}
 
+	pub(super) fn do_bond_extra(stash: &T::AccountId, additional: BalanceOf<T>) -> DispatchResult {
+		let mut ledger = Self::ledger(StakingAccount::Stash(stash.clone()))?;
+
+		// for virtual stakers, we don't need to check the balance. Since they are only accessed
+		// via low level apis, we can assume that the caller has done the due diligence.
+		let extra = if Self::is_virtual_staker(stash) {
+			additional
```
gpestana, line 160: If we add the field saying that a ledger is virtual to the staking ledger, I'd add a debug_assert here to double checking that the ledger is in fact virtual.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -149,6 +151,39 @@ impl<T: Config> Pallet<T> {
...
+		ensure!(ledger.active >= T::Currency::minimum_balance(), Error::<T>::InsufficientBond);
+
+		// NOTE: ledger must be updated prior to calling `Self::weight_of`.
+		ledger.update()?;
+		// update this staker in the sorted list, if they exist in it.
+		if T::VoterList::contains(stash) {
+			let _ = T::VoterList::on_update(&stash, Self::weight_of(stash)).defensive();
+		}
+
+		Self::deposit_event(Event::<T>::Bonded { stash: stash.clone(), amount: extra });
```
gpestana, line 182: nit but I could see it being helpful in the future for traceability to distinguish ledger-related events that are virtual and non-virtual. Maybe we could add a field `virtual: bool` at very low cost.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -149,6 +151,39 @@ impl<T: Config> Pallet<T> {
...
+			additional.min(
+				T::Currency::free_balance(stash)
+					.checked_sub(&ledger.total)
+					.ok_or(sp_runtime::ArithmeticError::Overflow)?,
+			)
+		};
+
+		ledger.total += extra;
+		ledger.active += extra;
+		// Last check: the new active amount of ledger must be more than ED.
```
gpestana, line 172:
```suggestion
		// last check: the new active amount of ledger must be more than ED.
```

for consistency

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1132,6 +1167,31 @@ impl<T: Config> Pallet<T> {
 	) -> Exposure<T::AccountId, BalanceOf<T>> {
 		EraInfo::<T>::get_full_exposure(era, account)
 	}
+
+	/// Whether `who` is a virtual staker whose funds are managed by another pallet.
+	pub(crate) fn is_virtual_staker(who: &T::AccountId) -> bool {
+		VirtualStakers::<T>::contains_key(who)
```
gpestana, line 1173: As stated before, I think we could get rid of this extra storage map and add this info directly into the ledger. In any case, I'd remove this one-liner for readability.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1748,6 +1808,23 @@ impl<T: Config> StakingInterface for Pallet<T> {
 			.map(|_| ())
 	}
 
+	fn update_payee(stash: &Self::AccountId, reward_acc: &Self::AccountId) -> DispatchResult {
```
gpestana, line 1811: I'd prefer this `update_payee` method to take a `RewardDestination<AccountId>` instead of a reward account. That way we could allow the caller to set payee to `RewardDestination::None`, `RewardDestination::Staked`, etc or any other variants we have have in the future.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1860,6 +1941,41 @@ impl<T: Config> StakingInterface for Pallet<T> {
 	}
 }
 
+impl<T: Config> sp_staking::StakingUnsafe for Pallet<T> {
+	fn migrate_to_virtual_staker(who: &Self::AccountId) {
```
gpestana, line 1945: Thinking out loud here, but I guess we don't want to allow the callers to do the reverse operation through this trait? e.g. `migrate_to_non_virtual_staker`. This can easily be done through unbonding the virtual account and bonding with staking though, so probably better to not implement it directly.

substrate/frame/staking/src/pallet/impls.rs
gpestana: In any case, it seems that the reward destination for virtual accounts should always be `Stash(accountid)`. But it would be helpful to have this trait as general as possible so that it can be used in any case.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1860,6 +1941,41 @@ impl<T: Config> StakingInterface for Pallet<T> {
...
+		T::Currency::remove_lock(crate::STAKING_ID, who);
+		VirtualStakers::<T>::insert(who, ());
+	}
+
+	fn virtual_bond(
+		who: &Self::AccountId,
+		value: Self::Balance,
+		payee: &Self::AccountId,
```
gpestana, line 1953: As per the comments above, `RewardDestination<AccountId>` instead of accountid would be great for generality. However, in this specific case it's not so important, since the payee of a virtual bond will always have to be `Stash(accountid)` (where accountId != who). This assumption could change in the future though.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1860,6 +1941,41 @@ impl<T: Config> StakingInterface for Pallet<T> {
...
+		// check if payee not same as who.
+		ensure!(who != payee, Error::<T>::RewardDestinationRestricted);
+
+		// mark this pallet as consumer of `who`.
+		frame_system::Pallet::<T>::inc_consumers(&who).map_err(|_| Error::<T>::BadState)?;
+
+		// mark who as a virtual staker.
+		VirtualStakers::<T>::insert(who, ());
+
+		Self::deposit_event(Event::<T>::Bonded { stash: who.clone(), amount: value });
```
gpestana, line 1968:
```suggestion
		Self::deposit_event(Event::<T>::Bonded { stash: who.clone(), amount: value, virtual: true });
```

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1984,6 +2100,7 @@ impl<T: Config> Pallet<T> {
 	/// * Staking ledger and bond are not corrupted.
 	fn check_ledgers() -> Result<(), TryRuntimeError> {
```
gpestana, line 2101: Are there any checks we could add for the virtual bonds? e.g.:
- Payee of virtual bonds should always be `Stash(accountId)` where `accountId != who`
- Staking lock of virtual bond account == 0
- others?

substrate/frame/staking/src/pallet/impls.rs
Ank4n: Yeah we shouldn't. For practical purposes, `DelegateStake` is the v2 strategy and we should never really be going back. But technically yeah its possible to write a migration to iterate over delegators and move funds to this virtual ledger account.

substrate/frame/staking/src/pallet/impls.rs
Ank4n: I agree it can be. Though not sure where the boundary is to overgeneralizing 😀 . Nobody else uses this `StakingInterface` and we don't have that requirement. This will only currently disable self compounding.

substrate/frame/staking/src/pallet/impls.rs
georgepisaltu: I'd argue the distinction between virtual and non-virtual bonds should be made one level higher, as in an event from the entity doing the delegation should deposit a "virtual" event. At the staking pallet level, it probably doesn't matter whether the bonded funds are virtual or not, so long as they are bonded, but I don't feel too strongly about this.

substrate/frame/staking/src/pallet/impls.rs
georgepisaltu: `RewardDestination` is an implementation detail, not just conceptually but also in the code (it's not part of staking primitives, probably for a reason). Looking at the rest of the interface, I don't think it fits naturally here and it wouldn't make sense to bring it over. If other (more general) use cases appear, I'd argue it's more ergonomic to add new functions serving them and the impls behind the scenes can use their own structures (like `RewardDestination`) to reuse code.

substrate/frame/staking/src/pallet/impls.rs
georgepisaltu: > we don't want to allow the callers to do the reverse operation
>
> This can easily be done through unbonding the virtual account and bonding with staking

We should surely not *disallow* this, and we probably don't even have a way to disallow it even if we wanted to. But yes, I agree that we shouldn't introduce unneeded functionality in the interface.

> `DelegateStake` is the v2 strategy and we should never really be going back

Our best bet is to incentivize users to use v2 through superior functionality/performance. If there was nothing to gain from using v2, we probably wouldn't be building it, but I think it's clear there's value here and people won't go back.

substrate/frame/staking/src/pallet/impls.rs
georgepisaltu: As stated above, I wouldn't introduce `RewardDestination` here because, in short:
1. it's an implementation detail of the higher impl type
2. we don't have a use case for another destination other than an account

substrate/frame/staking/src/pallet/mod.rs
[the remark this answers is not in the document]
georgepisaltu: Because the number of virtual stakers starts at 0 and they are expected to grow over time, but still be a minority compared to regular stakers, I think it's ok to have a separate storage item for this. It would feel bad to migrate all of the ledgers for a few accounts which will become virtual stakers.

Staking functions are storage intensive anyway, but are not called that often and I don't expect this extra read to alter the dynamic too much. Worth mentioning that the ledger logic is already pretty complex, so any addition there has to have a good reason IMO.

If there are a lot of virtual stakers and it starts to make sense to optimize space and compute, we could write a migration in the future.

substrate/frame/staking/src/pallet/mod.rs
georgepisaltu: I agree with @gpestana , I don't think there's anything that changed regarding those assumptions (anymore). The previous impl had some restricted rewards destination checks IIRC, but they are no longer there so we're back to the initial assumptions.

substrate/primitives/staking/src/lib.rs
georgepisaltu: Reiterating, `RewardDestination` is an implementation detail and isn't really necessary in this context of a staking interface.

Let's run a thought experiment. Say we bring `RewardDestination` to the staking primitives to allow for more general uses of functions like this. It is an interface anyone can implement, it's not constrained to `pallet-staking`. What if users implementing this interface have additional rewards destinations other than what we currently use in `pallet-staking`? What if they use just one destination out of the several we provide? We just constrained them in this interface to what we are using in `pallet-staking`.

In this example, let's switch the approach and make this an associated type. That doesn't help right now because we don't use other types. If there are users out there that would want other reward destinations, they can just extend `StakingInterface` themselves with that additional type or an additional function and bake in any extra logic. It's what we are doing with `StakingUnsafe` (pending rename) right now :grin:.

We should aim at keeping this interface as simple as possible while enabling anyone who implements it to interact with other substrate-based staking consumers. My personal feeling is that the function right now is a good balance of functionality and ergonomics.

substrate/primitives/staking/src/lib.rs
georgepisaltu: I like `StakingUnchecked` way more.

Ank4n: What do you think about StakingInsecure since it either removes security (deposit) or does not care about one.

substrate/frame/staking/src/pallet/mod.rs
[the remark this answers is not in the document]
Ank4n: Virtual stakers will be same as number of pools. Though there are like ~180 pools in prod today.

The reason to move it to ledger: 1) It is always read in conjunction with ledger. 2) Ledger impl maintains the lock so we could argue it functionally belongs in Ledger.
Reason not to move: Will require a MBM so we can batch it with Currency-Fungible migration. 

I kind of lean towards not moving since I don't think it is a huge overhead (it does not even try to read the value from storage key). And at some point later we can migrate it into ledger.

substrate/frame/staking/src/pallet/impls.rs
Ank4n: Added checks 👍

substrate/frame/staking/src/pallet/impls.rs
Ank4n: > Yeah we shouldn't. For practical purposes, `DelegateStake` is the v2 strategy and we should never really be going back. But technically yeah its possible to write a migration to iterate over delegators and move funds to this virtual ledger account.

Actually had to do the reverse op for testing, i.e. setting up a pool that has not been migrated yet.

substrate/frame/staking/src/pallet/mod.rs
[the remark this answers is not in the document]
kianenigma: No strong opinion, but I would also not over optimize thing by adding another migration here. Separate map is fine for now. If and when that extra read is an issue, we will deal with it. Quite possibly it will never be an issue ;)

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -149,6 +151,39 @@ impl<T: Config> Pallet<T> {
...
+		} else {
+			// additional amount or actual balance of stash whichever is lower.
+			additional.min(
+				T::Currency::free_balance(stash)
+					.checked_sub(&ledger.total)
+					.ok_or(sp_runtime::ArithmeticError::Overflow)?,
+			)
+		};
+
+		ledger.total += extra;
```
kianenigma, line 170:
```suggestion
		ledger.total += extra;
```

saturating?

substrate/frame/staking/src/pallet/impls.rs
kianenigma: So semantically, a normal staker and virtual staker have all the same data, except: 

- real stakers have a lock, but no entry in `VirtualStakers`
- virtual stakers have no lock, but an entry in `VirtualStakers`.

kianenigma: As for the rest, I would just encourage everyone to remember that these are more or less internal APIs so I would generally opt for simplicity rather than possibly over-engineering them.

Moreover, once pools' participation in governance + stake tracker is done, I would say the staking system is "safe and feature complete", and at that point we can freeze the existing code, and decide if we want to iterate further on it, or if need be write a new version from scratch. Given this possibility, I think it is acceptable to take a bit of compromise and add even a tiny bit more of debt to this code for the sake of simplicity.

substrate/primitives/staking/src/lib.rs
```diff
@@ -304,6 +310,33 @@ pub trait StakingInterface {
...
+/// doing.
+pub trait StakingUnsafe: StakingInterface {
+	/// Migrate an existing staker to a virtual staker.
+	///
+	/// It would release all funds held by the implementation pallet.
+	fn migrate_to_virtual_staker(who: &Self::AccountId);
+
+	/// Book-keep a new bond for `who` without applying any locks (hence virtual).
+	///
+	/// It is important that who is a keyless account and therefore cannot interact with staking
```
kianenigma, line 325:
```suggestion
	/// It is important that `keyless_who` is a keyless account and therefore cannot interact with staking
```

substrate/primitives/staking/src/lib.rs
```diff
@@ -304,6 +310,33 @@ pub trait StakingInterface {
...
+	/// pallet directly. Caller is responsible for ensuring the passed amount is locked and valid.
+	fn virtual_bond(
+		keyless_who: &Self::AccountId,
+		value: Self::Balance,
+		payee: &Self::AccountId,
```
kianenigma, line 330: Seems like there are some constraints on `payee` as well that are not explained here.

substrate/frame/staking/src/pallet/impls.rs
```diff
@@ -1860,6 +1941,54 @@ impl<T: Config> StakingInterface for Pallet<T> {
...
+		// mark this pallet as consumer of `who`.
+		frame_system::Pallet::<T>::inc_consumers(&who).map_err(|_| Error::<T>::BadState)?;
+
+		// mark who as a virtual staker.
+		VirtualStakers::<T>::insert(who, ());
+
+		Self::deposit_event(Event::<T>::Bonded { stash: who.clone(), amount: value });
+		let ledger = StakingLedger::<T>::new(who.clone(), value);
+
+		// You're auto-bonded forever, here. We might improve this by only bonding when
```
kianenigma, line 1971: This comment is pretty unclear to me.

substrate/frame/staking/src/pallet/impls.rs
kianenigma: I generally agree with @georgepisaltu more, but in general staking is already aware and leaking info about "virtual vs. not" in any case, so it is a bit of a lost cause :p

substrate/frame/staking/src/mock.rs
```diff
@@ -249,17 +249,22 @@ parameter_types! {
...
-		_pool_account: &AccountId,
+		pool_account: &AccountId,
 		slashed_bonded: Balance,
 		slashed_chunks: &BTreeMap<EraIndex, Balance>,
-		_total_slashed: Balance,
+		total_slashed: Balance,
 	) {
 		LedgerSlashPerEra::set((slashed_bonded, slashed_chunks.clone()));
+		// update the observer.
+		let mut slash_observer = SlashObserver::get();
```
kianenigma, line 265: `static` has `fn mutate` afaik.

substrate/frame/staking/src/mock.rs
```diff
@@ -598,6 +603,18 @@ pub(crate) fn bond_nominator(who: AccountId, val: Balance, target: Vec<AccountId
...
 }
 
+pub(crate) fn bond_virtual_nominator(
+	who: AccountId,
+	payee: AccountId,
+	val: Balance,
+	target: Vec<AccountId>,
+) {
+	// who is provided by another pallet in a real scenario.
+	System::inc_providers(&who);
```
kianenigma, line 613: A bit fishy. Can you explain more? if `who` has enough ED, that should be a provider?

substrate/frame/staking/src/slashing.rs
```diff
@@ -609,26 +609,28 @@ pub fn do_slash<T: Config>(
...
 
 	let value = ledger.slash(value, T::Currency::minimum_balance(), slash_era);
+	if value.is_zero() {
+		// nothing to do
+		return
+	}
 
-	if !value.is_zero() {
+	// Skip slashing for virtual stakers. The pallets managing them should handle the slashing.
+	if !Pallet::<T>::is_virtual_staker(stash) {
```
kianenigma, line 618:
```suggestion
	if !Pallet::<T>::is_virtual_staker(stash) && !value.is_zero() {
```

substrate/frame/staking/src/tests.rs
kianenigma: I like the systematic thinking here, yes we should have these tests.

substrate/frame/staking/src/tests.rs
```diff
@@ -6849,6 +6847,227 @@ mod staking_interface {
...
+			// Because slashing happened.
+			assert!(is_disabled(11));
+
+			// but virtual nominator's balance is not slashed.
+			assert_eq!(Balances::free_balance(&101), nominator_balance);
+			// but slash is broadcasted to slash observers.
+			assert_eq!(SlashObserver::get().get(&101).unwrap(), &nominator_share);
+		})
+	}
+}
```
kianenigma, line 7070: good test coverage in general though 👍

kianenigma (approved):
LGTM

kianenigma (approved):

substrate/primitives/staking/src/lib.rs
Ank4n: It is more of an implementation constraint and not a trait level constraint. 

Added doc to impl.

substrate/frame/staking/src/pallet/impls.rs
Ank4n: Inherited from the original bond. Removing it. But I think it means the bond happens before someone actually puts in the intention to nominate/validate.

substrate/frame/staking/src/mock.rs
Ank4n: who is a keyless acc in a real scenario and another pallet (delegation staking pallet) provides for it. Improved the comment.

georgepisaltu (approved):

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -379,6 +379,17 @@ pub mod pallet {
 	pub type Nominators<T: Config> =
 		CountedStorageMap<_, Twox64Concat, T::AccountId, Nominations<T>>;
 
+	/// Stakers whose funds are managed by other pallets.
+	///
+	/// This pallet does not apply any locks on them, therefore they are only virtually bonded. They
+	/// are expected to be keyless accounts and hence should not be allowed to mutate their ledger
+	/// directly via this pallet. Instead, these accounts are managed by other pallets and accessed
+	/// via low level apis. We keep track of them to do minimal integrity checks.
+	// TODO(ank4n): Can we keep this entry in `Ledger`? Or migrate later in conjunction with
```
gpestana, line 388: Since this PR will keep the metadata of whether a ledger is virtual or not in storage, wdyt about opening an (unconfirmed) issue to keep track of this TODO and link it here?

substrate/frame/staking/src/slashing.rs
```diff
@@ -609,26 +609,28 @@ pub fn do_slash<T: Config>(
...
+		.defensive_proof("ledger fetched from storage so it exists in storage; qed.");
 
-		// trigger the event
-		<Pallet<T>>::deposit_event(super::Event::<T>::Slashed {
-			staker: stash.clone(),
-			amount: value,
-		});
-	}
+	// trigger the event
+	<Pallet<T>>::deposit_event(super::Event::<T>::Slashed { staker: stash.clone(), amount: value });
```
gpestana, line 633: This event will be deposited if it is a virtual staker with an amount that actually was not slashed here. Maybe we could return earlier if the ledger is virtual instead? Or throw a different event here than `Slashed`.

gpestana (approved):

substrate/frame/staking/src/mock.rs
```diff
@@ -249,17 +249,19 @@ parameter_types! {
...
 	fn on_slash(
-		_pool_account: &AccountId,
+		pool_account: &AccountId,
 		slashed_bonded: Balance,
 		slashed_chunks: &BTreeMap<EraIndex, Balance>,
-		_total_slashed: Balance,
+		total_slashed: Balance,
 	) {
 		LedgerSlashPerEra::set((slashed_bonded, slashed_chunks.clone()));
+		SlashObserver::mutate(|map| map.insert(*pool_account, total_slashed));
```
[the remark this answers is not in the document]
Ank4n: Yes good catch.

substrate/frame/staking/src/pallet/impls.rs
Ank4n: Since there is already a corresponding event in delegated staking pallet, I am leaning towards leaving it as it is. A dapp can get all information they need by reading other pallet events as well.

substrate/frame/staking/src/pallet/mod.rs
Ank4n: https://github.com/paritytech/polkadot-sdk/issues/4217

substrate/frame/staking/src/ledger.rs
```diff
@@ -257,6 +260,7 @@ impl<T: Config> StakingLedger<T> {
 
 			<Bonded<T>>::remove(&stash);
 			<Payee<T>>::remove(&stash);
+			<VirtualStakers<T>>::remove(&stash);
```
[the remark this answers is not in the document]
Ank4n: Done

substrate/frame/staking/src/slashing.rs
Ank4n: From this pallet's point of view it is still slashed right? The staking ledger is slashed.

For virtual stakers, this staking pallet does not manage locks so it cannot slash but it sends a slashing event for lock managing pallet to apply slashes.

If I am too pedantic, I would say we can have different events for each type of staker, but I think it is an overkill. Also, slash of a virtual staker will emit two events, this and another by delegated staking pallet which is enough to know for a dapp that this is a virtual staker. Wdyt?

Polkadot-Forum:

PR #3902, opened 2024-03-29
`im-online` removal final cleanup

s0me0ne-unkn0wn:
Rejoice! Rejoice! The story is nearly over.

This PR removes stale migrations, auxiliary structures, and package dependencies, thus making Rococo and Westend totally free from any `im-online`-related stuff.

`im-online` still stays a part of the Substrate node and its runtime: https://github.com/paritytech/polkadot-sdk/blob/0d9324847391e902bb42f84f0e76096b1f764efe/substrate/bin/node/runtime/src/lib.rs#L2276-L2277 I'm not sure if it makes sense to remove it from there considering that we're not removing `im-online` from FRAME. Please share your opinion.


ordian (approved):
I think we should remove this pallet after some time from Substrate and maybe find new maintainers in https://github.com/open-web3-stack/open-runtime-module-library or whoever wants to use it.

bkchr (approved):

PR #3903, opened 2024-03-29
Test that account is not Endowed before receiving balance

ggwpez:


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5698934

ggwpez:
Okay this does not work for the case where an account exists for a non-balance reason - like sufficient asset.  
We dont seem to have a test that asserts on the endowed events, going to add one.

ggwpez:
Okay i think its not that important.

PR #3904, opened 2024-03-30
Introduces: Delegated Staking Pallet

Ank4n:
This is the second PR in preparation for https://github.com/paritytech/polkadot-sdk/issues/454.

## Also see
- **Precursor** https://github.com/paritytech/polkadot-sdk/pull/3889.
- **Follow up** https://github.com/paritytech/polkadot-sdk/pull/3905.

### Delegation Interface
Provides delegation primitives for staking. 

Introduces two new roles:
- Agent: These are accounts who receive delegation from other accounts (delegators) and stakes on behalf of them. The funds are held in delegator accounts.
- Delegator: Accounts who delegate their funds to an agent authorising them to use it for staking.

Supports
- A way for delegators to add or withdraw delegation to an agent.
- A way for an agent to slash a delegator during a slashing event.

### Pallet Delegated Staking
- Implements `DelegationInterface`.
- Lazy slashing: Any slashes to an Agent is posted in a ledger but not immediately slashed. The agent can call `DelegationInterface::delegator_slash` to slash the member and clear the corresponding slash from its ledger. 
- Consumes `StakingInterface` to provide `CoreStaking` features. In reality, this will be `pallet-staking`.
- Ensures bookkeeping for agent and delegator are correct but leaves the management of reward and slash logic upto the consumer of this pallet. 
- While it does not expose any calls yet, it is written with the intent of exposing these primitives via extrinsics.

## TODO
- [x] Improve unit tests in the pallet.
- [x] Separate slash reward perbill for rewarding the slash reporters?
- [x] Review if we should add more events.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699231

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699251

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699250

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699249

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699546

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699715

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699714

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5699713

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,309 @@
...
+		Delegation { agent: agent.clone(), amount }
+	}
+
+	/// Ensure the delegator is either a new delegator or they are adding more delegation to the
+	/// existing agent.
+	///
+	/// Delegators are prevented from delegating to multiple agents at the same time.
+	pub(crate) fn can_delegate(delegator: &T::AccountId, agent: &T::AccountId) -> bool {
+		Delegation::<T>::get(delegator)
+			.map(|delegation| delegation.agent == agent.clone())
```
georgepisaltu, line 60: No need to clone, just deref one or ref the other.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,309 @@
...
+	pub(crate) fn save_or_kill(self) -> Result<(), DispatchError> {
+		let key = self.key;
+		// see if delegate can be killed
+		if self.ledger.total_delegated == Zero::zero() {
+			ensure!(
+				self.ledger.unclaimed_withdrawals == Zero::zero() &&
+					self.ledger.pending_slash == Zero::zero(),
+				Error::<T>::BadState
+			);
+			<Agents<T>>::remove(key);
```
georgepisaltu, line 276: Should we also `dec_providers` here too if we remove the agent?

georgepisaltu (commented):
First pass, need to look at the tests and re-read the docs again.

substrate/frame/delegated-staking/src/impls.rs
```diff
@@ -0,0 +1,277 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Implementations of public traits, namely [StakingInterface], [DelegatedStakeInterface] and
+//! [OnStakingUpdate].
+
+use super::*;
+use sp_staking::{DelegatedStakeInterface, OnStakingUpdate};
+
+/// Wrapper `StakingInterface` implementation for `Agents`.
+impl<T: Config> StakingInterface for Pallet<T> {
```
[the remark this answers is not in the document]
Ank4n: Yeah I agree, `DelegatedStakeInterface` could be made independent. Will make the refactor. 👍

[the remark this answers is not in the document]
Ank4n: https://github.com/paritytech/polkadot-sdk/pull/3905/commits/3988cff3d7dfc891ea754107e5e42e8769350693

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5746379

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5779793

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5780087

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5833354

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+			origin: OriginFor<T>,
+			reward_account: T::AccountId,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			// Existing `agent` cannot register again.
+			ensure!(!Self::is_agent(&who), Error::<T>::NotAllowed);
+
+			// A delegator cannot become an `agent`.
+			ensure!(!Self::is_delegator(&who), Error::<T>::NotAllowed);
```
gpestana, line 282:
```suggestion
			// Existing `agent` cannot register again and a delegator cannot become an `agent`.
			ensure!(!Self::is_agent(&who) && !Self::is_delegator(&who), Error::<T>::NotAllowed);
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+
+			// They cannot be already a direct staker in the staking pallet.
+			ensure!(Self::not_direct_staker(&who), Error::<T>::AlreadyStaking);
```
gpestana, line 285:
```suggestion
			ensure!(!Self::direct_staker(&who), Error::<T>::AlreadyStaking);
```

In general I think it's simpler to avoid negations in the method names. e.g. `!not_direct_staker` becomes double negation and it's harder to read.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+
+	/// Returns true if who is delegating to an `Agent` account.
+	fn is_delegator(who: &T::AccountId) -> bool {
+		<Delegators<T>>::contains_key(who)
+	}
+
+	/// Returns true if who is not already staking on [`Config::CoreStaking`].
+	fn not_direct_staker(who: &T::AccountId) -> bool {
+		T::CoreStaking::status(who).is_err()
+	}
```
gpestana, line 443: As said above, I think it's simpler to avoid negations in the method names. e.g. !not_direct_staker becomes double negation and it's harder to read.

```suggestion
	/// Returns true if who is already staking on [`Config::CoreStaking`].
	fn is_direct_staker(who: &T::AccountId) -> bool {
		T::CoreStaking::status(who).is_ok()
	}
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+	/// functions defined in the following impl block should act as a good reference for how the
+	/// exposed calls would look like when exposed.
+	impl<T: Config> Pallet<T> {
+		/// Register an account to become a stake `Agent`. Sometimes also called a `Delegatee`.
+		///
+		/// Delegators can authorize `Agent`s to stake on their behalf by delegating their funds to
+		/// them. The `Agent` can then use the delegated funds to stake to [`Config::CoreStaking`].
+		///
+		/// Implementation note: This function allows any account to become an agent. It is
+		/// important though that accounts that call [`StakingUnsafe::virtual_bond`] are keyless
```
gpestana, line 267: > It is important though that accounts that call [`StakingUnsafe::virtual_bond`] are keyless

I really can't think of a nice way to check/enforce this but maybe something to think about for FRAME. I'd say au that not having a way to ensure/check that the origin is a keyless account may become problematic.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+			ensure!(!Self::is_agent(&who), Error::<T>::NotAllowed);
+
+			// A delegator cannot become an `agent`.
+			ensure!(!Self::is_delegator(&who), Error::<T>::NotAllowed);
+
+			// They cannot be already a direct staker in the staking pallet.
+			ensure!(Self::not_direct_staker(&who), Error::<T>::AlreadyStaking);
+
+			// Reward account cannot be same as `agent` account.
+			ensure!(reward_account != who, Error::<T>::InvalidRewardDestination);
```
gpestana, line 288: Should we also check if the `reward_account` is not an agent? That's what we are trying to avoid but there's nothing preventing that from happening if `reward_account != who`.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+		pub fn migrate_to_agent(
+			origin: OriginFor<T>,
+			reward_account: T::AccountId,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			// ensure who is not already an agent.
+			ensure!(!Self::is_agent(&who), Error::<T>::NotAllowed);
+
+			// and they should already be a nominator in `CoreStaking`.
+			ensure!(Self::is_direct_nominator(&who), Error::<T>::NotAllowed);
```
gpestana, line 316:
```suggestion
			let who = ensure_signed(origin)?;
			
			// ensure who is not already an agent and it is a nominator in `CoreStaking`.
			ensure!(!Self::is_agent(&who) && Self::is_direct_nominator(&who), Error::<T>::NotAllowed);
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+
+			// Reward account cannot be same as `agent` account.
+			ensure!(reward_account != who, Error::<T>::InvalidRewardDestination);
```
gpestana, line 319: As commented above, maybe we should make sure that `reward_account != who && !Self::is_agent(reward_account)`.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+		}
+
+		/// Delegate given `amount` of tokens to an `Agent` account.
+		///
+		/// If `origin` is the first time delegator, we add them to state. If they are already
+		/// delegating, we increase the delegation.
+		///
+		/// Conditions:
+		/// - Delegators cannot delegate to more than one agent.
+		/// - The `agent` account should already be registered as such. See [`Self::register_agent`]
```
gpestana, line 382:
```suggestion
		/// - The `agent` account should already be registered as such. See [`Self::register_agent`].
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+			Delegators::<T>::get(source_delegator).defensive_ok_or(Error::<T>::BadState)?;
+
+		// some checks that must have already been checked before.
+		ensure!(source_delegation.amount >= amount, Error::<T>::NotEnoughFunds);
+		debug_assert!(
+			!Self::is_delegator(destination_delegator) && !Self::is_agent(destination_delegator)
+		);
+
+		// update delegations
+		Delegation::<T>::from(&source_delegation.agent, amount).save_or_kill(destination_delegator);
```
gpestana, line 607: Instead of `amount`, should we instead move all the funds held by the proxy for this delegator into the new `destination_delegator`? That would basically mean to remove `amount` as input of this method and just max out the balance based on the delegator's balance in this proxy agent. 

I believe this is better as proxy delegations are temp and we want to clear them out as soon as possible.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,308 @@
...
+
+	/// Checked increase of delegation amount. Consumes self and returns a new copy.
+	#[allow(unused)]
+	pub(crate) fn increase_delegation(self, amount: BalanceOf<T>) -> Option<Self> {
+		let updated_delegation = self.amount.checked_add(&amount)?;
+		Some(Delegation::from(&self.agent, updated_delegation))
+	}
+
+	/// Save self to storage. If the delegation amount is zero, remove the delegation.
+	pub(crate) fn save_or_kill(self, key: &T::AccountId) {
```
gpestana, line 81: nit

```suggestion
	pub(crate) fn update_or_kill(self, key: &T::AccountId) {
```

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,308 @@
...
+		T::CoreStaking::stake(&self.key).is_ok()
+	}
+
+	/// Returns the reward account registered by the agent.
+	pub(crate) fn reward_account(&self) -> &T::AccountId {
+		&self.ledger.payee
+	}
+
+	/// Save self to storage.
+	pub(crate) fn save(self) {
```
gpestana, line 259: How about having a `checked_mutate` like fn here that does all the integrity checks before saving to storage? some checks could be:

- `stakeable >= bonded_stake`
- `pending_slash <= bonded_stake`
- etc..

gpestana: Ah, I see that `fn save_or_kill` below implements a similar approach before killing the ledger. I'd check the more broad ledger integrity checks here and fail with `BadState` if something is off. 

If we get into a similar situation with the corrupted staking ledger has we saw recently, at least we don't corrupt the ledger further.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,309 @@
...
+use super::*;
+use frame_support::traits::DefensiveSaturating;
+
+/// The type of pot account being created.
+#[derive(Encode, Decode)]
+pub(crate) enum AccountType {
+	/// A proxy delegator account created for a nominator who migrated to an `Agent` account.
+	///
+	/// Funds for unmigrated `delegator` accounts of the `Agent` are kept here.
+	ProxyDelegator,
```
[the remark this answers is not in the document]
Ank4n: When we expose this via extrinsics, what we should do is have agent accounts also derived (keyless). This would mean they could never access staking pallet directly and have to go via delegated staking pallet. This is also a pre-condition for virtual bond.

substrate/frame/delegated-staking/src/impls.rs
```diff
@@ -0,0 +1,144 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Implementations of public traits, namely [DelegationInterface] and [OnStakingUpdate].
```
kianenigma, line 19:
```suggestion
//! Implementations of public traits, namely [`DelegationInterface`] and [`OnStakingUpdate`].
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+			hold::{
+				Balanced as FunHoldBalanced, Inspect as FunHoldInspect, Mutate as FunHoldMutate,
+			},
+			Balanced, Inspect as FunInspect, Mutate as FunMutate,
+		},
+		tokens::{fungible::Credit, Precision, Preservation},
+		Defensive, DefensiveOption, Imbalance, OnUnbalanced,
+	},
+};
+
```
kianenigma, line 149:
```suggestion
```

kianenigma: Then RA will format them :)

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+//! # Delegated Staking Pallet
+//!
+//! This pallet implements [`sp_staking::DelegationInterface`] that provides delegation
+//! functionality to `delegators` and `agents`. It is designed to be used in conjunction with
+//! [`StakingInterface`] and relies on [`Config::CoreStaking`] to provide primitive staking
+//! functions.
+//!
+//! Currently, it does not expose any dispatchable calls but is written with a vision to expose them
+//! in the future such that it can be utilised by any external account, off-chain entity or xcm
+//! multi location such as a parachain or a smart contract.
```
kianenigma, line 27:
```suggestion
//! `MultiLocation` such as a parachain or a smart contract.
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		/// Operation not supported by this pallet.
+		NotSupported,
+	}
+
+	/// A reason for placing a hold on funds.
+	#[pallet::composite_enum]
+	pub enum HoldReason {
+		/// Funds held for stake delegation to another account.
+		#[codec(index = 0)]
+		Delegating,
```
kianenigma, line 230: Might be confused in a UI with governance delegation, so let's call it `StakingDelegation` or similar?

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		CountedStorageMap<_, Twox64Concat, T::AccountId, Delegation<T>, OptionQuery>;
+
+	/// Map of `Agent` to their `Ledger`.
+	#[pallet::storage]
+	pub(crate) type Agents<T: Config> =
+		CountedStorageMap<_, Twox64Concat, T::AccountId, AgentLedger<T>, OptionQuery>;
+
+	/// This pallet is not currently written with the intention of exposing any calls. But the
+	/// functions defined in the following impl block should act as a good reference for how the
+	/// exposed calls would look like when exposed.
```
kianenigma, line 259:
```suggestion
	// This pallet is not currently written with the intention of exposing any calls. But the
	// functions defined in the following impl block should act as a good reference for how the
	// exposed calls would look like when exposed.
```

substrate/frame/delegated-staking/src/lib.rs
kianenigma: > I don't know of a way to check if an account is keyless

Was thinking of the same. The only thing is that if we create a one-off sub-account with a "secure random derivation path" from the origin? 

in any case this is a problem of the future.

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,322 @@
...
+	bounds::{ElectionBounds, ElectionBoundsBuilder},
+	onchain, SequentialPhragmen,
+};
+use frame_support::dispatch::RawOrigin;
+use pallet_staking::CurrentEra;
+use sp_core::U256;
+use sp_runtime::traits::Convert;
+use sp_staking::{Stake, StakingInterface};
+
+pub type T = Runtime;
```
kianenigma, line 39: nice

substrate/frame/delegated-staking/src/lib.rs
kianenigma: In this case, all `reward_account`s also need to be kept in a map and always checked. Else, I can make alice my reward account, and register her as an agent (with an arbitrary reward account) later.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,760 @@
...
+			ensure!(reward_account != who, Error::<T>::InvalidRewardDestination);
+
+			Self::do_register_agent(&who, &reward_account);
+			Ok(())
+		}
+
+		/// Migrate from a `Nominator` account to `Agent` account.
+		///
+		/// The origin needs to
+		/// - be a `Nominator` with `CoreStaking`,
```
kianenigma, line 294:
```suggestion
		/// - be a `Nominator` with [`Config::CoreStaking`],
```

substrate/primitives/staking/src/lib.rs
```diff
@@ -455,4 +455,103 @@ pub struct PagedExposureMetadata<Balance: HasCompact + codec::MaxEncodedLen> {
...
+		amount: Self::Balance,
+	) -> DispatchResult;
+
+	/// Returns true if there are pending slashes posted to the `Agent` account.
+	///
+	/// Slashes to `Agent` account are not immediate and are applied lazily. Since `Agent`
+	/// has an unbounded number of delegators, immediate slashing is not possible.
+	fn has_pending_slash(agent: &Self::AccountId) -> bool;
+
+	/// Apply a pending slash to a `Agent` by slashing `value` from `delegator`.
```
kianenigma, line 527:
```suggestion
	/// Apply a pending slash to an `Agent` by slashing `value` from `delegator`.
```

substrate/primitives/staking/src/lib.rs
```diff
@@ -455,4 +455,103 @@ pub struct PagedExposureMetadata<Balance: HasCompact + codec::MaxEncodedLen> {
...
+	///
+	/// If a reporter is provided, the reporter will receive a fraction of the slash as reward.
```
kianenigma, line 529: where is this fraction dictated?

substrate/primitives/staking/src/lib.rs
```diff
@@ -455,4 +455,103 @@ pub struct PagedExposureMetadata<Balance: HasCompact + codec::MaxEncodedLen> {
...
+		delegator: &Self::AccountId,
+		value: Self::Balance,
+		maybe_reporter: Option<Self::AccountId>,
+	) -> sp_runtime::DispatchResult;
+
+	/// Migrate an existing `Nominator` to `Agent` account.
+	///
+	/// The implementation should ensure the `Nominator` account funds are moved to an escrow
+	/// from which `Agents` can later release funds to its `Delegators`.
+	fn migrate_nominator_to_agent(
```
kianenigma, line 541: if all of this abstractions help us have a easy journey where all existing nominator can become pools with 1 click, that is an amazing north star.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+		assert_ok!(DelegatedStaking::delegate_to_agent(
+			RawOrigin::Signed(delegator).into(),
+			agent,
+			100
+		));
+
+		// verify
+		assert!(DelegatedStaking::is_agent(&agent));
+		assert_eq!(DelegatedStaking::stakeable_balance(&agent), 100);
+		assert_eq!(Balances::balance_on_hold(&HoldReason::Delegating.into(), &delegator), 100);
```
kianenigma, line 50:
```suggestion
		assert_eq!(Balances::balance_on_hold(&HoldReason::Delegating.into(), &delegator), 100);
		assert_eq!(DelegatedStaking::held_balance_of(&delegator), 100);
```

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,322 @@
...
+		});
+
+		ext
+	}
+	pub fn build_and_execute(self, test: impl FnOnce()) {
+		sp_tracing::try_init_simple();
+		let mut ext = self.build();
+		ext.execute_with(test);
+		ext.execute_with(|| {
+			DelegatedStaking::do_try_state().unwrap();
```
kianenigma, line 245:
```suggestion
			<AllPallets as TryState>::try_state(now).unwrap();
```

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,322 @@
...
+	Staking::stake(&who).unwrap() == Stake { total, active } &&
+		get_agent(&who).ledger.stakeable_balance() == total
+}
+
+pub(crate) fn get_agent(agent: &AccountId) -> Agent<T> {
+	Agent::<T>::from(agent).expect("delegate should exist")
+}
+
+#[allow(unused)]
+pub(crate) fn held_balance(who: &AccountId) -> Balance {
```
kianenigma, line 302: seems duplicate to a fn on `Pallet`

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+			Error::<T>::InvalidDelegation
+		);
+	});
+}
+
+use sp_staking::DelegationInterface;
+#[test]
+fn apply_pending_slash() {
+	ExtBuilder::default().build_and_execute(|| {
+		start_era(1);
```
kianenigma, line 198: should move atop.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,308 @@
...
+		Delegation::<T>::get(delegator)
+			.map(|delegation| delegation.agent == agent.clone())
+			.unwrap_or(
+				// all good if it is a new delegator except it should not be an existing agent.
+				!<Agents<T>>::contains_key(delegator),
+			)
+	}
+
+	/// Checked decrease of delegation amount. Consumes self and returns a new copy.
+	pub(crate) fn decrease_delegation(self, amount: BalanceOf<T>) -> Option<Self> {
```
kianenigma, line 68: admittedly a bit of a strange from a pure Rust persepctive. Consuming self, returning `Option<Self>`, why? 

First comment, the fields of the struct are totally public. Possibly, you can just alter those in place, with checks. 

If you want to go the extra mile, I suggest this somewhat battle-tested FRAME pattern: 

- make the fields private, and make a standard with yourself that you nerver read the storage directly.
- have a `fn get(key)` and `fn put(self)`.
- All other mutable functions would work with `fn foo(&mut self)`

kianenigma: I would say the current code is reasonable IFF there is a strong reason where you don't want this item to remain in scope in case the addition fails. That is a good reason to consume `self`, and very Rust-idiomatic. But otherwise it is a bit against all idioms I know of, and at an optimization nitpick it has needless allocation to move objects around instead of refs.

substrate/frame/delegated-staking/src/types.rs
kianenigma: This is good, and same as `fn put(self)` I imagined: you are done with this, it should not remain in scope, therefore consume self, and write back to storage. If it is zero, writing is basically deleting, else updating.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,309 @@
...
+pub struct Agent<T: Config> {
+	/// storage key
+	pub key: T::AccountId,
+	/// storage value
+	pub ledger: AgentLedger<T>,
+}
+
+impl<T: Config> Agent<T> {
+	/// Get `Agent` from storage if it exists or return an error.
+	pub(crate) fn from(agent: &T::AccountId) -> Result<Agent<T>, DispatchError> {
```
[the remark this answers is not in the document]
kianenigma: Yes, `new` is more Rust-idiomatic from my experience. 

In other words, a constructor in Rust is typically called `fn new()`, a conversion method is called `from`.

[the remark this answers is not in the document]
kianenigma:
```
AgentLedger::<T>::get(agent)
			.ok_or(Error::<T>::NotAgent)
			.map(|ledger| Agent { key: agent.clone(), ledger })
```

yours is fewer lines tho :))

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,308 @@
...
+			return Ok(true)
+		}
+		self.ledger.save(&key);
+		Ok(false)
+	}
+
+	/// Reloads self from storage.
+	pub(crate) fn refresh(&self) -> Result<Agent<T>, DispatchError> {
+		Self::from(&self.key)
+	}
```
kianenigma, line 288: now _this_ is actually a function that I would expect it to consume self, because it is possibly dangerous for the call site to have two copies of this value.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,308 @@
...
+		let ledger = AgentLedger::<T>::get(agent).ok_or(Error::<T>::NotAgent)?;
+		Ok(Agent { key: agent.clone(), ledger })
+	}
+
+	/// Remove funds that are withdrawn from [Config::CoreStaking] but not claimed by a delegator.
+	///
+	/// Checked decrease of delegation amount from `total_delegated` and `unclaimed_withdrawals`
+	/// registers. Consumes self and returns a new instance of self if success.
+	pub(crate) fn remove_unclaimed_withdraw(
+		self,
```
kianenigma, line 178: sth to keep in mind is that all FRAME dispatchables are not executing in a transactional layer, so if you take a mutable reference, and still return `Result<(), Err>`, and the return is `Err`, eventually any change to `self` will revert.

But if you find this better, I also get it.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+			);
+			// assert_noop!(DelegatedStaking::release_delegation(RawOrigin::Signed(agent).into(),
+			// 200, 50, 0), Error::<T>::NotAllowed); active and total stake remains same
+			assert!(eq_stake(agent, total_staked, total_staked));
+
+			// 305 wants to unbond 50 in era 2, withdrawable in era 5.
+			assert_ok!(Staking::unbond(RawOrigin::Signed(agent).into(), 50));
+			// 310 wants to unbond 100 in era 3, withdrawable in era 6.
+			start_era(3);
+			assert_ok!(Staking::unbond(RawOrigin::Signed(agent).into(), 100));
```
kianenigma, line 347: nit: `assert_noop` implies this

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+				DelegatedStaking::release_delegation(RawOrigin::Signed(agent).into(), 301, 50, 0),
+				Error::<T>::NotEnoughFunds
+			);
+			// assert_noop!(DelegatedStaking::release_delegation(RawOrigin::Signed(agent).into(),
+			// 200, 50, 0), Error::<T>::NotAllowed); active and total stake remains same
+			assert!(eq_stake(agent, total_staked, total_staked));
+
+			// 305 wants to unbond 50 in era 2, withdrawable in era 5.
+			assert_ok!(Staking::unbond(RawOrigin::Signed(agent).into(), 50));
+			// 310 wants to unbond 100 in era 3, withdrawable in era 6.
```
kianenigma, line 345: this is some commented code?

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+			start_era(3);
+			assert_ok!(Staking::unbond(RawOrigin::Signed(agent).into(), 100));
+			// 320 wants to unbond 200 in era 4, withdrawable in era 7.
+			start_era(4);
```
kianenigma, line 349: it is worth noting and contemplating again on the fact that: 

This pallet does not bookkeeping on which delegator initiated which agen unbond operation. If a higher level pallet doesn't regualte this, delegators can front run each other in unbonding and withdrawing. 

Does that make sense?

substrate/frame/delegated-staking/src/types.rs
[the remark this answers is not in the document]
Ank4n: renamed to `get()` which is more appropriate here.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,809 @@
...
+//!  2) stake from pool account as an `Agent` account on the staking pallet.
+//!
+//! The difference being, in the second approach, the delegated funds will be locked in-place in
+//! user's account enabling them to participate in use cases that allows use of `held` funds such
+//! as participation in governance voting.
+//!
+//! Nomination pool still does all the heavy lifting around pool administration, reward
+//! distribution, lazy slashing and as such, is not meant to be replaced with this pallet.
+//!
+//! ## Limitations
```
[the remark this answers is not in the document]
Ank4n: I don't really see it as a limitation. Given that stash account is a keyless account, does not really mean anything by paying rewards there.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: Don't think it matters. The only real restriction we need to impose is that we cannot have `RewardDestination::Staked` (compounding) in `CoreStaking`. Reward cannot be stash is a bit conservative but also seems cleaner to impose stash and reward account to be always separate.

substrate/primitives/staking/src/lib.rs
Ank4n: @kianenigma For pools, my plan was to migrate them on runtime upgrade. It is possible to allow both strategies at the same time in NominationPool: `TransferStake` and `DelegateStake` and allow pool to upgrade one at a time. Leaning towards migrating all but looking forward to know what you think once you get the chance to review https://github.com/paritytech/polkadot-sdk/pull/3905.

For members themselves, they can claim their delegation in 1 click, or be made permissionless and we (bot script) do the claiming for them.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: same as above. removed the commented code.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: Yes, it is assuming that the agent manages this (which is what nomination pool does today). Added in the crate docs.

substrate/frame/delegated-staking/src/types.rs
Ank4n: All frame dispatchables are executing in a tx layer right?

Didn't really get the point though?

substrate/frame/delegated-staking/src/types.rs
Ank4n: Yeah not sure I was trying to do but removed these and added inline sub.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		let mut delegation = Delegation::<T>::get(delegator).ok_or(Error::<T>::NotDelegator)?;
+
+		// make sure delegation to be released is sound.
+		ensure!(&delegation.agent == who, Error::<T>::NotAgent);
+		ensure!(delegation.amount >= amount, Error::<T>::NotEnoughFunds);
+
+		// if we do not already have enough funds to be claimed, try withdraw some more.
+		if agent.ledger.unclaimed_withdrawals < amount {
+			// withdraw account.
+			let _ = T::CoreStaking::withdraw_unbonded(who.clone(), num_slashing_spans)
```
georgepisaltu, line 540: Maybe we save the return value of this as `stash_killed`...

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		}
+
+		// if we still do not have enough funds to release, abort.
+		ensure!(agent.ledger.unclaimed_withdrawals >= amount, Error::<T>::NotEnoughFunds);
+
+		// claim withdraw from agent. Kill agent if no delegation left.
+		// TODO(ank4n): Ideally if there is a register, there should be an unregister that should
+		// clean up the agent. Need to improve this.
+		if agent.remove_unclaimed_withdraw(amount)?.update_or_kill()? {
+			let _ = frame_system::Pallet::<T>::dec_providers(who).defensive();
```
georgepisaltu, line 553: ... and at least defensive assert that the stash was killed there as well.

georgepisaltu (commented):
Final review tomorrow :grin

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+			origin: OriginFor<T>,
+			delegator: T::AccountId,
+			amount: BalanceOf<T>,
+			num_slashing_spans: u32,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			Self::do_release(&who, &delegator, amount, num_slashing_spans)
+		}
+
+		/// Claim delegated funds that are held in `proxy_delegator` to the claiming delegator's
```
gpestana, line 338:
```suggestion
		/// Claim delegated funds that are held in `delegator` to the claiming delegator's
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		/// account. If successful, the specified funds will be delegated directly from `delegator`
+		/// account to the agent.
+		///
+		/// This can be called by `agent` accounts that were previously a direct `Nominator` with
+		/// [`Config::CoreStaking`] and has some remaining unclaimed delegations.
+		///
+		/// Internally, it moves some delegations from `pxoxy_delegator` account to `delegator`
+		/// account and reapplying the holds.
+		pub fn claim_delegation(
```
gpestana, line 347: How about renaming this to `fn migrate_delegation`? Or else rename `fn do_migrate_delegation` to `fn do_claim_delegation`.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+			ensure!(
+				Delegation::<T>::can_delegate(&delegator, &agent),
+				Error::<T>::InvalidDelegation
+			);
+			ensure!(!Self::is_direct_staker(&delegator), Error::<T>::AlreadyStaking);
+
+			// ensure agent is sane.
+			ensure!(Self::is_agent(&agent), Error::<T>::NotAgent);
+
+			// add to delegation
```
gpestana, line 399:
```suggestion
			// add to delegation.
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	/// Derive a (keyless) pot account from the given agent account and account type.
+	pub(crate) fn sub_account(account_type: AccountType, agent: T::AccountId) -> T::AccountId {
+		T::PalletId::get().into_sub_account_truncating((account_type, agent.clone()))
+	}
+
+	/// Balance of a delegator that is delegated.
```
gpestana, line 422:
```suggestion
	/// Held balance by a delegator.
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+	/// Returns true if who is delegating to an `Agent` account.
+	fn is_delegator(who: &T::AccountId) -> bool {
+		<Delegators<T>>::contains_key(who)
+	}
+
+	/// Returns true if who is already staking on [`Config::CoreStaking`].
+	fn is_direct_staker(who: &T::AccountId) -> bool {
+		T::CoreStaking::status(who).is_ok()
+	}
+
```
gpestana, line 441:
```suggestion
    /// Registers a new agent in the system.
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+	fn do_register_agent(who: &T::AccountId, reward_account: &T::AccountId) {
+		AgentLedger::<T>::new(reward_account).update(who);
+
+		// Agent does not hold balance of its own but this pallet will provide for this to exist.
+		// This is expected to be a keyless account and not created by any user directly so safe.
+		// TODO: Someday if we allow anyone to be an agent, we should take a deposit for
+		// being a delegator.
+		frame_system::Pallet::<T>::inc_providers(who);
+	}
+
```
gpestana, line 451: maybe add rustdocs for this and the remaining methods?

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+	/// Migrates delegation of `amount` from `source` account to `destination` account.
+	fn do_migrate_delegation(
+		source_delegator: &T::AccountId,
+		destination_delegator: &T::AccountId,
+		amount: BalanceOf<T>,
+	) -> DispatchResult {
+		let mut source_delegation =
+			Delegators::<T>::get(source_delegator).defensive_ok_or(Error::<T>::BadState)?;
```
gpestana, line 589: nit but not sure if `BadState` is the best error here.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,640 @@
...
+//! Tests for pallet-delegated-staking.
+
+use super::*;
+use crate::mock::*;
+use frame_support::{assert_noop, assert_ok, traits::fungible::InspectHold};
+use pallet_staking::Error as StakingError;
+use sp_staking::DelegationInterface;
+
+#[test]
+fn create_a_agent_with_first_delegator() {
```
georgepisaltu, line 27:
```suggestion
fn create_an_agent_with_first_delegator() {
```

georgepisaltu (approved):

substrate/frame/delegated-staking/src/lib.rs
Ank4n: Some account can become agent first and reward account later.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: There are two steps for a withdrawal:
- Agent Withdraw: Funds are withdrawn from staking ledger but not claimed yet by delegator.
- Delegator withdraw: This is when funds are actually released in delegator account.

The above represents `Agent Withdraw` which can kill the stash in staking pallet. The delegation staking pallet needs to keep the agent until funds are claimed by a delegator. Which is why we ignore the return value for `withdraw_unbonded`.

These two steps generally happens in sync but not always. If the staking ledger is filled up with `MaxUnlockingChunks`, it will do an implicit withdraw where these calls become async.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,763 @@
...
+		});
+
+		Ok(())
+	}
+
+	fn do_release(
+		who: &T::AccountId,
+		delegator: &T::AccountId,
+		amount: BalanceOf<T>,
+		num_slashing_spans: u32,
```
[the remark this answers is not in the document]
Ank4n: Yeah this needs to be fixed.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: I could check if agent exists as a staker in staking pallet. Though I think we still achieve the same thing.
`dec_providers` would error if there is a consumer reference to it. The consumer reference is cleaned when stash is killed in staking pallet.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: No strong opinions though the way I was thinking, the public function naming indicates what this function does from delegator's point of view. Internally it is a migration of delegated amount.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: Can add. Note though these are private functions and won't show up in the rust docs.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: The pattern I have tried to follow is:
- Public fn does all the sanity checks.
- Internal (pvt) fn assumes the passed parameters are sane.

The delegator checks are already done by the caller and this function assumes it must be a delegator or returns `BadState` defensively. Wdyt?

substrate/frame/delegated-staking/src/tests.rs
Ank4n: Result of all the renaming: `Delegatee` -> `Delegate` -> `Agent`

substrate/frame/delegated-staking/src/mock.rs
Ank4n: Seems `try_state` is gated behind `try_runtime`. And if I enable it for test, test feature is not propagated to dependency crates. :/

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		<Delegators<T>>::insert(key, self)
+	}
+}
+
+/// Ledger of all delegations to an `Agent`.
+///
+/// This keeps track of the active balance of the `Agent` that is made up from the funds that
+/// are currently delegated to this `Agent`. It also tracks the pending slashes yet to be
```
kianenigma, line 82: I thought I commented on this before but I can't find it, maybe elsewhere: 

Pattern of `self -> Self` is strange to me as opposed to simply doing `&mut self`.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,292 @@
...
+	/// Funds for unmigrated `delegator` accounts of the `Agent` are kept here.
+	ProxyDelegator,
+}
+
+/// Information about delegation of a `delegator`.
+#[derive(Default, Encode, Clone, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+pub struct Delegation<T: Config> {
+	/// The target of delegation.
+	pub agent: T::AccountId,
```
kianenigma, line 38:
```suggestion
	agent: T::AccountId,
```

In case you want to enforce creation only through `fn from`. But sure if it matters here.

substrate/frame/delegated-staking/src/types.rs
kianenigma: It was meant to encourage you to change more function signatures to `fn foo(&mut self) -> Result<(), Error>`.

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,292 @@
...
+		amount: BalanceOf<T>,
+	) -> Result<Self, DispatchError> {
+		let new_unclaimed_withdrawals = self
+			.ledger
+			.unclaimed_withdrawals
+			.checked_add(&amount)
+			.defensive_ok_or(ArithmeticError::Overflow)?;
+
+		Ok(Agent {
+			ledger: AgentLedger { unclaimed_withdrawals: new_unclaimed_withdrawals, ..self.ledger },
```
kianenigma, line 201: eg

```
	pub(crate) fn remove_unclaimed_withdraw(
		&mut self,
		amount: BalanceOf<T>,
	) -> Result<(), DispatchError> {
		self
			.ledger
			.total_delegated
			.checked_sub(&amount)
			.defensive_ok_or(ArithmeticError::Overflow)?;
		self
			.ledger
			.unclaimed_withdrawals
			.checked_sub(&amount)
			.defensive_ok_or(ArithmeticError::Overflow)?;

		Ok(())
	}

```

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	/// Get the total stake of agent bonded in [`Config::CoreStaking`].
+	pub(crate) fn bonded_stake(&self) -> BalanceOf<T> {
+		T::CoreStaking::total_stake(&self.key).unwrap_or(Zero::zero())
+	}
+
+	/// Returns true if the agent is bonded in [`Config::CoreStaking`].
+	pub(crate) fn is_bonded(&self) -> bool {
+		T::CoreStaking::stake(&self.key).is_ok()
```
kianenigma, line 237:
```suggestion
	pub(crate) fn remove_slash(&mut self, amount: BalanceOf<T>) {
```

substrate/frame/delegated-staking/src/types.rs
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	/// Returns the reward account registered by the agent.
+	pub(crate) fn reward_account(&self) -> &T::AccountId {
```
kianenigma, line 241:
```suggestion
```

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+//! - Slashes are lazy and hence there could be a period of time when an account can use funds for
+//!   operations such as voting in governance even though they should be slashed.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![deny(rustdoc::broken_intra_doc_links)]
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
```
kianenigma, line 131: tip: move all `use` statements next to one another so that RA format them properly.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+				100
+			),
+			Error::<T>::AlreadyStaking
+		);
+
+		// an existing nominator cannot become agent
+		assert_noop!(
+			DelegatedStaking::register_agent(
+				RawOrigin::Signed(mock::GENESIS_NOMINATOR_ONE).into(),
+				100
```
kianenigma, line 81: What if someone is staked but `Chill`-ed? can they become an agent? 

And in general, once you become an Agent, you are then exepcted to become a staker right?

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+			),
+			Error::<T>::InvalidDelegation
+		);
+
+		// delegator one tries to delegate to agent 2 as well (it already delegates to agent
+		// 1)
+		assert_noop!(
+			DelegatedStaking::delegate_to_agent(
+				RawOrigin::Signed(delegator_one).into(),
+				agent_two,
```
kianenigma, line 191: Would expect to see `fn delegator_restrictions` that ensures existing delegators:

1. cannot delegate to others
2. cannot delegate to non-agents.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+			// update old balance
+			old_reporter_balance = Balances::free_balance(reporter);
+		}
+
+		// nothing to slash anymore
+		assert_eq!(get_agent(&agent).ledger.pending_slash, 0);
+
+		// cannot slash anymore
+		assert_noop!(
+			<DelegatedStaking as DelegationInterface>::delegator_slash(&agent, &350, 1, None),
```
kianenigma, line 270: Suggest checking events in the end of tests to be more pedantic.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+			Error::<T>::NothingToSlash
+		);
+	});
+}
+
+/// Integration tests with pallet-staking.
+mod staking_integration {
+	use super::*;
+	use pallet_staking::RewardDestination;
```
kianenigma, line 279: is there something fundamentally different between this and `create_multiple_delegators`? they can both stay, but I want to know your mental model here.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+			let total_staked =
+				setup_delegation_stake(agent, reward_acc, delegators.clone(), 10, 10);
+
+			// lets go to a new era
+			start_era(2);
+
+			assert!(eq_stake(agent, total_staked, total_staked));
+			// Withdrawing without unbonding would fail.
+			assert_noop!(
+				DelegatedStaking::release_delegation(RawOrigin::Signed(agent).into(), 301, 50, 0),
```
kianenigma, line 341: ?

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+
+			// active stake is now reduced..
+			let expected_active = total_staked - (50 + 100 + 200);
+			assert!(eq_stake(agent, total_staked, expected_active));
+
+			// nothing to withdraw at era 4
+			assert_noop!(
+				DelegatedStaking::release_delegation(RawOrigin::Signed(agent).into(), 305, 50, 0),
+				Error::<T>::NotEnoughFunds
+			);
```
kianenigma, line 364: assert_noop implies this, but okay to be verbose

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+			ledger.update(who);
+		}
+
+		Ok(())
+	}
+
+	/// Bond `amount` to `agent_acc` in [`Config::CoreStaking`].
+	fn do_bond(agent_acc: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+		let agent = Agent::<T>::get(agent_acc)?;
+
```
kianenigma, line 509: might be easier to ask `Agent::get` to always to a withdraw unbonded under the hood.

substrate/frame/delegated-staking/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+		let new_delegation_amount =
+			if let Some(existing_delegation) = Delegation::<T>::get(delegator) {
+				ensure!(&existing_delegation.agent == agent, Error::<T>::InvalidDelegation);
+				existing_delegation
+					.amount
+					.checked_add(&amount)
+					.ok_or(ArithmeticError::Overflow)?
+			} else {
+				amount
+			};
```
kianenigma, line 539:
```suggestion
		defensive_assert_eq!(released, amount, "hold should have been released fully");
```

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+pub type Balance = u128;
+
+parameter_types! {
+	pub static ExistentialDeposit: Balance = 1;
+}
+impl pallet_balances::Config for Runtime {
+	type MaxLocks = ConstU32<128>;
+	type MaxReserves = ();
```
kianenigma, line 69: I believe we have good configs for this as well now.

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+impl delegated_staking::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type PalletId = DelegatedStakingPalletId;
+	type Currency = Balances;
+	type OnSlash = ();
+	type SlashRewardFraction = SlashRewardFraction;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type CoreStaking = Staking;
+}
```
kianenigma, line 152: Not needed?

substrate/frame/delegated-staking/src/mock.rs
kianenigma: Can you elaborate a bit more? 

```
<AllPallets as TryState>::try_state(now).unwrap();  
```

doesn't work?

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -0,0 +1,319 @@
...
+
+#[allow(unused)]
+pub(crate) fn events_since_last_call() -> Vec<crate::Event<Runtime>> {
+	let events = System::events()
+		.into_iter()
+		.map(|r| r.event)
+		.filter_map(
+			|e| if let RuntimeEvent::DelegatedStaking(inner) = e { Some(inner) } else { None },
+		)
+		.collect::<Vec<_>>();
```
kianenigma, line 315: yes, let's use this :D

kianenigma: And you can use this https://paritytech.github.io/polkadot-sdk/master/frame_system/pallet/struct.Pallet.html#method.read_events_for_pallet

kianenigma: And note you have to set the block number to 1 to be able to read events.

substrate/frame/delegated-staking/src/lib.rs
kianenigma: Least assumptions about the call site generally better.

substrate/frame/delegated-staking/src/types.rs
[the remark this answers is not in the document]
kianenigma: I suggest calling this something other than `Proxy` to avoid confusion with the actual `pallet-proxy` stuff.

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -0,0 +1,647 @@
...
+			assert_noop!(
+				DelegatedStaking::register_agent(RawOrigin::Signed(GENESIS_VALIDATOR).into(), 201),
+				Error::<T>::AlreadyStaking
+			);
+		});
+	}
+
+	#[test]
+	fn migration_works() {
```
kianenigma, line 564: How about a set of live tests using `remote_externalities`? that we can then easily add to the fellowship runtimes and test things with? bags-list pallet has similar ones.

kianenigma (approved):
I have some comments left, but all in all looks good. I will start reviewing the 3rd part next. 

Please do a good amount of live testing both for operations (eg. Chopsticks) and migration on the live data as well.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: Anyone directly staked cannot become an agent (they can go through the migration path though.).

Yeah when an account become an agent, the pallet registers the account as a virtual staker in the staking pallet.

Ank4n: Added some extra comments to clarify this.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: Not sure what you mean by fn `delegator_restrictions` but added new assertions to ensure
- delegator cannot delegate to non-agents.
- delegator cannot delegate to other delegators/self.
- agents cannot delegate to other agents or self.

> cannot delegate to others
there were already existing assertions for this.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: What is the question. :D
Don't like the error?

substrate/frame/delegated-staking/src/tests.rs
Ank4n: right, will remove the assertion for stake which I check before the call.
The other two checks are probably ok to stay, shows the ongoing setup and wasn't asserted before.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: Didn't really understand. Could you expand? 
For this pallet, its never really useful to do an implicit withdraw unbonded, but pools do it and it will be handled. Also funds that are withdrawn should be reused for bonding and is tracked in a `unclaimed_withdrawals` register.

substrate/frame/delegated-staking/src/lib.rs
Ank4n: probably commented at the wrong line? Will use `defensive_assert_eq` in places where it is checked.

Ank4n: Did you comment on the wrong line?

substrate/frame/delegated-staking/src/mock.rs
Ank4n: Needed for nomination pool which is the third PR. Removed it from here.

substrate/frame/delegated-staking/src/mock.rs
Ank4n: Yeah I set block number to 1 while building externalities.

substrate/frame/delegated-staking/src/tests.rs
Ank4n: Yeah I have been thinking about having a good set of remote externalities test for the whole integration. Probably will be part of next PR.

Polkadot-Forum:

PR #3905, opened 2024-03-30
Refactor Nomination Pool to support multiple staking strategies

Ank4n:
Third and final PR in the set, closes https://github.com/paritytech/polkadot-sdk/issues/454.

Original PR: https://github.com/paritytech/polkadot-sdk/pull/2680

## Precursors:
- https://github.com/paritytech/polkadot-sdk/pull/3889.
- https://github.com/paritytech/polkadot-sdk/pull/3904.

## Follow up issues/improvements
- https://github.com/paritytech/polkadot-sdk/issues/4404

### Pallet Staking
**Nominator**: An account that directly stakes on `pallet-staking` and nominates a set of validators.
**Stakers**: Common term for nominators and validators.
Virtual Stakers: Same as stakers, but they are keyless accounts and their locks are managed by a pallet external to `pallet-staking`.

### Pallet Delegated Staking
**Agent**: An account that receives delegation from other accounts (delegators) and stakes on their behalf. They are also Virtual Stakers in `pallet-staking` where `pallet-delegated-staking` manages its locks.
**Delegator**: An account that delegates some funds to an agent.

### Pallet Nomination Pools
**Pool account**: Keyless account of a pool where funds are pooled. Members pledge their funds towards the pools. These are going to become `Agent` accounts in `pallet-delegated-staking`.
**Pool Members**: They are individual members of the pool who contributed funds to it. They are also `Delegator` in `pallet-delegated-staking`.

### Multiple Stake strategies

**TransferStake**: The current nomination pool logic can be considered a staking strategy where delegators transfer funds to pool and stake. In this scenario, funds are locked in pool account, and users lose the control of their funds.

**DelegateStake**: With this PR, we introduce a new staking strategy where individual delegators delegate fund to pool. `Delegate` implies funds are locked in delegator account itself. Important thing to note is, pool does not have funds of its own, but it has authorization from its members to use these funds for staking.

We extract out all the interaction of pool with staking interface into a new trait `StakeStrategy`. This is the logic that varies between the above two staking strategies. We use the trait `StakeStrategy` to implement above two strategies: `TransferStake` and `DelegateStake`.

### NominationPool
Consumes an implementation of `StakeStrategy` instead of `StakingInterface`. I have renamed it from `Staking` to `StakeAdapter` to clarify the difference from the earlier used trait.

To enable delegation based staking in pool, Nomination pool can be configured as:
```
type StakeAdapter = pallet_nomination_pools::adapter::DelegateStake<Self, DelegatedStaking>;
```

Note that with the following configuration, the changes in the PR are no-op. 
```
type StakeAdapter = pallet_nomination_pools::adapter::TransferStake<Self, Staking>;
```

## Deployment roadmap
Plan to enable this only in Westend. In production runtimes, we can keep pool to use `TransferStake` which will be no functional change.

Once we have a full audit, we can enable this in Kusama followed by Polkadot.

## TODO
- [x] Runtime level (Westend) migration for existing nomination pools.
- [x] Permissionless call/ pallet::tasks for claiming delegator funds.
- [x] Add/update benches.
- [x] Migration tests.
- [x] Storage flag to mark `DelegateStake` migration and integrity checks to not allow `TransferStake` for migrated runtimes.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5700453

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5704713

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5704866

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5704865

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5704867

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5705004

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5706064

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5707404

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5707556

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5707724

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5707725

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5725677

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2239,13 +2238,18 @@ pub mod pallet {
...
-			let current_era = T::Staking::current_era();
+			let current_era = T::StakeAdapter::current_era();
 
 			let bonded_pool = BondedPool::<T>::get(member.pool_id)
 				.defensive_ok_or::<Error<T>>(DefensiveError::PoolNotFound.into())?;
 			let mut sub_pools =
 				SubPoolsStorage::<T>::get(member.pool_id).ok_or(Error::<T>::SubPoolsNotFound)?;
 
+			if bonded_pool.has_pending_slash() {
+				// apply slash if any before withdraw.
```
georgepisaltu, line 2249: Option 1 from above continued: we could do `<frame_system::Pallet<T>>::register_extra_weight_unchecked(T::WeightInfo::apply_slash())` here in this `if`.

georgepisaltu (commented):
Changes look good. The delegate staking tests need to switch to the new interface to become relevant, but it should be fairly easy as the testing logic should be the same, maybe a couple of extra corner cases here and there. Will re-review once that's done.

substrate/frame/nomination-pools/benchmarking/src/mock.rs
```diff
@@ -166,7 +166,7 @@ impl pallet_nomination_pools::Config for Runtime {
 	type RewardCounter = FixedU128;
 	type BalanceToU256 = BalanceToU256;
 	type U256ToBalance = U256ToBalance;
-	type Staking = Staking;
+	type StakeAdapter = pallet_nomination_pools::adapter::TransferStake<Self, Staking>;
```
Ank4n, line 169: Assuming DelegateStake will be worse case, should I becch against that or is there a good way to bench with different configurations for different runtime?

georgepisaltu: For the `SubstrateWeight` I think it's ok to bench against whatever adapter. Actual runtimes will bench this pallet again with their specific runtime config for `StakeAdapter`, so as long as the benchmark functions hit the right codepaths regardless of adapters, it's ok.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5748481

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5778803

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5833502

substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -107,6 +107,98 @@ pub mod unversioned {
...
+	/// [`adapter::StakeStrategyType::Delegate`].
+	///
+	/// Note: This only migrates the pools, the members are not migrated. They can use the
+	/// permission-less [`Pallet::claim_delegation()`] to migrate their funds.
+	///
+	/// This migration does not break any existing pool storage item, does not need to happen in any
+	/// sequence and hence can be applied unversioned on a production runtime.
+	pub struct DelegationStakeMigration<T>(sp_std::marker::PhantomData<T>);
+
+	// FIXME(ank4n) convert to MBM.
```
Ank4n, line 121: There are currently 224 pools in polkadot and 179 in Kusama so its probably totally fine to do this in single block.

Ank4n, line 121: There are currently 224 pools in polkadot and 179 in Kusama so its probably totally fine to do this in single block.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5859790

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5864723

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5868219

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5868523

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5868836

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5881837

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5882230

substrate/frame/nomination-pools/benchmarking/src/lib.rs
```diff
@@ -844,6 +883,146 @@ frame_benchmarking::benchmarks! {
 		assert!(&Pools::<T>::check_ed_imbalance().is_ok());
 	}
 
+	apply_slash {
```
Ank4n, line 886: @georgepisaltu I am not super happy with the benchmarks since some code paths are never reached with `TransferStake` as adapter. Let me know if you have some better ideas to manage this (or possibly refactor adapter setup in the pallet).

substrate/frame/nomination-pools/benchmarking/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 //! Benchmarks for the nomination pools coupled with the staking and bags list pallets.
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![recursion_limit = "256"]
```
georgepisaltu, line 21: Where is the default limit of `128` hit? :joy:

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -549,9 +551,18 @@ impl<T: Config> PoolMember<T> {
...
+	/// Worst case, iterates over [`TotalUnbondingPools`] member unbonding pools to calculate member
+	/// balance.
+	pub fn total_balance(&self) -> BalanceOf<T> {
+		let maybe_pool = BondedPool::<T>::get(self.pool_id);
+		// this internal function is always called with a valid pool id.
+		if maybe_pool.is_none() {
+			defensive!("pool should exist; qed");
+			return Zero::zero();
+		}
+		let pool = maybe_pool.expect("checked pool is not none; qed");
```
georgepisaltu, line 564: Cleaner than using `expect` IMO.
```suggestion
		let pool = match BondedPool::<T>::get(self.pool_id) {
			Some(p) => p,
			None => {
				// this internal function is always called with a valid pool id.
				defensive!("pool should exist; qed");
				return Zero::zero();
			}
		};
```

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2246,13 +2253,23 @@ pub mod pallet {
...
 				.defensive_ok_or::<Error<T>>(DefensiveError::PoolNotFound.into())?;
 			let mut sub_pools =
 				SubPoolsStorage::<T>::get(member.pool_id).ok_or(Error::<T>::SubPoolsNotFound)?;
 
+			let slash_weight = if bonded_pool.has_pending_slash() {
+				// apply slash if any before withdraw.
+				match Self::do_apply_slash(&member_account, None) {
+					Ok(_) => T::WeightInfo::apply_slash(),
+					Err(_) => T::WeightInfo::apply_slash_fail(),
+				}
```
georgepisaltu, line 2268: We should probably exit this function if there is a slash to apply, but we failed to do so. The `do_apply_slash` call has underlying effects in `pallet-delegated-staking`. Right now we're ignoring the error itself but keeping track of weight.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -3302,6 +3381,36 @@ impl<T: Config> Pallet<T> {
...
+	/// Slash member against the pending slash for the pool.
+	fn do_apply_slash(
+		member_account: &T::AccountId,
+		reporter: Option<T::AccountId>,
+	) -> DispatchResult {
+		// calculate points to be slashed.
+		let member =
+			PoolMembers::<T>::get(&member_account).ok_or(Error::<T>::PoolMemberNotFound)?;
+
+		let pool_account = Pallet::<T>::create_bonded_account(member.pool_id);
```
georgepisaltu, line 3393: A better name for this fn would be `generate_bonded_account_id` or something like that. Create account sounds like funding it or something similar.

georgepisaltu (approved):

georgepisaltu (commented):
Didn't mean to approve *yet*, sorry :grin:

substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -107,6 +107,109 @@ pub mod unversioned {
...
+				return T::DbWeight::get().reads_writes(1, 0)
+			}
+
+			let mut migrate_count: u32 = 0;
+			let mut fail_count: u32 = 0;
+			BondedPools::<T>::iter().for_each(|(id, _inner)| {
+				if Pallet::<T>::migrate_to_delegate_stake(id).is_ok() {
+					migrate_count.saturating_inc();
+				} else {
+					fail_count.saturating_inc();
```
gpestana, line 133: if a few pools migrate for some reason, what are the side effects of having them? the `StakeStrategyType` is global so that may be an issue, no? Have you considered keeping the `StakeStrategyType` as a per-pool setting?

gpestana: > what are the side effects of having them? 

I guess the `post_upgrade` will fail as the failed pool balances will not be 0. So it should not happen in the runtime.

gpestana (commented):
Looking good, left a few minor comments and questions.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: If it fails it means the member has no slash to be applied right? Or do you see a codepath where there is slash to apply but fails and continues to withdraw?

substrate/frame/nomination-pools/src/lib.rs
Ank4n: Yeah it confuses me too.

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,362 @@
...
+pub enum StakeStrategyType {
+	/// The older staking strategy where funds are transferred to pool account and staked.
+	Transfer,
+	/// The newer staking strategy where funds are delegated to pool account and staked.
+	Delegate,
+}
+
+impl Default for StakeStrategyType {
+	fn default() -> Self {
+		// default to the older strategy before migration.
```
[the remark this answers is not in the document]
Ank4n: could probably refer to as v0 and v1 :D

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	/// See [`StakingInterface::chill`].
+	fn chill(pool_account: &Self::AccountId) -> DispatchResult {
+		Self::CoreStaking::chill(pool_account)
+	}
+
+	/// Pledge `amount` towards `pool_account` and update the pool bond. Also see
+	/// [`StakingInterface::bond`].
+	fn pledge_bond(
```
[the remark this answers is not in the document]
Ank4n: Great question. This function basically does two things: delegate/transfer and bond. I was trying to find a word in the dictionary 😂 that could generalise the actions delegate and transfer. Here the delegator pledges the fund towards the pool account either by transferring or delegating. :D

If we call it only bond, it is also misleading as its doing more than just bond. 

But I am not sure of any of these names so I am happy to apply your suggestion if you stand by your opinion after listening to my rant. 😄

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2325,20 +2344,20 @@ pub mod pallet {
 
 				if member_account == bonded_pool.roles.depositor {
 					Pallet::<T>::dissolve_pool(bonded_pool);
-					None
+					Weight::default()
```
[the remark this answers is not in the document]
Ank4n: This is just a refactor and keeps the same weight logic as before. 

But good question. I am guessing we want to incentivise dissolving of pool so its free :D. It would remove items from storage so its a good thing for the chain.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2842,58 @@ pub mod pallet {
...
+		/// slash to be applied, caller may be rewarded with the part of the slash.
+		#[pallet::call_index(23)]
+		#[pallet::weight(T::WeightInfo::apply_slash())]
+		pub fn apply_slash(
+			origin: OriginFor<T>,
+			member_account: AccountIdLookupOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+			let member_account = T::Lookup::lookup(member_account)?;
+			Self::do_apply_slash(&member_account, Some(who))?;
```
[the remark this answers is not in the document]
Ank4n: Yeah the intention was to make this function permissionless where anybody can call and get rewarded from part of the slash. This would incentivise applying of pending slashes in the pool.

Happy to consider other approaches?

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2842,58 @@ pub mod pallet {
...
+
+		/// Allows a pool member to claim their delegation in their own account.
+		///
+		/// This is a permission-less call and refunds any fee if claim is successful.
+		///
+		/// If the pool has migrated to delegation based staking, the staked tokens of pool members
+		/// can be moved and held in their own account. See [`adapter::DelegateStake`]
+		#[pallet::call_index(24)]
+		#[pallet::weight(T::WeightInfo::claim_delegation())]
+		pub fn claim_delegation(
```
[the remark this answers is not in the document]
Ank4n: Yeah I am also fighting between these two terms. From a member's pov, they are claiming delegation that they transferred to the pool back into their own account. But agree we can try to use same term everywhere.

If you like migrate, I will make the changes everywhere.

substrate/frame/nomination-pools/src/migration.rs
Ank4n: There is a test that checks if a pool fails to migrate, the members still cannot withdraw. (since they don't exist in delegators map in delegation staking pallet).

But this is the refactor we were discussing last week. To recap, I am contemplating whether to make another refactor to:
- Keep a storage map per pool if the migration has happened.
- Make pallet support both `TransferStake` and `DelegateStake` at the same time. We could even separate out `DelegationInterface` and pass it separately providing an impl for `()` if the runtime wants to support only `TransferStake`. Have to try it out (in a separate branch) to see how this ends up.
 - We avoid doing one time migration, but pool can permissionlessly migrate to `DelegateStake` and the appropriate migration flag is updated.
 
## Reasons not to do this refactor?
- We might want to just clean up `TransferStake` completely at some point. If we can ensure that there are no footguns that would corrupt the state, we can argue that the code is fine in its current form and we are probably over engineering.

I will make a note of possible footguns in the main description and if it can lead to issues. The only big footgun I can think off top of my head is if the provided adapter to the pallet is `DelegateStake` but no migration was done to convert pools to `DelegateStake` (or some pools failed to migrate). This would lead to all member calls failing for existing pools since these accounts don't exist in `pallet-delegation-staking`. (scenario covered in test)

@georgepisaltu your opinion here might be valuable too.

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,362 @@
...
+/// Useful for determining current staking strategy of a runtime and enforce integrity tests.
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, RuntimeDebugNoBound, PartialEq)]
+pub enum StakeStrategyType {
+	/// The older staking strategy where funds are transferred to pool account and staked.
+	Transfer,
+	/// The newer staking strategy where funds are delegated to pool account and staked.
+	Delegate,
+}
+
+impl Default for StakeStrategyType {
```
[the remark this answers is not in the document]
Ank4n: Yeah fair, will remove it.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -545,9 +547,17 @@ impl<T: Config> PoolMember<T> {
 
 	/// Total balance of the member, both active and unbonding.
 	/// Doesn't mutate state.
-	#[cfg(any(feature = "try-runtime", feature = "fuzzing", test, debug_assertions))]
```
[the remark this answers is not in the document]
Ank4n: The function is part of PoolMember. I think it is fair to assume the pool and the member always exist, hence the function is infallible.

substrate/frame/nomination-pools/src/adapter.rs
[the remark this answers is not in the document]
Ank4n: removed default completely and using the storage only to mark migration.

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,362 @@
...
+/// [`DelegateStake`] for more detail.
+pub trait StakeStrategy {
+	type Balance: frame_support::traits::tokens::Balance;
+	type AccountId: Clone + sp_std::fmt::Debug;
+	type CoreStaking: StakingInterface<Balance = Self::Balance, AccountId = Self::AccountId>;
+
+	/// The type of staking strategy.
+	fn strategy_type() -> StakeStrategyType {
+		StakeStrategyType::default()
+	}
```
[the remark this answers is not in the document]
Ank4n: Removed

substrate/frame/nomination-pools/src/migration.rs
Ank4n: There is a test that checks if a pool fails to migrate, the members still cannot withdraw. (since they don't exist in delegators map in delegation staking pallet).

I have refactored the storage to be used only to make sure a migration of strategy type has occurred and to avoid redoing the migration.

Ank4n:
bot bench polkadot-pallet --pallet=pallet_nomination_pools

Ank4n:
bot cancel

Ank4n:
bot bench polkadot-pallet --pallet=pallet_nomination_pools --runtime=westend

Ank4n:
bot bench substrate-pallet --pallet=pallet_nomination_pools

substrate/frame/delegated-staking/src/tests.rs
```diff
@@ -638,3 +639,494 @@ mod staking_integration {
...
+			assert!(!Agents::<T>::contains_key(Pools::create_bonded_account(pool_id)));
+			assert!(!Delegators::<T>::contains_key(creator));
+			for i in 300..310 {
+				assert!(!Delegators::<T>::contains_key(i));
+			}
+		});
+	}
+
+	#[test]
+	fn pool_slashed() {
```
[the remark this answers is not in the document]
Ank4n: I was thinking what would happen when all amount is slashed. I think accounts don't get deleted automatically but there is a `Staking::reap_stash` to clean up data. We probably need something similar in `Delegation` pallet. Will make a test.

substrate/frame/nomination-pools/src/lib.rs
georgepisaltu: The `has_pending_slash` check is duplicated in `do_apply_slash`.

substrate/frame/nomination-pools/src/lib.rs
georgepisaltu: As discussed, let's use the `apply_slash_fail` when we get a `NothingToSlash` error and exit the function on any other error, as that would be bad state.

substrate/frame/nomination-pools/benchmarking/src/inner.rs
```diff
@@ -838,6 +876,146 @@ frame_benchmarking::benchmarks! {
...
+		for i in 0..T::MaxUnbonding::get() {
+			pallet_staking::CurrentEra::<T>::put(i + 2); // +2 because we already set the current era to 1.
+			assert!(Pools::<T>::unbond(RuntimeOrigin::Signed(joiner.clone()).into(), joiner_lookup.clone(), min_join_bond).is_ok());
+		}
+
+		pallet_staking::CurrentEra::<T>::put(T::MaxUnbonding::get() + 3);
+		whitelist_account!(joiner);
+
+	}: {
+		assert!(Pools::<T>::apply_slash(RuntimeOrigin::Signed(joiner.clone()).into(), joiner_lookup.clone()).is_err());
```
georgepisaltu, line 969: nit: Could do an `assert!(matches!(err, NothingToSlash))`.

georgepisaltu (approved):
LGTM, pending all of the yet unresolved comments.

Polkadot-Forum:

https://forum.polkadot.network/t/paritytech-update-for-april-2024/7646/3

substrate/frame/nomination-pools/src/lib.rs
gpestana: Yeah I agree with this, it's pretty `create` confusing here.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2850,58 @@ pub mod pallet {
...
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+			let member_account = T::Lookup::lookup(member_account)?;
+			Self::do_apply_slash(&member_account, Some(who))?;
+
+			// If successful, refund the fees.
+			Ok(Pays::No.into())
+		}
+
+		/// Allows a pool member to migrate their delegation from pool account in their own account.
```
gpestana, line 2872:
```suggestion
		/// Migrates delegated funds from `member_account` from a pool account to their account.
```

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2934,6 +3014,13 @@ impl<T: Config> Pallet<T> {
 		T::PalletId::get().into_sub_account_truncating((AccountType::Bonded, id))
 	}
 
+	pub fn migrate_to_delegate_stake(id: PoolId) -> DispatchResult {
+		T::StakeAdapter::migrate_nominator_to_agent(
+			&Self::create_bonded_account(id),
```
gpestana, line 3019: At this point, I think it would be useful to rename `fn create_bonded_account` to `fn generate_bonded_account_id` or similar. It is already confusing and it gets worse as its used more.

Same for `fn create_reward_acount`.

Perhaps something like this?

```rust
pub fn generate_account_id(id: PoolId, type: AccountType) -> T::AccountId {
  T::PalletId::get().into_sub_account_truncating((type, id))
}
```

substrate/frame/nomination-pools/src/migration.rs
gpestana: > I have refactored the storage to be used only to make sure a migration of strategy type has occurred and to avoid redoing the migration.

So basically, the pools that failed to migrate will be active but the members cannot migrate to the new strategy right? Will the failed pools works as expected with the transfer stake strat a i.e. can we have different strategies working in at the same time? (I think that would be great!)

substrate/frame/delegated-staking/src/mock.rs
```diff
@@ -129,7 +131,7 @@ impl pallet_staking::Config for Runtime {
 	type NominationsQuota = pallet_staking::FixedNominationsQuota<16>;
 	type MaxUnlockingChunks = ConstU32<10>;
 	type MaxControllersInDeprecationBatch = ConstU32<100>;
-	type EventListeners = DelegatedStaking;
+	type EventListeners = (Pools, DelegatedStaking);
```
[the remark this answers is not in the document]
Ank4n: Ideally order should not matter, right? Or do you see a scenario where it matters?

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,374 @@
...
+		who: &T::AccountId,
+		pool_account: &Self::AccountId,
+		reward_account: &Self::AccountId,
+		amount: BalanceOf<T>,
+		bond_type: BondType,
+	) -> DispatchResult {
+		match bond_type {
+			BondType::Create => {
+				// first bond
+				T::Currency::transfer(who, pool_account, amount, Preservation::Expendable)?;
```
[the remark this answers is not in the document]
Ank4n: Why should we remove this line? 😕

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,374 @@
...
+		Zero::zero()
+	}
+
+	fn pledge_bond(
+		who: &T::AccountId,
+		pool_account: &Self::AccountId,
+		reward_account: &Self::AccountId,
+		amount: BalanceOf<T>,
+		bond_type: BondType,
+	) -> DispatchResult {
```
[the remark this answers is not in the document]
Ank4n: ahh sorry get it now. But there is a subtle difference. We do not want to kill the member account when they join a pool (`BondType::Extra`) but the creator account is `expendable`. This is also just to keep backward compatibility.

substrate/frame/nomination-pools/src/adapter.rs
```diff
@@ -0,0 +1,374 @@
...
+		amount: BalanceOf<T>,
+		_num_slashing_spans: u32,
+	) -> DispatchResult {
+		T::Currency::transfer(pool_account, &who, amount, Preservation::Expendable)?;
+
+		Ok(())
+	}
+
+	fn has_pending_slash(_: &Self::AccountId) -> bool {
+		// for transfer stake strategy, slashing is greedy and never deferred.
```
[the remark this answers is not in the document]
Ank4n: This will be legit called while the pallet is using `TransferStake` adapter. So shouldn't really use defensive here.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2232,7 +2235,11 @@ pub mod pallet {
 		///
 		/// # Note
 		///
-		/// If the target is the depositor, the pool will be destroyed.
+		/// - If the target is the depositor, the pool will be destroyed.
+		/// - If the pool has any pending slash, we also try to slash the member before letting them
+		/// withdraw. This calculation can be expensive and is only defensive. In the ideal
```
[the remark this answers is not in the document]
Ank4n: It is actually not bad (max 32 iterations, single storage read). I probably exaggerated in the docs and will remove them. Don't think it should be an issue for staking parachain as well.

We shouldn't really fail as the (not so) expensive part was checking whether there is a pending slash for the member. If we have already checked, we might as well just apply it. Other way is to fail if the pool has any pending slash but that would affect every member of the pool even if just one member has a pending slash. 

Tldr: IMO I think this is fine.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: Refactored 
- `create_bonded_account` --> `generate_bonded_account`
- `create_reward_account` --> `generate_reward_account`

substrate/frame/nomination-pools/src/migration.rs
Ank4n: > Will the failed pools works as expected with the transfer stake strat 

No they will be kinda blocked (won't be able to do anything). I did try to support both adapters at the same time and have it configurable per pool, but honestly I felt it is adding too much complexity for something we don't really want to do (support both adapters at the same time). Ideal path is, we completely remove `TransferStake` at some point since `DelegateStake` is (unarguably) better than the previous strategy.

Also, if we do the try runtime upgrade tests properly, I think we should feel confident nothing would fail. The only thing we should do defensively (imo) here is to make sure nothing harmful happens if there are some failed migrations in the unexpected case.

substrate/frame/nomination-pools/src/lib.rs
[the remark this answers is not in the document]
gpestana: Sounds good!

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -1257,28 +1254,17 @@ impl<T: Config> BondedPool<T> {
...
-			// found, we exit early.
-			BondType::Later => T::Staking::bond_extra(&bonded_account, amount)?,
-		}
+		T::StakeAdapter::pledge_bond(
+			who,
+			&self.bonded_account(),
+			&self.reward_account(),
+			amount,
+			ty,
+		)?;
```
kianenigma, line 1267: nice naming :p

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -1750,6 +1738,13 @@ pub mod pallet {
 	pub type ClaimPermissions<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, ClaimPermission, ValueQuery>;
 
+	/// Records last migration of stake strategy. Stores the migrated strategy type.
+	///
+	/// This is used to ensure that the stake strategy is only migrated once.
+	#[pallet::storage]
+	pub type StrategyMigration<T: Config> =
```
kianenigma, line 1745: Okay I have not seen the migration, but can you elaborate why this is needed if the migration is a one-off MBM? Or is it not?

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2246,23 +2252,42 @@ pub mod pallet {
...
 				PoolMembers::<T>::get(&member_account).ok_or(Error::<T>::PoolMemberNotFound)?;
-			let current_era = T::Staking::current_era();
+			let current_era = T::StakeAdapter::current_era();
 
 			let bonded_pool = BondedPool::<T>::get(member.pool_id)
 				.defensive_ok_or::<Error<T>>(DefensiveError::PoolNotFound.into())?;
 			let mut sub_pools =
 				SubPoolsStorage::<T>::get(member.pool_id).ok_or(Error::<T>::SubPoolsNotFound)?;
 
+			let slash_weight =
```
kianenigma, line 2262: quesiton for you to think about: how as this not an issue in the past? 

probable answer: because slashes were immediate, and now they are more lazy?

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2849,58 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Apply a pending slash on a member.
+		///
+		/// This call can be dispatched permissionlessly (i.e. by any account). If the member has
+		/// slash to be applied, caller may be rewarded with the part of the slash.
```
kianenigma, line 2856: Would be great to write a simple PJS/python bot for this (you can add it to `polkadot-scripts`). And we should simulate slashes in westend.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2849,58 @@ pub mod pallet {
...
+
+		/// Migrates delegated funds from the pool account to the `member_account`.
+		///
+		/// This is a permission-less call and refunds any fee if claim is successful.
+		///
+		/// If the pool has migrated to delegation based staking, the staked tokens of pool members
+		/// can be moved and held in their own account. See [`adapter::DelegateStake`]
+		#[pallet::call_index(24)]
+		#[pallet::weight(T::WeightInfo::migrate_delegation())]
+		pub fn migrate_delegation(
```
kianenigma, line 2879: same here.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2849,58 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			member_account: AccountIdLookupOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let _caller = ensure_signed(origin)?;
+
+			let member_account = T::Lookup::lookup(member_account)?;
+			let member =
+				PoolMembers::<T>::get(&member_account).ok_or(Error::<T>::PoolMemberNotFound)?;
+			let pool_contribution = member.total_balance();
```
kianenigma, line 2888: Do we have tests for the scenario of: members is partially/fully unbonding and calls this?

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2849,58 @@ pub mod pallet {
...
+			ensure!(pool_contribution >= MinJoinBond::<T>::get(), Error::<T>::MinimumBondNotMet);
+
+			let delegation = T::StakeAdapter::member_delegation_balance(&member_account);
+			ensure!(pool_contribution > delegation, Error::<T>::NoDelegationToClaim);
+
+			let diff = pool_contribution.defensive_saturating_sub(delegation);
+			T::StakeAdapter::migrate_delegation(
+				&Pallet::<T>::generate_bonded_account(member.pool_id),
```
kianenigma, line 2896: these APIs are tricky in that you might mistakenly swap params.. but okay.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2930,12 +3009,19 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Create the main, bonded account of a pool with the given id.
-	pub fn create_bonded_account(id: PoolId) -> T::AccountId {
+	pub fn generate_bonded_account(id: PoolId) -> T::AccountId {
 		T::PalletId::get().into_sub_account_truncating((AccountType::Bonded, id))
 	}
 
+	pub fn migrate_to_delegate_stake(id: PoolId) -> DispatchResult {
```
kianenigma, line 3016:
```suggestion
	pub(crate) fn migrate_to_delegate_stake(id: PoolId) -> DispatchResult {
```

substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -107,6 +107,115 @@ pub mod unversioned {
...
+	impl<T: Config> OnRuntimeUpgrade for DelegationStakeMigration<T> {
+		fn on_runtime_upgrade() -> Weight {
+			if StrategyMigration::<T>::get() == Some(adapter::StakeStrategyType::Delegate) {
+				log!(info, "`pools::migration::unversioned::DelegationStakeMigration` already migrated and can be removed.");
+				return T::DbWeight::get().reads_writes(1, 0)
+			}
+
+			let mut migrate_count: u32 = 0;
+			let mut fail_count: u32 = 0;
+			BondedPools::<T>::iter().for_each(|(id, _inner)| {
```
kianenigma, line 130: This can still be attacked by creating many pools right before the migration... 

I don't think there is an easy and sound way to fix it. What comes to mind is `.take(DEFENSIVE_MAX_POOLS_I_KNOW)`. So we don't iterate too much and the chain won't brick, but then you are stuck in a situation when you have some pools migrated and som unmigrated.. 

This is only safe if there was a `MaxPools`, but we don't have one set in Polkadot. 

It seems to be that a one-off MBM that does pools + members is a bit more work in advance, but less work on agregate.

kianenigma: Unless, if: 

1. it is okay to have some pools migrated and some not
2. add a permissionelss transaction to also migrate any leftover pools 
3. run this migration with a `.take(SAFE_BOUND)`

So this migration is just a best effort, and even pools should be migrated lazily.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: It is single block migration of pools. Currently, there are 229 pools in polkadot and around same or less in Kusama so single block migration seems okay. It is also an unversioned migration (each prod runtime can follow a different migration sequence). 

Probably not super bad but this tries to avoid iterating over these 229 pools again (similar to version checks). Do you suggest to remove it?

substrate/frame/nomination-pools/src/lib.rs
Ank4n: The slashes were system initiated before. It still is for the pool, but yeah we did not had lazy slashes before.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2246,23 +2252,42 @@ pub mod pallet {
...
 				SubPoolsStorage::<T>::get(member.pool_id).ok_or(Error::<T>::SubPoolsNotFound)?;
 
+			let slash_weight =
+				// apply slash if any before withdraw.
+				match Self::do_apply_slash(&member_account, None) {
+					Ok(_) => T::WeightInfo::apply_slash(),
+					Err(e) => {
+						let no_pending_slash: DispatchResult = Err(Error::<T>::NothingToSlash.into());
+						// This is an expected error. We add appropriate fees and continue withdrawal.
+						if Err(e) == no_pending_slash {
```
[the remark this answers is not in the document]
Ank4n: There are two errors that are possible here:
- Nothing to Slash: this is perfectly reasonable error and we want to continue with withdraw.
- Any other error: This is unexpected, and if for whatever reason we cannot apply slash, its better to abort and not allow withdraw.

The second scenario is defensive and ideally should never happen.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: Yes absolutely. We should have this bot running all the time as well. Since it gets part of the slash as reward, we should probably see other bots from the community as well.

Ank4n: I have mentioned this in the followup issues, will create a separate issue for it at some point.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: There is a test for `migrate_delegation` in `nomination-pools/test-delegate-stake/src/lib.rs` but probably could do with adding more members to the pool in different unbonding states.

Will add.

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -2823,6 +2849,58 @@ pub mod pallet {
...
+		) -> DispatchResultWithPostInfo {
+			let _caller = ensure_signed(origin)?;
+
+			let member_account = T::Lookup::lookup(member_account)?;
+			let member =
+				PoolMembers::<T>::get(&member_account).ok_or(Error::<T>::PoolMemberNotFound)?;
+			let pool_contribution = member.total_balance();
+			ensure!(pool_contribution >= MinJoinBond::<T>::get(), Error::<T>::MinimumBondNotMet);
+
+			let delegation = T::StakeAdapter::member_delegation_balance(&member_account);
```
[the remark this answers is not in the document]
Ank4n: The first time it is called for a member, `delegation` would be zero and `pool_contribution` > 0. 
After this is called, all delegation would be migrated. From next time it will always be `pool_contribution` <= `delegation` (less if there is an unapplied slash), and the call would fail.

Does it make sense?

substrate/frame/nomination-pools/src/lib.rs
Ank4n: Yeah. I think we talked about wrapping them in a new type. It should be a good improvement, will make the change.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: I use them in the benchmarking crate as well :)

kianenigma (approved):

kianenigma (approved):
As usual I have some comments about further testing and such, but all in all looks good 🚀

So long as the testing suggestions are done, I'd be happy to see this merged and tested in Westend and other test nets, ready for Polkadot. 

https://github.com/paritytech/polkadot-sdk/pull/3905#discussion_r1604934619 is the only main grumble left.

substrate/frame/nomination-pools/src/migration.rs
Ank4n: Yeah good point. Was discussing with Goncalo as well about having a permissionless extrisinsic for the unlikely scenario where some pools fail to migrate (should really not happen though). But your suggestion does play well with it.

Practically though, we need 500 DOT to create a pool so it would be a very expensive attack which will only cause the migration block to be longer. :D

substrate/frame/nomination-pools/src/lib.rs
kianenigma: see my other comment about migration, otherwise this can be closed.

substrate/frame/nomination-pools/src/lib.rs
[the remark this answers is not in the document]
kianenigma: I would expect to see a `ensure!(delegtaion != zero, ...)` and that is it. 

What you have here also makes sense, but gives the false impression that a single member can be passed to this function multiple times.

substrate/frame/nomination-pools/fuzzer/src/call.rs
```diff
@@ -306,7 +306,7 @@ fn main() {
 					BondedPools::<T>::iter().for_each(|(id, _)| {
 						let amount = random_ed_multiple(&mut rng);
 						let _ =
-							Balances::deposit_creating(&Pools::create_reward_account(id), amount);
+							Balances::deposit_creating(&Pools::generate_reward_account(id), amount);
```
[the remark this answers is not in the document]
Ank4n: Added in the followup issue. 👍

substrate/frame/nomination-pools/src/lib.rs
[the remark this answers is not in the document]
Ank4n: I assume you meant `delegation == zero` so this can be called only once per pool member.

Have made the change.

substrate/frame/nomination-pools/src/migration.rs
```diff
@@ -118,51 +118,66 @@ pub mod unversioned {
 	/// sequence and hence can be applied unversioned on a production runtime.
 	pub struct DelegationStakeMigration<T>(sp_std::marker::PhantomData<T>);
 
+	/// We iterate over a maximum of `MAX_POOLS_BOUND` pools in a single block. At the time of
+	/// migration, there are currently less than `MAX_POOLS_BOUND` pools in the system. But if
+	/// there are more pools in the future, the remaining pools can be migrated via the
+	/// permission-less extrinsic [`Call::migrate_pool_to_delegate_stake`].
+	const MAX_POOLS_BOUND: u32 = 250;
```
kianenigma, line 125: sounds good. We can even make this a parameter of `DelegationStakeMigration`. Polkadot has 230 pools now, so pretty close.

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-digest-13-jun-2024/8598/1

Polkadot-Forum:

substrate/frame/nomination-pools/src/weights.rs
```diff
@@ -1249,9 +1330,56 @@ impl WeightInfo for () {
...
+		Weight::from_parts(15_004_000, 3702)
+			.saturating_add(RocksDbWeight::get().reads(1_u64))
+	}
+	fn pool_migrate() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 759_000 picoseconds.
+		Weight::from_parts(819_000, 0)
+	}
```
gui1117, line 1364: how does a pool migrate read and write nothing?

also we run the benchmark on master and `apply_slash` `apply_slash_fail`, `pool_migrate` and `migrate_delegation` all read and write nothing. when run against kitchen sink. I think something is wrong no?

https://github.com/paritytech/polkadot-sdk/pull/6418/commits/4dfe581539971010fb462ae1987cdb0a2a2ff734#diff-de1639ae973d95b58521e7385de22a790c15c6a2000b2034bd8f10f5e92b33c5

Ank4n: Pool has two strategies, the initial one `TransferStake` and the new `DelegateStake`. Unless it is set to `DelegateStake`, we don't allow migrating (there is nothing to migrate), or slashing (In TS slashing is greedy, ie whole pool is slashed instantly). Since prod is still TS, I left kitchensink with the same setup.

The new setup was tested in Westend and there the weight should have writes.

gui1117: ok thank you

Issue #3906, opened 2024-03-30
Regular connectivity reports make log reading more difficult


stakeworld:
Since #3677 a connectivity report is printed every 10 minutes in the logs, which is helpfull for debugging but makes it more difficult for node operators to filter out relevant data, since the report has a generic structure and is variable in length. 

Would it possible to make this optional or at least more easy to filter out?

> Mar 30 10:48:26 polkadot[395740]: 2024-03-30 10:48:26 Connectivity Report connected_ratio=98 absolute_connected=990 absolute_resolved=1005 unconnected_authorities=
> Mar 30 10:48:26 polkadot[395740]: GLYecbdgTwfpikpk9e5qUcXzVUHh6yxtjqgo9UV2LW3AA2q:
> Mar 30 10:48:26 polkadot[395740]:   /ip4/54.228.127.159/tcp/30333/p2p/12D3KooWJWf9dUdavb9yHoa2gVhY7os8P3ykQXa4yUjxLAx7sZb9
> Mar 30 10:48:26 polkadot[395740]:   /ip4/18.203.162.92/tcp/30333/p2p/12D3KooWJWf9dUdavb9yHoa2gVhY7os8P3ykQXa4yUjxLAx7sZb9
> Mar 30 10:48:26 polkadot[395740]: DpJXKC6nVJ5wQEwB3Qmi4sBKiWi5RCjfLj2YNUwjaTKBRnb:
> Mar 30 10:48:26 polkadot[395740]:   /ip4/104.129.128.95/tcp/30333/p2p/12D3KooWLKxaHSXgYtS7o7WrZt2XU5ThhVsd1sUr2k7hLSWY7i3A
> Mar 30 10:48:26 polkadot[395740]: EoAd1fr2Lcy7n8Bakg41ThseRTQLhEffM9a4oc6uiZeunEb:
> Mar 30 10:48:26 polkadot[395740]:   /ip4/52.60.56.141/tcp/30333/p2p/12D3KooWHk8vbVMHDYAoCtZVVwiUsnBRNEoaZKgCRsdQh7ihKJqB
> Mar 30 10:48:26 polkadot[395740]:   /ip4/52.60.255.89/tcp/30333/p2p/12D3KooWHk8vbVMHDYAoCtZVVwiUsnBRNEoaZKgCRsdQh7ihKJqB
> etc.

PR #3913, opened 2024-03-30
Revert log level changes

bkchr:
Closes: https://github.com/paritytech/polkadot-sdk/issues/3906

### DIFF
--- a/polkadot/node/network/gossip-support/src/lib.rs
+++ b/polkadot/node/network/gossip-support/src/lib.rs
@@ -508,7 +508,7 @@ where
 			);
 		}
 		let pretty = PrettyAuthorities(unconnected_authorities);
-		gum::info!(
+		gum::debug!(
 			target: LOG_TARGET,
 			?connected_ratio,
 			?absolute_connected,


ggwpez (approved):

liamaharon (approved):

sandreim (approved):

alexggh:
Did this cause any problem @bkchr? 

If not why wouldn't we want node operator understanding if their validators are properly connected or not

I guess we kind of disagree on this and this https://github.com/paritytech/polkadot-sdk/pull/3677 the fact that this information is not relevant for nodes.
> Debug output is not to be printed using info logs. If people come to you complaining, you should either tell them which cli flags to change or come up for example with a dedicated RPC endpoint that generates debug information.

Let me try to see if I can change your mind :), at least on my perspective part of operating a system with minimal downtime is the ability to understand what happened in production after it happened, metrics are a part of the story for that, but we also need the default log level to give us enough information to be able to pin-point problems or at least know for sure certain things are going correctly.

A lot of our bug reports include just the Import #<block_number> log, it is really hard to figure it out just from that what happened and for a certain categories of bugs you can't go and tell them to enable `DEBUG` level because they are not reproducing systematically or we don't know how to reproduce it.

Hence, I has thinking we should actually, go the opposite direction and re-think, a bit the information our INFO level prints, so that we make it easier to understand what is going on.

I'm aware that too much logging might cause other problems, but I don't think that was the issue with this line, because it is outputed every 10 minutes.

bkchr:
I get what you want to achieve. However, printing all authorities that you are not connected are debug information. If you go ahead and just print the connectivity to all authorities in percentage every 10 minutes it is fine.

Having a bad connectivity is almost never a result of what you are doing locally. If not having these issues because you regenerated your node key every restart and then authority discovery not telling the others your new key. However, this is a bug and nothing the operator can change. They could also not change anything for these missing authorities as they are not required to open slots in their firewall for each outgoing connection. Yes they could have configured their server in this way, but then they already failed there.

All in all, printing all this information versus just printing the connectivity every 10 minutes as percentage doesn't really makes any difference.

alexggh:
> All in all, printing all this information versus just printing the connectivity every 10 minutes as percentage doesn't really makes any difference.

It doesn't make a difference for that particular node, but I does it for the whole network and anyone trying to find out which nodes are the ones not properly connected since they would show repeatedly in the logs for all the others.

Anyways, this doesn't matter much because we figure it out the issue, the next one will probably be in other parts of the system, that's why I think that maybe we need to re-think our rules with what we output as `INFO`, currently there isn't much information in there or maybe recommend people running with `DEBUG`, but that is probably **dangerous** and costly because it would fill their disk logs.

sandreim:
@bkchr @alexggh you both raise valid points, but I think this is what we actually need: https://github.com/paritytech/polkadot-sdk/issues/816

bkchr:
> It doesn't make a difference for that particular node, but I does it for the whole network and anyone trying to find out which nodes are the ones not properly connected since they would show repeatedly in the logs for all the others.

And then? You see some ids and don't know who they are. Even if you resolve them to some validator name, it still doesn't solve any problem. You could just write a dedicated tool to query the DHT and then trying to connect to all validators.

PR #3915, opened 2024-03-31
Contracts: Only exec parsed code in benchmarks

pgherveou:
[Weights compare](https://weights.tasty.limo/compare?unit=weight&ignore_errors=true&threshold=10&method=asymptotic&repo=polkadot-sdk&old=master&new=pg%2Fbench_tweaks&path_pattern=substrate%2Fframe%2F**%2Fsrc%2Fweights.rs%2Cpolkadot%2Fruntime%2F*%2Fsrc%2Fweights%2F**%2F*.rs%2Cpolkadot%2Fbridges%2Fmodules%2F*%2Fsrc%2Fweights.rs%2Ccumulus%2F**%2Fweights%2F*.rs%2Ccumulus%2F**%2Fweights%2Fxcm%2F*.rs%2Ccumulus%2F**%2Fsrc%2Fweights.rs)

Note: Raw weights change does not mean much here, as this PR reduce the scope of what is benchmarked, they are therefore decreased by a good margin. One should instead print the Schedule using 

cargo test --features runtime-benchmarks  bench_print_schedule -- --nocapture
or following the instructions from the [README](https://github.com/paritytech/polkadot-sdk/tree/pg/bench_tweaks/substrate/frame/contracts#schedule) for looking at the Schedule of a specific runtime


pgherveou:
@athei  @Robbepop 
I think this approach address the comment of this PR

https://github.com/paritytech/polkadot-sdk/pull/2941#pullrequestreview-1836267022

I have only change one benchmark here `seal_now`, to test this out, but other benchmark should be updated as well.
Because of the tricky lifetime requirement, I am leaking some memory for now. depending how the benchmark are executed this might not be a big deal, and we can always clean this up with some unsafe code.

pgherveou:

pgherveou:
https://weights.tasty.limo/compare?unit=weight&ignore_errors=true&threshold=10&method=asymptotic&repo=polkadot-sdk&old=master&new=pg/bench_tweaks&path_pattern=substrate/frame/**/src/weights.rs,polkadot/runtime/*/src/weights/**/*.rs,polkadot/bridges/modules/*/src/weights.rs,cumulus/**/weights/*.rs,cumulus/**/weights/xcm/*.rs,cumulus/**/src/weights.rs

pgherveou:
Looking the Schedule result
```diff
-            now: 329 ns, 6 bytes,
+           now: 280 ns, 0 bytes,
```

substrate/frame/contracts/src/benchmarking/mod.rs
pgherveou, line 1: Note: We should update all host fn benchmark in this PR. Leaving a single update for now

pgherveou:

pgherveou:
bot cancel 2-11f65051-854c-4ba5-93f1-47f2d7449529

pgherveou:

substrate/frame/contracts/src/benchmarking/call_builder.rs
```diff
@@ -0,0 +1,104 @@
...
+			schedule: T::Schedule::get(),
+			gas_meter: GasMeter::new(Weight::MAX),
+			storage_meter: Default::default(),
+			input: vec![],
+		}
+	}
+
+	/// Prepare a call to the module.
+	/// Returns a a closure used to invoke the call.
+	pub fn build(self) -> impl FnOnce() {
```
pgherveou, line 78: Hack: 
I need to leak and reclaim some references here as the benchmark needs to move ownership of the code behind `#[block]` not sure if there is a better approach

Robbepop: I think you wouldn't need to leak anything or use `unsafe` Rust code if you change the API of your `CallBuilder`. Instead of a `fn build() -> impl FnOnce` method you could provide a `fn run()` method that simply executes the test instead of returning a closure to allow the caller to execute the test. So the `CallBuilder` itself can simply store all the data. I did something similar with Wasmi's translation tests: https://github.com/wasmi-labs/wasmi/blob/master/crates/wasmi/src/engine/translator/tests/op/block.rs#L98 and https://github.com/wasmi-labs/wasmi/blob/master/crates/wasmi/src/engine/translator/tests/driver.rs.

pgherveou: in the code you shared, it looks like you are also creating the module inside the `run` fn, which we want to do ahead of time here so that it's done outside of the benchmarked block.

Robbepop: On further note: In my opinion returning `impl X` usually also is a code smell in my opinion and should only ever be used for highly generic code (which again is kinda code smelly) if there is clearly no other possible way out.

pgherveou: > , I already see leaks, for example if the returned closure is accidentally never executed for whatever reason and simply dropped which again is a problem

right, I guess we could just mark this as unsafe to put the caller in charge of upholding these safety guarantees

> On further note: In my opinion returning impl X usually also is a code smell in my opinion and should only ever be used for some generic code

I had it setup a bit differently on the previous [commit](https://github.com/paritytech/polkadot-sdk/pull/3915/commits/d381d65992c7d0913fc65456479f6731a52c4af1), but ideally, because this code will be repeated a lot in the benchmark we can come up with an API as concise as the current one.

pgherveou: @ggwpez, if you have any suggestions for resolving the lifetime issues we're encountering, I'd be eager to hear your thoughts

pgherveou: To give more context, the issue here is that the benchmark framework generate this function

```rust
fn instance(
    &self,
    components: &[(frame_benchmarking::BenchmarkParameter, u32)],
    verify: bool,
) -> Result<frame_benchmarking::__private::Box<dyn FnOnce() -> Result<(), frame_benchmarking::BenchmarkError>>, frame_benchmarking::BenchmarkError> 
{
    let r = components
        .iter()
        .find(|&c| c.0 == frame_benchmarking::BenchmarkParameter::r)
        .ok_or("Could not find component during benchmark preparation.")?
        .1;

    let invoke = CallBuilder::<T>::new(WasmModule::getter("seal0", "seal_now", r)).build();
    Ok(
        frame_benchmarking::__private::Box::new(move || -> Result<(), frame_benchmarking::BenchmarkError> {
            {
                invoke();
            }
            if verify {}
            Ok(())
        }),
    )
}
```

Without tricking Rust with 'static ref we end up with the classic
```
returns a value referencing data owned by the current function
```

pgherveou: I guess an approach could be to update the impl_benchmark macro here 
https://github.com/paritytech/polkadot-sdk/blob/2d466329c7212311a43b6dcc51df7708db4f0f14/substrate/frame/benchmarking/src/v1.rs?plain=1#L1111-L1117

instead of returning a closure, we would just "inject" the start & end value around the benchmark block.

pgherveou:

substrate/frame/contracts/src/benchmarking/call_builder.rs
```diff
@@ -0,0 +1,103 @@
...
+			&mut self.storage_meter,
+			&self.schedule,
+			0u32.into(),
+			None,
+			Determinism::Enforced,
+		)
+	}
+
+	/// Prepare a call to the module.
+	/// Returns a a closure used to invoke the call.
```
pgherveou, line 88:
```suggestion
```

pgherveou:

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5780829

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5782708

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5787900

pgherveou:

substrate/frame/contracts/fixtures/contracts/return_with_data.rs
```diff
@@ -38,6 +38,9 @@ pub extern "C" fn call() {
 		output: [u8],
 	);
 
+	// burn some PoV
+	api::clear_storage_v1(b"");
+
```
athei, line 43: Why is that here?

substrate/frame/contracts/src/benchmarking/call_builder.rs
```diff
@@ -0,0 +1,170 @@
...
+	};
+    ($func:ident, setup: $setup: ident) => {
+		$crate::call_builder!($func, _contract, setup: $setup);
+	};
+    ($func:ident, $contract: ident, setup: $setup: ident) => {
+		let data = $setup.data();
+		let $contract = $setup.contract();
+		let (mut ext, module) = $setup.ext();
+		let $func = CallSetup::<T>::prepare_call(&mut ext, module, data);
+	};
```
athei, line 169: This will make the call side really hard to read because it just crates bindings which are not seen at the call side. I would prefer to return a tuple or a struct here.

substrate/frame/contracts/src/wasm/mod.rs
```diff
@@ -1806,6 +1851,9 @@ mod tests {
 			)
 		)
 
+	    ;; Burn some PoV
+		(call $clear_storage (i32.const 0))
+
```
athei, line 1856: Why is that here?

substrate/frame/contracts/fixtures/contracts/return_with_data.rs
pgherveou: arg looks like I messed up some merge, will remove

pgherveou: actually no,
This is related to the update below in the other test
This make the contract consume some PoV so that these asserts works
(They don't anymore since return_value does not consume PoV anymore)
https://github.com/paritytech/polkadot-sdk/blob/pg/bench_tweaks/substrate/frame/contracts/fixtures/contracts/caller_contract.rs?plain=1#L118

athei: It is kind of counter intuitive why `clear_storage` is used for that. It works because in order to write an item, we read its size before. However, just reading something would be more clear.

substrate/frame/contracts/src/benchmarking/call_builder.rs
pgherveou: Not sure how I would do that, without raising some borrowing compilation errors.
$func borrow from ext that borrows from $setup, if I make it a function or try to return something, rustc will complain that something does not live long enough.

substrate/frame/contracts/src/wasm/mod.rs
pgherveou: same reason as above

substrate/frame/contracts/fixtures/contracts/return_with_data.rs
pgherveou: Yes right I can convert to a get_storage and just ignore the value, 
I used it here because I used it in the .wat below where it's slightly easier since clear does not push anything on the stack

substrate/frame/contracts/src/benchmarking/call_builder.rs
athei: I didn't realize you input the indentifiers so it is not actually as bad as I thought. My bad.

athei (approved):

substrate/frame/contracts/src/benchmarking/call_builder.rs
pgherveou: mmm, to do so we would need to update 
```rust
pub struct Runtime<'a, E: Ext + 'a> {
	ext: &'a mut E,
	input_data: Option<Vec<u8>>,
	memory: Option<Memory>,
	chain_extension: Option<Box<<E::T as Config>::ChainExtension>>,
}
```
to take ownership of ext instead, is that what you had in mind ?

athei: That would be a big change and I am not sure if this would actually work. I meant just moving the `Ext` into the `PreparedCall` if that is easily possible.

pgherveou: well then I don't see how we could do that easily without bothering the borrow checker.
Not too sure what that would buy us as well ...

substrate/frame/contracts/fixtures/contracts/return_with_data.rs
```diff
@@ -38,6 +38,9 @@ pub extern "C" fn call() {
 		output: [u8],
 	);
 
+	// burn some PoV
```
xermicus, line 41: Nit: comment could include some of below discussion for context

xermicus (approved):

substrate/frame/contracts/fixtures/contracts/return_with_data.rs
pgherveou: Yep

pgherveou:

PR #3920, opened 2024-04-01
Update derive syn parse 0.2.0 (+ docify)

sam0x17:
derive-syn-parse v0.2.0 came out recently which (finally) adds support for syn 2x.

Upgrading to this will remove many of the places where syn 1x was still compiling alongside syn 2x in the polkadot-sdk workspace.

This also upgrades `docify` to 0.2.8 which is the version that upgrades derive-syn-pasre to 0.2.0.

Additionally, this consolidates the `docify` versions in the repo to all use the latest, and in one case upgrades to the 0.2x syntax where 0.1.x was still being used.


lexnv (approved):
LGTM

liamaharon (approved):

Issue #3921, opened 2024-04-01
pallet-scheduler: tests::postponed_named_task_cannot_be_rescheduled broken


sandreim:
More context: https://github.com/paritytech/polkadot-sdk/issues/3884

Blocks merging https://github.com/paritytech/polkadot-sdk/pull/3885

TRY 6 FAIL [   0.008s] pallet-scheduler tests::postponed_named_task_cannot_be_rescheduled

PR #3923, opened 2024-04-01
pallet-scheduler: fix test

sandreim:
fix https://github.com/paritytech/polkadot-sdk/issues/3921

### DIFF
--- a/substrate/frame/scheduler/src/tests.rs
+++ b/substrate/frame/scheduler/src/tests.rs
@@ -2316,11 +2316,12 @@ fn postponed_named_task_cannot_be_rescheduled() {
 		);
 
 		// Finally add the preimage.
-		assert_ok!(Preimage::note(call.encode().into()));
+		assert_ok!(Preimage::note_preimage(RuntimeOrigin::signed(0), call.encode()));
+
 		run_to_block(1000);
 		// It did not execute.
 		assert!(logger::log().is_empty());
-		assert!(Preimage::is_requested(&hash));
+		assert!(!Preimage::is_requested(&hash));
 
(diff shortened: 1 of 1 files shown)


substrate/frame/scheduler/src/tests.rs
```diff
@@ -2285,9 +2285,18 @@ fn postponed_named_task_cannot_be_rescheduled() {
...
-		assert!(Preimage::is_requested(&hash));
+
+		// Preimage was not available
+		assert_eq!(
+			System::events().last().unwrap().event,
+			crate::Event::CallUnavailable { task: (4, 0), id: Some(name) }.into()
+		);
+
+		// So it should not be requested.
+		assert!(!Preimage::is_requested(&hash));
```
bkchr, line 2299: Some lines below this it is requested again while it wasn't executed. This looks like some potential bug to me. However, I didn't check the logic yet.

bkchr (commented):
.

substrate/frame/scheduler/src/tests.rs
sandreim: yeah, but we do call `Preimage::note(call.encode().into()`

ggwpez: Calling note through the trait instead of the `note_bytes` function does request it indeed.

ggwpez (approved):

substrate/frame/scheduler/src/tests.rs
sandreim: changed to use the extrinsic.

bkchr (approved):

PR #3927, opened 2024-04-01
pallet-xcm: fix weights for all XTs and deprecate unlimited weight ones 

acatangiu:
Fix "double-weights" for extrinsics, use only the ones benchmarked in the runtime.

Deprecate extrinsics that don't specify WeightLimit, remove their usage across the repo.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5718168

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5718758

serban300 (approved):

acatangiu:
bot bench cumulus-assets --runtime asset-hub-westend --pallet pallet_xcm

svyatonik (commented):
Why `pallet::weight` attribute is removed at all from most of calls. Does it mean that the weight is now zero => near to zero fee + ability to create overweight block. I'm probably missing something important, just caught my eye

acatangiu:
> Why `pallet::weight` attribute is removed at all from most of calls. Does it mean that the weight is now zero => near to zero fee + ability to create overweight block. I'm probably missing something important, just caught my eye

It is replaced by [<T as Config>::WeightInfo here](https://github.com/acatangiu/polkadot-sdk/blob/cdb3886404088c137f578cd82430670c6b2daf9a/polkadot/xcm/pallet-xcm/src/lib.rs#L943). The default weight fn is fn with same name in `<T as Config>::WeightInfo`.

There are some places where I kept `pallet::weight` because it is non-default (e.g. see `execute` and `execute_blob`).

svyatonik (approved):
Ok, cool stuff! Didn't knew we can do weight definitions in the `#[pallet::call(..)]`

bkontur (approved):

PR #3929, opened 2024-04-01
Elastic scaling: add e2e test

sandreim:
On top of https://github.com/paritytech/polkadot-sdk/pull/3879

I've also moved the previous test where we ensure multiple cores per para doesn't break non elastic parachains.


polkadot/zombienet_tests/elastic_scaling/assign-core.js
```diff
@@ -1,7 +1,9 @@
-async function run(nodeName, networkInfo, _jsArgs) {
+async function run(nodeName, networkInfo, core) {
```
pepoviola, line 1: `core` here is an array (`["0"]` or `["1"]`), currently is working because `js` coerce an array of a single element with `Number(["0"])` for example, but could lead to errors in the future. We can change it to  

```suggestion
async function run(nodeName, networkInfo, jsArgs) {
    const core = jsArgs[0];
```

To ensure we are using the correct argument.
Thx!!

.gitlab/pipeline/zombienet/polkadot.yml
```diff
@@ -158,13 +158,21 @@ zombienet-polkadot-functional-0011-async-backing-6-seconds-rate:
...
   extends:
     - .zombienet-polkadot-common
   script:
     - /home/nonroot/zombie-net/scripts/ci/run-test-local-env-manager.sh
-      --local-dir="${LOCAL_DIR}/functional"
-      --test="0012-elastic-scaling-mvp.zndsl"
+      --local-dir="${LOCAL_DIR}/elastic_scaling"
+      --test="0001-basic-3cores-6s-blocks.zndsl"
+
+zombienet-polkadot-elastic-scaling-0002-elastic-scaling-mvp:
```
alindima, line 169: I would rename this test to smth else now that we can actually test elastic scaling

elastic-scaling-does-not-brick-old-collators or elastic-scaling-compatibility

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.toml
```diff
@@ -0,0 +1,42 @@
...
+  allowed_ancestry_len = 2
+
+[relaychain.genesis.runtimeGenesis.patch.configuration.config.scheduler_params]
+  max_validators_per_core = 1
+  scheduling_lookahead = 2
+  num_cores = 3
+
+[relaychain.genesis.runtimeGenesis.patch.configuration.config.approval_voting_params]
+  needed_approvals = 3
+  max_approval_coalesce_count = 5
```
alindima, line 15: is this relevant?

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.zndsl
```diff
@@ -0,0 +1,22 @@
...
+# Register 2 extra cores to this some-parachain.
+elastic-validator-0: js-script ./assign-core.js with "0" return is 0 within 600 seconds
+elastic-validator-0: js-script ./assign-core.js with "1" return is 0 within 600 seconds
+
+# Wait for 10 relay chain blocks 
+elastic-validator-0: reports substrate_block_height{status="best"} is at least 20 within 120 seconds
+
+# Parachain should progress with 3 blocks per relay chain block, so it's reasonable to expect state to be 
+# at least 50, assuming some tolerance 
+some-parachain: log line contains "BlockData { state: 50, add: 2 }" within 10 seconds
```
alindima, line 22: a more reliable test would be based on the parachain block height

alindima: also, we can only assert this for parachain 2000. we should check that the other parachain is progressing 3 times slower

alindima (commented):
Did you also make some changes to the adder-collator for this test to pass?

polkadot/zombienet_tests/elastic_scaling/assign-core.js
sandreim: I tried that initially, but it seems that the jsArgs contained just one value so I assumed this is the way it works. I'll give it one more shot. Does zombienet expand the params separated like this "arg1,arg2,arg3" ?

.gitlab/pipeline/zombienet/polkadot.yml
sandreim: yeah, makes sense

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.toml
sandreim: Not directly relevant to elastic scaling, but it is a good practice for these e2e integration tests to catch more potential issues outside the main focus.

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.zndsl
sandreim: Not really possible with test collators, since they are not based on substrate and have no graphana endpoint for the parachain.

sandreim:
> Did you also make some changes to the adder-collator for this test to pass?

No direct changes were needed as we just define the collator_fn which is used by collation-generation that has been already upgraded to support elastic scaling

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.zndsl
alindima: > Not really possible with test collators, since they are not based on substrate and have no graphana endpoint for the parachain.

hmm, this test for example: https://github.com/paritytech/polkadot-sdk/blob/d0ebb850ed2cefeb3e4ef8b8e0a16eb7fb6b3f3e/polkadot/zombienet_tests/smoke/0001-parachains-smoke-test.zndsl#L6, https://github.com/paritytech/polkadot-sdk/blob/d0ebb850ed2cefeb3e4ef8b8e0a16eb7fb6b3f3e/polkadot/zombienet_tests/smoke/0001-parachains-smoke-test.toml#L24
uses adder-collator and asserts the block height

alindima: but that looks like a very old test which isn't being run any more

sandreim: TBH, I am not sure that is correct.

sandreim: For example, querying the adder collator prometheus endpoint gives:
```
# HELP substrate_block_height Block height info of the chain
# TYPE substrate_block_height gauge
substrate_block_height{status="best",chain="rococo_local_testnet"} 47
substrate_block_height{status="finalized",chain="rococo_local_testnet"} 45
substrate_block_height{status="sync_target",chain="rococo_local_testnet"} 47
```

Which is the bundled full node metrics.

sandreim: added test for non-elastic para

polkadot/zombienet_tests/elastic_scaling/assign-core.js
sandreim: done

sandreim:
Seems like deployment error: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5751023

`Error: Timeout(0) for: waitLog(): pod: temp-collator, container: temp-collator, log: waiting for copy files to finish `

CC @pepoviola

polkadot/zombienet_tests/elastic_scaling/0001-basic-3cores-6s-blocks.toml
```diff
@@ -0,0 +1,42 @@
...
+  args = [ "-lparachain=debug,parachain::candidate-backing=trace,parachain::provisioner=trace,parachain::prospective-parachains=trace,runtime=debug"]
+
+{% for id in range(2000,2002) %}
+[[parachains]]
+id = {{id}}
+addToGenesis = true
+
+    [parachains.collator]
+    name = "some-parachain"
+    image = "{{CUMULUS_IMAGE}}"
```
pepoviola, line 38:
```suggestion
    image = "{{COL_IMAGE}}"
```
This should be `COL_IMAGE` to use adder-collator.

pepoviola:
> Seems like deployment error: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5751023
> 
> `Error: Timeout(0) for: waitLog(): pod: temp-collator, container: temp-collator, log: waiting for copy files to finish `
> 
> CC @pepoviola

Fixed in [a4531235d2808a8a614d96df9bf62a31ec283da1](https://github.com/paritytech/polkadot-sdk/pull/3929/commits/a4531235d2808a8a614d96df9bf62a31ec283da1). The config image was the problem.

Thx!

sandreim:
> > Seems like deployment error: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5751023
> > `Error: Timeout(0) for: waitLog(): pod: temp-collator, container: temp-collator, log: waiting for copy files to finish `
> > CC @pepoviola
> 
> Fixed in [a4531235d2808a8a614d96df9bf62a31ec283da1](https://github.com/paritytech/polkadot-sdk/pull/3929/commits/a4531235d2808a8a614d96df9bf62a31ec283da1). The config image was the problem.
> 
> Thx!

thank you @pepoviola I totally missed it.

eskimor (approved):
Do I read this right, that we are doing elastic scaling with 3 collators here instead of just one connecting to multiple backing groups? I guess that works with the adder collator. Do we also support one collator collating on 3 cores?

sandreim:
> Do I read this right, that we are doing elastic scaling with 3 collators here instead of just one connecting to multiple backing groups? I guess that works with the adder collator. Do we also support one collator collating on 3 cores?

Just single collator case, the test has 2 parachains each with one collator:
- 2000 has 3 cores assigned
- 2001 has 1 core assigned

tdimitrov (approved):

Issue #2209, opened 2023-11-07
Request revenue info


eskimor:
Implement [`request_revenue_info`](https://github.com/polkadot-fellows/RFCs/blob/33d45a809d12481c32b6b1d129caf609bb7db7c6/text/0005-coretime-interface.md#request_revenue_info_at).

- [ ] Keep track of revenue in on-demand
- [ ] Implement call

PR #3940, opened 2024-04-02
Coretime: Add request revenue info

antonva:
Enables the `request_revenue` and `notify_revenue` parts of [RFC 5 - Coretime Interface](https://polkadot-fellows.github.io/RFCs/approved/0005-coretime-interface.html)

TODO:
- [x] Finish first pass at implementation
- [x] ~~Need to explicitly burn uncollected and dropped revenue~~ Accumulate it instead
- [x] Confirm working on zombienet
- [x] Tests 
- [ ] Enable XCM `request_revenue` sending on Coretime chain on Kusama and Polkadot

Fixes: #2209


polkadot/runtime/parachains/src/assigner_on_demand/tests.rs
```diff
@@ -707,3 +707,39 @@ fn queue_status_size_fn_works() {
...
+#[test]
+fn add_revenue_info_works() {
+	let alice = 100u64;
+	let amt = 10_000_000u128;
+	new_test_ext(GenesisConfigBuilder::default().build()).execute_with(|| {
+		Balances::make_free_balance_be(&alice, amt);
+		assert_eq!(Balances::total_issuance(), amt);
+
+		// Revenue should be empty on block 0
+		assert_eq!(Revenue::<Test>::get().len(), 0);
```
eskimor, line 720: Try to use the public interface in tests as much as possible. The less you rely on implementation, the less your test will break in the future.

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -139,17 +144,24 @@ pub mod pallet {
...
+		/// after Relay-chain block number when whose until parameter is equal to `when`.
+		///
+		/// The period in to the past which when is allowed to be may be limited;
+		/// if so the limit should be understood on a channel outside of this proposal.
+		/// In the case that the request cannot be serviced because when is too old a block
+		/// then a `notify_revenue`` message must still be returned, but its `revenue` field
+		/// may be `None``.
+		#[pallet::weight(<T as Config>::WeightInfo::request_revenue_info_at())]
+		#[pallet::call_index(2)]
+		pub fn request_revenue_info_at(
```
eskimor, line 157: Let's rename this to `request_revenue_at`.

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -1028,13 +1028,18 @@ impl coretime::Config for Runtime {
 
 parameter_types! {
 	pub const OnDemandTrafficDefaultValue: FixedU128 = FixedU128::from_u32(1);
+	// Keep 2 timeslices worth of revenue information.
+	pub const MaxHistoricalRevenue: BlockNumber = 2 * 80;
```
eskimor, line 1032: We should be able to retrieve this value from [here](https://github.com/paritytech/polkadot-sdk/blob/99234440f0f8b24f7e4d1d3a0102a9b19a408dd3/cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs#L227)

s0me0ne-unkn0wn: We could, but I'm asking myself if it is right to create a dependency of a relay chain runtime to a parachain pallet. A common constant could work, and here I should ask which side should define that constant. IIUC, the timeslice length cannot be called a property of either parachain or relay chain; instead, it's a property of the whole system formed by the relay chain and the coretime parachain. Being put like that, it seems that putting it into `rococo-runtime-constants` (and separately into `westend-runtime-constants`) is the most natural move. `coretime-rococo` and `coretime-westend` runtimes naturally depend on the corresponding relay chain constants anyway. Let me know if I'm looking at the wrong angle.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -780,6 +828,53 @@ where
...
+
+		Ok(())
+	}
+
+	fn get_revenue(now: BlockNumberFor<T>, when: BlockNumberFor<T>) -> Balance {
+		let revenue: Balance = 0u128;
+		T::Revenue::mutate(|revenue| {
+			revenue.into_iter().enumerate().for_each(|(index, block_revenue)| {
+				//
+				if when >= now.saturating_sub(index) {
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: I tried implementing it with `split_at_mut`, but it introduced such a large amount of bound-checking code that I ended up keeping the current code, just fixing the condition.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -780,6 +828,53 @@ where
...
+					revenue.saturating_add(block_revenue);
+					block_revenue = 0;
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Well, that *might* make some sense. But I'm not really sure what is more optimal from the storage perspective: keeping a constant-length bounded vector or squeezing it for short periods of time when we don't need the tail elements. Trying to win on space here doesn't make sense, as numbers are too low, and it will be filled again during the next 160 blocks at max. And I'm afraid that constantly changing the length would make the vector jump between storage buckets. @kianenigma do you maybe have some insight?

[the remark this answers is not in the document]
s0me0ne-unkn0wn: Left an alternative implementation in comments: https://github.com/paritytech/polkadot-sdk/pull/3940/files#diff-96442c27c75e4ced42fe956c99adf0b5651f35865fef7f42ebe2791ec44bc3faR634-R643

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -244,11 +262,55 @@ impl<T: Config> Pallet<T> {
...
+		log::info!(target: LOG_TARGET, "Revenue info requested: {:?}", revenue);
+		match TryInto::<Balance>::try_into(revenue) {
+			Ok(raw_revenue) => {
+				log::info!(target: LOG_TARGET, "Revenue into balance success: {:?}", raw_revenue);
+				let message = Xcm(vec![
+					Instruction::UnpaidExecution {
+						weight_limit: WeightLimit::Unlimited,
+						check_origin: None,
+					},
+					mk_coretime_call::<T>(CoretimeCalls::NotifyRevenue((when, raw_revenue))),
```
eskimor, line 296: We will also need an accompanying teleport.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -386,15 +200,21 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_now: BlockNumberFor<T>) -> Weight {
+			// Update revenue information storage.
+			Revenue::<T>::mutate(|revenue| {
+				let _ = revenue.force_insert_keep_left(0, 0u32.into()).defensive_proof("Inserting a revenue value does not fail");
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Well, I do agree, but all those strings get compiled into the runtime, so I try to keep them short. Will fix.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
kianenigma: I am not familiar enough with the topic but overall my advice to you is to, unless if obviously inefficient and wrong, optimize for readability and maintenance. Micro optimizations at the FRAME storage level at the cost of hard to maintain code are IMO a mistake. 

In this case, the amount of time we spent to go to the storage is way more significant than the length of the vector, so I reside to the above advice, do whatever is more sensible as if there is no weight/resource restriction.

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::assigner_on_demand --runtime=rococo

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::coretime --runtime=rococo

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_common::coretime --runtime=westend

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::assigner_on_demand --runtime=westend

s0me0ne-unkn0wn:
bot bench substrate-pallet --pallet=pallet_broker

s0me0ne-unkn0wn:
bot bench substrate-pallet --pallet=pallet_broker

s0me0ne-unkn0wn:
bot bench substrate-pallet --pallet=pallet_broker

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: I definitely need some guidance here. Like, what is a teleport, what should it accompany, and how to implement it :slightly_smiling_face:

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::coretime --runtime=westend

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: Well, I took a crash course on assets, currencies, and teleports, and it looks like I'm starting to understand what it all is about. Still, I'm not sure we need that full-scale XCM teleport we were talking about.

The relay chain withdraws the funds for on-demand purchases from the buyer's account and deposits them to the pallet's account. At the same time, it accounts for the revenue (that accounting is just a number not touching economic stuff yet).

Then, the Coretime chain sends a request to the relay chain to provide the revenue info. At that moment, the relay chain burns funds from the pallet account and provides the exact amount of the revenue collected to the Coretime chain.

Thus, when receiving the answer from the relay chain:
1) The Coretime chain already knows the exact amount of revenue;
2) That amount has already been burnt on the relay chain.

It seems that the only thing that should be done here is to mint the amount in question into the Coretime pallet's account. No need for an additional XCM instruction as all the information required is already known to the receiver and is trusted.

Do I miss something?

eskimor: No. Burn -> mint is a teleport. If you think using the actual XCM is overkill here, that is totally fine. There is a difference later once we have credits as then we don't need to teleport anything, but we can deal with that once we are there.

polkadot/runtime/rococo/src/lib.rs
seadanda: Alternatively we could hardcode them both as we do now, but ensure that they are correct with a straightforward emulated test, since in emulated tests we have both runtimes as dependencies. That means if somebody changes one the test will fail. It's a similar strategy I'm planning to use for hardcoded weights between runtimes

polkadot/runtime/parachains/src/assigner_on_demand/tests.rs
```diff
@@ -342,8 +359,8 @@ fn pop_assignment_for_core_works() {
 
 		// Add enough assignments to the order queue.
 		for _ in 0..2 {
-			place_order(para_a);
-			place_order(para_b);
+			place_order_run_to_101(para_a);
```
eskimor, line 362: Why do we need to trigger a session change?

eskimor (commented):
Looks good!

cumulus/parachains/runtimes/coretime/coretime-westend/src/weights/pallet_broker.rs
```diff
@@ -491,6 +491,9 @@ impl<T: frame_system::Config> pallet_broker::WeightInfo for WeightInfo<T> {
 			.saturating_add(Weight::from_parts(0, 0))
 			.saturating_add(T::DbWeight::get().writes(1))
 	}
+	fn notify_revenue() -> Weight {
+		todo!()
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Oh my, good catch!

polkadot/runtime/parachains/src/assigner_on_demand/tests.rs
s0me0ne-unkn0wn: I'm not sure what the original idea was as this code was here before me. I've just checked and the test passes without this change.

substrate/frame/broker/src/tests.rs
```diff
@@ -449,6 +449,18 @@ fn renewals_affect_price() {
...
+// that the pot is topped up twice: first, when the credits are purchased, the value is transferred
+// from the buyer's account to the pot (and currently that doesn't really result in any credits
+// being acquired on the relay chain as the credit system is not implemented yet), and second,
+// when the revenue is teleported from the relay chain to the parachain, thus minting the revenue
+// value into the pot. Thus, current pot assertions are not correct and should be fixed back when
+// the credit system is implemented.
+//
+// The other way to handle that wold be burning the balance transferred from the buyer's account to
+// the pot in `do_purchase_credit()`, but, as it would affect the code that will go into
+// production, it's more error prone, so handling incorrect pot balance is the lesser evil.
```
eskimor, line 462: Or we just comment out the `do_purchase_credit` call?

s0me0ne-unkn0wn: In principle, yes, but that would require some severe changes to the mock implementation as well, as `TestCoretimeProvider::spend_instantaneous()` behaves like the credit system is the thing. But it would definitely be a cleaner solution, I'll look into how easy it is to implement.

s0me0ne-unkn0wn:
bot bench cumulus-coretime

s0me0ne-unkn0wn:
bot bench cumulus-coretime --pallet=pallet_broker

s0me0ne-unkn0wn:
bot bench cumulus-coretime --pallet=pallet_broker --runtime=coretime-westend

substrate/frame/broker/src/tests.rs
s0me0ne-unkn0wn: Yes, you were right. The changes were not that scary, and the tests are much cleaner now. At least there's no need to guess where the numbers come from.

eskimor (approved):

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -386,15 +200,29 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_now: BlockNumberFor<T>) -> Weight {
+			// Update revenue information storage.
+			Revenue::<T>::mutate(|revenue| {
+				if let Some(overdue) =
+					revenue.force_insert_keep_left(0, 0u32.into()).defensive_unwrap_or(None)
```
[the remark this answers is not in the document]
seadanda: Ahh yea I totally misunderstood. Thanks for the explanation, thought it was KEEPING 0 existing, not keeping them left. This makes more sense.

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -154,17 +168,21 @@ pub mod pallet {
...
-		//	Ok(())
-		//}
+		/// Requests that the Relay-chain send a notify_revenue message back at or soon
+		/// after Relay-chain block number when whose until parameter is equal to `when`.
+		///
+		/// The period in to the past which when is allowed to be may be limited;
+		/// if so the limit should be understood on a channel outside of this proposal.
+		/// In the case that the request cannot be serviced because when is too old a block
+		/// then a `notify_revenue`` message must still be returned, but its `revenue` field
+		/// may be `None``.
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: I do agree, and probably that's because it is a copy-paste from RFC-5 :grin:

substrate/frame/broker/src/tick_impls.rs
```diff
@@ -93,15 +97,27 @@ impl<T: Config> Pallet<T> {
 	}
 
 	pub(crate) fn process_revenue() -> bool {
-		let Some((until, amount)) = T::Coretime::check_notify_revenue_info() else { return false };
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Do I understand correctly, you propose to keep the interface but leave the `RevenueInbox` in the broker pallet for now and move it into another pallet as soon as it's implemented?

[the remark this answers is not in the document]
s0me0ne-unkn0wn: Implemented in https://github.com/paritytech/polkadot-sdk/pull/3940/commits/c263b197a45860d4b3b9deafe637425a149b3751

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -53,19 +55,23 @@ impl WeightInfo for TestWeightInfo {
...
+	//fn credit_account() -> Weight {
+	//	Weight::MAX
+	//}
 	fn assign_core(_s: u32) -> Weight {
 		Weight::MAX
 	}
 }
 
+/// Shorthand for the Balance type the runtime is using.
+pub type BalanceOf<T> =
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: I believe it's just a naming convention, we have literally tens of definitions of `BalanceOf` throughout the codebase, and all of them are different. I don't see an easy way to generalize it.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -790,6 +636,35 @@ where
...
+		});
+
+		let imbalance = T::Currency::burn(amount);
+		let _ = T::Currency::settle(
+			&Self::account_id(),
+			imbalance,
+			WithdrawReasons::FEE,
+			ExistenceRequirement::AllowDeath,
+		);
+		amount
```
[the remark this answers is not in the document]
joepetrowski: `burn` will handle total issuance, but I think it depends on intent. As in, if we want all the burning to be done at once from the Coretime Chain (like, "look how much we burned from Coretime sales") then teleporting probably makes sense and having Coretime manage all the revenue burning. cc @muharem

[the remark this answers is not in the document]
seadanda: This is a combination of system and private income. The private income is an incentive for people to pool cores, so it is distributed to the people who contributed to the pool by the coretime chain before the remaining system amount is dealt with in the same way as bulk revenue.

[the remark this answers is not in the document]
joepetrowski: Yeah I discussed a bit with @s0me0ne-unkn0wn in DM. Since these get minted on the Coretime chain, we should definitely use `teleport` and let Coretime handle what to do with revenue (which may mean teleporting it back to burn).

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -527,7 +373,8 @@ where
 	}
 
 	/// Helper function for `place_order_*` calls. Used to differentiate between placing orders
-	/// with a keep alive check or to allow the account to be reaped.
+	/// with a keep alive check or to allow the account to be reaped. The amount charged is
+	/// burnt from the `Currency` and stored to the pallet account.
```
s0me0ne-unkn0wn, line 377:
```suggestion
	/// stored to the pallet account to be later paid out as revenue.
```

Fixme

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -562,18 +409,35 @@ where
...
+
+			// Add the amount to the current block's (index 0) revenue information.
+			Revenue::<T>::mutate(|bounded_revenue| {
+				if let Some(current_block) = bounded_revenue.get_mut(0) {
+					*current_block = current_block.saturating_add(spot_price);
+				} else {
+					// Revenue has already been claimed in the same block, including the block
+					// itself. It shouldn't normally happen as the parachain part checks not to
+					// claim revenue in the future, but relay-chain-only implementations (e.g. mocks
+					// or some future implementations) may still do that.
```
s0me0ne-unkn0wn, line 436:
```suggestion
					// itself. It shouldn't normally happen as revenue claims in the future are
					// not allowed.
```

Fixme

s0me0ne-unkn0wn (approved):
LGTM :grin:

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -85,13 +108,19 @@ enum CoretimeCalls {
...
 	NotifyCoreCount(u16),
+	#[codec(index = 20)]
+	NotifyRevenue((BlockNumber, Balance)),
 	#[codec(index = 99)]
 	SwapLeases(ParaId, ParaId),
 }
 
 #[frame_support::pallet]
 pub mod pallet {
+
```
bkchr, line 119:
```suggestion
```

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -244,11 +288,50 @@ impl<T: Config> Pallet<T> {
...
 
+	/// Provide the amount of revenue accumulated from Instantaneous Coretime Sales from Relay-chain
+	/// block number last_until to until, not including until itself. last_until is defined as being
+	/// the until argument of the last notify_revenue message sent, or zero for the first call. If
+	/// revenue is None, this indicates that the information is no longer available. This explicitly
+	/// disregards the possibility of multiple parachains requesting and being notified of revenue
+	/// information.
+	///
+	/// The Relay-chain must be configured to ensure that only a single revenue information
+	/// destination exists.
```
bkchr, line 304: Docs talk about last_until and until, while the parameter is called `when`.

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -244,11 +288,50 @@ impl<T: Config> Pallet<T> {
...
+
+		let claim = <assigner_on_demand::Pallet<T>>::start_revenue_claim_until(when_bnf);
+		log::debug!(target: LOG_TARGET, "Revenue info requested: {:?}", claim.amount);
+
+		let raw_revenue: Balance = claim.amount.try_into().map_err(|_| {
+			log::error!(target: LOG_TARGET, "Converting on demand revenue for `NotifyRevenue` failed");
+			Error::<T>::AssetTransferFailed
+		})?;
+
+		with_transaction(|| -> TransactionOutcome<Result<_, DispatchError>> {
```
bkchr, line 320: Why do we call this in a transaction?

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -281,3 +364,55 @@ fn mk_coretime_call<T: Config>(call: crate::coretime::CoretimeCalls) -> Instruct
 		call: BrokerRuntimePallets::Broker(call).encode().into(),
 	}
 }
+
+fn do_notify_revenue<T: Config>(when: BlockNumber, raw_revenue: Balance) -> Result<(), XcmError> {
```
bkchr, line 368: Do we have a test for this?

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
```diff
@@ -205,26 +270,25 @@ impl CoretimeInterface for CoretimeAllocator {
...
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
-	fn ensure_notify_revenue_info(when: RCBlockNumberOf<Self>, revenue: Self::Balance) {
-		CoretimeRevenue::set(&Some((when, revenue)));
+	fn ensure_notify_revenue_info(
+		info: OnDemandRevenueRecord<RCBlockNumberOf<Self>, Self::Balance>,
+	) {
+		RevenueInbox::<Runtime>::put(info);
 	}
```
bkchr, line 283: This logic makes no sense. Why do we put this into the internal storage item and then require that the external trait implementation returns the internal storage value? Either this trait is implemented by some completely different pallet upstream or we remove these two methods.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
```diff
@@ -20,20 +20,89 @@ use cumulus_pallet_parachain_system::RelaychainDataProvider;
...
+
+pub struct BurnStash;
+impl NewTimesliceHook for BurnStash {
+	fn on_new_timeslice(_t: pallet_broker::Timeslice) {
+		let stash = BurnStashAccount::get();
+		let value = Balances::reducible_balance(&stash, Preservation::Expendable, Fortitude::Force);
+
+		if value > 0 {
+			log::debug!(target: "runtime::coretime", "Going to burn {value} stashed tokens at RC");
+			with_transaction(|| -> TransactionOutcome<Result<(), DispatchError>> {
```
bkchr, line 90: Transaction again?

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -87,214 +88,21 @@ impl WeightInfo for TestWeightInfo {
...
-	fn cmp(&self, other: &Self) -> Ordering {
-		match other.idx.cmp(&self.idx) {
-			Ordering::Equal => other.para_id.cmp(&self.para_id),
-			o => o,
-		}
+impl<T: Config> Default for RevenueClaim<T> {
+	fn default() -> Self {
+		Self { split_off: None, amount: 0u32.into() }
 	}
 }
```
bkchr, line 107:
```suggestion
```

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -87,214 +88,21 @@ impl WeightInfo for TestWeightInfo {
...
-	///
-	/// But have a hole to `smallest_index`, so we can not yet bump `smallest_index`. This binary
-	/// heap is roughly bounded in the number of on demand cores:
-	///
-	/// For a single core, elements will always be processed in order. With each core added, a
-	/// level of out of order execution is added.
-	freed_indices: BinaryHeap<ReverseQueueIndex>,
+/// A revenue claim that has been started but is not commited yet.
+/// Produced by `start_revenue_claim_until` and consumed then by `commit_revenue_claim`.
+pub struct RevenueClaim<T: Config> {
```
bkchr, line 93:
```suggestion
#[derive(DefaultNoBound)]
pub struct RevenueClaim<T: Config> {
```

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -244,11 +288,50 @@ impl<T: Config> Pallet<T> {
...
+			match do_notify_revenue::<T>(when, raw_revenue) {
+				Ok(()) => TransactionOutcome::Commit(Ok(())),
+				Err(err) => {
+					log::error!(target: LOG_TARGET, "notify_revenue failed: {err:?}");
+					TransactionOutcome::Rollback(Err(Error::<T>::AssetTransferFailed.into()))
+				},
+			}
+		})?;
+
+		<assigner_on_demand::Pallet<T>>::commit_revenue_claim(claim);
```
bkchr, line 330: Why isn't this directly committed in `start_revenue_claim_until`?

substrate/frame/broker/src/traits.rs
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Timeslice;
+
+pub trait NewTimesliceHook {
+	fn on_new_timeslice(t: Timeslice);
```
bkchr, line 21: Why not just put it into the `CoretimeInterface` trait?

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -562,18 +409,35 @@ where
...
+			// Charge the sending account the spot price. The amount will be teleported to the
+			// broker chain once it requests revenue information.
+			let amt = T::Currency::withdraw(
 				&sender,
 				spot_price,
 				WithdrawReasons::FEE,
 				existence_requirement,
 			)?;
+			// Consume the negative imbalance and deposit it into the pallet account.
+			T::Currency::resolve_creating(&Self::account_id(), amt);
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: We're screwed then. Not in this exact place of code, in many of them. The on-demand base fee should never be lower than the existential deposit. I think it should be documented.

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: Because funds are first withdrawn and then checked out, if the withdrawal is successful and the checkout is not, we end up with unbalanced funds.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
s0me0ne-unkn0wn: I think @seadanda had some reasoning why it should be kept like this :thinking: 
@seadanda ?

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
s0me0ne-unkn0wn: The same reasoning

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: If the XCM stuff between `start_revenue_claim_until` and `commit_revenue_claim` fails, we don't call `commit_revenue_claim`, thus leaving the revenue where it is so we can try to collect it again on the next iteration. Otherwise, we end up with funds sitting in the pot, not accounted for anywhere.

substrate/frame/broker/src/traits.rs
s0me0ne-unkn0wn: Because we want batching here. Without batching it may result in quite a few XCMs sent for every small reason.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -324,6 +132,21 @@ pub mod pallet {
...
 		type TrafficDefaultValue: Get<FixedU128>;
+
+		/// The maximum number of blocks some historical revenue
+		/// information stored for.
+		#[pallet::constant]
+		type MaxHistoricalRevenue: Get<u32>;
+
+		/// Identifier for the internal revenue balance.
+		#[pallet::constant]
+		type PalletId: Get<PalletId>;
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: That's a fair point, but I see the issue is still open and has some ongoing discussion. It's how it is implemented in many places (I do agree it's nasty), and here, beyond establishing the account address from the relay chain side, it's also used from the parachain side to give Coretime an idea of which account it should teleport funds to. If you have ideas on how to improve it quickly, please suggest. Otherwise, it just works.

substrate/frame/broker/src/types.rs
```diff
@@ -255,6 +255,21 @@ pub struct LeaseRecordItem {
...
 pub type LeasesRecordOf<T> = LeasesRecord<<T as Config>::MaxLeasedCores>;
 
+/// Record for On demand core sales.
+///
+/// The blocknumber is the relay chain block height `when` the original request
+/// for revenue was made.
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct OnDemandRevenueRecord<RelayBlockNumber, RelayBalance> {
+	/// The height of the Relay-chain at the time the revenue request was made.
+	pub until: RelayBlockNumber,
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Well, it's put out [in the RFC-5](https://polkadot-fellows.github.io/RFCs/approved/0005-coretime-interface.html#request_revenue_info_at) in that rather dubious way, and I think it's where all the confusion with this `when`/`until` duality comes from. Here, it's formulated explicitly in terms of the RFC:

> Requests that the Relay-chain send a `notify_revenue` message back at or soon after Relay-chain block number `when` whose `until` parameter is equal to `when`.

substrate/frame/broker/src/mock.rs
```diff
@@ -90,22 +92,25 @@ impl CoretimeInterface for TestCoretimeProvider {
...
+		mint_to_pot(total);
+		RevenueInbox::<Test>::put(OnDemandRevenueRecord { until: when, amount: total });
 	}
-	fn credit_account(who: Self::AccountId, amount: Self::Balance) {
-		CoretimeCredit::mutate(|c| c.entry(who).or_default().saturating_accrue(amount));
+	fn credit_account(_who: Self::AccountId, _amount: Self::Balance) {
+		// Commented out code is from the reference mock implementation and should be uncommented as
+		// soon as we have the credit system implemented
+
+		// CoretimeCredit::mutate(|c| c.entry(who).or_default().saturating_accrue(amount));
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: I probably started commenting out the credit-related code in tests (where it really needed to be commented out), got blown away, and ended up commenting out everything related. I'll check if it breaks something; if not, I'll bring it back.

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: We have a WIP test for everything not yet merged in but already working: https://github.com/s0me0ne-unkn0wn/coretime-revenue-test/blob/master/src/main.rs

Do you think a separate unit test could be useful here?

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Thinking about it a bit more... We have three potentially problematic accounts:
1) `CheckedAccount` at the relay chain. It's the least problem as in production environment we always have some inactive issuance, and the situation when it becomes empty is nearly impossible, so there's no problem with checking out amounts less than the existential deposit;
2) Broker pallet pot where the revenue is teleported into. Given that the on-demand base fee is always greater than the ED, it doesn't seem to be a concern. But there are corner cases, like if the on-demand base fee is equal or just a bit greater than the ED and the contributions are partially system and partially private, then we'll burn the system part, and the rest may fall under the ED.
3) Coretime chain burn stash, which batches funds to burn on the relay chain every timeslice. Concerns are similar to 2).

We can handle that in code, but it's much easier to handle 2) and 3) by just manually putting the ED to those two accounts. This requires only a tiny alteration of logic (the burn stash should not burn to zero but to the ED) and the problem is solved. WDYT?

CC @eskimor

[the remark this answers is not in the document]
s0me0ne-unkn0wn: Implemented mitigation for 3) in https://github.com/paritytech/polkadot-sdk/pull/3940/commits/3decbf37eeae90de0278b575a48d38b62f98c32b
Anyway, it makes things better. We don't have to kill and recreate the account every burning cycle.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
s0me0ne-unkn0wn: Found it: https://github.com/paritytech/polkadot-sdk/pull/3940/commits/3decbf37eeae90de0278b575a48d38b62f98c32b

substrate/frame/broker/src/mock.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: I ended up with a single commented-out line that was ruining assertions, all the other code fits well.

polkadot/runtime/parachains/src/coretime/mod.rs
bkchr: Yes

substrate/frame/broker/src/traits.rs
bkchr: I mean put the function there instead of having another trait.

polkadot/runtime/parachains/src/coretime/mod.rs
bkchr: The XCM stuff should not fail. If it fails, there is a bug.

polkadot/runtime/parachains/src/coretime/mod.rs
bkchr: Same as below. If it fails, there is a bug in your XCM code that is there inherently.

substrate/frame/broker/src/types.rs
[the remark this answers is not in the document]
bkchr: The line in the RFC makes sense, but not here. You can just call it `until` and that's it.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
bkchr: He says it should not be tightly coupled, but what you are doing here is still the same as before. It is still tightly coupled. You basically "cheated" to fix his comment, but the underlying solution is still the same.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
bkchr: Make the pallet a provider of the account. Then the account is not reaped when it goes below ED.

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: Okay, I'll remove it.

CC @joepetrowski I remember you insisted this part must be transactional

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
bkchr: > it's also used from the parachain side to give Coretime an idea of which account it should teleport funds to.

Then just make it an account directly, instead of passing a `PalletId`.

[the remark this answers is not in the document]
s0me0ne-unkn0wn: You mean hardcode some `[u8; 32]` there? Sounds even more hacky :/

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Any clue on how to achieve that? Never did that before

s0me0ne-unkn0wn:
bot bench cumulus-coretime --pallet=pallet_broker --runtime=coretime-westend

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::coretime --runtime=westend

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::coretime --runtime=rococo

s0me0ne-unkn0wn:
bot bench substrate-pallet --pallet=pallet_broker

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::coretime --runtime=rococo

s0me0ne-unkn0wn:
bot bench polkadot-pallet --pallet=runtime_parachains::assigner_on_demand --runtime=westend

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
s0me0ne-unkn0wn: Yes, exactly. He suggests we keep the methods where they are not to break the interface, and later, we move the `RevenueInbox` into the coretime allocator pallet when we have it, thus decoupling things and not introducing breaking changes to the public interface twice. I agree that the current implementation is kind of cheating, but it's for good.

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: Raised a follow-up: #4882

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
bkchr: It is not for "good"

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
bkchr: https://github.com/paritytech/polkadot-sdk/blob/63e264446f6cabff06be72912eae902662dcb699/substrate/frame/system/src/lib.rs#L1488

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
s0me0ne-unkn0wn: So what do you suggest? Remove it from the interface completely?

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
bkchr: No, but let's keep the `PalletId`...

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
bkchr: This entire code path makes no sense. Just remove the trait function and then it is good. You also added the dispatchable to report the revenue to the pallet. I really don't see any reasons for the gymnastics you are doing there.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -790,6 +632,29 @@ where
...
+		let mut amount: BalanceOf<T> = BalanceOf::<T>::zero();
+		Revenue::<T>::mutate(|revenue| {
+			while !revenue.is_empty() {
+				let index = (revenue.len() - 1) as u32;
+				if when > now.saturating_sub(index.into()) {
+					amount = amount.saturating_add(revenue.pop().defensive_unwrap_or(0u32.into()));
+				} else {
+					break
+				}
+			}
```
bkchr, line 647:
```suggestion
			while let Some(rev) = revenue.pop() {
				let index = revenue.len() as u32;
				if when > now.saturating_sub(index.into()) {
					amount = amount.saturating_add(rev);
				} else {
					break
				}
			}
```

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -562,18 +382,40 @@ where
...
 				existence_requirement,
 			)?;
 
-			ensure!(
-				queue_status.size() < config.scheduler_params.on_demand_queue_max_size,
-				Error::<T>::QueueFull
-			);
+			// Consume the negative imbalance and deposit it into the pallet account. Make sure the
+			// account preserves even without the existential deposit.
+			let pot = &Self::account_id();
```
bkchr, line 401:
```suggestion
			let pot = Self::account_id();
```

polkadot/runtime/parachains/src/coretime/mod.rs
```diff
@@ -281,3 +355,55 @@ fn mk_coretime_call<T: Config>(call: crate::coretime::CoretimeCalls) -> Instruct
...
+					beneficiary: T::BrokerPotLocation::get().into_location(),
+				},
+			]
+			.into_iter(),
+		);
+	}
+
+	message.push(mk_coretime_call::<T>(CoretimeCalls::NotifyRevenue((when, raw_revenue))));
+
+	send_xcm::<T::SendXcm>(dest.clone(), Xcm(message))?;
```
joepetrowski, line 402: We know that this will send with Root privilege? It will be necessary for the `UnpaidExecution` instruction. IMO this is where testing with xcm-emulator is worth it.

polkadot/runtime/parachains/src/mock.rs
```diff
@@ -411,6 +418,13 @@ parameter_types! {
 	pub MaxXcmTransactWeight: Weight = Weight::from_parts(10_000_000, 10_000);
 }
 
+pub struct BrokerPot;
+impl Get<InteriorLocation> for BrokerPot {
+	fn get() -> InteriorLocation {
+		todo!()
```
joepetrowski, line 424: Needs doing?

cumulus/parachains/runtimes/coretime/coretime-rococo/src/coretime.rs
```diff
@@ -66,11 +109,7 @@ enum CoretimeProviderCalls {
 
 parameter_types! {
 	pub const BrokerPalletId: PalletId = PalletId(*b"py/broke");
-}
-
-parameter_types! {
-	pub storage CoreCount: Option<CoreIndex> = None;
-	pub storage CoretimeRevenue: Option<(BlockNumber, Balance)> = None;
+	pub BurnStashAccount: AccountId = BrokerPalletId::get().into_sub_account_truncating(b"burnstash");
```
joepetrowski, line 112: Not crazy about the naming since it collides with our concept of a stash account in staking. Maybe `RevenueAccumulationAccount`? And then:

```rust
pub struct BurnCoretimeRevenue;
impl OnUnbalanced<Credit<AccountId, Balances>> for BurnCoretimeRevenue
```

joepetrowski (approved):

polkadot/runtime/parachains/src/mock.rs
s0me0ne-unkn0wn: Not really, it's never called. Will change it to `unimplemented`.

polkadot/runtime/parachains/src/coretime/mod.rs
s0me0ne-unkn0wn: Well, it works in manual zombienet tests and in integration tests, so I guess it's likely to be okay... But if someone could make it through the emulator that would be great!

seadanda (approved):

ggwpez:
How do i integrate this into a runtime? PRdoc and or MR description should provide some integration advise for downstream teams.

s0me0ne-unkn0wn:
> How do i integrate this into a runtime? PRdoc and or MR description should provide some integration advise for downstream teams.

Well, it's a fair point, although I doubt the PRdoc is a good place for it... Anyway, I was going to start a PR in the fellowship repo to integrate it into Kusama first.

Cargo.lock
```diff
@@ -22200,34 +22185,31 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.36"
```
nazar-pc, line 22203: Not sure why `time` upgrade had to be reverted, but it created compilation issues that was fixed in https://github.com/paritytech/polkadot-sdk/pull/4862, so `stable2407` is broken on nightly again. Does no one review lock file changes or what?

bkchr: Yeah that was an oversight. Downstream users should not have the problem as they can just upgrade the patch version. When you compile the stable in the repo, there is no other way around than bumping this locally for you.

PR #3941, opened 2024-04-02
Use 0.1.0 as minimum version for crates

Morganamilo:
CI will be enforcing this with next parity-publish release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1920,7 +1920,7 @@ dependencies = [
 
 [[package]]
 name = "bridge-hub-common"
-version = "0.0.0"
+version = "0.1.0"
 dependencies = [
  "cumulus-primitives-core",
  "frame-support",
@@ -10479,7 +10479,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-parameters"
-version = "0.0.1"
+version = "0.1.0"
 dependencies = [
  "docify",
  "frame-benchmarking",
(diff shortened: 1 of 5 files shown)


ordian (approved):

niklasad1 (approved):

ggwpez (approved):
Yes. Can you also please ban suffices for published crates? Or are they already banned?  
In the future we should also take crate-metadata into account. I drafted this [here](https://github.com/paritytech/polkadot-sdk/pull/3825/files#diff-ad79d250c3e6ba2c2a112962791239e297bb0c5f6bdd41f44b9a64ff0c50b4d1R46) but think that it should better be put in a proper tool.

lexnv (approved):

liamaharon (approved):

Morganamilo:
> Yes. Can you also please ban suffices for published crates? Or are they already banned? In the future we should also take crate-metadata into account. I drafted this [here](https://github.com/paritytech/polkadot-sdk/pull/3825/files#diff-ad79d250c3e6ba2c2a112962791239e297bb0c5f6bdd41f44b9a64ff0c50b4d1R46) but think that it should better be put in a proper tool.

I also added a check for pre release. I'm not sure what the use for the metadata is. So far I've been using 'is publish' to decide if a crate is internal. In what cases would 'is publish' and 'is internal' be different?

ggwpez:
> I also added a check for pre release. I'm not sure what the use for the metadata is. So far I've been using 'is publish' to decide if a crate is internal. In what cases would 'is publish' and 'is internal' be different?

There are some published crates that can still have breaking changes since we dont consider them as part of our public api.  
It is mentioned in the release docs, it will only be relevant later.

Issue #3145, opened 2024-01-30
[HRMP] Make `max_message_size` adjustable


ggwpez:
HRMP channels have a `max_message_size` and will possibly trunc/drop messages that are longer than this.  
I order to prepare for the revival of https://github.com/paritytech/polkadot-sdk/pull/2302, there needs to be a way to reduce the `max_message_size` of all open channels since the Message Queue would otherwise fail to enqueue them.

One possible way to do this between `Alice` (requesting a change to MMS) and `Bob` counter-party:
- `Alice`: `RequestConfigUpdate(MaxMessageSize(123))`. This will be called `ReqId` for brevity.
- `Bob`:
  - `Ok(ReqId)` if the new MMS value is larger or equal to the old one.
  - `TentativeOk(ReqId, LatestRelayBlockNumber)` otherwise and adjusts the limits locally.
- `Alice`: waits until `LatestRelayBlockNumber` has elapsed and adjusts the limits locally.

franciscoaguirre:
Are those XCM instructions?

ggwpez:
> Would those be XCM instructions?

Could be, for a start i would probably go with a new extrinsic + transact.

PR #2302, opened 2023-11-13
[Runtime] Bound XCMP queue

ggwpez:
Remove `without_storage_info` from the XCMP queue pallet. Part of https://github.com/paritytech/polkadot-sdk/issues/323  

Changes:
- Limit the number of channels that can be suspended at the same time.
- Limit the number of channels that can have messages or signals pending at the same time.

A No-OP migration is put in place to ensure that all `BoundedVec`s still decode and not truncate after upgrade. The storage version is thereby bumped to 4 to have our tooling remind us to deploy that migration.

Issue #323, opened 2021-04-16
[FRAME Core] Remove `without_storage_info` on pallets


shawntabrizi:
https://github.com/paritytech/substrate/pull/8556

Need to update all instances of `Vec` in the runtime storage to `BoundedVec`, which will allow us to start generating worse case scenario PoV sizes.

- [ ] Validators / Session Stuff: https://github.com/paritytech/substrate/issues/8640
    - [x] Aura
    - [x] Authority Discovery
    - [x] Babe
    - [x] Grandpa
    - [x] Im-Online
    - [ ] Session
- [x] Authorship @koute https://github.com/paritytech/substrate/pull/11610
- [x] Beefy @KiChjang paritytech/substrate#11584
- [ ] paritytech/polkadot-sdk#167
- [x] ConvictionVoting @gavofyork https://github.com/paritytech/substrate/pull/11631
- [x] Democracy @gavofyork https://github.com/paritytech/substrate/pull/11649
- [ ] ~~ElectionProviderMultiPhase~~
- [ ] ElectionsPhragmen @Szegoo https://github.com/paritytech/substrate/pull/12549
- [x] (Examples)
- [x] Lottery https://github.com/paritytech/substrate/pull/10594
- [x] Membership @skunert https://github.com/paritytech/substrate/pull/11591
- [x] Multisig @shawntabrizi https://github.com/paritytech/substrate/pull/12457
- [ ] NodeAuthorization @koute https://github.com/paritytech/substrate/pull/11621
- [x] Offences @ggwpez https://github.com/paritytech/substrate/pull/11585
- [x] Preimage @gavofyork https://github.com/paritytech/substrate/pull/11649
- [x] RankedCollective @shawntabrizi https://github.com/paritytech/substrate/pull/11602/files
- [x] Referenda @gavofyork https://github.com/paritytech/substrate/pull/11631
- [x] Remark @KiChjang paritytech/substrate#11590
- [x] Scheduler @gavofyork https://github.com/paritytech/substrate/pull/11649
- [x] ScoredPool @hbulgarini https://github.com/paritytech/substrate/pull/11996
- [x] Society @gavofyork https://github.com/paritytech/substrate/pull/11324
- [ ] Staking
- [x] StateTrieMigration @ggwpez https://github.com/paritytech/substrate/pull/11639
- [ ] Sudo @shawntabrizi https://github.com/paritytech/substrate/pull/11596
- [ ] System @gavofyork Blocked on (Infra 3).
[...]

PR #3952, opened 2024-04-02
[Runtime] Bound XCMP queue

ggwpez:
Re-applying #2302 after increasing the `MaxPageSize`.  

Remove `without_storage_info` from the XCMP queue pallet. Part of https://github.com/paritytech/polkadot-sdk/issues/323  

Changes:
- Limit the number of messages and signals a HRMP channel can have at most.
- Limit the number of HRML channels.

A No-OP migration is put in place to ensure that all `BoundedVec`s still decode and not truncate after upgrade. The storage version is thereby bumped to 5 to have our tooling remind us to deploy that migration.

## Integration

If you see this error in your try-runtime-cli:  
```pre
Max message size for channel is too large. This means that the V5 migration can be front-run and an
attacker could place a large message just right before the migration to make other messages un-decodable.
Please either increase `MaxPageSize` or decrease the `max_message_size` for this channel. Channel max:
102400, MaxPageSize: 65535
```

Then increase the `MaxPageSize` of the `cumulus_pallet_xcmp_queue` to something like this:
```rust
type MaxPageSize = ConstU32<{ 103 * 1024 }>;
```

There is currently no easy way for on-chain governance to adjust the HRMP max message size of all channels, but it could be done: https://github.com/paritytech/polkadot-sdk/issues/3145.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5738872

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5741533

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5776669

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -297,19 +319,28 @@ pub mod pallet {
 	/// case of the need to send a high-priority signal message this block.
 	/// The bool is true if there is a signal message waiting to be sent.
 	#[pallet::storage]
-	pub(super) type OutboundXcmpStatus<T: Config> =
-		StorageValue<_, Vec<OutboundChannelDetails>, ValueQuery>;
+	pub(super) type OutboundXcmpStatus<T: Config> = StorageValue<
+		_,
+		BoundedVec<OutboundChannelDetails, T::MaxActiveOutboundChannels>,
```
ggwpez, line 324: `Vec -> BoundedVec` the migration checks that `MaxActiveOutboundChannels` is above the current length of this.  
If governance decides to open more channels, then it will truncate on decode.

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -297,19 +319,28 @@ pub mod pallet {
...
 	#[pallet::storage]
-	pub(super) type OutboundXcmpMessages<T: Config> =
-		StorageDoubleMap<_, Blake2_128Concat, ParaId, Twox64Concat, u16, Vec<u8>, ValueQuery>;
+	pub(super) type OutboundXcmpMessages<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		ParaId,
+		Twox64Concat,
+		u16,
+		WeakBoundedVec<u8, T::MaxPageSize>,
```
ggwpez, line 336: `Vec -> WeakBoundedVec` since we dont want to truncate any messages.

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -297,19 +319,28 @@ pub mod pallet {
...
+		ValueQuery,
+	>;
 
 	/// Any signal messages waiting to be sent.
 	#[pallet::storage]
 	pub(super) type SignalMessages<T: Config> =
-		StorageMap<_, Blake2_128Concat, ParaId, Vec<u8>, ValueQuery>;
+		StorageMap<_, Blake2_128Concat, ParaId, WeakBoundedVec<u8, T::MaxPageSize>, ValueQuery>;
```
ggwpez, line 343: `Vec -> WeakBoundedVec` since we dont want to truncate any messages.

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -503,7 +535,9 @@ impl<T: Config> Pallet<T> {
 						if page.len() + encoded_fragment.len() > max_message_size {
 							return None
 						}
-						page.extend_from_slice(&encoded_fragment[..]);
+						for frag in encoded_fragment.iter() {
+							page.try_push(*frag).ok()?;
```
ggwpez, line 539: WeakBoundedVec is missing some convenience functions. Will fix.

cumulus/parachains/runtimes/coretime/coretime-rococo/src/lib.rs
```diff
@@ -299,7 +300,9 @@ impl pallet_message_queue::Config for Runtime {
 	// The XCMP queue pallet is only ever able to handle the `Sibling(ParaId)` origin:
 	type QueueChangeHandler = NarrowOriginToSibling<XcmpQueue>;
 	type QueuePausedQuery = NarrowOriginToSibling<XcmpQueue>;
-	type HeapSize = sp_core::ConstU32<{ 64 * 1024 }>;
+	// Most on-chain HRMP channels are configured to use 102400 bytes of max message size, so we
+	// need to set the page size larger than that until we reduce the channel size it on-chain.
+	type HeapSize = sp_core::ConstU32<{ 103 * 1024 }>;
```
ggwpez, line 305: Bumped the MQ page size so it wont drop HRMP pages.

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -132,6 +131,25 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxInboundSuspended: Get<u32>;
 
+		/// Maximal number of outbound XCMP channels that can have messages queued at the same time.
+		///
+		/// If this is reached, then no further messages can be sent to channels that do not yet
+		/// have a message queued. This should be set to the expected maximum of outbound channels
+		/// which is determined by [`Self::ChannelInfo`]. It is important to set this correctly,
```
franciscoaguirre, line 138: What does "correctly" mean in this case? Is there any other information we could add here to make it better for users setting this?

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -276,6 +294,10 @@ pub mod pallet {
 		AlreadySuspended,
 		/// The execution is already resumed.
 		AlreadyResumed,
+		/// There are too many active outbound channels.
+		TooManyOutboundChannels,
```
franciscoaguirre, line 298:
```suggestion
		TooManyActiveOutboundChannels,
```

Maybe?

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -664,13 +711,17 @@ impl<T: Config> OnQueueChanged<ParaId> for Pallet<T> {
 		let suspended = suspended_channels.contains(&para);
 
 		if suspended && fp.ready_pages <= resume_threshold {
-			Self::send_signal(para, ChannelSignal::Resume);
+			// If the resuming fails then it is not critical. We will retry in the future.
```
franciscoaguirre, line 714: Why can the resuming fail? Why is it not critical?

cumulus/pallets/xcmp-queue/src/migration/v5.rs
```diff
@@ -0,0 +1,119 @@
...
+					info.max_message_size, max_msg_len
+				);
+				//return Err("Migration can be front-run".into());
+			}
+		}
+
+		// Now check that all pages still fit into the new `BoundedVec`s:
+		for page in v4::OutboundXcmpMessages::<T>::iter_values() {
+			if page.len() >= T::MaxPageSize::get() as usize {
+				log::warn!("Too long message in storage. WeakBoundedVec should handle this.");
```
franciscoaguirre, line 106: How does `WeakBoundedVec` handle this?

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -478,7 +508,9 @@ impl<T: Config> Pallet<T> {
 		{
 			details
 		} else {
-			all_channels.push(OutboundChannelDetails::new(recipient));
+			all_channels
+				.try_push(OutboundChannelDetails::new(recipient))
+				.map_err(|_| MessageSendError::TooManyChannels)?;
```
franciscoaguirre, line 513: I know the error is very simple, but we could log it instead of ignoring it. What do you think?

cumulus/pallets/xcmp-queue/src/lib.rs
franciscoaguirre: Is there a difference between open channels and active channels? Do open channels already have a limit?

cumulus/pallets/xcmp-queue/src/lib.rs
ggwpez: It can go wrong when a chain has like 200 channels open but they set this value to 100.  
The migration does prevent this in the `pre_upgrade` hook, so it should be mostly fine. Unless opengov decides to open a channel right before the RA.  
Will add that as comment.

cumulus/pallets/xcmp-queue/src/migration/v5.rs
ggwpez: It just truncates 😬 will change comment.

cumulus/pallets/xcmp-queue/src/lib.rs
ggwpez: If it fails here then the channel is either broken or something else with HRMP delivery does not work.  
Best we can do is to retry, so from the perspective here "not critical" means that we keep buffering messages and not dropping them. Amending the comment.

ggwpez: Hm the code before is not so good either. Actually it should not remove the channel from the `InboundXcmpSuspended` in this case. Will change.

cumulus/pallets/xcmp-queue/src/lib.rs
ggwpez: Open channels are not really tracked in this pallet. They are just provided by `T::ChannelInfo::get_channel_info(para_id)`, so the pallet has no clue how many channels are currently open.  
But it does track all "active" channels, so to say all channels that have messages queued.

cumulus/pallets/xcmp-queue/src/lib.rs
ggwpez: Changed it to better handle the error. "not critical" in the sense that we dont have to invalidate the candidate for this (=panic).

muharem (approved):

cumulus/pallets/xcmp-queue/src/lib.rs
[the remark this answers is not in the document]
ggwpez: Yea true. In the past, there was the possibility for multiple signals to be enqueued, but i removed that, since later signals completely overwrite any previous ones.  
Now indeed it is wasting weight here... it could probably be changed to something very small, like 16 without an issue.

Issue #1756, opened 2023-09-29
SubscribeVersion from the Relay will be blocked by new parachains


mrshiposha:
The Relay chain [uses the `WithUniqueTopic` router](https://github.com/paritytech/polkadot-sdk/blob/f60318f68687e601c47de5ad5ca88e2c3f8139a7/polkadot/runtime/polkadot/src/xcm_config.rs#L123C22-L123C37), which [adds a `SetTopic` instruction](https://github.com/paritytech/polkadot-sdk/blob/bf90cb0b73a2d2e0b3e79b2956256831ae79c9f7/polkadot/xcm/xcm-builder/src/routing.rs#L44-L51) to every XCM message the Relay sends if the message doesn't end with such an instruction already.

This could break the XCM version subscriptions for new parachains that didn't receive the `SubscribeVersion` in the past.

The `SubscribeVersion` message will also be augmented with the `SetTopic` instruction. Currently, the `AllowSubscriptionsFrom` barrier used by the parachains [forbids any message that contains anything but the `SubscribeVersion`/`UnsubscribeVersion` instruction](https://github.com/paritytech/polkadot-sdk/blob/f60318f68687e601c47de5ad5ca88e2c3f8139a7/polkadot/xcm/xcm-builder/src/barriers.rs#L367-L371). So, any `SubscribeVersion` or `UnsubscribeVersion` from the Relay chain will be blocked.

bkchr:
CC @franciscoaguirre

franciscoaguirre:
That sounds bad, I'll look into it

bkontur:
@mrshiposha Yes, you're right. `AllowSubscriptionsFrom` accepts just `SubscribeVersion` / `UnsubscribeVersion`, but there is a [`TrailingSetTopicAsId` barrier](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/barriers.rs#L213-L240), which is exactly designed to handle this `SetTopic` added by `WithUniqueTopic` router. Therefore, the parachain should wrap its barrier with `TrailingSetTopicAsId`, for example: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs#L485-L506. Anyway, the parachain still has the possibility to write custom barriers according to its needs.

[...]

PR #3955, opened 2024-04-02
Added tests for XCM barriers: `AllowSubscriptions`, `WithUniqueTopic` and `TrailingSetTopicAsId`

bkontur:
Closes: https://github.com/paritytech/polkadot-sdk/issues/1756

### DIFF
--- a/polkadot/xcm/xcm-builder/src/tests/barriers.rs
+++ b/polkadot/xcm/xcm-builder/src/tests/barriers.rs
@@ -309,3 +309,62 @@ fn suspension_should_work() {
 	);
 	assert_eq!(r, Ok(()));
 }
+
+#[test]
+fn allow_subscriptions_from_should_work() {
+	// allow only parent
+	AllowSubsFrom::set(vec![Location::parent()]);
+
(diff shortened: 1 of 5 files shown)


svyatonik (approved):

polkadot/xcm/xcm-builder/src/tests/barriers.rs
```diff
@@ -309,3 +309,62 @@ fn suspension_should_work() {
...
+	let invalid_xcm_1 = Xcm::<TestCall>(vec![
+		SetAppendix(Xcm(vec![])),
+		SubscribeVersion { query_id: 42, max_response_weight: Weight::from_parts(5000, 5000) },
+	]);
+	let invalid_xcm_2 = Xcm::<TestCall>(vec![
+		SubscribeVersion { query_id: 42, max_response_weight: Weight::from_parts(5000, 5000) },
+		SetTopic([0; 32]),
+	]);
+
+	let test_data = vec![
```
serban300, line 332: Nit: Not very important, but personally I would find it easier to follow and I think it would also be more extensible if we just asserted each case separately instead of iterating through a `vec`. I don't think it would create a lot of duplication.

serban300 (approved):

Issue #3894, opened 2024-03-29
Allow Nomination pool configuration changes with StakingAdmin.


Ank4n:
`set_config` should be set by StakingAdmin instead of root.

iMac7:
Hi @Ank4n 
I'm a newcomer to this repo, not as new to the polkadot ecosystem. I like to pick up issues to give me a starting point to look around, this one looks promising :) 

Just a few noob questions (to you or assigned mentor).
Is this done in `/substrate` (pallet issue) or `/polkadot` or other folder? 
Where can I find the nomination pool config?
How would I test this locally? Assume I have cloned the repo and installed all dependencies. Nothing more satisfying that running a tiny blockchain on my pc  :smile:

Ank4n:
Hey @iMac7 thanks for your interest. Sorry I have already worked on it so this is unfortunately not up for mentor. But I hope you find another good first issue.

PR #3959, opened 2024-04-03
Nomination pool configurations can be managed by custom origin

Ank4n:
closes https://github.com/paritytech/polkadot-sdk/issues/3894

Allows Nomination Pool configuration to be set by a custom origin instead of root.

In runtimes, we would set this to be `StakingAdmin`, same as for pallet-staking.


prdoc/pr_3959.prdoc
```diff
@@ -0,0 +1,14 @@
...
+
+doc:
+  - audience: Runtime User
+    description: |
+      Adds a custom origin to Nomination pool configuration and allows StakingAdmin to be this origin in Westend. Other
+      runtimes could also set this origin to be the same that manages staking-pallet configurations.
+
+crates:
+  - name: pallet-nomination-pools
+    bump: minor
```
liamaharon, line 14: does it compile without `type AdminOrigin = ...`? otherwise I think it's major 
```suggestion
    bump: major
```

liamaharon (approved):

kianenigma (approved):

PR #3962, opened 2024-04-03
Change forks pruning algorithm.

shamil-gadelshin:
This PR changes the fork calculation and pruning algorithm to enable future block header pruning. It's required because the previous algorithm relied on the block header persistence. It follows the [related discussion](https://github.com/paritytech/polkadot-sdk/issues/1570)

The previous code contained this comment describing the situation:
```
	/// Note a block height finalized, displacing all leaves with number less than the finalized
	/// block's.
	///
	/// Although it would be more technically correct to also prune out leaves at the
	/// same number as the finalized block, but with different hashes, the current behavior
	/// is simpler and our assumptions about how finalization works means that those leaves
	/// will be pruned soon afterwards anyway.
	pub fn finalize_height(&mut self, number: N) -> FinalizationOutcome<H, N> {
```

The previous algorithm relied on the existing block headers to prune forks later and to enable block header pruning we need to clear all obsolete forks right after the block finalization to not depend on the related block headers in the future.


Polkadot-Forum:

https://forum.polkadot.network/t/block-header-pruning/7198/1

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -255,6 +247,58 @@ pub trait Backend<Block: BlockT>:
...
+				}
+
+				if let Some(parent_header) = self.header(*fork_block_header.parent_hash())? {
+					fork_block_header = parent_header;
+				} else {
+					// Sometimes routes can't be calculated. E.g. after warp sync.
+					needs_pruning = true;
+					break
+				}
+			}
```
bkchr, line 292:
```suggestion
			// All leaves will eventually have as ancestor either the finalized block or its
			// parent. All other forks were cleared on the previous block finalization.
			let needs_pruning = loop {
				// The block ends up in the finalized block. All forks are valid at this point.
				if fork_block_header.hash() == finalized_block_hash {
					break false
				}

				// The block ends up in the parent block of the finalized block. It's a stale fork.
				if *fork_block_header.number() <= finalized_block_number {
					break true
				}

				if let Some(parent_header) = self.header(*fork_block_header.parent_hash())? {
					fork_block_header = parent_header;
				} else {
					// Sometimes routes can't be calculated. E.g. after warp sync.
					break true
				}
			};
```

bkchr (commented):
While the logic is correct, it is much more costly doing this than doing the "lazy" pruning as before. If we would record the block at which a certain leaf forked off, we should probably be able to achieve the same without iterating always over all the headers from the leaf down to the finalized block?

substrate/primitives/blockchain/src/backend.rs
shamil-gadelshin: This looks much better! Thank you!

shamil-gadelshin:
> it is much more costly doing this than doing the "lazy" pruning as before

1. I assume that we traverse the block subtree which is relatively shallow with a small of leaves on each block finalization and I see the most expensive operation here is "read header from DB (disk)" in a loop. (Am I correct here?).
a) However, this operation is cached and is likely to be accessible from the cache in the majority of header acquisitions.
b) Both the previous and the current algorithms invoke this function chain to prune blocks: note_finalized()->prune_blocks()->prune_displaced_branches()->tree_route() and `tree_route` is invoked for each dispaced leaf and contains within it a similar loop with extracting header metadata (which invokes the same "read header" operation each time). Even if the new algorithm will read all the necessary headers from DB instead of the cache (likely on a restart) - the subsequent `tree_route` operations will be much faster. To be precise the `archive` pruning mode won't invoke the `prune_block` code but we consider the worst case anyway. 

> If we would record the block at which a certain leaf forked off, we should probably be able to achieve the same without iterating always over all the headers from the leaf down to the finalized block?

2. I understand your suggestion as follows: on each block import operation calculate a branching block for the fork, save it to a dedicated structure, and use this information later instead of calculation.
a) Because of the possible restarts we need to make it persistent, but it contradicts your previous guideline: https://github.com/paritytech/polkadot-sdk/issues/1570#issuecomment-1892336635
I understood that comment as "in general, we should avoid adding persistent data if we can calculate the result on the fly (similar to `tree_route` operation)". 
b) Even if we add the persistent cached data for current leaves it seems we will update this data in a similar loop but in the opposite direction: on each new block we would check whether it's a new best and recalculate saved "fork-points" for all the leaves. However, this time we won't be able to cache disk "write" operations. Did I understand it correctly? 

Summary:
The new proposed algorithm doesn't seem to worsen the operation cost much because of the mostly cached operations and already existing `tree_route` operation.

bkchr:
> 1. a) However, this operation is cached and is likely to be accessible from the cache in the majority of header acquisitions.

Yes, I mean I would assume the same as well. I think the best is to ask @arkpar on why he hasn't done it this way from the beginning. There is maybe something we oversee, otherwise I'm also fine with the approach you are proposing here.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5786697

arkpar:
This was written by someone else, but I guess it was done for simplicity, as the quoted comment explains.

As for this PR, it looks good for me. I wonder if it can be optimized though. `prune_displaced_branches` calls `sp_blockchain::tree_route` for each displaced leaf anyway. Instead of traversing the header chain, it should be possible to just call `tree_route` once somewhere and then store or pass around the `retracted` path for each leaf.

shamil-gadelshin:
I implemented the suggestions from @arkpar  The code seems simpler now.

arkpar (approved):

shamil-gadelshin:
The last commit removes the tests related to the removed method: `finalize_height`.
The new code is tested with other tests related to pruning blocks (displaced leaves).

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5818237

shamil-gadelshin:
The last commits fix tests by updating the expected block where stale heads appear. However, I also changed `expand_forks` to use `tree_route` function first. It seems to me that the function works incorrectly because it uses `hash(block_number)` function to work with the canonical chain to get a parent block by number. I don't think that `hash(block_number)` defines its behavior in the presence of forks (it just reads from DB). Please, correct me if I'm wrong. I didn't delete the original code and left it as a fallback option because it tries to return a partial route in case where some blocks are missing and `tree_route` doesn't do that.

shamil-gadelshin:
@arkpar Do you mind looking at the `expand_fork()` changes?

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -104,6 +107,18 @@ pub trait ForkBackend<Block: BlockT>:
...
+					for block in tree_route.retracted() {
+						expanded_forks.insert(block.hash);
+					}
+					continue
+				},
+				Err(_) => {
+					// Continue with fallback algorithm
+				},
+			}
+
```
bkchr, line 121:
```suggestion
```

See my comment in the pr.

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -255,6 +262,67 @@ pub trait Backend<Block: BlockT>:
...
+/// Calculation result of the displaced leaves after the block finalization.
+#[derive(Clone, Debug)]
+pub struct DisplacedLeavesAfterFinalization<Block: BlockT> {
+	/// A collection of hashes and block numbers for displaced leaves.
+	pub displaced_leaves: BTreeMap<Block::Hash, NumberFor<Block>>,
+
+	/// A collection of tree routes from the leaves to finalized block.
+	pub tree_routes: BTreeMap<Block::Hash, TreeRoute<Block>>,
+}
```
bkchr, line 313:
```suggestion
/// Result of  [`HeaderBackend::displaced_leaves_after_finalizing`].
#[derive(Clone, Debug)]
pub struct DisplacedLeavesAfterFinalization<Block: BlockT> {
	/// A collection of hashes and block numbers for displaced leaves.
	pub displaced_leaves: BTreeMap<Block::Hash, NumberFor<Block>>,

	/// A collection of tree routes from the leaves to finalized block.
	pub tree_routes: BTreeMap<Block::Hash, TreeRoute<Block>>,
}

impl<Block: BlockT> Default for DisplacedLeavesAfterFinalization<Block> {
	fn default() -> Self {
		Self { displaced_leaves: Default::default(), tree_routes: Default::default() }
	}
}
```

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -255,6 +262,67 @@ pub trait Backend<Block: BlockT>:
...
+	/// Returns a collection of hashes for the displaced leaves.
+	pub fn hashes(&self) -> Vec<Block::Hash> {
+		self.displaced_leaves.keys().map(|h| *h).collect()
+	}
+
+	/// Constructor. We need this explicit initialization to not introduce a requirement
+	/// `Block: Default`
+	pub fn new() -> Self {
+		Self { displaced_leaves: BTreeMap::new(), tree_routes: BTreeMap::new() }
+	}
```
bkchr, line 325:
```suggestion
```

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -255,6 +262,67 @@ pub trait Backend<Block: BlockT>:
...
+	pub displaced_leaves: BTreeMap<Block::Hash, NumberFor<Block>>,
+
+	/// A collection of tree routes from the leaves to finalized block.
+	pub tree_routes: BTreeMap<Block::Hash, TreeRoute<Block>>,
+}
+
+impl<Block: BlockT> DisplacedLeavesAfterFinalization<Block> {
+	/// Returns a collection of hashes for the displaced leaves.
+	pub fn hashes(&self) -> Vec<Block::Hash> {
+		self.displaced_leaves.keys().map(|h| *h).collect()
```
bkchr, line 318:
```suggestion
	pub fn hashes(&self) -> impl Iterator<Item = Block::Hash> {
		self.displaced_leaves.keys().map(|h| *h)
```

substrate/primitives/blockchain/src/backend.rs
```diff
@@ -255,6 +262,67 @@ pub trait Backend<Block: BlockT>:
...
 
 	fn block_indexed_body(&self, hash: Block::Hash) -> Result<Option<Vec<Vec<u8>>>>;
+
+	/// Returns all leaves that will be displaced after the block finalization.
+	fn displaced_leaves_after_finalizing(
+		&self,
+		finalized_block_hash: Block::Hash,
+		finalized_block_number: NumberFor<Block>,
+	) -> std::result::Result<DisplacedLeavesAfterFinalization<Block>, Error> {
+		let mut result = DisplacedLeavesAfterFinalization::new();
```
bkchr, line 272:
```suggestion
		let mut result = DisplacedLeavesAfterFinalization::default();
```

bkchr (approved):

prdoc/pr_3962.prdoc
```diff
@@ -0,0 +1,11 @@
+title: Change fork calculation algorithm.
+
+doc:
+  - audience: Node Dev
+    description: |
+      This PR changes the fork calculation and pruning algorithm to enable future block header pruning.
```
bkchr, line 6: Please also mention that forks are now pruned much faster. Best to give some example.

shamil-gadelshin:
@arkpar @bkchr Please, have another look when you have time.

arkpar (approved):

arkpar:
looks good here

bkchr:
> When we have only A3 finalized and we call `expand_forks(C5)` then `hash(block_number = 4)` will return C4, `parent_hash == meta.parent` condition will be incorrect and the final function result will be `expand_fork(C5) = [C5]`.
> 
> This is likely because fork3 is the longest chain.

Yes for sure this is related to the fork3 being the longest chain. Substrate knows a best chain/canonical chain ahead of the last finalized block (for BABE the chain with the highest block height). However, we can keep the code you added.

But then please update the docs of `expand_forks` to make clear that it goes back to the finalized block.

// G --- A1 --- A2 --- A3 --- A4           ( < fork1 )
//                                           \-----C4 --- C5    ( < fork3 )

If you have something like this and `A3` is the last finalized block, it will now return `A4`, `C4`, `C5`. Which is quite a difference. Maybe we should remove the function completely and people should directly use `tree_route`.

shamil-gadelshin:
The last commits update the recently added test and the comment for `expand_forks` function.

It seems there are two usages of the `expand_forks`: `sc-consensus-babe` and `mmr-gadget`. I don't think it's a good idea to change those in the scope of this PR. Please, let me know whether I should change those crates, create a refactoring issue, or leave it be.

bkchr:
@shamil-gadelshin the `expand_forks` function should have a consistent behavior, either based on tree route as what you have done in your pr or the old approach. However, currently both could be returned. Could you please remove one of the code paths?

shamil-gadelshin:
> Could you please remove one of the code paths?

Sure. I removed the old code and updated the function comment as well as its dependencies. I also updated one of the tests to counter its flaky behavior.

bkchr (approved):
Thank you!

bkchr:
@shamil-gadelshin the rustdoc jobs are still failing.

bkchr (approved):

PR #3568, opened 2024-03-04
Move cumulus zombienet tests to aura & async backing

skunert:
Cumulus test-parachain node and test runtime were still using relay chain consensus and 12s blocktimes. With async backing around the corner on the major chains we should switch our tests too.

Also needed to nicely test the changes coming to collators in #3168.

### Changes Overview
- Followed the [migration guide](https://wiki.polkadot.network/docs/maintain-guides-async-backing) for async backing for the cumulus-test-runtime
- Adjusted the cumulus-test-service to use the correct import-queue, lookahead collator etc.
- The block validation function now uses the Aura Ext Executor so that the seal of the block is validated
- Previous point requires that we seal block before calling into `validate_block`, I introduced a helper function for that
- Test client adjusted to provide a slot to the relay chain proof and the aura pre-digest

Issue #3168, opened 2024-02-01
Slot based collations


bkchr:
The current collator implementations depend on the relay chain block import as "clock". This means for every imported relay chain block, they will check the relay chain state if the parachain is allowed to build a block and then do so. However, the relay chain for example has a task that fires every 6 seconds and then builds a block. For sure parachains still need to include their blocks still in the relay chain, but with async backing parachains get more freedom when it comes to block production. So, we should split up the current collator implementation into two tasks:

1. Block production
2. Collation generation

The block production would run the same way as it is done on the relay chain. This means we have a fixed slot and this fire every X seconds to build a block. This mechanism needs to be implemented in a flexible way to support the following cases:

- One parachain block every X seconds, where X >= 6 seconds (the relay chain slot time). 
- Build a parachain block on demand on some trigger. The trigger could be for example that the transaction pool is full enough and we have bought some on demand slot.
- One parachain block every X seconds, where X < 6 seconds, but the block time is less than 2 seconds. This should be useful for parachains that want to run in a higher pace than the relay chain to give faster feedback on transaction inclusion.
- One parachain block every X seconds, where X < 6 seconds, but the block time is 2 seconds. This should be useful for parachains that want to build 3 blocks every 6 seconds because they have 3 cores constantly to achieve an higher throughput. 
- Future: Scale block production depending on the demand/special triggers.

This logic for block production should be implemented in a fairly generic way. Meaning that the actual block production is hidden behind some generic type as the stuff here is actually more like the trigger to build a block. The actual block production is then something like Aura that gets the key/builds/signs the block and returns it. Then we almost have like a normal chain that has its block production separated. However, we should consider to slow down the block production if our parachain is running too far in front of what is already enacted in the relay chain. This means that the collation tasks tells the block production to slow down. As each parachain block is build on a certain relay chain block that provides some context, we would use the best block of the relay chain as of starting the block production.

The collation task would work as the current collator task is working. This means it listens for relay chain block imports, checks if the parachain has a slot and if yes, creates the collation. After https://github.com/paritytech/polkadot-sdk/issues/3167 is finished, it would also be possible to build the collation early enough to send it on time to the relay chain. We could probably think about certain kind of optimizations, like keeping the storage proof from an imported/built block around to not require to re-run the block for creating the collation.

[...]

PR #3963, opened 2024-04-03
Introduce basic slot-based collator

skunert:
Part of #3168 
On top of #3568

### Changes Overview
- Introduces a new collator variant in `cumulus/client/consensus/aura/src/collators/slot_based/mod.rs`
- Two tasks are part of that module, one for block building and one for collation building and submission.
- Introduces a new variant of `cumulus-test-runtime` which has 2s slot duration, used for zombienet testing
- Zombienet tests for the new collator

**Note:** This collator is considered experimental and should only be used for testing and exploration for now.

### Comparison with `lookahead` collator
- The new variant is slot based, meaning it waits for the next slot of the parachain, then starts authoring
- The search for potential parents remains mostly unchanged from lookahead
- As anchor, we use the current best relay parent
- In general, the new collator tends to be anchored to one relay parent earlier. `lookahead` generally waits for a new relay block to arrive before it attempts to build a block. This means the actual timing of parachain blocks depends on when the relay block has been authored and imported. With the slot-triggered approach we are authoring directly on the slot boundary, were a new relay chain block has probably not yet arrived.

### Limitations
- Overall, the current implementation focuses on the "happy path"
- We assume that we want to collate close to the tip of the relay chain. It would be useful however to have some kind of configurable drift, so that we could lag behind a bit. https://github.com/paritytech/polkadot-sdk/issues/3965
- The collation task is pretty dumb currently. It checks if we have cores scheduled and if yes, submits all the messages we have received from the block builder until we have something submitted for every core. Ideally we should do some extra checks, i.e. we do not need to submit if the built block is already too old (build on a out of range relay parent) or was authored with a relay parent that is not an ancestor of the relay block we are submitting at. https://github.com/paritytech/polkadot-sdk/issues/3966
- There is no throttling, we assume that we can submit _velocity_ blocks every relay chain block. There should be communication between the collator task and block-builder task.
- The parent search and ConsensusHook are not yet properly adjusted. The parent search makes assumptions about the pending candidate which no longer hold. https://github.com/paritytech/polkadot-sdk/issues/3967
- Custom triggers for block building not implemented.

### TBD
- [ ] PRDOC

### DIFF
--- a/cumulus/client/consensus/common/src/lib.rs
+++ b/cumulus/client/consensus/common/src/lib.rs
@@ -46,6 +46,8 @@ pub use level_monitor::{LevelLimit, MAX_LEAVES_PER_LEVEL_SENSIBLE_DEFAULT};
 
 pub mod import_queue;
 
+const PARENT_SEARCH_LOG_TARGET: &str = "consensus::common::find_potential_parents";
+
 /// Provides the hash of validation code used for authoring/execution of blocks at a given
 /// hash.
 pub trait ValidationCodeHashProvider<Hash> {
@@ -350,6 +352,7 @@ pub async fn find_potential_parents<B: BlockT>(
 	let included_hash = included_header.hash();
 	let pending_hash = pending_header.as_ref().map(|hdr| hdr.hash());
 
+	tracing::trace!(target: PARENT_SEARCH_LOG_TARGET, ?included_hash, included_num = ?included_header.number(), ?pending_hash ,?rp_ancestry, "Searching relay chain ancestry.");
 	let mut frontier = vec![PotentialParent::<B> {
 		hash: included_hash,
 		header: included_header,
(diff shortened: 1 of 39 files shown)


sandreim (commented):
You also need to enable `elastic-scaling-experimental` feature when building `polkadot-parachain-bin` and you should be able to collate properly on all assigned cores.

cumulus/client/consensus/common/src/lib.rs
```diff
@@ -389,19 +484,19 @@ pub async fn find_potential_parents<B: BlockT>(
 		}
 
 		// push children onto search frontier.
-		for child in client.blockchain().children(hash).ok().into_iter().flatten() {
+		for child in backend.blockchain().children(hash).ok().into_iter().flatten() {
```
skunert, line 487: We could consider iterating from the leaves along the parents here. (not in this PR)

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5844325

Issue #3268, opened 2024-02-08
[FRAME] Prepare pallets for dynamic block durations


ggwpez:
Currently we just use `System::block_number()` in many pallet and derive a timestamp from it. This will not work anymore when parachains have changing block times, either from async backing or coretime.  

## Possible Solution

<sup>(the naming here are just placeholders)</sup>

- Create a new system config item: `BlockNumberProvider` which can then either be configured to `RelaychainDataProvider` when its a parachain runtime or `()` for relay runtimes.
- Add a function `System::provided_block_number() -> Number`
- Add a function `System::local_block_number() -> Number` (for migrations and to avoid ambiguity)
- Deprecate `System::block_number()`

We then need to adapt a ton of pallets and check whether their storage needs to be migrated.

kianenigma:
This is more or less the proposed solution that I extracted from the call, mostly my attempt to capture an idea that @gupnik expressed: 

```diff
diff --git a/substrate/frame/scheduler/src/lib.rs b/substrate/frame/scheduler/src/lib.rs
index e94f154eee..bcb5aff728 100644
--- a/substrate/frame/scheduler/src/lib.rs
+++ b/substrate/frame/scheduler/src/lib.rs
@@ -334,7 +334,9 @@ pub mod pallet {
 		#[pallet::weight(<T as Config>::WeightInfo::schedule(T::MaxScheduledPerBlock::get()))]
 		pub fn schedule(
 			origin: OriginFor<T>,
-			when: BlockNumberFor<T>,
+			// we provide this type to all dispatchables wishing to reference the future. Later on,
+			// `fn passed` can be used to check if this time is already passed.
+			when: T::RuntimeTime,
 			maybe_periodic: Option<schedule::Period<BlockNumberFor<T>>>,
 			priority: schedule::Priority,
 			call: Box<<T as Config>::RuntimeCall>,
diff --git a/substrate/frame/system/src/lib.rs b/substrate/frame/system/src/lib.rs
index 069217bcee..ab725a7dd5 100644
--- a/substrate/frame/system/src/lib.rs
+++ b/substrate/frame/system/src/lib.rs
@@ -570,6 +570,64 @@ pub mod pallet {
 
 		/// The maximum number of consumers allowed on a single account.
 		type MaxConsumers: ConsumerLimits;
+
+		/// Something that can represent a notion of time within this runtime.
+		///
+		/// It can be one's block number, timestamp or similar, depending on wether this is being
[...]

PR #3970, opened 2024-04-03
Update Treasury to Support Relay Chain Block Number Provider

shawntabrizi:
The goal of this PR is to have the treasury pallet work on a parachain which does not produce blocks on a regular schedule, thus can use the relay chain as a block provider.

Because blocks are not produced regularly, we cannot make the assumption that block number increases monotonically, and thus have new logic to handle multiple spend periods passing between blocks.


substrate/frame/treasury/src/lib.rs
```diff
@@ -451,17 +462,27 @@ pub mod pallet {
 			}
 
 			// Check to see if we should spend some funds!
-			if (n % T::SpendPeriod::get()).is_zero() {
-				Self::spend_funds()
+			let last_spend_period =
+				// AUDIT REVIEW TODO! Needs to handle migration story for this better.
+				LastSpendPeriod::<T, I>::get().unwrap_or(BlockNumberFor::<T>::zero());
```
shawntabrizi, line 467: call out here

probably needs migration, but perhaps there is a way to avoid it with some clever logic?

bkchr (dismissed):

substrate/frame/treasury/src/lib.rs
xlc:
```suggestion
				LastSpendPeriod::<T, I>::get().unwrap_or_else(|| calculate_last_spend_period());
```

in that way no migration is needed if the block number provider is the local block number

but we will need a migration when switch the block number provider

shawntabrizi: > but we will need a migration when switch the block number provider

Seems unavoidable

substrate/frame/treasury/src/lib.rs
```diff
@@ -437,7 +448,8 @@ pub mod pallet {
 	impl<T: Config<I>, I: 'static> Hooks<BlockNumberFor<T>> for Pallet<T, I> {
 		/// ## Complexity
 		/// - `O(A)` where `A` is the number of approvals
-		fn on_initialize(n: frame_system::pallet_prelude::BlockNumberFor<T>) -> Weight {
+		fn on_initialize(_do_not_use_local_block_number: BlockNumberFor<T>) -> Weight {
+			let block_number = T::BlockNumberProvider::current_block_number();
```
shawntabrizi, line 452: This might be a common error when people write pallets depending on a `BlockNumberProvider`

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5831372

substrate/frame/treasury/src/lib.rs
```diff
@@ -289,6 +293,9 @@ pub mod pallet {
 		/// Helper type for benchmarks.
 		#[cfg(feature = "runtime-benchmarks")]
 		type BenchmarkHelper: ArgumentsFactory<Self::AssetKind, Self::Beneficiary>;
+
+		/// Provider for the block number.
+		type BlockNumberProvider: BlockNumberProvider<BlockNumber = BlockNumberFor<Self>>;
```
muharem, line 298: may be a comment that this is the system pallet in general case

muharem (approved):

substrate/frame/treasury/src/lib.rs
```diff
@@ -289,6 +293,9 @@ pub mod pallet {
 		/// Helper type for benchmarks.
 		#[cfg(feature = "runtime-benchmarks")]
 		type BenchmarkHelper: ArgumentsFactory<Self::AssetKind, Self::Beneficiary>;
+
+		/// Provider for the block number.
```
shawntabrizi, line 297:
```suggestion
		/// Provider for the block number. Normally this is the `frame_system` pallet.
```

substrate/frame/treasury/src/lib.rs
shawntabrizi: done

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5867947

substrate/frame/treasury/src/lib.rs
```diff
@@ -453,17 +465,29 @@ pub mod pallet {
 			}
 
 			// Check to see if we should spend some funds!
-			if (n % T::SpendPeriod::get()).is_zero() {
-				Self::spend_funds()
+			let last_spend_period = LastSpendPeriod::<T, I>::get()
+				// This unwrap should only occur one time on any blockchain.
+				// `update_last_spend_period` will populate the `LastSpendPeriod` storage if it is
+				// empty.
+				.unwrap_or(Self::update_last_spend_period());
```
ggwpez, line 472: It is possible to use the [`OnEmpty` syntax](https://github.com/paritytech/polkadot-sdk/blob/cd91c6b782b60906597ed28c4e4a4fdddd8428c9/substrate/frame/examples/kitchensink/src/lib.rs#L130-L137) of the `LastSpendPeriod` and then change it to `ValueQuery` - it is not really optional.

ggwpez (approved):

substrate/frame/treasury/src/lib.rs
shawntabrizi: Feel like it is exactly the same thing, but more substrate magic.

or what am I missing?

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5888107

substrate/frame/treasury/src/lib.rs
ggwpez: Yup. Depends on whether you like more or less magic 😆

ggwpez:
Hey @shawntabrizi, the benchmarks CI is still failing: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6003177

shawntabrizi:
@ggwpez i think it should be good now, but probably requires a more careful review as treasury changes leak into multiple other pallets

substrate/frame/child-bounties/src/benchmarking.rs
```diff
@@ -114,7 +119,8 @@ fn activate_bounty<T: Config>(
 	let approve_origin =
 		T::SpendOrigin::try_successful_origin().map_err(|_| BenchmarkError::Weightless)?;
 	Bounties::<T>::approve_bounty(approve_origin, child_bounty_setup.bounty_id)?;
-	Treasury::<T>::on_initialize(BlockNumberFor::<T>::zero());
```
shawntabrizi, line 117: we shouldnt allow `on_initialize(0)`: https://github.com/paritytech/polkadot-sdk/issues/4316

prdoc/pr_3970.prdoc
```diff
@@ -0,0 +1,11 @@
...
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Update Treasury to Support Block Number Provider
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      The goal of this PR is to have the treasury pallet work on a parachain which does not produce blocks on a regular schedule, thus can use the relay chain as a block provider. Because blocks are not produced regularly, we cannot make the assumption that block number increases monotonically, and thus have new logic to handle multiple spend periods passing between blocks. To migrate existing treasury implementations, simply add `type BlockNumberProvider = System` to have the same behavior as before.
+
+crates: [ ]
```
muharem, line 11: we trying to be more smart with versioning now, here info - https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/prdoc.md#record-semver-changes

shawntabrizi: can you make a suggestion for what should be put here?

substrate/frame/bounties/src/lib.rs
```diff
@@ -551,8 +551,8 @@ pub mod pallet {
 						T::Currency::reserve(curator, deposit)?;
 						bounty.curator_deposit = deposit;
 
-						let update_due = frame_system::Pallet::<T>::block_number() +
-							T::BountyUpdatePeriod::get();
+						let update_due =
+							Self::treasury_block_number() + T::BountyUpdatePeriod::get();
```
ggwpez, line 555: I imagine most of the logic breaks when a parachain migrates from `System` to the `Relay` `BlockNumberProvider`...  
Not sure what can be done about it without re-writing the pallet to use timepoints. Maybe we can use this as first step and advise to still use `System`.

substrate/frame/treasury/src/lib.rs
ggwpez: Yea, `on_initialize` should be deprecated in favour of the new hooks from https://github.com/paritytech/polkadot-sdk/pull/1781.

prdoc/pr_3970.prdoc
muharem, line 11:
```suggestion
crates: 
- name: pallet-treasury
  bump: major
```

shawntabrizi: thank you

substrate/frame/bounties/src/lib.rs
ggwpez: I think in the long term we need some kind of block-number-abstraction like https://github.com/paritytech/polkadot-sdk/issues/3268, but its a bit stale now as low-prio. That is not necessarily timestamps, but could be implemented with Relay blocknumber or something else.

muharem:
@shawntabrizi should we merge the PR? all required CIs are green

shawntabrizi:
Its no problem from me. Perhaps requires an audit, but I leave it to you all to get this merged :)

I think the code makes sense, although adds yet another piece of complexity to pallet config.

muharem:
@shawntabrizi do we have a live or upcoming use case for this?

joepetrowski:
> @shawntabrizi do we have a live or upcoming use case for this?

Don't we need it for instances of Treasury on Collectives?

muharem:
> > @shawntabrizi do we have a live or upcoming use case for this?
> 
> Don't we need it for instances of Treasury on Collectives?

I think we do not, Collectives chain does produce the blocks, and we do not need the spend period to be in sync with Relay Treasury spend period.

shawntabrizi:
There isn't a world ending pressure to get this in, but I think this is a natural improvement to the treasury system for the new models of accessing and using coretime.

There was a project I was working on that would want this feature.

bkchr:
> @shawntabrizi do we have a live or upcoming use case for this?

This question is totally irrelevant for this pr here. With block times getting more dynamic, we need this fix. There are also other downstream users than "we" that may use or want to use it.

substrate/frame/treasury/src/lib.rs
kianenigma: I am tempted to suggest removing the parameter from `Hooks` -- even without this change, it was superfluous as you (almost) always have access to block number from `frame_system` directly.

PR #3972, opened 2024-04-03
Smooth transition to async backing

s0me0ne-unkn0wn:
Description TBD


cumulus/polkadot-parachain/src/service.rs
```diff
@@ -1168,6 +944,47 @@ fn start_lookahead_aura_consensus(
...
+			if has_unincluded_segment_api {
+				log::info!("Collation request forwarder detected async backing API");
+				return
+			}
+
+			if let Err(e) = collation_sender.send(request).await {
+				log::error!("Cannot forward collation request: {e:?}");
+			}
+		}
+	});
```
bkchr, line 986: I should have been more concrete. This was not what I had in mind :see_no_evil: 

Basically I thought that we can just guard this function: https://github.com/paritytech/polkadot-sdk/blob/0f4e849e0ac2de8c9880077c085985c5f656329c/cumulus/client/consensus/aura/src/collators/lookahead.rs#L433-L462

If the runtime api exists, we call this otherwise we take whatever block is in the validation data. This way we should build one block per relay chain block.

s0me0ne-unkn0wn: Too late, I cannot stop anymore! :stuck_out_tongue_closed_eyes: 

Have a look at the latest version. You're right, your approach is better in the sense that it's simplistic and just does what is needed. But in mine, I see two advantages:
1) Absolute zero overhead after async backing support is detected (no need to call `has_api` for every block)
2) Just look what a bunch of code this refactoring brings down! Instead of having a separate consensus starter for everything, it just detects what we have and what we don't have, and behaves accordingly.

Anyway, you decide which is better, I'll prepare another PR with guarding as you propose and let them fight, but I still love this one!

bkchr: > 1. Absolute zero overhead after async backing support is detected (no need to call `has_api` for every block)

We already have the `runtime_api` in the function. We are doing tons of these calls everywhere and `has_api` just looks into the `RuntimeVersion` which is cached internally, so it doesn't involve any call into the runtime.


> 2\. Just look what a bunch of code this refactoring brings down! Instead of having a separate consensus starter for everything, it just detects what we have and what we don't have, and behaves accordingly.

The `start_generic_aura` function can also be removed in my case as it is not needed. For the other changes, I think it gets a little bit hard to follow when we also include the shell to aura transition there.

This said, I still favor my solution :sweat_smile:

s0me0ne-unkn0wn: Okay, I've dug a little bit into that. Indeed, non-existing API is not so much of a problem:

https://github.com/paritytech/polkadot-sdk/blob/0f4e849e0ac2de8c9880077c085985c5f656329c/cumulus/client/consensus/aura/src/collators/lookahead.rs#L456-L458

This just returns `None` if API doesn't exist, nothing explodes. So, indeed, the chain not having that API still progresses even when run on the lookahead collator. The only problem is the runtime panics we were discussing before, which happens here:

https://github.com/paritytech/polkadot-sdk/blob/0f4e849e0ac2de8c9880077c085985c5f656329c/cumulus/pallets/parachain-system/src/lib.rs#L1337

I'll definitely need some time to understand how that part works and could use some help here :thinking:

bkchr: I would assume that this: https://github.com/paritytech/polkadot-sdk/blob/0f4e849e0ac2de8c9880077c085985c5f656329c/cumulus/client/consensus/aura/src/collators/lookahead.rs#L315 is the problem. It will just try to produce two blocks, while it actually should only allow one in maximum.

Basically `can_build_upon` should only return `Some()` once.

s0me0ne-unkn0wn:
After some investigation, it turned out it's not really needed. The lookahead collator perfectly handles everything by itself. So, it's enough to merge #3630, and we're good to go. Closing.

PR #3974, opened 2024-04-03
Make `BlockNumberProvider::set_block_number` available in tests / std

shawntabrizi:
This function is currently only exposed to runtime-benchmarks, where it should be available in all tests.

Also made it available in `std` because this is how `set_block_number` works in FRAME System, however, not sure if it is needed in the latest std/no_std paradigm.


substrate/primitives/runtime/src/traits.rs
```diff
@@ -2322,12 +2322,12 @@ pub trait BlockNumberProvider {
...
 
-	/// Utility function only to be used in benchmarking scenarios, to be implemented optionally,
-	/// else a noop.
+	/// Utility function only to be used in benchmarking scenarios or tests, to be implemented
+	/// optionally, else a noop.
 	///
 	/// It allows for setting the block number that will later be fetched
 	/// This is useful in case the block number provider is different than System
-	#[cfg(feature = "runtime-benchmarks")]
+	#[cfg(any(feature = "std", feature = "runtime-benchmarks", test))]
```
bkchr, line 2330:
```suggestion
	#[cfg(any(feature = "std", feature = "runtime-benchmarks"))]
```

`test` feature is useless as this is only activated when running the tests inside of this crate.

bkchr (approved):

substrate/primitives/runtime/src/traits.rs
shawntabrizi: yeah, forgot

acatangiu (approved):

ggwpez (approved):

Issue #3295, opened 2024-02-12
[RelEng] Aggregate PrDocs into CHANGELOG


ggwpez:
There should be a CI action that aggregates all PrDoc files into a CHANGELOG for the next stable release.  
Can maybe reuse the one from the fellowship.

bkchr:
@EgorPopelyaev was working on this, at least the aggregation and we have already used it for 1.7.0 AFAIK.

PR #3978, opened 2024-04-04
Github workflow to automate release draft creation

EgorPopelyaev:
This PR introduces the github flow which will create a release draft automatically when the rc tag is pushed. The flow contains the following steps:

- Gets the info about rust version used to build the node
- Builds the runtimes using `srtool`
- Extracts the info about each runtime 
- Aggregates the changelog from the prdocs
- Creates the release draft containing all the info related to the release (changelog, runtimes, rust versions)
- Attaches the runtimes to the draft
- Posts the message to the RelEng internal channel to inform that the build is done.

Related to the #3295


alvicsam (approved):

Morganamilo (approved):

rcny (approved):

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Current release/rc version
+
+jobs:
+  get-rust-versions:
+    runs-on: ubuntu-latest
+    container:
+      image: paritytech/ci-unified:bullseye-1.75.0-2024-01-22-v20240109
```
rcny, line 18: Do you really want to depend on an actual CI image to do that? What about the image version upgrade policy here?

rcny (changes requested):

.github/workflows/release-30_publish_release_draft.yml
rcny: https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-unified/Dockerfile#L4
For example, toolchain versions could be extracted from this `Dockerfile` part. In case if you want to continue exact version pinning in the workflow definition, it still needs to be known how the CI image version bumping would be handled here (possible options could be "with the image update in the rest of CI" / "after updating the image in CI", etc.)

EgorPopelyaev: Not really, but haven't found a better way. I track the versions of the CI image as it used in our cleanroom scripts and upgrade it when the new version is up. But it is a manual process as well

.github/scripts/common/lib.sh
```diff
@@ -369,7 +369,7 @@ function relative_parent() {
 # used as Github Workflow Matrix. This call is exposed by the `scan` command and can be used as:
 # podman run --rm -it -v /.../fellowship-runtimes:/build docker.io/chevdor/srtool:1.70.0-0.11.1 scan
 function find_runtimes() {
-    libs=($(git grep -I -r --cached --max-depth 20 --files-with-matches 'construct_runtime!' -- '*lib.rs'))
+    libs=($(git grep -I -r --cached --max-depth 20 --files-with-matches '[frame_support::runtime]!' -- '*lib.rs'))
```
ggwpez, line 372: I think we only moved over Rococo and Westend to the new syntax. For example the rococo-bridge-hub still uses the old one.  
Not sure if that one is important, otherwise maybe keep both?

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+    container:
+      image: paritytech/ci-unified:bullseye-1.75.0-2024-01-22-v20240222
+    outputs:
+      rustc-stable: ${{ steps.get-rust-versions.outputs.stable }}
+      rustc-nightly: ${{ steps.get-rust-versions.outputs.nightly }}
+    steps:
+      - id: get-rust-versions
+        run: |
+          echo "stable=$(rustc +stable --version)" >> $GITHUB_OUTPUT
+          echo "nightly=$(rustc +nightly --version)" >> $GITHUB_OUTPUT
```
ggwpez, line 26: We dont use nightly anymore for runtime building. Do we need it anywhere here in the script?

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+
+  build-runtimes:
+    uses: "./.github/workflows/srtool.yml"
+    with:
+       excluded_runtimes: "substrate-test bp cumulus-test kitchensink minimal-template parachain-template penpal polkadot-test seedling shell frame-try sp solochain-template"
```
ggwpez, line 31: If we forget to include one here, then it will also be published 🙈 ?

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+    outputs:
+      release_url: ${{ steps.create-release.outputs.html_url }}
+      asset_upload_url: ${{ steps.create-release.outputs.upload_url }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
+      - name: Prepare tooling
+        run: |
+          URL=https://github.com/chevdor/tera-cli/releases/download/v0.2.4/tera-cli_linux_amd64.deb
```
ggwpez, line 45: Is it using 0.2.4 since there is no deb package in the latest release? Is it much faster than `cargo install`?

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+          wget $URL -O tera.deb
+          sudo dpkg -i tera.deb
+          tera --version
+
+      - name: Download artifacts
+        uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
+      - name: Display structure of downloaded files
+        run: pwd; ls -la
```
ggwpez, line 53: This is just for debugging?

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Get runtime info
+        env:
+          JSON: release-notes-context/${{ matrix.chain }}-runtime/${{ matrix.chain }}-srtool-digest.json
+        run: |
+          >>$GITHUB_ENV echo ASSET=$(find ${{ matrix.chain }}-runtime -name '*.compact.compressed.wasm')
+          >>$GITHUB_ENV echo SPEC=$(<${JSON} jq -r .runtimes.compact.subwasm.core_version.specVersion)
+
+      - name: Upload compressed ${{ matrix.chain }} v${{ env.SPEC }} wasm
+        if: ${{ matrix.chain }} !~= 'rococo-parachain'
```
ggwpez, line 133: What is this doing?

prdoc/1.9.0/pr_3513.prdoc
```diff
@@ -5,6 +5,12 @@ title: Fix call enum's metadata regression
 
 doc:
   - audience: Runtime Dev
+    description: |
+      This PR fixes an issue where in the metadata of all FRAME-based chains, the documentation for
+      all extrinsic/call/transactions has been replaced by a single line saying "see Pallet::name".
+
+      Many wallets display the metadata of transactions to the user before signing, and this bug
+      might have affected this process.
   - audience: Runtime User
```
ggwpez, line 14:
```suggestion
  - audience: [Runtime User, Runtime Dev]
```

Added this syntax recently, so you dont have to copy&paste it 😄 (and made the description required)

scripts/release/build-changelogs.sh
```diff
@@ -21,7 +26,8 @@ OUTPUT="${TMP}/changelogs/$PRODUCT/$VERSION"
 echo -e "OUTPUT: \t\t$OUTPUT"
 mkdir -p $OUTPUT
 
-prdoc load -d "$PROJECT_ROOT/prdoc/$VERSION" --json > $DATA_JSON
+$ENGINE run --rm -v ${PROJECT_ROOT}:/repo paritytech/prdoc load -d "prdoc/$VERSION" --json > $DATA_JSON
+#prdoc load -d "$PROJECT_ROOT/prdoc/$VERSION" --json > $DATA_JSON
```
ggwpez, line 30:
```suggestion
```

ggwpez (approved):
Fantastic 🚀 💯

.github/workflows/release-30_publish_release_draft.yml
rcny: Yeah, we'll definitely come up with something more automated here after we'll migrate from GitLab to GHA.

rcny (approved):

.github/workflows/release-30_publish_release_draft.yml
ggwpez: You could tag the the `paritytech/ci-unified` with a `CI-latest` tag and then only use that tag for CI jobs.  
Has the downside that old jobs possibly are not re-runnable, for example when we want to do a patch release of an older version.

.github/scripts/common/lib.sh
EgorPopelyaev: Yep, I've tested it locally, and it gave all the runtimes as well, so I went for this one. If it is finem, I can keep it like this

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: If we do not use it, I think no. It is just the for info. I can delete it.

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: No, it just will be built. Filtering for the publishing is done in the `Prepare draft` step

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: Ah, yep, this one can be cleaned up

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: @rcny Vova, I changed the way to get the rust version as you suggested. Now it is parsed from the `Dockerfile`

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: It is here to exclude the `rococ-parachain` from publishing in the draft. Was made with typos here, already corrected it.

.github/workflows/release-30_publish_release_draft.yml
rcny: Nice, thanks!

Issue #3948, opened 2024-04-02
Remove non-sensical `para_id` function


eskimor:
The [para_id](https://github.com/paritytech/polkadot-sdk/blob/d6f68bb9062167537211cc05286809771fc8861a/polkadot/primitives/src/v7/mod.rs#L1090) function returns either the occupying para id or the scheduled one. This is nonsensical, if we assume core sharing paras and should be fixed.

PR #3979, opened 2024-04-04
Deprecate `para_id()` from `CoreState` in polkadot primitives

tdimitrov:
With Coretime enabled we can no longer assume there is a static 1:1 mapping between core index and para id. This mapping should be obtained from the scheduler/claimqueue on block by block basis.

This PR modifies `para_id()` (from `CoreState`) to return the scheduled `ParaId` for occupied cores and removes its usages in the code.

Closes https://github.com/paritytech/polkadot-sdk/issues/3948

### DIFF
--- a/cumulus/client/consensus/aura/src/collators/lookahead.rs
+++ b/cumulus/client/consensus/aura/src/collators/lookahead.rs
@@ -50,7 +50,7 @@ use polkadot_node_subsystem::messages::{
 };
 use polkadot_overseer::Handle as OverseerHandle;
 use polkadot_primitives::{
-	CollatorPair, CoreIndex, CoreState, Id as ParaId, OccupiedCoreAssumption,
+	AsyncBackingParams, CollatorPair, CoreIndex, CoreState, Id as ParaId, OccupiedCoreAssumption,
 };
 
 use futures::{channel::oneshot, prelude::*};
@@ -188,10 +188,14 @@ where
 
 			// TODO: Currently we use just the first core here, but for elastic scaling
 			// we iterate and build on all of the cores returned.
-			let core_index = if let Some(core_index) =
-				cores_scheduled_for_para(relay_parent, params.para_id, &mut params.overseer_handle)
-					.await
-					.get(0)
+			let core_index = if let Some(core_index) = cores_scheduled_for_para(
+				relay_parent,
+				params.para_id,
+				&mut params.overseer_handle,
+				&mut params.relay_client,
+			)
+			.await
+			.get(0)
 			{
 				*core_index
 			} else {
@@ -225,7 +229,10 @@ where
 			let parent_search_params = ParentSearchParams {
 				relay_parent,
 				para_id: params.para_id,
-				ancestry_lookback: max_ancestry_lookback(relay_parent, &params.relay_client).await,
+				ancestry_lookback: async_backing_params(relay_parent, &params.relay_client)
+					.await
+					.map(|c| c.allowed_ancestry_len as usize)
+					.unwrap_or(0),
 				max_depth: PARENT_SEARCH_DEPTH,
 				ignore_alternative_branches: true,
 			};
@@ -463,21 +470,19 @@ where
 	Some(SlotClaim::unchecked::<P>(author_pub, slot, timestamp))
 }
 
-/// Reads allowed ancestry length parameter from the relay chain storage at the given relay parent.
-///
-/// Falls back to 0 in case of an error.
-async fn max_ancestry_lookback(
+/// Reads async backing parameters from the relay chain storage at the given relay parent.
+async fn async_backing_params(
 	relay_parent: PHash,
 	relay_client: &impl RelayChainInterface,
-) -> usize {
+) -> Option<AsyncBackingParams> {
 	match load_abridged_host_configuration(relay_parent, relay_client).await {
-		Ok(Some(config)) => config.async_backing_params.allowed_ancestry_len as usize,
+		Ok(Some(config)) => Some(config.async_backing_params),
 		Ok(None) => {
 			tracing::error!(
 				target: crate::LOG_TARGET,
 				"Active config is missing in relay chain storage",
 			);
-			0
+			None
 		},
 		Err(err) => {
 			tracing::error!(
@@ -486,7 +491,7 @@ async fn max_ancestry_lookback(
 				?relay_parent,
 				"Failed to read active config from relay chain client",
 			);
-			0
+			None
 		},
 	}
 }
@@ -496,6 +501,7 @@ async fn cores_scheduled_for_para(
 	relay_parent: PHash,
 	para_id: ParaId,
 	overseer_handle: &mut OverseerHandle,
+	relay_client: &impl RelayChainInterface,
 ) -> Vec<CoreIndex> {
 	// Get `AvailabilityCores` from runtime
 	let (tx, rx) = oneshot::channel();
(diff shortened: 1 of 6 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5775997

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -681,14 +681,23 @@ pub(crate) async fn handle_active_leaves_update<Context>(
 				.map_err(JfyiError::FetchValidatorGroups)?
 				.1;
 
+		let maybe_claim_queue = fetch_claim_queue(ctx.sender(), new_relay_parent)
```
sandreim, line 684: now we are fetching it twice, first time here and also in `determine_groups_per_para`

polkadot/node/network/statement-distribution/src/v2/mod.rs
```diff
@@ -752,26 +761,39 @@ fn find_active_validator_state(
...
+			.flatten()
+	};
 	let group_validators = groups.get(our_group)?.to_owned();
 
 	Some(LocalValidatorState {
 		active: Some(ActiveValidatorState {
 			index: validator_index,
 			group: our_group,
-			assignment: para,
+			assignment: core_assignment,
```
sandreim, line 796: This implies that we gonna check our local assignment vs what a local statement core index is. But that requires extracing the core index which is done in backing.

I doubt that this change is useful, but only more work to be done in this PR, as the backing group is mapped 1:1 to a parachain at any relay parent.

polkadot/node/core/provisioner/src/tests.rs
```diff
@@ -918,7 +918,11 @@ mod select_candidates {
 		let committed_receipts: Vec<_> = (0..mock_cores.len())
 			.map(|i| {
 				let mut descriptor = dummy_candidate_descriptor(dummy_hash());
-				descriptor.para_id = mock_cores[i].para_id().unwrap();
+				descriptor.para_id = if let Scheduled(scheduled_core) = &mock_cores[i] {
+					scheduled_core.para_id
+				} else {
+					panic!("`moc_cores` is not initialized with `Scheduled`?")
```
sandreim, line 924:
```suggestion
					panic!("`mock_cores` is not initialized with `Scheduled`?")
```

cumulus/client/consensus/aura/src/collators/lookahead.rs
```diff
@@ -522,11 +525,49 @@ async fn cores_scheduled_for_para(
 		},
 	};
 
+	// Get `AsyncBackingParams` from runtime
+	let (tx, rx) = oneshot::channel();
+	let request = RuntimeApiRequest::AsyncBackingParams(tx);
+	overseer_handle
```
sandreim, line 531: Might be better to use `load_abridged_host_configuration` for the async backing params so the collator works without full node. It should also avoid spawning the runtime in favor of a faster direct storage read.

polkadot/node/network/statement-distribution/src/v2/mod.rs
sandreim: ok, i think `core_assignment` is vert misleading in this case, why not keep it as `para` ? I assumed we want the core index in there.

sandreim: `core_assigned_to_para` is what you really wanted :D

tdimitrov: Yes, definitely a bad choice. I'll fix this.

polkadot/node/network/statement-distribution/src/v2/mod.rs
tdimitrov: Fixed

cumulus/client/consensus/aura/src/collators/lookahead.rs
tdimitrov: Done

prdoc/pr_3979.prdoc
```diff
@@ -0,0 +1,19 @@
...
+doc:
+  - audience: "Node Dev"
+    description: |
+      With Coretime there is no more a static 1:1 mapping between core index and para id. This
+      mapping should be obtained from the scheduler/claimqueue on block by block basis. This change
+      removes `para_id()` from `CoreState` as it doesn't serve its purpose anymore.
+
+crates:
+  - name: polkadot-primitives
+    bump: major
```
tdimitrov, line 15: Not sure about this one. We are removing a pub method so it's should be a major bump?

sandreim: I think so, it is a breaking change

prdoc/pr_3979.prdoc
```diff
@@ -0,0 +1,19 @@
...
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Remove `para_id()` from `CoreState` in polkadot primitives
+
+doc:
+  - audience: "Node Dev"
+    description: |
+      With Coretime there is no more a static 1:1 mapping between core index and para id. This
+      mapping should be obtained from the scheduler/claimqueue on block by block basis. This change
+      removes `para_id()` from `CoreState` as it doesn't serve its purpose anymore.
```
sandreim, line 11:
```suggestion
     The motivation is that `para_id()` can be misused with Coretime which allows two parachains (A,B) to share a 
     core at an arbitrary relay chain block. This happens when a candidate of parachain A is pending availability 
     while parachain B is scheduled on the same core. In such a case the removed `para_id()` function would always
     return the id of A. Implementers should now use the `ClaimQueue` to query the scheduled `para_id`.
```

sandreim (approved):
LGTM, modulo the prdoc description.

prdoc/pr_3979.prdoc
eskimor: Good point. Let's deprecate it first then.

eskimor: And just in case people are actually using it, let's also fix it. (Returning the scheduled para_id also on an occupied core.) This should not break any code, as up until now this was the same thing.

eskimor (approved):
Approved. Module the breaking change. As a general rule, don't break things, if there is no real need.

(Sorry, I did not realize that this was publicly exposed and potentially used by others, when I  said "remove it".) :grimacing:

prdoc/pr_3979.prdoc
tdimitrov: @sandreim could you please have another look at the prdoc? Does the updated description make sense?

polkadot/primitives/src/v7/mod.rs
```diff
@@ -1087,9 +1087,12 @@ pub enum CoreState<H = Hash, N = BlockNumber> {
 
 impl<N> CoreState<N> {
 	/// If this core state has a `para_id`, return it.
```
sandreim, line 1089: Let's align the doc comment with the PRDoc, and state that it always returns the scheduled `ParaId` if any, and otherwise None.

sandreim (approved):
One more nit!

polkadot/primitives/src/v7/mod.rs
tdimitrov: Sure, forgot about the doc comment.

prdoc/pr_3979.prdoc
```diff
@@ -0,0 +1,21 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Deprecate `para_id()` from `CoreState` in polkadot primitives
+
+doc:
+  - audience: "Node Dev"
+    description: |
+     The motivation is that the old implementation of `para_id()` can be misused with Coretime which allows two
```
eskimor, line 9: Given that we just fixed it, it can not actually be misused anymore. So the only reason for deprecation is that claim queue is replacing it.

PR #3426, opened 2024-02-21
[dnm] detect closed substream

altonen:
Apply patch https://github.com/paritytech/substrate/pull/13396 to detect closed notification substreams without writing data into them. Disable peer eviction, as closed substream detection should serve the same purpose without causing sudden peer count drops to zero.

Issue #3346, opened 2024-02-15
Peers all drops when a bad peer try to connect


crystalin:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

Running a regular node (Moonriver, v0.35.0, based on polkadot-sdk 1.3.0), which maintains 40 peers (configured 20 in, 20 outs), when a bad node connect with a bad genesis, all the peers get dropped:

```
2024-02-15 23:06:31 [🌗] 💤 Idle (40 peers), best: #6153831 (0xf3ff…06ed), finalized #6153829 (0xb749…a45a), ⬇ 12.5kiB/s ⬆ 12.5kiB/s
2024-02-15 23:06:36 [Relaychain] Received imported block via RPC: #21894430 (0xef28…3f0c -> 0x59b1…1fad)
2024-02-15 23:06:36 [Relaychain] Failed to handle incoming network message err=ImplicitViewFetchError(ProspectiveParachainsUnavailable)
2024-02-15 23:06:36 [Relaychain] Received imported block via RPC: #21894430 (0xef28…3f0c -> 0x992a…2263)
2024-02-15 23:06:36 [Relaychain] Failed to handle incoming network message err=ImplicitViewFetchError(ProspectiveParachainsUnavailable)
2024-02-15 23:06:36 [🌗] 💤 Idle (40 peers), best: #6153831 (0xf3ff…06ed), finalized #6153829 (0xb749…a45a), ⬇ 11.1kiB/s ⬆ 11.4kiB/s
2024-02-15 23:06:38 [Relaychain] Received finalized block via RPC: #21894427 (0xdeae…4c01 -> 0x15ab…8cd9)
2024-02-15 23:06:40 Accepting new connection 1/100
2024-02-15 23:06:41 [🌗] 💤 Idle (40 peers), best: #6153831 (0xf3ff…06ed), finalized #6153830 (0x31fd…5597), ⬇ 15.7kiB/s ⬆ 15.8kiB/s
2024-02-15 23:06:42 [Relaychain] Received imported block via RPC: #21894431 (0x992a…2263 -> 0x0eed…e67f)
2024-02-15 23:06:42 [Relaychain] Failed to handle incoming network message err=ImplicitViewFetchError(ProspectiveParachainsUnavailable)
[...]

PR #3983, opened 2024-04-04
Detect closed notifications substream instead of evicting all peers

dmitry-markin:
This PR brings the fix https://github.com/paritytech/substrate/pull/13396 to polkadot-sdk.

In the past, due to insufficient inbound slot count on polkadot & kusama, this fix led to low peer count. The situation has improved since then after changing the default ratio between `--in-peers` & `--out-peers`.

Nevertheless, it's expected that the reported total peer count with this fix is going to be lower than without it. This should be seen as the correct number of working connections reported, as opposed to also reporting already closed connections, and not as lower count of working connections with peers.

This PR also removes the peer eviction mechanism, as closed substream detection is a more granular way of detecting peers that stopped syncing with us.

The burn-in has been already performed as part of testing these changes in https://github.com/paritytech/polkadot-sdk/pull/3426.


substrate/client/network/sync/src/engine.rs
```diff
@@ -99,23 +98,6 @@ const TICK_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(1100)
...
-/// evicted.
-///
-/// Parachain collator may incorrectly get evicted because it's waiting to receive a number of
-/// relaychain blocks before it can start creating parachain blocks. During this wait,
-/// `SyncingEngine` still counts it as active and as the peer is not sending blocks, it may get
-/// evicted if a block is not received within the first 30 seconds since the peer connected.
-///
-/// To prevent this from happening, define a threshold for how long `SyncingEngine` should wait
-/// before it starts evicting peers.
-const INITIAL_EVICTION_WAIT_PERIOD: Duration = Duration::from_secs(2 * 60);
```
lexnv, line 117: dq: The worst that can happen now is to stall indefinitely? And we presume that previously, we evicted peers because they closed their notification substreams but we didn't notice it on time (ie before the eviction timer triggered all peers to be dropped)? 🤔 

I think this is related to: https://github.com/paritytech/polkadot-sdk/issues/3346

lexnv (approved):
LGTM

substrate/client/network/sync/src/engine.rs
dmitry-markin: In order for this to happen, all peers have to keep the notification channel open and misbehave by not providing any useful data. This seems unlikely to me.

Yes, the peer eviction kicked in only if we had all notification streams half-closed, but didn't notice it.

I think we can revisit peer eviction and may be come up with a solution tracking individual peers (what has been done at some point i the past).

Also, people complained that suddenly the number of connected peers dropped to zero due to eviction. Now it should work on an individual peer level.

dmitry-markin: Also, we can keep the eviction as is for now and merge only the closed substream detection, and handle eviction in a separate PR.

@lexnv @bkchr what do you think?

bkchr: We just added this to work around the bug that this pr is fixing, thus I think it is fine to remove this. Especially as we already have seen issues with it.

bkchr (approved):

PR #3998, opened 2024-04-05
[prdoc] Update docs

ggwpez:
Updating the prdoc doc file to be a bit more useful for new contributors and adding a SemVer section.


docs/contributor/prdoc.md
```diff
@@ -65,7 +41,39 @@ The list of valid audiences is described and documented in the JSON schema as fo
...
+If you have a change that affects multiple audiences, you can either list them all, or write multiple sections and
+re-phrase the changes for each audience.
+
+## Record SemVer Changes
+
+All published crates that got modified need to have an entry in the `crates` section of your `PRDoc`. This entry tells
+the release team how to bump the crate version prior to the next release. It is very important that this information is
+correct, otherwise it could break downstream teams' code.
+
+The bump can either be `major`, `minor` or `patch`. There are exceptions for experimental and private API. These are
```
ggwpez, line 53: We thought about adding a `none` option for tests/typos etc. Do you think that is useful @bkchr?  
I was convinced in the past, but now i think we can also just do it idiomatically and do patch. I think crates like serde also do patch bumps for typos.

liamaharon: `patch` for changing a test file or whatever seems overkill to me. but happy to do either.

docs/contributor/prdoc.md
```diff
@@ -1,55 +1,31 @@
...
-## Tooling
-
-Users might find the following helpers convenient:
-- Setup VSCode to be aware of the prdoc schema: see [using VSCode](https://github.com/paritytech/prdoc#using-vscode)
-- Using the `prdoc` cli to:
-  - generate a `PRDoc` file from a [template defined in the Polkadot SDK
-    repo](https://github.com/paritytech/polkadot-sdk/blob/master/prdoc/.template.prdoc) simply providing a PR number
-  - check the validity of one or more `PRDoc` files
+When creating a PR, the author needs to decide with the `R0-silent` label whether the PR has to contain a prdoc. The
+`R0` label should only be placed for No-OP changes like correcting a typo in a comment or a doc file. If unsure, ping
```
liamaharon, line 10: doc change I think needs patch, so new version gets deployed for rust-analyzer and crates.io?

liamaharon (approved):

docs/contributor/prdoc.md
ggwpez: Ah yes, with "doc file" i meant something like markdown, but that could also be included into a crate release. Will change.

docs/contributor/prdoc.md
bkchr: > I think crates like serde also do patch bumps for typos.

Patch for typos in the docs makes sense, patch for adding/changing a test not that much ;) 

As Morgan is saying `None` being the technical correct solution, but `patch` also works. As long as you explain this, choose whatever you think is the best.

ggwpez: Okay since we have the `R0-silent`, we can use the `none` as `R0-silent` label on a per-crate basis.  
Added it now.

tdimitrov (approved):

michalkucharczyk (approved):

docs/contributor/prdoc.md
```diff
@@ -1,55 +1,31 @@
...
+1. Open a Pull Request and get the PR number.
+1. Generate the file with `prdoc generate <PR_NUMBER>`. The output filename will be printed.
+1. Optional: Install the `prdoc/schema_user.json` schema in your editor, for example
+[VsCode](https://github.com/paritytech/prdoc?tab=readme-ov-file#schemas).
+1. Edit your `.prdoc` file according to the [Audience](#pick-an-audience) and [SemVer](#record-semver-changes) sections.
+1. Check your prdoc with `prdoc check -n <PR_NUMBER>`. This is optional since the CI will also check it.
 
-where <NNNN> is the PR number.
+> **Tip:** GitHub CLI and jq can be used to provide the number of your PR to generate the correct file:  
+> `prdoc generate $(gh pr view --json number | jq '.number') -o prdoc`
```
bkchr, line 26: This should be done by `prdoc` automatically.

docs/contributor/prdoc.md
```diff
@@ -65,7 +41,41 @@ The list of valid audiences is described and documented in the JSON schema as fo
...
+```
+
+It means that downstream code using `frame-example-pallet` is still guaranteed to work as before, while code using
+`frame-example` might break.
+
+### Dependencies
 
-The PRDoc schema is defined in each repo and usually is quite restrictive.
-You cannot simply add a new property to a `PRDoc` file unless the Schema allows it.
+A crate that depends on another crate will automatically inherit its `major` bumps. This means that you do not need to
```
bkchr, line 78: How does the crate inherits the `major` bump?

bkchr (approved):

docs/contributor/prdoc.md
ggwpez: I think we do this as part of the release process, or @Morganamilo ?

muharem (approved):

docs/contributor/prdoc.md
bkchr: I mean you will need to figure out if the crate is in the public api to justify a major bump of the dependent crate?

ggwpez: "in the public api" is very difficult for a developer to see. It could also mean that some behaviour was changed, but all the API definitions stay the same.  

We have no way to detect this, and developers could make mistakes by not adding all transitive dependants. So for now i think we just major-bump all dependants.

Morganamilo: I ended up deciding it does not matter whether the crate is in the public API (though parity-publish can tell this ) as the dependency tree is so interwoven for polkadot. Consider we have a release where we have one and just one PRdoc that specifies `sp-io` has a major bump.

```
morganamilo@songbird % parity-publish prdoc ../polkadot-sdk ../polkadot-sdk/prdoc -d
sp-io (substrate/primitives/io):
    files
    Major

morganamilo@songbird % parity-publish plan --new ../polkadot-sdk --prdoc ../polkadot-sdk/prdoc 
looking up crate data, this may take a while....
434 packages changed 433 indirect
calculating order...
looking up crates...
calculating plan...
plan generated 500 packages 344 to publish
```

We have 3 options in terms of publishing here. We could:

- just major bump `sp-io`
- major bump `sp-io` and minor bump all 142 direct dependants
- major bump `sp-io` and major bump all 343 indirect dependants

Option one doesn't work as we want our crates to use the new version so we have to at least also update direct dependants.

Option two is also bad in my opinion. Because semver compatible versions can be mixed and matched. We could then end up with multiple versions of `sp-io` in the dependency tree. Multiply that by how many crates we have and how interwoven the dependencies are. we could end up with 2?, 3?, 4?, 5? different versions of dozens of packages in theory.

Also even if this crate is not exposed in the public API of any crates. If it manages any sort of global state, then each copy of the crate would have its own global state. If crates rely on some sort of shared global state all of a sudden it's strangely not shared any more.

Option 3 is the only* clean solution to me. While it sucks to have to update so mange packages because of one change,  think it the solution is to be aware of breaking changes and try to not make the often. 

*if there's need I think we could possibly have a whitelist of crates where we want to do option 2, though option 3 should be the default.

bkchr: `sp-io` is also quite special because we can not have multiple different major version of this crate in our tree :D But yeah, what you write makes sense and we don't have that many other options. TLDR, do not break stuff :P

PR #4007, opened 2024-04-05
Improve frame umbrella crate doc experience

liamaharon:
1. Add `#[doc(no_inline)]` to frame umbrella crate re-exports that eventually resolve to `frame_support_procedural` so docs don't look like the screenshot below and instead link to the proper `frame-support` docs.
Screenshot 2024-04-05 at 20 05 01

2. Remove `"Rust-Analyzer Users: "` prefix from `frame_support_procedural` doc comments, since these doc comments are visible in the web documentation and possible to stumble upon especially when navigating from the frame umbrella crate.


ggwpez (approved):

ggwpez:
Can you also post a screenshot how it looks now, or is it only available after deploying to docs-rs?

liamaharon:
@ggwpez here 

Screenshot 2024-04-06 at 09 13 09

clicking `pallet_macros` leads to the page with correct docs.

kianenigma (approved):

PR #4012, opened 2024-04-05
sp-api: `impl_runtime_apis!` replace the use of `Self` as a type argument

### LINKED ISSUE #1890: `impl_runtime_apis! {}`: prevent the use of `Self`
Assuming some type like `type HeaderFor<T>` is in scope, you would expect `HeaderFor<Runtime>` and `HeaderFor<Self>` would be equivalent, but it is not. I am not sure of the exact reason, but this can be confusing. 

A simple patch to `impl_runtime_apis` should simply prevent the use of `Self` in this block of code.

dastansam:
closes #1890 

### Overview 

Introduces similar checker struct to `CheckTraitDecls` in `decl_runtime_apis!` - `CheckTraitImpls`. Overrides `visit::visit_type_path` to detect usage of `Self` as a type argument within the scope of `impl_runtime_apis!`.

**Note**: only prevents the usage of `Self` as a type argument in an angle bracket `<>`, as it is the only use case that fails to compile. For example, the code [below](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-westend/src/lib.rs#L1002) compiles fine:

```rs
impl BridgeMessagesConfig<WithBridgeHubRococoMessagesInstance> for Runtime {
  fn is_relayer_rewarded(relayer: &Self::AccountId) -> bool {
	  let bench_lane_id = <Self as BridgeMessagesConfig<WithBridgeHubRococoMessagesInstance>>::bench_lane_id();
	  // ...
```

### Result

Given a block of code like this:

```rs
impl_runtime_apis! {
	impl apis::Core<Block> for Runtime {
		fn initialize_block(header: &HeaderFor<Self>) -> ExtrinsicInclusionMode {
			let _: HeaderFor<Self> = header.clone();
			RuntimeExecutive::initialize_block(header)
		}
		// ...
         }
// ...
```

<details open>

<summary>Output:</summary>

```bash
$ cargo build --release -p minimal-template-node
  error: `Self` can not be used as type argument in the scope of `impl_runtime_apis!`. Use `Runtime` instead.
     --> /polkadot-sdk/templates/minimal/runtime/src/lib.rs:133:11
      |
  133 |             let _: HeaderFor<Self> = header.clone();
      |                    ^^^^^^^^^^^^^^^

  error: `Self` can not be used as type argument in the scope of `impl_runtime_apis!`. Use `Runtime` instead.
     --> /polkadot-sdk/templates/minimal/runtime/src/lib.rs:132:32
      |
  132 |         fn initialize_block(header: &HeaderFor<Self>) -> ExtrinsicInclusionMode {
```

</details>


substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -81,6 +82,19 @@ fn generate_impl_call(
 	let ptypes = params.iter().map(|v| &v.1);
 	let pborrow = params.iter().map(|v| &v.2);
 
+	// usage of `Self` is not allowed in the scope of the trait implementation
+	params.iter().try_for_each(|v| {
+		if let Type::Path(path) = &v.1 {
```
ggwpez, line 87: But this only checks it for params, and not the function body?  
Or is it allowed there? Otherwise i would rather check it with [`syn::visit`](https://docs.rs/syn/latest/syn/visit/index.html).

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5811316

bkchr:
IMO, it would make more sense to replace `Self` with the impl type of the trait. So, the user can just use `Self` and doesn't see an error.

kianenigma (commented):
I would be happy with the existing approach, but maybe we should explore @bkchr's suggestion? at least see how much harder it is? 

in any case, this is great and we will submit a tip for it, thank you!

dastansam:
> any updates?

working on it

kianenigma:
again? :)

dastansam:
> again? :)

will try to push it through this weekend, if not, I am ok if someone takes over)

pkhry:
bot fmt

pkhry:
Changed the visitor to rename explicit `Self` to `Runtime` inside `sp_api::impl_runtime_apis`

pkhry:
bot fmt

substrate/primitives/api/proc-macro/src/utils.rs
```diff
@@ -381,7 +381,6 @@ mod tests {
 		assert_eq!(cfg_benchmarks, filtered[1]);
 	}
 
-	#[test]
```
kianenigma, line 384: Why removed?

kianenigma (commented):
Needs test(s) to demonstrate the usage before a final review.

substrate/primitives/api/proc-macro/src/utils.rs
pkhry: on accident, it's back again

pkhry:
> Needs test(s) to demonstrate the usage before a final review.

added a test on that operates on a `TokenStream` to show that macro will indeed substitute `Self` for Trait.Self_ty.  

Another possible way to test would be to use something like macrotest, but it would be capturing internal details inside expanded macro output.

kianenigma (approved):
LGTM based on the provided test

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -7,7 +7,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-// 	http://www.apache.org/licenses/LICENSE-2.0
+//  http://www.apache.org/licenses/LICENSE-2.0
```
lexnv, line 10: nit; unrelated change

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -787,17 +789,55 @@ fn generate_runtime_api_versions(impls: &[ItemImpl]) -> Result<TokenStream> {
...
+
+/// Rename `Self` to `ItemImpl.self_ty` in all items.
+fn rename_self_in_trait_impls(impls: &mut [ItemImpl]) -> Result<()> {
+	impls.iter_mut().for_each(|i| {
+		let mut checker = ReplaceSelfImpl { self_ty: i.self_ty.clone() };
+		checker.replace(i);
+	});
+
+	Ok(())
+}
```
jsdw, line 826: The comment isn't really accurate as we only rename `Self` to `ItemImpl.self_ty` in generic arguments.

I don't know the rest of this code but wonder whether there is any harm in being more general ie replacing `Self` whereever it's encountered and not just in generic args?

ie something like (untested):

```rust
impl VisitMut for ReplaceSelfImpl {
	fn visit_type_mut(&mut self, t: &mut syn::Type) {
        let syn::Type::Path(p) = t else { return };

        if p.path.get_ident().is_some_and(|ident| ident == "Self") {
            *t = self.self_ty.clone()
        }
	}
}
```

If not, as a nit I think we could tidy the exzisting code slightly to something like:

```rust
impl VisitMut for ReplaceSelfImpl {
	fn visit_generic_argument_mut(&mut self, i: &mut syn::GenericArgument) {
        let GenericArgument::Type(Type::Path(p)) else { return };

        if if p.path.get_ident().is_some_and(|ident| ident == "Self") {
            *i = GenericArgument::Type(*self.self_ty.clone());
        }
	}
}
```

As long as the borrow checker is happy, at least :)

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -945,4 +985,34 @@ mod tests {
...
+					let _: HeaderFor<Self> = header.clone();
+					example_fn::<Self>(header)
+				}
+			}
+		};
+		let expected = quote::quote! {
+			impl client::Core<Block> for Runtime {
+				fn initialize_block(header: &HeaderFor<Runtime>) -> Output<Runtime> {
+					let _: HeaderFor<Runtime> = header.clone();
+					example_fn::<Runtime>(header)
```
jsdw, line 1003: ie would it currently work if I had something like this?:

`let foo: Self = ...`

jsdw:
Looks good to me but I wonder if we can simplify the code a touch and whether we should replace `Self` more generally everywhere or else tweak the comment :)

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
pkhry: i'd keep things conservative in this regard

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
pkhry: applied 
```rust
impl VisitMut for ReplaceSelfImpl {
	fn visit_type_mut(&mut self, t: &mut syn::Type) {
        let syn::Type::Path(p) = t else { return };

        if p.path.get_ident().is_some_and(|ident| ident == "Self") {
            *t = self.self_ty.clone()
        }
	}
}
```

substrate/primitives/api/proc-macro/src/impl_runtime_apis.rs
pkhry: should be removed now

jsdw (approved):

PR #4017, opened 2024-04-07
Remove deprecated `TryRuntime` subcommand

liamaharon:
Completes the removal of `try-runtime-cli` logic from `polkadot-sdk`.


prdoc/pr_4017.prdoc
```diff
@@ -0,0 +1,15 @@
...
+
+doc:
+  - audience: Node Operator
+    description: |
+      The deprecated `try-runtime` subcommand and `try-runtime-cli` crate has been completely removed.
+      If you are not already, you should be using the standalone CLI at https://github.com/paritytech/try-runtime-cli.
+
+crates:
+  - name: polkadot-parachain-bin
+    bump: patch
```
liamaharon, line 11: Making these patch because the subcommands were already a noop.

ggwpez (approved):

substrate/bin/node/cli/src/cli.rs
```diff
@@ -61,11 +61,6 @@ pub enum Subcommand {
 	#[command(subcommand)]
 	Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
-	/// Try-runtime has migrated to a standalone CLI
-	/// (<https://github.com/paritytech/try-runtime-cli>). The subcommand exists as a stub and
-	/// deprecation notice. It will be removed entirely some time after January 2024.
```
kianenigma, line 66: damn we didn't mark this as deprecated before, so it will be a breaking chaneg in all node side code of all teams... wdyt about making it deprecated now, and removing one version later? we can still fully remove it from our codebase.

kianenigma (approved):
Looks good, just worried that we skipped one release in which we emit a warning, and then break everyone's code.

substrate/bin/node/cli/src/cli.rs
liamaharon: The actual command has been a noop for a few releases, it only logged that it's deprecated with instructions to use the dedicated cli :)

Polkadot-Forum:

PR #3512, opened 2024-02-28
[FRAME] Free Standing pallet benchmark runner 

ggwpez:
This MR contains two major changes and some maintenance cleanup.  
Status: trying to get all benchmarks to run with the RA genesis state.

## 1. Free Standing Pallet Benchmark Runner

Closes https://github.com/paritytech/polkadot-sdk/issues/3045

Introduces a new binary crate: `frame-benchmarking-freestanding-cli` (stupid name).  
It allows to directly benchmark a WASM blob - without needing a node or chain spec.  Example:  
```sh
cargo r -p frame-benchmarking-freestanding-cli -- --runtime target/release/wbuild/rococo-runtime/rococo_runtime.compact.compressed.wasm --pallet "" --extrinsic ""
```

This makes it much easier to generate pallet weights and should allow us to remove bloaty code from the node.  
It should work for all FRAME runtimes that dont use 3rd party host calls.

## 2. Building the Benchmark Genesis State in the Runtime

Closes https://github.com/paritytech/polkadot-sdk/issues/2664

This add `--runtime` and `--runtime-builder=none|runtime|spec` arguments to the `benchmark pallet` command to make it possible to use generate the genesis storage by the runtime. This can be used with the node and the freestanding benchmark runner. It utilizes the new `GenesisBuilder` RA and depends on having https://github.com/paritytech/polkadot-sdk/pull/3412 deployed.

## 3. Maintenance Change

- Add percent progress print when running benchmarks
- Dont immediately exot on benchmark error but try to run as many as possible and print errors last.
- Only require `Option<Box<dyn ChainSpec>>` instead of `sc_service::Configuration` for `PalletCmd`.

PR #4020, opened 2024-04-08
benchmarking: complete automatic weight sanity check PR

dastansam:
should complete #1493 


substrate/frame/benchmarking/src/utils.rs
```diff
@@ -230,6 +230,27 @@ pub struct BenchmarkMetadata {
 	pub pov_modes: Vec<(Vec<u8>, Vec<u8>)>,
 }
 
+/// Exhaustive benchmark info provided by the runtime.
+#[derive(Encode, Decode, Default, Clone, PartialEq, Debug, TypeInfo)]
+pub struct BenchmarkInfo {
```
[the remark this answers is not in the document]
ggwpez: `RuntimeBenchmarkInfo`? dunno honestly.

substrate/frame/benchmarking/src/utils.rs
```diff
@@ -230,6 +230,27 @@ pub struct BenchmarkMetadata {
...
+/// For backwards compatibility.
+impl From<(Vec<BenchmarkList>, Vec<StorageInfo>)> for BenchmarkInfo {
+	fn from((list, storage_info): (Vec<BenchmarkList>, Vec<StorageInfo>)) -> Self {
+		Self { list, storage_info, max_extrinsic_weight: None, db_weight: None }
+	}
+}
+
 sp_api::decl_runtime_apis! {
 	/// Runtime api for benchmarking a FRAME runtime.
 	pub trait Benchmark {
```
ggwpez, line 256:
```suggestion
	#[api_version(2)]
	pub trait Benchmark {
```

substrate/frame/benchmarking/src/utils.rs
```diff
@@ -238,7 +259,10 @@ sp_api::decl_runtime_apis! {
 		/// Parameters
 		/// - `extra`: Also list benchmarks marked "extra" which would otherwise not be
 		///            needed for weight calculation.
-		fn benchmark_metadata(extra: bool) -> (Vec<BenchmarkList>, Vec<StorageInfo>, Weight, RuntimeDbWeight);
+		///
+		/// Returns:
+		/// - `BenchmarkInfo` - Complete benchmark metadata available for this runtime.
+		fn benchmark_metadata(extra: bool) -> BenchmarkInfo;
```
ggwpez, line 265: Normally, we keep the old version of the runtime API around. But given that benchmarks operate on newly build runtimes, i guess its fine.

substrate/utils/frame/benchmarking-cli/src/pallet/command.rs
```diff
@@ -254,11 +254,9 @@ impl PalletCmd {
 		.execute()
 		.map_err(|e| format!("{}: {}", ERROR_METADATA_NOT_FOUND, e))?;
 
-		let (list, storage_info, max_extrinsic_weight, db_weight) =
```
ggwpez, line 257: Not sure how, but i think its possible to query the runtime for the version of the Runtime API.  
Would be good to first query the version of the api and check that it is Version 2. Otherwise the returned data here is of the wrong format.

ggwpez: The line was added since the base branch of this merge request is not master, but the branch from Daan. Should maybe be changed?

I just realized that if we do not keep the old version declaration, then the [omni-bencher](https://github.com/paritytech/polkadot-sdk/pull/3512) will not work with the runtimes of most parachains since they are using the old version.

So if possible:
- Check the version of the Runtime API
- if the version is 1 then dont do the weight-sanity check
- if the version is 2 then do the weight-sanity check

Then it should still work with all runtimes.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5899116

prdoc/pr_4020.prdoc
```diff
@@ -0,0 +1,65 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: "Benchmarking: automatic weight sanity check"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR introduces a new sanity weight check during benchmark execution to ensure that the weight of the function/extrinsic does not exceed the given limit.
```
ggwpez, line 9:
```suggestion
      This PR introduces a new sanity weight check after benchmark execution to ensure that the weight of the function/extrinsic does not exceed the block limit. 
```

prdoc/pr_4020.prdoc
```diff
@@ -0,0 +1,65 @@
...
+      --- a/path/to/runtime.rs
+      +++ b/path/to/runtime.rs
+      @@ -1,3 +1,4 @@
+      -fn benchmark_metadata(extra: bool) -> (Vec<BenchmarkList>, Vec<StorageInfo>) {
+      +fn benchmark_metadata(extra: bool) -> frame_benchmarking::RuntimeBenchmarkInfo {
+          /// ...
+          +let max_extrinsic_weight = RuntimeBlockWeights::get().per_class.get(DispatchClass::Normal).max_extrinsic.unwrap();
+          +let db_weight: frame_support::weights::RuntimeDbWeight = <Self as frame_system::Config>::DbWeight::get();
+          -(list, storage_info)
+          +RuntimeBenchmarkInfo {list, storage_info, max_extrinsic_weight: Some(max_extrinsic_weight), db_weight: Some(db_weight) }
```
ggwpez, line 32:
```suggestion
          +RuntimeBenchmarkInfo {list, storage_info, max_extrinsic_weight: Some(max_extrinsic_weight), db_weight: Some(db_weight), ..Default::default() }
```
We can be a bit smarter here and mark the `RuntimeBenchmarkInfo` struct as `#[non_exhaustive]`. Then we can add more `Option` fields and stay forwards compatible.

substrate/frame/benchmarking/src/utils.rs
```diff
@@ -227,15 +230,40 @@ pub struct BenchmarkMetadata {
 	pub pov_modes: Vec<(Vec<u8>, Vec<u8>)>,
 }
 
+/// Exhaustive benchmark info provided by the runtime.
+#[derive(Encode, Decode, Default, Clone, PartialEq, Debug, TypeInfo)]
+pub struct RuntimeBenchmarkInfo {
```
ggwpez, line 235:
```suggestion
#[non_exhaustive]
pub struct RuntimeBenchmarkInfo {
```

substrate/frame/benchmarking/src/utils.rs
```diff
@@ -227,15 +230,40 @@ pub struct BenchmarkMetadata {
 	pub pov_modes: Vec<(Vec<u8>, Vec<u8>)>,
 }
 
+/// Exhaustive benchmark info provided by the runtime.
```
ggwpez, line 233:
```suggestion
/// Benchmark info provided by the runtime.
///
/// This struct is marked a `#[non_exhaustive]` to ensure that doing a simple `..Default::default()` in
/// the constructor will allow downstream teams to stay forward compatible to future changes.
```

prdoc/pr_4020.prdoc
```diff
@@ -0,0 +1,65 @@
...
+          list_benchmarks!(list, extra);
+
+          let storage_info = AllPalletsWithSystem::storage_info();
+
+          -(list, storage_info)
+          +(list, storage_info).into()
+      }
+      ```
+
+      Once the node is compiled with `runtime-benchmarks` feature, you can pass the `--sanity-check` flag to the benchmarking CLI or `ombi-bencher`:
```
ggwpez, line 55:
```suggestion
      Once the node is compiled with `runtime-benchmarks` feature, you can pass the `--sanity-weight-check` flag to the benchmarking CLI or `ombi-bencher`:
```

substrate/client/cli/src/arg_enums.rs
```diff
@@ -65,6 +65,22 @@ impl std::fmt::Display for WasmExecutionMethod {
 	}
 }
 
+/// How to output the result of the sanity weight check and what to do when it fails.
+#[allow(missing_docs)]
```
ggwpez, line 69: Is it possible to remove the missing doc allow here?

substrate/utils/frame/benchmarking-cli/src/pallet/command.rs
```diff
@@ -255,20 +248,75 @@ impl PalletCmd {
...
+							&executor,
+							"Benchmark_benchmark_metadata",
+							&(self.extra).encode(),
+							&mut Self::build_extensions(executor.clone()),
+							&runtime_code,
+							CallContext::Offchain,
+						),
+						ERROR_METADATA_NOT_FOUND,
+					)?;
+				old_benchmark_data.into()
```
ggwpez, line 289: Yes this looks good. So now it should be possible to have old runtimes still work.  
I guess to be really precise, we should keep the old `fn benchmark_metadata` around in the `pub trait Benchmark`, since this is what we call into.  
There is a `changed_in` attribute that you can search for in the code base and add.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+	fn max_component(parameter: &ComponentSlope, component_ranges: &Vec<ComponentRange>) -> u64 {
+		for component in component_ranges {
+			if parameter.name == component.name {
+				return component.max.into()
+			}
+		}
+		0
+	}
+
+	color_print::cprintln!(
```
ggwpez, line 205: Should just do `log::warn!` as we cannot assume that the output terminal actually supports colors.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+		"\n<s>Sanity Weight Check 🧐:</> each extrinsic's weight function is executed \
```
ggwpez, line 206: This whole print is a bit pointless, i think it should only print when something is wrong.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+		each extrinsic is checked whether it will fit in an empty (meaning; empty of \
+		`DispatchClass::Normal` extrinsics) block. \n\nProvided values for the maximum extrinsic weight: \
+		{:?} (ref_time), {:?} (proof_size)\n\n<u>Results:</>\n",
+		max_extrinsic_weight.ref_time(),
+		max_extrinsic_weight.proof_size()
+	);
+	let mut sanity_weight_check_passed = true;
+	// Loop through all benchmark results.
+	for ((pallet, instance), results) in all_results.iter() {
+		println!("Pallet: {}\nInstance: {}\n", pallet, instance);
```
ggwpez, line 218: Was this code moved or added? If it was added then please delete again.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+				color_print::cprintln!(
+					"<r>The following extrinsic exceeds the maximum extrinsic weight:</>",
+				);
+			}
+			color_print::cprintln!("- <s>'{}'</>: {:?}\nPercentage of max. extrinsic weight: {:.2}% (ref_time), {:.2}% (proof_size)\n", 
+				result.name,
+				total_weight,
+				(total_weight.ref_time() as f64 / max_extrinsic_weight.ref_time() as f64) * 100.0,
+				(total_weight.proof_size() as f64 / max_extrinsic_weight.proof_size() as f64) * 100.0,
+			);
```
ggwpez, line 278: Just one `log::warn!` is enough per extrinsic. Looks like the previous code is a bit overkill.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+		}
+	}
+	if sanity_weight_check_passed {
+		color_print::cprintln!("<g>Your extrinsics passed the Sanity Weight Check 😃!</>\n");
```
ggwpez, line 282:
```suggestion
```
No needless noise.

substrate/utils/frame/benchmarking-cli/src/pallet/writer.rs
```diff
@@ -175,6 +179,121 @@ fn map_results(
...
+	} else {
+		color_print::cprintln!(
+			"<r>Your extrinsics failed the Sanity Weight Check, please review \
+			the extrinsic's logic and/or the associated benchmark function.</>\n",
+		);
```
ggwpez, line 287:
```suggestion
```

ggwpez (approved):
Going to approve since its technically correct, just some small improvements possible.

prdoc/pr_4020.prdoc
ggwpez: Yee my comment below 😄

ggwpez: Damn, i thought this was so smart. Then lets keep it as is - implementing a builder pattern here would be a bit overkill.

substrate/frame/benchmarking/src/utils.rs
ggwpez: nvm

substrate/utils/frame/benchmarking-cli/Cargo.toml
```diff
@@ -21,7 +21,7 @@ chrono = "0.4"
 clap = { version = "4.5.3", features = ["derive"] }
 codec = { package = "parity-scale-codec", version = "3.6.1" }
 comfy-table = { version = "7.1.0", default-features = false }
-handlebars = "5.1.0"
+handlebars = "5.1.0"
```
ggwpez, line 24:
```suggestion
handlebars = "5.1.0"
```

ggwpez (approved):
LGTM. cc @kianenigma in case you also want to take a look.

dastansam:
> hey @kianenigma would appreciate your review on this. thanks!

@kianenigma ping

kianenigma (commented):
Glad to see this completed! 

Can it be done without breaking the API? 

@ggwpez just worried if this will again introduce another versioing issue with omni bencher and runtimes out there. Some runtimes will not work because they are not up to date, etc etc. 

Tbh if that is the case, I don't think this feature is worth it. 

Ideally it can be added incrementally, as a new runtime api.

kianenigma:
\tip medium

Issue #922, opened 2021-10-24
xcm version auto discovery improvements


xlc:
There are few issues with the current implement of xcm version auto discovery:

- It doesn't work within `transactional` call. We revert all storage changes if the xcm message failed to sent, which reverts the update to `VersionDiscoveryQueue`.
- It always fail for the first message by default, which is poor UX.
- While there are governance calls to subscribe version, there is no permissionless way. It should allow user to trigger subscribe XCM version to all connected HRMP channel dest parachains.

Why not just automatically subscribe XCM version for all HRMP channel destination chains? Frontend can detect if the HRMP channel is created and if XCM version is discovered and pause feature until it is all good.

bkontur:
> Accept HRMP channel already triggers a XCM message. We can automatically subscribe XCM version with `HrmpChannelAccepted` message?
> 
> https://github.com/paritytech/polkadot/blob/c50bdc8f0b36d47b38341b332d75e315cdd658f8/runtime/parachains/src/hrmp.rs#L1115

@xlc It looks like I came to the same conclusion: https://github.com/paritytech/polkadot-sdk/issues/3692

bkontur:
> Accept HRMP channel already triggers a XCM message. We can automatically subscribe XCM version with `HrmpChannelAccepted` message?
> 
> https://github.com/paritytech/polkadot/blob/c50bdc8f0b36d47b38341b332d75e315cdd658f8/runtime/parachains/src/hrmp.rs#L1115

@xlc proposed solution here: https://github.com/paritytech/polkadot-sdk/pull/4025 - initiate version discovery on HRMP channel accepted message.

If we want to immediately subscribe and set version with `HrmpChannelAccepted`, then we need to:
- create RFC for changing `HrmpChannelAccepted` XCM instruction by adding XCM version here: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/src/v4/mod.rs#L537-L542  (means XCMv5)
- probably add `XcmVersion` to the all extrinsics which results in the calling [`fn accept_open_channel`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L1480)
  - and maybe this could be not enough, because e.g. governance calls `force_*` take some time and XCM version can change meantime and so on, it will maybe just complicate stuff

bkontur:
[...]

Issue #3692, opened 2024-03-14
Lowering XCM weights and fees


bkontur:
## Summary

The dedicated XCM-relevant benchmarks (`pallet_xcm`, `pallet_xcm_benchmarks::fungible`, `pallet_xcm_benchmarks::generic`) can utilize a `SendXcm` implementation at the end, which typically employs the `WrapVersion` implementation, often sourced from `pallet_xcm`. This also impacts all other extrinsics and benchmarks that utilize `SendXcm` + `WrapVersion`.

The `pallet_xcm`'s `WrapVersion` implementation adds unnecessary overhead to the weights - 2 reads and 1 write:

```
        /// Storage: `PolkadotXcm::VersionDiscoveryQueue` (r:1 w:1)
	/// Proof: `PolkadotXcm::VersionDiscoveryQueue` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
	/// Storage: `PolkadotXcm::SafeXcmVersion` (r:1 w:0)
	/// Proof: `PolkadotXcm::SafeXcmVersion` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
```

This overhead arises from checking the compatible XCM version for the destination. If no supported version is stored, we utilize `SafeXcmVersion` and store the `VersionDiscoveryQueue` key/value for later version discovery.
This process occurs **only the first** time when the **destination is not added** to the `SupportedVersion` in `pallet_xcm`.

The idea here is that this overhead should not be borne by the end user, but rather should be part of establishing the communication channel (e.g., opening HRMP).

## Proposed solution part 1 - reducing XCM weights

Implemented by: https://github.com/paritytech/polkadot-sdk/pull/3664

We utilize [`xcm_builder::EnsureDelivery`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/routing.rs#L111C1-L111C27) and `DeliveryHelper` features for XCM benchmarking, as demonstrated [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/src/benchmarking.rs#L37-L38) and [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm-benchmarks/src/lib.rs#L45-L46). These features ensure successful delivery for extrinsic benchmarks with the underlying `SendXcm` implementation.

[...]

PR #4025, opened 2024-04-08
Setup `HrmpChannelAccepted` handler with `pallet_xcm` for initiating XCM version discovery when openning HRMP channel

bkontur:
Closes: https://github.com/paritytech/polkadot-sdk/issues/922
Implements: https://github.com/paritytech/polkadot-sdk/issues/3692 (see: `Proposed solution part 2 - initiate VersionDiscoveryQueue when HRMP is accepted`).

When we utilize a parachain's `pallet_xcm` extrinsics or other functionality that employs `XcmpQueue` as an `XcmRouter`, there is a possibility of encountering a situation where it is the first message to the desired destination, and we do not know a supported XCM version of the destination. In such cases, `pallet_xcm` includes a feature called "version discovery," which is initiated. This is a one-time operation. Therefore, this fix aims to initiate "version discovery" as soon as possible when an HRMP channel is created, with best effort.

In other words, when parachains establish an HRMP channel, we immediately initiate XCM version discovery and don't wait for the very first user request.

This PR:
- introduces `AllowHrmpNotificationsFromRelayChain` barrier for allowing HRMP notifications just from the relay chain (to fulfill safety assumptions - [see](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/src/v4/mod.rs#L532))
- adds the implementation of `HandleHrmpChannelAccepted` for `pallet_xcm`, so `pallet_xcm` instance can be used for `xcm_executor::Config::HrmpChannelAcceptedHandler`(https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-executor/src/config.rs#L121-L122)
- sets `HandleHrmpChannelAccepted` for all cumulus/testnets


polkadot/xcm/xcm-builder/src/tests/barriers.rs
```diff
@@ -315,56 +315,150 @@ fn allow_subscriptions_from_should_work() {
...
+		Parachain(1).into_location(),
+		Err(ProcessMessageError::Unsupported),
+	);
+	assert_should_execute(
+		vec![UnsubscribeVersion],
+		Parachain(1).into_location(),
+		Err(ProcessMessageError::Unsupported),
+	);
 
-	let test_data = vec![
```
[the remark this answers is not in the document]
serban300: Thanks ! Will take a look.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -3049,6 +3050,35 @@ impl<T: Config> CheckSuspension for Pallet<T> {
...
 }
 
+/// The implementation of `HandleHrmpChannelAccepted` which can initiate version discovery (if
+/// needed).
+///
+/// Note: This is useful for parachains when establishing HRMP channels with other parachains.
+impl<T: Config> HandleHrmpChannelAccepted for Pallet<T> {
+	fn handle(recipient: u32) -> XcmResult {
+		// how would the actual runtime see the `recipient` as `Parachain(recipient)`
+		let dest = match T::UniversalLocation::get().global_consensus() {
```
svyatonik, line 3060: Just curious - I though that `UniversalLocation` is assumed to be started with a `GlobalConsensus`. Then why would we need that match here? A defensive programming or something else?

svyatonik (approved):

polkadot/xcm/xcm-builder/src/tests/barriers.rs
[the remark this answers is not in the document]
serban300: Looks good ! Thank you !

serban300 (approved):

polkadot/xcm/pallet-xcm/src/lib.rs
bkontur: > Just curious - I though that `UniversalLocation` is assumed to be started with a `GlobalConsensus`. Then why would we need that match here? A defensive programming or something else?

Yes, good point, exactly - defensive and generic.

Originally, I wanted to use just `let dest = ParentThen(Parachain(recipient).into()).into();`, but yeah there is a `Parent` which is also not that generic.

Yes, there is that thing with `GlobalConsensus` you mentioned, which I guess it's kind of unwritten rule or assumption that runtime's UniversalLocation starts with `GlobalConsensus` - I don't remember exactly but without `GlobalConsensus` something fails in `XcmExecutor` (maybe just `ExportMessage` or something with `WithComputedOrigin`...)

The Hrmp notifications have just `paraId: u32` and we need to convert it to `xcm::Location`, so I was thinking this way:
`*Parachain(recipient)*.relative_to(&T::UniversalLocation::get());`.


```
        #[test]
	fn test() {
		// E.g. we are on AssetHub
		frame_support::parameter_types! {
			pub UniversalLocation: InteriorLocation = [GlobalConsensus(Rococo), Parachain(1000)].into();
		}

		// We opened channel with 1003
		let recipient = 1003;

		// convert 1003 from the point of view of `UniversalLocation`
		let dest = match UniversalLocation::get().global_consensus() {
			Ok(global_consensus) =>
				Junctions::from([GlobalConsensus(global_consensus), Parachain(recipient)]),
			Err(_) => Junctions::from([Parachain(recipient)]),
		}.relative_to(&UniversalLocation::get());

		// expected result, how AssetHub sees sibling parachain
		assert_eq!(dest, Location { parents: 1, interior: Junctions::X1([Parachain(1003)].into()) });
	}
```

And when I was checking other runtimes, I saw that we have some minimal template without `GlobalConsensus`: https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/runtime/src/configs/xcm_config.rs#L29, so thats why I ended up with that match.

@svyatonik Do you have any other idea how to handle a conversion from `paraId: u32` to the `Location` with `T::UniversalLocation`? I was thinking about some adapter with which would wrap `Convert<u32, Location>`, `Self::get_version_for` with `Self::note_unknown_version`, but it could just complicate things and not sure if it is worth.

acatangiu: > I though that UniversalLocation is assumed to be started with a GlobalConsensus.

Yes, this is ideally where we should get, but not all parachains have done this yet..

Everything except bridging works fine without it too, so there is also no big incentive to do it.

acatangiu: > And when I was checking other runtimes, I saw that we have some minimal template without GlobalConsensus: https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/runtime/src/configs/xcm_config.rs#L29

We should sanitize whole repo and make sure all our runtimes test, mock, etc have the `GlobalConsensus` too so we create a consistent best-practice.

svyatonik: > Originally, I wanted to use just let dest = ParentThen(Parachain(recipient).into()).into();, but yeah there is a Parent which is also not that generic.

But OTOH we know that this notification comes from the relay chain (because it notifies us about new HRMP channels), right? So maybe `Parent` would be enough here :) Not sure if it'll be changed

> And when I was checking other runtimes, I saw that we have some minimal template without GlobalConsensus: https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/runtime/src/configs/xcm_config.rs#L29, so thats why I ended up with that match.

Maybe we should fix that in this case. E.g. some bridging adapters (that's where I got this assumption - e.g. `UnpaidRemoteExporter`) are assuming that `UniversalLocation` starts with `GlobalConsensus`. But that's not related to this PR ofc.

Thank you for the explanation - I've just been curios about that && I wasn't planning to request changes or smth like that :)

acatangiu (approved):

acatangiu:
@bkontur is this good to merge?

bkontur:
> @bkontur is this good to merge?

not yet, I want to check comments and add also closing part, where we clean storage for unneded stuff

polkadot/xcm/pallet-xcm/src/lib.rs
bkontur: > > And when I was checking other runtimes, I saw that we have some minimal template without GlobalConsensus: https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/runtime/src/configs/xcm_config.rs#L29
> 
> We should sanitize whole repo and make sure all our runtimes test, mock, etc have the `GlobalConsensus` too so we create a consistent best-practice.

Good point, attempt to sanitize (where possible): https://github.com/paritytech/polkadot-sdk/pull/4025/commits/9c5a91893a4fe37561bba3140e6fea4a4a858391

bkontur:
- [ ] investigating one more test if this is enough and aligned with `process_hrmp_open_channel_requests` / `initializer_on_new_session`

bkontur:
Change to draft, because `HrmpNotificationAccepted` won't work and need to find another solution: https://github.com/paritytech/polkadot-sdk/issues/922#issuecomment-2058868951

Opened another PR: https://github.com/paritytech/polkadot-sdk/pull/4156 which contains still relevant changes but with reverted `HrmpNotificationAccepted` handler stuff.

Issue #3576, opened 2024-03-05
Add new staging Runtime API: `candidates_pending_availability`


sandreim:
This is mandatory for elastic scaling support in Cumulus. This new API should obsolete the existing `candidate_pending_availability` API which just returns the first candidate as per https://github.com/paritytech/polkadot-sdk/pull/3479 .

Should work just like it's predecessor, but return all candidates pending availability:
`fn candidates_pending_availability(para_id: ParaId) -> Vec<CommittedCandidateReceipt<Hash>>`

More context about usage in Cumulus:
- pov-recovery
- block import (`handling handle_empty_block_announce_data`)

alindima:
for prototyping, we could use the backing_state API for this. it's already returning the vector of candidates pending availability for that paraid. arguably, it's less performant because it also returns some Contraints, which includes xcm messages

sandreim:
It includes the commitments for all candidates pending availabiliy anyway, so that can be pretty huge with code upgrade and all messages sent by the parachain. But yeah, it's good for the elastic scaling prototype.

PR #4027, opened 2024-04-08
Runtime API: introduce `candidates_pending_availability`

sandreim:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3576

Required by elastic scaling collators.
Deprecates old API: `candidate_pending_availability`.

TODO:
- [x] PRDoc


last_test.yaml
```diff
@@ -0,0 +1,15 @@
+TestConfiguration:
+- objective: !DataAvailabilityRead
+    fetch_from_backers: true
+  n_validators: 1000
```
sandreim, line 4: oops!

polkadot/zombienet_tests/async_backing/004-async-backing-6s.toml
```diff
@@ -0,0 +1,47 @@
+[settings]
```
sandreim, line 1: more oops

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5824879

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5829204

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5836408

polkadot/runtime/parachains/src/inclusion/mod.rs
```diff
@@ -1104,6 +1104,24 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	/// Returns all the `CommittedCandidateReceipt` pending availability for the para provided, if
+	/// any.
+	pub(crate) fn candidates_pending_availability(
```
sandreim, line 1109: Maybe we'd want to provide a parameter which is an enum to return `Last' or `All` depending on needs. These candidates can be pretty big due to messages and code upgrades, so tailoring the return type would be useful. WDYT ?

polkadot/primitives/src/runtime_api.rs
```diff
@@ -288,5 +288,10 @@ sp_api::decl_runtime_apis! {
 		/// Claim queue
 		#[api_version(11)]
 		fn claim_queue() -> BTreeMap<CoreIndex, VecDeque<ppp::Id>>;
+
+		/***** Added in v11 *****/
+		/// Elastic scaling support
+		#[api_version(11)]
```
alindima, line 294: is it fine to reuse the api version? haven't seen this before

polkadot/runtime/parachains/src/inclusion/mod.rs
alindima: If we want to be as generic as possible we could have an optional count. If the count is None, return all of them.
Anyway, it's a matter of how is this being used in cumulus.

AFAICT, in pov-recovery, we need all candidates.
OTOH, in `handle_empty_block_announce_data`, we only need the output head data of the last candidate, which can be retrieved already using the much lighter `persisted_validation_data` runtime API (which force enacts all candidates before returning the head data, so the result is the same).

@skunert WDYT?

polkadot/primitives/src/runtime_api.rs
sandreim: Yeah, it is per method. Check
Above version 7 which added async backing.

alindima: I guess this can be done only if there wasn't a runtime release in the meantime (between the claimqueue API and this API). Otherwise, versioning doesn't really make sense if we can add more things in a past version

polkadot/runtime/parachains/src/inclusion/mod.rs
skunert: The `handle_empty_block_announce_data` of `RequireSecondedInBlockAnnounce` is rather niche because it is only used for the relay chain consensus, AURA parachain nodes don't use it. However, even in that scenario we would be interested in whether the block that was announced was backed by the relay chain. Could for sure happen that a block is announced that was in the backed blocks but not the latest.

For `pov-recovery` and the parent search in lookahead collator and slot-based we also need all of them.

TBH I had not thought about the different return types of them before, in the [collator parent search](https://github.com/paritytech/polkadot-sdk/blob/c3a5488d3f610222cbbeb121f8d75a7cc98177e3/cumulus/client/consensus/common/src/lib.rs#L335-L342) we will need to switch from `persisted_validation_data` to `candidates_pending_availability`.

DQ: In theory we could for example have three candidates on three different para branches right? Or do we have some constraint that the para block backed in a single RCB slot must be dependent on each other? If not, a `Last` parameter does not make so much sense.

alindima: > DQ: In theory we could for example have three candidates on three different para branches right? Or do we have some constraint that the para block backed in a single RCB slot must be dependent on each other? If not, a Last parameter does not make so much sense.

The relay chain won't back parachain forks, it only backs chains.
Also, for proper elastic scaling to work at full throughput, we need parachains to not produce forks

skunert: > The relay chain won't back parachain forks, it only backs chains.

Okay, makes sense. So if I still submit three different forks to the relay two will be ignored.

>  Also, for proper elastic scaling to work at full throughput, we need parachains to not produce forks

That makes sense, this was a scenario that was in my head, not something I would actively support in the collator :+1:

polkadot/primitives/src/runtime_api.rs
sandreim: Yes it is only for staging APIs

polkadot/node/subsystem-types/src/messages.rs
```diff
@@ -739,6 +739,10 @@ pub enum RuntimeApiRequest {
 	/// Fetch the `ClaimQueue` from scheduler pallet
 	/// `V11`
 	ClaimQueue(RuntimeApiSender<BTreeMap<CoreIndex, VecDeque<ParaId>>>),
+	/// Get a the candidate pending availability for a particular parachain by parachain / core
```
skunert, line 742: This comment could be fixed

skunert (approved):

polkadot/primitives/src/runtime_api.rs
sandreim: Currently production is at 10: https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L2147 

I'd  prefer to conflate all of the current staging methods into a runtime api version release (11). After discussing with @tdimitrov it seems there are no strict rules, except not adding on top of an released to prod version. 

@eskimor WDYT ?

polkadot/runtime/parachains/src/inclusion/mod.rs
sandreim: Ok, so we just keep it simple like it is right now.

polkadot/node/core/runtime-api/src/lib.rs
```diff
@@ -531,6 +537,12 @@ where
 			ver = 1,
 			sender
 		),
+		Request::CandidatesPendingAvailability(para, sender) => query!(
+			CandidatesPendingAvailability,
+			candidates_pending_availability(para),
+			ver = 1,
```
tdimitrov, line 543: This should be 11.

Add a new const [here](https://github.com/paritytech/polkadot-sdk/pull/4027/files#diff-e4ba49b765a03768afe33710f5056b84a6e069775d58945771d450d3efc638fdR782):
```rust
/// `ClaimQueue`
	pub const CANDIDATE_PENDING_AVAILABILITY_REQUIREMENT: u32 = 11;
```
and then it can be used for version checks too.

sandreim: Nice catch!

polkadot/primitives/src/runtime_api.rs
alindima: as discussed, sounds good. I thought the changes were already included in a testnet runtime release

alindima (approved):

tdimitrov (approved):

PR #4029, opened 2024-04-08
Removed unused deps from Snowbridge deps

bkontur:


cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/Cargo.toml
```diff
@@ -27,11 +27,10 @@ sp-runtime = { path = "../../../../../../../substrate/primitives/runtime", defau
 # Polkadot
 xcm = { package = "staging-xcm", path = "../../../../../../../polkadot/xcm", default-features = false }
 pallet-xcm = { path = "../../../../../../../polkadot/xcm/pallet-xcm", default-features = false }
-xcm-executor = { package = "staging-xcm-executor", path = "../../../../../../../polkadot/xcm/xcm-executor", default-features = false }
+    xcm-executor = { package = "staging-xcm-executor", path = "../../../../../../../polkadot/xcm/xcm-executor", default-features = false }
```
bkchr, line 30:
```suggestion
xcm-executor = { package = "staging-xcm-executor", path = "../../../../../../../polkadot/xcm/xcm-executor", default-features = false }
```

bkchr (approved):

claravanstaden:
@bkontur if the tests pass it should be fine. How did you check this? Did you use some tool?

claravanstaden (approved):

bkontur:
> @bkontur if the tests pass it should be fine. How did you check this? Did you use some tool?

Well, it was kind of coincidence, I am working on some fixing of `assets-common`, and I just checked where we just this deps and I saw that it is used in your `test-common`, so I looked there and there were lots of different pallets which seem strange, so I used `cargo machete` for snowbridge subtree to see other stuff. So, yeah I usually use `cargo machete` (sure, there are other tools for this) but it is not 100% sure, there could lots of false positive (sp-core, sp-std, codec, scale) because it cannot detect stuff in marcos and so on, so I just do try_remove-and-compile stuff.

serban300 (approved):

acatangiu (approved):

PR #4031, opened 2024-04-08
peer_store: Increase peer ban time until escapes banned threshold

lexnv:
This is a tiny PR to increase the time a peer remains banned.

A peer is banned when the reputation drops below a threshold.
With every second, the peer reputation is exponentially decayed towards zero.

For the previous setup:
- decaying to zero from (i32::MAX or i32::MIN) would take 948 seconds (15mins 48seconds)
- from i32::MIN to escaping the banned threshold would take 10 seconds
This means we are decaying reputation a bit too aggressive and misbehaving peers can misbehave again in 10 seconds. 
Another side effect of this is that we have encountered multiple warnings caused by a few misbehaving peers.

In the new setup:
- decaying to zero from (i32::MAX or i32::MIN) would take 3544 seconds (59 minutes)
- from i32::MIN to escaping the banned threshold would take ~69 seconds

This is a followup of:
- https://github.com/paritytech/polkadot-sdk/pull/4000.

### Testing Done
- Created a misbehaving client with [subp2p-explorer](https://github.com/lexnv/subp2p-explorer), the client is banned for approx 69seconds until it is allowed to connect again.

cc @paritytech/networking

### DIFF
--- a/substrate/client/network/src/peer_store.rs
+++ b/substrate/client/network/src/peer_store.rs
@@ -428,7 +428,7 @@ mod tests {
 	#[test]
 	fn decaying_max_reputation_finally_yields_zero() {
 		const INITIAL_REPUTATION: i32 = i32::MAX;
-		const SECONDS: u64 = 1000;
+		const SECONDS: u64 = 3544;
 
 		let mut peer_info = PeerInfo::default();
 		peer_info.reputation = INITIAL_REPUTATION;
@@ -443,7 +443,7 @@ mod tests {
 	#[test]
 	fn decaying_min_reputation_finally_yields_zero() {
 		const INITIAL_REPUTATION: i32 = i32::MIN;
-		const SECONDS: u64 = 1000;
+		const SECONDS: u64 = 3544;
 
 		let mut peer_info = PeerInfo::default();
 		peer_info.reputation = INITIAL_REPUTATION;
--- a/substrate/client/network/src/peer_store.rs
+++ b/substrate/client/network/src/peer_store.rs
@@ -43,11 +43,18 @@ pub const BANNED_THRESHOLD: i32 = 82 * (i32::MIN / 100);
 /// Reputation change for a node when we get disconnected from it.
 const DISCONNECT_REPUTATION_CHANGE: i32 = -256;
 /// Relative decrement of a reputation value that is applied every second. I.e., for inverse
-/// decrement of 50 we decrease absolute value of the reputation by 1/50. This corresponds to a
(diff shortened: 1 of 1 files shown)


substrate/client/network/src/peer_store.rs
```diff
@@ -39,15 +39,22 @@ use crate::protocol_controller::ProtocolHandle;
 pub const LOG_TARGET: &str = "peerset";
 
 /// We don't accept nodes whose reputation is under this value.
-pub const BANNED_THRESHOLD: i32 = 82 * (i32::MIN / 100);
+pub const BANNED_THRESHOLD: i32 = 71 * (i32::MIN / 100);
```
dmitry-markin, line 42: Worth double-checking we don't have this constant hardcoded anywhere in reputation costs.

dmitry-markin (approved):

substrate/client/network/src/peer_store.rs
```diff
@@ -39,15 +39,22 @@ use crate::protocol_controller::ProtocolHandle;
...
+/// This corresponds to a factor of `k = 0.955`, where k = 1 - 1 / INVERSE_DECREMENT.
+///
+/// It takes ~ `ln(0.5) / ln(k)` seconds to reduce the reputation by half, or 138.63 seconds for the
+/// values above.
+///
+/// In this setup:
+/// - `i32::MAX` becomes 0 in exactly 3544 seconds, or approximately 59 minutes
+/// - `i32::MIN` becomes 0 in exactly 3544 seconds, or approximately 59 minutes
+/// - `i32::MIN` escapes the banned threshold in 69 seconds
+const INVERSE_DECREMENT: i32 = 200;
```
lexnv, line 57: We could also include a more complex mechanism for this (where the numbers are also configurable constants):

- if a peer is reported as banned at least 3 times within 5 minutes
- then we wait 10 minutes until the reputation is decayed (basically banning the peer for around 10 mins + 69 seconds)

It may be overkill for now, but something to think about if we still have misbehaving peers (like shown in kusama). Would love to get your thoughts on this @dmitry-markin 🙏

dmitry-markin: The goal is to conserve resources not handling misbehaving peers' messages/requests. It seems already fine with ~1 min banning, but we yet to observe the real DoS attack to make informed decisions regarding the banning mechanism in general.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5828418

bkchr (approved):

PR #4032, opened 2024-04-08
Adjust zombienet test resources and logic

sandreim:
One more try to make this test robust from a resource perspective.


sandreim:
Got 2 succesfull runs, but 3rd was really bad:

image

pepoviola:
_Small update_: Our hunch is that the intermittent failures are due to the performance of the vms disks we currently use, we are going to add a new tier to the cluster with more performing disks to run these tests. We will be doing this work tonight at midnight.
Thx!

rcny (approved):

.gitlab/pipeline/zombienet/polkadot.yml
```diff
@@ -161,6 +161,8 @@ zombienet-polkadot-functional-0011-async-backing-6-seconds-rate:
 zombienet-polkadot-elastic-scaling-0001-basic-3cores-6s-blocks:
   extends:
     - .zombienet-polkadot-common
+  variables:
+    FORCED_INFRA_INSTANCE: "spot-iops"
```
sandreim, line 165: We should use better naming, something like `high-iops-instance` does more justice

sandreim: cc @pepoviola

pepoviola (approved):

pepoviola (approved):

pepoviola:
> Tried many times and it looks stable now. I'd say we should merge it.

I tried a couple of time more and works as expected :)

alindima (approved):

Issue #3219, opened 2024-02-06
prospective-parachains: limit count of nodes in a fragment tree


alindima:
As discovered in https://github.com/paritytech/polkadot-sdk/pull/3160#discussion_r1476412151, the fragment tree size is an exponential function of `O(num_forks ^ core_count)`.

This is not a problem for non-elastic-scaling, because core_count will always be 1 and num_forks is a constant equal to the backing group size multiplied by the configured async backing max candidate depth.

For secure elastic scaling, we should add a limit to the number of nodes in a fragment tree, to not allow for an unreasonable amount of forks, which could potentially DOS block authoring. This is needed because num_forks and core_count will be influenced by how many cores a parachain acquires. For example if they acquire 100 cores, the fragment tree size could grow to O(500^100) in size. It's reasonable to enforce that parachains that have forks over a sensible amount will not be able to make use of elastic scaling.

alindima:
will most likely be tackled as part of: https://github.com/paritytech/polkadot-sdk/issues/3541

Issue #3699, opened 2024-03-14
Elastic Scaling: Guide Changes


eskimor:
Document elastic scaling changes:

1. Core handling in general
2. Provisioner changes
3. Runtime dependency tracking
4. Collator protocol changes
5. Prospective Parachains changes: Fork + Cycle support removal + support chains with holes.
6. Backing changes

sandreim:
@alindima didn't we update all of this for the MVP ?

alindima:
> @alindima didn't we update all of this for the MVP ?

nope

PR #4035, opened 2024-04-09
prospective-parachains rework

alindima:
Reworks prospective-parachains so that we allow a number of unconnected candidates (for which we don't know the parent candidate yet). Needed for elastic scaling: https://github.com/paritytech/polkadot-sdk/issues/3541. Without this, candidate B will not be validated and backed until candidate A (its parent) is validated and a backing statement reaches the validator.

Due to the high complexity of the subsystem, I rewrote parts of it so that we don't concern ourselves with candidates which form cycles or which form parachain forks. We now have "Fragment chains" instead of "Fragment trees". This greatly simplifies some of the code and is a compromise we can make. We just need to  make sure that cycle-producing parachains don't brick the relay chain and that fork-producing parachains can still make some progress (on one core at least). The only forks that are allowed are those on the relay chain, obviously.

Unconnected candidates are kept in the `CandidateStorage` and whenever a new candidate is introduced, we try to repopulate the chain with as many candidates as we can.

Also fixes https://github.com/paritytech/polkadot-sdk/issues/3219

Guide changes will be done as part of: https://github.com/paritytech/polkadot-sdk/issues/3699

TODOs:

- [x] see if we can replace the `Cow` over the candidate commitments with an `Arc` over the entire `ProspectiveCandidate`. It's only being overwritten in unit tests. We can work around that.
- [x] finish fragment_chain unit tests
- [x] add more prospective-parachains subsystem tests
- [x] test with zombienet what happens if a parachain is creating cycles (it should not brick the relay chain).
- [x] test with zombienet a parachain that is creating forks. it should keep producing blocks from time to time (one bad collator should not DOS the parachain, even if throughput decreases)
- [x] add some more logs and metrics
- [x] add prdoc and remove the "silent" label


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5835328

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5860831

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5878857

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5897378

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5900336

alindima:
I added https://github.com/paritytech/polkadot-sdk/pull/4148 hoping that github will be able to tell that the fragment_tree folder was renamed to fragment_chain. Doesn't look like it helped unfortunately

sandreim:
> I added #4148 hoping that github will be able to tell that the fragment_tree folder was renamed to fragment_chain. Doesn't look like it helped unfortunately

Yeah, maybe it would have worked if it was renamed to `frament_chain` in #4148

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -407,20 +432,20 @@ async fn preprocess_candidates_pending_availability<Context>(
...
+	let IntroduceSecondedCandidateRequest {
 		candidate_para: para,
 		candidate_receipt: candidate,
 		persisted_validation_data: pvd,
 	} = request;
 
 	// Add the candidate to storage.
-	// Then attempt to add it to all trees.
+	// Then attempt to add it to all chains.
 	let storage = match view.candidate_storage.get_mut(&para) {
```
sandreim, line 449: use `let Some(..)` to avoid match.

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -430,81 +455,121 @@ async fn handle_candidate_introduced<Context>(
...
-			// hashing but this branch indicates something is seriously wrong elsewhere
-			// so it's doubtful that it would affect debugging.
+	let actual_relay_parent = candidate.descriptor.relay_parent;
+	let parent_head_hash = pvd.parent_head.hash();
+	let output_head_hash = Some(candidate.commitments.head_data.hash());
+	let candidate_hash =
+		match storage.add_candidate(candidate.clone(), pvd, CandidateState::Seconded) {
+			Ok(c) => c,
+			Err(CandidateStorageInsertionError::CandidateAlreadyKnown(_)) => {
+				gum::warn!(
```
sandreim, line 471: Maybe change to debug, doesn't sound like an error condition as we return true.

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -430,81 +455,121 @@ async fn handle_candidate_introduced<Context>(
...
+
+				let _ = tx.send(false);
+				return
+			},
+		};
+
+	let mut keep_in_storage = false;
+	for (relay_parent, leaf_data) in view.active_leaves.iter_mut() {
+		if let Some(chain) = leaf_data.fragment_chains.get_mut(&para) {
+			gum::debug!(
```
sandreim, line 500:
```suggestion
			gum::trace!(
```

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -430,81 +455,121 @@ async fn handle_candidate_introduced<Context>(
...
+					"kept unconnected candidate",
+				);
 
-	let mut membership = Vec::new();
-	for (relay_parent, leaf_data) in &mut view.active_leaves {
-		if let Some(tree) = leaf_data.fragment_trees.get_mut(&para) {
-			tree.add_and_populate(candidate_hash, &*storage);
-			if let Some(depths) = tree.candidate(&candidate_hash) {
-				membership.push((*relay_parent, depths));
+				gum::debug!(
```
sandreim, line 539:
```suggestion
				gum::trace!(
```

alindima:
> > I added #4148 hoping that github will be able to tell that the fragment_tree folder was renamed to fragment_chain. Doesn't look like it helped unfortunately
> 
> Yeah, maybe it would have worked if it was renamed to `frament_chain` in #4148

yeah, I think that would've worked. But since this PR will take longer until it's merged, master will be in a weird state with everything being called FragmentTree but the folder name being fragment_chain. We could nevertheless do that if you and other reviewers think it's worth it

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -430,81 +455,121 @@ async fn handle_candidate_introduced<Context>(
...
-		},
-		Err(CandidateStorageInsertionError::PersistedValidationDataMismatch) => {
-			// We can't log the candidate hash without either doing more ~expensive
-			// hashing but this branch indicates something is seriously wrong elsewhere
-			// so it's doubtful that it would affect debugging.
+	let actual_relay_parent = candidate.descriptor.relay_parent;
+	let parent_head_hash = pvd.parent_head.hash();
+	let output_head_hash = Some(candidate.commitments.head_data.hash());
+	let candidate_hash =
+		match storage.add_candidate(candidate.clone(), pvd, CandidateState::Seconded) {
```
sandreim, line 468: It seems a bit counter intuitive to why we need to add the candidate to storage then remove it. It would be better if we do some proper renaming and adding some comments about how we use the candidate storage.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+	/// Whether the candidate exists.
+	pub(crate) fn contains_candidate(&self, candidate: &CandidateHash) -> bool {
+		self.candidates.contains(candidate)
+	}
+
+	/// Try accumulating more candidates onto the chain.
+	///
+	/// Candidates can only be added if they build on the already existing chain.
+	pub(crate) fn repopulate(&mut self, storage: &CandidateStorage) {
```
sandreim, line 599: This public method name seems wrong, I was assuming we remove everthing than add everthing back in. All that it seems to do is extend the chain based on new stuff in storage. 

`extend_from_storage` should do more justice.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+			self.scope
+				.ancestor_by_hash(&last_candidate.relay_parent())
+				.or_else(|| {
+					// if the relay-parent is out of scope _and_ it is in the chain,
+					// it must be a candidate pending availability.
+					self.scope
+						.get_pending_availability(&last_candidate.candidate_hash)
+						.map(|c| c.relay_parent.clone())
+				})
+				.expect("All nodes in chain are either pending availability or within scope; qed")
```
sandreim, line 768: Handling this and logging `error` is better to avoid node crashes, if this invariant is every violated.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+	}
+
+	// Return the earliest relay parent a new candidate can have in order to be added to the chain.
+	// This is the relay parent of the last candidate in the chain.
+	// The value returned may not be valid if we want to add a candidate pending availability, which
+	// may have a relay parent which is out of scope. Special handling is needed in that case.
+	fn earliest_relay_parent(&self) -> RelayChainBlockInfo {
+		if let Some(last_candidate) = self.chain.last() {
+			self.scope
+				.ancestor_by_hash(&last_candidate.relay_parent())
```
sandreim, line 760: Why the relay parent of the last candidate is not good in general ?

sandreim (commented):
First pass, looking good so far.

polkadot/node/subsystem-types/src/messages.rs
```diff
@@ -1156,25 +1148,21 @@ impl ParentHeadData {
...
-	/// The response sender accepts the candidate membership, which is the existing
-	/// membership of the candidate if it was already known.
-	IntroduceCandidate(IntroduceCandidateRequest, oneshot::Sender<FragmentTreeMembership>),
-	/// Inform the Prospective Parachains Subsystem that a previously introduced candidate
-	/// has been seconded. This requires that the candidate was successfully introduced in
-	/// the past.
-	CandidateSeconded(ParaId, CandidateHash),
+	/// The response sender returns false if the candidate was rejected by prospective parachains,
+	/// true otherwise (if it was accepted or already present)
+	IntroduceSecondedCandidate(IntroduceSecondedCandidateRequest, oneshot::Sender<bool>),
```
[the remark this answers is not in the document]
alindima: I can do that

polkadot/node/core/backing/src/lib.rs
```diff
@@ -1247,28 +1209,28 @@ async fn construct_per_relay_parent_state<Context>(
 
 enum SecondingAllowed {
 	No,
-	Yes(Vec<(Hash, Vec<usize>)>),
+	// On which leaves is seconding allowed.
+	Yes(Vec<Hash>),
```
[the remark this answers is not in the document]
alindima: we can't always know the depth, because we may second a candidate for which we don't have the parent.
so we only limit the total number. we can no longer enforce one per depth

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -673,72 +757,25 @@ fn answer_hypothetical_frontier_request(
...
-			let depths = fragment_tree.hypothetical_depths(
-				candidate_hash,
-				hypothetical,
-				candidate_storage,
-				request.backed_in_path_only,
-			);
 
-			if !depths.is_empty() {
-				membership.push((*active_leaf, depths));
+			if fragment_chain.hypothetical_membership(candidate_hash, c.into(), candidate_storage) {
```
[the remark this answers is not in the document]
alindima: > Feels confusing to have 2 definitions for HypotheteticalCandidate

It wasn't added by this PR and I don't think they can't be easily merged into one

> can we have the candidate hash in the crate level definition of HypotheteticalCandidate 

This I can do

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -860,8 +896,6 @@ async fn fetch_upcoming_paras<Context>(
 	for core in cores {
 		match core {
 			CoreState::Occupied(occupied) => {
-				pending_availability.insert(occupied.candidate_hash);
-
 				if let Some(next_up_on_available) = occupied.next_up_on_available {
```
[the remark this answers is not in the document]
alindima: > Would be a good time to use the ClaimQueue.

fair

> I am not sure how next_up_ok_time_out works wrt the ClaimQueue. Could we potentially add a para_id twice here ?

It's a `HashSet`, so we don't get duplicates

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: agreed 👍🏻

polkadot/node/core/prospective-parachains/src/lib.rs
alindima: may be a bit counter-intuitive at first, but I did this to avoid duplicating a lot of code between the `populate` and `repopulate` methods.

I'll rename the repopulate method as you suggested and add more comments here

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: if the candidate is pending availability, we keep its relay parent in `self.pending_availability`.
We need to differentiate between pending candidates and new candidates because pending candidates can have a relay parent which is out of scope (older than allowed_ancestry_len)

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+		self.by_parent_head
+			.entry(entry.parent_head_data_hash())
+			.or_default()
+			.insert(candidate_hash);
+		self.by_output_head
+			.entry(entry.output_head_data_hash())
+			.or_default()
+			.insert(candidate_hash);
+		// sanity-checked already.
+		self.by_candidate_hash.insert(candidate_hash, entry);
```
sandreim, line 187: I'd extract this into a fn called `add_entry`.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+impl CandidateStorage {
+	/// Create a new `CandidateStorage`.
+	pub fn new() -> Self {
+		CandidateStorage {
+			by_parent_head: HashMap::new(),
+			by_output_head: HashMap::new(),
+			by_candidate_hash: HashMap::new(),
+		}
+	}
```
sandreim, line 144: Why not just derive `Default` inatead of this ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+				self.by_parent_head
+					.get(hash)
+					.and_then(|m| m.iter().next())
+					.and_then(|a_candidate| self.by_candidate_hash.get(a_candidate))
+					.map(|e| &e.candidate.persisted_validation_data.parent_head)
+			})
+	}
+
+	/// Returns candidate's relay parent, if present.
+	pub(crate) fn relay_parent_by_candidate_hash(
```
sandreim, line 275:
```suggestion
	pub(crate) fn relay_parent_of_candidate(
```

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+pub(crate) struct PendingAvailability {
+	/// The candidate hash.
+	pub candidate_hash: CandidateHash,
+	/// The block info of the relay parent.
+	pub relay_parent: RelayChainBlockInfo,
+}
+
+/// The scope of a [`FragmentChain`].
+#[derive(Debug, Clone)]
+pub(crate) struct Scope {
```
sandreim, line 351: some docs for struct members helps.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+	pub fn earliest_relay_parent(&self) -> RelayChainBlockInfo {
+		self.ancestors
+			.iter()
+			.next()
+			.map(|(_, v)| v.clone())
+			.unwrap_or_else(|| self.relay_parent.clone())
+	}
+
+	/// Get the relay ancestor of the fragment chain by hash.
+	pub fn ancestor_by_hash(&self, hash: &Hash) -> Option<RelayChainBlockInfo> {
```
sandreim, line 436:
```suggestion
	pub fn ancestor(&self, hash: &Hash) -> Option<RelayChainBlockInfo> {
```

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+				relay_parent: *candidate_relay_parent,
+				parent_head_data_hash: *parent_head_data_hash,
+			},
+		}
+	}
+}
+
+pub struct FragmentNode {
+	fragment: Fragment<'static>,
+	pub candidate_hash: CandidateHash,
```
sandreim, line 523: Let's make it consistent, remove `pub` and provide accessor fn.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+}
+
+/// This is a chain of candidates based on some underlying storage of candidates and a scope.
+///
+/// All nodes in the chain must be either pending availability or within the scope. Within the scope
+/// means it's built off of the relay-parent or an ancestor.
+pub(crate) struct FragmentChain {
+	scope: Scope,
+
+	pub chain: Vec<FragmentNode>,
```
sandreim, line 540: Here as well, let's have an accessor for the chain.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+	// Index from head data hash to candidate hashes with that head data as a parent.
+	by_parent_head: HashMap<Hash, CandidateHash>,
+	// Index from head data hash to candidate hashes outputting that head data.
+	by_output_head: HashMap<Hash, CandidateHash>,
+}
+
+impl FragmentChain {
+	/// Create a new [`FragmentChain`] with given scope and populated from the storage.
+	pub fn populate(scope: Scope, storage: &CandidateStorage) -> Self {
```
sandreim, line 552: Not sure this is a good name for constructor. Why not just `new` ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+/// A hypothetical candidate, which may or may not exist in
+/// the fragment chain already.
+pub(crate) enum HypotheticalCandidate<'a> {
+	Complete {
+		receipt: Cow<'a, CommittedCandidateReceipt>,
+		persisted_validation_data: Cow<'a, PersistedValidationData>,
+	},
+	Incomplete {
+		relay_parent: Hash,
```
sandreim, line 466: Don't we know at least the candidate hash ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+	/// - the candidate is already present
+	/// - the candidate can be added to the chain
+	/// - the candidate could potentially be added to the chain in the future (its ancestors are
+	///   still unknown but it doesn't violate other rules).
+	///
+	/// If this returns false, the candidate could never be added to the current chain (not now, not
+	/// ever)
+	pub(crate) fn hypothetical_membership(
+		&self,
+		candidate_hash: CandidateHash,
```
sandreim, line 616: It feels weird to require the candidate hash passed separately, it should be known in `HypotheticalCandidate`

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+	/// Select `count` candidates after the given `ancestors` which pass
+	/// the predicate and have not already been backed on chain.
+	///
+	/// The intention of the `ancestors` is to allow queries on the basis of
+	/// one or more candidates which were previously pending availability becoming
+	/// available or candidates timing out.
+	pub(crate) fn find_backable_chain(
+		&self,
+		ancestors: Ancestors,
+		count: u32,
```
sandreim, line 710: Size of the chain we want 
```suggestion
		size: u32,
```

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+				})
+				.expect("All nodes in chain are either pending availability or within scope; qed")
+		} else {
+			self.scope.earliest_relay_parent()
+		}
+	}
+
+	// Checks if this candidate could be added in the future to this chain.
+	// This assumes that the chain does not already contain this candidate.
+	pub(crate) fn can_add_candidate_as_potential(
```
sandreim, line 776:
```suggestion
	pub(crate) fn is_potential_candidate(
```

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+		// If we've got enough candidates for the configured depth, no point in adding more.
+		if self.chain.len() > self.scope.max_depth {
+			return false
+		}
+
+		if !self.check_potential(relay_parent, parent_head_hash, output_head_hash) {
+			return false
+		}
+
+		// Todo: if max depth is 0, a potential candidate is only a connected candidate.
```
sandreim, line 792: Something we still need to address ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -0,0 +1,1047 @@
...
+	// The candidates which are present in `CandidateStorage`, are not part of this chain but could
+	// become part of this chain in the future. Capped at the max depth minus the existing chain
+	// length.
+	pub(crate) fn find_unconnected_potential_candidates(
+		&self,
+		storage: &CandidateStorage,
+	) -> Vec<CandidateHash> {
+		let mut candidates = vec![];
+		for candidate in storage.candidates() {
+			// We stop at max_depth + 1 with the search. There's no point in looping further.
```
sandreim, line 816: I'd rewrite this with `take_while` -> `filter_map` -> `collect` the candidates.

polkadot/node/core/prospective-parachains/src/lib.rs
alindima: these are still only draft logs. I have a bulletpoint in the PR description to add/fix logs

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: I find `populate` more expressive in this case. Also, this name wasn't introduced here

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: or the candidate count :P

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: even a potential candidate may not be always kept, depending on the number of potential candidates we already have. that's why I went for this name

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: this one is already called `add_candidate`. I don't think it helps

polkadot/node/core/prospective-parachains/src/lib.rs
alindima: done

polkadot/node/subsystem-types/src/messages.rs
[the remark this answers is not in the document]
alindima: thinking a bit more about this, I think it adds some complexity without much benefit. we're already logging the failure reasons in `handle_introduce_seconded_candidate`

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
sandreim: Yeah, but wrt fn name size sounds proper

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: yes, working on it

alindima: done

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
sandreim: Depends on what is the definition of potential. At least for me, it includes having more space for unconnected candidates.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
sandreim: Yeah, you are right, I missed the `candidates.len()` check.

polkadot/node/core/prospective-parachains/src/fragment_chain/tests.rs
```diff
@@ -1415,37 +1709,49 @@ fn pending_availability_in_scope() {
...
+
+	assert!(chain.hypothetical_membership(
+		HypotheticalCandidate::Incomplete {
+			parent_head_data_hash: HeadData::from(vec![0x0c]).hash(),
+			candidate_relay_parent: relay_parent_b,
+			candidate_para: para_id,
+			candidate_hash: candidate_d_hash,
+		},
+		&storage,
+	));
```
alexggh, line 1756: Not sure if it is covered by other tests, can we also add a test that checks work ok with forks on the relay chain and parachain cycles gets rejected.

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -323,13 +334,27 @@ fn prune_view_candidate_storage(view: &mut View, metrics: &Metrics) {
...
+						"Keeping unconnected candidates in storage: {:?}",
+						unconnected_potential
+					);
+				}
+				live_candidates.extend(unconnected_potential);
+			}
+		}
 	}
 
 	view.candidate_storage.retain(|para_id, storage| {
```
alexggh, line 360: The garbage collection, is really non-obvious here, could we add some debug logs  or metrics with the size of this objects. Thank you!

alexggh (commented):
Great Job on simplifying this log, it definitely is easier to digest now with the simplifying assumptions. Also, really good unittest coverage.

To make sure we don't miss anything, on top of of zombienet testing I would also recommend baking this PR for a few days in versi with gluttons and ticks and make sure the rate of parachains being accepted on the relay chain doesn't get somehow negatively impacted.

tdimitrov:
Great work, @alindima. This PR is quite an effort!

The changes look good based on my limited knowledge about the prospective parachains subsystem. I left one nit regarding the tests.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +658,349 @@ impl FragmentTree {
...
+			let mut added_child = false;
+			for candidate in possible_children {
+				// Add one node to chain if
+				// 1. it does not introduce a fork or a cycle.
+				// 2. parent hash is correct.
+				// 3. relay-parent does not move backwards.
+				// 4. all non-pending-availability candidates have relay-parent in scope.
+				// 5. candidate outputs fulfill constraints
+
+				if !self.is_fork_or_cycle(
```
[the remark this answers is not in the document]
alindima: you're right

alindima:
> To make sure we don't miss anything, on top of of zombienet testing I would also recommend baking this PR for a few days in versi with gluttons and ticks and make sure the rate of parachains being accepted on the relay chain doesn't get somehow negatively impacted.

Good idea, this is being done 👍🏻

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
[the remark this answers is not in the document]
alindima: done, flipped the return values (it now returns true if it's a fork or cycle)

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: I added some clarifications in comments about what a potential candidate means

polkadot/node/core/prospective-parachains/src/lib.rs
alindima: done

alindima:
no issues found when testing on versi. it's ready for some approvals @sandreim @tdimitrov @alexggh 😁

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -292,25 +294,38 @@ impl CandidateStorage {
...
-#[derive(Debug)]
-struct CandidateEntry {
+#[derive(Debug, Clone)]
+pub(crate) struct CandidateEntry {
 	candidate_hash: CandidateHash,
+	parent_head_data_hash: Hash,
+	output_head_data_hash: Hash,
 	relay_parent: Hash,
-	candidate: ProspectiveCandidate<'static>,
+	candidate: Arc<ProspectiveCandidate>,
```
sandreim, line 311: better :)

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -430,353 +452,207 @@ impl Scope {
...
-				.find(|(_, n)| &n.candidate_hash == candidate_hash)
-				.map(|(i, _)| NodePointer::Storage(i)),
-			NodePointer::Storage(ptr) =>
-				self.nodes.get(ptr).and_then(|n| n.candidate_child(candidate_hash)),
-		}
-	}
-
 	/// Returns an O(n) iterator over the hashes of candidates contained in the
 	/// tree.
 	pub(crate) fn candidates(&self) -> impl Iterator<Item = CandidateHash> + '_ {
```
sandreim, line 527: Does the caller always have to clone the hashes ? If not, and hope for we should return iterator over references.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -430,353 +452,207 @@ impl Scope {
...
-
-			if candidate_storage.get(candidate_hash).map_or(true, |candidate_entry| {
-				!matches!(candidate_entry.state, CandidateState::Backed)
-			}) {
-				return false
-			}
-			parent_pointer = node.parent;
-		}
+	/// Return a vector of the chain's candidate hashes, in-order.
+	pub(crate) fn chain(&self) -> Vec<CandidateHash> {
```
sandreim, line 542: Isn't `as_vec()` better since this is a method of `FragmentChain` and implies it is a chain already.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +665,335 @@ impl FragmentTree {
...
+		let unconnected = self
+			.find_unconnected_potential_candidates(
+				storage,
+				present_in_storage.then_some(candidate_hash),
+			)
+			.len();
+
+		if (self.chain.len() + unconnected) < self.scope.max_depth {
+			PotentialAddition::Anyhow
+		} else if (self.chain.len() + unconnected) == self.scope.max_depth {
```
sandreim, line 757: Why if we are already at max_depth we can add one more ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +665,335 @@ impl FragmentTree {
...
+				if ignore_candidate == &candidate.candidate_hash {
+					continue
+				}
+			}
+			// We stop at max_depth + 1 with the search. There's no point in looping further.
+			if (self.chain.len() + candidates.len()) > self.scope.max_depth {
+				break
 			}
+			if !self.candidates.contains(&candidate.candidate_hash) {
+				if self.check_potential(
```
sandreim, line 786: Can we merge both conditions in single if ?

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +665,335 @@ impl FragmentTree {
...
-						.map(|(index, node)| (NodePointer::Storage(index), node.candidate_hash))
-						.collect::<Vec<_>>();
+	// Check if adding a candidate which transitions `parent_head_hash` to `output_head_hash` would
+	// introduce a fork or a cycle in the parachain.
+	// `output_head_hash` is optional because we sometimes make this check before retrieving the
+	// collation.
+	fn is_fork_or_cycle(&self, parent_head_hash: Hash, output_head_hash: Option<Hash>) -> bool {
+		if self.by_parent_head.contains_key(&parent_head_hash) {
+			// fork. our parent has another child already
+			return true;
```
sandreim, line 806: Might want to run a round of `fmt`
```suggestion
			return true
```

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +665,335 @@ impl FragmentTree {
...
 
-			Err(())
-		} else {
-			Ok(next.copied())
+		let Some(earliest_rp) = self.earliest_relay_parent() else { return false };
+
+		let Some(relay_parent) = self.scope.ancestor(relay_parent) else { return false };
+
+		if relay_parent.number < earliest_rp.number {
+			return false // relay parent moved backwards.
```
sandreim, line 851: Does it make sense to log trace here just in case for debugging ?

polkadot/node/core/prospective-parachains/src/metrics.rs
```diff
@@ -46,6 +78,34 @@ impl metrics::Metrics for Metrics {
...
+				)?,
+				registry,
+			)?,
+			candidate_storage_unconnected_count: prometheus::register(
+				Gauge::new(
+					"polkadot_parachain_prospective_parachains_candidate_storage_unconnected_count",
+					"Number of unconnected candidates in the prospective parachains candidate storage".to_string(),
+				)?,
+				registry,
+			)?,
```
sandreim, line 108: I would define a single metric and have labels name `kind` with values `connected`, `unconnected`.

polkadot/node/core/prospective-parachains/src/metrics.rs
```diff
@@ -46,6 +78,34 @@ impl metrics::Metrics for Metrics {
...
+				prometheus::Histogram::with_opts(prometheus::HistogramOpts::new(
+					"polkadot_parachain_prospective_parachains_hypothetical_membership",
+					"Time spent responding to `GetHypotheticalMembership`",
+				))?,
+				registry,
+			)?,
+			candidate_storage_count: prometheus::register(
+				Gauge::new(
+					"polkadot_parachain_prospective_parachains_candidate_storage_count",
+					"Total size of the prospective parachains candidate storage".to_string(),
```
sandreim, line 98: Description does not specify the unit of measurement. Only implied by metric name.

polkadot/node/network/collator-protocol/src/validator_side/collation.rs
```diff
@@ -121,19 +121,15 @@ impl PendingCollation {
...
-	pub collator_id: CollatorId,
-	/// The relay-parent of the candidate.
-	pub candidate_relay_parent: Hash,
-	/// Hash of the candidate.
-	pub candidate_hash: CandidateHash,
+/// An identifier for a fetched collation that was blocked from being seconded because we don't have
+/// access to the parent's HeadData. Can be retried once the candidate outputting this head data is
+/// seconded.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct BlockedCollationId {
```
sandreim, line 128: Is this really an ID ? Just `BlockedCollation` maybe ?

prdoc/pr_4035.prdoc
```diff
@@ -0,0 +1,21 @@
+title: "Prospective parachains rework"
+
+doc:
+  - audience: Node Dev
+    description: |
+      Reworks prospective-parachains so that we allow a number of unconnected candidates (for which we don't yet know
+      the parent candidate). Needed for elastic scaling. Also simplifies it to not allow parachain forks and cycles.
```
sandreim, line 7: I'd first mention that this fundamentally changes what information the subsystem stores and operates on. From a tree to just a chain and bunch of unconnected candidates.

sandreim (approved):
Awesome work here and testing on Versi @alindima!

Just have some more nits and questions, but anyway good to go.

polkadot/node/network/collator-protocol/src/validator_side/collation.rs
alindima: it's only the key used in the HashMap, hence it's an ID. It maps to a fetched collation

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: that's a good question.

Even prior to this PR, we were accepting max_depth + 1 unincluded candidates at all times.
Not sure why it was designed this way. I assume it's because a max_candidate_depth of 0 is equal to synchronous backing and would still allow for one candidate to be backed (therefore the +1 is needed)

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: yes

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: removed these specific occurrences. however, this doesn't seem a rule that rustfmt forbids as configured in the CI, since it's passing anyway and my local formatter does not change this either

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
sandreim: I see that makes some sense indeed.

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: I don't think it helps, because this function is also called when pruning the candidate storage on a new leaf update.
therefore, it's bound to happen most of the times under normal circumstances

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
sandreim: The question is that now we have async backing enabled do we still want to keep this ?

alindima: > The question is that now we have async backing enabled do we still want to keep this ?

I don't have a good reason for chaning this. Moreover, we seem to have discovered that we need a max_depth + 1 unincluded segment size also on the collator. So maybe it's here for a reason.
Nevertheless, I'd leave it as is unless we find a good reason to change it

prdoc/pr_4035.prdoc
alindima: reworked the prdoc, let me know if it sounds better now

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: renamed to `to_vec()`

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
alindima: replaced this with the `to_vec` method on the FragmentChain

polkadot/node/core/prospective-parachains/src/fragment_chain/mod.rs
```diff
@@ -789,362 +659,334 @@ impl FragmentTree {
...
+				// 2. parent hash is correct.
+				// 3. relay-parent does not move backwards.
+				// 4. all non-pending-availability candidates have relay-parent in scope.
+				// 5. candidate outputs fulfill constraints
+
+				if self.is_fork_or_cycle(
+					candidate.parent_head_data_hash(),
+					Some(candidate.output_head_data_hash()),
+				) {
 					continue
```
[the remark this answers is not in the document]
alindima: well it can happen if there's a parachain fork or cycle on another relay chain fork. the candidate storage holds all candidates across all relay parents. so when populating the chain, we may attempt to introduce a candidate that is present on another relay chain fork (but those candidates will probably not pass these sanity checks). but it may not neccesarily mean something's wrong

polkadot/node/core/prospective-parachains/src/lib.rs
```diff
@@ -407,104 +447,139 @@ async fn preprocess_candidates_pending_availability<Context>(
...
 	view: &mut View,
-	request: IntroduceCandidateRequest,
-	tx: oneshot::Sender<FragmentTreeMembership>,
-) -> JfyiErrorResult<()> {
-	let IntroduceCandidateRequest {
+	request: IntroduceSecondedCandidateRequest,
+	tx: oneshot::Sender<bool>,
+	metrics: &Metrics,
+) {
+	metrics.time_introduce_seconded_candidate();
```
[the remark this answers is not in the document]
alindima: good catch!

tdimitrov (approved):
Excellent work @alindima

kamilsa:
Are fragment chains backwards compatible with fragment chains? Should we do the corresponding updates in KAGOME ASAP?

alindima:
Fragment trees would maintain parachain forks of backable candidates. In the end, the provisioner subsystem on the block author would only choose the first one to back.

Now, fragment chains enable validators to not maintain parachain forks. If the block author gets two backable candidates that have the same parent, it will not even fetch/store the second one.

We make the assumption that parachains shouldn't be creating forks (often or ever) and if they do, they won't utilise the full throughput.

This shouldn't break compatibility with nodes that still hold fragment trees. We don't expect every validator to use this implementation rightaway.

That being said, this PR is more than a switch from fragment-trees to fragment chains. It enables parallel validation of candidates of the same para across different backing groups (needed for elastic scaling).
Whether or not kagome should implement this ASAP depends on your goals, but it shoulnd't break compatibility.

cryptokin92 (commented):
Good update

PR #4036, opened 2024-04-09
Lookahead collator: do not try to build upon an unknown block

s0me0ne-unkn0wn:


cumulus/client/consensus/aura/src/collators/lookahead.rs
```diff
@@ -315,13 +315,14 @@ where
 			let mut parent_header = initial_parent.header;
 			let overseer_handle = &mut params.overseer_handle;
 
-			// We mainly call this to inform users at genesis if there is a mismatch with the
-			// on-chain data.
-			collator.collator_service().check_block_status(parent_hash, &parent_header);
```
skunert, line 320: Could we not directly call `continue` here? Why move it to the loop below?
Parent hash and parent header below will only change if we built a new block outselves, so that should always return true in subsequent calls anyway (or  I am missing something).

skunert (approved):

cumulus/client/consensus/aura/src/collators/lookahead.rs
s0me0ne-unkn0wn: `parent_hash` and `parent_header` get updated at the bottom of the inner loop. On the other hand, they are updated to values we get after a successful collation, which are unlikely to be unknown or bad  :thinking: Need to think a bit more, thanks for pointing out!

alexggh (approved):

cumulus/client/consensus/aura/src/collators/lookahead.rs
skunert: > `parent_hash` and `parent_header` get updated at the bottom of the inner loop. On the other hand, they are updated to values we get after a successful collation, which are unlikely to be unknown or bad 🤔 Need to think a bit more, thanks for pointing out!

Exactly, that is what I meant. We build a block successfully and set parent to it. But since we built it ourselves it will for sure be in-chain. If the first parent itself if messed up we will abort before entering the `0..2` loop.

cumulus/client/consensus/aura/src/collators/lookahead.rs
```diff
@@ -315,13 +315,14 @@ where
...
-
 			// This needs to change to support elastic scaling, but for continuously
 			// scheduled chains this ensures that the backlog will grow steadily.
 			for n_built in 0..2 {
+				// Do not try to build upon an unknown, pruned or bad block
+				if !collator.collator_service().check_block_status(parent_hash, &parent_header) {
+					break;
```
sandreim, line 323: Does this happen often ? I'd still add a trace here just in case.

sandreim (approved):

sandreim:
I think it depends on the exact reason for block not being present. If for some reason we didn't see it yet but it's present on the relay chain we should just fetch it or expect that it is announced by another collator. @skunert does this make any sense ?

skunert:
I did not think enough about this first time I checked here, I am not sure how this scenario should even occur. The parent we are checking in the code was found via `find_potential_parents`, which starts at the included block and checks all child branches to find a parent. So parents that we find should generally be in the local db. @s0me0ne-unkn0wn did you see this problem in the wild or was it added as precaution?

> I think it depends on the exact reason for block not being present. If for some reason we didn't see it yet but it's present on the relay chain we should just fetch it or expect that it is announced by another collator. @skunert does this make any sense ?

So if we see a new pending para block in the relay chain and it is not announced anywhere soon, we will automatically start fetching it with the pov-recovery mechanism. This already happens.

s0me0ne-unkn0wn:
@sandreim, if you could point me out to the interfaces that would help me achieve that, that could be a great addition. Still, I feel like it is better to have it as a follow-up. Testing the soundness of such a solution could be a real hell IIUC :worried: 

@skunert I came across that during the live testing on Kusama (paritytech/devops#3261). It looked like this:
image

cumulus/client/consensus/aura/src/collators/lookahead.rs
s0me0ne-unkn0wn: Not often at all, and `check_block_status()` has very detailed debug-level messages inside.

bkchr:
Generally weird that the block didn't make it in time over the parachain network to the collator. However, this stuff is not predictable. Also given that there are now that many different collator implementations are running together, makes it more complicated to reason about on what happened.

bkchr (approved):

s0me0ne-unkn0wn:
Merged into #3630

PR #4040, opened 2024-04-09
Synchronize templates

rzadp:
- Progresses https://github.com/paritytech/polkadot-sdk/issues/3155

### What's inside

A job, that will take each of the three [templates](https://github.com/paritytech/polkadot-sdk/tree/master/templates), yank them out of the monorepo workspace, and push to individual repositories ([1](https://github.com/paritytech/polkadot-sdk-minimal-template), [2](https://github.com/paritytech/polkadot-sdk-parachain-template), [3](https://github.com/paritytech/polkadot-sdk-solochain-template)).

In case the build/test does not succeed, a PR such as [this one](https://github.com/paritytech-stg/polkadot-sdk-solochain-template/pull/2) gets created instead.

I'm proposing a manual dispatch trigger for now - so we can test and iterate faster - and change it to fully automatic triggered by releases later.

The manual trigger looks like this:

### How it works

The job replaces dependencies [referenced by git](https://github.com/paritytech/polkadot-sdk/blob/d733c77ee2d2e8e2d5205c552a5efb2e5b5242c8/templates/minimal/pallets/template/Cargo.toml#L25) with a reference to released crates using [psvm](https://github.com/paritytech/psvm).

It creates a new workspace for the template, and adapts what's needed from the `polkadot-sdk` workspace.

### See the results

The action has been tried out in staging, and the results can be observed here:

- [minimal stg](https://github.com/paritytech-stg/polkadot-sdk-minimal-template/)
- [parachain stg](https://github.com/paritytech-stg/polkadot-sdk-parachain-template/)
- [solochain stg](https://github.com/paritytech-stg/polkadot-sdk-solochain-template/)

These are based on the `1.9.0` release (using `release-crates-io-v1.9.0` branch).

### DIFF
--- a/.github/workflows/sync-templates.yml
+++ b/.github/workflows/sync-templates.yml
@@ -48,6 +48,8 @@ jobs:
         id: app_token
         uses: actions/create-github-app-token@v1.9.3
         with:
+          owner: "paritytech"
+          repositories: "paritytech/polkadot-sdk-${{ matrix.template }}-template"
           app-id: ${{ secrets.TEMPLATE_APP_ID }}
           private-key: ${{ secrets.TEMPLATE_APP_KEY }}
       - uses: actions/checkout@v3
--- a/.github/workflows/sync-templates.yml
+++ b/.github/workflows/sync-templates.yml
@@ -120,7 +120,7 @@ jobs:
 
       # 3. Verify the build. Push the changes or create a PR.
 
-      # We've run into out-of-disk error when compiling Polkadot in the next step, so we free up some space this way.
+      # We've run into out-of-disk error when compiling in the next step, so we free up some space this way.
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # 1.3.1
         with:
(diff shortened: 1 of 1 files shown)


gupnik (approved):

AndWeHaveAPlan (approved):

.github/workflows/sync-templates.yml
```diff
@@ -0,0 +1,158 @@
...
+          sudo apt update
+          sudo apt install -y \
+            protobuf-compiler
+          rustup target add wasm32-unknown-unknown
+          rustup component add rustfmt clippy rust-src
+      
+      # 2. Yanking the template out of the monorepo workspace.
+
+      - name: Use psvm to replace git references with released creates.
+        run: find . -type f -name 'Cargo.toml' -exec psvm -o -v ${{ github.event.inputs.crate_release_version }} -p {} \;
```
alvicsam, line 75: It would be nice to validate input before this command

rzadp: It is sort of indirectly validated, because it'll fail before on the `checkout` step, if a branch `release-crates-io-v${{ github.event.inputs.crate_release_version }}` does not exist in `polkadot-sdk`.

Do you think we should add direct validation?

alvicsam: Sounds reasonable, I think it should be enough.

alvicsam (approved):

EgorPopelyaev (approved):

PR #4044, opened 2024-04-09
`CheckNonce`: Provide a way to pass an optional default nonce

bkchr:


substrate/frame/system/src/extensions/check_nonce.rs
```diff
@@ -79,13 +99,14 @@ where
 		_info: &DispatchInfoOf<Self::Call>,
 		_len: usize,
 	) -> Result<(), TransactionValidityError> {
-		let mut account = crate::Account::<T>::get(who);
+		let mut account = crate::Account::<T>::try_get(who)
+			.unwrap_or_else(|_| AccountInfo { nonce: DefaultNonce::get(), ..Default::default() });
 		if account.providers.is_zero() && account.sufficients.is_zero() {
 			// Nonce storage not paid for
 			return Err(InvalidTransaction::Payment.into())
 		}
```
georgepisaltu, line 107: If the account doesn't exist and `Account::<T>::try_get(who)` returns `None`, the `DefaultNonce` will be used but the next check of providers and sufficients is basically guaranteed to fail because the default account data for these is `0`.

If the account exists, a nonce also must exist (most likely created [here](https://github.com/paritytech/polkadot-sdk/blob/8b4cfda7589325d1a34f70b3770ab494a9d4052c/substrate/frame/balances/src/lib.rs#L937) through an [ED funding](https://github.com/paritytech/polkadot-sdk/blob/8b4cfda7589325d1a34f70b3770ab494a9d4052c/substrate/frame/support/src/traits/tokens/fungible/regular.rs#L227) or in some other place providing `providers` or `sufficients` refs), so the `DefaultNonce` introduced here has no effect.

It feels like the place to put this default nonce logic would be in the `AccountData` default impl. Am I missing anything? I encountered this issue while trying to write a test for this.

bkchr: That is a really good point.

bkchr: Yes then this solution here is actually not that great.

gupnik: Yeah, seems like https://github.com/paritytech/polkadot-sdk/pull/4034 is better then?

bkchr:
Doesn't work. This time for 100% :D

PR #4049, opened 2024-04-09
Look further in `ClaimQueue` during collation generation

tdimitrov:
Part of https://github.com/paritytech/polkadot-sdk/issues/1797

Implementation for https://github.com/paritytech/polkadot-sdk/issues/1797#issuecomment-1991099502

When generating a collation the first two elements from the `ClaimQueue` are examined:
* depth 0 - this is the `ParaId` scheduled next. As suggested in the linked comment we will build a collation for this slot on startup or if our `ParaId` was not scheduled for two (or more) slots. Providing a collation for depth 0 will probably time out on the first try but the collation will be ready for the retry,
* depth 1 - this is the most relevant slot from collation generation point of view. As discussed providing collation for depth 0 is challenging so we also prepare one more collation for the next slot if our `ParaId` is scheduled there.

To keep track of the work done in previous iterations of the `collation-generation` run loop a  `HashSet` is passed to `handle_new_activations`. It persist all work scheduled in the current iteration and is overwritten on the next iteration. I used this instead of a `bool` indicating if this is a first iteration or not (as suggested in the comment) because I want to cover the case when the collator's `ParaId` has not been scheduled for two consecutive `ClaimQueue` slots.

This PR is still work in progress. 

TODOS:
- [ ] Collect feedback on the initial approach
- [ ] Fix compilation errors in unit tests.
- [ ] Write new unit tests.
- [ ] Write end-to-end test.

### DIFF
--- a/polkadot/node/collation-generation/src/lib.rs
+++ b/polkadot/node/collation-generation/src/lib.rs
@@ -50,10 +50,10 @@ use polkadot_node_subsystem_util::{
 use polkadot_primitives::{
 	collator_signature_payload, CandidateCommitments, CandidateDescriptor, CandidateReceipt,
 	CollatorPair, CoreIndex, CoreState, Hash, Id as ParaId, OccupiedCoreAssumption,
-	PersistedValidationData, ScheduledCore, ValidationCodeHash,
+	PersistedValidationData, ValidationCodeHash,
 };
 use sp_core::crypto::Pair;
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
 mod error;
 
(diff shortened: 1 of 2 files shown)


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5843905

bkchr:
Collation generation is a subsystem that slowly has reached the end of its lifetime. The changes you are trying to incorporate here are not really necessary. None of the new collator logic will make use of the collation generation subsystem for triggering its block production. The new collator logic is being build based on what is described [here](https://github.com/paritytech/polkadot-sdk/issues/3168), which already includes the stuff that is mentioned in the linked issue comment.

PR #4050, opened 2024-04-09
tx: Remove tx_broadcast transaction from the pool

lexnv:
This PR ensures that broadcast future cleans-up the submitted extrinsic from the pool, iff the `broadcast_stop` operation has been called.

This effectively cleans-up transactions from the pool when the `broadcast_stop` is called.

cc @paritytech/subxt-team


substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -111,12 +113,21 @@ where
...
 			let Ok(decoded_extrinsic) = TransactionFor::<Pool>::decode(&mut &bytes[..]) else {
 				return;
 			};
 
+			// Save the tx hash to remove it later.
+			let tx_hash = pool.hash_of(&decoded_extrinsic);
+			broadcast_ids
+				.write()
+				.get_mut(&current_id)
+				.map(|state| state.tx_hash = Some(tx_hash));
```
bkchr, line 129: You can move this out of the future and store the `tx_hash` directly. Then you also don't need to use an `Option` for the the `tx_hash`.

lexnv: Yep, that makes sense! Thanks for the review!

jsdw (approved):

substrate/client/rpc-spec-v2/src/transaction/transaction_broadcast.rs
```diff
@@ -106,17 +108,19 @@ where
 		// The unique ID of this operation.
 		let id = self.generate_unique_id();
 
+		// There is nothing we could do with an extrinsic of invalid format.
```
niklasad1, line 111: Maybe improve this comment to something like that spec states because this logic was not super obvious to me:

> The JSON-RPC server might check whether the transaction is valid before broadcasting it. If it does so and if the transaction is invalid, the server should silently do nothing and the JSON-RPC client is not informed of the problem. Invalid transactions should still count towards the limit to the number of simultaneously broadcasted transactions.

niklasad1 (approved):

PR #4051, opened 2024-04-09
net/strategy: Log bad peerId from on_validated_block_announce

lexnv:
This tiny PR extends the `on_validated_block_announce` log with the bad PeerID.
Used to identify if the peerID is malicious by correlating with other logs (ie peer-set).

While at it, have removed the `\n` from a multiline log, which did not play well with [sub-triage-logs](https://github.com/lexnv/sub-triage-logs/tree/master).

cc @paritytech/networking


dmitry-markin (approved):

substrate/client/network/sync/src/strategy/chain_sync.rs
```diff
@@ -1063,7 +1063,7 @@ where
 		let peer = if let Some(peer) = self.peers.get_mut(&peer_id) {
 			peer
 		} else {
-			error!(target: LOG_TARGET, "💔 Called `on_validated_block_announce` with a bad peer ID");
+			error!(target: LOG_TARGET, "💔 Called `on_validated_block_announce` with a bad peer ID {}", peer_id);
```
bkchr, line 1066:
```suggestion
			error!(target: LOG_TARGET, "💔 Called `on_validated_block_announce` with a bad peer ID {peer_id}");
```

bkchr (approved):

alexggh (approved):

dmitry-markin:
> Used to identify if the peerID is malicious by correlating with other logs (ie peer-set).

It's unlikely the peer id is malicious — it either means a logic error (we would have seen it more often then), or that the peer has already disconnected and was removed from the map before the block announcement was asynchronously validated.

PR #4060, opened 2024-04-09
Fix nostd build of several crates

ggwpez:
Preparation for https://github.com/paritytech/polkadot-sdk/pull/3935

Changes:
- Add some `default-features = false` for the case that a crate and that dependency both support nostd builds.
- Shuffle files around of some benchmarking-only crates. These conditionally disabled the `cfg_attr` for nostd and pulled in libstd. Example [here](https://github.com/ggwpez/zepter/pull/95). The actual logic is moved into a `inner.rs` to preserve nostd capability of the crate in case the benchmarking feature is disabled.
- Add some `use sp_std::vec` where needed.
- Remove some `optional = true` in cases where it was not optional.
- Removed one superfluous `cfg_attr(not(feature = "std"), no_std..`.

All in all this should be logical no-op.


cumulus/parachains/runtimes/assets/asset-hub-rococo/Cargo.toml
```diff
@@ -36,7 +36,7 @@ pallet-nfts = { path = "../../../../../substrate/frame/nfts", default-features =
 pallet-nfts-runtime-api = { path = "../../../../../substrate/frame/nfts/runtime-api", default-features = false }
 pallet-proxy = { path = "../../../../../substrate/frame/proxy", default-features = false }
 pallet-session = { path = "../../../../../substrate/frame/session", default-features = false }
-pallet-state-trie-migration = { path = "../../../../../substrate/frame/state-trie-migration", default-features = false, optional = true }
+pallet-state-trie-migration = { path = "../../../../../substrate/frame/state-trie-migration", default-features = false }
```
ggwpez, line 39: This should still be gated by the construct_runtime! feature gate. But i think that is not actually working, otherwise this change would not be needed 🤷‍♂️

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5851965

prdoc/pr_4060.prdoc
```diff
@@ -0,0 +1,56 @@
+title: "Fix nostd build of several crates"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Fixes feature and dependency configuration of several crate. This should allow for better no-std build capabilities.
+
+crates:
+  - name: cumulus-pallet-session-benchmarking
+    bump: patch
```
ggwpez, line 10: Not sure if `patch` is technically correct, but it is the intention here.

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5852280

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5852438

liamaharon (approved):

substrate/frame/contracts/Cargo.toml
```diff
@@ -44,7 +44,7 @@ frame-benchmarking = { path = "../benchmarking", default-features = false, optio
 frame-support = { path = "../support", default-features = false }
 frame-system = { path = "../system", default-features = false }
 pallet-balances = { path = "../balances", default-features = false, optional = true }
-pallet-contracts-uapi = { path = "uapi" }
+pallet-contracts-uapi = { path = "uapi", default-features = false, features = ["scale"] }
```
athei, line 47: This crate is `no_std` only. It doesn't even have a `std` feature. Do you still want to add the `default-features = false`.

ggwpez: Oh right that crate is nostd-only, thanks!

substrate/frame/try-runtime/src/lib.rs
```diff
@@ -18,36 +18,9 @@
 //! Supporting types for try-runtime, testing and dry-running commands.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg(feature = "try-runtime")]
```
bkchr, line 21: It makes no sense that this crate has a `try-runtime` feature.. This crate should only be activated if you use `try-runtime`.

substrate/frame/system/benchmarking/src/lib.rs
```diff
@@ -15,221 +15,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Benchmarks for Utility Pallet
+//! Frame System benchmarks.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg(feature = "runtime-benchmarks")]
```
bkchr, line 21: Same. This crate should only be activated when you are compiling runtime benchmarks.

substrate/frame/src/lib.rs
```diff
@@ -51,317 +51,11 @@
 //! Please note that this crate can only be imported as `polkadot-sdk-frame` or `frame`.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg(feature = "experimental")]
 
-#[doc(no_inline)]
-pub use frame_support::pallet;
+// NOTE that we put the actual code into another module to keep the no-std capability of this create
+// even when the `experimental` feature is disabled.
```
bkchr, line 56: I don't get this comment. Also generally it doesn't make sense that the `experimental` feature disables everything of this crate. 

What is the purpose of this feature then? Just don't use the crate if you don't want to use experimental stuff.

substrate/frame/try-runtime/src/lib.rs
ggwpez: Then it would enable the `try-runtime` feature of all dependencies by default?  
I dont remember if this can cause issues with feature unification, but maybe it works since the crate would disabled otherwise.

ggwpez: Okay actually this will make `--workspace` not compile anymore, since it enables all crates, you can check it with `cargo tree --workspace` and see that `frame-system-benchmarking` is enabled as top-level @bkchr.

cumulus/parachains/runtimes/assets/asset-hub-rococo/Cargo.toml
ggwpez: It is still deployed though, will remove it then.

substrate/frame/try-runtime/src/lib.rs
bkchr: Hmm yeah okay... :seal: 

Maybe having the `try_runtime` types in `frame-support` not behind a feature would be a better solution.

ggwpez: We have some traits with feature gated functions. Maybe that is an anti-pattern in itself... should probably be an extension trait that is only selectively required.

substrate/frame/root-offences/Cargo.toml
```diff
@@ -24,8 +24,9 @@ pallet-staking = { path = "../staking", default-features = false }
 
 frame-support = { path = "../support", default-features = false }
 frame-system = { path = "../system", default-features = false }
-sp-runtime = { path = "../../primitives/runtime" }
+sp-runtime = { path = "../../primitives/runtime", default-features = false }
 sp-staking = { path = "../../primitives/staking", default-features = false }
+sp-std = { path = "../../primitives/std", default-features = false }
```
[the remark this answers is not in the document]
ggwpez: Using the `alloc` crate now for type `Vec`.

cumulus/parachains/runtimes/assets/asset-hub-rococo/Cargo.toml
ggwpez: Uhf, did we really genesis the asset-hub-rococo with state_version 0?  
I just queried `Core::version` of the 1.1 runtime and it returns state_version=0...

Looks like it was copy&pasted from AHK. @cheme can you double-check please?

cheme: Ouch https://github.com/polkadot-fellows/runtimes/blob/31ba26287ec752574244f0d690167f7ae8430c8b/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L153 . (kusuma is fine).
Statemint on rococo (I guess it is now assethub, not sure) was migrated IIRC or was it only on westend cc @PierreBesson ?

ggwpez: I meant it here for the Asset-hub-rococo: https://github.com/paritytech/polkadot-sdk/blob/5601f2865b3f867131654189c08e7776baf83d16/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L110

Not sure, since when i queried the runtime api it returned version 0. Maybe we can check with the RPC but the public node does not support it.

cheme: maybe https://github.com/polkadot-fellows/runtimes/issues/74

substrate/frame/try-runtime/src/lib.rs
bkchr: I mean the functions is fine, I would probably just not hide the traits etc in `try_runtime.rs`. But in the end this is all discussable and we can keep it like it is done right now.

bkchr: Generally just don't like these `inner.rs` files.

cumulus/parachains/runtimes/assets/asset-hub-rococo/Cargo.toml
```diff
@@ -101,14 +101,6 @@ substrate-wasm-builder = { path = "../../../../../substrate/utils/wasm-builder",
...
 [features]
 default = ["std"]
-# When enabled the `state_version` is set to `1`.
-# This means that the chain will start using the new state format. The migration is lazy, so
-# it requires to write a storage value to use the new state format. To migrate all the other
-# storage values that aren't touched the state migration pallet is added as well.
-# This pallet will migrate the entire state, controlled through some account.
-#
-# This feature should be removed when the main-net will be migrated.
-state-trie-version-1 = []
```
cheme: oh now I remembre, this is used for multiple network so the feature was needed to have one running V0 while another one was already migrated (likely westend/rococo a long time ago).
Sounds ok to remove it so migration got started everywhere. One point I am not sure is related to the account use as controller: currently it is hardcoded to :
```
pub const MigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
	pub const RootMigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
```
probably would need a different one per network, this is not really convenient.
Worst case, what can happen with this chain is some parachain moving to hybrid mode (both v0 and v1 in tree, if warpsync is not use by the parachain this should not be an issue) and not being able to migrate because no account to process the typescript driven migration.

substrate/frame/src/lib.rs
ggwpez: Okay i reverted the changes to the FRAME crate. Apparently it is not published anyway - so should not be needed for the umbrella crate until then.  
Also increased the warning in the docs. I could not rename it since the name is used in some crate-access generation macros and repeated often in docs... can fix in the future once we start publishing.

cumulus/parachains/runtimes/assets/asset-hub-rococo/Cargo.toml
ggwpez: Okay thanks for the context. For Rococo it is probably fine, as it will be decommissioned soon anyway.  
Just going to leave it as is not and hope that DevOps has the keys. Otherwise we do another upgrade or root call etc.

PierreBesson: I confirm we have the key to `0x8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52` to be able to process the migration.

PR #1403, opened 2023-09-05
Fix nothing scheduled on session boundary

eskimor:
We clear out claim queues at the end of the session. This leads to nodes not seeing anything scheduled for the next block. This is fixed by explicitly updating the claim queues in the `availability_cores` runtime API.

This ended up including a few further fixes and cleanups. Code is still not clean, but cleaner.

Issue #4002, opened 2024-04-05
Cleanup the runtime


eskimor:
Runtime is spaghetti code lacking in proper abstractions.

Examples: 

- Code in inclusion implicitly assumes that some filtering function has been called in paras_inherent.
- Any [new filtering](https://github.com/paritytech/polkadot-sdk/pull/3231/files#diff-43b438747de567b45960033dff7827f97b0639f0aa4e8953ab715891b73ea972R609) needs to remember to check for calling context - this should be top-level.
- Weights via benchmarking of the enter function does not work at all: https://github.com/paritytech/polkadot-sdk/issues/849

In general we are lacking proper structure.

PR #4065, opened 2024-04-10
Fix ClaimQueue case of nothing scheduled on session boundary 

sandreim:
Same issue but about av-cores was fixed in https://github.com/paritytech/polkadot-sdk/pull/1403


sandreim:
We better do https://github.com/paritytech/polkadot-sdk/issues/4002 as the code rot is increasing.

alindima (approved):

eskimor (approved):

alexggh (approved):

PR #4068, opened 2024-04-10
Run subsystem-benchmark without network latency

AndreiEres:
Implements the idea from https://github.com/paritytech/polkadot-sdk/pull/3899
- Removed latencies
- Number of runs reduced from 50 to 5, according to local runs it's quite enough
- Network message always sends in a spawned task even if latency is zero. Without it, CPU time sometimes spikes.

### DIFF
--- a/polkadot/node/subsystem-bench/src/lib/network.rs
+++ b/polkadot/node/subsystem-bench/src/lib/network.rs
@@ -474,19 +474,17 @@ impl EmulatedPeer {
 	pub async fn send_message(&mut self, message: NetworkMessage) {
 		self.tx_limiter.reap(message.size()).await;
 
-		if self.latency_ms == 0 {
-			self.to_node.unbounded_send(message).expect("Sending to the node never fails");
-		} else {
-			let to_node = self.to_node.clone();
-			let latency_ms = std::time::Duration::from_millis(self.latency_ms as u64);
+		let to_node = self.to_node.clone();
+		let latency_ms = std::time::Duration::from_millis(self.latency_ms as u64);
 
-			// Emulate RTT latency
-			self.spawn_handle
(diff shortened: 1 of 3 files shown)


alvicsam (approved):

polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -474,19 +474,17 @@ impl EmulatedPeer {
...
-			let latency_ms = std::time::Duration::from_millis(self.latency_ms as u64);
-
-			// Emulate RTT latency
-			self.spawn_handle
-				.spawn("peer-latency-emulator", "test-environment", async move {
+		let to_node = self.to_node.clone();
+		let latency_ms = std::time::Duration::from_millis(self.latency_ms as u64);
+
+		self.spawn_handle
+			.spawn("peer-latency-emulator", "test-environment", async move {
```
alexggh, line 481: I'm not sure I understand how this should help, we would actually get scheduling noise even if latency is zero, so this change would work against making the test more deterministic by setting latency to None.

sandreim: I have the same question. How much of an improvement are we seeing and do we have any clue why that is ?

AndreiEres: You are right, I used to receive weird results with sending without spawning, but now checked again and everything is ok. I suppose I just had some "noise" on my laptop because of rust-analyzer.

sandreim: I would use a VM with ref hw specs to run the tests and establish gold values for the asserts.

AndreiEres: Yep, I reverted that

alexggh (approved):

PR #4075, opened 2024-04-11
Adds ability to trigger tasks via unsigned transactions

gupnik:
This PR updates the `validate_unsigned` hook for `frame_system` to allow valid tasks to be submitted as unsigned transactions. It also updates the task example to be able to submit such transactions via an off-chain worker.


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5876611

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5880528

ggwpez (approved):

substrate/frame/system/src/lib.rs
```diff
@@ -742,8 +742,6 @@ pub mod pallet {
 		#[pallet::call_index(8)]
 		#[pallet::weight(task.weight())]
 		pub fn do_task(origin: OriginFor<T>, task: T::RuntimeTask) -> DispatchResultWithPostInfo {
-			ensure_signed(origin)?;
-
 			if !task.is_valid() {
 				return Err(Error::<T>::InvalidTask.into())
```
ggwpez, line 746: Is this defensive? I think the `validate_unsigned` should have prevented this?

substrate/frame/system/src/lib.rs
```diff
@@ -1032,6 +1030,12 @@ pub mod pallet {
 					})
 				}
 			}
+			#[cfg(feature = "experimental")]
+			if let Call::do_task { ref task } = call {
+				if task.is_valid() {
```
ggwpez, line 1035: Can you maybe mention in the doc of this function that it **MUST** be cheap, or there is a Dos vector IIUC?  
Like "no storage reads" cheap would be best, but probably impossible...

substrate/frame/system/src/lib.rs
gupnik: Yeah, `validate_unsigned` only allows those to be passed that meet the validity criteria. Anyone can still send the signed ones though, in which case they pay for execution.

substrate/frame/system/src/lib.rs
gupnik: You mean that `is_valid` check should be cheap, correct?

kianenigma: Yes.

gupnik: Done

substrate/frame/system/src/lib.rs
ggwpez: I see. In theory it would only need to check for validity in the signed case?  
That would make it slightly cheaper for unsigned ones. But okay, thanks for the explanation.

substrate/frame/support/src/traits/tasks.rs
```diff
@@ -46,6 +46,9 @@ pub trait Task: Sized + FullCodec + TypeInfo + Clone + Debug + PartialEq + Eq {
 	fn iter() -> Self::Enumeration;
 
 	/// Checks if a particular instance of this `Task` variant is a valid piece of work.
```
bkchr, line 48:
```suggestion
	/// Checks if a particular instance of this `Task` variant is a valid piece of work.
	///
```

substrate/frame/system/src/lib.rs
```diff
@@ -1032,6 +1030,12 @@ pub mod pallet {
...
+					return Ok(ValidTransaction::default())
```
bkchr, line 1036:
```suggestion
					return Ok(ValidTransaction { priority: u64::max_value(), requires: Vec::new(), provides: vec![T::Hash::hash_of(&task)], longevity: TransactionLongevity::max_value(), propagate: false })
```

No need to propagate this transaction because this will created on every validator by the offchain worker. Or do you assume that someone sends this `do_task` manually?

substrate/frame/examples/tasks/src/lib.rs
```diff
@@ -59,9 +63,33 @@ pub mod pallet {
...
+				let task = Task::<T>::AddNumberIntoTotal { i: key };
+				let runtime_task = <T as Config>::RuntimeTask::from(task);
+				let call = frame_system::Call::<T>::do_task { task: runtime_task.into() };
+
+				// Submit the task as an unsigned transaction
+				SubmitTransaction::<T, frame_system::Call<T>>::submit_unsigned_transaction(
+					call.into(),
+				)
+				.map_err(|()| "Unable to submit unsigned transaction.")
+				.unwrap();
```
bkchr, line 81: We should teach people to not use unwrap :D

bkchr: Maybe just log it.

bkchr (approved):

substrate/frame/system/src/lib.rs
gupnik: > Or do you assume that someone sends this do_task manually?

Yeah, they have an option to do as well, especially when the validator chooses to not run the off-chain worker?

bkchr: Not sure any one deactivates it :D But yeah, fine. Then only change `propagate` to `true`.

bkchr: (But the other stuff I added there is important) (maybe we should make the `longevity` smaller, otherwise the stay around in the state quite long.)

bkchr: Generally we should document that tasks can be executed in random order :D

gupnik: Yeah, sounds good. Thanks!

substrate/frame/examples/tasks/src/lib.rs
Ank4n, line 81:
```suggestion
				.defensive();
```

Ank4n: Defensive adds a log in prod and panics in tests :)

gupnik: Oh, I didn't know. Thanks :))

PR #4079, opened 2024-04-11
[FRAME] Storage derive macro `#[frame_support::stored]`

ggwpez:
WIP DNM

### DIFF
--- a/substrate/frame/conviction-voting/src/types.rs
+++ b/substrate/frame/conviction-voting/src/types.rs
@@ -32,18 +32,8 @@ use super::*;
 use crate::{AccountVote, Conviction, Vote};
 
 /// Info regarding an ongoing referendum.
-#[derive(
-	CloneNoBound,
-	PartialEqNoBound,
-	EqNoBound,
-	RuntimeDebugNoBound,
-	TypeInfo,
-	Encode,
-	Decode,
-	MaxEncodedLen,
-)]
-#[scale_info(skip_type_params(Total))]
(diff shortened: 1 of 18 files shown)


substrate/frame/conviction-voting/src/types.rs
```diff
@@ -32,18 +32,8 @@ use super::*;
...
-	TypeInfo,
-	Encode,
-	Decode,
-	MaxEncodedLen,
-)]
-#[scale_info(skip_type_params(Total))]
-#[codec(mel_bound(Votes: MaxEncodedLen))]
+#[derive(CloneNoBound, PartialEqNoBound, EqNoBound, RuntimeDebugNoBound)]
+#[frame_support::stored(skip(Total), mel(Votes))]
 pub struct Tally<Votes: Clone + PartialEq + Eq + Debug + TypeInfo + Codec, Total> {
```
ggwpez, line 37: The difficult part about this was that it tries to re-add the `Clone` bound for all parameters that are not listed here.  
And that requires some changes to all of our `*NoBound` derives. Now i am thinking that we can just scrap that and require the developer to write `Votes: Clone` themselves like we do here.

muharem: if I read it right, `mel(Votes)` here means do not generate anything for `Votes` and use whatever is set for it below? wonder why `MaxEncodedLen` is not needed then?

ggwpez:
Superseded by https://github.com/paritytech/polkadot-sdk/pull/10195

Issue #4062, opened 2024-04-10
Pallet-MMR should use custom `block_hash` getter instead of using `frame_system::block_hash` for non-canonicalised blocks


vedhavyas:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Motivation

Currently `pallet_mmr` uses `frame_system::block_hash` to get the hash of a given block number while constructing the non-canonicalised keys. This works fine with Polkadot and other blockchains where finality is quicker. Currently, polkadot stores [4096](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/common/src/lib.rs#L88) number of block hashes on the runtime and before the block_hashes are pruned, the blocks are already finalized.

At subspace, we do not have a quick finality and we cannot put a static number at which we can gaurantee block finalization happens. With this context, when an MMR proof is being verified, if the particular block is not canonicalised yet, it will look at `frame_system::block_hashes` but by then our runtime already prunes the block hash due to `frame_system::Config::BlockHashCount` and the storage returns an Default value for the Hash since the `frame_system::block_hash` is a ValueQuery. With the default value given, the MMR proof verification fails on our end.

### Request

`pallet_mmr` should ideally define a custom getter to fetch the block_hash instead of directly using the `frame_system`. This will provide the projects using the `pallet_mmr` can impl the custom getter to fetch the said `block_hash`.

For polkadot and likes, the impl will just wrap `frame_system::block_hashes` function and should not change anything while Subspace and other project can impl custom logic to fetch the correct `block_hash` for a given non-canonicalised block number.

### Solution

Solution would be introduce a new type for `pallet_mmr::Config` that can fetch `block_hash` for a given block_number

### Are you willing to help with this request?

Yes!

acatangiu:
IIUC this https://github.com/paritytech/polkadot-sdk/blob/817870e3b25cebb3dd9c0934bcafcafb2532d96a/substrate/frame/merkle-mountain-range/src/mmr/storage.rs#L90 is what is causing problems for you.

So instead of `<frame_system::Pallet<T>>::block_hash` (mapping `block-number -> block-hash` on current fork), I am guessing you will use some other **on-chain** map `block_number -> some-unique-identifier` that runs deeper than 4096, right?

> Solution would be introduce a new type for pallet_mmr::Config that can fetch block_hash for a given block_number

[...]

PR #4080, opened 2024-04-11
define block hash provider and default impl using frame_system

vedhavyas:
This PR introduces `BlockHashProvider` into `pallet_mmr::Config`
This type is used to get `block_hash` for a given `block_number` rather than directly using `frame_system::Pallet::block_hash`

The `DefaultBlockHashProvider` uses `frame_system::Pallet::block_hash` to get the `block_hash`

Closes: #4062

### DIFF
--- a/polkadot/runtime/rococo/src/lib.rs
+++ b/polkadot/runtime/rococo/src/lib.rs
@@ -1190,6 +1190,7 @@ impl pallet_mmr::Config for Runtime {
 	type OnNewRoot = pallet_beefy_mmr::DepositBeefyDigest<Runtime>;
 	type WeightInfo = ();
 	type LeafData = pallet_beefy_mmr::Pallet<Runtime>;
+	type BlockHashProvider = pallet_mmr::DefaultBlockHashProvider<Runtime>;
 }
 
 parameter_types! {
--- a/polkadot/runtime/westend/src/lib.rs
+++ b/polkadot/runtime/westend/src/lib.rs
@@ -333,6 +333,7 @@ impl pallet_mmr::Config for Runtime {
 	type OnNewRoot = pallet_beefy_mmr::DepositBeefyDigest<Runtime>;
 	type WeightInfo = ();
 	type LeafData = pallet_beefy_mmr::Pallet<Runtime>;
+	type BlockHashProvider = pallet_mmr::DefaultBlockHashProvider<Runtime>;
 }
 
 /// MMR helper types.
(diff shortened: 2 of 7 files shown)


acatangiu (approved):

paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5881092

serban300 (commented):
The PR makes sense for MMR. I was also wondering if in order to make it more generic it would make sense to change `frame_system::Config::BlockHashCount` to a method or something more generic. Not sure. But we could explore this in a separate issue. And related to this I was curious how your `BlockHashProvider` implementation looks like. Are you keeping a separate item in the storage with more `block_num -> block_hash ` items then the `frame_system` ? And how do you decide when to prune that ?

vedhavyas:
> I was also wondering if in order to make it more generic it would make sense to change frame_system::Config::BlockHashCount to a method or something more generic.

I'm not quiet sure how this would work for chains that do not have quick finality. Subspace for example, Runtime would not know at which point finality comes so our `BlockHashCount` method should ideally be able to Host function or something else for this. This seems to introduce more tricky situations on the first thought but we will need to think more on this.

> But we could explore this in a separate issue.
That is fair. Thanks!

> And related to this I was curious how your BlockHashProvider implementation looks like. Are you keeping a separate item in the storage with more block_num -> block_hash  items then the frame_system ? And how do you decide when to prune that ?

At the moment, we have not decided on the approach of how to fetch yet but we have some options.
- Our domains, enshrined rollups, would need to know consensus hash so we have storage for these hashes and we are planning to use these but there are potential concerns that came up with this usage
- Other option is to use a Host function to fetch the canonical hash using the block_number since, I believe, we will have access to those until the finality occurs.

serban300 (approved):

vedhavyas:
@serban300 seems like check is failing, IIUC, is unrelated to this PR. Is this expected ?

serban300:
> @serban300 seems like check is failing, IIUC, is unrelated to this PR. Is this expected ?

Yes, the failing checks are unrelated to the PR so far

vedhavyas:
@acatangiu @serban300 I'm not why the the `required` test failed. Seems like the change is not related to it. I would really appreciate your help on this. Thank you!

serban300:
`test-linux-oldkernel-stable 1/3` has been having issues lately. I'll retry it

PR #4081, opened 2024-04-11
[pallet-broker] Force-unpool provisionally pooled regions before redispatching them

seadanda:
This PR force removes regions from the pool before allowing them to be redispatched (through `partition`/`interlace`/`assign`) in the case that a region was pooled with `Provisional` finality.

This PR does not account for the case where a pooled region already entitles the benefactor to a contribution reward before the point of it being redispatched. However, `claim_revenue` should be called before trying to redispatch a region anyway.
Otherwise there would be ambiguity as to what should be done with the first part of a partitioned region. The first of the two new regions actually has the exact same `region_id` as the original region -- leave it pooled/withdraw from pool/claim contribution then withdraw?

### DIFF
--- a/prdoc/pr_4081.prdoc
+++ b/prdoc/pr_4081.prdoc
@@ -13,4 +13,4 @@ doc:
 
 crates:
 - name: pallet-broker
-  bump: patch
+  bump: major


substrate/frame/broker/src/utility_impls.rs
```diff
@@ -118,4 +118,33 @@ impl<T: Config> Pallet<T> {
...
+	// read. No-op for regions which have not been pooled.
+	pub(crate) fn force_unpool_region(region_id: RegionId, region: &RegionRecordOf<T>) {
+		// We don't care if this fails or not, just that it is removed if present. This is to
+		// account for the case where a region is pooled provisionally and redispatched.
+		if let Some(_) = InstaPoolContribution::<T>::take(region_id) {
+			let current_timeslice = Self::current_timeslice();
+			// Do no more for regions that have ended.
+			if region.end < current_timeslice {
+				return
+			};
```
bkchr, line 135:
```suggestion
			}
```

substrate/frame/broker/src/utility_impls.rs
```diff
@@ -118,4 +118,33 @@ impl<T: Config> Pallet<T> {
...
+	// Remove a region from on-demand pool contributions. Useful in cases where it was pooled
+	// provisionally and it is being redispatched (partition/interlace/assign).
+	//
+	// Takes both the region_id and (a reference to) the region as arguments to avoid another DB
+	// read. No-op for regions which have not been pooled.
+	pub(crate) fn force_unpool_region(region_id: RegionId, region: &RegionRecordOf<T>) {
+		// We don't care if this fails or not, just that it is removed if present. This is to
+		// account for the case where a region is pooled provisionally and redispatched.
+		if let Some(_) = InstaPoolContribution::<T>::take(region_id) {
+			let current_timeslice = Self::current_timeslice();
```
bkchr, line 131: Using here the current timeslice is incorrect? This timeslice was already scheduled and trying to change this will not change anything? We should use `last_committed_timeslice + 1` as the timeslice to process.

seadanda: It took me a while to decide which timeslice to use, I thought this would be the best one as it's the next one to be committed, but actually it might already have been committed. But what you're saying makes sense as the tick might have already run for this timeslice and processed the contributions.

bkchr: We are also committing `advance_notice` blocks before, thus, `current_slice()` will still return the old timeslice for `advance_notice` blocks.

seadanda: OK it actually needs to be `current_timeslice() + 2` 😳
It makes sense when you think about it though:
The tick that happens in timeslice N processes the history for [timeslice N+1](https://github.com/paritytech/polkadot-sdk/blob/donal-broker-force-unpool/substrate/frame/broker/src/tick_impls.rs#L55) and so when we want to force unpool in timeslice N, we need to register the change in `InstaPoolIo` in timeslice N+2 or the tick never sees the change and the `InstaPoolHistory` keeps them forever.

We could also use `last_committed_timeslice + 1` but is it worth the two db reads (config and status) for the sake of sometimes working 1 timeslice sooner when they beat the tick?

substrate/frame/broker/src/lib.rs
```diff
@@ -390,6 +390,14 @@ pub mod pallet {
 			/// The Region whose contribution is no longer exists.
 			region_id: RegionId,
 		},
+		/// A region has been force-removed from the pool. This is usually due to a provisionally
+		/// pooled region being redeployed.
+		RegionUnpooled {
```
seadanda, line 395: This is the reason for the major bump, but since there is an event for a region being pooled we should also have one for it being unpooled.

bkchr (approved):

eskimor (commented):
Tests seem to fail?

seadanda:
Hmm yea I just updated from master today after a while of waiting for audit, and it seems that https://github.com/paritytech/polkadot-sdk/pull/3940 changed the mock for these tests

substrate/frame/broker/src/utility_impls.rs
```diff
@@ -138,4 +138,39 @@ impl<T: Config> Pallet<T> {
...
+			if region.end <= end_timeslice {
+				return
+			}
+
+			// Account for the change in `InstaPoolIo` either from the start of the region or from
+			// the current timeslice if we are already part-way through the region.
+			let size = region_id.mask.count_ones() as i32;
+			let unpooled_at = end_timeslice.max(region_id.begin);
+			InstaPoolIo::<T>::mutate(unpooled_at, |a| a.private.saturating_reduce(size));
+			InstaPoolIo::<T>::mutate(region.end, |a| a.private.saturating_accrue(size));
```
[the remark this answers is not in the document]
seadanda: When the region was first pooled, it was accounted for in the `InstaPoolIo` by adding its size at the `region_id.start`, and reducing it at the `region.end` timeslices. Now that we're unpooling it from the `unpooled_at` timeslice, we need to reduce the `InstaPoolIo` by its size to show it exits the pool there, then we need to undo the original IO accounting at that timeslice by adding it back `region.end`

seadanda:
OK finally got some time to fix the mock and tests now that on-demand credits have been properly implemented.

substrate/frame/broker/src/utility_impls.rs
```diff
@@ -138,4 +138,39 @@ impl<T: Config> Pallet<T> {
...
+			// this timeslice would have region.end == unpooled_at below.
+			if region.end <= end_timeslice {
+				return
+			}
+
+			// Account for the change in `InstaPoolIo` either from the start of the region or from
+			// the current timeslice if we are already part-way through the region.
+			let size = region_id.mask.count_ones() as i32;
+			let unpooled_at = end_timeslice.max(region_id.begin);
+			InstaPoolIo::<T>::mutate(unpooled_at, |a| a.private.saturating_reduce(size));
```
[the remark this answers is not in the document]
seadanda: Yes, the caller will change the Workplan via `assign`. If they are just interlacing or partitioning their core then it doesn't matter that the core is not reassigned away from the pool until the point they `assign` it, in which case the workplan will be updated.
If somebody has a core that they have provisionally pooled, they'll need to `assign(Idle)` if they want to explicitly remove it from the pool, and this again is an operation that updates the Workplan.

prdoc/pr_4081.prdoc
```diff
@@ -0,0 +1,16 @@
...
+  - audience: Runtime User
+    description: |
+      This PR force removes regions from the pool before allowing them to be redispatched (through
+      `partition`/`interlace`/`assign`) for regions pooled with `Provisional` finality. To claim
+      any revenue from before this point, `claim_revenue` should be called before
+      partitioning/interleaving/reassigning as it cannot be claimed afterwards.
+
+crates:
+- name: pallet-broker
+  bump: patch
```
seadanda:
```suggestion
  bump: major
```

bkchr:
@eskimor please finish your review.

eskimor (approved):

PR #4085, opened 2024-04-11
[Draft] Fix usage of ROCOCO_EPOCH_DURATION for different builds

alexggh:
All  rococo blocks produced by `build-linux-stable`  seems to be identical and the session time is 1minute, instead of 1min, 10min and 60min.


polkadot/scripts/build-only-wasm.sh
```diff
@@ -27,11 +23,17 @@ else
...
+  fl_cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" --features fast-runtime \
     | grep -vE "cargo:rerun-if-|Executing build command"
 else
-  fl_cargo build --release -p $1
+  fl_cargo build --release -p $1 --features fast-runtime
 fi
+
+echo "ROCOCO_EPOCH_DURATION=$ROCOCO_EPOCH_DURATION checksums"
+
+md5sum $WASM_TARGET_DIRECTORY/*
```
bkchr, line 39:
```suggestion
```

Not sure what you want to achieve by this?

polkadot/scripts/build-only-wasm.sh
```diff
@@ -27,11 +23,17 @@ else
...
+
 if [ -d $WASM_BUILDER_RUNNER ]; then
   export DEBUG=false
   export OUT_DIR="$PROJECT_ROOT/target/release/build"
-  fl_cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" \
+  fl_cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" --features fast-runtime \
     | grep -vE "cargo:rerun-if-|Executing build command"
 else
-  fl_cargo build --release -p $1
+  fl_cargo build --release -p $1 --features fast-runtime
```
bkchr, line 34: This script is generic. If you want to be able to pass `--features`, you should make it possible from the outside to pass arguments.

bkchr (changes requested):

polkadot/scripts/build-only-wasm.sh
alexggh: Should have mark it as a draft I was debugging it.

polkadot/scripts/build-only-wasm.sh
alexggh: Yeah, this was not ready for review, forgot to mark it as a draft.

PR #4091, opened 2024-04-12
Remove `pallet::getter` usage from authority-discovery pallet

PolkadotDom:
As per #3326, removes pallet::getter usage from the pallet authority-discovery. The syntax `StorageItem::<T, I>::get()` should be used instead.

cc @muraca


muraca:
@kianenigma @liamaharon @ggwpez

prdoc/pr_4091.prdoc
```diff
@@ -0,0 +1,13 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Remove `pallet::getter` usage from authority-discovery pallet
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR removes `pallet::getter` usage from `pallet-authority-discovery`.
```
liamaharon, line 9:
```suggestion
      This PR removes `pallet::getter`s from `pallet-authority-discovery`.
```

PolkadotDom:
@liamaharon resolved the previous comment if you have a second for re-review

prdoc/pr_4091.prdoc
```diff
@@ -0,0 +1,13 @@
...
+title: Remove `pallet::getter` usage from authority-discovery pallet
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR removes `pallet::getter`s from `pallet-authority-discovery`.
+      The syntax `StorageItem::<T, I>::get()` should be used instead.
+
+crates:
+  - name: pallet-authority-discovery
```
muharem, line 13: you also need to provide a `bump` item here, please check the schema

PolkadotDom: Ah, good looks. I've added the appropriate bump now 👍

prdoc/pr_4091.prdoc
```diff
@@ -0,0 +1,14 @@
...
+    bump: minor
```
liamaharon, line 14: this is a breaking change
```suggestion
    bump: major
```

PolkadotDom: Updated 👍 Though the storage values are private, no? Or is it that someone might have customized the pallet and this could break their version?

liamaharon: The pallet exposes the getters in a public api, so this breaks that public api

liamaharon (approved):

PolkadotDom:
Realized I need to implement getters for the storage values since they were previously public. Hold on review for the moment.

muharem:
@PolkadotDom I did not know they were public, good catch. when you introduce them, these changes wont be breaking anymore, you can change bump to minor

PolkadotDom:
> @PolkadotDom I did not know they were public, good catch. when you introduce them, these changes wont be breaking anymore, you can change bump to minor

Yeah, I believe the getter macro makes the storage item read-only outside the crate. But that'd mean it should still be major since it'll break the pre-existing api and users would have to switch to my new explicitly implemented getters. Unless I'm thinking of this wrong :shrug:

PolkadotDom:
@muharem Actually, it looks like there was a bit of a redundancy. The pallet previously had both macro generated getters as well as explicit getters. So instead of implementing new ones for the lost macros I'm just going to keep the bump major and update the prdoc to reflect that the explicit getters should be used moving forward. Let me know if you think it should be otherwise.

bkchr (approved):

Issue #3967, opened 2024-04-03
slot-based-collator: Adjust ConsensusHook and parent search


skunert:
When a new parachain block is authored, we search for a parent that fits all constraints of the relay chain. This means that the parachain parent we want to build on needs to be anchored to a relay parent that is within the defined ancestry parameters. The logic currently assumes that we only have one pending candidate, which is not longer true.
https://github.com/paritytech/polkadot-sdk/blob/cdacfb9d33710093d67e96cd6767dba0243ff450/cumulus/client/consensus/common/src/lib.rs#L365

We should make this aware of multiple pending candidates, after we have https://github.com/paritytech/polkadot-sdk/issues/3576

In addition, adjust (or create a new one) the `FixedVelocityConsensusHook` to take into account sub slots < 6s. It currently does its checks based on blocks authored so far in the current parachain slot: 
https://github.com/paritytech/polkadot-sdk/blob/cdacfb9d33710093d67e96cd6767dba0243ff450/cumulus/pallets/aura-ext/src/consensus_hook.rs#L103

Issue #3966, opened 2024-04-03
slot-based-collator: Collation & block-builder communication


skunert:
Collation and block-builder task in the parachain node should be able to communicate with each other:
- Collation task can tell block-builder to stop building if we are not able to submit collations for a while
- Collation task can tell block-builder to start authoring if a new scheduled core is coming up in the future (could maybe be handled in the block-builder task itself, we need to check).

skunert:
Not needed anymore, similar ideas in #6066

Issue #3965, opened 2024-04-03
slot-based-collator: Allow slot drift


skunert:
When authoring with the slot based collator, it would be nice to allow building on some older relay parents, limited by the asynchronous backing config of the relay chain of course. Also should be allowed to specify some fixed time-drift.

Part of #3168

skunert:
Is now tracked in https://github.com/paritytech/polkadot-sdk/issues/7780

PR #4097, opened 2024-04-12
Introduce basic slot-based collator

skunert:
Part of #3168 
On top of #3568

### Changes Overview
- Introduces a new collator variant in `cumulus/client/consensus/aura/src/collators/slot_based/mod.rs`
- Two tasks are part of that module, one for block building and one for collation building and submission.
- Introduces a new variant of `cumulus-test-runtime` which has 2s slot duration, used for zombienet testing
- Zombienet tests for the new collator

**Note:** This collator is considered experimental and should only be used for testing and exploration for now.

### Comparison with `lookahead` collator
- The new variant is slot based, meaning it waits for the next slot of the parachain, then starts authoring
- The search for potential parents remains mostly unchanged from lookahead
- As anchor, we use the current best relay parent
- In general, the new collator tends to be anchored to one relay parent earlier. `lookahead` generally waits for a new relay block to arrive before it attempts to build a block. This means the actual timing of parachain blocks depends on when the relay block has been authored and imported. With the slot-triggered approach we are authoring directly on the slot boundary, were a new relay chain block has probably not yet arrived.

### Limitations
- Overall, the current implementation focuses on the "happy path"
- We assume that we want to collate close to the tip of the relay chain. It would be useful however to have some kind of configurable drift, so that we could lag behind a bit. https://github.com/paritytech/polkadot-sdk/issues/3965
- The collation task is pretty dumb currently. It checks if we have cores scheduled and if yes, submits all the messages we have received from the block builder until we have something submitted for every core. Ideally we should do some extra checks, i.e. we do not need to submit if the built block is already too old (build on a out of range relay parent) or was authored with a relay parent that is not an ancestor of the relay block we are submitting at. https://github.com/paritytech/polkadot-sdk/issues/3966
- There is no throttling, we assume that we can submit _velocity_ blocks every relay chain block. There should be communication between the collator task and block-builder task.
- The parent search and ConsensusHook are not yet properly adjusted. The parent search makes assumptions about the pending candidate which no longer hold. https://github.com/paritytech/polkadot-sdk/issues/3967
- Custom triggers for block building not implemented.


skunert:
I made a new round of adjustments to the collator based on the testing done locally and on versi by @sandreim .

### Collation Submission
In the first version of this PR we were building blocks when we had some cores scheduled and let the collation task figure out which blocks should be submitted for which cores. This can make sense when we are only running a single collator (like mythical MVP scenario). In that scenario we could keep blocks around for later submission to the relay chain. However in general, this approach is flawed (at least for now) since we don't know what the other collators have submitted. I changed the logic to just submit "now-or-never" style. We build a block for a fixed core and inform the collation task to submit immediately to that core. 

This raises the question if it was necessary to split these tasks in the first place. I think it is not strictly necessary. However I am in favor of keeping it this way, since it keeps the preparation of the collation out of the block building loop and provides some nice separation of concerns. In the future we might want to submit collations that were not only build on our own collator, so I think it still makes sense. 

### Parent Search
I  iterated on the potential parent search a bit. In previous commits I have integrated support for pending blocks that are not an immediate child of the included block. This is needed for elastic scaling, since we submit multiple blocks at once and a whole chain can be backed simultaneously. This introduced some more edge cases to the algorithm, so I added comments and factored out some methods to make it easier to follow. We now also check that the included and pending block are actually known to the local node. If we don't know included or pending block, we should not consider them potential parents.

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,453 @@
...
+			.await
+		else {
+			tracing::error!(target: crate::LOG_TARGET, "Unable to build block at slot.");
+			continue;
+		};
+
+		let new_block_hash = candidate.block.header().hash();
+
+		// Announce the newly built block to our peers.
+		collator.collator_service().announce_block(new_block_hash, None);
```
davxy, line 384: Is correct to announce the block here?
I was thinking about announcing it in the collator task if `build_collation` goes well.
(I see that the "lookahead" does it in this way (i.e. announces if collation goes well)

cumulus/client/consensus/aura/src/collators/slot_based/collation_task.rs
```diff
@@ -0,0 +1,140 @@
...
+	pub collator_receiver: TracingUnboundedReceiver<CollatorMessage<Block>>,
+}
+
+/// Asynchronously executes the collation task for a parachain.
+///
+/// This function initializes the collator subsystems necessary for producing and submitting
+/// collations to the relay chain. It listens for new best relay chain block notifications and
+/// handles collator messages. If our parachain is scheduled on a core and we have a candidate,
+/// the task will build a collation and send it to the relay chain.
+pub async fn run_collation_task<Block, RClient, CS>(mut params: Params<Block, RClient, CS>)
```
davxy, line 58: Would be nice if a task like this is usable by the other engines as well. Maybe is not worth limiting this task to this particular "slot-based" engine and also extend its usage to "basic" and "lookahead".

What about adding this task directly to the `Collator` struct? So that one can decide to optionally spawn it (if the user wants to use a task based mechanism to build and announce collations)

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
skunert: Hmm indeed it is a change to announce it here. Its a bit of a hassle. If we are unable to build the collation the block can not be submitted to the relay chain and has no chance of getting included. On the other hand, the faster we announce the faster the next block builder can start importing and building. In theory, the collation building can fail but I would say its quite unlikely, we are just fetching some data and compressing the PoV.
Will think about it.

Polkadot-Forum:

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,390 @@
...
+
+		let scheduled_cores = cores_scheduled_for_para(relay_parent, para_id, &relay_client).await;
+		if scheduled_cores.is_empty() {
+			tracing::debug!(target: LOG_TARGET, "Parachain not scheduled, skipping slot.");
+			continue;
+		}
+
+		let core_index_in_scheduled: u64 = *para_slot.slot % expected_cores;
+		let Some(core_index) = scheduled_cores.get(core_index_in_scheduled as usize) else {
+			tracing::debug!(target: LOG_TARGET, core_index_in_scheduled, core_len = scheduled_cores.len(), "Para is scheduled, but not enough cores available.");
```
sandreim, line 229: For fixed scaling might make sense to have this as a warning, so we can act on it.

davxy (commented):
New review iteration. Mostly really good.
I'd just like some clarification on some stuff :-)

cumulus/zombienet/tests/assign-core.js
```diff
@@ -0,0 +1,46 @@
...
+  await zombie.util.cryptoWaitReady();
+
+  // Submit transaction with Alice accoung
+  const keyring = new zombie.Keyring({ type: "sr25519" });
+  const alice = keyring.addFromUri("//Alice");
+
+  // Wait for this transaction to be finalized in a block.
+  await new Promise(async (resolve, reject) => {
+    const unsub = await api.tx.sudo
+      .sudo(api.tx.coretime.assignCore(core, 0, [[{ task: para }, 57600]], null))
```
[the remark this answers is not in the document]
skunert: This call goes to the coretime pallet of the relay chain directly: https://github.com/paritytech/polkadot-sdk/blob/8d8c0e13a7dc8d067367ac55fb142b12ac8a6d13/polkadot/runtime/parachains/src/coretime/mod.rs#L195

It is not part of the broker chain.

cumulus/pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -65,8 +65,17 @@ where
...
+		// during the relay chain slot, we can allow for `V` parachain slots into the future.
+		if *slot > *para_slot_from_relay + u64::from(velocity) {
+			panic!(
+				"Parachain slot is too far in the future: parachain_slot: {:?}, derived_from_relay_slot: {:?} velocity: {:?}",
+				slot,
+				para_slot_from_relay,
+				velocity
+			);
+		}
+
```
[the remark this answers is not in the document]
skunert: Good catch!

cumulus/zombienet/tests/assign-core.js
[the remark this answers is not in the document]
davxy: Ops yeah is in the relay. Ty

cumulus/client/consensus/aura/src/collators/slot_based/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+mod block_builder_task;
+mod collation_task;
+
+/// Parameters for [`run`].
+pub struct Params<BI, CIDP, Client, Backend, RClient, CHP, Proposer, CS> {
+	/// Inherent data providers. Only non-consensus inherent data should be provided, i.e.
+	/// the timestamp, slot, and paras inherents should be omitted, as they are set by this
+	/// collator.
+	pub create_inherent_data_providers: CIDP,
```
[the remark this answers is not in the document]
skunert: Hmm I don't fully understand your confusion. What you are describing is the same as what  the comment says. The  data is set set byt this collator aka this collator implementation.

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,400 @@
...
+				tracing::warn!(target: PARENT_SEARCH_LOG_TARGET, "Included block not an ancestor of pending block. This should not happen.");
+				return Ok(Default::default())
+			}
+
+			// Add all items on the path included -> pending - 1 to the potential parents, but
+			// not more than `max_depth`.
+			let num_parents_on_path =
+				route_to_pending.enacted().len().saturating_sub(1).min(params.max_depth);
+			for (num, block) in
+				route_to_pending.enacted().iter().take(num_parents_on_path).enumerate()
```
bkchr, line 181:
```suggestion
			for (num, block) in
				route_to_pending.enacted().rev().skip(1).take(params.max_depth - 1).enumerate()
```

Should we not favor blocks with a higher block number as parent? The function also does not give any information on the "ranking" of the returned values.

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,400 @@
...
+					aligned_with_pending: true,
+				}],
+				potential_parents,
+			)
+		},
+		_ => (only_included, Default::default()),
+	};
+
+	if potential_parents.len() > params.max_depth {
+		return Ok(potential_parents);
```
bkchr, line 209: This means we don't include the `pending` block? Sounds wrong to me?

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,400 @@
...
+	relay_client: &impl RelayChainInterface,
+) -> Result<Vec<(RelayHash, RelayHash)>, RelayChainError> {
+	let mut ancestry = Vec::with_capacity(ancestry_lookback + 1);
+	let mut current_rp = relay_parent;
+	let mut required_session = None;
+	while ancestry.len() <= ancestry_lookback {
+		let header = match relay_client.header(RBlockId::hash(current_rp)).await? {
+			None => break,
+			Some(h) => h,
+		};
```
bkchr, line 291:
```suggestion
		let Some(header) = relay_client.header(RBlockId::hash(current_rp)).await? else {
			break
		};
```

cumulus/client/consensus/aura/src/collators/slot_based/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+		collator_service: params.collator_service,
+		authoring_duration: params.authoring_duration,
+		collator_sender: tx,
+		relay_chain_slot_duration: params.relay_chain_slot_duration,
+	};
+
+	let block_builder_fut =
+		run_block_builder::<Block, P, _, _, _, _, _, _, _, _>(block_builder_params);
+
+	(collation_task_fut, block_builder_fut)
```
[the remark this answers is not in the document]
bkchr: We should not use a `select!` here. We want that these futures are running in different blocking tasks. However, we could just spawn them here internally. I don't see a reason to return them.

cumulus/client/consensus/aura/src/collators/slot_based/collation_task.rs
```diff
@@ -0,0 +1,140 @@
...
+				return;
+			},
+		};
+
+	tracing::info!(
+		target: LOG_TARGET,
+		"PoV size {{ header: {:.2}kB, extrinsics: {:.2}kB, storage_proof: {:.2}kB }}",
+		block_data.header().encode().len() as f64 / 1024f64,
+		block_data.extrinsics().encode().len() as f64 / 1024f64,
+		block_data.storage_proof().encode().len() as f64 / 1024f64,
```
bkchr, line 115:
```suggestion
		block_data.header().encoded_len() as f64 / 1024f64,
		block_data.extrinsics().encoded_len() as f64 / 1024f64,
		block_data.storage_proof().encoded_len() as f64 / 1024f64,
```

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,390 @@
...
+	pub proposer: Proposer,
+	/// The generic collator service used to plug into this consensus engine.
+	pub collator_service: CS,
+	/// The amount of time to spend authoring each block.
+	pub authoring_duration: Duration,
+	/// Channel to send built blocks to the collation task.
+	pub collator_sender: sc_utils::mpsc::TracingUnboundedSender<CollatorMessage<Block>>,
+	/// Slot duration of the relay chain
+	pub relay_chain_slot_duration: Duration,
+	/// Drift every slot by this duration.
```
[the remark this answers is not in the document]
bkchr: Yes! :D

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,492 @@
...
+/// Calculate the expected core count based on the slot duration of the relay and parachain.
+///
+/// If `slot_duration` is smaller than `relay_chain_slot_duration` that means that we produce more
+/// than one parachain block per relay chain block. In order to get these backed, we need multiple
+/// cores. This method calculates how many cores we should expect to have scheduled under the
+/// assumption that we have a fixed number of cores assigned to our parachain.
+fn expected_core_count(
+	relay_chain_slot_duration: Duration,
+	slot_duration: SlotDuration,
+) -> Result<u64, ()> {
```
bkchr, line 400:
```suggestion
) -> Option<u64> {
```

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,492 @@
...
+			relay_parent,
+			para_id,
+			&*para_backend,
+			&relay_client,
+		)
+		.await
+		{
+			Some(value) => value,
+			None => continue,
+		};
```
bkchr, line 275:
```suggestion
		let Some((included_block, parent)) = crate::collators::find_parent(
			relay_parent,
			para_id,
			&*para_backend,
			&relay_client,
		)
		.await else
		{
			continue
		};
```

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
bkchr: Yeah it is fine to announce it here directly. The set is restricted and thus, others can reject if there are "too many" blocks. The runtime also ensures that we are not building too far ahead into the future.

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,492 @@
...
+	pub relay_parent_hash: RelayHash,
+}
+
+/// Simple helper to fetch relay chain data and cache it based on the current relay chain best block
+/// hash.
+struct RelayChainCachingFetcher<RI> {
+	relay_client: RI,
+	para_id: ParaId,
+	last_seen_hash: Option<RelayHash>,
+	last_data: Option<RelayChainData>,
```
bkchr, line 426: Could we one `Option`. Then we don't need nested `if`s

cumulus/client/consensus/aura/src/collators/slot_based/collation_task.rs
```diff
@@ -0,0 +1,140 @@
...
+	/// The collator key used to sign collations before submitting to validators.
+	pub collator_key: CollatorPair,
+	/// The para's ID.
+	pub para_id: ParaId,
+	/// Whether we should reinitialize the collator config (i.e. we are transitioning to aura).
+	pub reinitialize: bool,
+	/// Collator service interface
+	pub collator_service: CS,
+	/// Receiver channel for communication with the block builder task.
+	pub collator_receiver: TracingUnboundedReceiver<CollatorMessage<Block>>,
```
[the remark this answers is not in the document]
bkchr: Block production is already limited by the unincluded segment. Aka if nothing is free, the block production will stop.

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,492 @@
...
+		collator.collator_service().announce_block(new_block_hash, None);
+
+		if let Err(err) = collator_sender.unbounded_send(CollatorMessage {
+			relay_parent,
+			parent_header,
+			parachain_candidate: candidate,
+			validation_code_hash,
+			core_index: *core_index,
+		}) {
+			tracing::error!(target: crate::LOG_TARGET, ?err, "Unable to send block to collation task.");
```
bkchr, line 386:
```suggestion
			tracing::error!(target: crate::LOG_TARGET, ?err, "Unable to send block to collation task.");
			break
```

This means that the collation task is dead and thus, we should also end the block production.

bkchr (commented):
Nice work! Generally everything looks good! I will need to do another pass over the parent search to ensure that I understood everything ;)

cumulus/test/runtime/src/lib.rs
```diff
@@ -83,8 +88,23 @@ impl_opaque_keys! {
...
+
+#[cfg(feature = "elastic-scaling")]
+const UNINCLUDED_SEGMENT_CAPACITY: u32 = 6;
+#[cfg(feature = "elastic-scaling")]
+const BLOCK_PROCESSING_VELOCITY: u32 = 3;
+
+#[cfg(not(feature = "elastic-scaling"))]
+pub const MILLISECS_PER_BLOCK: u64 = 6000;
+#[cfg(feature = "elastic-scaling")]
+pub const MILLISECS_PER_BLOCK: u64 = 2000;
```
alindima, line 104: we could derive all these constants from `MILLISECS_PER_BLOCK` and `RELAY_CHAIN_SLOT_DURATION_MILLIS`. we'd also avoid most of these feature guards

cumulus/test/runtime/src/lib.rs
```diff
@@ -83,8 +88,23 @@ impl_opaque_keys! {
...
 pub const PARACHAIN_ID: u32 = 100;
 
-const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
+#[cfg(not(feature = "elastic-scaling"))]
+const UNINCLUDED_SEGMENT_CAPACITY: u32 = 4;
+#[cfg(not(feature = "elastic-scaling"))]
 const BLOCK_PROCESSING_VELOCITY: u32 = 1;
+
+#[cfg(feature = "elastic-scaling")]
+const UNINCLUDED_SEGMENT_CAPACITY: u32 = 6;
```
alindima, line 97: we used 7 for this in tests. did we find the root-cause of this? was it fixed?

cumulus/zombienet/tests/assign-core.js
```diff
@@ -0,0 +1,46 @@
+// Assign a parachain to a core.
```
alindima, line 1: could we reuse the assign-core script from the polkadot folder? maybe move it to some common zombienet helper scripts folder in the root folder

cumulus/zombienet/tests/0008-elastic_authoring.zndsl
```diff
@@ -0,0 +1,15 @@
...
+collator-elastic: is up
+collator-single-core: is up
+
+
+# configure relay chain
+alice: js-script ./assign-core.js with "2100,0" return is 0 within 600 seconds
+alice: js-script ./assign-core.js with "2100,1" return is 0 within 600 seconds
+
+collator-single-core: reports block height is at least 20 within 225 seconds
+collator-elastic: reports block height is at least 40 within 225 seconds
```
alindima, line 15: I think this second check should have a close to zero timeout since the checks are sequential (because if the single core had enough time to back and include 20 blocks already, the elastic scaling one should have done twice the blocks in the same amount of time)

alindima:
now that https://github.com/paritytech/polkadot-sdk/pull/4733 was merged, please also include a zombienet test for pov recovery while using elastic scaling, if it's not a big ask. Otherwise, I can do it separately after this is merged

cumulus/client/consensus/aura/src/collators/slot_based/collation_task.rs
alindima: This should be encoded_size()

cumulus/test/runtime/src/lib.rs
skunert: IIRC we used 7 here together with the slot drift to fix the block spikes every 8 blocks. Also we found that max_candidate_depth=6 equals an unincluded segment of 7.

skunert:
@davxy @bkchr I think this is ready for approval farming and last nits now.

cumulus/client/consensus/aura/src/collators/slot_based/collation_task.rs
```diff
@@ -0,0 +1,140 @@
...
+
+	tracing::info!(
+		target: LOG_TARGET,
+		"PoV size {{ header: {}kb, extrinsics: {}kb, storage_proof: {}kb }}",
+		block_data.header().encode().len() as f64 / 1024f64,
+		block_data.extrinsics().encode().len() as f64 / 1024f64,
+		block_data.storage_proof().encode().len() as f64 / 1024f64,
+	);
+
+	if let MaybeCompressedPoV::Compressed(ref pov) = collation.proof_of_validity {
```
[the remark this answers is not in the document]
skunert: I agree, however lets do it in a follow up. There should be metrics added to the RPC client too.

cumulus/pallets/aura-ext/src/consensus_hook.rs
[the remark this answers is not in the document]
skunert: Ha, unresolving here for this note: The +1 here was for a situation where we have not seen a new relay block notification yet, so we need to build one more block on an older relay parent. To allow that we need to allow one more para slot here. However, the +1 should not apply to the consensus hook in general, the better solution for this MVP is to increase the velocity in the runtime by 1.

cumulus/zombienet/tests/0008-elastic_authoring.toml
```diff
@@ -0,0 +1,50 @@
...
+  args = ["-laura=trace,runtime=info,cumulus-consensus=trace,consensus::common=trace,parachain::collation-generation=trace,parachain::collator-protocol=trace,parachain=debug", "--force-authoring", "--experimental-use-slot-based"]
+
+# Slot based authoring with 1 core and 6s slot duration
+[[parachains]]
+id = 2000
+add_to_genesis = true
+
+  [[parachains.collators]]
+  name = "collator-single-core"
+  image = "{{CUMULUS_IMAGE}}"
```
pepoviola, line 48:
```suggestion
  image = "{{COL_IMAGE}}"
```

cumulus/zombienet/tests/0008-elastic_authoring.toml
```diff
@@ -0,0 +1,50 @@
...
+
+# Slot based authoring with 3 cores and 2s slot duration
+[[parachains]]
+id = 2100
+chain = "elastic-scaling"
+add_to_genesis = true
+
+  [[parachains.collators]]
+  name = "collator-elastic"
+  image = "{{CUMULUS_IMAGE}}"
```
pepoviola, line 37:
```suggestion
  image = "{{COL_IMAGE}}"
```

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,398 @@
...
+	}];
+
+	if params.max_depth == 0 {
+		return Ok(only_included)
+	};
+
+	// Pending header and hash.
+	let maybe_pending = {
+		// Fetch the pending header from the relay chain. We use `OccupiedCoreAssumption::Included`
+		// so the candidate pending availability gets enacted before being returned to  us.
```
bkchr, line 120:
```suggestion
		// so the candidate pending availability gets enacted before being returned to us.
```

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,398 @@
...
+					.map_or(false, is_root_in_ancestry);
+
+			is_hash_in_ancestry_check || is_root_in_ancestry_check
+		};
+
+		let parent_aligned_with_pending = entry.aligned_with_pending;
+		let child_depth = entry.depth + 1;
+		let hash = entry.hash;
+
+		tracing::trace!(target: PARENT_SEARCH_LOG_TARGET, root_in_ancestry = is_potential && !is_pending && !is_included, ?hash, is_pending, is_included, "Checking potential parent.");
```
bkchr, line 363: `root_is_ancestry`, as I understand it the root could still be in ancestry, even if the block is pending or is included? Maybe rename or change the conditions?

cumulus/client/consensus/aura/src/collators/slot_based/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! A collator for Aura that looks ahead of the most recently included parachain block
```
bkchr, line 17: This is still copy pasta?

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,488 @@
...
+{
+	pub fn new(relay_client: RI, para_id: ParaId) -> Self {
+		Self { relay_client, para_id, last_data: None }
+	}
+
+	/// Fetch required [`RelayChainData`] from the relay chain.
+	/// If this data has been fetched in the past for the incoming hash, it will reuse
+	/// cached data.
+	pub async fn get_relay_chain_data(&mut self) -> Result<RelayChainData, ()> {
+		let Ok(relay_parent) = self.relay_client.best_block_hash().await else {
```
bkchr, line 438: A problem with this could be that if the relay chain block comes too fast, we may did not use all the available cores of the parent and then we are wasting one core. Not really required for this pull request, but we should make this more robust in the future.

bkchr (approved):
Nice work!

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,488 @@
...
+	drift: Duration,
+	_marker: std::marker::PhantomData<(Block, Box<dyn Fn(P) + Send + Sync + 'static>)>,
+}
+
+/// Returns current duration since Unix epoch.
+fn duration_now() -> Duration {
+	use std::time::SystemTime;
+	let now = SystemTime::now();
+	now.duration_since(SystemTime::UNIX_EPOCH).unwrap_or_else(|e| {
+		panic!("Current time {:?} is before unix epoch. Something is wrong: {:?}", now, e)
```
davxy, line 117:
```suggestion
		panic!("Current time {:?} is before Unix epoch. Something is wrong: {:?}", now, e)
```

cumulus/client/consensus/aura/src/collators/slot_based/block_builder_task.rs
```diff
@@ -0,0 +1,488 @@
...
+	pub collator_service: CS,
+	/// The amount of time to spend authoring each block.
+	pub authoring_duration: Duration,
+	/// Channel to send built blocks to the collation task.
+	pub collator_sender: sc_utils::mpsc::TracingUnboundedSender<CollatorMessage<Block>>,
+	/// Slot duration of the relay chain
+	pub relay_chain_slot_duration: Duration,
+	/// Drift every slot by this duration.
+	/// This can be used to shift slots in relation to relay chain slots,
+	/// making it less likely to to encounter unlucky notification arrival timings.
```
davxy, line 94:
```suggestion
	/// This is a time quantity that is subtracted from the actual timestamp when computing
	/// the time left to enter a new slot. In practice, this *left-shifts* the clock time with the intent
	/// to keep our "clock" slightly behind the relay chain one and thus reducing the likelihood of
	/// encountering unfavorable notification arrival timings (i.e. we don't want to wait for relay
	/// chain notifications because we woke up too early).
```
I know is verbose :-)  But as a user I'd otherwise find very difficult to understand this

davxy (approved):
LGTM

cumulus/client/consensus/common/src/parent_search.rs
```diff
@@ -0,0 +1,398 @@
...
+		aligned_with_pending: true,
+	}];
+
+	if params.max_depth == 0 {
+		return Ok(only_included)
+	};
+
+	// Pending header and hash.
+	let maybe_pending = {
+		// Fetch the pending header from the relay chain. We use `OccupiedCoreAssumption::Included`
```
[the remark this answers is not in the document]
skunert: The comment correction you suggested is right. However, the unincluded segment can be longer. The pending header has already been backed. There can be blocks in the unincluded segment that are not yet pending availability.

skunert:
bot fmt

Polkadot-Forum:

https://forum.polkadot.network/t/elastic-scaling-mvp-launched/9392/1

PR #4089, opened 2024-04-11
pallet_broker: Support renewing leases expired in a previous period

bkchr:
Part of: https://github.com/paritytech/polkadot-sdk/issues/4107

PR #4099, opened 2024-04-12
[pallet-broker] add tests for renewing leases

seadanda:
The first test proves that parachains who were migrated over on a legacy lease can renew without downtime.

The exception is if their lease expires in period 0 - aka within `region_length` timeslices after `start_sales` is called. The second test is designed such that it passes if the issue exists and should be fixed.
This will require an intervention on Kusama to add these renewals to storage as it is too tight to schedule a runtime upgrade before the start_sales call. All leases will still have at least two full regions of coretime.


substrate/frame/broker/src/tests.rs
```diff
@@ -891,6 +891,164 @@ fn short_leases_are_cleaned() {
...
+		// Sale 1 starts at block 7, Sale 2 starts at 13.
+
+		// Add a core.
+		assert_ok!(Broker::request_core_count(RuntimeOrigin::root(), 1));
+
+		// Set lease to expire in sale 1 and start sales.
+		assert_ok!(Broker::do_set_lease(2001, 9));
+		assert_eq!(Leases::<Test>::get().len(), 1);
+		// Start the sales but don't offer any more cores.
+		assert_ok!(Broker::do_start_sales(100, 0));
```
eskimor, line 908: Not just not any more, but no cores at all.

substrate/frame/broker/src/tests.rs
```diff
@@ -891,6 +891,164 @@ fn short_leases_are_cleaned() {
...
+						.unwrap()
+				)
+			})
+		);
+		// And the lease has been removed from storage.
+		assert_eq!(Leases::<Test>::get().len(), 0);
+
+		// Advance to sale period 2, where we can renew.
+		advance_to(13);
+		assert_ok!(Broker::do_renew(1, 0));
```
eskimor, line 929: do_renew actually checks [cores for sale](https://github.com/paritytech/polkadot-sdk/blob/1c85bfe901741f9c456af1ac92008d647660a2f4/substrate/frame/broker/src/dispatchable_impls.rs#L134).

This only works, because of the request_core_count above, which takes effect after the `do_start_sales`. The 0 is basically overridden more or less immediately.

eskimor (approved):

substrate/frame/broker/src/tests.rs
```diff
@@ -891,6 +891,164 @@ fn short_leases_are_cleaned() {
...
+	TestExt::new().endow(1, 1000).execute_with(|| {
+		// Timeslice period is 2.
+		//
+		// Sale 1 starts at block 7, Sale 2 starts at 13.
+
+		// Add a core.
+		assert_ok!(Broker::request_core_count(RuntimeOrigin::root(), 1));
+
+		// Set lease to expire in sale 1 and start sales.
+		assert_ok!(Broker::do_set_lease(2001, 9));
```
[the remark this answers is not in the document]
eskimor: Like:

```rust
Broker::do_set_lease(TaskId(2001), Timeslice(9))
```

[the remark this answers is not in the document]
bkchr: Actually it should be `Region(9)` :P

substrate/frame/broker/src/tests.rs
```diff
@@ -891,6 +891,164 @@ fn short_leases_are_cleaned() {
...
+		assert_eq!(Leases::<Test>::get().len(), 1);
+		// Start the sales but don't offer any more cores.
+		assert_ok!(Broker::do_start_sales(100, 0));
+
+		// Advance to sale period 1, we should get an AllowedRenewal for task 2001 for the next
+		// sale.
+		advance_to(7);
```
bkchr, line 912: Maybe better to use what I introduced here: https://github.com/paritytech/polkadot-sdk/pull/4089/files#diff-9f7f1020b99ae1a246e61572423e1fac95ee6d42802901d822745c74a54488beR213 :D

substrate/frame/broker/src/tests.rs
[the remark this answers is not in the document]
eskimor: :rofl: Ok, thanks for proving my point. And to prove It even more: It should really be `TimeSlice(9)`: https://github.com/paritytech/polkadot-sdk/blob/1c85bfe901741f9c456af1ac92008d647660a2f4/substrate/frame/broker/src/dispatchable_impls.rs#L63

substrate/frame/broker/src/tests.rs
seadanda: Added in [8ab34b0](https://github.com/paritytech/polkadot-sdk/pull/4099/commits/8ab34b0655c68249ead5ea17f9bb1225b7858658)

seadanda: Ah jeez that resets the reviews now 🤦‍♂️

ggwpez (approved):
(approving to unblock)

PR #4102, opened 2024-04-12
Bridge: make some headers submissions free

svyatonik:
supersedes https://github.com/paritytech/parity-bridges-common/pull/2873

Draft because of couple of TODOs:
- [x] fix remaining TODOs;
- [x] double check that all changes from https://github.com/paritytech/parity-bridges-common/pull/2873 are correctly ported;
- [x] create a separate PR (on top of that one or a follow up?) for https://github.com/paritytech/polkadot-sdk/tree/sv-try-new-bridge-fees;
- [x] fix compilation issues (haven't checked, but there should be many).


paritytech-cicd-pr:
**Job name**: cargo-clippy
**Logs**: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/5906926

bridges/bin/runtime-common/src/extensions/check_obsolete_extension.rs
```diff
@@ -101,29 +282,72 @@ macro_rules! generate_bridge_reject_obsolete_headers_and_messages {
...
+			fn post_dispatch(
+				to_post_dispatch: Option<Self::Pre>,
+				info: &sp_runtime::traits::DispatchInfoOf<Self::Call>,
+				post_info: &sp_runtime::traits::PostDispatchInfoOf<Self::Call>,
+				len: usize,
+				result: &sp_runtime::DispatchResult,
+			) -> Result<(), sp_runtime::transaction_validity::TransactionValidityError> {
+				use tuplex::PopFront;
+				let Some((relayer, to_post_dispatch)) = to_post_dispatch else { return Ok(()) };
+				let has_failed = result.is_err();
```
svyatonik, line 340: Important difference with original PR here: previously there was a `return` here if `has_failed` has been false: https://github.com/paritytech/parity-bridges-common/pull/2873/files#diff-07de356c7872e1dee65e25f62dadf83ba489866c7a603ca733cd6df13d5e06ceR353-R355. But since our parachains pallet still may return `Ok(_)` even if it has not updated anything, I've removed that `return`. The `has_failed` is now extended with all required side effect checks (e.g. that the parachain head has been actually updated to bundled) and it is done inside "internal" extensions. Added some tests to make sure that we call "internal" `post_dispatch` even if `has_failed` is `false`.

bridges/bin/runtime-common/src/extensions/refund_relayer_extension.rs
```diff
@@ -66,7 +65,7 @@ type CallOf<R> = <R as frame_system::Config>::RuntimeCall;
 /// coming from this parachain.
 pub trait RefundableParachainId {
 	/// The instance of the bridge parachains pallet.
-	type Instance;
+	type Instance: 'static;
```
svyatonik, line 68: In the original PR, this trait and its implementation were moved to the `lib.rs` of this crate (https://github.com/paritytech/parity-bridges-common/pull/2873/files#diff-ec715be7be40d34a338721f0743eaeecf138cd3b007fb75a4a622ec605c63b3cR41), but it makes no sense - I've reverted this move

svyatonik:
All commits above are reviewed in https://github.com/paritytech/parity-bridges-common/pull/2873, modulo two comments I've left above. All commits below will need a proper review. Hopefully, all changes will be localized in bridge hub runtimes

prdoc/pr_4102.prdoc
```diff
@@ -0,0 +1,43 @@
+title: "Bridge: make some headers submissions free"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Adds `FreeHeadersInterval` configuration constant to the `pallet_bridge_grandpa`.
+      Transactions that improve best known header by at least `FreeHeadersInterval` headers
+      are now free for the submitter. Additionally, we allow single free parachain header
+      update per every free relay chain header. Bridge signed extensions are adjusted
```
acatangiu, line 9: I think a malicious relayer can do high-tip free txs that advance the view of the bridged relay chain by `FreeHeadersInterval`, but only advance some parachain view by 1 block, effectively DoS-ing the free update of the parachain.
Or the updated parachain head have to match the `para_heads` of the imported relay block proof?

Even so, attacker(s) would be spending money (the tx tips) for keeping the attack ongoing, while any honest relayer can still update the parachain view (but not for free).

acatangiu (approved):
really good test coverage too! 💪

serban300 (approved):
Looks good !

prdoc/pr_4102.prdoc
svyatonik: > Or the updated parachain head have to match the para_heads of the imported relay block proof?

Maybe I'b bad in describing behavior. Parachain head submission is free (`Pays::No`) iff:
1) best parachain head, known to the pallet is `PH_old` and it is proved at relay chain header `RH_old`;
2) new parachain head `PH_new` is proved at relay chain head `RH_new`;
3) the `RH_new - RH_old` is larger than the `FreeHeadersInterval`.

So no, attacker will be able to increase parachain head number by one for free only if it has actually been increased by one during `FreeHeadersInterval` relay chain headers.

We also have this new signed extension that boosts priority of relay chain header/parachain head submission transactions based on that `RH_new - RH_old` expression. So transaction of registered relayer that submits free header `1000` will have larger priority than transaction of registered relayer that submits free header `900`. So if there's a honest relayer, which tries to submit best available rela/para header `1000`, it will be costly for attacker to submit headers `900`, `910`, `920`, ... for free (assuming `FreeHeadersInterval = 10`) - honest collators will include transaction with `1000`  before other transactions

bridges/primitives/runtime/src/chain.rs
```diff
@@ -236,6 +236,12 @@ where
 pub trait Parachain: Chain {
 	/// Parachain identifier.
 	const PARACHAIN_ID: u32;
+	/// Maximal size of the parachain header.
+	///
+	/// This isn't a strict limit. The relayer may submit larger headers and the
+	/// pallet will accept the call. The limit is only used to compute whether
+	/// the refund can be made.
+	const MAX_HEADER_SIZE: u32;
```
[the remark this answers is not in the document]
svyatonik: Exactly - it may mean multiple things. But you'll probably be computing it in a way I've described in the referenced comment (either theoretical limit w.r.t. current code  OR experimental with some reserve). We may later use it for some other purposes as well. The fact that it is a soft limit for some other code doesn't affect the way it is computed. I can remove this comment if it confuses you and leave it where this "soft" criteria is actually applied, but it won't change the logic or naming or something else.

.gitlab/pipeline/zombienet.yml
```diff
@@ -12,4 +12,4 @@ include:
   # polkadot tests
   - .gitlab/pipeline/zombienet/polkadot.yml
   # bridges tests
-  - .gitlab/pipeline/zombienet/bridges.yml
+  # - .gitlab/pipeline/zombienet/bridges.yml
```
bkontur, line 15: this is temporary?

bridges/bin/runtime-common/src/extensions/check_obsolete_extension.rs
```diff
@@ -18,55 +18,229 @@
...
-	fn validate(call: &Call) -> TransactionValidity;
+pub trait BridgeRuntimeFilterCall<AccountId, Call> {
+	/// Data that may be passed from the validate to `post_dispatch`.
+	type ToPostDispatch;
+	/// Called during validation. Needs to checks whether a runtime call, submitted
+	/// by the `who` is valid. `who` may be `None` if transaction is not signed
+	/// by a regular account.
+	fn validate(who: &AccountId, call: &Call) -> (Self::ToPostDispatch, TransactionValidity);
+	/// Called after transaction is dispatched.
+	fn post_dispatch(_who: &AccountId, _has_failed: bool, _to_post_dispatch: Self::ToPostDispatch) {
```
bkontur, line 51: default impl "doing nothing" is intentional?

.gitlab/pipeline/zombienet.yml
svyatonik: Yes, let me leave a comment. Actually, we **may** do a hack here and release relayer that uses that branch earlier than this PR is merged. But I feel that we may lost track of what relayer and where it is used. So IMO better to comment and uncommet soon after (I hope minutes/hours after) this PR and #4157 will be merged

svyatonik: https://github.com/paritytech/polkadot-sdk/pull/4102/commits/18e15fb42e92fa4d1b92fe32cda567e8fc5eadf8

bridges/bin/runtime-common/src/extensions/check_obsolete_extension.rs
svyatonik: Yes - `post_dispatch` is not used by default pallet-level implementations for pallets (see `impl BridgeRuntimeFilterCall for pallet_bridge_*`)

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-westend/tests/tests.rs
```diff
@@ -271,22 +270,6 @@ fn relayed_incoming_message_works() {
 	)
 }
 
-#[test]
-pub fn complex_relay_extrinsic_works() {
-	from_parachain::complex_relay_extrinsic_works::<RuntimeTestsAdapter>(
```
bkontur, line 276: so the complex relayer won't work with older `submit_finality_proof` even if it pays ?

bkontur (approved):

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-westend/tests/tests.rs
svyatonik: It should work, but his transactions won't be refunded + we do not check for obsolete stuff inside batches anymore. This tests checks that after complex transaction, relayer receives a refund, so it fails. There is a test somewhere nearby for a standalone transactions

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/Cargo.toml
```diff
@@ -22,6 +22,7 @@ scale-info = { version = "2.11.1", default-features = false, features = [
 	"derive",
 ] }
 serde = { optional = true, features = ["derive"], workspace = true, default-features = true }
+tuplex = { version = "0.1", default-features = false }
```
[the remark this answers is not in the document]
bkontur: aha, ok, I am not sure if it is possible to hide/re-export `tuplex` inside that macro, I think we did something similar for other macros for paste?
```
// Re-export macro to avoid include paste dependency everywhere
pub use sp_runtime::paste;
```
but also ok to leave it as it is

cumulus/parachains/runtimes/bridge-hubs/bridge-hub-westend/tests/tests.rs
svyatonik: @bkontur I've added tests in this commit: https://github.com/paritytech/polkadot-sdk/pull/4102/commits/cde347f7eb86eee4bab73638a4fc5133226394f0 PTAL. Thank you!

bkontur: > @bkontur I've added tests in this commit: [cde347f](https://github.com/paritytech/polkadot-sdk/commit/cde347f7eb86eee4bab73638a4fc5133226394f0) PTAL. Thank you!

ok, checked, cool, thank you :)

bridges/bin/runtime-common/src/extensions/check_obsolete_extension.rs
svyatonik: Yes, it makes sense. Actually, I'd prefer to rename it along with those 'obsolete' extensions, which are not just about 'obsolete' anymore. Maybe `PreDispatchFilter` would be better.

However, it is the large refactoring and it is better to do that in a separate PR, which will contain no logic changes. Serban has already suggested to do that [here](https://github.com/paritytech/parity-bridges-common/pull/2873#discussion_r1557703246).

I'll file an issue.

svyatonik:
The plan is to deploy it on Rococo <> Westend before the next release. So I've prepared a branch: https://github.com/paritytech/polkadot-sdk/tree/sv-new-bridges-burning. It is forked from last release and all required commits are cherry-picked:
```sh
    git checkout polkadot-v1.11.0
    git checkout -b sv-new-bridges-burning
    git cherry-pick a633e954f3b88697aa797d9792e8a5b5cf310b7e
    git cherry-pick 7e68b2b8da9caf634ff4f6c6d96d2d7914c44fb7
    git cherry-pick a99948939785eb62377a235b448a7f4ae65c01b1
    git cherry-pick 871281783c1be03157319d5143096fd3dd860d0a
    + bump Rococo and Westend BH versions to 1_011_001
```

I'll build Rococo and Westend BH runtimes on monday and we'll do a upgrade. Then we need to migrate all our relayers to our new scheme - guess it can take a couple of days. After that, we need to update our dashboards and alerts (it should be easy).

bridges/bin/runtime-common/src/extensions/check_obsolete_extension.rs
```diff
@@ -101,29 +283,72 @@ macro_rules! generate_bridge_reject_obsolete_headers_and_messages {
...
 			fn pre_dispatch(
 				self,
-				who: &Self::AccountId,
+				relayer: &Self::AccountId,
 				call: &Self::Call,
 				info: &sp_runtime::traits::DispatchInfoOf<Self::Call>,
 				len: usize,
 			) -> Result<Self::Pre, sp_runtime::transaction_validity::TransactionValidityError> {
-				self.validate(who, call, info, len).map(drop)
+				use tuplex::PushBack;
```
ggwpez, line 316: Macros should always export their own dependencies, otherwise it wont build when the dependency is not present or when its a wrong version.

bkontur: @ggwpez fixed here: https://github.com/paritytech/polkadot-sdk/pull/4774

PR #4112, opened 2024-04-12
rococo_contracts: Adds missing migration

bkchr:


pgherveou (approved):

liamaharon (approved):
Did CI not catch this? Maybe it should be mandatory to pass now, since it seems stable (except westend until that is fixed)

pgherveou:
To catch it you would need to dry run the migration with try-runtime as the list of migration that you need to setup here depend of what has already been deployed (v16 implied you are already on v15)

bkchr:
> Did CI not catch this? Maybe it should be mandatory to pass, since it has been stable for a while now (except westend until that is fixed)

CI catched this, the test was failing all the time

PR #4118, opened 2024-04-15
pallet assets: Fix errors

thiolliere:
`LiveAsset` is an error to be returned when an asset is not supposed to be live.
And `AssetNotLive` is an error to be returned when an asset is supposed to be live, I don't think frozen qualifies as live.

### DIFF
--- a/prdoc/pr_4118.prdoc
+++ b/prdoc/pr_4118.prdoc
@@ -1,4 +1,4 @@
-title: pallet assets: minor improvement on errors returned for some calls
+title: "pallet assets: minor improvement on errors returned for some calls"
 
 doc:
   - audience: Runtime Dev
--- a/prdoc/pr_4118.prdoc
+++ b/prdoc/pr_4118.prdoc
@@ -4,6 +4,9 @@ doc:
   - audience: Runtime Dev
     description: |
       Some calls in pallet assets have better errors. No new error is introduced, only more sensible choice are made.
+  - audience: Runtime User
+    description: |
+      Some calls in pallet assets have better errors. No new error is introduced, only more sensible choice are made.
 
 crates:
   - name: pallet-assets
--- a/prdoc/pr_4118.prdoc
+++ b/prdoc/pr_4118.prdoc
@@ -1,0 +1,10 @@
+title: pallet assets: minor improvement on errors returned for some calls
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Some calls in pallet assets have better errors. No new error is introduced, only more sensible choice are made.
+
+crates:
+  - name: pallet-assets
+    bump: minor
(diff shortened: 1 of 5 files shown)


joepetrowski (approved):

substrate/frame/assets/src/functions.rs
```diff
@@ -491,7 +491,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		let d = Asset::<T, I>::get(&id).ok_or(Error::<T, I>::Unknown)?;
 		ensure!(
 			d.status == AssetStatus::Live || d.status == AssetStatus::Frozen,
-			Error::<T, I>::AssetNotLive
+			Error::<T, I>::IncorrectStatus
```
gupnik, line 494: Can't we say `AssetNotLiveOrFrozen`? Not sure if `IncorrectStatus` is much helpful.

liamaharon: I'm personally fine with `IncorrectStatus`. It's non-ambiguous since it's only used in one place, and allows for us to add new statuses in the future without needing to update the error.

liamaharon (approved):

liamaharon:
@thiolliere could you please add a prdoc?

substrate/frame/assets/src/functions.rs
thiolliere: `AssetNotLiveNorFrozen` or `AssetNotLiveOrFrozen` ?
I will do.

gupnik: >  allows for us to add new statuses in the future without needing to update the error.

Yeah, it helps us, but not sure someone using a UI would get any info with the `IncorrectStatus` error. Either ways, its fine but I would prefer the errors to be more meaningful for the users.

liamaharon: I don't think Rust enum errors are supposed to be shown directly to end-users in UIs? I always understood them as more for developers.

thiolliere: Some UI might show the doc: `/// The asset status is not the expected status.`

Considering there is 3 status I think it is ok

thiolliere:
I updated the minor version of pallet assets, I don't know if it is wanted.

prdoc/pr_4118.prdoc
```diff
@@ -0,0 +1,13 @@
+title: "pallet assets: minor improvement on errors returned for some calls"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      Some calls in pallet assets have better errors. No new error is introduced, only more sensible choice are made.
+  - audience: Runtime User
+    description: |
+      Some calls in pallet assets have better errors. No new error is introduced, only more sensible choice are made.
```
thiolliere, line 9: I feel information is not useful, but in theory it impacts both dev for maybe test to udpate and frontend.

thiolliere: in theory it is kind of a major breaking change :-(

PR #4119, opened 2024-04-15
Improvements in minimal template

gupnik:
This PR makes a few improvements in the docs for the minimal template.


substrate/frame/balances/src/lib.rs
```diff
@@ -232,6 +232,17 @@ pub mod pallet {
...
 			type WeightInfo = ();
 		}
+
+		/// Default configurations of this pallet in a solochain environment.
+		pub struct SolochainDefaultConfig;
+
+		#[derive_impl(frame_system::config_preludes::SolochainDefaultConfig, no_aggregated_types)]
+		impl frame_system::DefaultConfig for SolochainDefaultConfig {}
+
+		/// It currently uses the same configuration as `TestDefaultConfig`.
```
kianenigma, line 242:
```suggestion
		/// It currently uses the same configuration as [`TestDefaultConfig`].
```

substrate/frame/sudo/src/lib.rs
```diff
@@ -167,6 +167,17 @@ pub mod pallet {
...
 			type RuntimeCall = ();
 		}
+
+		/// Default configurations of this pallet in a solochain environment.
+		pub struct SolochainDefaultConfig;
+
+		#[derive_impl(frame_system::config_preludes::SolochainDefaultConfig, no_aggregated_types)]
+		impl frame_system::DefaultConfig for SolochainDefaultConfig {}
+
+		/// It currently uses the same configuration as `TestDefaultConfig`.
```
kianenigma, line 177:
```suggestion
		/// It currently uses the same configuration as [`TestDefaultConfig`].
```

substrate/frame/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -541,6 +542,7 @@ pub(crate) fn decl_all_pallets<'a>(
 			attrs.extend(TokenStream2::from_str(&feat).expect("was parsed successfully; qed"));
 		}
 		let type_decl = quote!(
+			#( #[doc = #docs] )*
```
kianenigma, line 545: nice touch

substrate/frame/timestamp/src/lib.rs
```diff
@@ -164,6 +164,17 @@ pub mod pallet {
...
 			type WeightInfo = ();
 		}
+
+		/// Default configurations of this pallet in a solochain environment.
+		pub struct SolochainDefaultConfig;
+
+		#[derive_impl(frame_system::config_preludes::SolochainDefaultConfig, no_aggregated_types)]
+		impl frame_system::DefaultConfig for SolochainDefaultConfig {}
+
+		/// It currently uses the same configuration as `TestDefaultConfig`.
```
kianenigma, line 174:
```suggestion
		/// It currently uses the same configuration as [`TestDefaultConfig`].
```

substrate/frame/transaction-payment/src/lib.rs
```diff
@@ -336,6 +336,17 @@ pub mod pallet {
...
 			type OperationalFeeMultiplier = ();
 		}
+
+		/// Default configurations of this pallet in a solochain environment.
+		pub struct SolochainDefaultConfig;
+
+		#[derive_impl(frame_system::config_preludes::SolochainDefaultConfig, no_aggregated_types)]
+		impl frame_system::DefaultConfig for SolochainDefaultConfig {}
+
+		/// It currently uses the same configuration as `TestDefaultConfig`.
```
kianenigma, line 346:
```suggestion
		/// It currently uses the same configuration as [`TestDefaultConfig`].
```

templates/minimal/runtime/src/lib.rs
```diff
@@ -54,61 +58,107 @@ pub fn native_version() -> NativeVersion {
...
 parameter_types! {
 	pub const Version: RuntimeVersion = VERSION;
 }
 
+/// Implements the types required for the system pallet.
 #[derive_impl(frame_system::config_preludes::SolochainDefaultConfig)]
 impl frame_system::Config for Runtime {
 	type Block = Block;
 	type Version = Version;
 	type BlockHashCount = ConstU32<1024>;
```
kianenigma, line 134:
```suggestion
	type BlockHashCount = ConstU32<1024>;
```
```suggestion
```

templates/minimal/runtime/src/lib.rs
```diff
@@ -54,61 +58,107 @@ pub fn native_version() -> NativeVersion {
...
+#[derive_impl(pallet_timestamp::config_preludes::SolochainDefaultConfig)]
 impl pallet_timestamp::Config for Runtime {}
 
-#[derive_impl(pallet_transaction_payment::config_preludes::TestDefaultConfig)]
+// Implements the types required for the transaction payment pallet.
+#[derive_impl(pallet_transaction_payment::config_preludes::SolochainDefaultConfig)]
 impl pallet_transaction_payment::Config for Runtime {
 	type OnChargeTransaction = pallet_transaction_payment::FungibleAdapter<Balances, ()>;
 	type WeightToFee = NoFee<<Self as pallet_balances::Config>::Balance>;
 	type LengthToFee = FixedFee<1, <Self as pallet_balances::Config>::Balance>;
```
kianenigma, line 158: An explanation of this would be good.

kianenigma (approved):
Great start. I think last step you should run this, and see what details you need to provide to the user about how to run etc.

kianenigma (changes requested):
- https://github.com/paritytech/polkadot-sdk/pull/4119/files#r1565970574
 
Regarding running the node: 

- The block time being configurable
- Fixed transaction fees

templates/minimal/README.md
```diff
@@ -0,0 +1,13 @@
...
+This is a minimal template for creating a blockchain using the Polkadot SDK.
+
+# Docs
+
+You can generate and view the [Rust
+Docs](https://doc.rust-lang.org/cargo/commands/cargo-doc.html) for this template
+with this command:
+
+```sh
+cargo +nightly doc -p minimal-template --open
```
kianenigma, line 12:
```suggestion
cargo doc -p minimal-template --open
```

Is it really needed?

kianenigma (approved):

templates/minimal/README.md
```diff
@@ -0,0 +1,13 @@
+# Minimal Template
```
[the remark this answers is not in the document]
gupnik: Thanks! Will take these up as a follow-up.

substrate/frame/balances/src/lib.rs
```diff
@@ -232,6 +232,17 @@ pub mod pallet {
...
+		/// Default configurations of this pallet in a solochain environment.
+		pub struct SolochainDefaultConfig;
+
+		#[derive_impl(frame_system::config_preludes::SolochainDefaultConfig, no_aggregated_types)]
+		impl frame_system::DefaultConfig for SolochainDefaultConfig {}
+
+		/// It currently uses the same configuration as [`TestDefaultConfig`].
+		#[derive_impl(TestDefaultConfig, no_aggregated_types)]
+		#[frame_support::register_default_impl(SolochainDefaultConfig)]
+		impl DefaultConfig for SolochainDefaultConfig {}
```
liamaharon, line 245: Why do pallets need to be concerned with the default config of the solochain?

Shouldn't config related logic for the solochain be in the solochain crate?

Where does this stop? Will we start adding default configs to every pallet for all our 4,5,6,7+ runtime variations?

gupnik: > Shouldn't config related logic for the solochain be in the solochain crate?

Which crate are you referring here? Is the expectation then for the runtime dev to import that crate as well? How does that reduce the friction?

> Where does this stop? Will we start adding default configs to every pallet for all our 4,5,6,7+ runtime variations?

I would imagine that we provide a few (Test, Solo, Para and Relay) and let the developers build out the rest as needed for custom use cases.

templates/minimal/README.md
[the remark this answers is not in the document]
liamaharon: Where will the readme.md be hosted? Could you remind me the distinction between deciding between whether to put content in the readme or rustdoc?

templates/minimal/runtime/src/lib.rs
```diff
@@ -54,61 +58,108 @@ pub fn native_version() -> NativeVersion {
 	NativeVersion { runtime_version: VERSION, can_author_with: Default::default() }
 }
 
+// The signed extensions that are added to the runtime.
```
liamaharon, line 61:
```suggestion
/// The signed extensions that are added to the runtime.
```

substrate/frame/balances/src/lib.rs
liamaharon: My initial reaction to this is I really don't like it. 

1. I don't think pallet developers should be concerned about needing to write a default config for every runtime variation the pallet could be used in, the entire point of pallets exposing a config is so the customisation happens in the **runtime**?

2. This also encourages lazy, or at best 'blind' runtime development. Critical configuration of your runtime is decided for you in a crates.io package you'll probably never look at. Even if you do check the downstream package, what if a new default config item is added to the downstream package in an upgrade that doesn't suit your runtime? It will be added without you even knowing. 

I'm all for default configs in the context of testing and mocking, but I think for production systems with value on the line it is much better for runtime developers to explicitly specify their configs.

templates/minimal/runtime/src/lib.rs
gupnik: Not a public type

gupnik:
> Where will the readme.md be hosted? Could you remind me the distinction between deciding between whether to put content in the readme or rustdoc?

My preference would be to move most of it to rustdoc and keep readme as a light intro.

substrate/frame/balances/src/lib.rs
gupnik: I think this is a longer discussion outside the scope of this PR. Could you open a separate issue to take that further?

templates/minimal/README.md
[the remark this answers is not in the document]
gupnik: > Where will the readme.md be hosted? Could you remind me the distinction between deciding between whether to put content in the readme or rustdoc?

My preference would be to move most of it to rustdoc and keep readme as a light intro.

substrate/frame/balances/src/lib.rs
liamaharon: OK. Before merging this PR I don't think the production chain config should extend the test config though.

Changing the test config seems like it would be an innocent change, I wouldn't expect that to impact production chains.

franciscoaguirre: I agree inheriting from the testconfig could be very bad.
Sane defaults (as restrictive as possible) for not having to always type out unreasonably large config impls is what we need. We don't want only mock runtimes to benefit from this.
If we agree, I would only have a `TestDefaultConfig` and a `RestrictiveDefaultConfig` or something like that.

gupnik: This was done based on this comment: https://github.com/paritytech/polkadot-sdk/pull/3946#discussion_r1548062675

Since this is not the core focus of this PR, I will revert it so that this can move ahead. We can continue this discussion on the original issue.

kianenigma:
> I don't think pallet developers should be concerned about needing to write a default config for every runtime variation the pallet could be used in, the point of pallets exposing a config is so the customisation happens in the runtime?

Is a correct comment in general. I foresee that a pallet can expose _optionally, at most_ 3 configs: 

1. Test, non-important and should be permissive to allow things in test
2. Solochain config, restrictive. 
3. Parachain config, mostly same as Solochain, but with higher emphasis on weights or other concepts that differ between parachain and Solochains. 

And anything more can live in independent crate, possibly even outside `polkadot-sdk`. A crate that uses `pallet-x` as a dependency and exposes defaults for `pallet-x`. Imagine many variants of `pallet-balances-default-config`, provided by different teams with varying degrees of opinionation. 

I've also raised that a team (eg. OZ) can instead of building a full runtime with 10 pallets, simply provide a set of defaults for those 10 pallets. 

> This also encourages lazy, or blind runtime development. Critical configuration of your runtime is decided for you in a crates.io package you'll probably only skim read. Even if you do check it thoroughly, what if a new default config item is added to the downstream package in an upgrade that doesn't suit your runtime? There is a risk that decisions will be made for you without you knowing.

Indeed any default configs added to `polkadot-sdk` should be carefully reviewed for safety. If any doubt remains, we should not provide a default. 

A downstream team can always opt out of using `derive_impl` if they feel more comfortable not using arbitrary defaults from upsteam. 

Possibly, we can even provide a default type that does nothing and raises a `compiler_error!` such that a donwstream teams, even if they opted into `derive_impl` _have to provide their own_.

substrate/frame/balances/src/lib.rs
gupnik: This is done. Thanks.

templates/minimal/runtime/src/lib.rs
gupnik: Added

gupnik:
@liamaharon @franciscoaguirre @ggwpez Could you take a look at this again please?

liamaharon (approved):

substrate/frame/support/procedural/src/pallet/expand/tt_default_parts.rs
```diff
@@ -198,9 +198,9 @@ pub fn expand_tt_default_parts(def: &mut Def) -> proc_macro2::TokenStream {
 		macro_rules! #default_parts_unique_id_v2 {
 			{
 				$caller:tt
-				frame_support = [{ $($frame_support:ident)::* }]
+				your_tt_return = [{ $my_tt_return:path }]
```
ggwpez, line 201: I actually had the same issue on the polkadot-sdk umbrella crate and ended up using `::tt *` since path seems to also not be enough.

PR #4122, opened 2024-04-15
FRAME: Meta Transaction (pallet based version)

muharem:
Meta transaction implementation based on a pallet's call and configurable via transaction extensions.

To see an example, refer to the `sign_and_execute_meta_tx` test case in `substrate/frame/meta-tx/src/tests.rs` file.

This implementation serves to demonstrate the concept and not production ready.

Based on: https://github.com/paritytech/polkadot-sdk/pull/3685
RFC: https://github.com/paritytech/polkadot-sdk/issues/4123
Alternative solution: https://github.com/paritytech/polkadot-sdk/pull/3712

TODOs:
- benchmarks
- more tests


substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+				return Err(Error::<T>::TODO.into());
+			}
+
+			let origin = T::RuntimeOrigin::signed(signer);
+			let (call, tx_ext, _) = signed_payload.deconstruct();
+			let info = call.get_dispatch_info();
+			// TODO: to get the len we have to encode the original `meta_tx`.
+			let len = 0;
+			let mut ctx = T::TxContext::default();
+
```
muharem, line 154: we can use set the `DispatchTransaction` bound for `Config::TxExtension` and use it's impl for everything below. here, for the demonstration purpose, I'm not utilizing it.

substrate/frame/meta-tx/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+[package]
```
kianenigma, line 1: This whole crate is great, and should live in `frame/examples` if it remains in any form.

substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+			+ GetDispatchInfo
+			+ Dispatchable<
+				Info = DispatchInfo,
+				PostInfo = PostDispatchInfo,
+				RuntimeOrigin = Self::RuntimeOrigin,
+			> + IsType<<Self as frame_system::Config>::RuntimeCall>;
+		/// Signature type for meta transactions.
+		type Signature: Parameter + Verify<Signer = Self::PublicKey>;
+		/// Public key type used for signature verification.
+		type PublicKey: IdentifyAccount<AccountId = Self::AccountId>;
```
kianenigma, line 97: unrelated to this PR: Kind of a pitty that we can't fetch `PublicKey` and `Signature` from `AccountId`. The reverse is possible. You can always go from `Signature` (eg. `MultiSignature`) to `PublicKey` and then `AccountId` via `IdentifyAccount`.

muharem: in this scope I do not even need a public key type, but have no other way to ensure the signature type implemented over the account id the system operates with.

substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+};
+use frame_system::pallet_prelude::*;
+use sp_runtime::traits::TransactionExtension;
+use sp_runtime::traits::TransactionExtensionBase;
+use sp_runtime::traits::{Dispatchable, IdentifyAccount, Verify};
+
+/// Meta transaction type.
+// TODO: The `MetaTx` type is similar to `sp_runtime::generic::UncheckedExtrinsic`. However,
+// `MetaTx` cannot replace generic::UncheckedExtrinsic because we need to box the call type,
+// given that `MetaTx` is used as an argument type for a call.
```
georgepisaltu, line 43: We could probably have this "destructured" in the extrinsic argument list if it looks funny because of the `UncheckedExtrinsic` similarity. I think it's fine.

substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+	) -> Self {
+		Self { proof: Proof::Signed(address, signature), call: Box::new(call), tx_ext }
+	}
+}
+
+#[derive(Encode, Decode, PartialEq, Eq, TypeInfo, Clone, RuntimeDebug)]
+pub enum Proof<Address, Signature> {
+	Signed(Address, Signature),
+	// TODO `General` as in `sp_runtime::generic::Preamble`.
+}
```
georgepisaltu, line 66: I think we can branch off the `UncheckedExtrinsic` and `Preamble` logic a bit here and consider introducing a variant for any type of custom validation logic supported by the pallet. In `TransactionExtension`s, they would be individual extensions in the pipeline, one for each validation type, each wrapping a `General` transaction.

Here, we care only about the call and validation type, which is nice.

substrate/frame/meta-tx/src/lib.rs
georgepisaltu: `DispatchTransaction` is meant to work only for "legacy" extensions which use a `()` context. I believe the current impl is the right thing :+1: .

substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+			.map_err(|_| Error::<T>::TODO)?;
+
+			let pre = T::TxExtension::prepare(tx_ext, val, &origin, &call, &info, len, &ctx)
+				.map_err(|_| Error::<T>::TODO)?;
+
+			let res = call.dispatch(origin);
+			let post_info = res.unwrap_or_else(|err| err.post_info);
+			let pd_res = res.map(|_| ()).map_err(|e| e.error);
+
+			T::TxExtension::post_dispatch(pre, &info, &post_info, len, &pd_res, &ctx)
```
georgepisaltu, line 174: This would allow a graceful and harmless failure in cases like the one described [here](https://github.com/paritytech/polkadot-sdk/pull/4165#discussion_r1568226192) when a [`post_dispatch` fail is critical](https://github.com/paritytech/polkadot-sdk/blob/ba0f8de0c74b1308a53942c406137a748ef79925/substrate/primitives/runtime/src/traits.rs#L1576).

substrate/frame/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+
+/// Meta transaction type.
+// TODO: The `MetaTx` type is similar to `sp_runtime::generic::UncheckedExtrinsic`. However,
+// `MetaTx` cannot replace generic::UncheckedExtrinsic because we need to box the call type,
+// given that `MetaTx` is used as an argument type for a call.
+#[derive(Encode, Decode, PartialEq, Eq, TypeInfo, Clone, RuntimeDebug)]
+pub struct MetaTx<Address, Signature, Call, TxExtension> {
+	proof: Proof<Address, Signature>,
+	call: Box<Call>,
+	tx_ext: TxExtension,
```
georgepisaltu, line 48: Is there any reason why we can't get away with just a `CheckNonce` here? The other stuff will be checked when the agent submits the tx anyway, we just want the replay protection for Alice.

If we want to move off the extension approach completely, we could do the nonce check and `inc` without the extension even, just run the `frame_system::Account` mutation inside the extrinsic, but the signed payload would need to include the nonce. Maybe it's too complicated, just a thought.

georgepisaltu (commented):
Looks great :+1:

muharem:
Clippy CI check fails on the file that is not edited by this PR. I will wait for the https://github.com/paritytech/polkadot-sdk/pull/3685 to be updated from master, to update this branch, which should address the issue. 

Currently version of this branch compiles, tests pass.

jasl:
Kindly ping.
Since Token Extension has merged, can it push Meta Transaction forward?

muharem:
@jasl I am working on it

muharem:
in favor of https://github.com/paritytech/polkadot-sdk/pull/6428

Issue #4128, opened 2024-04-15
Migration from v3::Location to v4::Location for ForeignAssets


franciscoaguirre:
ForeignAssets in the asset hubs use XCM Locations.
They don't use VersionedLocation, they are fixed to v3::Location.
When we introduced XCMv4, we left the assets using v3::Location instead of moving them to v4::Location, since they needed a migration.
We should migrate them.

acatangiu:
To be clear, migrate to `VersionedLocation`

PR #4129, opened 2024-04-15
Migrate foreign assets v3::Location to v4::Location

franciscoaguirre:
In the move from XCMv3 to XCMv4, the `AssetId` for `ForeignAssets` in `asset-hub-rococo` and `asset-hub-westend` was left as `v3::Location` to be later migrated to `v4::Location`.

This is that migration PR.

Because the encoding of `v3::Location` and `v4::Location` is the same, we don't need to do any data migration, the keys will still be decodable.
The [original idea by Jan](https://github.com/paritytech/polkadot/pull/7236) was to make the v4 changes in v3 since the ABI (the encoding/decoding) didn't change.
Corroborated the ABI is the same iterating over all storage, the code is on [another branch](https://github.com/paritytech/polkadot-sdk/blob/cisco-assert-v3-v4-encodings-equal/cumulus/parachains/runtimes/assets/migrations/src/foreign_assets_to_v4/mod.rs).

We will need a data migration when we want to update from `v4::Location` to `v5::Location` because of [the accepted RFC changing the NetworkId enum](https://github.com/polkadot-fellows/RFCs/pull/108).
I'll configure MBMs (Multi-Block Migrations) then and make the actual migration.

Fixes https://github.com/paritytech/polkadot-sdk/issues/4128


franciscoaguirre:
> Can we migrate to `VersionedLocation` instead?

We can. The only issue with that is we might have two versions of the same asset.
Say we do migrate to `VersionedLocation`, then we could have both `VersionedLocation::V3(usdt)` and `VersionedLocation::V4(usdt)` for example.

We'd still need a migration to change values to the newest version. It's either that or migrating to change types to the newest version. I say fixing the version in this case doesn't sound like a bad idea. What do you think?

acatangiu:
I was thinking to keep assets with mixed versions in storage, but you're right that it won't work (or rather it could work but with many corner cases and ultimately more complexity).

I agree we keep it simple and migrate assets to latest version.

franciscoaguirre:
> I was thinking to keep assets with mixed versions in storage, but you're right that it won't work (or rather it could work but with many corner cases and ultimately more complexity).
> 
> I agree we keep it simple and migrate assets to latest version.

Rethinking this, it might be better to migrate them to VersionedAssets. The only way to create foreign assets is via XCM and the sender doesn't specify the asset id. We can just use V4 all the time and change it when we migrate. This means we'll have more leeway whenever a new XCM version comes out. What do you think?

acatangiu:
@xlc I remember seeing another thread on this topic where you had some input, but can't find it now.
(I might also be imagining things 😆 , but tagging you here just in case)

bkontur:
> @xlc I remember seeing another thread on this topic where you had some input, but can't find it now. (I might also be imagining things 😆 , but tagging you here just in case)

This is the original issue for this foreign asset id problem: https://github.com/paritytech/polkadot-sdk/issues/1130,

> Rethinking this, it might be better to migrate them to VersionedAssets. The only way to create foreign assets is via XCM and the sender doesn't specify the asset id. We can just use V4 all the time and change it when we migrate. This means we'll have more leeway whenever a new XCM version comes out. What do you think?

I was proposing there something like this - to use `VersionedLocation` for `Transact(ForeignAssets::create(asset_id as VersionedLocation, ...)`:
```
        type AssetId = xcm::v3::Location;                    // requires storage migration
	type AssetIdParameter = VersionedLocation;           // this is used for extrinsics
```
which would solve `xcm::Transact` versioning problem from other parachains,
but another part of the problem is how the dapps will read `ForeignAssets` with `v3/v4::Location`:
- if they read directly from storage, they could have a problem when we migrate to new version
- if they use new dedicated runtime API (with input parameter as desired XcmVersion - as we do for `query_acceptable_payment_assets(xcm_version: xcm::Version)`), where we can return `VersionedLocation` -> no problem
- something else?

If we use `VersionedLocation` for AssetId:
```
        type AssetId = VersionedLocation
```
I don't know, it could lead maybe to other problems, at least we would need to use something like for `pallet_xcm` with [`LatestVersionedLocation`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/src/lib.rs#L614)

franciscoaguirre:
I was initially convinced by the argument that if we use `VersionedLocation`, we could have two versions of the same asset. However, now I'm not so sure. Since we can only create foreign assets via an XCM and the runtime assigns the id, we can always assign the latest version. When we upgrade, we bump the version and migrate all ids to `VersionedLocation::V{n + 1}`.

xlc:
> @xlc I remember seeing another thread on this topic where you had some input, but can't find it now. (I might also be imagining things 😆 , but tagging you here just in case)

Yeah I made a comment somewhere but also can't find it anymore...
But basically there isn't much reason to migrate from v3 to v4 as they are exactly the same thing except version number.
Yeah there is no harm to do so but also no benefit and less work is better than more work.
Also I always use a fixed XCM version for storage item and Versioned for extrinsic (this applies to Acala and ORML).
This means it is less likely to have unexpected storage item format change as a fixed version type should never change but Versioned type is a moving target. e.g. when XCM V2 is removed, runtime using V2 type will not be able to compile but runtime using Versioned type that is V2 will get bad decoding error. It also means there could be multiple representation of a same thing and that can makes consumer of the storage item harder (e.g. UI, block explorer). It could be a critical issue when such thing is used as storage key.

cumulus/parachains/runtimes/assets/migrations/src/v1/benchmarks.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn step() {
+        let key = v3::Location::new(1, [v3::Junction::Parachain(2004)]);
+        let mock_asset_details = mock_asset_details();
+        dbg!(&std::any::type_name::<<T as pallet_assets::Config>::AssetId>());
+        // old::Asset::<T, ()>::insert(key.clone(), mock_asset_details);
+
+        let mut meter = WeightMeter::new();
+        #[block]
+        {
+            Migration::<T, ()>::step(None, &mut meter).unwrap();
```
franciscoaguirre, line 40: I get an error here saying the trait bounds are not satisfied. It says `<T as pallet_assets::Config>::AssetId` needs to be `xcm::v4::Location`. But it is. Don't know what's wrong.

cumulus/parachains/runtimes/assets/migrations/src/v1/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+		AssetDetails<
+			<T as Config<I>>::Balance,
+			<T as frame_system::Config>::AccountId,
+			DepositBalanceOf<T, I>,
+		>,
+	>;
+}
+
+pub struct Migration<T: Config<I>, I: 'static = ()>(PhantomData<(T, I)>);
+impl<T: Config<I, AssetId = v4::Location>, I: 'static> SteppedMigration
```
franciscoaguirre, line 62: This is the trait bound

cumulus/parachains/runtimes/assets/migrations/src/v1/tests.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[derive_impl(pallet_balances::config_preludes::TestDefaultConfig)]
+impl pallet_balances::Config for Runtime {
+    type AccountStore = System;
+    type Balance = Balance;
+}
+
+#[derive_impl(pallet_assets::config_preludes::TestDefaultConfig)]
+impl pallet_assets::Config for Runtime {
+    type AssetId = v4::Location;
```
franciscoaguirre, line 62: `AssetId = v4::Location`

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1094,7 +1127,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	Migrations,
+	(),
```
ggwpez, line 1130: Maybe also mention in the MR description that you now use the new convention of specifying the migrations in `frame-system`.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -175,8 +179,36 @@ impl frame_system::Config for Runtime {
...
+	#[cfg(not(feature = "runtime-benchmarks"))]
+	type Migrations = MultiBlockMigrations;
+	#[cfg(feature = "runtime-benchmarks")]
+	type Migrations = pallet_migrations::mock_helpers::MockedMigrations;
+	type CursorMaxLen = ConstU32<1024>;
+	type IdentifierMaxLen = ConstU32<256>;
+	type MigrationStatusHandler = ();
+	type FailedMigrationHandler = FreezeChainOnFailedMigration;
+	type MaxServiceWeight = MbmServiceWeight;
+	type WeightInfo = ();
```
ggwpez, line 202: Can probably set this once there is a weight file, but can be done afterwards.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -175,8 +179,36 @@ impl frame_system::Config for Runtime {
...
 }
 
+#[cfg(not(feature = "runtime-benchmarks"))]
+type MultiBlockMigrations = assets_common_migrations::v1::Migration<
```
ggwpez, line 206: I think migrations dont currently know about pallet versions, so you probably had to write the pallet version manually?  
We can make a wrapper for that.

cumulus/parachains/runtimes/assets/migrations/Cargo.toml
```diff
@@ -0,0 +1,80 @@
...
+
+# Cumulus
+parachains-common = { workspace = true }
+
+[dev-dependencies]
+sp-io = { workspace = true }
+hex-display = "0.3.0"
+
+[build-dependencies]
+substrate-wasm-builder = { workspace = true, default-features = true }
```
ggwpez, line 42: I think this is only needed for runtime crates.

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
franciscoaguirre: I just chose `v1` as a place to put this first migration. We could have more in the future. It's not tied to the version of any pallet. I could just give it a name instead of using `v1`.

cumulus/parachains/runtimes/assets/migrations/src/v1/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+			// If there is no cursor, start iterating from the beginning.
+			old::Asset::<T, I>::iter()
+		};
+
+		if let Some((key, value)) = iter.next() {
+			// Most likely all locations will be able to be converted, but if they can't
+			// we log them to try again later.
+			let maybe_new_key: Result<v4::Location, _> = key.try_into();
+			if let Ok(new_key) = maybe_new_key {
+				old::Asset::<T, I>::remove(&key);
```
ggwpez, line 113: I am not sure if it is safe to use an iterator `.next` *while* removing elements from the map at the same time or if it breaks the iterator...  
Could be safer to have a second step afterwards where it removes them, @cheme do you know?

cumulus/parachains/runtimes/assets/migrations/src/v1/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+pub struct Migration<T, I, W>(PhantomData<(T, I, W)>);
+impl<T: Config<I>, I: 'static, W: weights::WeightInfo> SteppedMigration for Migration<T, I, W>
+where
+	<T as Config<I>>::AssetId: From<v4::Location>,
+{
+	type Cursor = v3::Location;
+	type Identifier = MigrationId<13>;
+
+	fn id() -> Self::Identifier {
+		MigrationId { pallet_id: *PALLET_MIGRATIONS_ID, version_from: 0, version_to: 1 }
```
ggwpez, line 66: I think pallet-assets is already at version 1?  
We dont have try-runtime-cli support yet to check these things.

franciscoaguirre: Do these versions relate to the pallet version?

What if asset-hub-rococo and asset-hub-westend have different pallet versions? There wouldn't be a way to make a generic migration like this

ggwpez: Ah okay. I thought it would, yes then maybe change the name entirely to `pallet-assets-xcm-location` or something.  
Otherwise there could be another `pallet-assets` MBM. These version numbers here dont mean anything in that caes.

franciscoaguirre: I'd still like to keep this as its own pallet that holds a lot of migrations. For the future everytime we need to migrate from an older XCM version to a newer one

franciscoaguirre:
bot help

franciscoaguirre:
bot bench cumulus-assets --runtime asset-hub-westend --pallet assets_common_migrations

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1019,8 +1022,8 @@ pub type SignedExtra = (
 pub type UncheckedExtrinsic =
 	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
 /// Migrations to apply on runtime upgrade.
-#[allow(deprecated)]
 pub type Migrations = (
+	pallet_collator_selection::migration::v1::MigrateToV1<Runtime>,
```
acatangiu, line 1026: added by accident maybe?
```suggestion
```

acatangiu (approved):

prdoc/pr_4129.prdoc
```diff
@@ -0,0 +1,18 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Update ForeignAssets from xcm::v3::Location to xcm::v4::Location
+
+doc:
+  - audience: Runtime Dev
```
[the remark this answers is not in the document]
franciscoaguirre: Wasn't sure if that worked, thanks!

acatangiu:
needs a review from @paritytech/frame-coders

bkontur (approved):

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
```diff
@@ -1781,64 +1784,3 @@ cumulus_pallet_parachain_system::register_validate_block! {
 	Runtime = Runtime,
 	BlockExecutor = cumulus_pallet_aura_ext::BlockExecutor::<Runtime, Executive>,
 }
-
-#[cfg(test)]
-mod tests {
```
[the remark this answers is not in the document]
seadanda: Or even just a `tests.rs` file, instead of `tests/mod.rs`.

seadanda (approved):

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs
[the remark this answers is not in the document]
franciscoaguirre: Guess I ended up moving tests early in the PR for some reason. I'll put them back

serban300:
@franciscoaguirre I think this PR broke the `zombienet-bridges-0001-asset-transfer-works ` test.

Can you take a look please ?

cumulus/parachains/runtimes/assets/migrations/src/v1/mod.rs
cheme: I am late at the party, but code was written to support this (calling next on a state that change: either adding or removing content).

franciscoaguirre:
Sure, I'll take a look

bkontur:
> Sure, I'll take a look

hmm, strange, https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6991678
```
2024-08-13 14:37:42             VEC: Unable to decode on index 0 createType(Lookup37):: Struct: failed on interior: {"_enum":{"Here":"Null","X1":"[Lookup40;1]","X2":"[Lookup40;2]","X3":"[Lookup40;3]","X4":"[Lookup40;4]","X5":"[Lookup40;5]","X6":"[Lookup40;6]","X7":"[Lookup40;7]","X8":"[Lookup40;8]"}}:: Enum(X1):: decodeU8aVec: failed at 0x5b6f626a656374204f626a6563745d… (index 1/1): {"_enum":{"Parachain":"Compact<u32>","AccountId32":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"id\":\"[u8;32]\"}","AccountIndex64":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"index\":\"Compact<u64>\"}","AccountKey20":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"key\":\"[u8;20]\"}","PalletInstance":"u8","GeneralIndex":"Compact<u128>","GeneralKey":"{\"length\":\"u8\",\"data\":\"[u8;32]\"}","OnlyChild":"Null","Plurality":"{\"id\":\"XcmV3JunctionBodyId\",\"part\":\"XcmV3JunctionBodyPart\"}","GlobalConsensus":"StagingXcmV4JunctionNetworkId"}}:: Unable to create Enum via index 91, in Parachain, AccountId32, AccountIndex64, AccountKey20, PalletInstance, GeneralIndex, GeneralKey, OnlyChild, Plurality, GlobalConsensus
2024-08-13 14:37:42        RPC-CORE: queryStorageAt(keys: Vec<StorageKey>, at?: BlockHash): Vec<StorageChangeSet>:: createType(Vec<StorageKey>):: createType(Lookup37):: Struct: failed on interior: {"_enum":{"Here":"Null","X1":"[Lookup40;1]","X2":"[Lookup40;2]","X3":"[Lookup40;3]","X4":"[Lookup40;4]","X5":"[Lookup40;5]","X6":"[Lookup40;6]","X7":"[Lookup40;7]","X8":"[Lookup40;8]"}}:: Enum(X1):: decodeU8aVec: failed at 0x5b6f626a656374204f626a6563745d… (index 1/1): {"_enum":{"Parachain":"Compact<u32>","AccountId32":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"id\":\"[u8;32]\"}","AccountIndex64":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"index\":\"Compact<u64>\"}","AccountKey20":"{\"network\":\"Option<StagingXcmV4JunctionNetworkId>\",\"key\":\"[u8;20]\"}","PalletInstance":"u8","GeneralIndex":"Compact<u128>","GeneralKey":"{\"length\":\"u8\",\"data\":\"[u8;32]\"}","OnlyChild":"Null","Plurality":"{\"id\":\"XcmV3JunctionBodyId\",\"part\":\"XcmV3JunctionBodyPart\"}","GlobalConsensus":"StagingXcmV4JunctionNetworkId"}}:: Unable to create Enum via index 91, in Parachain, AccountId32, AccountIndex64, AccountKey20, PalletInstance, GeneralIndex, GeneralKey, OnlyChild, Plurality, GlobalConsensus
 Error running script: /home/nonroot/bridges-polkadot-sdk/bridges/testing/framework/js-helpers/wrapped-assets-balance.js
```

probably this lines: https://github.com/paritytech/polkadot-sdk/blob/master/bridges/testing/framework/js-helpers/wrapped-assets-balance.js#L10-L11:
```
const foreignAssetAccount = await api.query.foreignAssets.account(
            { parents: 2, interior: { X1: { GlobalConsensus: bridgedNetworkName } } },
``` 
are not compatible with `xcm::v4`?

franciscoaguirre:
Ahh indeed, that must be the issue. In v4, interior is now an array, so it should be:

```js
const foreignAssetAccount = await api.query.foreignAssets.account(
            { parents: 2, interior: { X1: [{ GlobalConsensus: bridgedNetworkName }] } },
```

I'll try that to see if that solves it

PR #4131, opened 2024-04-15
Deprecate XCMv2

franciscoaguirre:


polkadot/xcm/src/lib.rs
```diff
@@ -22,6 +22,9 @@
 // Hence, `no_std` rather than sp-runtime.
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// Because of XCMv2.
+#![allow(deprecated)]
```
bkontur, line 26: @franciscoaguirre why not just remove V2 when V5? I think we should maybe start deprecating when adding new one.
E.g let's say we have v2 v3 and when adding v4, we just mark v2 as deprecated, so when adding v5 we just remove deprecated v2 and so on.

polkadot/xcm/src/lib.rs
```diff
@@ -428,6 +431,7 @@ pub type VersionedMultiAssets = VersionedAssets;
 #[scale_info(replace_segment("staging_xcm", "xcm"))]
 pub enum VersionedXcm<RuntimeCall> {
 	#[codec(index = 2)]
+	#[deprecated]
```
bkontur, line 434: what about other V2 types? VersionedAssets/VersionedAsset/VersionedLocation? Why just `VersionedXcm`?

polkadot/xcm/src/v2/mod.rs
```diff
@@ -262,6 +267,8 @@ pub const VERSION: super::Version = 2;
 /// An identifier for a query.
 pub type QueryId = u64;
 
+/// DEPRECATED. Please use XCMv3 or XCMv4 instead.
+#[deprecated(note = "XCMv2 will be removed after June 2024. Please use XCMv3 or XCMv4 instead.")]
```
bkontur, line 271: when v5 won't be ready in the June, will we still remove v2? :)
I would say, still, we remove v2 when adding v5 in the same branch.

polkadot/xcm/src/v3/mod.rs
```diff
@@ -1576,118 +1576,6 @@ impl<Call> TryFrom<OldInstruction<Call>> for Instruction<Call> {
 #[cfg(test)]
 mod tests {
 	use super::{prelude::*, *};
-	use crate::v2::{
```
bkontur, line 1579: I would maybe remove these tests really when removing the whole v2, can we use `#[allow(deprecated)]` for tests?

polkadot/xcm/src/lib.rs
franciscoaguirre: I agree with the idea. That was the missing part. We added v4 but didn't deprecate v2. So now we deprecate v2 and remove it when we have v5.

polkadot/xcm/src/lib.rs
franciscoaguirre: Usually that's the type that determines whether you're using v2. I can mark all the others as deprecated though.

polkadot/xcm/src/v2/mod.rs
franciscoaguirre: The notice is saying it won't be removed before that date, but it can certainly be kept for longer after that date, so we're covered.
It would be good to have it ready around june though 😅

polkadot/xcm/src/v3/mod.rs
franciscoaguirre: I removed the test in the spirit of "we don't guarantee anything with deprecated versions". The code is still there, but we already don't want the maintenance burden of making sure these tests pass. Although we won't push new changes to the version, so they should always pass.

That was my reasoning, but I can just allow deprecated.

polkadot/xcm/src/lib.rs
bkontur: > I agree with the idea. That was the missing part. We added v4 but didn't deprecate v2. So now we deprecate v2 and remove it when we have v5.

exactly, ok, very cool, we are on the same boat :)

To my other comments about - reverting removed tests from XCM modules, marking `#[deprecated]` other types... My point here is to make this PR as a very good example for deprecating future versions and I would go with minimal modifications - e.g. just to add `#[deprecated]` with note "will be removed when adding new v5", I would not mention June, but also it is ok.

E.g. isn't it enough just to deprecated the whole `pub mod v2;` [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/src/lib.rs#L31) instead of deprecating particular types?

Also, we use `xcm::v2` outside of `xcm` directory, so yes, there I would maybe remove or rewrite those tests to use xcm::v3/v4.

franciscoaguirre: Yeah I could not mention June. I was just following the [deprecation checklist](https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/DEPRECATION_CHECKLIST.md) suggestions.
I can just mention v5.

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/send_xcm.rs
```diff
@@ -61,10 +61,10 @@ fn send_xcm_from_westend_relay_to_rococo_asset_hub_should_fail_on_not_applicable
...
 fn send_xcm_through_opened_lane_with_different_xcm_version_on_hops_works() {
 	// Initially set only default version on all runtimes
-	AssetHubRococo::force_default_xcm_version(Some(xcm::v2::prelude::XCM_VERSION));
-	BridgeHubRococo::force_default_xcm_version(Some(xcm::v2::prelude::XCM_VERSION));
-	BridgeHubWestend::force_default_xcm_version(Some(xcm::v2::prelude::XCM_VERSION));
-	AssetHubWestend::force_default_xcm_version(Some(xcm::v2::prelude::XCM_VERSION));
+	AssetHubRococo::force_default_xcm_version(Some(xcm::v3::prelude::XCM_VERSION));
+	BridgeHubRococo::force_default_xcm_version(Some(xcm::v3::prelude::XCM_VERSION));
+	BridgeHubWestend::force_default_xcm_version(Some(xcm::v3::prelude::XCM_VERSION));
+	AssetHubWestend::force_default_xcm_version(Some(xcm::v3::prelude::XCM_VERSION));
```
bkontur, line 67:
```suggestion
    let newer_xcm_version = xcm::prelude::XCM_VERSION;
    let older_xcm_version = newer_xcm_version - 1;
    
	AssetHubRococo::force_default_xcm_version(Some(older_xcm_version));
	BridgeHubRococo::force_default_xcm_version(Some(older_xcm_version));
	BridgeHubWestend::force_default_xcm_version(Some(older_xcm_version));
	AssetHubWestend::force_default_xcm_version(Some(older_xcm_version));
```

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/send_xcm.rs
```diff
@@ -87,42 +87,12 @@ fn send_xcm_through_opened_lane_with_different_xcm_version_on_hops_works() {
...
-		assert_err!(
-			send_asset_from_asset_hub_westend(destination.clone(), (native_token.clone(), amount)),
-			DispatchError::Module(sp_runtime::ModuleError {
-				index: 31,
-				error: [1, 0, 0, 0],
-				message: Some("SendFailure")
-			})
-		);
-	}
+	AssetHubWestend::force_xcm_version(destination.clone(), xcm::v4::prelude::XCM_VERSION);
```
bkontur, line 90:
```suggestion
	AssetHubWestend::force_xcm_version(destination.clone(), newer_xcm_version);
```

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/send_xcm.rs
```diff
@@ -87,42 +87,12 @@ fn send_xcm_through_opened_lane_with_different_xcm_version_on_hops_works() {
...
 
 	// set version with `ExportMessage` for BridgeHubWestend
 	AssetHubWestend::force_xcm_version(
 		ParentThen(Parachain(BridgeHubWestend::para_id().into()).into()).into(),
-		xcm::v3::prelude::XCM_VERSION,
+		xcm::v4::prelude::XCM_VERSION,
```
bkontur, line 95:
```suggestion
		newer_xcm_version,
```

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/send_xcm.rs
```diff
@@ -136,12 +106,12 @@ fn send_xcm_through_opened_lane_with_different_xcm_version_on_hops_works() {
 	// set version for remote BridgeHub on BridgeHubWestend
 	BridgeHubWestend::force_xcm_version(
 		bridge_hub_rococo_location(),
-		xcm::v3::prelude::XCM_VERSION,
+		xcm::v4::prelude::XCM_VERSION,
```
bkontur, line 109:
```suggestion
		newer_xcm_version,
```

cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/send_xcm.rs
```diff
@@ -136,12 +106,12 @@ fn send_xcm_through_opened_lane_with_different_xcm_version_on_hops_works() {
...
 	);
 	// set version for AssetHubRococo on BridgeHubRococo
 	BridgeHubRococo::force_xcm_version(
 		ParentThen(Parachain(AssetHubRococo::para_id().into()).into()).into(),
-		xcm::v3::prelude::XCM_VERSION,
+		xcm::v4::prelude::XCM_VERSION,
```
bkontur, line 114:
```suggestion
		newer_xcm_version,
```

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -313,13 +312,12 @@ benchmarks! {
...
-			part: v2::BodyPart::Voice,
+		let bad_loc: v3::Location = v3::Junction::Plurality {
+			id: v3::BodyId::Unit,
+			part: v3::BodyPart::Voice,
 		}
 		.into();
-		let bad_loc = VersionedLocation::from(bad_loc);
+		let bad_loc = VersionedLocation::V3(bad_loc);
 		let current_version = T::AdvertisedXcmVersion::get();
 		VersionNotifyTargets::<T>::insert(current_version, bad_loc, (0, Weight::zero(), current_version));
```
bkontur, line 322:
```suggestion
		let loc = VersionedLocation::from(Location::from(Parent));
		let current_version = T::AdvertisedXcmVersion::get();
		VersionNotifyTargets::<T>::insert(current_version, loc, (0, Weight::zero(), current_version));
```

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -313,13 +312,12 @@ benchmarks! {
...
 	notify_target_migration_fail {
-		let bad_loc: v2::MultiLocation = v2::Junction::Plurality {
-			id: v2::BodyId::Named(WeakBoundedVec::<u8, ConstU32<32>>::try_from(vec![0; 32])
-				.expect("vec has a length of 32 bits; qed")),
-			part: v2::BodyPart::Voice,
+		let bad_loc: v3::Location = v3::Junction::Plurality {
+			id: v3::BodyId::Unit,
+			part: v3::BodyPart::Voice,
 		}
 		.into();
```
bkontur, line 319: I think, that we don't need to do this hack with xcm versions, this `set_send_xcm_artificial_failure` should work

```suggestion
		// introduce artificial error in sending outbound XCM
		crate::mock::set_send_xcm_artificial_failure(true);
```

polkadot/xcm/pallet-xcm/src/benchmarking.rs
```diff
@@ -15,12 +15,11 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use super::*;
-use bounded_collections::{ConstU32, WeakBoundedVec};
 use frame_benchmarking::{benchmarks, whitelisted_caller, BenchmarkError, BenchmarkResult};
 use frame_support::{assert_ok, weights::Weight};
 use frame_system::RawOrigin;
 use sp_std::prelude::*;
-use xcm::{latest::prelude::*, v2};
+use xcm::{latest::prelude::*, v3};
```
bkontur, line 22:
```suggestion
use xcm::latest::prelude::*;
```

cumulus/pallets/xcmp-queue/src/mock.rs
```diff
@@ -337,12 +337,12 @@ pub(crate) fn mk_page() -> Vec<u8> {
...
-	let instr = xcm::v2::Instruction::<()>::ClearOrigin;
-	VersionedXcm::V2(xcm::v2::Xcm::<()>(vec![instr; 3]))
-}
-
 pub(crate) fn v3_xcm() -> VersionedXcm<()> {
 	let instr = xcm::v3::Instruction::<()>::Trap(1);
 	VersionedXcm::V3(xcm::v3::Xcm::<()>(vec![instr; 3]))
 }
+
+pub(crate) fn v4_xcm() -> VersionedXcm<()> {
```
bkontur, line 345: I would maybe refactor this tests to avoid using concrete versions, something like this should be enough:

```
pub(crate) fn versioned_xcm(v: XcmVersion) -> VersionedXcm<()> {
	let instr = Instruction::<()>::Trap(1);
    VersionedXcm::from(Xcm(vec![instr; 3])).into_version(v).expect("...")
}
```

and also use:
```
    let newer_xcm_version = xcm::prelude::XCM_VERSION;
    let older_xcm_version = newer_xcm_version - 1;
```

so when adding new XCMs we won't need to come back here again

bkontur (approved):
I added some suggestions/nits to avoid using concrete XCM versions where not needed, so we don't need to do this again with newer versions

Polkadot-Forum:

https://forum.polkadot.network/t/psa-xcmv2-deprecation/8367/1

PR #4146, opened 2024-04-16
Improve doc for pallet macro and config macro

thiolliere:
Improve doc:

* the pallet macro is actually referring to 2 places, for the module and for the struct placeholder but doesn't really clarify it (I should have named the latter just `pallet_struct` or something but it is a bit late)

* The doc of `with_default` is a bit confusing too IMO.

CC @kianenigma

### DIFF
--- a/substrate/frame/support/src/lib.rs
+++ b/substrate/frame/support/src/lib.rs
@@ -971,6 +971,10 @@ pub mod pallet_prelude {
 /// ```
 /// #[frame_support::pallet(dev_mode)]
 /// mod pallet {
+/// # 	 #[pallet::pallet]
+/// # 	 pub struct Pallet<T>(_);
+/// # 	 #[pallet::config]
+/// # 	 pub trait Config: frame_system::Config {}
 /// 	/* ... */
 /// }
 /// ```
@@ -1494,6 +1498,8 @@ pub mod pallet_macros {
 	/// # 	use frame_system::pallet_prelude::*;
 	/// # 	use core::fmt::Debug;
 	/// # 	use frame_support::traits::Contains;
+	/// #
+	/// # 	pub trait SomeMoreComplexBound {}
 	/// #
 	/// 	#[pallet::pallet]
 	/// 	pub struct Pallet<T>(_);
--- a/substrate/frame/support/src/lib.rs
+++ b/substrate/frame/support/src/lib.rs
@@ -1001,8 +1001,8 @@ pub mod pallet_prelude {
 /// <div class="example-wrap" style="display:inline-block"><pre class="compile_fail"
 /// style="white-space:normal;font:inherit;">
 /// <strong>WARNING</strong>:
-/// You should not deploy or use dev mode pallets in production. Doing so can break your
-/// chain and therefore should never be done. Once you are done tinkering, you should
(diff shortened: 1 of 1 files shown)


substrate/frame/support/src/lib.rs
```diff
@@ -936,6 +936,79 @@ pub mod pallet_prelude {
...
+/// Note that the `dev_mode` argument can only be supplied to the `#[pallet]` or
+/// `#[frame_support::pallet]` attribute macro that encloses your pallet module. This
+/// argument cannot be specified anywhere else, including but not limited to the
+/// `#[pallet::pallet]` attribute macro.
+///
+/// <div class="example-wrap" style="display:inline-block"><pre class="compile_fail"
+/// style="white-space:normal;font:inherit;">
+/// <strong>WARNING</strong>:
+/// You should not deploy or use dev mode pallets in production. Doing so can break your
+/// chain and therefore should never be done. Once you are done tinkering, you should
```
liamaharon, line 1005:
```suggestion
/// You should never deploy or use dev mode pallets in production. Doing so can break your
/// chain. Once you are done tinkering, you should
```

substrate/frame/support/src/lib.rs
```diff
@@ -936,6 +936,79 @@ pub mod pallet_prelude {
...
+///   these will be replaced by `Blake2_128Concat`. In case of explicit key-binding, `Hasher`
+///   can simply be ignored when in `dev_mode`.
+///
+/// Note that the `dev_mode` argument can only be supplied to the `#[pallet]` or
+/// `#[frame_support::pallet]` attribute macro that encloses your pallet module. This
+/// argument cannot be specified anywhere else, including but not limited to the
+/// `#[pallet::pallet]` attribute macro.
+///
+/// <div class="example-wrap" style="display:inline-block"><pre class="compile_fail"
+/// style="white-space:normal;font:inherit;">
```
liamaharon, line 1002: Could you please post a screenshot of how this looks?

thiolliere: ok, but this is not my code this is cut/paste from below. I only added the syntax part


Capture d’écran du 2024-04-16 14-44-56

thiolliere: Capture d’écran du 2024-04-16 14-47-14

liamaharon (approved):

substrate/frame/support/src/lib.rs
```diff
@@ -936,6 +936,83 @@ pub mod pallet_prelude {
...
+///
+/// 	#[pallet::call]
+/// 	impl<T: Config> Pallet<T> {
+/// 	}
+///
+/// 	/* ... */
+/// }
+/// ```
+///
+/// The documentation for each individual part can be found at [frame_support::pallet_macros]
```
franciscoaguirre, line 965: Does this result in a link?

franciscoaguirre (approved):

substrate/frame/support/src/lib.rs
thiolliere: Yes, I think rust doc link directly to item when there are in scope.

gupnik (approved):
Thank you!

PR #4147, opened 2024-04-16
Pallet macro: Better error for default config trait generation.

gui1117:
Default config implementation is using code from user to generate the default config trait.

The generated default config trait is not guaranteed to compiles. In this case user must remove some type from the default config trait.

Currently: In case the generated code fails to compile, the error message never link to the `with_default` attributes. This is confusing as adding an associated type to a pallet with `with_default` leads to error which are unexpected without knowledge about `with_default`.

Proposal: This PR generate the default config trait 2 times. One with the span of `with_default` and one as currently. So if an error is raised then the first error will link to the attribute `with_default` so that user knows the error comes from the macro generation related to this attribute. The second error will link to the associated type. Helping the user to know which associated type is to blame.

In retrospect I think it would have been better to whitelist the associated type to put into the default config instead of blacklisting them so we could link to those individual item attributes. (e.g. instead of `no_default` and `no_default_bounds` we should have `with_default` and `with_default_no_bounds` and if it fails error would link to `with_default` attribute of the trait item).

Also I added the instance on the default config trait just so that in case `I` is used in a trait item then it doesn't error with: `I` type is unknown.
But we could also scan for it and raise a precise error.
Or we could do nothing and considering that it points to `with_default` it shouldn't be less confusing.

CC @kianenigma


gui1117:
Ideally rust could handle multiple span/context, so that an error from generated code could point to both the macro and the user input inside the macro.

Issue #1625, opened 2023-09-19
[AHM] Requires a deposit for setting session keys


xlc:
Previous issue https://github.com/paritytech/substrate/issues/9267

shawntabrizi:
@bkchr i believe this should be marked as easy / medium.

Might need a storage migration, but this should be a good issue for students out of the academy.

Krayt78:
Hi Guys, i could take care of that if you want.
What would need to be done ? Just staking some tokens on set_keys and freeing them once the the keys get purged ?
And create a migration script

bkchr:
> What would need to be done ? Just staking some tokens on set_keys and freeing them once the the keys get purged ?

Yes exactly, reserve some funds.

> And create a migration script

Not needed.

PR #4155, opened 2024-04-16
On keys deregistered hook

girazoki:
Just initial draft on the idea of having a hook that is applied after session-keys are purged. This would allow to wire it with, e.g., pallet-invulnerables to immediately remove the invulnerable whose keys have been purged.

I want to capture initial opinions before I move on with this PR

### DIFF
--- a/substrate/frame/session/src/lib.rs
+++ b/substrate/frame/session/src/lib.rs
@@ -143,13 +143,17 @@ pub use weights::WeightInfo;
 
 /// Decides whether the session should be ended.
 pub trait ShouldEndSession<BlockNumber> {
-	/// Return `true` if the session should be ended.
+	// Return `true` if the session should be ended.
 	fn should_end_session(now: BlockNumber) -> bool;
 }
 
 /// Hook to be applied when an account deregisters keys
 pub trait OnKeysDeregistered<AccountId> {
 	fn on_keys_deregisterd(account: AccountId);
+}
+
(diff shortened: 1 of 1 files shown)


bkchr:
> pallet-invulnerables to immediately remove the invulnerable whose keys have been purged.

I think the general pr is fine, but I don't get the example :D Not sure we should remove an `invulnerable` in this case.

girazoki:
> > pallet-invulnerables to immediately remove the invulnerable whose keys have been purged.
> 
> I think the general pr is fine, but I don't get the example :D Not sure we should remove an `invulnerable` in this case.

So correct me if I am wrong here, but before adding an invulnerable I think we check that the invulnerable account has session keys registered in https://github.com/paritytech/polkadot-sdk/blob/e81322bc3e8192b536067fed3ef9e20f2752c376/cumulus/pallets/collator-selection/src/lib.rs#L396.

However if the invulnerable somehow purges keys, it will still show as a valid invulnerable to be selected. One potential usecase for this hook is to immediately de-register the invulnerable if it purges keys. But maybe I am wrong here with my assumptions?

bkchr:
Thank you for the explanation. I think it would be better to implement this: https://github.com/paritytech/polkadot-sdk/issues/1625 and then have some kind of "locking" to prevent that they can prune the keys as long as they are registered for staking/collators.

girazoki:
Closing this as we probably work in favour of having a "locked" deposit

Issue #4123, opened 2024-04-15
RFC: Meta Transaction Implementation


muharem:
## Summary

This document proposes several potential solutions for implementing Meta Transactions for FRAME, accompanied by draft implementations. The first solution relies solely on a runtime’s general transaction and its transaction extensions, also known as signed extensions. In contrast, the second and third solutions introduce a new pallet call. In these solutions, validation pipeline for a meta transaction is abstracted via transaction extensions or a new contract/trait and can be defined through the pallet's configuration.

## Motivation

The concept of Meta Transaction is well-established in the Ethereum ecosystem, referring to a transaction authorised by one party (signer) and executed by an untrusted third party that covers the transaction fees (relayer). This concept proves useful in scenarios where the signer lacks the assets to cover the fee or lacks the incentive to do so. 
Examples include (source: https://github.com/paritytech/polkadot-sdk/issues/266):
- dApp covering transaction fees for its users;
- proxy accounts lacking balance;
- transaction fees paid in any asset when a signer incentivizes a relayer to cover the fee, as demonstrated by a transaction such as `batch([sendCustomToken(relayerAddr, amount), doTheActualWork()])`;
- signer delegating voting power without covering the transaction fee.

The solution must ensure that a relayer can execute only actions authorized by the signer and is protected from attacks such as replay attacks. Since it does concerns the transaction layer, it must be aligned with the Extrinsic Horizon (https://github.com/paritytech/polkadot-sdk/issues/2415).

Additionally, an important consideration is the complexity of implementation and integration for clients. Furthermore, this proposal addresses the issue where a signer lacks an on-chain account, preventing the storage of information such as nonce (an issue initially raised and discussed here - https://github.com/paritytech/polkadot-sdk/issues/266).

The initial solution, which relies solely on extending the extension type of a general transaction, has sparked some scepticism regarding its complexity. Therefore, alternative solutions are presented here, along with draft implementations.

### Solution 1: Meta Transaction as Extension to General Transaction.

Draft Implementation: https://github.com/paritytech/polkadot-sdk/pull/3712

The first solution extends the runtime’s transaction extension type with two new extensions. Additionally, it introduces a new context type for transaction extension to share a relevant information down the pipeline.
```
TxExtension = (VerifyRelayerSig, VerifyAccountSig, MetaTxExtension);
MetaTxExtension = (CheckNonZeroSender, CheckSpecVersion, ..., RelayTransactionPayment);
```
[...]

PR #4165, opened 2024-04-17
Fixes nonce issue with feeless calls

### LINKED ISSUE #3991: `CheckNonce` should not enforce the account creation requirement for feeless calls
As raised in https://substrate.stackexchange.com/questions/11259/how-to-enable-free-execution-for-specific-extrinsics-using-the-skipcheckiffeeles/11262#11262, `SkipCheckIfFeeless`, though skips the payment extensions, doesn't work as intended due to these requirements in `CheckNonce`:

https://github.com/paritytech/polkadot-sdk/blob/0bbda78d86bc6210cda123042d817aeaf45b3d77/substrate/frame/system/src/extensions/check_nonce.rs#L83-L86

So, we need a mechanism for `CheckNonce` to skip these checks for feeless calls.

gupnik:
Fixes https://github.com/paritytech/polkadot-sdk/issues/3991

`feeless_if` was introduced as a mechanism to trigger feeless calls without requiring an existence of the account. However, the requirement for sufficients/providers to be non-zero in `CheckNonce` makes it impossible to do so.

This PR removes this check for an account with 0 nonce if the ED has been provided as a result of the execution. This is validated during `post_dispatch`. 

NOTE: This check in `post_dispatch` should NEVER fail, otherwise this will lead to wasted work by the block producers.

### DIFF
--- a/substrate/frame/system/src/extensions/check_nonce.rs
+++ b/substrate/frame/system/src/extensions/check_nonce.rs
@@ -17,7 +17,10 @@
 
 use crate::Config;
 use codec::{Decode, Encode};
-use frame_support::dispatch::DispatchInfo;
+use frame_support::{
+	dispatch::{CheckIfFeeless, DispatchInfo},
+	traits::OriginTrait,
+};
 use scale_info::TypeInfo;
 use sp_runtime::{
 	traits::{DispatchInfoOf, Dispatchable, One, SignedExtension, Zero},
@@ -60,12 +63,16 @@ impl<T: Config> sp_std::fmt::Debug for CheckNonce<T> {
 
 impl<T: Config> SignedExtension for CheckNonce<T>
 where
-	T::RuntimeCall: Dispatchable<Info = DispatchInfo>,
+	T::RuntimeCall: Dispatchable<Info = DispatchInfo>
+		+ CheckIfFeeless<Origin = crate::pallet_prelude::OriginFor<T>>,
 {
 	type AccountId = T::AccountId;
 	type Call = T::RuntimeCall;
 	type AdditionalSigned = ();
-	type Pre = ();
+	type Pre = (
+		Self::AccountId, // who
+		bool,            // is feeless with zero nonce
+	);
 	const IDENTIFIER: &'static str = "CheckNonce";
 
 	fn additional_signed(&self) -> sp_std::result::Result<(), TransactionValidityError> {
@@ -75,14 +82,19 @@ where
 	fn pre_dispatch(
 		self,
 		who: &Self::AccountId,
-		_call: &Self::Call,
+		call: &Self::Call,
 		_info: &DispatchInfoOf<Self::Call>,
 		_len: usize,
-	) -> Result<(), TransactionValidityError> {
+	) -> Result<Self::Pre, TransactionValidityError> {
 		let mut account = crate::Account::<T>::get(who);
+		let mut is_feeless_with_zero_nonce = false;
 		if account.providers.is_zero() && account.sufficients.is_zero() {
-			// Nonce storage not paid for
-			return Err(InvalidTransaction::Payment.into())
+			if !(call.is_feeless(&T::RuntimeOrigin::signed(who.clone())) && account.nonce.is_zero())
+			{
+				return Err(InvalidTransaction::Payment.into())
+			} else {
+				is_feeless_with_zero_nonce = true;
+			}
 		}
 		if self.0 != account.nonce {
 			return Err(if self.0 < account.nonce {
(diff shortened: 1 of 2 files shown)


substrate/frame/system/src/extensions/check_nonce.rs
```diff
@@ -92,22 +104,52 @@ where
...
+		if let Some((who, is_feeless_with_zero_nonce)) = _pre {
+			if !is_feeless_with_zero_nonce {
+				Ok(())
+			} else {
+				let mut account = crate::Account::<T>::get(who.clone());
+				if account.providers.is_zero() && account.sufficients.is_zero() {
+					return Err(InvalidTransaction::Payment.into())
+				} else {
+					account.nonce += T::Nonce::one();
+					crate::Account::<T>::insert(who, account);
```
liamaharon, line 130: Would this not lead to accounts submitting feeless txns bloating the chain at no cost? There should be very clear docs explaining this trade off, and potential vulnerability here.

gupnik: > Would this not lead to accounts submitting feeless txns bloating the chain at no cost? 

Not if ED is being provided during the execution of the call. This is what is being validated in the `post_dispatch`.

> There should be very clear docs explaining this trade off, and potential vulnerability here.

Indeed. 

Open to suggestions if there is a better way to do this.

liamaharon: Probably best to mention here https://paritytech.github.io/polkadot-sdk/master/frame_support/pallet_macros/attr.feeless_if.html

gupnik: > Probably best to mention here https://paritytech.github.io/polkadot-sdk/master/frame_support/pallet_macros/attr.feeless_if.html

Thanks. I meant if there are alternate approaches that could be tried :)

liamaharon: Could we destroy the account after the extrinsic is processed and it's no longer needed?

gupnik: > Could we destroy the account after the extrinsic is processed and it's no longer needed?

But the feeless call might have been used for an airdrop (based on a certain check) which could have paid for the ED. This is one of the major use cases. In that case, we don't want to destroy it.

georgepisaltu:
Commented on the original issue too but other arguments against doing this would be:
- this feature is a huge footgun; users can easily DOS their chain if they see this extension, assume it can make some calls free based on conditional code (which it does), and then attach this to some extrinsic they have (one that doesn't endow)
- it's bad to condition the behavior of a crucial extension like `CheckNonce` based on what an optional extension might or might not do; extensions are in a pipeline and are independent for a reason
- we introduce feeless logic into an extension that is supposed to check nonces to guard against replay attacks; this is a "separation of concerns" problem as one should have nothing to do with the other
- this will make more sense and be easier to implement in the context of `General` transactions and `TransactionExtension` (see the POCs for meta transactions [here](https://github.com/paritytech/polkadot-sdk/pull/3712) and [here](https://github.com/paritytech/polkadot-sdk/pull/4122))

substrate/frame/system/src/extensions/check_nonce.rs
georgepisaltu: Doesn't this change make the `SkipCheckIfFeeless` extension usable only when the underlying extrinsic does endow because of the attack vector introduced by [invalidating entire blocks on `post_dispatch`](https://github.com/paritytech/polkadot-sdk/blob/ba0f8de0c74b1308a53942c406137a748ef79925/substrate/primitives/runtime/src/traits.rs#L1576)?

gupnik: Currently, no transaction can be triggered at all when `SkipCheckIfFeeless` and `CheckNonce` are used at the same time, cause the check for nonce would cause all such transactions to be invalid. This just provides an additional mechanism to be able to at least trigger those calls where the endowment occurs during the execution. 

Runtime can devs can definitely customise their CheckNonce extension to handle this specifically, but I am not sure if there's a downside to make it a bit easier?

gupnik:
Replied on the issue as well, but for these points specifically:

> it's bad to condition the behavior of a crucial extension like CheckNonce based on what an optional extension might or might not do; extensions are in a pipeline and are independent for a reason

I think there's a confusion between `pallet::feeless_if` macro and `SkipCheckIfFeeless` extension. The former marks a call in your pallet as feeless while the latter uses this information to skip the transaction payment. The change here only uses the former and does not rely on the presence of the latter.

> we introduce feeless logic into an extension that is supposed to check nonces to guard against replay attacks; this is a "separation of concerns" problem as one should have nothing to do with the other

Same answer as above

> this feature is a huge footgun; users can easily DOS their chain if they see this extension, assume it can make some calls free based on conditional code (which it does), and then attach this to some extrinsic they have (one that doesn't endow)

Indeed, but I was hoping that this could be solved by docs. Happy to hear other suggestions as well.

> this will make more sense and be easier to implement in the context of General transactions and TransactionExtension (see the POCs for meta transactions https://github.com/paritytech/polkadot-sdk/pull/3712 and https://github.com/paritytech/polkadot-sdk/pull/4122)

I would explore this further to see how a POC with the transaction extension would like. Thanks.

substrate/frame/system/src/extensions/check_nonce.rs
georgepisaltu: > Currently, no transaction can be triggered at all when SkipCheckIfFeeless and CheckNonce are used at the same time, cause the check for nonce would cause all such transactions to be invalid.

Even transactions where the user already exists? If I have a funded account and I submit an extrinsic with a call that passes the `feeless_if` checks, does it not work? As I said here, the two problems are different and allowing nonexistent accounts to run signed transactions for free is difficult to solve and requires a robust solution because of the attack vector introduced. IMO the approach here is too big of a compromise.

georgepisaltu:
Again, the problems are separate and I'd like to focus only on the first one. If this PR is meant to target the problem of allowing nonexistent accounts to run signed transactions for free, then I am against the approach and I favor what is described in the [meta transaction RFC](https://github.com/paritytech/polkadot-sdk/issues/4123) as solution 2. Initial issue for reference #266.

> I was hoping that this could be solved by docs

I don't think docs can realistically cover this corner case given the amount of hidden complexity pertaining not only to this PR, but to the entire extension logic.

> Happy to hear other suggestions as well

Please refer to #4123.

> I would explore this further to see how a POC with the transaction extension would like.

The linked POCs above do the first part of relaying a transaction, with the mechanism for providing for a new account before relaying described in the RFC, it's just an idea right now but it would be a provider pallet based on deposits, or the relayer can pay for the ED. Unfortunately, I'm too short on time to come up with a POC for it in the timeframe of this PR.

substrate/frame/system/src/extensions/check_nonce.rs
liamaharon: > > Could we destroy the account after the extrinsic is processed and it's no longer needed?
> 
> But the feeless call might have been used for an airdrop (based on a certain check) which could have paid for the ED. This is one of the major use cases. In that case, we don't want to destroy it.

This would bypass ED requirements, and the acc would get destroyed next ED check? :/

I think if we want to allow airdrops to new accounts, that airdrop asset needs to be sufficient otherwise we are just throwing away the point of EDs and sufficient assets..

muharem: To me this logic looks valid. I think it even valid without `is_feeless` condition (but without it it will fail later).
But I do not like the way it complicates the general implementation used by most of the runtime. The use case is quite specific (airdrop claims). May be we provide another implementation which handles this use case.

gupnik: > To me this logic looks valid. I think it even valid without `is_feeless` condition (but without it it will fail later). But I do not like the way it complicates the general implementation used by most of the runtime. The use case is quite specific (airdrop claims). May be we provide another implementation which handles this use case.

Yes, I was thinking to implement a wrapper similar to the `SkipCheckIfFeeless` extension itself such that the wrapped `CheckNonce` will only be called if the feeless conditions are met and the nonce is 0. This would remove the coupling here, while providing an extensible mechanism.

rossbulat:
This PR looks stale now. It specifically changes `CheckNonce` to allow a zero-nonce feeless call from an account with no providers/sufficients, then verifies in `post_dispatch` that the account was endowed before incrementing the nonce.

Based on the review discussion, this approach seems unlikely to be the right direction: it couples `CheckNonce` to `#[pallet::feeless_if]` behavior and adds feeless-call logic into the nonce extension. The suggested direction in the thread was to handle this through the meta-transaction / `TransactionExtension` path instead.

Given there has been no follow-up since June 2024, I think we can close this PR. If issue #3991 is still relevant, it would be better to reopen it as a fresh, focused PR against the current transaction extension design.

PR #4168, opened 2024-04-17
chainHead: Stabilize chainHead to version 1

lexnv:
This PR stabilizes the chainHead API to version 1.

Needs:
- https://github.com/paritytech/polkadot-sdk/pull/3667

cc @paritytech/subxt-team


jsdw (approved):
I'm happy with this!

~~I wonder whether, in this initial PR, we should alias eg chainHead_unstable_foo to call chainHead_v1_foo, just so that clients like PAPI/Subxt will continue to work until they update. I'm not personally too fussed from Subxt's PoV though.~~ - Since PAPI uses Smoldot anyway it probably doesn't care, and I don't care for Subxt so I think we can just change them to v1 without any extra effort.

niklasad1 (approved):

Issue #4126, opened 2024-04-15
Determine the sweet spot for num execute_workers_max_num & prepare_workers_max_num


alexggh:
With the upcoming async backing changes where we are increasing the
parachain authoring time to 2 seconds instead of the max 500ms, 2
execution workers might prove not to be enough when multiple parachains
produce blocks that take 2 seconds to verify we will create very easy a
backlog of candidates we need to verify.

In the best case scenario a validator has to verify at least 7
candidates, the 6 tranch0 assignments and the candidate it helps with
backing, so if all of them take 2 seconds on the worst case scenario you
end-up needing 14s of execution time each block, so spliting that
between two workers you would need 7s of execution each 6s, that get us
in a situation where the pvf execution workers become the bottleneck of
the system.

Old PR where we changed this in the past: https://github.com/paritytech/polkadot/pull/4273

## Remaining work
- [x] After https://polkadot.subsquare.io/referenda/1051?tab=timeline goes into effect bump the number of PVF execution  workers from 2 to 4 on kusama and polkadot.

eskimor:
Thanks @alexggh and @s0me0ne-unkn0wn  for raising the issue.

Let's double the workers and go for Kusama first. 

Other things to consider:

1. New networking stack has been merged and if successfully enabled on Parity Kusama validators, we should make it the default soon.
2. Together with our optimizations in approval voting (and more to come), we should be able to scale up the number of validators. Would be good to reach 1000 this year, this also reduces the amount of validation work needed per validator - by a factor.

alexggh:
> Let's double the workers and go for Kusama first.

Unfortunately, the number of workers is hardcoded in the node excutable, https://github.com/paritytech/polkadot-sdk/blob/6f73b746d3caee00030e55369f9e5e0ddac15d83/polkadot/node/core/pvf/src/host.rs#L204, so we can't really deploy it just for kusama since it would go in a node release.

However, I do think increasing the executor numbers from 2 to 4 should be uber low risk.

Here we can seee  that https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot#reference-hardware, our recommended HW spec is 4 hardware cores and 32GiB of ram, so we definitely should have space for `4` executions in parallel.

sandreim:
> > Let's double the workers and go for Kusama first.
> 
> Unfortunately, the number of workers is hardcoded in the node excutable,
> 
[...]

PR #4172, opened 2024-04-17
Use higher priority for PVF preparation in dispute/approval context

sandreim:
Related to https://github.com/paritytech/polkadot-sdk/issues/4126 discussion

Currently all preparations have same priority and this is not ideal in all cases. This change should improve the finality time in the context of on-demand parachains and when `ExecutorParams` are updated on-chain and a rebuild of all artifacts is required. The desired effect is to speed up approval and dispute PVF executions which require preparation and delay backing executions which require preparation.


polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -870,11 +892,13 @@ impl ValidationBackend for ValidationHost {
 		pvf: PvfPrepData,
 		exec_timeout: Duration,
 		encoded_params: Vec<u8>,
+		// The priority for the preparation job.
+		prepare_priority: polkadot_node_core_pvf::Priority,
 	) -> Result<WasmValidationResult, ValidationError> {
-		let priority = polkadot_node_core_pvf::Priority::Normal;
```
alexggh, line 874: So, we never used the priority logic.

alexggh (approved):
The changes look sane to me, however I don't think the impact is significant in the situation where you have to recompile all, because at any point PVF backing jobs are way less than PVF approval jobs, because a validator is usually allocated to a single parachain for backing while it is doing approval for at least an order of magnitude more.

polkadot/node/core/pvf/src/priority.rs
```diff
@@ -20,11 +20,12 @@ pub enum Priority {
...
 	/// done pretty quick.
 	///
-	/// Approvals and disputes fall into this category.
+	/// Backing falls into this category.
 	Normal,
 	/// This priority is used for requests that are required to be processed as soon as possible.
 	///
-	/// For example, backing is on a critical path and requires execution as soon as possible.
+	/// Disputes and approval are on a critical path and requires execution as soon as
+	/// possible to not delay finality.
```
Overkillus, line 28: Backing was deemed more important than approvals and disputes? :fearful: 

If I understand it correctly that is an awesome find and fix. I was not aware of it. It also aligns much better with out backpressure plans - first thing you should stop doing when overworked is backing, NOT approvals. 

But despite the comments indicating that... I can't find any anything in the original code that sets anything to `Priority::Critical`. Huh.

polkadot/node/core/candidate-validation/src/lib.rs
sandreim: yup

Overkillus (commented):
Good find. 

One thing we need to be conscious about is that this effectively introduces a form of backing back-off when nodes are overworked. Not an issue but if we ever notice some weird fluctuations in backing frequency this should be a prime candidate for investigation.

s0me0ne-unkn0wn (approved):
How the hell did that happen? :thinking: 
I'll try to dig into history to investigate if we lost in at some point or never had indeed

sandreim:
> One thing we need to be conscious about is that this effectively introduces a form of backing back-off when nodes are overworked. Not an issue but if we ever notice some weird fluctuations in backing frequency this should be a prime candidate for investigation.

Yes, we need that back pressure, but this change is only relevant for PVF preparation. Execution remains unaffected so I wouldn't say node overworked, more like preparation thread is overworked :)

polkadot/node/core/pvf/src/priority.rs
sandreim: I am looking at the execution path as well, but that seem to require a lot more changes. 
However the backing execution backpressure to activate at the network level requires at least the backing group to be overloaded, but that also depends on the validator hardware, some might be faster or some might provision more CPUs.

s0me0ne-unkn0wn: I've checked the original paritytech/polkadot#2710 and it seems it was like that from the very beginning :shrug: 
Maybe @pepyakin could remember why it was implemented like that

Overkillus:
> Yes, we need that back pressure, but this change is only relevant for PVF preparation. Execution remains unaffected so I wouldn't say node overworked, more like preparation thread is overworked :)

Very good point. Missed it that this is only prep related. Removes most of the worry then

polkadot/node/core/pvf/src/priority.rs
sandreim: Yeah, you might be right that this was far worse before async backing, but even now it is not so great either. Depending on the actual PVF, the preparation can take 10-20s so the relay parent could go out of scope. Given that we have 1 single worker doing this right now, finality can be delayed for quite a lot of time and this would affect finality on all parachains vs just one parachain that won't be able to get it's blocks included.

sandreim: Actually, we could also bump `prepare_workers_hard_max_num` from 1 to 2. This would be used only for preparing PVFs tagged with `Critical` priority.

s0me0ne-unkn0wn: Sounds like a really good idea, but again, it would be good to have a number of CPU cores reserved in the hardware requirements for the maximum number of workers that may run in parallel.

polkadot/node/core/pvf/src/priority.rs
```diff
@@ -20,11 +20,12 @@ pub enum Priority {
 	/// Normal priority for things that do not require immediate response, but still need to be
 	/// done pretty quick.
 	///
-	/// Approvals and disputes fall into this category.
+	/// Backing falls into this category.
```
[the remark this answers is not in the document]
sandreim: Yes this is currently used for preparation only. You ate right, with pre-checking the artifact should be already there. It helps mostly with missing artifacts due to prunning or executor parameter changes.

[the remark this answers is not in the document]
s0me0ne-unkn0wn: > maybe the PVF artifact pruning is too aggressive?

I wouldn't say it is too aggressive, we prune an artifact if it hasn't been used for 24 hours, which seems like a fair measure to prevent artifact cache bloating.

> I would expect them to have the artifacts even for new PVFs due to pre-checking

Generally, yes, but if a validator pre-checked a PVF and didn't use it for 24 hours, then again, it has to re-compile it.

polkadot/node/core/candidate-validation/src/lib.rs
```diff
@@ -776,8 +785,15 @@ trait ValidationBackend {
...
 
-		let mut validation_result =
-			self.validate_candidate(pvf.clone(), exec_timeout, params.encode()).await;
+		// Use `Priority::Critical` as finality trumps parachain liveliness.
+		let mut validation_result = self
+			.validate_candidate(
+				pvf.clone(),
+				exec_timeout,
+				params.encode(),
+				polkadot_node_core_pvf::Priority::Critical,
```
eskimor, line 794: Nah. This is way to implicit. The docs of `validate_candidate_with_retry` say that is is used in approvals, but this is just an assumption about its usage, one the might even make sense to remove. The docs should focus on what the function does and the caller then decides whether it is suitable for the job at hand.

TL;DR: This function should take a priority argument, instead of assuming calling context.

sandreim: yeah it makes sense

polkadot/node/core/pvf/src/priority.rs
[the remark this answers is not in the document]
sandreim: > If we have them compiled already, then we should actually prioritize compiling new PVFs as soon as we see them (backing). So again, this change would not have helped in this case. Bumping preparation workers would though.

For the executor parameter changed case . We are more often assigned to validate candidates rather than back candidates for which the PVF needs to be recompiled. I don't understand why would backing priority helps in this case. Assuming we prioritize backing, we'd basically no-show on more of the assignments we have for many blocks, leading to other tranches triggering increasing the work of approval subsystems needlessly as finality lag increases. Bumping prepration workers hard cap might be a better choice for the time being along with this change.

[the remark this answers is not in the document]
ordian: > Assuming we prioritize backing, we'd basically no-show on more of the assignments we have for many blocks, leading to other tranches triggering increasing the work of approval subsystems needlessly as finality lag increases. Bumping prepration workers hard cap might be a better choice for the time being along with this change.

I was describing a case when we have PVFs already compiled for unapproved (old) blocks. In that case, we would not no-show. What I meant is that we should put new PVF for preparation as soon as possible.

ordian (approved):
Approving, but I think we should also bump the prep worker limit, at least the hard one, to have an effect on the situation when we need to recompile all PVFs at the same time in real time.

AndreiEres (approved):
Should help to keep finality. But I agree that it would be good to consider separately the case where we need to recompile many pvfs.

sandreim:
> Approving, but I think we should also bump the prep worker limit, at least the hard one, to have an effect on the situation when we need to recompile all PVFs at the same time in real time.

bumped hard limit to `2`

PR #4183, opened 2024-04-18
remove unnessisary use statements due to 2021 core prelude

gilescope:
Some traits are already included in the 2021 prelude and so shouldn't be needed to use explicitly:

use `convert::TryFrom`, `convert::TryInto`, and `iter::FromIterator` are removed.

( https://doc.rust-lang.org/core/prelude/rust_2021/ )

No breaking changes or change of functionality, so I think no PR doc is needed in this case.

(Motivation: Removes some references to `sp-std`)


bridges/bin/runtime-common/src/messages.rs
```diff
@@ -35,7 +35,7 @@ use frame_support::{traits::Get, weights::Weight};
 use hash_db::Hasher;
 use scale_info::TypeInfo;
 use sp_runtime::RuntimeDebug;
-use sp_std::{convert::TryFrom, marker::PhantomData, vec::Vec};
+use sp_std::{marker::PhantomData, vec::Vec};
```
ggwpez, line 38: We can also use `core::marker`.

ggwpez (approved):
`sp-core-fuzzer` is on edition 2015. Could you please also fix that while youre at it?

bridges/bin/runtime-common/src/messages.rs
gilescope: Agreed but that's one for another PR. I want to keep each PR as easy to review as possible by not fixing too many things at once. (There's a lot of references that need changing.)

gilescope:
Good spot. There were a few editions set to 2021 but not pointing to the workspace so I've repointed them also.

bkchr (approved):

muharem (approved):

seadanda (approved):

gilescope:
@bkontur I need a thumbs-up from the bridges team if you're ok with the changes?

acatangiu (approved):

PR #4185, opened 2024-04-18
State trie migration on asset-hub westend and collectives westend

cheme:


cheme:
@PierreBesson I guess 5F4EbSkZz18X36xhbsjvDNs6NuZ82HyYtq5UiJ1h9SBHJXZD would be fine for asset-hub westend, is it the case for collectives westend, or which key can be used?

PierreBesson:
@cheme Yes, you can use the same key for all of them.

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
```diff
@@ -1770,3 +1772,46 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+impl pallet_state_trie_migration::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Currency = Balances;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type SignedDepositPerItem = MigrationSignedDepositPerItem;
+	type SignedDepositBase = MigrationSignedDepositBase;
+	// An origin that can control the whole pallet: should be Root, or a part of your council.
+	type ControlOrigin = frame_system::EnsureSignedBy<RootMigController, AccountId>;
+	// specific account for the migration, can trigger the signed migrations.
+	type SignedFilter = frame_system::EnsureSignedBy<MigController, AccountId>;
```
ggwpez, line 1792: For Polkadot and Kusama parachains we probably want this to be permissionless, or?  
Then we could also test it here?

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
```diff
@@ -1770,3 +1772,46 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	pub const RootMigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
+}
+
+#[test]
+fn ensure_key_ss58() {
+	use frame_support::traits::SortedMembers;
+	use sp_core::crypto::Ss58Codec;
+	let acc =
+		AccountId::from_ss58check("5F4EbSkZz18X36xhbsjvDNs6NuZ82HyYtq5UiJ1h9SBHJXZD").unwrap();
+	//panic!("{:x?}", acc);
```
ggwpez, line 1811:
```suggestion
```

cumulus/parachains/runtimes/collectives/collectives-westend/src/lib.rs
```diff
@@ -1080,3 +1082,46 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	use frame_support::traits::SortedMembers;
+	use sp_core::crypto::Ss58Codec;
+	let acc =
+		AccountId::from_ss58check("5F4EbSkZz18X36xhbsjvDNs6NuZ82HyYtq5UiJ1h9SBHJXZD").unwrap();
+	//panic!("{:x?}", acc);
+	assert_eq!(acc, MigController::sorted_members()[0]);
+	let acc =
+		AccountId::from_ss58check("5F4EbSkZz18X36xhbsjvDNs6NuZ82HyYtq5UiJ1h9SBHJXZD").unwrap();
+	assert_eq!(acc, RootMigController::sorted_members()[0]);
+	//panic!("{:x?}", acc);
```
ggwpez, line 1126:
```suggestion
```

ggwpez (approved):
LGTM, thanks!

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
```diff
@@ -1770,3 +1772,46 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	assert_eq!(acc, MigController::sorted_members()[0]);
+	let acc =
+		AccountId::from_ss58check("5F4EbSkZz18X36xhbsjvDNs6NuZ82HyYtq5UiJ1h9SBHJXZD").unwrap();
+	assert_eq!(acc, RootMigController::sorted_members()[0]);
+	//panic!("{:x?}", acc);
```
liamaharon, line 1816:
```suggestion
```

liamaharon (approved):

cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs
cheme: I think the idea is to use specific account eg bridge one (I would still prefer https://github.com/polkadot-fellows/runtimes/issues/74#issuecomment-1943400824 but this is more practical probably).
Maybe ControlOrigin is not strictly necessary, but SignedFilter is, otherwise we could have frontrunning issue (two ext on same block, then only one get refund, also we don't want multiple ext).

PR #2129, opened 2023-11-02
XCM MultiAssets: sort after reanchoring

serban300:
Fixes https://github.com/paritytech/polkadot-sdk/issues/2123

PR #4186, opened 2024-04-18
Ensure outbound XCMs are decodable with limits + add `EnsureDecodableXcm` router (for testing purposes)

bkontur:
This PR:
- adds `EnsureDecodableXcm` (testing) router that attempts to *encode* and *decode* passed XCM `message` to ensure that the receiving side will be able to decode, at least with the same XCM version.
- fixes `pallet_xcm` / `pallet_xcm_benchmarks` assets data generation

Relates to investigation of https://substrate.stackexchange.com/questions/11288 and missing fix https://github.com/paritytech/polkadot-sdk/pull/2129 which did not get into the fellows 1.1.X release.

## Questions/TODOs

- [x] fix XCM benchmarks, which produces undecodable data - new router catched at least two cases
  - `BoundedVec exceeds its limit`
  - `Fungible asset of zero amount is not allowed`  
- [x] do we need to add `sort` to the  `prepend_with` as we did for reanchor [here](https://github.com/paritytech/polkadot-sdk/pull/2129)? @serban300 (**created separate/follow-up PR**: https://github.com/paritytech/polkadot-sdk/pull/4235)
- [x] We added decoding check to `XcmpQueue` -> `validate_xcm_nesting`, why not to added to the `ParentAsUmp` or `ChildParachainRouter`? @franciscoaguirre  (**created separate/follow-up PR**: https://github.com/paritytech/polkadot-sdk/pull/4236)
- [ ] `SendController::send_blob` replace `VersionedXcm::<()>::decode(` with `VersionedXcm::<()>::decode_with_depth_limit(MAX_XCM_DECODE_DEPTH, data)` ?


polkadot/xcm/xcm-builder/src/routing.rs
```diff
@@ -139,3 +139,69 @@ impl EnsureDelivery for Tuple {
...
 }
+
+/// A wrapper router that attempts to *encode* and *decode* passed XCM `message` to ensure that the
+/// receiving side will be able to decode, at least with the same XCM version.
+///
+/// This is designed to be at the top-level of any routers which do the real delivery. While other
+/// routers can manipulate the `message`, we cannot access the final XCM due to the generic
+/// `Inner::Ticket`. Therefore, this router aims to validate at least the passed `message`.
+///
+/// NOTE: For testing purposes.
```
acatangiu, line 150:
```suggestion
/// NOTE: For use in mock runtimes which don't have the DMP/UMP/HRMP XCM validations.
```

cumulus/pallets/xcmp-queue/src/lib.rs
```diff
@@ -952,7 +952,7 @@ impl<T: Config> SendXcm for Pallet<T> {
 
 /// Checks that the XCM is decodable with `MAX_XCM_DECODE_DEPTH`.
 ///
-/// Note that this uses the limit of the sender - not the receiver. It it best effort.
+/// Note that this uses the limit of the sender - not the receiver. It is a best effort.
 pub(crate) fn validate_xcm_nesting(xcm: &VersionedXcm<()>) -> Result<(), ()> {
```
bkontur, line 956: @ggwpez I see that you added `validate_xcm_nesting` [here](https://github.com/paritytech/polkadot-sdk/pull/1246/files#diff-31d504e3776a85fc47e3c11e673819cc87bdaf712bb613ea366360fe8b8a1786R929) but only for `XcmpQueue`. Don't you remember if there was any reason not to add it also to the [`ParentAsUmp`](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/primitives/utility/src/lib.rs#L70-L72) or [`ChildParachainRouter`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/common/src/xcm_sender.rs#L122)?

And another question, is that [`defensive_assert`](https://github.com/paritytech/polkadot-sdk/pull/1246/files#diff-31d504e3776a85fc47e3c11e673819cc87bdaf712bb613ea366360fe8b8a1786R945) still needed? I think, that check in `fn validate` is enough, could we remove it?

polkadot/xcm/xcm-builder/src/routing.rs
```diff
@@ -139,3 +139,69 @@ impl EnsureDelivery for Tuple {
...
 }
+
+/// A wrapper router that attempts to *encode* and *decode* passed XCM `message` to ensure that the
+/// receiving side will be able to decode, at least with the same XCM version.
+///
+/// This is designed to be at the top-level of any routers which do the real delivery. While other
+/// routers can manipulate the `message`, we cannot access the final XCM due to the generic
+/// `Inner::Ticket`. Therefore, this router aims to validate at least the passed `message`.
+///
+/// NOTE: For use in mock runtimes which don't have the DMP/UMP/HRMP XCM validations.
```
ggwpez, line 150: Why would this only be used for testing? I think it is more important for produciton deployment to prevent invalid messages from ever being sent.

polkadot/xcm/xcm-builder/src/routing.rs
```diff
@@ -139,3 +139,69 @@ impl EnsureDelivery for Tuple {
...
+	fn deliver(ticket: Self::Ticket) -> Result<XcmHash, SendError> {
+		Inner::deliver(ticket)
+	}
+}
+impl<Inner> EnsureDecodableXcm<Inner> {
+	fn ensure_encode_decode(xcm: &Xcm<()>) -> Option<SendError> {
+		use parity_scale_codec::Decode;
+		let encoded = xcm.encode();
+		match Xcm::<()>::decode(&mut &encoded[..]) {
+			Ok(decoded_xcm) => match decoded_xcm.eq(xcm) {
```
ggwpez, line 176: It needs to check the nesting limit for encoded XCM.

polkadot/xcm/xcm-builder/src/routing.rs
bkontur: well, the thing is that when we wrap `EnsureDecodableXcm<XcmpQueue>` it will be able to check message which **is passed** to the `XcmpQueue`, which could be ok, but we have other routers which internally can modify message and possibly break validity.  

We could just wrap `EnsureDecodableXcm<XcmpQueue>`, but we pass `XcmpQueue` on several places:
```
SovereignPaidRemoteExporter<bridging::EthereumNetworkExportTable, XcmpQueue, UniversalLocation>
...
type ToBridgeHubSender = XcmpQueue;
...
```

Yes, we could wrap "everywhere", but it can be easily missed.
So, I think, if we want to be pretty sure, this validation should be as a part of `XcmpQueue` / `ParentAsUmp` / `ChildParachainRouter` internal implementation, and I think that `SendXcm::validate` is the best place (as you put it for `XcmpQueue`) and also there is no need to do double-check in `SendXcm::deliver`

ggwpez: > So, I think, if we want to be pretty sure, this validation should be as a part of XcmpQueue / ParentAsUmp / ChildParachainRouter internal implementation

Yea 👍 did not want to suggest it since i though that Router would be better, but if it always needs to wrapped then just put it directly in the pallet.

cumulus/pallets/xcmp-queue/src/lib.rs
bkontur: > I did not think about the fact that other delivery protocols could encounter the same issue, so i only added it for HRMP outbound messages.

@ggwpez I created a separate [PR](https://github.com/paritytech/polkadot-sdk/pull/4236) for adding `validate_xcm_nesting` elsewhere, but for completeness, what kind of issue was that? I mean, why does it only affect HRMP and not DMP or UMP? Could you please provide more info to that PR (if possible)?

polkadot/xcm/xcm-builder/src/routing.rs
bkontur: fixed by reused with `validate_xcm_nesting`, but moved to the separate PR: https://github.com/paritytech/polkadot-sdk/pull/4236

polkadot/xcm/xcm-builder/src/routing.rs
bkontur: `validate_xcm_nesting` changes are moved to the separate PR: https://github.com/paritytech/polkadot-sdk/pull/4236

polkadot/xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -157,15 +155,29 @@ benchmarks_instance_pallet! {
 		);
 		let sender_account_balance_before = T::TransactAsset::balance(&sender_account);
 
+		// generate holding and add possible required fees
```
franciscoaguirre, line 158: Where do you add the possible required fee?

polkadot/xcm/pallet-xcm-benchmarks/src/lib.rs
```diff
@@ -50,6 +50,8 @@ pub trait Config: frame_system::Config {
 	fn valid_destination() -> Result<Location, BenchmarkError>;
 
 	/// Worst case scenario for a holding account in this runtime.
+	/// - `depositable_count` specifies the count of assets we plan to add to the holding on top of
+	///   those generated by the `worst_case_holding` implementation.
```
franciscoaguirre, line 54: 😄

franciscoaguirre (approved):

polkadot/xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
bkontur: right bellow - `required fees = expected_assets_in_holding`
```
            for a in expected_assets_in_holding.into_inner() {
				holding.push(a);
			}
```

polkadot/xcm/pallet-xcm-benchmarks/src/lib.rs
bkontur: @franciscoaguirre :), any other suggestion?

polkadot/xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
bkontur: there was just `executor.set_holding(expected_assets_in_holding`) before, which could possibly remove previously `executor.set_holding(holding.clone()`

polkadot/xcm/xcm-builder/src/routing.rs
```diff
@@ -139,3 +139,69 @@ impl EnsureDelivery for Tuple {
...
+pub struct EnsureDecodableXcm<Inner>(sp_std::marker::PhantomData<Inner>);
+impl<Inner: SendXcm> SendXcm for EnsureDecodableXcm<Inner> {
+	type Ticket = Inner::Ticket;
+
+	fn validate(
+		destination: &mut Option<Location>,
+		message: &mut Option<Xcm<()>>,
+	) -> SendResult<Self::Ticket> {
+		if let Some(msg) = message {
+			if let Some(e) = Self::ensure_encode_decode(msg) {
```
acatangiu, line 160: Isn't this already covered by `validate_xcm_nesting()` in https://github.com/paritytech/polkadot-sdk/pull/4236?

It already does encode+decode (with limits) inside HRMP, DMP, UMP.

I believe ^ covers everything, no? Do we still need `EnsureDecodableXcm` wrapper for anything extra?

bkontur: when https://github.com/paritytech/polkadot-sdk/pull/4236 is merged, I plan to remove `fn ensure_encode_decode` and replace it with `VersionedXcm::<()>::from(msg.clone()).validate_xcm_nesting()` here.

yes, #4236 covers real cases for HRMP, DMP, UMP with real runtimes, but still does not cover cases for lots of testing/example/mocking stuff, where we use just "dummy" routers

bkontur: Important benchmarks fixes were moved to the #4236, so this PR contains now just `EnsureDecodableXcm` feature for XCM mocks/tests/samples where we don't use HRMP,UMP,DMP routers.

@acatangiu merge or close :D? I am fine with both :)

acatangiu (approved):

PR #4188, opened 2024-04-18
WithMaxSize

### LINKED ISSUE #4092: A way to annotate and enforce a max size of call parameters
#3749 essentially added a max size of the call parameter. It is done in such way that makes the DX miserable as we lost the type information.

We should be able to come up a way that don't sacrifice the type information but can still enforce the max parameter size.

Could be something like 

```rs
struct WithMaxSize<T: Parameter, S: Get<usize>>;
```

and custom implementation of encode/decode and helper to construct inner type with safety in mind.

xlc:
closes #4092
we may want to revert #3749 

TODOs:
- [ ] docs
- [ ] tests
- [ ] revert changes from #3749 either fully or partially

### DIFF
--- a/polkadot/xcm/xcm-builder/src/controller.rs
+++ b/polkadot/xcm/xcm-builder/src/controller.rs
@@ -48,6 +48,7 @@ pub trait ExecuteControllerWeightInfo {
 
 parameter_types! {
 	pub const MaxXcmEncodedSize: u32 = xcm::MAX_XCM_ENCODED_SIZE;
+	pub const MaxXcmEncodedSizeUsize: usize = xcm::MAX_XCM_ENCODED_SIZE as usize;
 }
 
 /// Execute an XCM locally, for a given origin.
--- a/substrate/frame/support/src/utils.rs
+++ b/substrate/frame/support/src/utils.rs
@@ -1,0 +1,91 @@
+// This file is part of Substrate.
+
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use codec::{Encode, Decode, Input, MaxEncodedLen};
+use frame_support_procedural::{EqNoBound, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound};
+use scale_info::TypeInfo;
+use sp_core::Get;
+
+use crate::Parameter;
+
+/// TODO: docs
+pub struct InputWithMaxLength<'a, I: Input, S: Get<usize>> {
+	input: &'a mut I,
+	len: usize,
+	_phantom: core::marker::PhantomData<S>,
+}
+
+impl<'a, I: Input, S: Get<usize>> InputWithMaxLength<'a, I, S> {
+	pub fn new(input: &'a mut I) -> Self {
+		Self { input, len: S::get(), _phantom: Default::default() }
+	}
+}
+
+impl<'a, I: Input, S: Get<usize>> Input for InputWithMaxLength<'a, I, S> {
+	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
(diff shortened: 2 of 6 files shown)


polkadot/xcm/xcm-builder/src/controller.rs
```diff
@@ -48,6 +48,7 @@ pub trait ExecuteControllerWeightInfo {
 
 parameter_types! {
 	pub const MaxXcmEncodedSize: u32 = xcm::MAX_XCM_ENCODED_SIZE;
+	pub const MaxXcmEncodedSizeUsize: usize = xcm::MAX_XCM_ENCODED_SIZE as usize;
```
xlc, line 51: Not sure what's the best way to handle this

xlc:
Screenshot 2024-04-18 at 10 01 17 PM

substrate/frame/support/src/utils.rs
```diff
@@ -0,0 +1,91 @@
...
+use frame_support_procedural::{EqNoBound, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound};
+use scale_info::TypeInfo;
+use sp_core::Get;
+
+use crate::Parameter;
+
+/// TODO: docs
+pub struct InputWithMaxLength<'a, I: Input, S: Get<usize>> {
+	input: &'a mut I,
+	len: usize,
```
bkchr, line 28:
```suggestion
	remaining_len: usize,
```

substrate/frame/support/src/utils.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl<'a, I: Input, S: Get<usize>> Input for InputWithMaxLength<'a, I, S> {
+	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
+		let len = self.input.remaining_len()?;
+		Ok(len.map(|l| l.min(self.len)))
+	}
+
+	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
+		let max_len = self.len.min(into.len());
```
bkchr, line 45: This can directly fail if `into.len() > self.len`

substrate/frame/support/src/utils.rs
```diff
@@ -0,0 +1,91 @@
...
+		self.len -= max_len;
+		self.input.read(&mut into[..max_len])?;
+		Ok(())
+	}
+}
+
+/// TODO: docs
+#[derive(Encode, EqNoBound, PartialEqNoBound, CloneNoBound, TypeInfo, RuntimeDebugNoBound)]
```
bkchr, line 53: IMO we should forward the `TypeInfo` of the inner type?

xlc:
Please let me know what should I do with the changes in pallet-xcm
1. Only keep the WithMaxSize change in this PR and fix pallet-xcm in next PR.
2. Fix send and execute and keep send_blob and execute_blob.
3. Completely revert #3749 and then fix send and execute

substrate/frame/support/src/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+	pub fn new(value: T) -> Result<Self, ()> {
+		if value.encoded_size() <= S::get() {
+			Ok(Self { value, _phantom: Default::default() })
+		} else {
+			Err(())
+		}
+	}
+
+	pub fn value(self) -> T {
+		self.value
```
ggwpez, line 73: Could make sense to impl the `Deref` trait and mark with `WrapperTypeEncode` and `WrapperTypeDecode`.

xlc:
I guess @franciscoaguirre you are the right person to ping?

acatangiu:
@xlc let me verify this fully fixes the issue please before putting in more effort here, can't promise to do it today, but latest Monday

substrate/frame/support/src/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+impl<'a, I: Input, S: Get<usize>> Input for InputWithMaxLength<'a, I, S> {
+	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
+		self.input.remaining_len().map(|l| l.map(|l| l.min(self.remaining_len)))
+	}
+
+	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
+		if into.len() > self.remaining_len {
+			return Err("Not enough data to fill buffer".into());
+		}
+		self.input.read(into)
```
bkchr, line 46:
```suggestion
		let res = self.input.read(into)?;
		self.remaining_len -= into.len();
		Ok(res)
```

We actually need to remove it from the remaining length :D

xlc: I really need some tests XD
and do we need to remove it before the read or it wouldn’t matter?

bkchr: I thought about it. If this returns an error, I would assume that we bubble that up any way and the decoding failed.

xlc:
lazy decode the Call will be a better way

PR #4192, opened 2024-04-18
Fix next_retry busy waiting on first retry

alexggh:
The `next_retry_time` gets populated when a request receives an error timeout or any other error, after thatn next_retry would check all requests in the queue returns the smallest one, which then gets used to move the main loop by creating a Delay
```
futures_timer::Delay::new(instant.saturating_duration_since(Instant::now())).await,
```

However when we retry a task for the first time we still keep it in the queue an mark it as in flight so its next_retry_time would be the oldest and it would be small than `now`, so the Delay will always triggers, so that would make the main loop essentially busy wait untill we received a response for the retry request.

Fix this by excluding the tasks that are already in-flight.

### DIFF
--- a/polkadot/node/network/statement-distribution/src/v2/requests.rs
+++ b/polkadot/node/network/statement-distribution/src/v2/requests.rs
@@ -288,7 +288,7 @@ impl RequestManager {
 	/// Returns an instant at which the next request to be retried will be ready.
 	pub fn next_retry_time(&mut self) -> Option<Instant> {
 		let mut next = None;
-		for (_id, request) in self.requests.iter().filter(|(_, request)| !request.in_flight) {
+		for (_id, request) in self.requests.iter().filter(|(_id, request)| !request.in_flight) {
 			if let Some(next_retry_time) = request.next_retry_time {
 				if next.map_or(true, |next| next_retry_time < next) {
 					next = Some(next_retry_time);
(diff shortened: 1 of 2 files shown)


eskimor (approved):

polkadot/node/network/statement-distribution/src/v2/requests.rs
```diff
@@ -553,7 +553,6 @@ impl UnhandledResponse {
 		let UnhandledResponse {
 			response: TaggedResponse { identifier, requested_peer, props, response },
 		} = self;
-
```
sandreim, line 556: why remove newline ?

polkadot/node/network/statement-distribution/src/v2/requests.rs
```diff
@@ -288,7 +288,7 @@ impl RequestManager {
 	/// Returns an instant at which the next request to be retried will be ready.
 	pub fn next_retry_time(&mut self) -> Option<Instant> {
 		let mut next = None;
-		for (_id, request) in &self.requests {
+		for (_id, request) in self.requests.iter().filter(|(_, request)| !request.in_flight) {
```
sandreim, line 291:
```suggestion
		for (_id, request) in self.requests.iter().filter(|(_id, request)| !request.in_flight) {
```

sandreim (approved):

polkadot/node/network/statement-distribution/src/v2/requests.rs
alexggh: Touched those while I was debugging, I shouldn't had included them in the commit, fixed now.

Issue #240, opened 2022-12-01
State Diff Recorder + Guard


ggwpez:
In migrations and tests it is sometimes desirable to know and restrict which storage keys change. Further it is nice to get a *diff* of all changes, if some unexpected changes happened.  
For example a pallet migration would whitelist all storage keys with a specific prefix as "expected to change". Then later on when the migration finishes it is possible to assert that only exactly these keys changed. Otherwise a *diff* is printed and it panics.  

Migration example of `Pallet` where `OtherPallet` can also change but nothing else:
```rust
let state_guard = StateDiffGuard::builder()
           .must_change(Prefix(Pallet::storage_prefix())
           .can_change(Prefix(OtherPallet::storage_prefix())
           .must_not_change(AnythingElse)
           .build();

// Run the migration with all the changes.

// Dropping will panic and print the *diff* iff there are unexpected changed.
sp_std::mem::drop(state_guard);
```

You could also try it without a builder and instead use single `assert!` statements, but then something like `AnythingElse` does not work anymore. The builder pattern helps here by holding a context.

cc @cor @kianenigma

ggwpez:
> I am guessing StorageNoopGuard be used as an inspiration, right?

Yea, but its rather small. It is a big design space, but starting small with a good foundation should be fine since it can then be extended later. Note that this is rather difficult to get right first try, but if you can do it then its great 😄

ggwpez:
> yeah, I see. I was thinking of writing something that simply checks if must change prefixes are actually changed. And then gradually tackle the `AnythingElse` and `can_change` cases, and the difference. This should be simple enough, no? Or maybe I am missing something?

Yes sounds fine. One way would be to take a snapshot of the storage and then compare it. Maybe creating a storage transaction could also work 🤔 yea just try it out a bit.

PR #4204, opened 2024-04-18
[FRAME] State Diff Guard

dastansam:
### Motivation

Adds a guard that prevents unexpected storage changes. `StateDiffGuard`:
- initializes map of prefixes to the mutation policy (can/not change)
- upon initialization takes the snapshot of the current state
- when dropped, iterates every storage key, gets its mutation policy, defaulting to `AnythingElse` if present. 
- applies the mutation policy, if present

### Example

Code below panics since migration is changing `SomeDoubleMap`
```rs
use frame_support::storage::StateDiffGuard;

pub struct UncheckedMigrateToV1<T: Config>(PhantomData<T>);

impl<T: Config> UncheckedOnRuntimeUpgrade for UncheckedMigrateToV1<T> {
        // migration logic here
        SomeDoubleMap::remove(_);

        Weight::default()
    }

    fn try_on_runtime_upgrade() -> Result<Weight, &'static str> {
        // Using StateDiffGuard to ensure storage items are correctly migrated
        let _guard = StateDiffGuard::builder()
            .must_change_if_exists(SomeMap::<T>::storage_info())
            .must_not_change(SomeDoubleMap::<T>::storage_info())
            .can_not_change(GuardSubject::AnythingElse)
            .build();

        // try runtime upgrade logic here
        let weight = Self::on_runtime_upgrade();

        // any other logic
        Ok(weight)
    }
}
```

part of #240 

### DIFF
--- a/substrate/frame/support/src/storage/state_diff_guard.rs
+++ b/substrate/frame/support/src/storage/state_diff_guard.rs
@@ -114,9 +114,9 @@
 //! In the example above, the guard will panic if any storage entry that doesn't match `SomeMap` or
 //! `SomeDoubleMap` prefixes is changed.
 
+use alloc::{collections::btree_map::BTreeMap, vec::Vec};
 use array_bytes::bytes2hex;
 use core::fmt::Debug;
-use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
 
 use crate::traits::StorageInfo;
 
@@ -335,6 +335,7 @@ impl Drop for StateDiffGuard {
 		let check_passed = self.apply_mutation_policy();
 
 		// No need to double panic, eg. inside a test assertion failure.
+		#[cfg(feature = "std")]
 		if sp_std::thread::panicking() {
 			return
 		}


substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,381 @@
...
+use super::storage_prefix;
+
+/// Raw storage key.
+type StorageKey = Vec<u8>;
+/// Raw storage value.
+type StorageValue = Vec<u8>;
+
+/// Storage prefix: pallet name and storage name.
+#[derive(Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+pub struct StoragePrefix {
```
ggwpez, line 133: We should already have something like this struct in frame-support. `trait StorageInstance` or `struct StorageInfo` for example.

substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,381 @@
...
+			sp_std::str::from_utf8(self.storage_name).unwrap_or("")
+		)
+	}
+}
+
+/// A guard that asserts that a specific storage prefix has been mutated or not.
+#[derive(Default, Debug)]
+pub struct StateDiffGuard {
+	// Storage prefixes that are expected to change.
+	whitelisted_prefixes: BTreeSet<StoragePrefix>,
```
ggwpez, line 162:
```suggestion
	must_change_prefixes: BTreeSet<StoragePrefix>,
```
The name here is not really expressing that they must *change*. It is also possible to make this more abstract by having a `Map<Prefix, Change>` where `Change` can be `{ Can | Must } x { Change | NotChange }`, but i think its fine for the beginning.

kianenigma:
This looks like a very good PR. @dastansam sorry for the delay, are you still around for finishing? @ggwpez @liamaharon and I can provide reviews.

kianenigma:
What I see missing is integrating this into `polkadot_sdk_docs::reference_docs::migrations`. It could be as simple as a basic mention.

dastansam:
hey @ggwpez, thanks for the updates. I am willing to finish this off if you have any other comments?

ggwpez (approved):

ggwpez:
Just wanted to improve the API a bit but I think it should be good now.

substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,612 @@
...
+	pub fn builder() -> StateDiffGuardBuilder {
+		StateDiffGuardBuilder { mutation_policy: BTreeMap::new() }
+	}
+
+	/// Take a snapshot of the current storage state.
+	fn read_state(&self) -> State {
+		let mut state = BTreeMap::new();
+
+		let mut previous_key = Vec::new();
+		while let Some(next) = sp_io::storage::next_key(&previous_key) {
```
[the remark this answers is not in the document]
dastansam: so that we can get the keys that were not meant to be changed. see this [line](https://github.com/dastansam/polkadot-sdk/blob/feat/state-diff-guard/substrate/frame/support/src/storage/state_diff_guard.rs#L266)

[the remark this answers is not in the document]
kianenigma: Not from the runtime, but from the node side we could probably do this more elegently by using the ProofRecording types that are used for state proofs. 

I think from the runtime there is no way other than (naively) reading the entire state pre and post operation, and compare them.

kianenigma (approved):

dastansam:
@bkchr one more review, please

substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,613 @@
...
+//!     UncheckedMigrateToV1<T>,
+//!     Pallet<T>,
+//!     <T as frame_system::Config>::DbWeight
+//! >;
+//! ```
+//!
+//! In the example above, the guard will panic if any storage entry that doesn't match `SomeMap` or
+//! `SomeDoubleMap` prefixes is changed.
+
+use alloc::{collections::btree_map::BTreeMap, vec::Vec};
```
bkchr, line 117:
```suggestion
use alloc::{collections::btree_map::BTreeMap, vec::Vec, vec};
```

substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,613 @@
...
+//! # How it works
+//!
+//! When the guard is instantiated, it takes the current storage state snapshot. When the guard is
+//! dropped, it reads every storage entry and compares it with the snapshot to collect any storage
+//! entry has been changed, removed or added. It then asserts that those mutations have the
+//! one of the whitelisted prefixes.
+//!
+//! # Example
+//!
+//! Use the `StateDiffGuard` in a migration:
```
bkchr, line 39: IMO this can not be used in any migration with a reasonable state, because of the behavior of reading the *entire state*. 

This would take ages.

substrate/frame/support/src/storage/state_diff_guard.rs
```diff
@@ -0,0 +1,472 @@
...
+		state_diff_guard
+	}
+}
+
+impl Drop for StateDiffGuard {
+	fn drop(&mut self) {
+		let check_passed = self.apply_mutation_policy();
+
+		// No need to double panic, eg. inside a test assertion failure.
+		if sp_std::thread::panicking() {
```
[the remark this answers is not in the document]
bkchr: You did not apply this.

bkchr (commented):
Given the current way it is implemented, we can never use this with any production chain. The time it would take to run the migrations would grow by a lot if every migration would read the entire state.

IMO if we really want to have this, we need to change the internals. The user needs to declare which storage entries are allowed to change. At the start we would read these keys from the state. (Could still hit a big map, but we can not really go better with what we have at hand) 

After the operation is finished, we could do the checks that the keys that should have changed, indeed have changed and also record the current keys/values. Now we could do a trick and run the `migration` and the checking in a storage transaction. After the migration and the checking succeeded, we would record the storage root and revert the transaction. Then we would  apply the changed keys again and check that the storage root matches. If the storage root doesn't match, something else was touched. Problem would be that we would not know which key changed.

PR #4208, opened 2024-04-19
Fixed GrandpaConsensusLogReader::find_scheduled_change

svyatonik:


bkontur (approved):

serban300 (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/45

acatangiu:
This needed a `prdoc` instead of "silent"

svyatonik:
> This needed a prdoc instead of "silent"

I've added it for 4209, but it is in a release branch, so copied here: https://github.com/paritytech/polkadot-sdk/pull/4239

kianenigma:
> so we haven't seen it on Rialto/Millau/Rococo/Westend.

We can use the `pallet-root-offences` that have been built for more or less this purpose to make sure on all testnets we have regularly slash a random set of validators. Even if not with this pallet, I think, we can easily scarp together the code to do this.

PR #4215, opened 2024-04-19
Add semver CI check 

Morganamilo:
This checks changed files against API surface changes against what the
prdoc says.

It will error if the detected semver change is greater than the one
listed in the prdoc. It will also error if any crates were touched but
not mentioned in the prdoc.


.github/workflows/check-semver.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  check-publish:
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/paritytech/ci-unified:bullseye-1.77.0-2024-04-10-v20240408
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          fetch-depth: 0
```
ggwpez, line 16: Complete history is probably too slow. 250 or so should be fine - otherwise the dev just has to rebase.

PR #4216, opened 2024-04-19
[subsystem-benchmark] Add approval-voting benchmark to CI

AndreiEres:


alvicsam (approved):

AndWeHaveAPlan (approved):

polkadot/node/core/approval-voting/benches/approval-voting-regression-bench.rs
```diff
@@ -0,0 +1,96 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+// TODO
```
sandreim, line 17: Is there any TODO left ?

polkadot/node/core/approval-voting/benches/approval-voting-regression-bench.rs
```diff
@@ -0,0 +1,96 @@
...
+//! approval-voting throughput test
+//!
+//! Approval Voting benchmark based on Kusama parameters and scale.
```
sandreim, line 20: These are future parameters, we're not there yet, but hopefully we'll be there soon.

polkadot/node/core/approval-voting/benches/approval-voting-regression-bench.rs
```diff
@@ -0,0 +1,96 @@
...
+};
+use std::io::Write;
+
+const BENCH_COUNT: usize = 10;
+
+fn main() -> Result<(), String> {
+	let mut messages = vec![];
+	let mut config = TestConfiguration::default();
+	// A single node effort roughly
+	config.n_cores = 100;
```
sandreim, line 41: Are we approving 100 candidates per block ? If so, then the comment above is not correct.

sandreim (approved):
Before merging we should be very confident it's not flaky.

polkadot/node/core/approval-voting/benches/approval-voting-regression-bench.rs
alexggh: No, it shouldn't be.

polkadot/node/core/approval-voting/benches/approval-voting-regression-bench.rs
alexggh: Yes, this would generate assignments and approvals for 100 occupied core  each block

alexggh (approved):

PR #4221, opened 2024-04-19
pallet_broker: Let `start_sales` calculate and request the correct core count

bkchr:


eskimor (commented):
Comment on start_sales need to be fixed as well, as it now actually does request the core count.

eskimor (approved):
Thanks!

seadanda (approved):

Issue #3968, opened 2024-04-03
`fungibles` methods should not consume `AssetId`


liamaharon:
Methods such as https://github.com/paritytech/polkadot-sdk/blob/cdacfb9d33710093d67e96cd6767dba0243ff450/substrate/frame/support/src/traits/tokens/fungibles/regular.rs#L84 should take a ref rather than consume the AssetId.

It was probably written this way back when AssetIds were small, but that's no longer the case with the increasing usage of XCM Locations as AssetIds.

All relevant `fungibles` methods should be updated, and unnecessary cloning of large types removed (e.g. in the `asset-conversion` pallet).

bkontur:
Related issue for XCM `Location`: https://github.com/paritytech/polkadot-sdk/issues/3038

dianpopa:
should be fixed by https://github.com/paritytech/polkadot-sdk/pull/4224

Nathy-bajo:
@liamaharon Took the liberty of fixing this again here https://github.com/paritytech/polkadot-sdk/pull/11787
Please review

ggwpez:
We cannot break all Asset interfaces for a single copy reduction. *IF* it is really a problem (and not just an assumed one) then we should add new interfaces and slowly migrate over.  

But i did not see any benchmark so far that identifies asset clones as problematic. We should instead accept this shortcoming and think harder the next time when we design interfaces.

PR #4224, opened 2024-04-20
`fungibles` methods do not consume `AssetId`

dianpopa:
`Fungibles` methods now take `AssetId` by reference instead of value where the value is not needed.

`pallet-assets` methods now also take `AssetId` by reference instead of value where the value is not needed.

This avoids a lot of cloning across the codebase.

There are more objects and traits that could do the same, but this PR already blew up in size as I was changing the most relevant ones. This "small" change is actually pretty big..

fixes https://github.com/paritytech/polkadot-sdk/issues/3968


substrate/frame/support/src/traits/tokens/fungibles/hold.rs
```diff
@@ -39,6 +39,7 @@ use sp_runtime::{DispatchError, DispatchResult, Saturating, TokenError};
 
 use super::*;
 
+// TODO
```
ggwpez, line 42: ?

cumulus/parachains/runtimes/test-utils/src/lib.rs
```diff
@@ -534,9 +534,9 @@ pub fn assert_metadata<Fungibles, AccountId>(
 	Fungibles: frame_support::traits::fungibles::metadata::Inspect<AccountId>
 		+ frame_support::traits::fungibles::Inspect<AccountId>,
 {
-	assert_eq!(Fungibles::name(asset_id.clone().into()), Vec::from(expected_name),);
-	assert_eq!(Fungibles::symbol(asset_id.clone().into()), Vec::from(expected_symbol),);
-	assert_eq!(Fungibles::decimals(asset_id.into()), expected_decimals);
+	assert_eq!(Fungibles::name(&asset_id.clone().into()), Vec::from(expected_name),);
+	assert_eq!(Fungibles::symbol(&asset_id.clone().into()), Vec::from(expected_symbol),);
```
liamaharon, line 538: Possible to remove the `.clones` in scenarios like this or?

substrate/frame/support/src/traits/tokens/fungible/union_of.rs
```diff
@@ -115,80 +115,80 @@ impl<
 	type AssetId = AssetKind;
 	type Balance = Left::Balance;
 
-	fn total_issuance(asset: Self::AssetId) -> Self::Balance {
-		match Criterion::convert(asset) {
+	fn total_issuance(asset: &Self::AssetId) -> Self::Balance {
+		match Criterion::convert(asset.clone()) {
```
liamaharon, line 119: Hmm I wonder if `Criterion::convert` should also take a ref?

liamaharon (commented):
Thanks for helping with this!

gilescope:
Context: This is good clean up. It's because `AssetId` used to be `Copy` but we needed it to not be (we wanted to use `MultiLocation` as an AssetId but certainly for xcm v2 `MultiLocation` wasn't `Copy`).

cumulus/parachains/runtimes/test-utils/src/lib.rs
dianpopa: All these tests helper functions provide `asset_id` as `impl Into<Fungibles::AssetId> + Clone`. I would need to change the API of these functions, then a lot of test callsites.
I was thinking it's not worth the effort and the noise since these are only tests.

substrate/frame/support/src/traits/tokens/fungible/union_of.rs
dianpopa: I initially tried, but got lost in the sea of errors. There are so many generic types and trait bounds used here, hard to follow.

I will give it another try.

dianpopa:
PR has _a lot_ of changes, but the vast majority is test code because of the changed inner functions APIs. The "main" API changes are small I think.

substrate/frame/support/src/traits/tokens/fungible/union_of.rs
dianpopa: Can't make it work.. got as far as https://github.com/dianpopa/polkadot-sdk/commit/090c74efa31d71604124ed3e660aba74e288d944, maybe someone can help or do it in another PR.

dianpopa:
@liamaharon @gilescope is there anything else required here?

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-rococo/src/tests/reserve_transfer.rs
```diff
@@ -723,7 +723,10 @@ fn reserve_transfer_native_asset_from_system_para_to_para() {
 	let sender_balance_before = test.sender.balance;
 	let receiver_assets_before = PenpalA::execute_with(|| {
 		type ForeignAssets = <PenpalA as PenpalAPallet>::ForeignAssets;
-		<ForeignAssets as Inspect<_>>::balance(system_para_native_asset_location.clone(), &receiver)
+		<ForeignAssets as Inspect<_>>::balance(
+			&system_para_native_asset_location.clone(),
```
liamaharon, line 727: Does this need `.clone()`?

liamaharon (approved):
Looks good, maybe some redundant `.clones` still in there?

dianpopa:
@liamaharon is this fix for #3968 still wanted?
I already put in considerable time, I will only continue if actually needed..
If yes, can you help with reviewers?

franciscoaguirre:
I agree this is a good fix and a tedious one (the amount of compiler errors, number of files), we should push for it to get merged before more conflicts pile up. @liamaharon thoughts?

bkchr:
Looks like the pr is stale.

PR #4228, opened 2024-04-21
Define and implement `OnPollStatusChange`

### LINKED ISSUE #4227: `OnPollStatusChange` for `pallet-referenda`

### Motivation

For pallets that use `pallet_referenda` as `Polling` implementor might be benefitted with having knowledge of the change of a poll status (i.e. for releasing locks of `Completed` polls)

### Request

Implement a trait that allows pallets to listen for the poll status change event and act accordingly.

### Solution

Such trait might have the following structure, where `poll_status` is a reference to alllow implement for tuples without `Clone` constraints:

```rs
pub trait OnPollStatusChange<Tally, Moment, Class, Index: Copy> {
	fn on_poll_status_change(index: Index, poll_status: &PollStatus<Tally, Moment, Class>);
}
```

### Are you willing to help with this request?

Yes!

pandres95:
Closes #4227


substrate/frame/referenda/src/lib.rs
```diff
@@ -199,6 +199,13 @@ pub mod pallet {
 			+ Debug
 			+ TypeInfo
 			+ MaxEncodedLen;
+		// A Hook thaet announces when a poll changes status
```
liamaharon, line 202:
```suggestion
			+ MaxEncodedLen;
			
		/// A Hook that announces when a poll changes status
```

substrate/frame/support/src/traits/voting.rs
```diff
@@ -126,3 +127,16 @@ pub trait Polling<Tally> {
...
+}
+
+#[cfg_attr(all(not(feature = "tuples-96"), not(feature = "tuples-128")), impl_for_tuples(64))]
+#[cfg_attr(all(feature = "tuples-96", not(feature = "tuples-128")), impl_for_tuples(96))]
+#[cfg_attr(feature = "tuples-128", impl_for_tuples(128))]
+impl<Tally, Moment, Class, Index: Copy> OnPollStatusChange<Tally, Moment, Class, Index> for Tuple {
+	fn on_poll_status_change(index: Index, poll_status: &PollStatus<Tally, Moment, Class>) {
+		for_tuples!( #( Tuple::on_poll_status_change(index, poll_status); )* );
+	}
+}
```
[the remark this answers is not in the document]
pandres95: I took it from [`frame_support::traits::hooks`](https://github.com/paritytech/polkadot-sdk/blob/92a348f57deed44789511df73d3fbbbcb58d98cb/substrate/frame/support/src/traits/hooks.rs#L340) 😁

substrate/frame/referenda/src/lib.rs
```diff
@@ -499,6 +506,12 @@ pub mod pallet {
 				alarm: Self::set_alarm(nudge_call, now.saturating_add(T::UndecidingTimeout::get())),
 			};
 			ReferendumInfoFor::<T, I>::insert(index, ReferendumInfo::Ongoing(status));
+			let (status, track) = Self::as_ongoing(index)
+				.expect("has been added to ReferendumInfoFor as ongoing; qed");
```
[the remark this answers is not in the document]
pandres95: 🤯 You're right! The `status` that's here is not actually a status, but an instance of `TallyOf<T>`.

What I can do to avoid this redundant read (since the tally has just been created anyways, is to input a copy of the created tally).

pandres95:
@liamaharon just added a test [here](https://github.com/paritytech/polkadot-sdk/pull/4228/commits/ccd2d73fa9e11a6d94a5f7431a06730a5a7b3a28)

substrate/frame/referenda/src/mock.rs
```diff
@@ -55,6 +55,43 @@ impl Contains<RuntimeCall> for BaseFilter {
...
+#[frame_support::storage_alias]
+type CounterForListNodes<T: crate::Config<I>, I: 'static> = StorageMap<
+	crate::Pallet<T, I>,
+	frame_support::pallet_prelude::Blake2_128Concat,
+	PollStatusChangeType,
+	u32,
+	frame_support::pallet_prelude::ValueQuery,
+>;
+
+pub struct CountPollStatusChanges;
```
liamaharon, line 75:
```suggestion
pub struct StatusChangeConsumer;
```

substrate/frame/referenda/src/mock.rs
```diff
@@ -55,6 +55,43 @@ impl Contains<RuntimeCall> for BaseFilter {
...
+// Tests that hooks for OnPollStatusChange are being observed
+#[derive(Clone, Encode, Decode, MaxEncodedLen)]
+pub enum PollStatusChangeType {
+	None,
+	Ongoing,
+	Completed,
+}
+
+#[frame_support::storage_alias]
+type CounterForListNodes<T: crate::Config<I>, I: 'static> = StorageMap<
```
liamaharon, line 67:
```suggestion
type StatusChangeCounterStorage<T: crate::Config<I>, I: 'static> = StorageMap<
```

prdoc/pr_4228.prdoc
```diff
@@ -0,0 +1,27 @@
...
+        type OnPollStatusChange = ();
+        // ...
+      }
+      ```
+
+crates:
+  - name: frame-support
+    bump: minor
+  - name: pallet-referenda
+    bump: minor
```
liamaharon, line 27: breaking change, due to new mandatory config item 
```suggestion
    bump: major
```

substrate/frame/referenda/src/tests.rs
```diff
@@ -67,6 +67,9 @@ fn basic_happy_path_works() {
 		run_to(13);
 		// #10: Proposal should be executed.
 		assert_eq!(Balances::free_balance(&42), 1);
+
+		assert_eq!(CountPollStatusChanges::count_for(PollStatusChangeType::Ongoing), 4);
+		assert_eq!(CountPollStatusChanges::count_for(PollStatusChangeType::Completed), 1);
```
liamaharon: I suggest some dedicated tests for the hook, one for each time the hook is suggested to be triggered

substrate/frame/referenda/src/lib.rs
```diff
@@ -200,6 +200,14 @@ pub mod pallet {
 			+ TypeInfo
 			+ MaxEncodedLen;
 
+		// A Hook thaet announces when a poll changes status
```
liamaharon, line 203:
```suggestion
		// Subscribers to notify when the referenda poll status changes.
```

liamaharon (commented):
Generally I think this looks good :)

However I have some concerns about the general idea of pallets calling into each other, it seems potentially dangerous from PoV perspective and bug prone. 

I'd like to get @bkchr 's opinion on whether this is something we want, and if it is a good approach to achieving it before approving.

substrate/frame/referenda/src/tests.rs
pandres95: Changed it to a test called `on_poll_status_change_works`.

kianenigma (commented):
Complicated pallet such a s staking provide similar interfaces, and in general I am not against this if it solves a reasonable issue. What is the end goal here? 

About granularity, possibly you can consider one trait encapsulating all possible changes that for now has only one function. Then, adding new hooks in the future will not be a breaking change. 

Also, be aware that such hooks impose strict benchmarking requirements. In that the code behind the hooks must also be benchmarked. It would be good to keep that in mind.

olanod:
> Complicated pallet such a s staking provide similar interfaces, and in general I am not against this if it solves a reasonable issue. What is the end goal here?

For Kreivo we have a couple of use cases at the moment and more will likely follow, a simple one is related to what Liam mentions, have a better UX(always the priority!) for cleaning up storage in another pallet, we have a communities-pallet(DAOs) with voting logic that should clean votes of completed/failed/timed-out polls. Another one a bit more interesting is for community referenda to trigger actions mainly outside of Kreivo, e.g. a DAO wants to participate in the relay chain's OpenGov, we could send 3 different XCM to the relay on behalf of the plurality based on the outcome of a local poll, if it was approved send an *aye* vote to the relevant poll in Kusama/Polkadot, if local poll failed, send *nay* and if it timed-out send *abstain*.

> About granularity, possibly you can consider one trait encapsulating all possible changes that for now has only one function. Then, adding new hooks in the future will not be a breaking change.

That's the plan and was my initial feedback when discussing this with Pablo, instead of having a hook for different things we just care about the (end?) status of the poll.

> Also, be aware that such hooks impose strict benchmarking requirements. In that the code behind the hooks must also be benchmarked. It would be good to keep that in mind.

One implementation of the trait I'm thinking about is one that is more passive and uses `on_idle` to *eventually* execute actions with "leftover weight" assuming they are not critical and can wait to be executed one or more blocks in the future. In that sense, maybe I would suggest name that better reflects that passiveness like `notify_poll_status_change` ? it would suggest *don't mind me, finish what your are doing, I'm cheap and won't take your precious weight*

Polkadot-Forum:

https://forum.polkadot.network/t/decoded-2024-sponsorship-for-active-community-members/7654/71

Issue #4206, opened 2024-04-19
System chains: all kusama system chain permissionless collators were kicked out of the active collator list


stakeworld:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

All the kusama system chain permissionless collators were deleted from the [collators lists](https://polkadot.js.org/apps/?rpc=wss://ksm-rpc.stakeworld.io/assethub#/collators) somewere in the last hours. It is not yet clear what caused this; did collators stop producing blocks and were kicked out (why did they stop producing blocks?) is there a relation to the new runtime? After being kicked out the required 50 ksm stays locked, so re-entering is only possible with new funds. 

For example [assethub](https://polkadot.js.org/apps/?rpc=wss://ksm-rpc.stakeworld.io/assethub#/collators):
![image](https://github.com/paritytech/polkadot-sdk/assets/106294967/81fcebf1-3a2a-471e-8b6b-3a4a05c32f90)

stakeworld:
When I look a the logs (kusama asset hub) the last block was around 02h in the night (UTC+2), after that no errors; it just stopped producing  new blocks.

> Apr 19 02:01:48 polkadot-parachain\[2186672\]: 2024-04-19 02:01:48 \[Parachain\] 🙌 Starting consensus session on top of parent 0x156748f4adab3d0c09b25090d00315a7de7ef09d9f21c87a70e553032331>
> Apr 19 02:01:48 polkadot-parachain\[2186672\]: 2024-04-19 02:01:48 \[Parachain\] 🎁 Prepared block for proposing at 6811192 (1 ms) \[hash: 0x976ecaf77af06b744d8c9c56cbf719350c0c1ce5fc69169cd9>
> Apr 19 02:01:48 polkadot-parachain\[2186672\]: 2024-04-19 02:01:48 \[Parachain\] 🔖 Pre-sealed block for proposal at 6811192. Hash now 0xe3aec173b84f3a03da30830d89d92a48ed9fda3958338a25a846d>
> Apr 19 02:01:48 polkadot-parachain\[2186672\]: 2024-04-19 02:01:48 \[Parachain\] ✨ Imported #6811192 (0xe3ae…42dc)
[...]

PR #1340, opened 2023-08-31
Add simple collator election mechanism

georgepisaltu:
Fixes https://github.com/paritytech/polkadot-sdk/issues/106

Port of cumulus PR https://github.com/paritytech/cumulus/pull/2960

This PR adds the ability to bid for collator slots even after the max number of collators have already registered. This eliminates the first come, first served mechanism that was in place before.

Key changes:
- added `update_bond` extrinsic to allow registered candidates to adjust their bonds in order to dynamically control their bids
- added `take_candidate_slot` extrinsic to try to replace an already existing candidate by bidding more than them
- candidates are now kept in a sorted list in the pallet storage, where the top `DesiredCandidates` out of `MaxCandidates` candidates in the list will be selected by the session pallet as collators

# Checklist

- [ ] My PR includes a detailed description as outlined in the "Description" section above
- [ ] My PR follows the [labeling requirements](https://github.com/paritytech/polkadot-sdk/blob/master/docs/CONTRIBUTING.md#process) of this project (at minimum one label for `T` required)
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] If this PR alters any external APIs or interfaces used by Polkadot, the corresponding Polkadot PR is ready as well as the corresponding Cumulus PR (optional)

PR #4229, opened 2024-04-21
Fix Stuck Collator Funds

joepetrowski:
Fixes https://github.com/paritytech/polkadot-sdk/issues/4206

In #1340 one of the storage types was changed from `Candidates` to `CandidateList`. Since the actual key includes the hash of this value, all of the candidates stored here are (a) "missing" and (b) unable to unreserve their candidacy bond.

This migration kills the storage values and refunds the deposit held for each candidate.

### DIFF
--- a/cumulus/pallets/collator-selection/src/migration.rs
+++ b/cumulus/pallets/collator-selection/src/migration.rs
@@ -55,7 +55,7 @@ pub mod v2 {
 	impl<T: Config + pallet_balances::Config> UncheckedOnRuntimeUpgrade for UncheckedMigrationToV2<T> {
 		fn on_runtime_upgrade() -> Weight {
 			let mut weight = Weight::zero();
-			let mut count: u32 = 0;
+			let mut count: u64 = 0;
 			// candidates who exist under the old `Candidates` key
 			let candidates = Candidates::<T>::take();
 
@@ -86,7 +86,7 @@ pub mod v2 {
 					count.saturating_inc();
 				}
 				weight.saturating_accrue(
-					<<T as pallet_balances::Config>::WeightInfo as pallet_balances::WeightInfo>::force_unreserve(),
+					<<T as pallet_balances::Config>::WeightInfo as pallet_balances::WeightInfo>::force_unreserve() * count,
 				);
 			}
 
--- a/cumulus/pallets/collator-selection/src/migration.rs
+++ b/cumulus/pallets/collator-selection/src/migration.rs
@@ -29,6 +29,8 @@ pub mod v2 {
 		traits::{Currency, ReservableCurrency},
 	};
 	use sp_runtime::traits::{Saturating, Zero};
+	#[cfg(feature = "try-runtime")]
+	use sp_std::vec::Vec;
 
 	#[storage_alias]
 	pub type Candidates<T: Config> = StorageValue<
--- a/cumulus/pallets/collator-selection/src/migration.rs
+++ b/cumulus/pallets/collator-selection/src/migration.rs
@@ -81,7 +81,7 @@ pub mod v2 {
 					if err > Zero::zero() {
 						log::info!(
 							target: LOG_TARGET,
-							"{:?} balance was unable to be unreserved from {}",
+							"{:?} balance was unable to be unreserved from {:?}",
 							err, &candidate.who,
 						);
 					}
@@ -222,14 +222,17 @@ mod tests {
 			Balances::make_free_balance_be(&one, 100u64);
 			Balances::make_free_balance_be(&two, 100u64);
 			Balances::make_free_balance_be(&three, 100u64);
+
 			// Reservations: 10 for the "old" candidacy and 10 for the "new"
 			Balances::reserve(&one, 10u64).unwrap(); // old
 			Balances::reserve(&two, 20u64).unwrap(); // old + new
 			Balances::reserve(&three, 10u64).unwrap(); // new
+
 			// Candidate info
 			let candidate_one = CandidateInfo { who: one, deposit };
 			let candidate_two = CandidateInfo { who: two, deposit };
 			let candidate_three = CandidateInfo { who: three, deposit };
+
 			// Storage lists
 			let bounded_candidates =
 				BoundedVec::<CandidateInfo<u64, u64>, ConstU32<20>>::try_from(vec![
@@ -243,9 +246,11 @@ mod tests {
 					candidate_three.clone(),
 				])
 				.expect("it works");
+
 			// Set storage
 			Candidates::<Test>::put(bounded_candidates);
 			CandidateList::<Test>::put(bounded_candidate_list.clone());
+
 			// Sanity check
 			assert_eq!(Balances::free_balance(one), 90);
 			assert_eq!(Balances::free_balance(two), 80);
@@ -281,12 +286,15 @@ mod tests {
 			// Set balance to 100
 			Balances::make_free_balance_be(&one, 100u64);
 			Balances::make_free_balance_be(&two, 100u64);
+
 			// Reservations
 			Balances::reserve(&one, 10u64).unwrap(); // old
 			Balances::reserve(&two, 10u64).unwrap(); // old
+
 			// Candidate info
 			let candidate_one = CandidateInfo { who: one, deposit };
 			let candidate_two = CandidateInfo { who: two, deposit };
+
 			// Storage lists
 			let bounded_candidates =
 				BoundedVec::<CandidateInfo<u64, u64>, ConstU32<20>>::try_from(vec![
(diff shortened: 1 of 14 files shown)


cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -90,3 +196,54 @@ pub mod v1 {
...
+		mock::{new_test_ext, Balances, Test},
+	};
+	use frame_support::{
+		traits::{Currency, ReservableCurrency, StorageVersion},
+		BoundedVec,
+	};
+	use sp_runtime::traits::ConstU32;
+
+	#[test]
+	fn migrate_to_v2() {
```
georgepisaltu, line 214: We should do a simple test with an empty `CandidateList`, just to be sure.

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -90,3 +196,54 @@ pub mod v1 {
...
+
+			let who = 1u64;
+
+			// Set balance to 100
+			Balances::make_free_balance_be(&who, 100u64);
+			// Reserve 20 -> 10 for the "old" candidacy and 10 for the "new"
+			Balances::reserve(&who, 20u64).unwrap();
+			// Candidate info
+			let candidate = CandidateInfo { who, deposit: 10u64 };
+			let bounded_candidates =
```
georgepisaltu, line 227: Let's also add 2 more candidates with the same sanity checks:
- one present only in `Candidates`
- one present only in `CandidateList`

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+				}
+
+				StorageVersion::new(2).put::<Pallet<T>>();
+				log::info!(
+					target: LOG_TARGET,
+					"Unreserved locked bond of {} candidates, upgraded storage to version 2",
+					count,
+				);
+				// 1 read/write for storage version
+				weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
```
georgepisaltu, line 96: `StorageVersion` r/w
`Candidates` r/w (`take`)
`CandidateList` r

```suggestion
				weight.saturating_accrue(T::DbWeight::get().reads_writes(3, 2));
```

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+	>;
+
+	/// Migrate to V2.
+	///
+	/// Note: This should have been in https://github.com/paritytech/polkadot-sdk/pull/1340.
+	pub struct UncheckedMigrationToV2<T>(sp_std::marker::PhantomData<T>);
+	impl<T: Config> UncheckedOnRuntimeUpgrade for UncheckedMigrationToV2<T> {
+		fn on_runtime_upgrade() -> Weight {
+			let on_chain_version = Pallet::<T>::on_chain_storage_version();
+			if on_chain_version == 1 {
```
ggwpez, line 60: No need to check this anymore as its done by the wrapping `VersionedMigration`.

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+					// just put the candidates into the new storage item.
+					CandidateList::<T>::put(&candidates);
+					// 1 write for the new list
+					weight.saturating_accrue(T::DbWeight::get().reads_writes(0, 1));
+				} else {
+					// Oops, the runtime upgraded without the migration. There are new candidates in
+					// `CandidateList`. So, let's just refund the old ones and assume they have
+					// already started participating in the new system.
+					for candidate in candidates {
+						let _err = T::Currency::unreserve(&candidate.who, candidate.deposit);
```
ggwpez, line 82:
```suggestion
						let _err = T::Currency::unreserve(&candidate.who, candidate.deposit).defensive();
```
Just so we see it in the console.

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+						count.saturating_inc();
+					}
+					// TODO: set each accrue to weight of `unreserve`
+					weight.saturating_accrue(T::DbWeight::get().reads_writes(0, count as u64));
+				}
+
+				StorageVersion::new(2).put::<Pallet<T>>();
```
ggwpez, line 89:
```suggestion
```
The wrapped `VersionedMigration` does this.

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade(_number_of_candidates: Vec<u8>) -> Result<(), sp_runtime::DispatchError> {
+			let new_number_of_candidates = Candidates::<T>::get().to_vec().len();
+			assert_eq!(
+				new_number_of_candidates, 0 as usize,
+				"after migration, the candidates map should be empty"
+			);
+
+			let on_chain_version = Pallet::<T>::on_chain_storage_version();
+			frame_support::ensure!(on_chain_version >= 1, "must_upgrade");
```
ggwpez, line 122:
```suggestion
```

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,115 @@
...
+		}
+
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade(_number_of_candidates: Vec<u8>) -> Result<(), sp_runtime::DispatchError> {
+			let new_number_of_candidates = Candidates::<T>::get().to_vec().len();
+			assert_eq!(
+				new_number_of_candidates, 0 as usize,
+				"after migration, the candidates map should be empty"
+			);
+
```
ggwpez, line 120: Maybe check the number of entries in `CandidateList`?

joepetrowski: I think the only check we can add here is that it's greater than zero, although even that is questionable. If someone was using this pallet with only Invulnerables, this migration would be a noop and `CandidateList` would be empty.

Likewise if someone applies this "late" then it will just kill `Candidates`, but it doesn't mean `CandidateList` has the same members.

ggwpez: Ah okay, yea then its fine.

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,107 @@
...
+				// 1 write for the new list
+				weight.saturating_accrue(T::DbWeight::get().reads_writes(0, 1));
+			} else {
+				// Oops, the runtime upgraded without the migration. There are new candidates in
+				// `CandidateList`. So, let's just refund the old ones and assume they have already
+				// started participating in the new system.
+				for candidate in candidates {
+					let err = T::Currency::unreserve(&candidate.who, candidate.deposit);
+					if err > Zero::zero() {
+						log::info!(
```
ggwpez, line 80:
```suggestion
						log::error!(
```
Not sure what log level the try-runtime CI runs with, but error should show.

ggwpez (approved):

cumulus/pallets/collator-selection/src/migration.rs
```diff
@@ -17,9 +17,107 @@
...
+							target: LOG_TARGET,
+							"{:?} balance was unable to be unreserved from {:?}",
+							err, &candidate.who,
+						);
+					}
+					count.saturating_inc();
+				}
+				weight.saturating_accrue(
+					<<T as pallet_balances::Config>::WeightInfo as pallet_balances::WeightInfo>::force_unreserve() * count,
```
joepetrowski, line 89:
```suggestion
					<<T as pallet_balances::Config>::WeightInfo as pallet_balances::WeightInfo>::force_unreserve().saturating_mul(count.into()),
```

joepetrowski:
bot fmt

georgepisaltu (approved):

Polkadot-Forum:

wischli:
Can someone please explain to me why this fix is only part of SDK V1.11.0 and not any previous version? The error was introduced in v1.7.0 and should be backported to that version! 

On Centrifuge Chain, the v1 migration leads to collator candidates being removed from the session whitelist (i.e. `queued keyes` is reduced to the invulnerables). It wasn't a smooth debugging process as part of our effort to upgrade to v1.7.2 because the v2 migration is simply not part of the codebase until SDK v1.11.0.

joepetrowski:
> Can someone please explain to me why this fix is only part of SDK V1.11.0 and not any previous version? The error was introduced in v1.7.0 and should be backported to that version!
> 
> On Centrifuge Chain, the v1 migration leads to collator candidates being removed from the session whitelist (i.e. `queued keyes` is reduced to the invulnerables). It wasn't a smooth debugging process as part of our effort to upgrade to v1.7.2 because the v2 migration is simply not part of the codebase until SDK v1.11.0.

It was backported to 1.7: https://github.com/paritytech/polkadot-sdk/pull/4288

PR #4232, opened 2024-04-22
[dnm] ci: check westend migration

alvicsam:
...


liamaharon:
Yeah this is still failing, due to another error now 🙈

@gpestana is aware and looking into it.

Issue #4163, opened 2024-04-17
contracts: Make host function benchmarks architecture independent


athei:
## The problem
We don't benchmark the host functions callable by contracts of pallet-contracts in isolation. We measure the execution of a wasm module (with wasmi) which just calls the host function `t` times. We then do some linear regression to determine how expensive it is to add one call. This means that in addition to the host function call itself, it measures all the wasm instructions to load the parameters for the host function onto the stack as this grows with `t`, too.

This is not optimal for two reason: It adds noise to the host function benchmarks as we just benchmarking a small thing in a big operation (executing one host function vs. calling a whole contract). Secondly, it makes our benchmarks architecture dependent since we need to generate wasm modules for each's host function benchmark. The [upcoming support of another architecture](https://github.com/paritytech/polkadot-sdk/issues/115) makes this inconvenient.

## Proposal

Instead of benchmarking the whole module we can call the host function directly. This will probably entail changing the macro that defines our host functions to (additionally) generate some code so that we can call host functions directly. Instead of only registering them with wasmi as it is right now. I think this will mostly get most of the job done. Except for memory access from within host functions which depends on the code being run within wasmi. One solution would be to make [`Runtime`](https://github.com/paritytech/polkadot-sdk/blob/e81322bc3e8192b536067fed3ef9e20f2752c376/substrate/frame/contracts/src/wasm/runtime.rs#L375-L380) generic over the memory access so we can mock it from the benchmarks. Something we would need to do for PolkaVM anyways in order to support both byte codes.

This will cut down the architecture dependent benchmarks  down to those:
- Benchmark the time it takes wasmi to execute one instruction (already exists)
- Benchmark the time it takes wasmi to execute an empty host function (no work on the host, no args, no return value)
- Benchmark the time it takes to access `n` bytes of memory in wasmi as we mocked this in our test

PR #4233, opened 2024-04-22
Contracts: Rework host fn benchmarks

pgherveou:
fix https://github.com/paritytech/polkadot-sdk/issues/4163

This PR does the following:
Update to pallet-contracts-proc-macro: 
- Parse #[cfg] so we can add a dummy noop host function for benchmark.
- Generate BenchEnv::<host_fn> so we can call host functions directly in the benchmark.
- Add the weight of the noop host function before calling the host function itself

Update benchmarks:
- Update all host function benchmark, a host function benchmark now simply call the host function, instead of invoking the function n times from within a contract.
- Refactor RuntimeCosts & Schedule, for most host functions, we can now use the generated weight function directly instead of computing the diff with the cost! macro

```rust
// Before
#[benchmark(pov_mode = Measured)]
fn seal_input(r: Linear<0, API_BENCHMARK_RUNS>) {
    let code = WasmModule::<T>::from(ModuleDefinition {
        memory: Some(ImportedMemory::max::<T>()),
        imported_functions: vec![ImportedFunction {
            module: "seal0",
            name: "seal_input",
            params: vec![ValueType::I32, ValueType::I32],
            return_type: None,
        }],
        data_segments: vec![DataSegment { offset: 0, value: 0u32.to_le_bytes().to_vec() }],
        call_body: Some(body::repeated(
            r,
            &[
                Instruction::I32Const(4), // ptr where to store output
                Instruction::I32Const(0), // ptr to length
                Instruction::Call(0),
            ],
        )),
        ..Default::default()
    });

    call_builder!(func, code);

    let res;
    #[block]
    {
        res = func.call();
    }
    assert_eq!(res.did_revert(), false);
}
```

```rust
// After
fn seal_input(n: Linear<0, { code::max_pages::<T>() * 64 * 1024 - 4 }>) {
    let mut setup = CallSetup::<T>::default();
    let (mut ext, _) = setup.ext();
    let mut runtime = crate::wasm::Runtime::new(&mut ext, vec![42u8; n as usize]);
    let mut memory = memory!(n.to_le_bytes(), vec![0u8; n as usize],);
    let result;
    #[block]
    {
        result = BenchEnv::seal0_input(&mut runtime, &mut memory, 4, 0)
    }
    assert_ok!(result);
    assert_eq!(&memory[4..], &vec![42u8; n as usize]);
}
``` 

[Weights compare](https://weights.tasty.limo/compare?unit=weight&ignore_errors=true&threshold=10&method=asymptotic&repo=polkadot-sdk&old=master&new=pg%2Frework-host-benchs&path_pattern=substrate%2Fframe%2Fcontracts%2Fsrc%2Fweights.rs%2Cpolkadot%2Fruntime%2F*%2Fsrc%2Fweights%2F**%2F*.rs%2Cpolkadot%2Fbridges%2Fmodules%2F*%2Fsrc%2Fweights.rs%2Ccumulus%2F**%2Fweights%2F*.rs%2Ccumulus%2F**%2Fweights%2Fxcm%2F*.rs%2Ccumulus%2F**%2Fsrc%2Fweights.rs)


substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -629,6 +629,27 @@ mod benchmarks {
...
+		let (mut ext, _) = setup.ext();
+		let mut runtime = crate::wasm::Runtime::new(&mut ext, vec![]);
+		let mut memory = [0u8; 36];
+		memory[0..4].copy_from_slice(&32u32.to_le_bytes());
+
+		let result;
+		#[block]
+		{
+			result = crate::wasm::BenchEnv::seal0_caller(&mut runtime, &mut memory, 4, 0);
+		}
```
pgherveou, line 644: > One solution would be to make [Runtime](https://github.com/paritytech/polkadot-sdk/blob/e81322bc3e8192b536067fed3ef9e20f2752c376/substrate/frame/contracts/src/wasm/runtime.rs#L375-L380) generic over the memory access so we can mock it from the benchmarks.

@athei do we really need to make memory generic? Doesn't this approach above just work?

pgherveou: ah nvm I guess we need it so that we abstract gas counting for when we read and write to memory sandbox

pgherveou:

pgherveou:

athei (commented):
At first glance it looks like weights go down. But when looking at the formular we can see that the weights go up for many host functions:
Screenshot 2024-05-07 at 12 50 02

Before it was 280k and not it is 680k. They kind of should go down because no we are running the host functions directly.

pgherveou:
> At first glance it looks like weights go down. But when looking at the formular we can see that the weights go up for many host functions: Screenshot 2024-05-07 at 12 50 02
> 
> https://weights.tasty.limo/compare?unit=weight&ignore_errors=true&threshold=10&method=asymptotic&repo=polkadot-sdk&old=master&new=pg%2Frework-host-benchs&path_pattern=substrate%2Fframe%2Fcontracts%2Fsrc%2Fweights.rs%2Cpolkadot%2Fruntime%2F*%2Fsrc%2Fweights%2F**%2F*.rs%2Cpolkadot%2Fbridges%2Fmodules%2F*%2Fsrc%2Fweights.rs%2Ccumulus%2F**%2Fweights%2F*.rs%2Ccumulus%2F**%2Fweights%2Fxcm%2F*.rs%2Ccumulus%2F**%2Fsrc%2Fweights.rs
> 
> Before it was 280k and not it is 680k. They kind of should go down because no we are running the host functions directly.

Not sure if it should go down, but it should be roughly equal for sure.
To be correct you would need to add as well the noop_host_fn weight which adds another 70k, since the previous measured time was (fn time + base cost of calling a host fn) repeated r times.

athei:
Yes it should be roughly the same. Even slightly less after the rework because we still add the `noop_host_fn` onto the new weight as you pointed out.

Could be attributed to caching behaviour. Before we ran the same code over and over again. That could make each of them faster.

pgherveou:
Yes exactly for host fn where there is io, there are some trie node cached that amortized the cost but for other like gas_left where there are no PoV this must be something else 🤔

athei (changes requested):
For the benchmarks which touch storage (`seal_call, `seal_instantiate`, `seal_delegate_call`, ...) we need to whitelist the callers address and probably other keys.

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -256,85 +268,58 @@ impl<T: Config> Token<T> for RuntimeCosts {
...
+			BlockNumber => T::WeightInfo::seal_block_number(),
+			Now => T::WeightInfo::seal_now(),
+			WeightToFee => T::WeightInfo::seal_weight_to_fee(),
+			Terminate => T::WeightInfo::seal_terminate(),
+			Random => T::WeightInfo::seal_random(),
+			DepositEvent { num_topic, len } => T::WeightInfo::seal_deposit_event(num_topic, len),
+			DebugMessage(len) => T::WeightInfo::seal_debug_message(len),
+			SetStorage { new_bytes, old_bytes } =>
+				T::WeightInfo::seal_set_storage_per_new_byte(new_bytes)
+					.saturating_add(T::WeightInfo::seal_set_storage_per_old_byte(old_bytes)),
```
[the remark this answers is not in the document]
pgherveou: right good point

pgherveou:
> For the benchmarks which touch storage (`seal_call, `seal_instantiate`, `seal_delegate_call`, ...) we need to whitelist the callers address and probably other keys.

Added that to the host fn benchmark setup.
Storage access for the contract account_id and  ContractInfo (that is touched when we add / pop a frame) are already taken into account with the base call extrinsic benchmark.

Everything else should be fine (will double check once the benchmark are updated)

I have also removed from the kitchensink in this PR #4489 the dynamic parameters that should not be used with pallet-contracts until there is a proper way to not double count them in the host fn benchmarks

```rs
		// Whitelist contract account, as it is already accounted for in the call benchmark
		benchmarking::add_to_whitelist(
			frame_system::Account::<T>::hashed_key_for(&contract.account_id).into(),
		);

		// Whitelist the contract's contractInfo as it is already accounted for in the call
		// benchmark
		benchmarking::add_to_whitelist(
			crate::ContractInfoOf::<T>::hashed_key_for(&contract.account_id).into(),
		);
```

pgherveou:

pgherveou:

pgherveou:
bot cancel 1-621accc2-6fcb-4497-a21d-db259c223c49

pgherveou:

athei:
Screenshot 2024-05-17 at 12 50 08

`seal_call` still has **3** extra reads and **1** extra write from before.

pgherveou:
> Screenshot 2024-05-17 at 12 50 08
> `seal_call` still has **3** extra reads and **1** extra write from before.

I traced the execution this seems correct, let me check the old equation

athei (commented):
Yes you are right. The storage accesses look correct. They were too low before.  One last thing. There should be only one access to `Account` for `seal_instantiate` to bring it in line with `seal_call`.

Screenshot 2024-05-17 at 14 49 08

pgherveou:
> Yes you are right. The storage accesses look correct. They were too low before. One last thing. There should be only one access to `Account` for `seal_instantiate` to bring it in line with `seal_call`.
> Screenshot 2024-05-17 at 14 49 08

Yes reason for that is that the bench_call setup is using the contract caller as origin, I am changing it to the root_contract for delegate/call/instantiate, so that the origin account get whitelisted

pgherveou:

athei (commented):
Something still doesn't add up with the `seal_instantiate` benchmark. The base weight is super low (lower than `seal_call`)  and then a lot is added in case of transfer:
Screenshot 2024-05-18 at 07 31 52

pgherveou:
> Something still doesn't add up with the `seal_instantiate` benchmark. The base weight is super low (lower than `seal_call`) and then a lot is added in case of transfer: Screenshot 2024-05-18 at 07 31 52

mm interesting, will look into this tomorrow.
Relaunching one more time to check if this is consistent

pgherveou:

pgherveou:
> Something still doesn't add up with the `seal_instantiate` benchmark. The base weight is super low (lower than `seal_call`) and then a lot is added in case of transfer: Screenshot 2024-05-18 at 07 31 52

@athei Actually I think it does adds up
We are just forgetting the RocksDbWeight (which confusingly are actually adding ref_time and not pov)
on `seal_call` they are attached to `t` since  we only touch accounts if we do a transfer. 
On `instantiate` it happens all the time and does not depend on `t`

so if you count these 3 lines on `seal_call` 

```rust
.saturating_add(Weight::from_parts(44_066_869, 0).saturating_mul(t.into()))
.saturating_add(RocksDbWeight::get().reads((1_u64).saturating_mul(t.into())))
.saturating_add(RocksDbWeight::get().writes((1_u64).saturating_mul(t.into())))

// 44_000_000 + 25_000_000 + 100_000_000 ~=  169_000_000
```

You roughly get what we have in `seal_instantiate`
```rust
.saturating_add(Weight::from_parts(159_867_027, 0).saturating_mul(t.into()))
```

athei (approved):
You are right. I wasn't accounting for the reads and writes which are added unconditionally for `seal_instantiate`.

Robbepop (approved):
This change is very reasonable and I am glad that it has been done. The new resulting numbers look reasonable and should be more precise than before. Thanks @pgherveou !

smiasojed (approved):

xermicus (approved):
LGTM!

PR #4234, opened 2024-04-22
metaspan boot nodes for coretime-[westend, kusama]

dcolley:


prdoc/pr_4234.prdoc
```diff
@@ -0,0 +1,11 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
```
bkchr, line 1: Not required for this change.

bkchr (approved):

prdoc/pr_4234.prdoc
bkchr: (The entire prdoc, please remove)

dcolley: prdoc removed

bkontur (approved):

PR #4235, opened 2024-04-22
[xcm] Assets: sort after `prepend_with`

bkontur:
Adds sorting to the XCM Assets' `prepend_with`, which could modify the order of `AssetId` locations.

Relates to: https://github.com/paritytech/polkadot-sdk/pull/4186 (the same fix for `reanchored`)
Part of: https://github.com/paritytech/polkadot-sdk/pull/2129

### DIFF
--- a/polkadot/xcm/src/v4/asset.rs
+++ b/polkadot/xcm/src/v4/asset.rs
@@ -1041,7 +1041,7 @@ mod tests {
 		assert!(assets.using_encoded(|mut enc| Assets::decode(&mut enc).map(|_| ())).is_ok());
 
 		assert!(assets.reanchor(&dest, &reanchor_context).is_ok());
-		assert_eq!(assets, vec![asset_2_reanchored, asset_3_reanchored, asset_1_reanchored].into());
+		assert_eq!(assets.0, vec![asset_2_reanchored, asset_3_reanchored, asset_1_reanchored]);
 
 		// decoding respects limits and sorting
 		assert!(assets.using_encoded(|mut enc| Assets::decode(&mut enc).map(|_| ())).is_ok());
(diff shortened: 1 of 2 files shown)


polkadot/xcm/src/v4/asset.rs
```diff
@@ -1035,8 +1037,61 @@ mod tests {
...
+		);
+
+		// `From` impl does sorting.
+		let mut assets: Assets = vec![asset_1, asset_2, asset_3].into();
+		// decoding respects limits and sorting
+		assert!(assets.using_encoded(|mut enc| Assets::decode(&mut enc).map(|_| ())).is_ok());
+
+		// let's do `prepend_with`
+		assert!(assets.prepend_with(&prefix).is_ok());
+		assert_eq!(assets, vec![asset_2_prepended, asset_1_prepended, asset_3_prepended].into());
```
serban300, line 1091: `Assets::From::<Vec>`, sorts the `vec`. So I think it would be better to do:

```
		let mut assets: Assets = vec![asset_1.clone(), asset_2.clone(), asset_3.clone()].into();
		// decoding respects limits and sorting
		assert!(assets.using_encoded(|mut enc| Assets::decode(&mut enc).map(|_| ())).is_ok());
		assert_eq!(assets.0, vec![asset_1, asset_2, asset_3]);

		// let's do `prepend_with`
		assert!(assets.prepend_with(&prefix).is_ok());
		assert_eq!(assets.0, vec![asset_2_prepended, asset_1_prepended, asset_3_prepended]);
```

Also, can you fix the other tests related to sorting as well please ? I didn't realize this when I wrote them.

serban300 (approved):

franciscoaguirre (approved):

acatangiu (approved):

Issue #336, opened 2021-02-23
Avoid burning tokens in pallet-staking and other runtime modules


sorpaas:
cc @thiolliere @xlc, rel paritytech/substrate#8168 

Currently, pallet-staking runtime module, in some cases, will burn reward payout tokens (via unclaimed reward, lower than existence deposit, or other methods). As discussed in [another PR](https://github.com/paritytech/substrate/pull/8168#issuecomment-783978233), this is still a point of confusion.

I'm not convinced this issue is of concern (as I think burning a small amount of tokens should be totally fine). In the mean time, this issue tries to document all methods right now that tokens might be burned through pallet-staking reward payouts, and aims at reaching one of the two outcomes:

* So that we consider current burning tokens of no issue in pallet-staking, and properly document it to avoid future confusions.
* If we really consider burning tokens to be an issue, audit all cases and fix them.

---

Currently known ways of burning tokens:

* Staking module's payout period can be set to longer to the bonding period (and it's indeed the case on Polkadot). So a validator/nominator can generate some rewards and leave them unclaimed. Then unbond, and claim rewards when all the bonding is cleared. The reward will be burned but not sent to the payee, because `Self::bond` at that time will return None.
* One can also generate a large number of nominators, and make it so that each day's reward is less than the existence balance. The nominator then set the reward destination to a new account and claim rewards every day. The reward will be burned, because while the sum might be large enough, each individual payment will be less than the existence balance.
* Another really simple method to burn tokens right now is just -- not to claim the reward. After the payout deadline ends, the reward will be removed from staking module, and burned.

rossbulat:
This issue has been inactive for a long time and was already marked stale. The original discussion also seems inconclusive on whether the behavior is something we still want to change, rather than document or accept.

Unless someone can confirm that this is still relevant and actionable today, I think it would be reasonable to close this as stale/unowned. It can always be reopened, or replaced with a narrower issue, if there is a current concrete problem to address.

sigurpol:
Closing the issue - with `pallet-dap`, we have now (on Polkadot) dap and not staking responsible of minting and DAP is also responsible to collect burns / fees / unclaimed rewards / slashes ... (we really don't want burns with capped issuance) .

PR #4236, opened 2024-04-22
Add `validate_xcm_nesting` to the `ParentAsUmp` and `ChildParachainRouter`

bkontur:
This PR:
- moves `validate_xcm_nesting` from `XcmpQueue` into the `VersionedXcm`
- adds `validate_xcm_nesting` to the `ParentAsUmp`
- adds `validate_xcm_nesting` to the `ChildParachainRouter`

Based on discussion [here](https://github.com/paritytech/polkadot-sdk/pull/4186#discussion_r1571344270) and/or [here](https://github.com/paritytech/polkadot-sdk/pull/4186#discussion_r1572076666) and/or [here]()

## Question/TODO

- [x] To the [comment](https://github.com/paritytech/polkadot-sdk/pull/4186#discussion_r1572072295) - Why was `validate_xcm_nesting` added just to the `XcmpQueue` router and nowhere else? What kind of problem `MAX_XCM_DECODE_DEPTH` is solving? (see [comment](https://github.com/paritytech/polkadot-sdk/pull/4236#discussion_r1574605191))


polkadot/runtime/common/src/xcm_sender.rs
```diff
@@ -119,7 +119,9 @@ where
 		let config = configuration::ActiveConfig::<T>::get();
 		let para = id.into();
 		let price = P::price_for_delivery(para, &xcm);
-		let blob = W::wrap_version(&d, xcm).map_err(|()| DestinationUnsupported)?.encode();
+		let versioned_xcm = W::wrap_version(&d, xcm).map_err(|()| DestinationUnsupported)?;
+		versioned_xcm.validate_xcm_nesting().map_err(|()| ExceedsMaxMessageSize)?;
+		let blob = versioned_xcm.encode();
```
ggwpez, line 124: So why is it being added to the router if having it in the pallet is better? Just for validation to return an error earlier?

cumulus/primitives/utility/src/lib.rs
```diff
@@ -526,6 +529,8 @@ impl<
 mod test_xcm_router {
 	use super::*;
 	use cumulus_primitives_core::UpwardMessage;
+	use frame_support::assert_ok;
```
ggwpez, line 532: You can look into srlabs finding [#336](https://github.com/paritytech-secops/srlabs_findings/issues/336).  
It is to prevent the sending of messages that will surely not decode, because of nesting limits.

polkadot/runtime/common/src/xcm_sender.rs
bkontur: > ...if having it in the pallet is better

I don't quite understand what you mean by this, could you elaborate more? Or what do you suggest here?

---
You added `validate_xcm_nesting()` to the XcmpQueue for `SendXcm::validate`, and I just added it exactly like that to the `ParentAsUmp` and `ChildParachainRouter`.
Thank you for pointing me to the that sr_labs issue, now I see it was really about HRMP. So, if we don't have this issue for UMP/DMP, we can close this PR.

My point/question is that, without this PR, we don't allow "too-much-nested" XCM between paras, why do we allow to "too-much-nested" XCM from para to relaychain or from relaychain to para? Again, if it is ok, we can close this PR :)

Also there are other PRs which allows/relaxes XCM execution e.g.: https://github.com/polkadot-fellows/runtimes/pull/261, so maybe better safe then sorry?

E.g. can we have a scenario like this? Like the relaychain allows to execute XCM which send "too-much-nested" XCM to the parachain (**DMP works**), which would result re-sending the "too-much-nested" XCM to the parachain (**HRMP fails**)? Is that ok?


@ggwpez what do you suggest with this PR? :)

franciscoaguirre (approved):

polkadot/runtime/common/src/xcm_sender.rs
ggwpez: >Like the relaychain allows to execute XCM which send "too-much-nested" XCM to the parachain

"too-much-nested" <=> fails-to-decode.  
An XCM that is too deeply nested will fail to `Decode::decode`. It is undecodable, that is the whole issue.  
I think we should also not send such messages over any other protocol, including DMP/UMP. I just did not think about this when writing it for HRMP.

acatangiu:
> moves validate_xcm_nesting from XcmpQueue into the VersionedXcm

this is not accurate anymore, it is also done inside `XcmpQueue` validations

PR #4240, opened 2024-04-22
Use default branch of `psvm` when synchronizing templates

rzadp:
We cannot lock to a specific version of `psvm`, because we will need to keep it up-to-date - each release currently requires a change in `psvm` such as [this one](https://github.com/paritytech/psvm/pull/2/files).

There is no `stable` branch in `psvm` repo or anything so using the default branch.


rzadp:
Actually I looked again and looks like [as of now](https://github.com/paritytech/psvm/pull/11), the `psvm` no longer needs to be updated with each new release, it will fetch appropriate versions online.

We still need to use more recent version than the one currently in the workflow.

AndWeHaveAPlan (approved):

alvicsam (approved):

PR #4244, opened 2024-04-22
Add PoV-reclaim enablement guide to polkadot-sdk-docs

skunert:
This adds a small guide on how to prepare your parachain for pov-reclaim usage.


docs/sdk/src/guides/enable_pov_reclaim.rs
```diff
@@ -0,0 +1,81 @@
...
+//! > set up correctly.
+//!
+//! ## 3. Add the SignedExtension to your runtime
+//!
+//! In your runtime, you will find a list of SignedExtensions.
+//! To enable the reclaiming,
+//! add [StorageWeightReclaim](`cumulus_primitives_storage_weight_reclaim::StorageWeightReclaim`)
+//! to that list. The extension will check the size of the storage proof before and after an
+//! extrinsic execution. It reclaims the difference between the calculated size and the benchmarked
+//! size.
```
ggwpez, line 65: One note please if it matters where to add it in the list of SEs. I assume at the end, but could be mentioned.

docs/sdk/src/guides/enable_pov_reclaim.rs
```diff
@@ -0,0 +1,81 @@
...
+//! proof-size consumption.
+//!
+//!
+//! # How to enable PoV reclaim
+//! ## 1. Add the host function to your node
+//!
+//! To reclaim excess storage weight, a parachain runtime needs the
+//! ability to fetch the size of the storage proof from the node. The reclaim
+//! mechanism uses the
+//! [storage_proof_size](`cumulus_primitives_proof_size_hostfunction::storage_proof_size`)
```
skunert, line 24: Hmm when I build locally, this link and other cumulus links below are broken, but I am unable to figure out why :thinking:. Anyone got an idea? They look correct to me.

docs/sdk/src/guides/enable_pov_reclaim.rs
skunert: I was thinking about it during implementation and came to the conclusion that the order should not be important. But maybe I failed. Why do you think it should be at the end?

ggwpez: >To answer my own question, when it is placed at the end it will be able to operate more accurately because storage changes performed by the other SignedExtensions can be reflected as well.

Yes that is what i thought. Having it before that could miss some of the following consumed weight. So i guess end is best?  

I would still like to use it to actually refund the overspent fee payment. Since we will need to bump the PoV weights a bit, and then refunding that amount would be nice.

skunert: > I would still like to use it to actually refund the overspent fee payment. Since we will need to bump the PoV weights a bit, and then refunding that amount would be nice. I can try to do it.

Slightly offtopic for this PR, but do you have an idea how we could deal with the MEV implications? Imagine multiple tx load some expensive storage item. Only the first executed would be charged since the other ones are not growing the proof size at all. Collators could benefit from reordering the transactions, which we wanted to avoid.

dmitry-markin:
CC @s0me0ne-unkn0wn

alexggh (approved):

docs/sdk/src/guides/enable_pov_reclaim.rs
ggwpez: > Only the first executed would be charged since the other ones are not growing the proof size at all. Collators could benefit from reordering the transactions, which we wanted to avoid.

Yea... i mentioned MEV and secondary market at my last Sub0 talk as consequence of this weight fee reclaim.  
I think it needs a wider discussion. There are probably some tricks that can be done, did not check how other ecosystems solve this. cc @kianenigma

sandreim (approved):

templates/parachain/node/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = { features = ["derive"], workspace = true, default-features = true }
 jsonrpsee = { version = "0.22", features = ["server"] }
 futures = "0.3.28"
 serde_json = { workspace = true, default-features = true }
+docify = "0.2.8"
```
bkchr, line 27: This is a template used by external people. We should clearly not add this docify stuff here.

templates/parachain/runtime/Cargo.toml
```diff
@@ -28,6 +28,7 @@ scale-info = { version = "2.11.1", default-features = false, features = [
 	"derive",
 ] }
 smallvec = "1.11.0"
+docify = "0.2.8"
```
bkchr, line 31: Same

docs/sdk/src/guides/enable_pov_reclaim.rs
```diff
@@ -0,0 +1,83 @@
...
+//! # How to enable PoV reclaim
+//! ## 1. Add the host function to your node
+//!
+//! To reclaim excess storage weight, a parachain runtime needs the
+//! ability to fetch the size of the storage proof from the node. The reclaim
+//! mechanism uses the
+//! [`storage_proof_size`](cumulus_primitives_proof_size_hostfunction::storage_proof_size)
+//! hostfunction for this purpose. For convenience, cumulus provides
+//! [`ParachainHostFunctions`](cumulus_client_service::ParachainHostFunctions), a set of
+//! hostfunctions typically used by cumulus-based parachains. In the binary crate of your parachain,
```
bkchr, line 27:
```suggestion
//! host functions typically used by cumulus-based parachains. In the binary crate of your parachain,
```

docs/sdk/src/guides/enable_pov_reclaim.rs
```diff
@@ -0,0 +1,83 @@
...
+//! find the instantiation of the [`WasmExecutor`](sc_executor::WasmExecutor) and set the correct
+//! generic type.
+//!
+//! This example from the parachain-template shows a type definition that includes the correct
+//! hostfunctions.
```
bkchr, line 32:
```suggestion
//! host functions.
```

docs/sdk/src/guides/enable_pov_reclaim.rs
```diff
@@ -0,0 +1,83 @@
+//! This guide will teach you how to enable storage weight reclaiming for a parachain. The
```
kianenigma, line 1: Only think I would add here is possbly a link to the original PR or any other relevant resources, maybe your sub0 talk.

kianenigma (approved):

kianenigma:
cc @SBalaguer this page can be useful for parachains to intergrate this.

Issue #3688, opened 2024-03-14
[Frame Core] Construct Runtime V2 follow-ups


gupnik:
https://github.com/paritytech/polkadot-sdk/pull/1378 introduced the new runtime macro. This issue tracks the follow-up items:

- [ ] Abstract away the executive similar to https://github.com/paritytech/substrate/pull/14742
- [ ] Either support `cfg` attrs or throw an error if they should not be used CC: @ggwpez 
- [ ] Remove the `experimental` flag once its tested in Rococo/Westend

kianenigma:
> Abstract away the executive similar to https://github.com/paritytech/substrate/pull/14742

I am not sure about this. but it is a good candidate. 

Alternatively, what I see missing rn is mostly that `AllPallets` is used in `Type Executive` but is never declared. Bringing it in scope might help.

I have also pointed out before that this macro seems to do some automatic `use *` that makes it not intuitive to use, and not in par with `frame::pallet` macro. Would be great if you can log this as well. 

Here's my mentality: 

You import `polkadot-sdk-frame`. This docs for this crate should be informative. 

```rust
#[frame::pallet] 
pub mod pallet1 {
    use frame::pallet_prelude::*;
} 

#[frame::pallet] 
pub mod pallet2 {
    use frame::pallet_prelude::*;
}

#[cfg(test)]
pub mod tests {
    pub use frame::testing_prelude::*; 

    #[frame::runtime]
    pub mod runtime {
        use frame::runtime_prelude::*;
    }
}
```

So the experience and behavior you would get from all of these should be as similar as possible.

gupnik:
> I have also pointed out before that this macro seems to do some automatic `use *` that makes it not intuitive to use, and not in par with `frame::pallet` macro. Would be great if you can log this as well.

It actually maintains the same scope as with the previous CRV1. I can scope those under a `mod` and expose all of the generated types such as `RuntimeEvent` by default, but I am not sure if that adds a lot of value?

>  use frame::runtime_prelude::*;

What does this `runtime_prelude` provide?

kianenigma:
> What does this runtime_prelude provide?

Types that are needed for a typical runtime, the existing code should to some extent describe this.

PR #4249, opened 2024-04-23
Moves runtime macro out of experimental flag

gupnik:
Step in https://github.com/paritytech/polkadot-sdk/issues/3688

Now that the `runtime` macro (Construct Runtime V2) has been successfully deployed on Westend, this PR moves it out of the experimental feature flag  and makes it generally available for runtime devs.


prdoc/pr_4249.prdoc
```diff
@@ -0,0 +1,17 @@
...
+    description: |
+      Now that the runtime macro (Construct Runtime V2) has been successfully deployed on Westend, 
+      this PR moves it out of the experimental feature flag and makes it generally available for
+      runtime devs.
+
+crates:
+  - name: frame-support
+    bump: patch
+  - name: frame-support-procedural
+    bump: patch
```
bkchr, line 17:
```suggestion
    bump: minor
  - name: frame-support-procedural
    bump: minor
```

bkchr (commented):
Generally not sure if just because it was deployed on one network, it should be taken out of the experimental flag? Given this argument, we would have never needed to add it there in the first place, because no downstream user had even the time to try this out.

gupnik:
> Generally not sure if just because it was deployed on one network, it should be taken out of the experimental flag? Given this argument, we would have never needed to add it there in the first place, because no downstream user had even the time to try this out.

Yeah, makes sense. @ggwpez had suggested to put in under the experimental flag to make sure that the metadata is not impacted before and after the change. I felt that this has been validated now but happy to wait in case you feel something else is needed first.

kianenigma (commented):
Which test networks are using this now? 

I have less of a strong opinion on letting it remain experimental, probably would opt for the safer option of letting it be there for now.

Instead, I would put the focus on making sure this is used as much as possible within our codebase/templates.

gupnik:
> Which test networks are using this now?

Currently on westend.

> Instead, I would put the focus on making sure this is used as much as possible within our codebase/templates.

Yeah, my hope was to add it in all test runtimes, but the experimental flag makes it difficult.

kianenigma (approved):
I agree that the `feature = experimental` has very limited use in this context, maybe we should not use it at all. But that is OT and this should not remain blocked.

PR #4251, opened 2024-04-23
MBM `try-runtime` support

liamaharon:
# MBM try-runtime support

This MR adds support to the try-runtime trait such that the try-runtime-CLI will be able to support MBM testing [here](https://github.com/paritytech/try-runtime-cli/pull/90). It mainly adds two feature-gated hooks to the `SteppedMigration` hook to facilitate testing. These hooks are named `pre_upgrade` and `post_upgrade` and have the same signature and implications as for single-block migrations.

## Integration

To make use of this in your Multi-Block-Migration, just implement the two new hooks and test pre- and post-conditions in them:

```rust
fn pre_upgrade() -> Result<Vec<u8>, frame_support::sp_runtime::TryRuntimeError> {
	// ...
}

fn post_upgrade(prev: Vec<u8>) -> Result<(), frame_support::sp_runtime::TryRuntimeError> {
    // ...
}
```

You may return an error or panic in these functions to indicate failure. This will then show up in the try-runtime-CLI and can be used in CI for testing.

Changes:
- Adds `try-runtime` gated methods `pre_upgrade` and `post_upgrade` on `SteppedMigration`
- Adds `try-runtime` gated methods `nth_pre_upgrade` and `nth_post_upgrade` on `SteppedMigrations`
- Modifies `pallet_migrations` implementation to run pre_upgrade and post_upgrade steps at the appropriate times, and panic in the event of migration failure.


substrate/frame/executive/src/tests.rs
```diff
@@ -460,6 +460,11 @@ impl MultiStepMigrator for MockedModeGetter {
 	fn step() -> Weight {
 		Weight::zero()
 	}
+
+	#[cfg(feature = "try-runtime")]
+	fn try_mbms() -> Result<(), sp_runtime::TryRuntimeError> {
```
ggwpez, line 465: We used this approach a lot in the past, but i think it can cause some difficult to debug issues.  
Maybe we can try this time to use an extension trait, like `trait MultiStepMigratorTryRuntime : MultiStepMigrator ` that has this function, and then requiring that trait instead.  
You think it could work?

substrate/frame/migrations/src/lib.rs
```diff
@@ -785,4 +785,100 @@ impl<T: Config> MultiStepMigrator for Pallet<T> {
...
+			let migration_index = match Cursor::<T>::get() {
+				Some(MigrationCursor::Active(active)) => active.index,
+				_ => unreachable!("migration must be active"),
+			};
+
+			// Simulate a step.
+			log::debug!("Block {}", System::<T>::block_number());
+			System::<T>::set_block_number(System::<T>::block_number() + 1u32.into());
+			System::<T>::on_initialize(System::<T>::block_number());
+			crate::Pallet::<T>::on_initialize(System::<T>::block_number());
```
ggwpez, line 864: Not sure if we need these. The `step` could also be called multiple times per block i think.

substrate/frame/migrations/src/lib.rs
```diff
@@ -785,4 +785,100 @@ impl<T: Config> MultiStepMigrator for Pallet<T> {
...
+			Self::step();
+			<crate::Pallet<T>>::on_finalize(System::<T>::block_number());
+			<frame_system::Pallet<T>>::on_finalize(System::<T>::block_number());
+
+			// Check events for `UpgradeCompleted` or `UpgradeFailed`.
+			for e in System::<T>::events().iter() {
```
ggwpez, line 870:
```suggestion
			for e in frame_system::Events::<T>::take().into_iter() {
```
Should make it a bit faster to not hold old events.

substrate/frame/executive/src/tests.rs
liamaharon: What kinds of issues are you referring to and what use case is there? 

My initial reaction is I'm not sure about it, because `MultiStepMigratorTryRuntime` would just be `MultiStepMigrator` with the `try-runtime` feature enabled.

substrate/frame/support/src/migrations.rs
```diff
@@ -649,6 +685,16 @@ impl SteppedMigrations for () {
 		None
 	}
 
+	#[cfg(feature = "try-runtime")]
+	fn nth_pre_upgrade(_n: u32) -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
+		Ok(Vec::new())
```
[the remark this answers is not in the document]
liamaharon: Good point, I'll update it

substrate/frame/support/src/migrations.rs
```diff
@@ -416,6 +416,23 @@ pub trait SteppedMigration {
 		})
 		.map_err(|()| SteppedMigrationError::Failed)?
 	}
+
+	/// Executed prior to the migration starting.
+	///
+	/// Returns some bytes which are passed into `post_upgrade` after the migration is completed.
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
```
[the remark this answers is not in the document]
liamaharon: I'm not sure about creating an extension trait that just adds some feature gated methods, what's the use case?

Generally I like to keep code simple with minimal indirection, only adding complexity and indirection when there is tangible benefit.

bkchr:
I remember that you asked somewhere which approach to take. I would take the approach of building multiple blocks until the MBMs are finished. I know that this is more complicated, but this directly checks that your runtime can handle this properly and also that you stay always in the POV limits. 

(Didn't check the pr yet :D)

liamaharon:
> I remember that you asked somewhere which approach to take. I would take the approach of building multiple blocks until the MBMs are finished. I know that this is more complicated, but this directly checks that your runtime can handle this properly and also that you stay always in the POV limits.
> 
> (Didn't check the pr yet :D)

Here's the discussion https://github.com/paritytech/try-runtime-cli/issues/17#issuecomment-2061746572

I missed that Kian also supports building proper blocks. Although more involved, I will proceed with that approach :)

substrate/frame/examples/multi-block-migrations/src/migrations/v1/mod.rs
```diff
@@ -115,4 +121,29 @@ impl<T: Config, W: weights::WeightInfo> SteppedMigration for LazyMigrationV1<T,
...
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(prev: Vec<u8>) -> Result<(), frame_support::sp_runtime::TryRuntimeError> {
+		use codec::Decode;
+
+		// Check the state of the storage after the migration.
+		let prev_map = BTreeMap::<u32, u32>::decode(&mut &prev[..])
+			.expect("Failed to decode the previous storage state");
+
+		for (key, value) in MyMap::<T>::iter() {
```
ggwpez, line 141: This only checks that all migrated values got migrated correctly, but not that all values got migrated.  
I know that its an example, but still 😄

substrate/frame/migrations/src/lib.rs
```diff
@@ -69,22 +69,22 @@
 //! either be [`MigrationCursor::Active`] or [`MigrationCursor::Stuck`]. In the active case it
 //! points to the currently active migration and stores its inner cursor. The inner cursor can then
 //! be used by the migration to store its inner state and advance. Each time when the migration
-//! returns `Some(cursor)`, it signals the pallet that it is not done yet.  
+//! returns `Some(cursor)`, it signals the pallet that it is not done yet.
```
ggwpez, line 72: Its markdown, so this changes the formatting into one large blog:  

Screenshot 2024-05-29 at 22 16 30

substrate/frame/migrations/src/lib.rs
```diff
@@ -260,13 +260,26 @@ pub type IdentifierOf<T> = BoundedVec<u8, <T as Config>::IdentifierMaxLen>;
...
+#[impl_trait_for_tuples::impl_for_tuples(30)]
 pub trait MockedMigrations: SteppedMigrations {
 	/// The migration should fail after `n` steps.
 	fn set_fail_after(n: u32);
 	/// The migration should succeed after `n` steps.
 	fn set_success_after(n: u32);
 }
 
+#[cfg(feature = "try-runtime")]
+/// Wrapper for pre-upgrade bytes, allowing us to impl MEL on it.
```
ggwpez, line 272: Can you mention that it is basically a mock? Does it have to be public? Nobody should be using this really.

substrate/frame/migrations/src/lib.rs
```diff
@@ -378,6 +391,12 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type Historic<T: Config> = StorageMap<_, Twox64Concat, IdentifierOf<T>, (), OptionQuery>;
 
+	/// Data stored by the pre-upgrade hook of the MBM.
+	#[cfg(feature = "try-runtime")]
+	#[pallet::storage]
+	pub type PreUpgradeBytes<T: Config> =
```
ggwpez, line 397: It could also be a storage alias - give that it is not a proper storage value but only temporarily for testing.

substrate/frame/migrations/src/lib.rs
```diff
@@ -698,6 +717,14 @@ impl<T: Config> Pallet<T> {
 		}
 
 		let max_steps = T::Migrations::nth_max_steps(cursor.index);
+
+		// If this is the first time running this migration, exec the pre-upgrade hook.
```
ggwpez, line 721: The line code above `Historic::<T>::contains_key` should already prevent it from executing a migration twice, so this is implicit?

substrate/frame/migrations/src/lib.rs
```diff
@@ -756,10 +792,20 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Fail the current runtime upgrade, caused by `migration`.
+	///
+	/// When the `try-runtime` feature is enabled, this function will panic.
+	// Allow unreachable code so it can compile without warnings when `try-runtime` is enabled.
+	#[allow(unreachable_code)]
```
ggwpez, line 798: Could be that it requires this to be `cfg_attr`, depending on whether clippy forbids unused lints.

substrate/frame/migrations/src/tests.rs
```diff
@@ -333,3 +352,92 @@ fn migration_timeout_errors() {
 		assert_eq!(upgrades_started_completed_failed(), (0, 0, 1));
 	});
 }
+
+#[cfg(feature = "try-runtime")]
```
ggwpez, line 356: Having the tests split by feature is slightly problematic, as we probably dont check both cases in the CI...

substrate/frame/migrations/src/lib.rs
liamaharon: If the migration runs multiple steps, I think this is still needed to ensure pre_upgrade isn't run again, because the migration is only added to Historic after it finishes.

substrate/frame/migrations/src/lib.rs
ggwpez: There is a clippy lint for unused clippy lints. So if you put the `unreachable_code` without there being any unreachable code, then it would error.  
But we will see, maybe it does not trigger 🤷‍♂️

substrate/frame/migrations/src/tests.rs
ggwpez: Yea me neither... its a recurring problem that we have not just with try-runtime but also a lot with the benchmarks feature.  
Maybe adding another job to the test-pipeline that only tests this crate without the feature could work.

substrate/frame/migrations/src/lib.rs
liamaharon: Seems like no issue with current CI

substrate/frame/migrations/src/tests.rs
liamaharon: Added this https://github.com/paritytech/polkadot-sdk/pull/4251/commits/028aa60ed26e6687e39527765fca6f081a49cc85

Let's see if it works 😁

liamaharon: Worked

substrate/frame/migrations/src/lib.rs
liamaharon: It wasn't erroring, but I've added cfg_attr anyways in case we change how we use clippy

claravanstaden:
@liamaharon is this still being worked on?

bkchr:
Ty!

bkchr:
I assume @ggwpez can help

substrate/frame/migrations/src/lib.rs
```diff
@@ -700,6 +730,16 @@ impl<T: Config> Pallet<T> {
...
 
 		let max_steps = T::Migrations::nth_max_steps(cursor.index);
+
+		// If this is the first time running this migration, exec the pre-upgrade hook.
+		#[cfg(feature = "try-runtime")]
+		if !PreUpgradeBytes::<T>::contains_key(&bounded_id) {
+			let bytes = T::Migrations::nth_pre_upgrade(cursor.index)
+				.expect("Invalid cursor.index")
+				.expect("Pre-upgrade failed");
+			PreUpgradeBytes::<T>::insert(&bounded_id, PreUpgradeBytesWrapper(bytes));
```
ggwpez, line 740: This part is a but ugly - since it puts the intermediary state into storage. But i dont think there is a spot where we can put intermediary data that is persisted between blocks.  
The threadlocal stuff gets reset after each block i think...

ggwpez (approved):
As good as it gets for now, going to push some more small cleanups though.

georgepisaltu (approved):
LGTM

claravanstaden (approved):

ggwpez:
Waiting for https://github.com/paritytech/polkadot-sdk/pull/5649 to merge as to not block the merge queue.

alvicsam (commented):
Would be nice to move this check to [github action](https://github.com/paritytech/polkadot-sdk/blob/master/.github/workflows/tests-linux-stable.yml). If this is not urgent I can do it in this PR, if it's urgent - I can do it later.

ggwpez:
I am trying to get it in soon since i then want to finally merge this https://github.com/paritytech/polkadot-sdk/pull/5676 without any conflicts. I enabled auto-merge last week, but apparently nothing happened.

I will re-try it one more time and if that does not work then i guess we have to move the check?

PR #4257, opened 2024-04-23
Update `RuntimeVerison` type and use `system_version` to derive extrinsics root `StateVersion` instead of `V0`

vedhavyas:
This PR 
- Renames `RuntimeVersion::state_version` to `system_version`
- Uses `Runtime::system_version` to derive extrinsics root `StateVersion` instead of default `StateVersion::V0`

This PR should not be breaking any existing chains so long as they use same `RuntimeVersion::state_version` for `Runtime::system_version`

Using `RuntimeVersion::system_version = 2` will make the extrinsics root to use `StateVersion::V1` instead of `V0`

RFC for this change - https://github.com/polkadot-fellows/RFCs/pull/42


vedhavyas:
@bkchr I had looked through the failed tests and not sure if the failures are related to changes in this PR. Can you confirm? If not where should I look at to fix these tests ?

vedhavyas:
Hey @koute @bkchr would appreciate some thoughts on the failing issues.
Also, any chance you can do a first pass/review this PR ?

vedhavyas:
Hey @bkchr @koute 
Would be really great to this PR in. I will fix the conflicts in sometime but please have a look so that this is landed

vedhavyas:
@bkchr @koute bumping this again

vedhavyas:
@bkchr @koute bumping this again

vedhavyas:
@bkchr @koute bumping this again :)

cumulus/parachains/runtimes/starters/seedling/src/lib.rs
```diff
@@ -79,7 +79,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 2,
-	state_version: 0,
+	system_version: 0,
```
bkchr, line 82:
```suggestion
	system_version: 1,
```

cumulus/parachains/runtimes/starters/shell/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
-	state_version: 0,
+	system_version: 0,
```
bkchr, line 90:
```suggestion
	system_version: 1,
```

bkchr (approved):

bkchr:
@vedhavyas sorry for the delay. Can you please fix the CI issues and merge master?

vedhavyas:
@bkchr yes will do. Thank you!

prdoc/pr_4257.prdoc
```diff
@@ -0,0 +1,21 @@
...
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR renames `state_version` in `RuntimeVersion` to `system_version`. `system_version=2` signifies
+      that extrinsic root derivation uses `StateVersion::V1`.
+
+  - audience: Runtime User
+    description: |
+      `RuntimeVersion`'s `state_version` is renames to `system_version`. Applications using that type and its field
```
koute, line 14:
```suggestion
      `RuntimeVersion`'s `state_version` is renamed to `system_version`. Applications using that type and its field
```

substrate/primitives/version/proc-macro/src/decl_runtime_version.rs
```diff
@@ -124,8 +124,8 @@ impl ParseRuntimeVersion {
 			parse_once(&mut self.impl_version, field_value, Self::parse_num_literal)?;
 		} else if field_name == "transaction_version" {
 			parse_once(&mut self.transaction_version, field_value, Self::parse_num_literal)?;
-		} else if field_name == "state_version" {
-			parse_once(&mut self.state_version, field_value, Self::parse_num_literal_u8)?;
```
koute, line 128: Would be nice to still accept the old `state_version` for backwards compatibility and emit a deprecation message prompting the users to change to `system_version`.

Also, since this is a procedural macro it could be nice to allow people to specify the versions in a "nice" way instead of just having a hardcoded magic number, for example:

```
    // instead of "system_version: 0":
    state_version: StateVersion::V0,
    extrinsics_root_state_version: StateVersion::V0,

    // instead of "system_version: 1":
    state_version: StateVersion::V1,
    extrinsics_root_state_version: StateVersion::V0,

    // instead of "system_version: 2":
    state_version: StateVersion::V1,
    extrinsics_root_state_version: StateVersion::V1,
```

This would be nicer from a users' point of view, but I suppose it makes the macro even more "magic", so \*shrug\*. I'll leave it up to whatever @bkchr prefers.

vedhavyas: This is similar to my initial proposal but decided to not go that route since this adds another extra field to Runtimeversion. More here - https://github.com/polkadot-fellows/RFCs/pull/42#discussion_r1372588369

koute: I'm not talking about adding/changing the physical fields - the RFC's already accepted so that train already left the station. I'm just talking about providing potential syntax sugar in the macro which would turn the `state_version` + `extrinsics_root_state_version` into `system_version`.

A less magic alternative would be something like a `const fn system_version(state_version: StateVersion, extrinsics_root_state_version: StateVersion) -> u8` or `SystemVersion { state_version: ..., extrinsics_root_state_version: ... }.into()` helper or something like that.

(But, again, I don't have a particularly strong opinion here which way to go here. But at very least I'd like to not force additional churn on people by breaking backwards compatibility.)

bkchr: > `state_version` + `extrinsics_root_state_version` into `system_version`.

Not sure we should do this.

The general idea of still supporting `state_version` sounds like a good proposal. So, it prints some warning and internally converts it to `system_version`.

bkchr: However, I would not separate the fields.

vedhavyas:
@bkchr seems like CI is stuck. Is there an way to restart it ?

koute:
> seems like CI is stuck. Is there an way to restart it ?

Done.

vedhavyas:
@koute any idea why some tests failed. Looking at the CI logs of each failed test does not give much info. Any pointers pls ?

bkchr:
I think the zombienet tests are just flaky. @vedhavyas please give us merge rights to your branch, then we can also merge master for example

substrate/primitives/version/proc-macro/src/decl_runtime_version.rs
bkchr: @vedhavyas and could you do this?

nazar-pc:
> @vedhavyas please give us merge rights to your branch, then we can also merge master for example

I don't think that'll work because PR was created from an organization fork, not from a personal one.

I just merged `master` in though.

nazar-pc:
Merged `master` again, hoping to push this across the finish line soon :crossed_fingers:

koute:
> @koute how can I get you interested in reviewing this before cutting the next stable release?

Make it backwards compatible (have it emit a deprecation warning when the old field name is used) and you'll have a :+1: from me.

bkchr:
What @koute is referring to is, this discussion: https://github.com/paritytech/polkadot-sdk/pull/4257#discussion_r1690053835

vedhavyas:
@bkchr @koute Should be updated now. Can you take a look if I missed anything ?

substrate/primitives/version/proc-macro/src/decl_runtime_version.rs
```diff
@@ -125,7 +130,10 @@ impl ParseRuntimeVersion {
 		} else if field_name == "transaction_version" {
 			parse_once(&mut self.transaction_version, field_value, Self::parse_num_literal)?;
 		} else if field_name == "state_version" {
-			parse_once(&mut self.state_version, field_value, Self::parse_num_literal_u8)?;
+			log::warn!("{STATE_VERSION_DEPRECATION_WARN}");
```
bkchr, line 133: https://github.com/ggwpez/proc-macro-warning please use this here.

bkchr (commented):
One small nitpick and then this is good to go.

substrate/primitives/version/proc-macro/src/decl_runtime_version.rs
vedhavyas: @bkchr I have updated to use the `proc-macro-warning` crate until now. I derived the change from the examples. Let me know if something can be improved there

nazar-pc:
@bkchr @koute are we good to go here?

bkchr (approved):
Looks good. Some CI tests are still unhappy.

nazar-pc:
Not sure what to do with zombienet checks (not clear what is wrong from CI logs alone), but the rest of the CI should be fixed assuming I didn't miss anything in prdoc

bkchr:
The zombienet errors seem to be valid: 

> tokio-runtime-worker cumulus-pov-recovery: [Parachain] Failed to fetch relay chain runtime version. error=RpcCallError("state_getRuntimeVersion")

The bridges zombienet errors are maybe also related. Probably PJS tries to fetch the `RuntimeVersion` and it not having the `state_version` field is maybe the problem.

nazar-pc:
I have added `stateVersion` alias to `systemVersion` field, which should fix the decoding issue. If desirable customized serialization can be implemented to produce both `stateVersion` and `systemVersion` in the serialized JSON for smoother migration process.

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +230,10 @@ pub struct RuntimeVersion {
 	/// This number should never decrease.
 	pub transaction_version: u32,
 
-	/// Version of the state implementation used by this runtime.
+	/// Version of the system implementation used by this runtime.
 	/// Use of an incorrect version is consensus breaking.
-	pub state_version: u8,
+	#[cfg_attr(feature = "serde", serde(alias = "stateVersion",))]
```
bkchr, line 235:
```suggestion
	#[cfg_attr(feature = "serde", serde(alias = "stateVersion"))]
```

bkchr:
> If desirable customized serialization can be implemented to produce both `stateVersion` and `systemVersion` in the serialized JSON for smoother migration process.

Yeah I also thought about this.

nazar-pc:
So do you want me to do that?

bkchr:
> So do you want me to do that?

Yeah please do it. Can you also try to merge master again? I hope that this fixes the bridges tests.

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,394 @@ pub struct RuntimeVersion {
...
+						},
+						Field::TransactionVersion => {
+							if transaction_version.is_some() {
+								return Err(<A::Error as serde::de::Error>::duplicate_field(
+									"transactionVersion",
+								));
+							}
+							transaction_version = Some(map.next_value()?);
+						},
+						Field::SystemVersion => {
```
bkchr, line 558: If both exists, they should be the same value

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,394 @@ pub struct RuntimeVersion {
...
+							system_version = Some(map.next_value()?);
```
bkchr, line 559: Also you are missing here the check if this is already some.

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,394 @@ pub struct RuntimeVersion {
...
+		struct FieldVisitor;
+
+		impl<'de> serde::de::Visitor<'de> for FieldVisitor {
+			type Value = Field;
+
+			fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+				formatter.write_str("field identifier")
+			}
+
+			fn visit_u64<E>(self, value: u64) -> Result<Self::Value, E>
```
bkchr, line 297: Do we need this?

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,394 @@ pub struct RuntimeVersion {
...
+					"specVersion" => Ok(Field::SpecVersion),
+					"implVersion" => Ok(Field::ImplVersion),
+					"apis" => Ok(Field::Apis),
+					"transactionVersion" => Ok(Field::TransactionVersion),
+					"systemVersion" | "stateVersion" => Ok(Field::SystemVersion),
+					_ => Ok(Field::Ignore),
+				}
+			}
+
+			fn visit_bytes<E>(self, value: &[u8]) -> Result<Self::Value, E>
```
bkchr, line 331: And this?

substrate/primitives/version/src/lib.rs
nazar-pc: I didn't write it, I used macro expansion as the basis and this is what it spit out, I just made it a bit more readable

substrate/primitives/version/src/lib.rs
nazar-pc: Same here, this is from `Deserialize` derive

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,406 @@ pub struct RuntimeVersion {
...
+							}
+							transaction_version = Some(map.next_value()?);
+						},
+						Field::SystemVersion =>
+							if let Some(system_version) = system_version {
+								let new_value = map.next_value::<u8>()?;
+								if system_version != new_value {
+									return Err(<A::Error as serde::de::Error>::custom(
+										alloc::format!(
+											r#"Duplicated "stateVersion" and "systemVersion" \
```
bkchr, line 564: You don't really know this. It could also be that there are more than one `stateVersion`.

substrate/primitives/version/src/lib.rs
```diff
@@ -230,9 +221,406 @@ pub struct RuntimeVersion {
...
+	/// Version of the system implementation used by this runtime.
 	/// Use of an incorrect version is consensus breaking.
-	pub state_version: u8,
+	pub system_version: u8,
+}
+
+// Manual implementation in order to sprinkle `stateVersion` at the end for migration purposes
+// after the field was renamed from `state_version` to `system_version`
+#[cfg(feature = "serde")]
+impl serde::Serialize for RuntimeVersion {
```
bkchr, line 232: Can we move all this code into a separate file? 400 lines is a lot.

bkchr (approved):

substrate/primitives/version/src/lib.rs
nazar-pc: You're right, but it is one of those two fields. And I decided to make this assumption in an error message since it is the most likely. Open for suggestions.

substrate/primitives/version/src/lib.rs
nazar-pc: Would `serde` module work? I generally try to keep impls next to the definition, but agree that in this case it is quite lengthy.

koute, line 232: Hmm.... I don't think we need this `serde` boilerplate? Something like this should work:

```rust
use serde::{Serialize, Deserialize};

#[derive(Clone)]
pub struct RuntimeVersion {
    // ...
    pub system_version: u8
}

#[derive(Clone, Serialize, Deserialize)]
struct RuntimeVersionSerde {
    // ...
    system_version: Option<u8>,
    state_version: Option<u8>,
}

impl From<RuntimeVersion> for RuntimeVersionSerde {
    fn from(value: RuntimeVersion) -> Self {
        Self {
            // ...
            system_version: Some(value.system_version),
            state_version: Some(value.system_version),
        }
    }
}

impl From<RuntimeVersionSerde> for RuntimeVersion {
    fn from(value: RuntimeVersionSerde) -> Self {
        Self {
            // ..
            system_version: value.system_version.or(value.state_version)
                .unwrap() // TODO: Properly handle if both/none are there.
        }
    }
}

impl serde::Serialize for RuntimeVersion {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        RuntimeVersionSerde::from(self.clone()).serialize(serializer)
    }
}

impl<'de> serde::Deserialize<'de> for RuntimeVersion {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        Ok(RuntimeVersionSerde::deserialize(deserializer)?.into())
    }
}
```

Basically - use an auxiliary struct to do the serialization/deserialization, and then just convert between the real `RuntimeVersion` and the auxiliary one. You still need some boilerplate to copy the fields, but it should be much less than the serde junk and be much easier to maintain.

nazar-pc: This looks appealing, but `Option` and missing field are not 100% the same for `serde`. How big of an issue is this really?

nazar-pc:
If there are no blockers here, I'd really appreciate finally merging this and be done with it :pray:

substrate/primitives/version/src/lib.rs
bkchr: Yeah, just let's take the current code.

bkchr:
@koute needs to approve :P

substrate/primitives/version/src/lib.rs
koute: > This looks appealing, but `Option` and missing field are not 100% the same for `serde`. How big of an issue is this really?

@nazar-pc What do you mean are not the same? What exactly is the problem? You can implement strictly the same behavior as you've currently implemented, just with a lot less code (and more elegant) if you use a wrapper like I've suggested.

I'm fine with taking the existing code *if* it's truly necessary, but if not I'd rather have a simplified version that won't be pain to maintain and won't require you to read a ton of serde boilerplate to understand what it does.

nazar-pc: Serde is not used just for JSON, so while `Some(u8)` and `u8` encode identically in JSON, I do not know if it is actually the case for all formats that serde supports, so the change is not strictly equivalent in serialization case.

koute: > Serde is not used just for JSON, so while `Some(u8)` and `u8` encode identically in JSON, I do not know if it is actually the case for all formats that serde supports, so the change is not strictly equivalent in serialization case.

Not entirely sure if we care (do we?), but, okay, fair enough. In that case you can just use two internal structs - one for serialization with both fields as `u8`, and one for deserialization with them being `Option<u8>`. That would solve this problem and still be less code/boilerplate, no?

So, basically, this (quick copy-paste so might not compile):

```rust
/// Runtime version.
/// This should not be thought of as classic Semver (major/minor/tiny).
/// This triplet have different semantics and mis-interpretation could cause problems.
/// In particular: bug fixes should result in an increment of `spec_version` and possibly
/// `authoring_version`, absolutely not `impl_version` since they change the semantics of the
/// runtime.
#[derive(Clone, PartialEq, Eq, Encode, Default, sp_runtime::RuntimeDebug, TypeInfo)]
pub struct RuntimeVersion {
	/// Identifies the different Substrate runtimes. There'll be at least polkadot and node.
	/// A different on-chain spec_name to that of the native runtime would normally result
	/// in node not attempting to sync or author blocks.
	pub spec_name: RuntimeString,

	/// Name of the implementation of the spec. This is of little consequence for the node
	/// and serves only to differentiate code of different implementation teams. For this
	/// codebase, it will be parity-polkadot. If there were a non-Rust implementation of the
	/// Polkadot runtime (e.g. C++), then it would identify itself with an accordingly different
	/// `impl_name`.
	pub impl_name: RuntimeString,

	/// `authoring_version` is the version of the authorship interface. An authoring node
	/// will not attempt to author blocks unless this is equal to its native runtime.
	pub authoring_version: u32,

	/// Version of the runtime specification.
	///
	/// A full-node will not attempt to use its native runtime in substitute for the on-chain
	/// Wasm runtime unless all of `spec_name`, `spec_version` and `authoring_version` are the same
	/// between Wasm and native.
	///
	/// This number should never decrease.
	pub spec_version: u32,

	/// Version of the implementation of the specification.
	///
	/// Nodes are free to ignore this; it serves only as an indication that the code is different;
	/// as long as the other two versions are the same then while the actual code may be different,
	/// it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are
	/// about the only changes that could be made which would result in only the `impl_version`
	/// changing.
	///
	/// This number can be reverted to `0` after a [`spec_version`](Self::spec_version) bump.
	pub impl_version: u32,

	/// List of supported API "features" along with their versions.
	pub apis: ApisVec,

	/// All existing calls (dispatchables) are fully compatible when this number doesn't change. If
	/// this number changes, then [`spec_version`](Self::spec_version) must change, also.
	///
	/// This number must change when an existing call (pallet index, call index) is changed,
	/// either through an alteration in its user-level semantics, a parameter
	/// added/removed, a parameter type changed, or a call/pallet changing its index. An alteration
	/// of the user level semantics is for example when the call was before `transfer` and now is
	/// `transfer_all`, the semantics of the call changed completely.
	///
	/// Removing a pallet or a call doesn't require a *bump* as long as no pallet or call is put at
	/// the same index. Removing doesn't require a bump as the chain will reject a transaction
	/// referencing this removed call/pallet while decoding and thus, the user isn't at risk to
	/// execute any unknown call. FRAME runtime devs have control over the index of a call/pallet
	/// to prevent that an index gets reused.
	///
	/// Adding a new pallet or call also doesn't require a *bump* as long as they also don't reuse
	/// any previously used index.
	///
	/// This number should never decrease.
	pub transaction_version: u32,

	/// Version of the system implementation used by this runtime.
	/// Use of an incorrect version is consensus breaking.
	pub system_version: u8,
}

#[cfg(feature = "serde")]
#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
struct RuntimeVersionSer {
    spec_name: RuntimeString,
    impl_name: RuntimeString,
    authoring_version: u32,
    spec_version: u32,
    impl_version: u32,
    #[serde(serialize_with = "apis_serialize::serialize")]
    apis: ApisVec,
    transaction_version: u32,
    system_version: u8,
    state_version: u8,
}

#[cfg(feature = "serde")]
#[derive(Deserialize)]
#[serde(rename_all = "camelCase")]
struct RuntimeVersionDe {
    spec_name: RuntimeString,
    impl_name: RuntimeString,
    authoring_version: u32,
    spec_version: u32,
    impl_version: u32,
    #[serde(deserialize_with = "apis_serialize::deserialize"]
    apis: ApisVec,
    transaction_version: u32,
    system_version: Option<u8>,
    state_version: Option<u8>,
}

#[cfg(feature = "serde")]
impl serde::Serialize for RuntimeVersion {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        RuntimeVersionSer {
            spec_name: self.spec_name.clone(),
            impl_name: self.impl_name.clone(),
            authoring_version: self.authoring_version,
            spec_version: self.spec_version,
            impl_version: self.impl_version,
            apis: self.apis.clone(),
            transaction_version: self.transaction_version,
            system_version: self.system_version,
            state_version: self.system_version,
        }.serialize(serializer)
    }
}

#[cfg(feature = "serde")]
impl<'de> serde::Deserialize<'de> for RuntimeVersion {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let v = RuntimeVersionDe::deserialize(deserializer)?;
        let system_version = match (v.system_version, v.state_version) {
            (Some(version), None) | (None, Some(version)) => version,
            (Some(system_version), Some(state_version)) => {
                if system_version == state_version {
                    system_version
                } else {
                    return Err(<D::Error as serde::de::Error>::custom(
                        alloc::format!(
                            r#"Duplicated "stateVersion" and "systemVersion" \
                            fields must have the same value, but different values \
                            were provided: {system_version} vs {new_value}"#
                        ),
                    ));
                }
            },
            (None, None) => return Err(<D::Error as serde::de::Error>::missing_field("systemVersion")),
        };
        Ok(Self {
            spec_name: v.spec_name,
            impl_name: v.impl_name,
            authoring_version: v.authoring_version,
            spec_version: v.spec_version,
            impl_version: v.impl_version,
            apis: v.apis,
            transaction_version: v.transaction_version,
            system_version,
        })
    }
}
```

Isn't this much cleaner?

nazar-pc: Same thing with deserialization: you can deserialize a number into both `Option<u8>` and `u8` from JSON, but I do not think it is guaranteed to be symmetrical that way with other formats. If we do not care then fine, but I was trying to make it actually equivalent and didn't find a clean way to do so without manual implementation.

koute: Well, as far as I know the only format we ever explicitly supported was JSON (we just used serde because that's the "standard" way to support JSON), and if we care about formats other than JSON then ideally we should have tests for serialization/deserialization of those formats, otherwise considering how flexible serde is we essentially always risk breaking some unknown code somewhere. And we're already partially breaking compatibility anyway by having it serialize an extra field!

So, to reiterate my policy here: if there's a good reason to keep things complex, sure, go for it, but if it's something that's "just in case" and we don't have any idea at all whether it's necessary, then I just prefer to keep things as simple as possible.

Anyway, honestly, taking a step back: the simplest thing would have been to leave the name `state_version` as `state_version` and just introduce the possibility of it being `2`, and then all of this could have been avoided! We wouldn't have to worry whether this breaks anything, we wouldn't need custom serialization/deserialization, we wouldn't need deprecation warnings, consumers of `RuntimeVersion` that don't expect the extra field wouldn't break (if there are any then they will break now!), we wouldn't have to remove the `state_version` from the serialized `RuntimeVersion` in the future potentially breaking consumers *yet again* (or, alternatively, we wouldn't leave those two fields in there forever - do you have a plan for when to remove the `state_version`?), and this PR would have been already merged long ago. The only advantage of renaming it is, as far as I can see, a slightly nicer name, and a big list of pain and disadvantages to go with it.

The RFC was already accepted by the fellowship, so I **won't** block this, but my opinion is that I would *highly* prefer to just leave the old name as it was, delete all of the new extra code and call it a day.

Anyway, okay, sorry, rant over. If Big Boss @bkchr's fine with this then I'm also fine; I don't like it, but I don't see any hard blockers. :P

koute (approved):

substrate/primitives/version/src/lib.rs
koute: Although, now that I think about it again, isn't emitting both `state_version` and `system_version` non-compliant with the RFC? The RFC stated that only `system_version` should be there. For `polkadot-sdk` internally we're fine to do whatever we want, but on external community/Polkadot interfaces (like the serialized `RuntimeVersion` struct) what is in the Polkadot spec and in fellowship RFCs takes precedence over what we want, so shouldn't we only emit `system_version` or update the RFC?

bkchr: For SCALE the encoding is as documented in the RFC. For others consuming the type via JSON (which is self descriptive and it is observable that the field name changed), they can continue to use their old code without it breaking.

bkchr: (I'm also not 100% happy with the 400 lines of useless code, but there is no real better solution...)

bkchr:
@nazar-pc we need a master merge again :see_no_evil:

liuchengxu:
There is a leftover in chain sync :( https://github.com/paritytech/polkadot-sdk/blob/ec9a734f9a1f9aad839f0fd1d549179d7b1354a4/substrate/client/network/sync/src/strategy/chain_sync.rs#L2290-L2293

cc @nazar-pc

bkchr:
https://github.com/paritytech/polkadot-sdk/pull/5686

Issue #4255, opened 2024-04-23
`UncheckedExtrinsic` decoding improvements


serban300:
We can modify `UncheckedExtrinsic` in order to get the bytes for the `Call`, without decoding it.

Credits to @xlc and @bkchr for the idea.

serban300:
## 18 March 2025 update:
- opened another  PR that addresses this and also sets a memory limit for the `UncheckedExtrinsic` . I marked it as draft for the moment. I just need to polish some things. Will get it ready for review soon. The PR is: https://github.com/paritytech/polkadot-sdk/pull/7902

serban300:
## 29 October 2025 update:

- The PR related to [https://github.com/paritytech/polkadot-sdk/pull/9480](Lazy decoding block extrinsics)  was merged. This was the main part of the work here

- Remaining work: https://github.com/paritytech/polkadot-sdk/issues/8675

PR #4260, opened 2024-04-23
pallet-xcm::transfer_assets_using_type() supports custom actions on destination

acatangiu:
Change `transfer_assets_using_type()` to not assume `DepositAssets` as the intended use of the assets on the destination.

Instead provides the caller with the ability to specify custom XCM that be executed on `dest` chain as the last step of the transfer, thus allowing custom usecases for the transferred assets. E.g. some are used/swapped/etc there, while some are sent further to yet another chain.

Note: this is a follow-up on https://github.com/paritytech/polkadot-sdk/pull/3695, bringing in an API change for `transfer_assets_using_type()`. This is ok as the previous version has not been yet released. Thus, its first release will include the new API proposed by this PR.

This allows usecases such as: https://forum.polkadot.network/t/managing-sas-on-multiple-reserve-chains-for-same-asset/7538/4

BTW: all this pallet-xcm asset transfers code will be massively reduced once we have https://github.com/paritytech/xcm-format/pull/54

### DIFF
--- a/polkadot/xcm/pallet-xcm/src/lib.rs
+++ b/polkadot/xcm/pallet-xcm/src/lib.rs
@@ -1422,20 +1422,21 @@ pub mod pallet {
 		///    mint/teleport assets and deposit them to `beneficiary`.
 		///
 		/// On the destination chain, as well as any intermediary hops, `BuyExecution` is used to
-		/// buy execution using transferred `assets` identified by `fees_id`.
-		/// Make sure enough of the specified `fees_id` asset is included in the given list of
-		/// `assets`. `fees_id` should be enough to pay for `weight_limit`. If more weight is needed
-		/// than `weight_limit`, then the operation will fail and the sent assets may be at risk.
+		/// buy execution using transferred `assets` identified by `remote_fees_id`.
+		/// Make sure enough of the specified `remote_fees_id` asset is included in the given list
+		/// of `assets`. `remote_fees_id` should be enough to pay for `weight_limit`. If more weight
+		/// is needed than `weight_limit`, then the operation will fail and the sent assets may be
+		/// at risk.
 		///
-		/// `fees_id` may use different transfer type than rest of `assets` and can be specified
-		/// through `fees_transfer_type`.
+		/// `remote_fees_id` may use different transfer type than rest of `assets` and can be
+		/// specified through `fees_transfer_type`.
 		///
 		/// The caller needs to specify what should happen to the transferred assets once they reach
 		/// the `dest` chain. This is done through the `custom_xcm_on_dest` parameter, which
 		/// contains the instructions to execute on `dest` as a final step.
-		///     This is usually as simple as:
-		///     `Xcm::<()>(vec![DepositAsset { assets: Wild(AllCounted(assets.len())), beneficiary
-		/// }])`,     but could be something more exotic like sending the `assets` even further.
+		///   This is usually as simple as:
+		///   `Xcm(vec![DepositAsset { assets: Wild(AllCounted(assets.len())), beneficiary }])`,
+		///   but could be something more exotic like sending the `assets` even further.
 		///
 		/// - `origin`: Must be capable of withdrawing the `assets` and executing XCM.
 		/// - `dest`: Destination context for the assets. Will typically be `[Parent,
@@ -1445,14 +1446,14 @@ pub mod pallet {
 		/// - `assets`: The assets to be withdrawn. This should include the assets used to pay the
 		///   fee on the `dest` (and possibly reserve) chains.
 		/// - `assets_transfer_type`: The XCM `TransferType` used to transfer the `assets`.
-		/// - `fees_id`: One of the included `assets` to be be used to pay fees.
+		/// - `remote_fees_id`: One of the included `assets` to be be used to pay fees.
 		/// - `fees_transfer_type`: The XCM `TransferType` used to transfer the `fees` assets.
 		/// - `custom_xcm_on_dest`: The XCM to be executed on `dest` chain as the last step of the
 		///   transfer, which also determines what happens to the assets on the destination chain.
 		/// - `weight_limit`: The remote-side weight limit, if any, for the XCM fee purchase.
 		#[pallet::call_index(15)]
 		#[pallet::weight(T::WeightInfo::transfer_assets())]
-		pub fn transfer_assets_using_type(
+		pub fn transfer_assets_using_type_and_then(
 			origin: OriginFor<T>,
 			dest: Box<VersionedLocation>,
 			assets: Box<VersionedAssets>,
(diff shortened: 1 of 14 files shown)


svyatonik (approved):

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1421,50 +1421,59 @@ pub mod pallet {
...
 		/// - `dest`: Destination context for the assets. Will typically be `[Parent,
 		///   Parachain(..)]` to send from parachain to parachain, or `[Parachain(..)]` to send from
 		///   relay to parachain, or `(parents: 2, (GlobalConsensus(..), ..))` to send from
 		///   parachain across a bridge to another ecosystem destination.
-		/// - `beneficiary`: A beneficiary location for the assets in the context of `dest`. Will
-		///   generally be an `AccountId32` value.
 		/// - `assets`: The assets to be withdrawn. This should include the assets used to pay the
 		///   fee on the `dest` (and possibly reserve) chains.
 		/// - `assets_transfer_type`: The XCM `TransferType` used to transfer the `assets`.
 		/// - `fees_id`: One of the included `assets` to be be used to pay fees.
```
bkontur, line 1448:
```suggestion
		/// - `remote_fees_id`: One of the included `assets` to be be used to pay fees.
```

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1421,50 +1421,59 @@ pub mod pallet {
...
-		/// given list of `assets`. `fees_id` should be enough to pay for `weight_limit`. If more
-		/// weight is needed than `weight_limit`, then the operation will fail and the sent assets
-		/// may be at risk.
+		/// On the destination chain, as well as any intermediary hops, `BuyExecution` is used to
+		/// buy execution using transferred `assets` identified by `fees_id`.
+		/// Make sure enough of the specified `fees_id` asset is included in the given list of
+		/// `assets`. `fees_id` should be enough to pay for `weight_limit`. If more weight is needed
+		/// than `weight_limit`, then the operation will fail and the sent assets may be at risk.
 		///
 		/// `fees_id` may use different transfer type than rest of `assets` and can be specified
```
bkontur, line 1430:
```suggestion
		/// `remote_fees_id` may use different transfer type than rest of `assets` and can be specified
```

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1421,50 +1421,59 @@ pub mod pallet {
...
-		/// Fee payment on the source, destination and all intermediary hops, is specified through
-		/// `fees_id`, but make sure enough of the specified `fees_id` asset is included in the
-		/// given list of `assets`. `fees_id` should be enough to pay for `weight_limit`. If more
-		/// weight is needed than `weight_limit`, then the operation will fail and the sent assets
-		/// may be at risk.
+		/// On the destination chain, as well as any intermediary hops, `BuyExecution` is used to
+		/// buy execution using transferred `assets` identified by `fees_id`.
+		/// Make sure enough of the specified `fees_id` asset is included in the given list of
+		/// `assets`. `fees_id` should be enough to pay for `weight_limit`. If more weight is needed
+		/// than `weight_limit`, then the operation will fail and the sent assets may be at risk.
```
bkontur, line 1428:
```suggestion
		/// On the destination chain, as well as any intermediary hops, `BuyExecution` is used to
		/// buy execution using transferred `assets` identified by `remote_fees_id`.
		/// Make sure enough of the specified `remote_fees_id` asset is included in the given list of
		/// `assets`. `remote_fees_id` should be enough to pay for `weight_limit`. If more weight is needed
		/// than `weight_limit`, then the operation will fail and the sent assets may be at risk.
```

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1814,22 +1834,22 @@ impl<T: Config> Pallet<T> {
...
-				let (local, remote) = Self::teleport_assets_program(
+				let (local, mut remote) = Self::teleport_assets_program(
 					origin.clone(),
 					dest.clone(),
-					beneficiary,
 					assets,
 					fees,
 					weight_limit,
 				)?;
+				remote.0.extend(custom_remote_xcm.into_iter());
```
bkontur, line 1852: just one nit or question, now the `Self::teleport_assets_program` is kind of incomplete, and there could be a potential problem if somebody reuse it elsewhere without extending it with `beneficiary` or `custom_remove_xcm` with `DepositAsset` - I know, it's very hypotetical :)

What I am trying to say, we have four TransferType and we do this `remote.0.extend` just for three of them and just only for `Self::remote_reserve_transfer_program` you basically changed `beneficiary` -> `custom_remote_xcm`.
What about doing the same change `beneficiary` -> `custom_remote_xcm` for all `Self::remote_reserve_transfer_program` / `Self::teleport_assets_program` / `Self::destination_reserve_transfer_programs` / `Self::local_reserve_transfer_programs`, so we don't need to do `remote.0.extend` at all? That way also `Self::teleport_assets_program` will become kind of complete.

bkontur (approved):

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -1421,50 +1421,59 @@ pub mod pallet {
...
 		pub fn transfer_assets_using_type(
 			origin: OriginFor<T>,
 			dest: Box<VersionedLocation>,
-			beneficiary: Box<VersionedLocation>,
 			assets: Box<VersionedAssets>,
 			assets_transfer_type: Box<TransferType>,
-			fees_id: Box<VersionedAssetId>,
+			remote_fees_id: Box<VersionedAssetId>,
 			fees_transfer_type: Box<TransferType>,
+			custom_xcm_on_dest: Box<VersionedXcm<()>>,
```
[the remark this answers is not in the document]
acatangiu: 1. With https://github.com/paritytech/polkadot-sdk/issues/4255 we don't need the blob hack anymore.
2. No, a `ClearOrigin` is run before these user-controlled instructions. Only thing they can do is operate on the holding register (which they loaded in the first place).
3. Yes, I like the name suggestion will rename to `transfer_assets_using_type_and_then`

polkadot/xcm/pallet-xcm/src/lib.rs
acatangiu: I wanted to avoid duplicating that extend logic everywhere since these fns are not used elsewhere, but you are right, maybe in the future we will use them elsewhere and they're "incomplete"

acatangiu:
@bkontur @serban300 I made the suggested changes, edited the prdoc from the original PR, and added emulated tests. PTAL

serban300 (approved):

bkontur (approved):

PR #4265, opened 2024-04-24
Try state: log errors instead of loggin the number of error and discarding them

thiolliere:
Currently we discard errors content
We should at least log it.

Code now is more similar to what is written in try_on_runtime_upgrade.

### DIFF
--- a/substrate/frame/support/src/traits/try_runtime/mod.rs
+++ b/substrate/frame/support/src/traits/try_runtime/mod.rs
@@ -161,22 +161,31 @@ impl<BlockNumber: Clone + sp_std::fmt::Debug + AtLeast32BitUnsigned> TryState<Bl
 		match targets {
 			Select::None => Ok(()),
 			Select::All => {
-				let mut error_count = 0;
+				let mut errors = Vec::new();
+
 				for_tuples!(#(
-					if let Err(_) = Tuple::try_state(n.clone(), targets.clone()) {
-						error_count += 1;
+					if let Err(err) = Tuple::try_state(n.clone(), targets.clone()) {
+						errors.push(err);
 					}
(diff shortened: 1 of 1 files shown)


bkchr (approved):

substrate/frame/support/src/traits/try_runtime/mod.rs
```diff
@@ -161,22 +161,31 @@ impl<BlockNumber: Clone + sp_std::fmt::Debug + AtLeast32BitUnsigned> TryState<Bl
...
 			Select::None => Ok(()),
 			Select::All => {
-				let mut error_count = 0;
+				let mut errors = Vec::<TryRuntimeError>::new();
+
 				for_tuples!(#(
-					if let Err(_) = Tuple::try_state(n.clone(), targets.clone()) {
-						error_count += 1;
+					if let Err(err) = Tuple::try_state(n.clone(), targets.clone()) {
+						errors.push(err);
```
xlc, line 168: why not just log the error here?

ggwpez: There are a lot of intermediate logs from migrations. I think it makes sense to have a comprehensive error report at the end that shows all the errors at once instead of needing to scroll.

ggwpez (approved):

substrate/frame/support/src/traits/try_runtime/mod.rs
thiolliere: Anything is good to me.

Ideally try_state implementations should just log its error with precise file:line information and just return an Error with maybe the pallet name. But currently there is a lot of use of `ensure!(..., "error message")` and this just goes inside the error.

Issue #291, opened 2021-12-13
Signed Metadata


Slesarew:
Currently the Signer needs trusted authority to verify correctness of metadata used. We could make this trustless if we ~require~ allow (edit 1: Signer should not be allowed to sign transactions with trivial blank metadata hash then) every transaction to include metadata hash into signed payload - this way the chain will be constantly verifying metadata validity instead of a trusted party.

This bears very little additional costs, deprecates lots of complex and confusing logic in Signer and makes automated generation of updates possible. Metadata portal might be reduced to a simple untrusted page with generator script since any attack on it will lead to Signer's DoS which is desired behavior.

kianenigma:
I belive this, and all related issues can be closed now as https://github.com/polkadot-fellows/RFCs/blob/bkchr-merkelized-metadata/text/0078-merkleized-metadata.md is formalized.

PR #4274, opened 2024-04-24
Implement `CheckMetadataHash` extension

bkchr:
This implements the `CheckMetadataHash` extension as described in [RFC78](https://polkadot-fellows.github.io/RFCs/approved/0078-merkleized-metadata.html).

Besides the signed extension, the `substrate-wasm-builder` is extended to support generating the metadata-hash.

Closes: https://github.com/paritytech/polkadot-sdk/issues/291


substrate/frame/metadata-hash-extension/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+///
+/// The extension adds one byte (the `mode`) to the size of the extrinsic. This one byte is
+/// controlling if the metadata hash should be added to the signed data or not. Mode `0` means that
+/// the metadata hash is not added and `1` means that it is added. Further values of `mode` are
+/// reserved for future changes.
+///
+/// The metadata hash is read from the environment variable `RUNTIME_METADATA_HASH`. This
+/// environment variable is for example set by the `substrate-wasm-builder` when the feature for
+/// generating the metadata hash is enabled. If the environment variable is not set and `mode = 1`
+/// is passed, the transaction is rejected with [`UnknownTransaction::CannotLookup`].
```
ggwpez, line 95: And when a transaction sets `mode = 0` but the runtime knows the hash? Is it optional or will it also be rejected?  
I guess we need to wait for wallet support first.

substrate/frame/metadata-hash-extension/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+	fn additional_signed(&self) -> Result<Self::AdditionalSigned, TransactionValidityError> {
+		match self.mode {
+			0 => Ok(EncodeNoneToEmpty(None)),
+			1 => match option_env!("RUNTIME_METADATA_HASH") {
+				Some(hash) => Ok(EncodeNoneToEmpty(Some(array_bytes::hex2array_unchecked(hash)))),
+				None => Err(UnknownTransaction::CannotLookup.into()),
+			},
+			// Unknown `mode`, let's reject it.
+			_ => Err(UnknownTransaction::CannotLookup.into()),
```
ggwpez, line 125: Could it not use an enum for mode, so that it already fails to decode an invalid mode?

substrate/frame/metadata-hash-extension/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+	type AdditionalSigned = EncodeNoneToEmpty;
+	type Pre = ();
+	const IDENTIFIER: &'static str = "CheckMetadataHash";
+
+	fn additional_signed(&self) -> Result<Self::AdditionalSigned, TransactionValidityError> {
+		match self.mode {
+			0 => Ok(EncodeNoneToEmpty(None)),
+			1 => match option_env!("RUNTIME_METADATA_HASH") {
+				Some(hash) => Ok(EncodeNoneToEmpty(Some(array_bytes::hex2array_unchecked(hash)))),
+				None => Err(UnknownTransaction::CannotLookup.into()),
```
ggwpez, line 122: Not sure if this error variant help so much with debugging... alternatively there is only `Other(_)`, which is not so nice either.

substrate/frame/metadata-hash-extension/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+	type Call = <T as Config>::RuntimeCall;
+	type AdditionalSigned = EncodeNoneToEmpty;
+	type Pre = ();
+	const IDENTIFIER: &'static str = "CheckMetadataHash";
+
+	fn additional_signed(&self) -> Result<Self::AdditionalSigned, TransactionValidityError> {
+		match self.mode {
+			0 => Ok(EncodeNoneToEmpty(None)),
+			1 => match option_env!("RUNTIME_METADATA_HASH") {
+				Some(hash) => Ok(EncodeNoneToEmpty(Some(array_bytes::hex2array_unchecked(hash)))),
```
ggwpez, line 121: Could it be a static or something? Probably makes no big difference, but like this it decodes the hash every time.

substrate/frame/support/procedural/src/pallet/expand/constants.rs
```diff
@@ -85,6 +85,8 @@ pub fn expand_constants(def: &mut Def) -> proc_macro2::TokenStream {
 		let default_byte_impl = &const_.default_byte_impl;
 
 		quote::quote!({
+			#frame_support::__private::log::debug!(concat!("Fetching constant: ", #ident_str));
```
ggwpez, line 88: Still needed?

substrate/utils/wasm-builder/src/metadata_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+		// The hash is only used for caching and thus, not that important for our use case here.
+		hash: vec![1, 2, 3],
+	};
+
+	let metadata = executor
+		.call(
+			&mut sp_io::TestExternalities::default().ext(),
+			&runtime_code,
+			"Metadata_metadata_at_version",
+			&15u32.encode(),
```
ggwpez, line 67: Why does it need to be exactly 15 and not the default `Metadata_metadata`? (I think Polkadot returns 14 per default).

substrate/utils/wasm-builder/src/metadata_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+			CallContext::Offchain,
+		)
+		.0
+		.expect("`Metadata::metadata_at_version` should exist.");
+
+	let metadata = Option::<Vec<u8>>::decode(&mut &metadata[..])
```
ggwpez, line 73: Nit: AFAIK it is `frame_metadata::OpaqueMetadata`, but is just a newtype around `Vec<u8>`, so does not really matter.

substrate/utils/wasm-builder/src/metadata_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+		decimals: extra_info.decimals,
+		token_symbol: extra_info.token_symbol,
+	};
+
+	generate_metadata_digest(&metadata, extra_info).unwrap().hash()
+}
+
+/// Extract the `SS58` from the constants in the given `metadata`.
+fn extract_ss58_prefix(metadata: &RuntimeMetadata) -> u16 {
+	let RuntimeMetadata::V15(ref metadata) = metadata else {
```
ggwpez, line 110: OT: what do you use for Runtime API debugging?  
I have been using some CLI like this, but nothing proper: https://github.com/ggwpez/frame-runtime-api

substrate/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -775,12 +814,15 @@ fn offline_build() -> bool {
...
 /// Build the project and create the bloaty runtime blob.
+///
+/// Returns the path to the generated bloaty runtime blob.
 fn build_bloaty_blob(
 	target: RuntimeTarget,
 	blob_build_profile: &Profile,
 	project: &Path,
 	default_rustflags: &str,
 	cargo_cmd: CargoCommandVersioned,
+	#[cfg(feature = "metadata-hash")] metadata_hash: Option<[u8; 32]>,
```
ggwpez, line 825: Why do you need to feature gate the existence of the arg instead of just setting it to `None` when the feature is disabled?

substrate/utils/wasm-builder/src/metadata_hash.rs
bkchr: The RFC was written with 15 being the minimum required version. With `15` we introduced this `metadata_at_version`. If you call this on Polkadot, you can request `v15` as return value. This was done to not always break all apps etc that parse the metadata. `metadata()` will continue returning `v14` forever.

substrate/utils/wasm-builder/src/metadata_hash.rs
bkchr: Yes

substrate/utils/wasm-builder/src/metadata_hash.rs
bkchr: Debugging for what? I mean what did you debug? Just that the return value is correct?

substrate/utils/wasm-builder/src/wasm_project.rs
bkchr: First I had it this way, but then I was doing feature gating all the other things and have done it here as well.

substrate/frame/metadata-hash-extension/src/lib.rs
bkchr: The point of `mode = 0` being that you can disable it. Not everyone will need this and I don't see any reason to force people to use this.

substrate/frame/metadata-hash-extension/src/lib.rs
bkchr: Ha! smart idea :D

substrate/frame/metadata-hash-extension/src/lib.rs
bkchr: Yeah, maybe something in general that we could fix with a better "system" for error reporting in this case. Just returning strings is also not such a great idea.

substrate/frame/metadata-hash-extension/src/lib.rs
bkchr: Yeah I thought about this as well, but `array_bytes` doesn't support const fn right now.

substrate/utils/wasm-builder/src/metadata_hash.rs
ggwpez: Yea just calling runtime APIs and inspecting the return values.

substrate/frame/metadata-hash-extension/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "frame-metadata-hash-extension"
+version = "0.1.0"
+edition = "2021"
```
ggwpez, line 4:
```suggestion
edition.workspace = true
```
CI is requiring License and Description as well.

ggwpez (approved):
LGTM

(probably waiting for audit i guess)

mustermeiszer:
@bkchr will this be backported to earlier releases of the polkadot-sdk and if so, which ones?

joepetrowski:
I think at least back to 1.7.0 for use in https://github.com/polkadot-fellows/runtimes/issues/288

mustermeiszer:
Thanks, thats great!

bkchr:
We probably even go back until 1.1. Need to check how complicated that will get.

substrate/utils/wasm-builder/src/metadata_hash.rs
bkchr: Never really needed to debug them, so I did not require anything like that so far :see_no_evil:

docs/sdk/src/guides/enable_metadata_hash.rs
```diff
@@ -0,0 +1,81 @@
...
+//! [RFC78](https://polkadot-fellows.github.io/RFCs/approved/0078-merkleized-metadata.html). At compile
+//! time the metadata hash is generated and "backed" into the runtime. This makes it extremely cheap
+//! for the runtime to verify on chain that the metadata hash is correct. By having the runtime
+//! verify the hash on chain, the user also doesn't need to trust the offchain metadata. If the
+//! metadata hash doesn't match the on chain metadata hash the transaction will be rejected. The
+//! metadata hash itself is added to the data of the transaction that is signed, this means the
+//! actual hash does not appear in the transaction. On chain the same procedure is repeated with the
+//! metadata hash that is known by the runtime and if the metadata hash doesn't match the signature
+//! verification will fail. As the metadata hash is actually the root of a merkle tree, the offline
+//! wallet can get proofs of inidivdual types to decode an transaction. This means that the offline
```
Ank4n, line 35:
```suggestion
//! wallet can get proofs of individual types to decode an transaction. This means that the offline
```

TarikGul:
> @TarikGul @josepot can you please take care to make pjs/papi ready for this new signed extension? Right now the easiest way would be to just pass `0u8` as `mode` (to disable any metadata hash checking).

Sounds good, I will make it a priority!

substrate/frame/metadata-hash-extension/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+/// environment variable is for example set by the `substrate-wasm-builder` when the feature for
+/// generating the metadata hash is enabled. If the environment variable is not set and `mode = 1`
+/// is passed, the transaction is rejected with [`UnknownTransaction::CannotLookup`].
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo, DebugNoBound)]
+#[scale_info(skip_type_params(T))]
+pub struct CheckMetadataHash<T> {
+	_phantom: core::marker::PhantomData<T>,
+	mode: Mode,
+	#[codec(skip)]
+	metadata_hash: MetadataHash,
```
muharem, line 130: May be worth having a type parameter to choose between `FetchFromEnv` and `Custom`. Now you can go with custom only in tests.

josepot:
> @TarikGul @josepot can you please take care to make pjs/papi ready for this new signed extension? Right now the easiest way would be to just pass `0u8` as `mode` (to disable any metadata hash checking).

Will do! for sure!

substrate/utils/wasm-builder/src/metadata_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+
+	let extra_info = ExtraInfo {
+		spec_version: runtime_version.spec_version,
+		spec_name: runtime_version.spec_name.into(),
+		base58_prefix,
+		decimals: extra_info.decimals,
+		token_symbol: extra_info.token_symbol,
+	};
+
+	generate_metadata_digest(&metadata, extra_info).unwrap().hash()
```
muharem, line 105: may be worth having some sanity check test for this function since it is an external library

muharem (approved):

substrate/utils/wasm-builder/src/metadata_hash.rs
bkchr: I have written tests on the other side :D

polkadot/runtime/rococo/build.rs
```diff
@@ -24,5 +24,18 @@ fn main() {
...
+fn main() {
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.enable_metadata_hash("ROC", 12)
+		.build();
+
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.set_file_name("fast_runtime_binary.rs")
+		.enable_feature("fast-runtime")
+		.enable_metadata_hash("ROC", 12)
+		.build();
```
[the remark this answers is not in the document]
bkchr: Yes

substrate/frame/metadata-hash-extension/src/lib.rs
bkchr: If you use `CheckMetadataHash::new` you will get `FetchFromEnv`.

adelove-ledger:
Is this a breaking change? If that's the case, shouldn't the tx version be updated for each chain?

bkchr:
> Is this a breaking change? If that's the case, shouldn't the tx version be updated for each chain?

I can bump this, but this will require the new ledger app any way that will not require the `transaction_version` anymore.

docs/sdk/src/guides/enable_metadata_hash.rs
```diff
@@ -0,0 +1,89 @@
...
+//! From the runtime side only the
+//! [`CheckMetadataHash`](frame_metadata_hash_extension::CheckMetadataHash) needs to be added to the
+//! list of signed extension:
+#![doc = docify::embed!("../../templates/parachain/runtime/src/lib.rs", template_signed_extra)]
+//!
+//! > **Note:**
+//! >
+//! > Adding the signed extension changes the encoding of the transaction and adds one extra byte
+//! > per
+//! > transaction!
```
kianenigma, line 55:
```suggestion
```

docs/sdk/src/guides/enable_metadata_hash.rs
```diff
@@ -0,0 +1,89 @@
...
+//!
+//! From the runtime side only the
+//! [`CheckMetadataHash`](frame_metadata_hash_extension::CheckMetadataHash) needs to be added to the
+//! list of signed extension:
+#![doc = docify::embed!("../../templates/parachain/runtime/src/lib.rs", template_signed_extra)]
+//!
+//! > **Note:**
+//! >
+//! > Adding the signed extension changes the encoding of the transaction and adds one extra byte
+//! > per
```
kianenigma, line 54:
```suggestion
//! > per transaction!
```

kianenigma (approved):
docs look good!

jsdw:
Should we add the `CheckMetadataHash` extension to the kitchensink runtime too (ie https://github.com/paritytech/polkadot-sdk/blob/b2724bd7ace6b61081ef8c35c7766522dc027cad/substrate/bin/node/runtime/src/lib.rs#L2518)?

bkchr:
> Should we add the `CheckMetadataHash` extension to the kitchensink runtime too?

Done

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-generic-ledger-app/4295/26

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-generic-ledger-app/4295/27

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-digest-13-jun-2024/8598/1

PR #4277, opened 2024-04-25
Contracts: revert reverted changes from 4266

pgherveou:
revert some reverted changes from #4266


franciscoaguirre (approved):
I thought after reverting those changes you might have wanted this 😅

athei (approved):

Issue #4003, opened 2024-04-05
HRMP notifications from relay chain should use `WrapVersion` instead of latest `VersionedXcm`


bkontur:
The HRMP channel are handled on the relay chain and the relay chain sends notifications to the child parachains:
- [HrmpNewChannelOpenRequest](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L1452-L1462)
- [HrmpChannelAccepted](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L1516-L1520)
- [HrmpChannelClosing](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L1588-L1598)

The potential problem is using `VersionedXcm::from(Xcm` which uses `xcm::latest::*` stuff. E.g. when the relay chain is migrated to the higher XCM version than the parachains, it could lead to the undecodable and `ProcessMessageError::Corrupt`  when processing message on the parachain [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/process_xcm_message.rs#L47).

Everywhere where we send XCM (ChildParachainRouter, SiblingParachainRouter, ...) we use `WrapVersion` which uses implementation from [`pallet_xcm`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/src/lib.rs#L2732-L2741) (`SupportedVersion` for destination or `safeXcmVersion` of runtime) to avoid using latest XCM version.

So the solution here is to add `WrapVersion` to the HRMP pallet [Config](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L253-L275) and use this `wrap_version` for all notifications mentioned above. The result will be that we `VersionedXcm` would be at least `safeXcmVersion` or the previously stored on in `SupportedVersion` (if parachain communicated before with the relay chain).

---

_Note: We added HRMP notification handler just recently here: https://github.com/paritytech/polkadot-sdk/pull/3696, but there could already exist custom implementations of XcmExecutor, so until this fix comes live, for them the hotfix would be to upgrade their XCM version as soon as possible according to the relay chain version (there is XCMv4 coming to the Kusama in few days)._

cc: @JuaniRios

PR #4281, opened 2024-04-25
Add support for versioned notification for HRMP pallet

bkontur:
Closes: https://github.com/paritytech/polkadot-sdk/issues/4003 (please see for the problem description)

## TODO
- [x] add more tests covering `WrapVersion` corner cases (e.g. para has lower version, ...)
- [x] regenerate benchmarks `runtime_parachains::hrmp` (fix for Rococo is here: https://github.com/paritytech/polkadot-sdk/pull/4332)

## Questions / possible improvements
- [ ] A `WrapVersion` implementation for `pallet_xcm` initiates version discovery with [note_unknown_version](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/src/lib.rs#L2527C5-L2527C25), there is possibility to avoid this overhead in this HRMP case to create new `WrapVersion` adapter for `pallet_xcm` which would not use `note_unknown_version`. Is it worth to do it or not?
- [ ] There's a possibility to decouple XCM functionality from the HRMP pallet, allowing any relay chain to generate its own notifications. This approach wouldn't restrict notifications solely to the XCM. However, it's uncertain whether it's worthwhile or desirable to do so? It means making HRMP pallet more generic. E.g. hiding HRMP notifications behind some trait:
	```
	trait HrmpNotifications {

		fn on_channel_open_request(
			sender: ParaId,
			proposed_max_capacity: u32,
			proposed_max_message_size: u32) -> primitives::DownwardMessage;

		fn on_channel_accepted(recipient: ParaId) -> primitives::DownwardMessage;

		fn on_channel_closing(initiator: ParaId, sender: ParaId, recipient: ParaId) -> primitives::DownwardMessage;
	}
	```
	and then we could have whatever adapter, `impl HrmpNotifications for VersionedXcmHrmpNotifications {...}`, 
	```
	impl parachains_hrmp::Config for Runtime {
	..
		type HrmpNotifications = VersionedXcmHrmpNotifications;
	..
	}
	```


bkontur:
bot help

bkontur:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::hrmp
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::hrmp

bkontur:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::hrmp

bkontur:
bot clean

bkontur:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::hrmp
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::hrmp

polkadot/runtime/parachains/src/hrmp.rs
```diff
@@ -278,6 +278,9 @@ pub mod pallet {
 		/// parachain.
 		type DefaultChannelSizeAndCapacityWithSystem: Get<(u32, u32)>;
 
+		/// Means of converting an `Xcm` into a `VersionedXcm`.
```
bkchr, line 281: A little bit more docs would be nice. For example that `pallet-xcm` implements this for example.

polkadot/runtime/parachains/src/hrmp.rs
```diff
@@ -1499,28 +1502,19 @@ impl<T: Config> Pallet<T> {
...
-			debug_assert!(false);
-		}
+		Self::send_to_para(
+			"init_open_channel",
+			&config,
+			recipient,
+			Self::wrap_notification(|| {
+				use xcm::opaque::latest::{prelude::*, Xcm};
+				Xcm(vec![HrmpNewChannelOpenRequest {
+					sender: u32::from(origin),
```
bkchr, line 1512:
```suggestion
					sender: origin.into(),
```

polkadot/runtime/parachains/src/hrmp.rs
```diff
@@ -1562,23 +1556,15 @@ impl<T: Config> Pallet<T> {
...
-			);
-			debug_assert!(false);
-		}
+		Self::send_to_para(
+			"accept_open_channel",
+			&config,
+			sender,
+			Self::wrap_notification(|| {
+				use xcm::opaque::latest::{prelude::*, Xcm};
+				Xcm(vec![HrmpChannelAccepted { recipient: u32::from(origin) }])
```
bkchr, line 1565:
```suggestion
				Xcm(vec![HrmpChannelAccepted { recipient: origin.into() }])
```

polkadot/runtime/parachains/src/hrmp.rs
```diff
@@ -1633,30 +1619,22 @@ impl<T: Config> Pallet<T> {
...
+		Self::send_to_para(
+			"close_channel",
+			&config,
+			opposite_party,
+			Self::wrap_notification(|| {
+				use xcm::opaque::latest::{prelude::*, Xcm};
+				Xcm(vec![HrmpChannelClosing {
+					initiator: u32::from(origin),
+					sender: u32::from(channel_id.sender),
+					recipient: u32::from(channel_id.recipient),
```
bkchr, line 1634:
```suggestion
					initiator: origin.into(),
					sender: channel_id.sender.into(),
					recipient: channel_id.recipient.into(),
```

bkchr (approved):

prdoc/pr_4281.prdoc
```diff
@@ -0,0 +1,16 @@
...
+      which controls the encoding of XCM notifications related to the opening/closing of HRMP channels.
+      If your runtime does not concern itself with the XCM version used for notifications,
+      you can set it as `type VersionWrapper = ()` to always use the latest XCM.
+      If your runtime does care about the XCM version when sending to child parachains,
+      you can provide an instance of the `pallet_xcm` with `type VersionWrapper = XcmPallet`,
+      which can manage XCM versions for destinations.
+
+crates:
+- name: polkadot-runtime-parachains
+  bump: minor
```
bkchr, line 16:
```suggestion
  bump: major
```

acatangiu (approved):

prdoc/pr_4281.prdoc
```diff
@@ -0,0 +1,16 @@
+title: "Add support for versioned notification for HRMP pallet"
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      The configuration of the HRMP pallet has been expanded to include the `VersionWrapper` type,
+      which controls the encoding of XCM notifications related to the opening/closing of HRMP channels.
+      If your runtime does not concern itself with the XCM version used for notifications,
+      you can set it as `type VersionWrapper = ()` to always use the latest XCM.
+      If your runtime does care about the XCM version when sending to child parachains,
```
franciscoaguirre, line 10: So if your runtime is a parachain you just put `()`, right?
What does it mean to concern oneself about versions? Is it allowing older versions of XCM?

franciscoaguirre (approved):

polkadot/runtime/parachains/src/hrmp.rs
bkontur: thank you, updated: https://github.com/paritytech/polkadot-sdk/pull/4281/commits/d8637569808eb3881cfa8aba9c919584acf98644

prdoc/pr_4281.prdoc
bkontur: > So if your runtime is a parachain you just put `()`, right? 

The HRMP pallet is dedicated to relay chain runtimes, I don't think that any parachain runtime uses this HRMP pallet.


> What does it mean to concern oneself about versions? Is it allowing older versions of XCM?

The problem could arise from a higher version on the relay chain than on the receiving parachain. For example, imagine that we update Polkadot with `xcm::v5`, but Polimec still uses `xcm::v4`. Without this fix, they could lose messages because they would not be able to decode `xcm::v5`.

bkontur: Perhaps it could be possible to replace most of the functionality related to the WrapVersion with just the `SendXcm` trait. In this scenario, the HRMP pallet would use `type XcmSender = ChildParachainRouter`, potentially simplifying things. However, this approach would introduce another layer of complexity, particularly regarding handling transport fees. Nevertheless, I think these HRMP notifications are treated as system notifications, which is why they are sent directly by `dmp::Pallet::<T>::queue_downward_message`.

PR #4295, opened 2024-04-25
Make parachain template great again (and async backing ready)

### LINKED ISSUE #4272: Fix parachain node template to make it async backing ready
Out [parachain node template](https://github.com/paritytech/polkadot-sdk/tree/master/templates/parachain/node) is not async backing ready, although the documentation [says the opposite](https://wiki.polkadot.network/docs/maintain-guides-async-backing). The template should use the lookahead collator to make it async backing ready.

s0me0ne-unkn0wn:
Closes #4272.

It turned out that not only the node part but also the runtime wasn't async backing ready. Now, both of them are using proper APIs and producing 6-second blocks out of the box.

### DIFF
--- a/templates/parachain/node/src/service.rs
+++ b/templates/parachain/node/src/service.rs
@@ -12,6 +12,7 @@ use parachain_template_runtime::{
 
 // Cumulus Imports
 use cumulus_client_collator::service::CollatorService;
+use cumulus_client_consensus_aura::collators::lookahead::{self as aura, Params as AuraParams};
 use cumulus_client_consensus_common::ParachainBlockImport as TParachainBlockImport;
 use cumulus_client_consensus_proposer::Proposer;
 use cumulus_client_service::{
@@ -19,7 +20,10 @@ use cumulus_client_service::{
 	BuildNetworkParams, CollatorSybilResistance, DARecoveryProfile, ParachainHostFunctions,
 	StartRelayChainTasksParams,
 };
-use cumulus_primitives_core::{relay_chain::CollatorPair, ParaId};
+use cumulus_primitives_core::{
+	relay_chain::{CollatorPair, ValidationCode},
+	ParaId,
+};
 use cumulus_relay_chain_interface::{OverseerHandle, RelayChainInterface};
 
 // Substrate Imports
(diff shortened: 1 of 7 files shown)


templates/parachain/node/src/service.rs
```diff
@@ -204,13 +209,12 @@ fn start_consensus(
 		relay_chain_slot_duration,
 		proposer,
 		collator_service,
-		// Very limited proposal time.
-		authoring_duration: Duration::from_millis(500),
-		collation_request_receiver: None,
+		authoring_duration: Duration::from_millis(1500),
```
s0me0ne-unkn0wn, line 212: Should we go ahead and bump it to 2000? All the mainnets and Rococo are already using 2.5 sec backing timeout. It's still 2 seconds for Westend.

alexggh (approved):

kianenigma (approved):
Would love to see the polkadot-launch stuff replaced here with Zombienet, but please coordinate with @gupnik if he's already working on it.

pepoviola:
> Would love to see the polkadot-launch stuff replaced here with Zombienet, but please coordinate with @gupnik if he's already working on it.

@gupnik  ping me if you need help to setup zombienet. Would be good to add examples for using both v1/v2.
Thx!

templates/parachain/node/src/service.rs
bkchr: Yes.

bkchr (approved):

templates/parachain/node/src/service.rs
skunert: I am wondering if we should even increase it to 2.5s. 

If I see it correctly we are subtracting 1/3 of the time specified here for "evaluation and block finalization": https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/basic-authorship/src/basic_authorship.rs#L291

So that means of the 2s we will hit the timeout for extrinsic application after  ~1.3s. However the runtime specified a ref time weight of 2s. So even with a correctly benchmarked runtime we could hit this limit easily, assuming reference hardware. Probably this works well in reality because our reference hardware is not super fast and most collators outperform it?

For backing a higher authoring should not be problem, because the ref_time will still limit the execution time.

s0me0ne-unkn0wn: > I am wondering if we should even increase it to 2.5s.

We're still trying to keep a gap between backing timeout and authoring duration as the backing has bigger overhead, and that's why we bumped the backing timeout in the first place -- with both being 2 sec, backers were timing out on Versi sometimes.

PR #4296, opened 2024-04-25
Unchecked extrinsic: lazily decode the Call

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/4255

I hope I understood the idea correctly.

There are a 2 things I should mention:

1. I think in the initial discussion, the idea was to store the encoded `Call` directly as a blob, without decoding it. However I think we need to decode it, in order to check if it's valid. I'm talking about [this line](https://github.com/serban300/polkadot-sdk/blob/f06ebdb8776c128fcd5a31e3748f56ddc72d250f/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L319-L320). If we're not sure if it's valid, the `decode_function()` method will have to return a `Result` and I don't know if this would be acceptable for some of the call sites. 

2. With this new approach we have to decode the `Call` each time it's needed. So there can be a performance penalty more or less significant depending on how often the decoded call is actually needed. If it's not accessed very often, this implementation should be ok. If it's accessed often, we can try to cache the decoded call also. However for caching the decoded call there are 2 options, and both are problematic:
    - modifying the signature of the `decode_call()` method to take a `mut` reference to `self`. And in many places we don't have access to the `mut UncheckedExtrinsic`
    - using a mutex for internal mutability. Doesn't seem to be easy to do this from a `no-std` environment

### DIFF
--- a/cumulus/pallets/parachain-system/src/validate_block/implementation.rs
+++ b/cumulus/pallets/parachain-system/src/validate_block/implementation.rs
@@ -240,25 +240,21 @@ where
 	B::Extrinsic: ExtrinsicCall,
 	<B::Extrinsic as Extrinsic>::Call: IsSubType<crate::Call<PSC>>,
 {
-	let xts = block
+	block
 		.extrinsics()
 		.iter()
 		// Inherents are at the front of the block and are unsigned.
 		//
 		// If `is_signed` is returning `None`, we keep it safe and assume that it is "signed".
 		// We are searching for unsigned transactions anyway.
-		.take_while(|e| !e.is_signed().unwrap_or(true));
-	for xt in xts {
-		if let Some(call) = xt.call().is_sub_type() {
-			match call {
-				crate::Call::set_validation_data { data: validation_data } =>
-					return validation_data.clone(),
-				_ => {},
-			}
-		}
-	}
-
-	panic!("Could not find `set_validation_data` inherent")
+		.take_while(|e| !e.is_signed().unwrap_or(true))
+		.filter_map(|e| match e.call().is_sub_type() {
+			Some(crate::Call::set_validation_data { data: validation_data }) =>
+				Some(validation_data.clone()),
+			_ => None,
+		})
+		.next()
+		.expect("Could not find `set_validation_data` inherent")
 }
 
 /// Validate the given [`PersistedValidationData`] against the [`MemoryOptimizedValidationParams`].
--- a/cumulus/pallets/parachain-system/src/validate_block/implementation.rs
+++ b/cumulus/pallets/parachain-system/src/validate_block/implementation.rs
@@ -235,25 +235,30 @@ where
 /// Extract the [`ParachainInherentData`].
 fn extract_parachain_inherent_data<B: BlockT, PSC: crate::Config>(
 	block: &B,
-) -> &ParachainInherentData
+) -> ParachainInherentData
 where
 	B::Extrinsic: ExtrinsicCall,
 	<B::Extrinsic as Extrinsic>::Call: IsSubType<crate::Call<PSC>>,
 {
-	block
+	let xts = block
 		.extrinsics()
 		.iter()
 		// Inherents are at the front of the block and are unsigned.
 		//
 		// If `is_signed` is returning `None`, we keep it safe and assume that it is "signed".
 		// We are searching for unsigned transactions anyway.
-		.take_while(|e| !e.is_signed().unwrap_or(true))
-		.filter_map(|e| e.call().is_sub_type())
-		.find_map(|c| match c {
-			crate::Call::set_validation_data { data: validation_data } => Some(validation_data),
-			_ => None,
-		})
-		.expect("Could not find `set_validation_data` inherent")
+		.take_while(|e| !e.is_signed().unwrap_or(true));
+	for xt in xts {
+		if let Some(call) = xt.call().is_sub_type() {
+			match call {
+				crate::Call::set_validation_data { data: validation_data } =>
+					return validation_data.clone(),
+				_ => {},
+			}
+		}
+	}
+
+	panic!("Could not find `set_validation_data` inherent")
 }
 
 /// Validate the given [`PersistedValidationData`] against the [`MemoryOptimizedValidationParams`].
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -158,8 +158,9 @@ use frame_support::{
 	migrations::MultiStepMigrator,
 	pallet_prelude::InvalidTransaction,
 	traits::{
-		BeforeAllRuntimeMigrations, ExecuteBlock, IsInherent, OffchainWorker, OnFinalize, OnIdle,
-		OnInitialize, OnPoll, OnRuntimeUpgrade, PostInherents, PostTransactions, PreInherents,
+		BeforeAllRuntimeMigrations, ExecuteBlock, ExtrinsicCall, IsInherent, OffchainWorker,
+		OnFinalize, OnIdle, OnInitialize, OnPoll, OnRuntimeUpgrade, PostInherents,
+		PostTransactions, PreInherents,
 	},
 	weights::{Weight, WeightMeter},
 };
@@ -167,7 +168,7 @@ use frame_system::pallet_prelude::BlockNumberFor;
 use sp_runtime::{
 	generic::Digest,
 	traits::{
-		self, Applyable, CheckEqual, Checkable, Dispatchable, Extrinsic, Header, NumberFor, One,
+		self, Applyable, CheckEqual, Checkable, Dispatchable, Header, NumberFor, One,
 		ValidateUnsigned, Zero,
 	},
 	transaction_validity::{TransactionSource, TransactionValidity, TransactionValidityError},
@@ -259,7 +260,7 @@ impl<
 	> ExecuteBlock<Block>
 	for Executive<System, Block, Context, UnsignedValidator, AllPalletsWithSystem, COnRuntimeUpgrade>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
+	Block::Extrinsic: Checkable<Context> + ExtrinsicCall + Codec,
 	CheckedOf<Block::Extrinsic, Context>: Applyable + GetDispatchInfo,
 	CallOf<Block::Extrinsic, Context>:
 		Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
@@ -299,7 +300,7 @@ impl<
 		COnRuntimeUpgrade: OnRuntimeUpgrade,
 	> Executive<System, Block, Context, UnsignedValidator, AllPalletsWithSystem, COnRuntimeUpgrade>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
+	Block::Extrinsic: Checkable<Context> + ExtrinsicCall + Codec,
 	CheckedOf<Block::Extrinsic, Context>: Applyable + GetDispatchInfo,
 	CallOf<Block::Extrinsic, Context>:
 		Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
@@ -521,7 +522,7 @@ impl<
 		COnRuntimeUpgrade: OnRuntimeUpgrade,
 	> Executive<System, Block, Context, UnsignedValidator, AllPalletsWithSystem, COnRuntimeUpgrade>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
+	Block::Extrinsic: Checkable<Context> + ExtrinsicCall + Codec,
 	CheckedOf<Block::Extrinsic, Context>: Applyable + GetDispatchInfo,
 	CallOf<Block::Extrinsic, Context>:
 		Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
@@ -698,7 +699,7 @@ where
 		let mut num_extrinsics = 0;
 		for (i, mut e) in extrinsics.into_iter().enumerate() {
 			num_extrinsics += 1;
-			e.pre_fetch();
+			ExtrinsicCall::cache_call(&mut e);
 
 			let is_inherent = System::is_inherent(&e);
 			if is_inherent {
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -158,9 +158,8 @@ use frame_support::{
 	migrations::MultiStepMigrator,
 	pallet_prelude::InvalidTransaction,
 	traits::{
-		BeforeAllRuntimeMigrations, EnsureInherentsAreFirst, ExecuteBlock, OffchainWorker,
-		OnFinalize, OnIdle, OnInitialize, OnPoll, OnRuntimeUpgrade, PostInherents,
-		PostTransactions, PreInherents,
+		BeforeAllRuntimeMigrations, ExecuteBlock, IsInherent, OffchainWorker, OnFinalize, OnIdle,
+		OnInitialize, OnPoll, OnRuntimeUpgrade, PostInherents, PostTransactions, PreInherents,
 	},
 	weights::{Weight, WeightMeter},
 };
@@ -168,11 +167,11 @@ use frame_system::pallet_prelude::BlockNumberFor;
 use sp_runtime::{
 	generic::Digest,
 	traits::{
-		self, Applyable, CheckEqual, Checkable, Dispatchable, Header, NumberFor, One,
+		self, Applyable, CheckEqual, Checkable, Dispatchable, Extrinsic, Header, NumberFor, One,
 		ValidateUnsigned, Zero,
 	},
-	transaction_validity::{TransactionSource, TransactionValidity},
-	ApplyExtrinsicResult, ExtrinsicInclusionMode,
+	transaction_validity::{TransactionSource, TransactionValidity, TransactionValidityError},
+	ApplyExtrinsicResult, ExtrinsicInclusionMode, RuntimeDebug,
 };
 use sp_std::{marker::PhantomData, prelude::*};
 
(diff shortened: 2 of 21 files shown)


xlc:
I think we can do a quick check on how many times the decode_function is getting called per each extrinisc? If it is 2 or 3, I think it is fine. If it is say 10, then we should optimize something.

serban300:
> I think we can do a quick check on how many times the decode_function is getting called per each extrinisc? Just add a log statement and run the node and submit transaction and check the logs. If it is 2 or 3, I think it is fine. If it is say 10, then we should optimize something.

Sounds good. Will do that.

serban300:
Added a log statement and started a Rococo + Bridge Hub zombienet. This is what I'm getting in the bridge hub collator logs for a random block for example:

```
2024-04-30 11:12:30 Executed decode_function(): "0xfb4d44fb3460263937e57f5d0bf95d2ce4340019bdab7276e6fdb554d46933f4"    
2024-04-30 11:12:30 Executed decode_function(): "0xfb4d44fb3460263937e57f5d0bf95d2ce4340019bdab7276e6fdb554d46933f4"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0xfb4d44fb3460263937e57f5d0bf95d2ce4340019bdab7276e6fdb554d46933f4"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0xfb4d44fb3460263937e57f5d0bf95d2ce4340019bdab7276e6fdb554d46933f4"    
2024-04-30 11:12:30 Executed decode_function(): "0xfb4d44fb3460263937e57f5d0bf95d2ce4340019bdab7276e6fdb554d46933f4"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0x47b2318c44183fd580190a33ee1a7a75c97cecd53de3c6e7c92903ecb3bc4969"    
2024-04-30 11:12:30 Executed decode_function(): "0x52a397e48c0f9e75afcfc18a5692591016a25b02e3bbb578c041ff6095e46960"    
2024-04-30 11:12:30 Executed decode_function(): "0x52a397e48c0f9e75afcfc18a5692591016a25b02e3bbb578c041ff6095e46960"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0x52a397e48c0f9e75afcfc18a5692591016a25b02e3bbb578c041ff6095e46960"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0x52a397e48c0f9e75afcfc18a5692591016a25b02e3bbb578c041ff6095e46960"    
2024-04-30 11:12:30 Executed decode_function(): "0x52a397e48c0f9e75afcfc18a5692591016a25b02e3bbb578c041ff6095e46960"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0xf1e033d8058a7db6def218da2e536b3b33e1adaacb0cc2cccfb9f5cb12d075c0"    
2024-04-30 11:12:30 Executed decode_function(): "0xcea9a5e826d41aea4a00f1bd9f0aae9551bdeeb2427d4d665fb6c22a9953369f"    
2024-04-30 11:12:30 Executed decode_function(): "0x0ebd58300ee0d9b2588a4b435457e9dccc54c7387a05d88f3ab42515379ebef0"    
2024-04-30 11:12:30 Executed decode_function(): "0xd10927d4c9b07f5c219bed41f1d1f4c45e37df88bc1107fc474b706c69c1a7d5"    
2024-04-30 11:12:30 Executed decode_function(): "0x0c206bf261b17f99b6f8351cb9cdd501cc85ad910587fa8bbafb927150390228"    
2024-04-30 11:12:30 Executed decode_function(): "0xbb9bd4926a0077e2c75072a2cbe9d34dc018c1d5f8f5d3dd16baaff29dd2bd13"    
2024-04-30 11:12:30 Executed decode_function(): "0xbb9bd4926a0077e2c75072a2cbe9d34dc018c1d5f8f5d3dd16baaff29dd2bd13"    
```

Some calls are decoded 5 times, some 7 times, some 1 time. I'll check where this calls come from.

serban300:
Did some optimizations [in this commit](https://github.com/paritytech/polkadot-sdk/pull/4296/commits/d8b240e9eb6593fffb88327b143d00839fa51222)

Now it's something like:
```
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0x7aa7bfde63a9072f9602b83cdbd56e9ec69ce235f2c7b7b474b5c2e574e498bb"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0xeebd7bbb74f07e7fc789b4b43390f0a33b1b3d303f2e68fd125326f8947160de"    
2024-04-30 17:38:54 Executed decode_function(): "0xe9d6757ff16002fbefb89475eccaa35add7b73c5e83f4b2f15f00fb727b66db2"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0xe9d6757ff16002fbefb89475eccaa35add7b73c5e83f4b2f15f00fb727b66db2"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0xe9d6757ff16002fbefb89475eccaa35add7b73c5e83f4b2f15f00fb727b66db2"    
2024-04-30 17:38:54 Executed decode_function(): "0xe9d6757ff16002fbefb89475eccaa35add7b73c5e83f4b2f15f00fb727b66db2"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0x7ddc64ac5789980a9cec6d5e0d557a044e60e061d021f8747c1c58638d09662a"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0x7ddc64ac5789980a9cec6d5e0d557a044e60e061d021f8747c1c58638d09662a"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0x7ddc64ac5789980a9cec6d5e0d557a044e60e061d021f8747c1c58638d09662a"    
2024-04-30 17:38:54 Executed decode_function(): "0x7ddc64ac5789980a9cec6d5e0d557a044e60e061d021f8747c1c58638d09662a"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977"    
2024-04-30 17:38:54 Executed decode_function(): "0x4b436b034c8e83ae0860d39f04827f4698f2d8f5dd4ac58cbb975084a6d6e977" 
```

The number of decodings for each function is down to around 8 times.

But from what I understand this is only happening for inherents because of calls to `check_extrinsics()` and `is_inherent()`. I guess that the inherents are normally pretty small.

serban300:
@xlc addressed/answered to your comment above. Please let me know if you have any thoughts on that.

@bkchr could you please take a look on the PR ? Any feedback would be helpful.

acatangiu (dismissed):

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -149,34 +150,31 @@ pub fn expand_outer_inherent(
...
-					// Inherents are before any other extrinsics.
-					// No module marked it as inherent thus it is not.
-					if !is_inherent {
-						break
+
+					// If there is no pallet apart from System, the true branch will be unreachable.
+					#[allow(unreachable_code)]
+					match is_inherent {
+						true => {
+							inherents.push(call);
```
svyatonik, line 158: Hope we won't ever have any inherents with `Xcm<()>` arguments :)

substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs
```diff
@@ -365,7 +391,7 @@ impl<Address: Encode, Signature: Encode, Call: Encode, Extra: SignedExtension> s
 }
 
 #[cfg(feature = "serde")]
-impl<'a, Address: Decode, Signature: Decode, Call: Decode, Extra: SignedExtension>
+impl<'a, Address: Decode, Signature: Decode, Call: Encode + Decode, Extra: SignedExtension>
```
svyatonik, line 394: Seems unneeded?

svyatonik (approved):

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
serban300: Yes. I don't know if this can happen. Also `Xcm<()>` variables can be ok as long as they are not very big.

substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs
```diff
@@ -121,22 +123,41 @@ where
...
 
-impl<Address: TypeInfo, Call: TypeInfo, Signature: TypeInfo, Extra: SignedExtension + TypeInfo>
-	Extrinsic for UncheckedExtrinsic<Address, Call, Signature, Extra>
+impl<Address, Call: Decode, Signature, Extra: SignedExtension>
+	UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+	/// Get the function that should be called in a decoded form.
+	pub fn decode_function(&self) -> Call {
+		Call::decode(&mut &self.encoded_function[..]).expect(
+			"We always check if the function is valid before setting the encoded_function field",
```
bkchr, line 150:
```suggestion
			"We always check if the function is valid before setting the encoded_function field; qed",
```

substrate/primitives/io/src/lib.rs
```diff
@@ -1822,6 +1822,37 @@ pub type SubstrateHostFunctions = (
 	transaction_index::HostFunctions,
 );
 
+/// Chainable Input implementation that reads from an underlying input
+/// and also stores the read bytes in an internal buffer.
+pub struct CopyReader<'a, I> {
```
bkchr, line 1827: Please move this next to the unchecked extrinsic and make it private.

substrate/frame/support/src/traits/misc.rs
```diff
@@ -916,30 +916,30 @@ pub trait IsInherent<Extrinsic> {
 /// An extrinsic on which we can get access to call.
 pub trait ExtrinsicCall: sp_runtime::traits::Extrinsic {
 	/// Get the call of the extrinsic.
-	fn call(&self) -> &Self::Call;
+	fn call(&self) -> Self::Call;
```
bkchr, line 919: We should probably return `Result` here to make it more obvious that this maybe includes decoding or maybe any better approach.

substrate/frame/support/procedural/src/construct_runtime/expand/inherent.rs
```diff
@@ -214,11 +212,11 @@ pub fn expand_outer_inherent(
...
+				let call = <#unchecked_extrinsic as ExtrinsicCall>::call(ext);
 				#(
 					#pallet_attrs
 					{
-						let call = <#unchecked_extrinsic as ExtrinsicCall>::call(ext);
-						if let Some(call) = IsSubType::<_>::is_sub_type(call) {
+						if let Some(call) = IsSubType::<_>::is_sub_type(&call) {
 							if <#pallet_names as ProvideInherent>::is_inherent(&call) {
 								return true;
 							}
```
bkchr, line 222: `EnsureInherentsAreFirst` below should be able to be replaced with some logic in `Executive`. Otherwise we are decoding twice.

substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs
```diff
@@ -291,10 +314,13 @@ where
 		}
 
 		let signature = is_signed.then(|| Decode::decode(input)).transpose()?;
-		let function = Decode::decode(input)?;
 
-		if let Some((before_length, after_length)) =
-			input.remaining_len()?.and_then(|a| before_length.map(|b| (b, a)))
+		let mut input = CopyReader::new(input);
+		// We have to check that the function decodes correctly.
+		Call::decode(&mut input)?;
```
bkchr, line 320: We really finally need to get some proper `skip` implemented.

cumulus/pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -235,25 +235,30 @@ where
...
-		.expect("Could not find `set_validation_data` inherent")
+		.take_while(|e| !e.is_signed().unwrap_or(true));
+	for xt in xts {
+		if let Some(call) = xt.call().is_sub_type() {
+			match call {
+				crate::Call::set_validation_data { data: validation_data } =>
+					return validation_data.clone(),
+				_ => {},
+			}
+		}
```
bkchr, line 258: Please keep the old logic with an early return on finding the `set_validation_data`. Now we decode multiple calls before we check them.

bkchr: Also you don't need to clone `validation_data` because you own the `call`

serban300 (commented):
Addressed only some of the easier comments so far

substrate/frame/executive/src/lib.rs
```diff
@@ -779,16 +779,14 @@ where
 
 		// We use the dedicated `is_inherent` check here, since just relying on `Mandatory` dispatch
 		// class does not capture optional inherents.
-		let is_inherent = System::is_inherent(&uxt);
+		if !<frame_system::Pallet<System>>::inherents_applied() && !System::is_inherent(&uxt) {
```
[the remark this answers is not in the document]
serban300: Fixed together with the comment regarding `EnsureInherentsAreFirst`. Added a way of caching the decoded function.

substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs
serban300: If you mean to just skip the encoded function bytes without decoding it, the problem is that we will have to return a `Result<>` from the methods that decode the function and:
1. I'm not sure how hard it will be to propagate this `Result<>` to all the call sites 
2. I'm not sure if it's ok to panic at a later time for example if the function can't be decoded

I will look into it

serban300: Ok, I think I finally got it. I replaced this with `Call::skip(&mut input)`, but from what I understand it still does `decode` internally. I agree we should implement a proper skip, but maybe in a separate PR ? I guess it's something that we should do in the `parity-scale-codec` crate, right ?

Issue #4073, opened 2024-04-10
Inconsistencies of XCM NonFungible adapters. And an easy-to-introduce vulnerability


mrshiposha:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

>Note: this issue describes an **easy-to-introduce** vulnerability. It's not an actual "live" vulnerability. That's why it is expressed publicly.

### 1. Different behavior of `T` and `(T,)`

Both the [NonFungible*s*Adapter](https://github.com/paritytech/polkadot-sdk/blob/92e142555d45f97aa88d241665d9952d12f4ae40/polkadot/xcm/xcm-builder/src/nonfungibles_adapter.rs#L345-L354) and the [NonFungibleAdapter](https://github.com/paritytech/polkadot-sdk/blob/92e142555d45f97aa88d241665d9952d12f4ae40/polkadot/xcm/xcm-builder/src/nonfungible_adapter.rs#L316-L325) implement the `transfer_asset` function of the `TransactAsset` trait. This leads to the inconsistency between the implementations of `TransactAsset` for `T` and for a tuple `(T,)`.
Why this happens:
* The implementation for a non-fungible adapter doesn't implement the `internal_transfer_asset` function, so [the default implementation](https://github.com/paritytech/polkadot-sdk/blob/92e142555d45f97aa88d241665d9952d12f4ae40/polkadot/xcm/xcm-executor/src/traits/transact_asset.rs#L111-L118) is used that returns the `Err(XcmError::Unimplemented)`. The non-fungible adapters implement the `transfer_asset` directly instead. The `transfer_asset` function has a non-trivial default implementation which will attempt to call the `internal_transfer_asset` first, and if failed, will resort to the combination of the `withdraw_asset` + `deposit_asset`. This default logic is substituted. So, if we call the `transfer_asset` directly on an adapter type, this particular implementation will be used.
[...]

PR #4300, opened 2024-04-25
Feature/asset ops traits

mrshiposha:
This PR introduces a new set of traits that represent different asset operations in a granular and abstract way. The main aim of the PR is to provide a granular general interface for non-fungible collections and tokens.

This is the PR I mentioned in the https://github.com/paritytech/polkadot-sdk/issues/4073 issue.

The new `asset_ops` module is added to `frame_support::traits::tokens`.
It defines the following operations on assets:
* InspectMetadata
* UpdateMetadata
* Create
* Transfer
* Destroy

Also, it defines the `AssetDefinition` trait used by all the above operations to retrieve the `Id` type of the asset (except the `Create` operation).

Each asset operation can be implemented for different asset kinds (for instance, one asset kind is a collection of NFTs and another kind is an individual non-fungible token).

Also, each operation can be implemented multiple times using different strategies associated with this operation.

For details, see the following (best viewed in the same order):
* doc comments in `substrate/frame/support/src/traits/tokens/asset_ops.rs`
* the example implementation in the `pallet-uniques`
* new XCM adapters
    * their doc comments
    * the use of the `RecreateableInstanceAdapter` for `pallet-uniques` in Rococo and Westend.

----

To keep this PR simple, it doesn't contain implementation for `pallet-nfts` and `pallet-nft-fractionalization`.
However, you can examine such implementation in [the separate PR opened in the fork](https://github.com/UniqueNetwork/polkadot-sdk/pull/8).
It also includes using the `TransferableInstanceAdapter` for `pallet-nfts` in Rococo and Westend.

Note: the `RecreateableInstanceAdapter` can't be used with `pallet-nfts` because this pallet [may burn data along with the NFT](https://github.com/paritytech/polkadot-sdk/blob/8b4cfda7589325d1a34f70b3770ab494a9d4052c/substrate/frame/nfts/src/features/create_delete_item.rs#L240-L241), so the NFT can't be fully recreated.

### DIFF
--- a/cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
+++ b/cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
@@ -45,12 +45,12 @@ use polkadot_runtime_common::xcm_sender::ExponentialPrice;
 use sp_runtime::traits::{AccountIdConversion, ConvertInto};
 use xcm::latest::prelude::*;
 use xcm_builder::{
-	AccountId32Aliases, AllowExplicitUnpaidExecutionFrom, AllowKnownQueryResponses,
-	AllowSubscriptionsFrom, AllowTopLevelPaidExecutionFrom, DenyReserveTransferToRelayChain,
-	DenyThenTry, DescribeFamily, DescribePalletTerminal, EnsureXcmOrigin,
-	FrameTransactionalProcessor, FungibleAdapter, FungiblesAdapter,
-	GlobalConsensusParachainConvertsFor, HashedDescription, IsConcrete, LocalMint,
-	NetworkExportTableItem, NoChecking, NonFungiblesAdapter, ParentAsSuperuser, ParentIsPreset,
+	unique_instances::RecreateableInstanceAdapter, AccountId32Aliases,
+	AllowExplicitUnpaidExecutionFrom, AllowKnownQueryResponses, AllowSubscriptionsFrom,
+	AllowTopLevelPaidExecutionFrom, DenyReserveTransferToRelayChain, DenyThenTry, DescribeFamily,
+	DescribePalletTerminal, EnsureXcmOrigin, FrameTransactionalProcessor, FungibleAdapter,
+	FungiblesAdapter, GlobalConsensusParachainConvertsFor, HashedDescription, InstancesOfClasses,
+	IsConcrete, LocalMint, NetworkExportTableItem, NoChecking, ParentAsSuperuser, ParentIsPreset,
 	RelayChainAsNative, SiblingParachainAsNative, SiblingParachainConvertsVia,
 	SignedAccountId32AsNative, SignedToAccountId32, SovereignSignedViaLocation, StartsWith,
 	StartsWithExplicitGlobalConsensus, TakeWeightCredit, TrailingSetTopicAsId, UsingComponents,
@@ -142,19 +142,11 @@ pub type UniquesConvertedConcreteId =
 	assets_common::UniquesConvertedConcreteId<UniquesPalletLocation>;
 
 /// Means for transacting unique assets.
-pub type UniquesTransactor = NonFungiblesAdapter<
-	// Use this non-fungibles implementation:
-	Uniques,
-	// This adapter will handle any non-fungible asset from the uniques pallet.
-	UniquesConvertedConcreteId,
-	// Convert an XCM Location into a local account id:
-	LocationToAccountId,
-	// Our chain's account ID type (we can't get away without mentioning it explicitly):
+pub type UniquesTransactor = RecreateableInstanceAdapter<
 	AccountId,
-	// Does not check teleports.
-	NoChecking,
-	// The account to use for tracking teleports.
-	CheckingAccount,
+	LocationToAccountId,
+	InstancesOfClasses<UniquesConvertedConcreteId>,
+	Uniques,
 >;
 
 /// `AssetId`/`Balance` converter for `ForeignAssets`.
--- a/polkadot/xcm/xcm-builder/src/asset_conversion.rs
+++ b/polkadot/xcm/xcm-builder/src/asset_conversion.rs
@@ -20,7 +20,9 @@ use frame_support::traits::{Contains, Get};
 use sp_runtime::traits::MaybeEquivalence;
 use sp_std::{marker::PhantomData, prelude::*, result};
 use xcm::latest::prelude::*;
-use xcm_executor::traits::{Error as MatchError, MatchesFungibles, MatchesNonFungibles};
+use xcm_executor::traits::{
+	Error as MatchError, MatchesFungibles, MatchesInstance, MatchesNonFungible, MatchesNonFungibles,
+};
 
 /// Converter struct implementing `AssetIdConversion` converting a numeric asset ID (must be
 /// `TryFrom/TryInto<u128>`) into a `GeneralIndex` junction, prefixed by some `Location` value.
@@ -149,6 +151,26 @@ impl<
 		let instance =
 			ConvertInstanceId::convert(instance).ok_or(MatchError::InstanceConversionFailed)?;
 		Ok((what, instance))
+	}
+}
+
+pub struct InstancesOfClasses<Matcher>(PhantomData<Matcher>);
+
+impl<ClassId, InstanceId, Matcher: MatchesNonFungibles<ClassId, InstanceId>>
+	MatchesInstance<(ClassId, InstanceId)> for InstancesOfClasses<Matcher>
+{
+	fn matches_instance(a: &Asset) -> result::Result<(ClassId, InstanceId), MatchError> {
+		Matcher::matches_nonfungibles(a)
+	}
+}
+
+pub struct ClasslessInstances<Matcher>(PhantomData<Matcher>);
+
+impl<InstanceId, Matcher: MatchesNonFungible<InstanceId>> MatchesInstance<InstanceId>
+	for ClasslessInstances<Matcher>
+{
+	fn matches_instance(a: &Asset) -> result::Result<InstanceId, MatchError> {
+		Matcher::matches_nonfungible(a).ok_or(MatchError::AssetNotHandled)
 	}
 }
 
--- a/polkadot/xcm/xcm-builder/src/unique_instances/backed_derivative.rs
+++ b/polkadot/xcm/xcm-builder/src/unique_instances/backed_derivative.rs
@@ -1,0 +1,172 @@
+use super::LOG_TARGET;
+use core::marker::PhantomData;
+use frame_support::traits::{
+	tokens::asset_ops::{
+		common_asset_kinds::{Class, Instance},
+		common_strategies::{DeriveIdFrom, FromTo, Owned},
+		AssetDefinition, Create, Transfer,
+	},
+	Get,
+};
+use xcm::latest::prelude::*;
+use xcm_executor::traits::{ConvertLocation, Error as MatchError, MatchesInstance, TransactAsset};
+
+/// The status of a derivative instance.
+pub enum DerivativeStatus<ClassId, InstanceId> {
+	/// The derivative can be deposited (created) in the given class.
+	DepositableIn(ClassId),
+
+	/// The derivative already exists and it has the given ID.
+	Exists(InstanceId),
+}
+
+/// The `BackedDerivativeInstanceAdapter` implements the `TransactAsset` for unique instances
+/// (NFT-like entities).
+///
+/// The adapter uses the following asset operations:
+/// * [`Create`] with the [`Owned`] strategy that uses the [`DeriveIdFrom`] id assignment.
+///     * The [`DeriveIdFrom`] accepts the value of the `Id` type retrieved from the class [asset
+///       definition](AssetDefinition)
+///     (`ClassDef` generic parameter, represents a class-like entity such as a collection of NFTs).
+/// * [`Transfer`] with [`FromTo`] strategy
+///
+/// This adapter assumes that a new asset can be created and an existing asset can be transferred.
+/// Also, the adapter assumes that the asset can't be destroyed.
+/// So, it transfers the asset to the `StashLocation` on withdrawal.
+///
+/// On deposit, the adapter consults the [`DerivativeStatus`] returned from the `Matcher`.
+/// If the asset is depositable in a certain class, it will be created within that class.
+/// Otherwise, if the asset exists, it will be transferred from the `StashLocation` to the
+/// beneficiary.
+///
+/// Transfers work as expected, transferring the asset from the `from` location to the beneficiary.
+///
+/// This adapter is meant to be used in non-reserve locations where derivatives
+/// can't be properly destroyed and then recreated.
+///
+/// For instance, an NFT engine on the chain (a pallet or a smart contract)
+/// can be incapable of recreating an NFT with the same ID.
+/// So, we can only create a derivative with a new ID.
+/// In this context, if we burn a derivative on withdrawal:
+/// 1. we could exhaust the ID space for derivatives
+/// 2. if "burning" means transferring to a special address (like the zero address in EVM),
+/// we also waste the active storage space for burned derivatives
+///
+/// To avoid that situation, the `StashLocation` is used to hold the withdrawn derivatives.
+///
+/// Also, this adapter can be used in the NFT engine that simply doesn't support burning NFTs.
+pub struct BackedDerivativeInstanceAdapter<
+	AccountId,
+	AccountIdConverter,
+	Matcher,
+	ClassDef,
+	InstanceOps,
+	StashLocation,
+>(PhantomData<(AccountId, AccountIdConverter, Matcher, ClassDef, InstanceOps, StashLocation)>);
+
+impl<AccountId, AccountIdConverter, Matcher, ClassDef, InstanceOps, StashLocation> TransactAsset
+	for BackedDerivativeInstanceAdapter<
+		AccountId,
+		AccountIdConverter,
+		Matcher,
+		ClassDef,
+		InstanceOps,
+		StashLocation,
+	> where
+	AccountIdConverter: ConvertLocation<AccountId>,
+	Matcher: MatchesInstance<DerivativeStatus<ClassDef::Id, InstanceOps::Id>>,
+	ClassDef: AssetDefinition<Class>,
+	for<'a> InstanceOps: AssetDefinition<Instance>
+		+ Create<Instance, Owned<'a, DeriveIdFrom<'a, ClassDef::Id, InstanceOps::Id>, AccountId>>
+		+ Transfer<Instance, FromTo<'a, AccountId>>,
+	StashLocation: Get<Location>,
+{
+	fn deposit_asset(what: &Asset, who: &Location, context: Option<&XcmContext>) -> XcmResult {
+		log::trace!(
+			target: LOG_TARGET,
+			"BackedDerivativeInstanceAdapter::deposit_asset what: {:?}, who: {:?}, context: {:?}",
+			what,
+			who,
+			context,
+		);
+
+		let derivative_status = Matcher::matches_instance(what)?;
+		let to = AccountIdConverter::convert_location(who)
+			.ok_or(MatchError::AccountIdConversionFailed)?;
+
+		let result = match derivative_status {
+			DerivativeStatus::DepositableIn(class_id) =>
+				InstanceOps::create(Owned::new(DeriveIdFrom::parent_id(&class_id), &to))
+					.map(|_id| ()),
+			DerivativeStatus::Exists(instance_id) => {
+				let from = AccountIdConverter::convert_location(&StashLocation::get())
+					.ok_or(MatchError::AccountIdConversionFailed)?;
+
+				InstanceOps::transfer(&instance_id, FromTo(&from, &to))
+			},
+		};
+
+		result.map_err(|e| XcmError::FailedToTransactAsset(e.into()))
+	}
+
+	fn withdraw_asset(
+		what: &Asset,
+		who: &Location,
+		maybe_context: Option<&XcmContext>,
+	) -> Result<xcm_executor::AssetsInHolding, XcmError> {
+		log::trace!(
+			target: LOG_TARGET,
+			"BackedDerivativeInstanceAdapter::withdraw_asset what: {:?}, who: {:?}, context: {:?}",
+			what,
+			who,
+			maybe_context,
+		);
+
+		let derivative_status = Matcher::matches_instance(what)?;
+		let from = AccountIdConverter::convert_location(who)
+			.ok_or(MatchError::AccountIdConversionFailed)?;
+
+		if let DerivativeStatus::Exists(instance_id) = derivative_status {
+			let to = AccountIdConverter::convert_location(&StashLocation::get())
+				.ok_or(MatchError::AccountIdConversionFailed)?;
+
+			InstanceOps::transfer(&instance_id, FromTo(&from, &to))
+				.map_err(|e| XcmError::FailedToTransactAsset(e.into()))?;
+
+			Ok(what.clone().into())
+		} else {
+			Err(XcmError::NotWithdrawable)
+		}
+	}
+
+	fn internal_transfer_asset(
+		what: &Asset,
+		from: &Location,
+		to: &Location,
+		context: &XcmContext,
+	) -> Result<xcm_executor::AssetsInHolding, XcmError> {
+		log::trace!(
+			target: LOG_TARGET,
+			"BackedDerivativeInstanceAdapter::internal_transfer_asset what: {:?}, from: {:?}, to: {:?}, context: {:?}",
+			what,
+			from,
+			to,
+			context,
+		);
+
+		let derivative_status = Matcher::matches_instance(what)?;
+		let from = AccountIdConverter::convert_location(from)
+			.ok_or(MatchError::AccountIdConversionFailed)?;
+		let to = AccountIdConverter::convert_location(to)
+			.ok_or(MatchError::AccountIdConversionFailed)?;
+
+		if let DerivativeStatus::Exists(instance_id) = derivative_status {
+			InstanceOps::transfer(&instance_id, FromTo(&from, &to))
+				.map_err(|e| XcmError::FailedToTransactAsset(e.into()))?;
+
+			Ok(what.clone().into())
+		} else {
+			Err(XcmError::NotWithdrawable)
+		}
+	}
+}
--- a/polkadot/xcm/xcm-executor/src/traits/token_matching.rs
+++ b/polkadot/xcm/xcm-executor/src/traits/token_matching.rs
@@ -101,7 +101,22 @@ impl<AssetId, Instance> MatchesNonFungibles<AssetId, Instance> for Tuple {
 		for_tuples!( #(
 			match Tuple::matches_nonfungibles(a) { o @ Ok(_) => return o, _ => () }
 		)* );
-		log::trace!(target: "xcm::matches_non_fungibles", "did not match fungibles asset: {:?}", &a);
+		log::trace!(target: "xcm::matches_non_fungibles", "did not match non-fungibles asset: {:?}", &a);
+		Err(Error::AssetNotHandled)
+	}
+}
+
+pub trait MatchesInstance<Id> {
+	fn matches_instance(a: &Asset) -> result::Result<Id, Error>;
+}
+
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+impl<Id> MatchesInstance<Id> for Tuple {
+	fn matches_instance(a: &Asset) -> result::Result<Id, Error> {
+		for_tuples!( #(
+			match Tuple::matches_instance(a) { o @ Ok(_) => return o, _ => () }
+		)* );
+		log::trace!(target: "xcm::matches_instance", "did not match an asset instance: {:?}", &a);
 		Err(Error::AssetNotHandled)
 	}
 }
--- a/substrate/frame/support/src/traits/tokens/asset_ops.rs
+++ b/substrate/frame/support/src/traits/tokens/asset_ops.rs
@@ -1,0 +1,543 @@
+//! Abstract asset operations traits.
+//!
+//! The following operations are defined:
+//! * [`InspectMetadata`]
+//! * [`UpdateMetadata`]
+//! * [`Create`]
+//! * [`Transfer`]
+//! * [`Destroy`]
+//!
+//! Also, all the operations above (except the `Create` operation) use
+//! the [`AssetDefinition`] to retrieve the `Id` type of the asset.
+//!
+//! Each asset operation can be implemented for different asset kinds
+//! such as [`Class`](common_asset_kinds::Class) and [`Instance`](common_asset_kinds::Instance).
+//!
+//! Also, an asset operation can be implemented multiple times
+//! using different strategies associated with this operation.
+//!
+//! A strategy defines the operation behavior,
+//! may supply additional parameters,
+//! and may define a return value type of the operation.
+
+use crate::dispatch::DispatchResult;
+use core::marker::PhantomData;
+use sp_runtime::DispatchError;
+use sp_std::vec::Vec;
+
+/// Trait for defining an asset of a certain kind.
+/// The definition must provide the `Id` type to identify the asset.
+///
+/// The common asset kinds are:
+/// * The [`Class`](common_asset_kinds::Class) asset kind is of assets that resemble class-like
+/// entities. For example, a collection of non-fungible tokens belongs to this kind.
+/// * The [`Instance`](common_asset_kinds::Instance) asset kind is of assets that resemble concrete
+/// instances of something. For example, a non-fungible token (which may or may not be part of a
+/// certain class) belongs to this kind.
+///
+/// Other asset kinds can be defined.
+pub trait AssetDefinition<AssetKind> {
+	/// Type for identifying the asset.
+	type Id;
+}
+
+/// A strategy for use in the [`InspectMetadata`] implementations.
+///
+/// The common inspect strategies are:
+/// * [`Bytes`](common_strategies::Bytes)
+/// * [`Ownership`](common_strategies::Ownership)
+/// * [`CanCreate`](common_strategies::CanCreate)
+/// * [`CanTransfer`](common_strategies::CanTransfer)
+/// * [`CanDestroy`](common_strategies::CanDestroy)
+/// * [`CanUpdateMetadata`](common_strategies::CanUpdateMetadata)
+pub trait MetadataInspectStrategy {
+	/// The type to return from the [`InspectMetadata::inspect_metadata`] function.
+	type Value;
+}
+
+/// A trait representing the ability of a certain asset kind to **provide** its metadata
+/// information.
+///
+/// This trait can be implemented multiple times using different [`inspect
+/// strategies`](MetadataInspectStrategy).
+///
+/// An inspect strategy defines how the asset metadata is identified/retrieved
+/// and what [`Value`](MetadataInspectStrategy::Value) type is returned.
+pub trait InspectMetadata<AssetKind, Strategy: MetadataInspectStrategy>:
+	AssetDefinition<AssetKind>
+{
+	/// Inspect metadata information of the asset
+	/// using the given `id` and the inspect `strategy`.
+	///
+	/// The ID type is retrieved from the [`AssetDefinition`].
+	fn inspect_metadata(
+		id: &Self::Id,
+		strategy: Strategy,
+	) -> Result<Strategy::Value, DispatchError>;
+}
+
+/// A strategy for use in the [`UpdateMetadata`] implementations.
+///
+/// The common update strategies are:
+/// * [`Bytes`](common_strategies::Bytes)
+/// * [`CanCreate`](common_strategies::CanCreate)
+/// * [`CanTransfer`](common_strategies::CanTransfer)
+/// * [`CanDestroy`](common_strategies::CanDestroy)
+/// * [`CanUpdateMetadata`](common_strategies::CanUpdateMetadata)
+pub trait MetadataUpdateStrategy {
+	/// The type of metadata update to accept in the [`UpdateMetadata::update_metadata`] function.
+	type Update<'u>;
+}
+
+/// A trait representing the ability of a certain asset kind to **update** its metadata information.
+///
+/// This trait can be implemented multiple times using different [`update
+/// strategies`](MetadataUpdateStrategy).
+///
+/// An update strategy defines how the asset metadata is identified
+/// and what [`Update`](MetadataUpdateStrategy::Update) type is used.
+pub trait UpdateMetadata<AssetKind, Strategy: MetadataUpdateStrategy>:
+	AssetDefinition<AssetKind>
+{
+	/// Update metadata information of the asset
+	/// using the given `id`, the update `strategy`, and the `update` value.
+	///
+	/// The ID type is retrieved from the [`AssetDefinition`].
+	fn update_metadata(
+		id: &Self::Id,
+		strategy: Strategy,
+		update: Strategy::Update<'_>,
+	) -> DispatchResult;
+}
+
+/// A strategy for use in the [`Create`] implementations.
+///
+/// The common "create" strategies are:
+/// * [`Owned`](common_strategies::Owned)
+/// * [`Adminable`](common_strategies::Adminable)
+pub trait CreateStrategy {
+	/// This type represents successful asset creation.
+	/// It will be the return type of the [`Create::create`] function.
+	type Success;
+}
+
+/// An ID assignment approach to use in the "create" strategies.
+///
+/// The common ID assignments are:
+/// * [`AutoId`](common_strategies::AutoId)
+/// * [`PredefinedId`](common_strategies::PredefinedId)
+/// * [`DeriveIdFrom`](common_strategies::DeriveIdFrom)
+pub trait IdAssignment {
+	/// The reported ID type.
+	///
+	/// Examples:
+	/// * [`AutoId`](common_strategies::AutoId) returns ID of the newly created asset
+	/// * [`PredefinedId`](common_strategies::PredefinedId) returns `()` since the ID is already
+	///   defined
+	/// * [`DeriveIdFrom`](common_strategies::DeriveIdFrom) returns the derived ID
+	type ReportedId;
+}
+
+/// A trait representing the ability of a certain asset kind to be created.
+///
+/// This trait can be implemented multiple times using different [`"create"
+/// strategies`](CreateStrategy).
+///
+/// A create strategy defines all aspects of asset creation including how an asset ID is assigned.
+pub trait Create<AssetKind, Strategy: CreateStrategy> {
+	/// Create a new asset using the provided `strategy`.
+	fn create(strategy: Strategy) -> Result<Strategy::Success, DispatchError>;
+}
+
+/// A strategy for use in the [`Transfer`] implementations.
+///
+/// The common transfer strategies are:
+/// * [`JustTo`](common_strategies::JustTo)
+/// * [`FromTo`](common_strategies::FromTo)
+pub trait TransferStrategy {}
+
+/// A trait representing the ability of a certain asset kind to be transferred.
+///
+/// This trait can be implemented multiple times using different [`transfer
+/// strategies`](TransferStrategy).
+///
+/// A transfer strategy defines transfer parameters.
+pub trait Transfer<AssetKind, Strategy: TransferStrategy>: AssetDefinition<AssetKind> {
+	/// Transfer the asset identified by the given `id` using the provided `strategy`.
+	///
+	/// The ID type is retrieved from the [`AssetDefinition`].
+	fn transfer(id: &Self::Id, strategy: Strategy) -> DispatchResult;
+}
+
+/// A strategy for use in the [`Destroy`] implementations.
+///
+/// The common destroy strategies are:
+/// * [`JustDestroy`](common_strategies::JustDestroy)
+/// * [`IfOwnedBy`](common_strategies::IfOwnedBy)
+/// * [`WithWitness`](common_strategies::WithWitness)
+/// * [`IfOwnedByWithWitness`](common_strategies::IfOwnedByWithWitness)
+pub trait DestroyStrategy {
+	/// This type represents successful asset destruction.
+	/// It will be the return type of the [`Destroy::destroy`] function.
+	type Success;
+}
+
+/// A trait representing the ability of a certain asset kind to be destroyed.
+///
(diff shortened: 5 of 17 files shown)


mrshiposha:
CC @franciscoaguirre

polkadot/xcm/xcm-builder/src/unique_instances/backed_derivative.rs
```diff
@@ -0,0 +1,172 @@
...
+/// So, we can only create a derivative with a new ID.
+/// In this context, if we burn a derivative on withdrawal:
+/// 1. we could exhaust the ID space for derivatives
+/// 2. if "burning" means transferring to a special address (like the zero address in EVM),
+/// we also waste the active storage space for burned derivatives
+///
+/// To avoid that situation, the `StashLocation` is used to hold the withdrawn derivatives.
+///
+/// Also, this adapter can be used in the NFT engine that simply doesn't support burning NFTs.
+pub struct BackedDerivativeInstanceAdapter<
```
mrshiposha, line 58: Ignore this for now, please. It will be changed/removed.

mrshiposha:
@franciscoaguirre @xlc This PR adds new granular NFT traits to frame-support and provides new XCM types, giving us the instruments to implement XCM NFT in any chain regardless of differences in NFT solutions used in the ecosystem (different pallets, which represent NFTs differently or incompatible with each other, or smart contract-based solutions).

This PR could undoubtedly be divided into at least two. However, I decided to provide all the pieces at once so we can discuss the complete picture. I can separate the PR if needed.

polkadot/xcm/pallet-xnft/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+		type RuntimeEvent: From<Event<Self, I>>
+			+ IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+		type DerivativeIdParams: Member + Parameter + MaxEncodedLen;
+		type DerivativeId: Member + Parameter + MaxEncodedLen;
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn foreign_asset_to_derivative_id_params)]
+	pub type ForeignAssetToDerivativeIdParams<T: Config<I>, I: 'static = ()> =
```
franciscoaguirre, line 71: Why is it necessary to map `AssetId` to a `DerivativeId`. Is it possible to just use `AssetId`? As we do with `ForeignAssets` in asset hub.

polkadot/xcm/xcm-builder/src/unique_instances/ops.rs
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+/// The `StashOnDestroy` implements the [`Destroy`] operation by utilizing the [`Stash`] operation.
+/// The destroy operation is implemented using the [`IfOwnedBy`] strategy
+/// and  modeled by the [`Stash`] operation using the same strategy.
+///
+/// The implemented [`Destroy`] operation can be used in the
+/// [`UniqueInstancesAdapter`](super::UniqueInstancesAdapter) via the [`UniqueInstancesOps`].
+pub struct StashOnDestroy<InstanceOps>(PhantomData<InstanceOps>);
```
franciscoaguirre, line 140: I like the idea of implementing destroy using a stash instead of burning the asset. However, should stash be its own operation? I also have a hard time wrapping my head around strategies, some seem to make sense, but some seem like they're more implementation details

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
```diff
@@ -101,7 +101,22 @@ impl<AssetId, Instance> MatchesNonFungibles<AssetId, Instance> for Tuple {
...
 			match Tuple::matches_nonfungibles(a) { o @ Ok(_) => return o, _ => () }
 		)* );
-		tracing::trace!(target: "xcm::matches_non_fungibles", asset = ?a, "did not match fungibles asset");
+		tracing::trace!(target: "xcm::matches_non_fungibles", asset = ?a, "did not match non-fungibles asset");
+		Err(Error::AssetNotHandled)
+	}
+}
+
+pub trait MatchesInstance<Id> {
+	fn matches_instance(a: &Asset) -> result::Result<Id, Error>;
```
franciscoaguirre, line 110: This is an interesting addition. Some NFTs do indeed come without a collection, like coretime regions. Do they use the same `AssetId` and an index or do they use different `AssetId`s and the `Undefined` instance, which is supposed to be used for NFTs with only one instance?

substrate/frame/support/src/traits/tokens/asset_ops.rs
```diff
@@ -0,0 +1,646 @@
...
+
+	/// The `Ownership` [inspect](MetadataInspectStrategy) metadata strategy allows getting the
+	/// owner of an asset.
+	pub struct Ownership<Owner>(PhantomData<Owner>);
+	impl<Owner> Default for Ownership<Owner> {
+		fn default() -> Self {
+			Self(PhantomData)
+		}
+	}
+	impl<Owner> MetadataInspectStrategy for Ownership<Owner> {
```
franciscoaguirre, line 374: I see what you mean by "strategy" here, it's more of a view inside the metadata. You specify some values that are reasonable for the metadata to have, and a catch-all `bytes` one for custom stuff. It's maybe more of a `Field`?

franciscoaguirre (commented):
Did a shallow pass on the adapter, ops and strategies, I'll make more review passes

polkadot/xcm/pallet-xnft/src/lib.rs
mrshiposha: This is not strictly necessary. In fact, chains like AssetHub can use `AssetId` everywhere.

However, I introduced this optional pallet that can map `AssetId` to a different `DerivativeId` to support the use case of mapping foreign NFTs to the already existing NFT engine on a given chain. For instance, we at Unique want a unified interface for working with all kinds of NFTs on the chain, both ours and foreign ones. E.g., we want to use the existing methods for nesting and fractionalization.

This way, our clients can think about a single interface without worrying about the kind of token they are working with.

mrshiposha: Maybe the pallet's name should be refined to reflect its optionality.

polkadot/xcm/xcm-builder/src/unique_instances/ops.rs
mrshiposha: I thought the `Stash` / `Restore` operations could be reasonably frequent. For instance, the fractionalization operation implies "stashing" the NFT and depositing associated fungibles (and symmetrically, burning fungibles and "restoring" the NFT).

The non-fungibles could be stashed during other, more high-level operations (like fractionalization). For instance, an NFT could be used as an entry ticket to a dedicated on-chain logic, and this ticket is considered "active" when the NFT is stashed. I feel that more use cases could be derived around stashing/restoring.

Nonetheless, I agree that these two strategies are more semantically complex than `Transfer` or `Destroy`.

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
mrshiposha: I think it depends on the chain that provides such NFTs :)

Yet, I'd like to see the following principle. Suppose we have collection-less NFTs on a chain. In that case, in XCM, they should be identified by the same `AssetId` and varying `AssetInstance` if these NFTs are different instances of the same "type" (where "type" is defined by the operations you can do with its objects).

So, applying this principle, the combination of `AssetId` and `AssetInstance::Undefined` should identify a singleton object with associated operations unique to this particular asset.

In short, I'd say this:
* `AssetId` should identify the NFT engine (and possibly some interior "group" of NFT in that engine, like a collection)
* `AssetInstance` should identify the particular NFT instance inside the specified NFT engine.
Note: I assume the chain can host multiple NFT engines that provide different NFT-associated operations.

substrate/frame/support/src/traits/tokens/asset_ops.rs
mrshiposha: I agree. It indeed feels like a `Field` or something.

polkadot/xcm/pallet-xnft/src/lib.rs
mrshiposha: >I don't think it belongs in this repository then

Providing support for derivatives of a general form seems like a good thing. Also, the pallet is minimal, so maintaining it will be easy.

polkadot/xcm/xcm-builder/src/unique_instances/ops.rs
mrshiposha: It could be. To me, the `Stash` seems more general than the `Lock`. Maybe I'm wrong, but it looks like the `Lock` should preserve ownership (the `Unlock` should return the token to the before-lock-owner). `Stash` isn't obliged to provide such guarantees; hence, it is more general.

Polkadot-Forum:

https://forum.polkadot.network/t/persistent-parachains-for-a-long-term-testnet-on-v1-14/9886/1

polkadot/xcm/xcm-builder/src/unique_instances/ops.rs
jsidorenko: I do agree with @mrshiposha here, `Stash` is more general or even abstract compared to `Lock`. You could even delete the asset on stash and restore it later, while locking assumes the asset will always be here under the same owner until you unlock it.

jsidorenko: I'm wondering whether this could be simply named `OnDestroy`, since stashing is an implementation detail

mrshiposha: I believe these adapters can be simplified when #5620 is merged.

For instance, the `Stash` trait can be used directly in the primary adapter (`UniqueInstancesAdapter`) instead of the `Destroy` trait.
So, there will be no need for the additional  `StashOnDestroy` type.

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
jsidorenko: That's very confusing... `AssetId` is a synonym of `AssetIdentifier`, which mentally sounds nearly the same as `AssetInstance`.
It becomes even more confusing when you say the `AssetId` could identify some collection, while `AssetInstance` will act as the asset's identifier (a particular NFT) :)

mrshiposha: I don't like it either, but this terminology already exists and is documented in XCM.

See the [`Asset` struct](https://github.com/paritytech/polkadot-sdk/blob/27fbb06d4b17178da50cc043ef5ec60b0cdfead8/polkadot/xcm/src/v5/asset.rs#L446-L452) and [`Fungibility::NonFungible`](https://github.com/paritytech/polkadot-sdk/blob/27fbb06d4b17178da50cc043ef5ec60b0cdfead8/polkadot/xcm/src/v5/asset.rs#L267-L269) enum variant.

gztensor:
I see quite a few very useful changes here:

- Traits for flexible NFT ID handling
- Support for creation of derivative NFTs -> a fun example use case is enabled when NFT created on one network nests inside an NFT created in another network
- UniqueInstancesAdapter for reserve-based NFT transfers nicely abstracts NFT types

All good changes overall!

xlc:
LGTM

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
franciscoaguirre: This was created back when all NFTs were assumed to be in a collection. Fungible assets have an id and an amount, non-fungible ones have a collection id and an instance.

I feel like collection-less assets could just be an `AssetId` with the `Undefined` instance.

polkadot/xcm/pallet-derivatives/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = "pallet-derivatives"
```
franciscoaguirre:
```suggestion
name = "pallet-xcm-derivatives"
```

Even though it's in the `xcm` folder, the crate name has to include it.

polkadot/xcm/pallet-derivatives/README.md
```diff
@@ -0,0 +1 @@
+# derivatives
```
franciscoaguirre: I would add some README detailing the purpose of this and how to use it.

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+//!
+//! The `pallet-derivatives` can be helpful when another pallet, which hosts the derivative assets,
+//! doesn't provide a good enough way to create new assets in the context of them being derivatives.
+//!
+//! This mainly concerns derivative NFT collections because they should be creatable by an
+//! unprivileged user, in contrast to how fungible derivative assets are usually registered using a
+//! privileged origin (root or some collective).
+//!
+//! Fungible derivatives require a privileged origin to be registered since they could be used as
+//! fee payment assets. Conversely, Derivative NFT collections contain unique derivative objects
```
franciscoaguirre: This is, not true? While they could be used for fee payment, they can only be used like that if they're present in a liquidity pool with DOT in the case of AH. Other chains could configure it differently and might even need them to be registered from a privileged origin, but it's not at all mandatory.

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+//! that don't affect the chain's fee system. They don't represent a payment asset but rather some
```
franciscoaguirre: They could actually be used for fee payment. A chain might create a system with NFT coupons for free execution.

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+//!
+//! However, a local NFT-hosting pallet might not provide a way for a regular user to create a
+//! derivative collection without giving that user ownership and collection configuration
+//! capabilities (e.g., `pallet-nfts` creates a collection owned by the transaction signer and
+//! always requires the signer to supply the collection's initial configuration). A regular user
+//! should not be able to influence the properties of a derivative collection and, of course, should
+//! not be its owner. The chain itself should manage such a collection in some way (e.g., by
+//! providing ownership to the reserve location's sovereign account).
+//!
+//! In this case, an intermediate API is needed to create derivative NFT collections safely. The
```
franciscoaguirre:
```suggestion
//! In this case, an intermediate API is needed to create derivative NFT collections safely.
```

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+//!
+//! Fungible derivatives require a privileged origin to be registered since they could be used as
+//! fee payment assets. Conversely, Derivative NFT collections contain unique derivative objects
+//! that don't affect the chain's fee system. They don't represent a payment asset but rather some
+//! logical entity that can interact with the given chain's functionality (e.g., NFT
+//! fractionalization). These interactions can be the reason why a user might want to transfer an
+//! NFT.
+//!
+//! Requiring a privileged origin in this case is raising an unreasonable barrier for NFT
+//! interoperability between chains.
```
franciscoaguirre: Why do you want anyone to create a derivative NFT representing any real remote NFT?

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+
+	fn destroy_derivative() -> Weight {
+		Weight::from_parts(100_000_000, 0)
+	}
+}
+
+/// The `NoStoredMapping` adapter calls the `CreateOp` (which should take the `Original` value and
+/// return a `Derivative` one) and returns `None`, indicating that the mapping between the original
+/// and the derivative shouldn't be saved.
+pub struct NoStoredMapping<CreateOp>(PhantomData<CreateOp>);
```
franciscoaguirre: Why might you want to not store the mapping? Because the ids are the same? Can you add that to this doc comment?

franciscoaguirre (commented):
I did a first iteration, didn't review the whole PR. Will continue reviewing later.

acatangiu (approved):

polkadot/xcm/pallet-derivatives/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+type OriginalOf<T, I> = <T as Config<I>>::Original;
+type DerivativeOf<T, I> = <T as Config<I>>::Derivative;
+type DerivativeExtraOf<T, I> = <T as Config<I>>::DerivativeExtra;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+
+	/// The in-code storage version.
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```
bkontur: not needed for new pallet
```suggestion
```

polkadot/xcm/xcm-builder/src/asset_conversion.rs
```diff
@@ -158,6 +160,26 @@ impl<
 	}
 }
 
+pub struct MatchInClassInstances<Matcher>(PhantomData<Matcher>);
```
bkontur, line 171: please add some docs here for `pub` stuff

polkadot/xcm/pallet-derivatives/Cargo.toml
bkontur: ~Yes, agree `pallet-derivatives` is  not a good name, `pallet-xcm-derivatives`, what about `pallet-asset-derivatives` or `pallet-xcm-asset-derivatives`.~

bkontur: > Even though it's in the `xcm` folder

Why does it even need to be under the `xcm` folder? Is it really part of the `xcm` module?

From what I see, the pallet itself only uses a few imports from `xcm-builder`:
```rust
use xcm_builder::unique_instances::{
	derivatives::{DerivativesRegistry, IterDerivativesRegistry},
	DerivativesExtra,
};
```

For example, the `DerivativesRegistry` trait and most of the code in `xcm-builder/src/unique_instances/derivatives.rs` don’t actually use any `xcm` types. This could easily be moved out of the XCM modules — maybe into `frame-support/traits` or somewhere else.

So at first glance, I’d say this pallet is entirely XCM-agnostic.

As a comparison, `pallet-revive` and `pallet-contracts` also use `xcm-builder`, and they live under `substrate/frame`.
There are quite a lot of new traits and structs introduced in this PR — I’d prefer to do some triage/clean-up and move things around a bit.

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
```diff
@@ -0,0 +1,234 @@
...
+/// * a map between a [`NonFungibleAsset`] and a derivative instance ID
+/// to create a new derivative instance
+pub trait DerivativesRegistry<Original, Derivative> {
+	fn try_register_derivative(original: &Original, derivative: &Derivative) -> DispatchResult;
+
+	fn try_deregister_derivative_of(original: &Original) -> DispatchResult;
+
+	fn get_derivative(original: &Original) -> Result<Derivative, DispatchError>;
+
+	fn get_original(derivative: &Derivative) -> Result<Original, DispatchError>;
```
bkontur: First of all, I don't like exposing `DispatchResult` and `DispatchError` in the XCM traits here. If you look at the `DispatchError` variants, which one would you use for `try_register_derivative`? In most cases, you'll likely end up probably using `DispatchError::Other`. XCM modules should be as independent of FRAME-specific types as possible. (I know we use some - for example, `xcm-builder` only uses `DispatchResult` for testing salary - ~which I also think should be removed and replaced also~).

XCM processing isn't about dispatchables; it's about interpreting and executing XCM programs and instructions. So if this logic is truly XCM-related, you should use something like `Result<T, XcmError>` instead.

In fact, the `DerivativesRegistry` and related logic seem more FRAME-oriented and could probably be moved out of the XCM module altogether, along with the pallet derivatives, as I mentioned earlier.


```
/// This is the legacy return type of `Dispatchable`. It is still exposed for compatibility reasons.
/// The new return type is `DispatchResultWithInfo`. FRAME runtimes should use
/// `frame_support::dispatch::DispatchResult`.
pub type DispatchResult = core::result::Result<(), DispatchError>;

/// Reason why a dispatch call failed.
pub enum DispatchError {
```

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
```diff
@@ -0,0 +1,234 @@
...
+/// The `DerivativeToOriginalConvert` uses the provided [DerivativesRegistry] to
+/// convert the `Derivative` value to the `Original` one.
+pub struct DerivativeToOriginalConvert<R>(PhantomData<R>);
+impl<Original, Derivative, R: DerivativesRegistry<Original, Derivative>>
+	FallibleConvert<Derivative, Original> for DerivativeToOriginalConvert<R>
+{
+	fn fallible_convert(a: Derivative) -> Result<Original, DispatchError> {
+		R::get_original(&a)
+	}
+}
```
bkontur: I think all this stuff around `FallibleConvert` is not really necessary, and everything what is needed can be done with existing tools e.g. `TryConvert` / `TryConvertBack`.

For example, if you use `TryConvertBack`, then we don't need two `OriginalToDerivativeConvert` and `DerivativeToOriginalConvert` - you can have just one `OriginalToDerivativeConvert` which implements both `TryConvert` and `TryConvertBack` - this way + removing `FallibleConvert` - lots of code can go away, which just simplifies the stuff.

```suggestion
```

For `FallibleConvert`, if the conversion fails, which variant are you planning to use?
```
pub enum DispatchError {
	/// Some error occurred.
	Other(
		#[codec(skip)]
		#[cfg_attr(feature = "serde", serde(skip_deserializing))]
		&'static str,
	),
	/// Failed to lookup some data.
	CannotLookup,
	/// A bad origin.
	BadOrigin,
	/// A custom error in a module.
	Module(ModuleError),
	/// At least one consumer is remaining so the account cannot be destroyed.
	ConsumerRemaining,
	/// There are no providers so the account cannot be created.
	NoProviders,
	/// There are too many consumers so the account cannot be created.
	TooManyConsumers,
	/// An error to do with tokens.
	Token(TokenError),
	/// An arithmetic error.
	Arithmetic(ArithmeticError),
	/// The number of transactional layers has been reached, or we are not in a transactional
	/// layer.
	Transactional(TransactionalError),
	/// Resources exhausted, e.g. attempt to read/write data which is too large to manipulate.
	Exhausted,
	/// The state is corrupt; this is generally not going to fix itself.
	Corruption,
	/// Some resource (e.g. a preimage) is unavailable right now. This might fix itself later.
	Unavailable,
	/// Root origin is not allowed.
	RootNotAllowed,
	/// An error with tries.
	Trie(TrieError),
}
```

polkadot/xcm/pallet-derivatives/src/lib.rs
mrshiposha: True, it could be. Though there is no real-world example at the moment, the XCM Payment API doesn't support that. Additionally, I believe there will be issues with weight refunds.
But yeah, still possible. I need to reformulate this.

polkadot/xcm/pallet-derivatives/src/lib.rs
mrshiposha: Moving NFT assets across chains securely via XCM is a very innovative capability (bridges aside), and use cases for it are basically unknown.
In such a case, the loose permissions should encourage the development of use cases, which can then drive the limitations that should be put on rights to create the derivative collection. The capabilities to accomplish that technically will grow over time, and it will become easier to implement the fine-tuned rights management if the need arises. At this stage, the most important thing is to get use cases in the first place. Therefore, it seems there is no need to create extra limitations at this moment.

In the future, we could add the following safeguards:
1. `create_derivative` will only start the registration process, sending the `ReportMetadata` XCM instruction to the corresponding chain and registering a handler-extrinsic that will finalize the registration.
2. The handler-extrinsic can check the specific metadata fields and decide if it is okay to register such a derivative collection.
3. Also, the reserve chain could decide not to report the metadata, effectively forbidding the creation of a derivative collection (e.g., such a chain could provide a way for a collection's owner to forbid metadata reporting).

This way, the registration process will be very simple without the need to involve any voting process, yet it will be secure. But given that XCM Asset Metadata isn't here yet, I'd suggest starting to experiment with a simple unprivileged registration.

polkadot/xcm/pallet-derivatives/src/lib.rs
mrshiposha: Yes, when IDs are the same, we don't need a mapping. In this scenario, the pallet only provides an API for derivative creation/destruction.

I'll highlight this in the doc comment

Polkadot-Forum:

https://forum.polkadot.network/t/2025-06-20-technical-fellowship-opendev-call/13443/1

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
mrshiposha: >For FallibleConvert, if the conversion fails, which variant are you planning to use?

The `Module` variant. The `FallibleConvert` is supposed to propagate pallet errors.

For instance, as it is used in the [`DeriveStrategyThenCreate`](https://github.com/UniqueNetwork/polkadot-sdk/blob/5ee9516eb4239399f1dedba4db9411fe0f17a280/substrate/frame/support/src/traits/tokens/asset_ops/utils.rs#L205-L226) type.
The NFT traits (`Create`, `Update`, `Destroy`, etc.) are supposed to usually be called from an extrinsic context, so they return a `DispatchError` when something goes wrong (containing a `Module` error since it will always be some pallet that implements the corresponding NFT operation).
The `FallibleConvert`'s purpose is to return a proper pallet error, so we could return it inside an NFT operation.

For example, [in the pallet tests](https://github.com/UniqueNetwork/polkadot-sdk/blob/5ee9516eb4239399f1dedba4db9411fe0f17a280/polkadot/xcm/pallet-derivatives/src/mock/mod.rs#L260-L263), the `CreateDerivativeOwnedBySovAcc` instance will return the `pallet_derivatives::Error::InvalidAsset` error if there is no way to get a sovereign account for the given asset ID, and the error will be propagated through the `create_derivative` extrinsic back to the client.
Also, if an error occurs inside the NFT engine, it will be propagated as well.

Another util type where `FallibleConvert` is used is [`MapId`](https://github.com/UniqueNetwork/polkadot-sdk/blob/5ee9516eb4239399f1dedba4db9411fe0f17a280/substrate/frame/support/src/traits/tokens/asset_ops/utils.rs#L106-L125). `FallibleConvert` helps to avoid using `DispatchError::Other` inside MapId's NFT traits implementations.
[In the pallet tests](https://github.com/UniqueNetwork/polkadot-sdk/blob/5ee9516eb4239399f1dedba4db9411fe0f17a280/polkadot/xcm/pallet-derivatives/src/mock/mod.rs#L377-L386), the `OriginalToDerivativeConvert` is used where `FallibleConvert` is needed.  
(However, I acknowledge that in the XCM context, it is of little use since if an error occurs, it will be converted to `FailedToTransactAsset` anyway. But it could be useful if this were called from an extrinsic, not from an XCM context, as in the case above).

I'm all for improving the error reporting and don't mind refactoring this. I just want to be able to propagate the pallet errors in NFT operations so that extrinsics where they are called could report the proper errors (especially when operations are nested under other pallets' APIs, such as pallet-nft-fractionalization).

polkadot/xcm/pallet-derivatives/src/tests.rs
```diff
@@ -0,0 +1,349 @@
...
+		let id = AssetId(Location::new(1, [Parachain(1111), PalletInstance(42), GeneralIndex(1)]));
+
+		// An invalid origin is rejected.
+		assert_err!(
+			PredefinedIdDerivativeCollections::create_derivative(RuntimeOrigin::none(), id.clone()),
+			DispatchError::BadOrigin,
+		);
+
+		PredefinedIdDerivativeCollections::create_derivative(RuntimeOrigin::signed(1), id.clone())
+			.unwrap();
```
bkontur:
```suggestion
		assert_ok!(PredefinedIdDerivativeCollections::create_derivative(RuntimeOrigin::signed(1), id.clone()));
```

polkadot/xcm/pallet-derivatives/src/tests.rs
```diff
@@ -0,0 +1,349 @@
...
+		assert_err!(
+			PredefinedIdDerivativeCollections::destroy_derivative(
+				RuntimeOrigin::signed(1),
+				id.clone()
+			),
+			DispatchError::BadOrigin,
+		);
+
+		PredefinedIdDerivativeCollections::destroy_derivative(RuntimeOrigin::root(), id.clone())
+			.unwrap();
```
bkontur:
```suggestion
		assert_ok!(PredefinedIdDerivativeCollections::destroy_derivative(RuntimeOrigin::root(), id.clone()));
```

polkadot/xcm/pallet-derivatives/src/tests.rs
```diff
@@ -0,0 +1,349 @@
...
+		let id_b =
+			AssetId(Location::new(1, [Parachain(3333), PalletInstance(42), GeneralIndex(2)]));
+
+		// An invalid origin is rejected.
+		assert_err!(
+			AutoIdDerivativeCollections::create_derivative(RuntimeOrigin::none(), id_a.clone()),
+			DispatchError::BadOrigin,
+		);
+
+		AutoIdDerivativeCollections::create_derivative(RuntimeOrigin::signed(3), id_a.clone())
```
bkontur: @mrshiposha please use better `asset_ok` instead of unwrap where possible

polkadot/xcm/pallet-derivatives/src/lib.rs
bkontur: @mrshiposha I see that you’ve put quite a bit of work into this — it’s really cool, and I like it. I’d definitely like to support you on this. To move forward, let’s start by moving `pallet-derivatives` out of the XCM subfolder - for example, to `polkadot-sdk/substrate/frame/derivatives`, or if it’s specifically about NFTs, then to `polkadot-sdk/substrate/frame/nfts/derivatives`:
- It currently doesn’t provide any XCM functionality (yet). If I understand correctly, it’s more of a FRAME-level pallet used to store some mappings, without any cross-chain or consensus-related functionality.
- Keeping it under the XCM folder could lead to the false assumption that the XCM team maintains or supports it.
- As for the Uniques `AssetTransactor` integration supporting XCM assets/instructions — that can stay in `xcm-builder`.

I’d suggest we start there :point_up_2: , and then revisit the other traits/stuff added. 

---

> In the future, we could add the following safeguards:
> `create_derivative` will only start the registration process, sending the ReportMetadata XCM instruction to the corresponding..

Exactly. Then we can add a `type SendXcm` to the `pallet-derivatives`, or introduce some other trait that will craft an XCM program. Again, this aligns with my earlier comments that `pallet-derivatives` doesn’t belong under the XCM modules. I still see pallet-derivatives as a kind of utility pallet to support NFT-related functionality, not as an XCM module.

Do we want to handle NFT transfers via XCM? That would be great! What are we missing to make that happen? Do we need new XCM instruction(s)? If so, why not start by defining them - perhaps through an XCM RFC? We should begin with a clear narrative and use cases, and/or propose an RFC for any new instructions that would be needed.

polkadot/xcm/xcm-builder/src/unique_instances/adapter.rs
```diff
@@ -0,0 +1,150 @@
+use core::marker::PhantomData;
```
bkontur, line 17: @mrshiposha please also check all other files, they are missing headers (licenses and stuff)

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
bkontur: If I understand correctly, nothing from the file `polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs` is used by either `polkadot/xcm/xcm-builder/src/unique_instances/adapter.rs` or `polkadot/xcm/xcm-builder/src/unique_instances/ops.rs`. So I’d say the entire `derivatives.rs` file can be cleanly moved into `pallet-derivatives`.

Once `pallet-derivatives` is moved to `substrate/frame` (as previously suggested), you can also properly take advantage of using `DispatchResult` and `DispatchError`. I also think that `polkadot/xcm/xcm-builder/src/unique_instances/ops.rs` could be moved directly into `pallet-derivatives` as well.

**I would really recommend splitting this PR into the parts:**
- XCM-related components
  - (From what I see, it's mainly XCM asset transacting — that can stay in `xcm-builder` as an implementation of `TransactAsset`.)
- Components related to **the configuration** **and logic** of `pallet-derivatives`
  - This includes most (if not all) of the code in these files (`unique_instances/ops.rs`, unique_instances/derivatives.rs`
  * For example, see this part from your mock file:
    https://github.com/UniqueNetwork/polkadot-sdk/blob/5ee9516eb4239399f1dedba4db9411fe0f17a280/polkadot/xcm/pallet-derivatives/src/mock/mod.rs#L71-L75

substrate/primitives/runtime/src/traits/mod.rs
```diff
@@ -587,6 +587,21 @@ impl<A, B: Default> Convert<A, B> for () {
 	}
 }
 
+/// Fallible conversion returning either the converted value or the
+/// [DispatchError](crate::DispatchError) on failure.
+pub trait FallibleConvert<A, B> {
```
bkontur: Usually, from what I’ve seen, a lot of conversions — especially in the XCM module - simply return the original `A` value as an error when they fail. So in such cases, you don’t really need a `FallibleConvert`.

If we wanted to go deeper, we could introduce something more generic (and arguably over-complicated), like:
```
pub trait FallibleConvert<A, B, E> {
	/// Attempt to perform the conversion.
	fn fallible_convert(a: A) -> Result<B, E>;
}
```
But then, when you have something like `SomeImplFallibleConvert<A, B, DispatchResult>` and/or `OtherImplFallibleConvert<A, B, XcmError>`, you’d need to deal with error conversions and additional boilerplate. Let’s not overthink it unless it’s truly necessary.

That said, I’d definitely start with the cleanup here:
https://github.com/paritytech/polkadot-sdk/pull/4300/files#r2160173529

And after that, maybe we will see that this `FallibleConvert` can also go in the `pallet-derivatives`.

polkadot/xcm/pallet-derivatives/src/lib.rs
mrshiposha: >I’d suggest we start there 👆 , and then revisit the other traits/stuff added.

Will do!

>Do we need new XCM instruction(s)?

The current instruction set is enough for transferring NFTs. At least via reserve-based transfers, since they just transfer the "ownership" without relocating the original object, while teleports require the metadata transfer. The data transfer is addressed by the XCM Asset Metadata RFC: https://polkadot-fellows.github.io/RFCs/approved/0125-xcm-asset-metadata.html.

The only thing we need are clear and general tools for making AssetTransactors for the existing NFT solutions.

It's just that NFT implementations are diverse because there could be a lot of aspects to the unique objects: different ID assignment approaches and metadata management, such as whether to destroy the metadata on destroying an NFT or not. The ecosystem already has different implementations: pallet-nfts, pallet-uniques, Unique Network, ORML NFTs, Aventus NFTs. Each of them follows their own path. There are similarities as well, of course, but generally they don't fall into "one big NFT interface".

>We should begin with a clear narrative and use cases, and/or propose an RFC for any new instructions that would be needed.

Hmm. Maybe I should revisit my Sub0 presentations (note that at the time this was an early concept) [Lisbon 2023](https://www.youtube.com/watch?v=4DAHuWUvbwI&list=PLOyWqupZ-WGsIJNIJ2hxTgyd4i9vui0mP), [Bangkok 2024](https://www.youtube.com/watch?v=xmLWjiujNAg) and make a corresponding text that will include the "big picture," including the future XCM Asset Metadata instruction uses.

mrshiposha:
@bkontur @franciscoaguirre please take a look at this commit: https://github.com/paritytech/polkadot-sdk/pull/4300/commits/cd4df1235649fdb70a4f7a8f05f9100d03ac1a5b.

It should address the current review comments.

NOTE: `<XCM_BUILDER>/unique_instances/ops.rs` was not moved to the `pallet-derivatives`. Only the `OwnerConvertedLocation` was moved there. Other types can be used for AssetTransactors that deal with both originals and derivatives. So, it seems they are more XCM-specific, so I left them there.

bkontur:
> @bkontur @franciscoaguirre please take a look at this commit: [cd4df12](https://github.com/paritytech/polkadot-sdk/commit/cd4df1235649fdb70a4f7a8f05f9100d03ac1a5b).
> 
> It should address the current review comments.
> 
> NOTE: `<XCM_BUILDER>/unique_instances/ops.rs` was not moved to the `pallet-derivatives`. Only the `OwnerConvertedLocation` was moved there. Other types can be used for AssetTransactors that deal with both originals and derivatives. So, it seems they are more XCM-specific, so I left them there.

@mrshiposha as we discussed on the call this week, here are my suggestions: https://github.com/UniqueNetwork/polkadot-sdk/pull/14

substrate/primitives/runtime/src/traits/mod.rs
bkontur: @mrshiposha so this `FallibleConvert` can be perfectly replaced by simple `Convert<A, Result<B, DispatchError>>`, so we don't need it anymre, please check the commit: https://github.com/UniqueNetwork/polkadot-sdk/pull/14/commits/01695ad73fa838807a8137695af43308a6088f2e as a part of https://github.com/UniqueNetwork/polkadot-sdk/pull/14

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
bkontur: FallibleConvert refactored here: https://github.com/paritytech/polkadot-sdk/pull/4300#discussion_r2245245528

polkadot/xcm/xcm-builder/src/unique_instances/derivatives.rs
bkontur: Refactor proposed here: https://github.com/UniqueNetwork/polkadot-sdk/pull/14

cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -48,18 +48,20 @@ use westend_runtime_constants::{
...
 	FrameTransactionalProcessor, FungibleAdapter, FungiblesAdapter, HashedDescription, IsConcrete,
-	LocalMint, MatchedConvertedConcreteId, MintLocation, NetworkExportTableItem, NoChecking,
-	NonFungiblesAdapter, OriginToPluralityVoice, ParentAsSuperuser, ParentIsPreset,
-	RelayChainAsNative, SendXcmFeeToAccount, SiblingParachainAsNative, SiblingParachainConvertsVia,
-	SignedAccountId32AsNative, SignedToAccountId32, SingleAssetExchangeAdapter,
-	SovereignSignedViaLocation, StartsWith, StartsWithExplicitGlobalConsensus, TakeWeightCredit,
-	TrailingSetTopicAsId, UnpaidRemoteExporter, UsingComponents, WeightInfoBounds,
-	WithComputedOrigin, WithLatestLocationConverter, WithUniqueTopic, XcmFeeManagerFromComponents,
+	LocalMint, MatchInClassInstances, MatchedConvertedConcreteId, MintLocation,
+	NetworkExportTableItem, NoChecking, NonFungiblesAdapter, OriginToPluralityVoice,
```
bkontur:
```suggestion
	NetworkExportTableItem, NoChecking, OriginToPluralityVoice,
```

substrate/frame/derivatives/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = "pallet-derivatives"
+version = "1.0.0"
+authors.workspace = true
+edition.workspace = true
+license = "Apache-2.0"
+homepage = "https://substrate.io"
```
bkontur:
```suggestion
homepage.workspace = true
```

substrate/frame/derivatives/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+readme = "README.md"
+
+[lints]
+workspace = true
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.6.12", default-features = false }
```
bkontur:
```suggestion
codec = { workspace = true }
```

substrate/frame/derivatives/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+log = { workspace = true }
+scale-info = { version = "2.11.1", default-features = false, features = ["derive"] }
```
bkontur:
```suggestion
scale-info = { features = ["derive"], workspace = true }
```

substrate/frame/derivatives/src/mock/auto_id_nfts.rs
```diff
@@ -0,0 +1,82 @@
...
+	for PredefinedIdNfts
+{
+	fn create(
+		strategy: WithConfig<ConfigValue<Owner<AccountId>>, PredefinedId<NftFullId>>,
+	) -> Result<NftFullId, DispatchError> {
+		let WithConfig { config: ConfigValue(owner), extra: id_assignment } = strategy;
+		let id = id_assignment.params;
+
+		unique_items::ItemOwner::<Test, PredefinedIdNftsInstance>::try_mutate(
+			id.clone(),
```
bkontur:
```suggestion
			id,
```

substrate/frame/derivatives/src/tests.rs
```diff
@@ -0,0 +1,375 @@
...
+			foreign_collection_id.clone(),
+		));
+
+		let derivative_collection_id =
+			AutoIdDerivativeCollections::get_derivative(&foreign_collection_id).unwrap();
+
+		// There is no derivative NFT yet
+		assert!(DerivativeNfts::get_derivative(&(
+			foreign_collection_id.clone(),
+			foreign_nft_id.clone()
```
bkontur:
```suggestion
			foreign_nft_id
```

substrate/frame/derivatives/src/tests.rs
```diff
@@ -0,0 +1,375 @@
...
+		))
+		.is_err());
+
+		let nft_beneficiary = 1;
+
+		let nft_asset: Asset = (foreign_collection_id.clone(), foreign_nft_id.clone()).into();
```
bkontur:
```suggestion
		let nft_asset: Asset = (foreign_collection_id.clone(), foreign_nft_id).into();
```

substrate/frame/derivatives/src/tests.rs
```diff
@@ -0,0 +1,375 @@
...
+			&mut hash,
+			Weight::MAX,
+			Weight::zero(),
+		);
+		assert_ok!(outcome.ensure_complete());
+
+		// The derivative NFT should exist now
+		let derivative_full_nft_id = DerivativeNfts::get_derivative(&(
+			foreign_collection_id.clone(),
+			foreign_nft_id.clone(),
```
bkontur:
```suggestion
			foreign_nft_id,
```

bkontur (changes requested):
@mrshiposha please add prdoc to pass SemVer and prdoc jobs:
https://github.com/paritytech/polkadot-sdk/actions/runs/16653697084/job/47133193884?pr=4300
https://github.com/paritytech/polkadot-sdk/actions/runs/16653697062/job/47132985589?pr=4300

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
bkontur: @franciscoaguirre @mrshiposha @jsidorenko guys, what about this comment? Any fix needed or just resolve conversation?

bkontur (approved):

polkadot/xcm/xcm-executor/src/traits/token_matching.rs
franciscoaguirre: I'm fine with anything really, but I think different `AssetId`s and `Unknown` as instance is the way to go for collection-less NFTs

franciscoaguirre: At the end of the day it's actual users who will create the standard by using it

mrshiposha: @franciscoaguirre I agree.
Should I make any changes to the PR, though? I could add a comment explaining this. The current API appears to support the collection-less NFTs the way we want them.

mrshiposha: @franciscoaguirre I extended the doc comment with this:

```rust
/// NOTE: The trait implementors should follow the convention of identifying the collection-less
/// NFTs by an XCM `Asset` of the form `{ asset_id: NFT_ID, fun: Fungibility::NonFungible(AssetInstance::Undefined) }`.
```

https://github.com/paritytech/polkadot-sdk/pull/4300/commits/c9d5cbffb003ecde0442f0dee3732b5b79c87675

cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -143,19 +144,11 @@ pub type UniquesConvertedConcreteId =
...
 /// Means for transacting unique assets.
-pub type UniquesTransactor = NonFungiblesAdapter<
-	// Use this non-fungibles implementation:
-	Uniques,
-	// This adapter will handle any non-fungible asset from the uniques pallet.
-	UniquesConvertedConcreteId,
-	// Convert an XCM Location into a local account id:
-	LocationToAccountId,
-	// Our chain's account ID type (we can't get away without mentioning it explicitly):
+pub type UniquesTransactor = UniqueInstancesAdapter<
```
franciscoaguirre, line 147: Is this a breaking change? Is the behavior different than before?

franciscoaguirre: 👍  you added tests to check the behavior is the same

substrate/frame/uniques/src/functions.rs
```diff
@@ -242,6 +242,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			|maybe_collection_details| -> Result<T::AccountId, DispatchError> {
 				let collection_details =
 					maybe_collection_details.as_mut().ok_or(Error::<T, I>::UnknownCollection)?;
+
+				// TODO should it be UnknownItem instead of UnknownCollection?
```
franciscoaguirre, line 246: Probably but changing it is a breaking change

Issue #426, opened 2023-02-22
Staking runtime apis to help with paged reward payouts


Ank4n:
Once we start supporting paged reward payouts (https://github.com/paritytech/polkadot-sdk/pull/1189/), we should expose runtime apis for staking pallet to help with the following use cases.

- [x] ~As a nominator, I want to know which page my exposure belongs to for a given era, so I could payout rewards only for that page.~ (this is not possible on on-chain)
- [x] As a validator, I want to know how many pages of exposure I have so I can payout all my nominators.
- [x] ~Api to get full exposure of a validator~ potentially unsafe as one api will call multiple storage items, also potentially bad practice to return such a huge chunk of data via runtime api, better for front ends to query storage directly.
- [ ] Api to check if there is pending pages of rewards to be claimed for an era.

PR #4301, opened 2024-04-25
[Staking] Runtime api if era rewards are pending to be claimed

Ank4n:
closes https://github.com/paritytech/polkadot-sdk/issues/426.
related to https://github.com/paritytech/polkadot-sdk/pull/1189.

Would help offchain programs to query if there are unclaimed pages of rewards for a given era.

The logic could look like below

```js
// loop as long as all era pages are claimed.
while (api.call.stakingApi.pendingRewards(era, validator_stash)) {
  api.tx.staking.payout_stakers(validator_stash, era)
}
```

### DIFF
--- a/prdoc/pr_4301.prdoc
+++ b/prdoc/pr_4301.prdoc
@@ -1,7 +1,9 @@
 title: New runtime api to check if a validator has pending pages of rewards for an era.
 
 doc:
-  - audience: Node Dev | Runtime User
+  - audience:
+    - Node Dev
+    - Runtime User
     description: |
       Creates a new runtime api to check if reward for an era is pending for a validator. Era rewards are paged and this
       api will return true as long as there is one or more pages of era reward which are not claimed.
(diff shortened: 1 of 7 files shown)


acatangiu (approved):
Nice!

kianenigma (approved):
How can we do even better, you might ask?  

ask @TarikGul and @rossbulat to tell you in which places in their codebase they have to re-implement some formula of some sort and expose those as APIs as well.

prdoc/pr_4301.prdoc
```diff
@@ -0,0 +1,11 @@
+title: New runtime api to check if a validator has pending pages of rewards for an era.
+
+doc:
+  - audience: Node Dev
```
kianenigma, line 4:
```suggestion
  - audience: Node Dev | Runtime Dev
```

I think it is worth pursuing a new audience type here, wdyt?

Ank4n: Added Runtime user which is defined in schema as `Anyone using the runtime. This can be a token holder or a dev writing a front end for a chain.`

TarikGul:
This is awesome, I'll make sure this is supported in the API when ready. Do we know what runtime this will be apart of?

PR #4302, opened 2024-04-25
migrations: `take()`should consume read and write operation weight

dastansam:
#### Problem
`take()` consumes only 1 read worth of weight in `single-block-migrations` example, while `take()` [is](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/src/storage/unhashed.rs#L63) `get() + kill()`, i.e should be 1 read + 1 write. I think this could mislead developers who follow this example to write their migrations


liamaharon (approved):

ggwpez (approved):
I think it depends on the Trie version, but okay. Many chains are still on V0.

bkchr (approved):

PR #4305, opened 2024-04-26
[pallet-migrations] fix: add missing storage version

clangenb:
This minor detail seems to have been forgotten.


liamaharon (commented):
Did you run into any issue due to the in-code storage version not being specified? This is generally considered OK, it just defaults to zero.

clangenb:
I didn't know this default to be honest, but I thought it might be an issue because I thought the log said so. However, it seems that I misremembered, now I look at the code and I see that the log does not imply that this is a bad thing. I will close that issue.

PR #4306, opened 2024-04-26
Fix assert_expected_events macro in xcm_emulator

ParthDesai:
## Description
This PR aims to fix `assert_expected_events` by only setting `meet_conditions` variable when we detect a matching event and have evaluated all conditions specified by user. This fixes an issue where `assert_expected_events` will not panic in case our event is not the last event in the `Chain::System::events()` vector. 

### Context
`assert_expected_events` first fetch events emitted by runtime lets call it system_events ; It also intializes message variable as vector of string that contains the panic message. For every event `user_event` we declared in the `vec!`  along with conditions, it does following:
1. Declare `match_conditions` variable with initial value of true.
2. Start for loop to iterate through `system_events`. For each system event, it does following in the for loop block:
     a. re-initialize `match_conditions` to true // <-- This is the bug, it simply over-writes past conditions match failures without checking current event matches `user_event` and also without evaluating any conditions.
     b. if match is found, it tries to check if conditions are satisfied, if any one of the condition is not satisfied meet_conditions variable is set to false.
3. If `meet_conditions` is `false` and we received the event, it pushes an error string in `message` vector.

### DIFF
--- a/cumulus/xcm/xcm-emulator/src/lib.rs
+++ b/cumulus/xcm/xcm-emulator/src/lib.rs
@@ -1225,8 +1225,8 @@ macro_rules! assert_expected_events {
 			let mut event_message: Vec<String> = Vec::new();
 
 			for (index, event) in events.iter().enumerate() {
-				// Have to reset the variable to override a previous partial match
-				meet_conditions = true;
+				// Variable to record current event's meet conditions
+				let mut current_event_meet_conditions = true;
 				match event {
 					$event_pat => {
 						event_received = true;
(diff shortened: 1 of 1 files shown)


acatangiu:
Can you add a unit-test showcasing the fixed scenario? A regression test that fails without your fix and passes with it. It will also help reviewers better understand the problem.

ParthDesai:
> Can you add a unit-test showcasing the fixed scenario? A regression test that fails without your fix and passes with it. It will also help reviewers better understand the problem.

Tried to write unit test, but problem is the macro expects implementation of `Chain` trait and I could not find a test/mock implementation of `Chain` trait in which I can inject some system event and then call this macro.  Can you point me to right direction here?

liamaharon (commented):
~test?~ I see above review already mentioned this, taking a look

franciscoaguirre:
@ParthDesai It's true we don't have a special emulated chain for these types of things, it should probably go in the xcm-emulator.
You could create an emulated chain as a mock to test this, like the ones we have in `cumulus/parachains/integration-tests/emulated/chains/relays/westend/src/lib.rs`

bkchr:
STale

PR #4311, opened 2024-04-26
[Staking] Not allow reap stash for virtual stakers

Ank4n:
Related to https://github.com/paritytech/polkadot-sdk/pull/3905.

Since virtual stakers does not have a real balance, they should not be allowed to be reaped.

The proper reaping for agents slashed will be done in a separate PR.


georgepisaltu (approved):

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -871,6 +871,8 @@ pub mod pallet {
 		RewardDestinationRestricted,
 		/// Not enough funds available to withdraw.
 		NotEnoughFunds,
+		/// Operation not allowed for virtual stakers.
+		VirtualStaker,
```
gpestana, line 875:
```suggestion
		VirtualStakerNotAllowed,
```

substrate/frame/staking/src/pallet/mod.rs
```diff
@@ -1997,6 +2002,9 @@ pub mod pallet {
 		) -> DispatchResult {
 			T::AdminOrigin::ensure_origin(origin)?;
 
+			// cannot restore ledger for virtual stakers.
+			ensure!(!Self::is_virtual_staker(&stash), Error::<T>::VirtualStaker);
```
gpestana, line 2006: 👍

gpestana (approved):

PR #4312, opened 2024-04-26
Assets Events for `Deposited` and `Withdrawn`

### LINKED ISSUE #4308: Assets Events for `Deposited` and `Withdrawn`
We have events for balances when they are deposited or withdrawn from an account. A lot of services like block indexers rely on these for account balance computation rather than reading on-chain state. We should provide the same events for assets so that they can use the same logic in integration.

Assigning @liamaharon but could also be a good mentor/first issue.

pandres95:
Closes #4308

Polkadot address: 12gMhxHw8QjEwLQvnqsmMVY1z5gFa54vND74aMUbhhwN6mJR


substrate/frame/assets/src/lib.rs
```diff
@@ -495,6 +495,10 @@ pub mod pallet {
 		Created { asset_id: T::AssetId, creator: T::AccountId, owner: T::AccountId },
 		/// Some assets were issued.
 		Issued { asset_id: T::AssetId, owner: T::AccountId, amount: T::Balance },
+		/// Some assets were deposited (e.g. for transaction fees).
+		Deposit { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
+		/// Some assets were withdrawn from the account (e.g. for transaction fees).
+		Withdraw { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
```
liamaharon, line 501: Just to be consistent with the other events
```suggestion
		Deposited { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
		/// Some assets were withdrawn from the account (e.g. for transaction fees).
		Withdrawn { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
```

liamaharon (approved):
Thanks! Great idea hooking into the `fungibles` methods :)

liamaharon:
bot fmt

liamaharon (approved):
PRDoc also please

prdoc/pr_4312.prdoc
```diff
@@ -0,0 +1,19 @@
...
+      New events were added to `pallet-assets`: `Deposited` and `Withdrawn`. This indicates
+      a change in the balance of an account.
+
+crates:
+  - name: pallet-assets
+    bump: patch
+  - name: pallet-asset-tx-payment
+    bump: patch
+  - name: pallet-asset-conversion-tx-payment
+    bump: patch
```
liamaharon, line 19:
```suggestion
    bump: minor
  - name: pallet-asset-tx-payment
    bump: minor
  - name: pallet-asset-conversion-tx-payment
    bump: minor
```

kianenigma (approved):

substrate/frame/assets/src/lib.rs
```diff
@@ -495,6 +495,10 @@ pub mod pallet {
 		Created { asset_id: T::AssetId, creator: T::AccountId, owner: T::AccountId },
 		/// Some assets were issued.
 		Issued { asset_id: T::AssetId, owner: T::AccountId, amount: T::Balance },
+		/// Some assets were deposited (e.g. for transaction fees).
+		Deposited { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
+		/// Some assets were withdrawn from the account (e.g. for transaction fees).
+		Withdrawn { asset_id: T::AssetId, who: T::AccountId, amount: T::Balance },
```
joepetrowski, line 501: Should these be added to the end of the enum? Technically this is a breaking change.

pandres95: I guess not, since decoding applies on each code update AFAIK. However, I just moved them to the end of the enum, it makes sense to reduce breaking as much as possible.

joepetrowski: > since decoding applies on each code update

What do you mean here? Sure in the runtime, but there are external libraries that depend on the variant encoding.

pandres95: Yes, you're right! I agree reducing breakage from end-to-end is necessary. I ended up moving those variants to the end of the enum to ensure they don't affect runtime users.

joepetrowski (approved):

prdoc/pr_4312.prdoc
```diff
@@ -0,0 +1,19 @@
+title: Add `Deposited`/`Withdrawn` events for `pallet-assets`
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      New events were added to `pallet-assets`: `Deposited` and `Withdrawn`. Make sure
+      to cover those events on tests if necessary.
+  - audience: Runtime User
+    description: |
+      New events were added to `pallet-assets`: `Deposited` and `Withdrawn`. This indicates
```
franciscoaguirre, line 10:
```suggestion
      New events were added to `pallet-assets`: `Deposited` and `Withdrawn`. These indicate
```

franciscoaguirre (approved):

PR #4318, opened 2024-04-28
Runtime Api for ideal staking rate

Ank4n:


kianenigma:
I also played with this in https://github.com/paritytech/polkadot-sdk/pull/4938/commits/8843d73b6e6d9b3224491c8b49881b7b5b782fb7

I think we should think carefully about the API and then commit to it. 

I think the struct I have in the above commit is a good start.

PR #4320, opened 2024-04-28
Whitelist for signed solution submitters in Staking

gpestana:
This PR implements a whitelisting feature for signed solution submitters in the staking pallet.

Up to `SignedWhitelistMax: Get<u32>` accounts may be whitelisted. Whenever a signed submission is accepted, the submitter is added to a storage bounded vec `SignedWhitelist`. The whitelist is updated in a FIFO manner so that:
- Up to `SignedWhitelistMax` of the last successful submitters are part of the whitelist.
- If the new account to add to the whitelist already exists, move it to the front of the queue without touching the order of the remaining accounts.
- If a non-whitelisted account is added to a saturated list, the last (oldest) submitter will be removed.

The base signed deposit fee for whitelisted accounts is `SignedDepositWhitelist`. Deposits for non-whitelisted accounts are calculated using the progressive geometric formula based on the number of queued submissions.


substrate/frame/election-provider-multi-phase/src/signed.rs
```diff
@@ -509,6 +509,11 @@ impl<T: Config> Pallet<T> {
 		let positive_imbalance =
 			T::Currency::deposit_creating(who, reward.saturating_add(call_fee));
 		T::RewardHandler::on_unbalanced(positive_imbalance);
+
+		// register submitter as whitelisted for the next round(s).
+		SignedWhitelist::<T>::mutate(|w| {
+			w.force_push(who.clone());
```
kianenigma, line 515: hmm, why? I think whitelisted accounts should be determined by the governance, no?

gpestana: The idea of this PR is to whitelist the accounts that have recently (i.e. submitted the last accepted signed solutions). So this works as a sort of temporary reputation system, where recent submitters have the part of the deposit waived.

The main goal is to increase considerably the geometric progression deposit without affecting the set of submitters that usually cooperate with good solutions.

We could also have this whitelist set by governance, but I think that's an overkill for this use case and very much static.

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -873,6 +909,9 @@ pub mod pallet {
 			// `SignedMaxSubmissions` is a red flag that the developer does not understand how to
 			// configure this pallet.
 			assert!(T::SignedMaxSubmissions::get() >= T::SignedMaxRefunds::get());
+
+			// idem.
+			assert!(T::SignedMaxSubmissions::get() >= T::SignedWhitelistMax::get());
```
Ank4n, line 914: Wondering if we require `SignedWhitelistMax` as a perbill of `SignedMaxSubmissions`, would it be better? One less thing a new dev could do wrong with the configuration?

substrate/frame/election-provider-multi-phase/src/signed.rs
Ank4n, line 515: Could you add a FIFO test? I tried and does not work as expected.
This is what I got:

```
whitelist(1); // BoundedVec([1], 2)
whitelist(2); // BoundedVec([1, 2], 2) 
whitelist(3); // BoundedVec([1, 3], 2)
```

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -616,6 +642,12 @@ pub mod pallet {
 			MaxWinners = Self::MaxWinners,
 		>;
 
+		/// Maximum number of whitelisted accounts.
```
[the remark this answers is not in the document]
gpestana: Added directly to the PR. thanks for the suggestion.

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -93,6 +93,32 @@
 //! Note that both of the bottom solutions end up being discarded and get their deposit back,
 //! despite one of them being *invalid*.
 //!
+//! ### Signed submission deposits
+//!
+//! Each signed submission must hold a deposit that will be returned to the submitter if their
+//! solution was accepted or not processed. The deposit will be slashed if the submission is not
+//! valid. The amount to deposit per submission *may* depend on multiple factors, namely the size of
+//! the submission (in bytes and weight), the deposit base, the inclusion or not of the submitter in
+//! a whitelist and, potentially, on the size of the submission queue.
```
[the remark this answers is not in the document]
gpestana: Thanks for the suggestion, added it manually.

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -642,6 +674,10 @@ pub mod pallet {
 		#[pallet::constant]
 		type SignedRewardBase: Get<BalanceOf<Self>>;
 
+		/// Fixed deposit for a signed solution.
```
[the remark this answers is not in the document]
gpestana: The non-whitelisted submissions need to deposit the fix amount + the geometric increment parcel. The whitelisted submissions *only* deposit the fixed deposit.

substrate/frame/election-provider-multi-phase/src/signed.rs
gpestana: I have fixed the code and added a test to ensure we have a FIFO mechanism in place. It should be good now!

prdoc/pr_4320.prdoc
```diff
@@ -6,7 +6,8 @@ doc:
       Implementation of a whitelisting feature for signed solution submitters for staking.
       When a signed solution is validated and accepted, its account is whitelisted.
       A whitelisted submitter's based deposit is fixed and it does not depend on the size of the
-      submission queue.
+      submission queue. In summary, this allows a dynamic set of previouslt successful signed
```
Ank4n, line 9:
```suggestion
      submission queue. In summary, this allows a dynamic set of previously successful signed
```

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -98,8 +98,9 @@
 //! Each signed submission must hold a deposit that will be returned to the submitter if their
 //! solution was accepted or not processed. The deposit will be slashed if the submission is not
 //! valid. The amount to deposit per submission *may* depend on multiple factors, namely the size of
-//! the submission (in bytes and weight), the deposit base, the inclusion or not of the submitter in
-//! a whitelist and, potentially, on the size of the submission queue.
+//! the submission (in bytes and weight), the deposit base, if the submitters have submitted a
+//! successult solution (and therefore are whitelisted) and, potentially, on the size of the
```
Ank4n, line 102:
```suggestion
//! successful solution (and therefore are whitelisted) and, potentially, on the size of the
```

Ank4n (approved):

substrate/frame/election-provider-multi-phase/src/lib.rs
```diff
@@ -93,6 +93,32 @@
...
+//! solution was accepted or not processed. The deposit will be slashed if the submission is not
+//! valid. The amount to deposit per submission *may* depend on multiple factors, namely the size of
+//! the submission (in bytes and weight), the deposit base, the inclusion or not of the submitter in
+//! a whitelist and, potentially, on the size of the submission queue.
+//!
+//! The base deposit of a submission is calculated based on whether a submitter is whitelisted. Up
+//! to [`pallet::Config::SignedWhitelistMax`] can be whitelisted at a given time. Everytime a
+//! submitter's solution is verified, valid and thus accepted, the submitter is added to the
+//! whitelist. The whitelist is a bounded FIFO with the list of the last submitters (note: there may
+//! be duplicate submitters in the whitelist, in case their submissions were accepted in the last
```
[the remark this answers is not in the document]
gpestana: @Ank4n after some discussions, we decided to change the FIFO logic so that there are no duplicate entries and the submitters can now remain for a long time in the whitelist -- similar to what you suggestions. So now the whitelist is updated in a FIFO manner so that:

- Up to `SignedWhitelistMax` of the last successful submitters are part of the whitelist.
- If the new account to add to the whitelist already exists, move it to the front of the queue without touching the order of the remaining accounts.
- If a non-whitelisted account is added to a saturated list, the last (oldest) submitter will be removed.

So no more duplicates.

substrate/frame/election-provider-multi-phase/src/signed.rs
```diff
@@ -510,6 +510,17 @@ impl<T: Config> Pallet<T> {
 		let positive_imbalance =
 			T::Currency::deposit_creating(who, reward.saturating_add(call_fee));
 		T::RewardHandler::on_unbalanced(positive_imbalance);
+
+		// `who` is force inserted at the top of the whitelist if it is not whitelisted yet,
+		// potentially pushing out older whitelisted accounts. If `who` is already part of the
+		// whitelist, move it to the head of the whitelist, while keeping the order of the remaining
+		// elements.
+		SignedWhitelist::<T>::mutate(|w| {
```
kianenigma, line 518: alternative: 
- use `BoundedVecDeq` as it has `push_front`. 
- always start with removing the existing one. 
- then pushing to front

kianenigma (approved):

PR #4323, opened 2024-04-29
Subsystem benchmark template to estimate CPU hardware requirements

alexggh:


polkadot/node/subsystem-bench/examples/estimate_cpu_requirements.yaml
```diff
@@ -0,0 +1,47 @@
+## Estimate CPU requirements on a network with 500 validators 100 cores
```
sandreim, line 1: I think we should merge this file to not have to recreate it every time we want to do such tests.

alexggh: Yeah, that was my plan, I just wanted to see the results first.

polkadot/node/subsystem-bench/examples/estimate_cpu_requirements.yaml
```diff
@@ -0,0 +1,47 @@
...
+  peer_bandwidth: 524288000000
+  bandwidth: 524288000000
+  num_blocks: 3
+  connectivity: 100
+  latency: null
+
+- objective: DataAvailabilityWrite
+  n_validators: 500
+  n_cores: 100
+  max_validators_per_core: 5
```
burdges, line 27: Is this backing?

You won't have much of a CPU requirement estimate without approvals, but maybe they're set elsewhere?

alexggh: This is the one measuring the approval voting CPU: https://github.com/paritytech/polkadot-sdk/pull/4323/files#diff-3173d0f01ac97698eec2cfc0da96349100b8cdfefa58f42369e142b39f0997b7R5.

burdges: I meant the `needed_approvals` setting.

alexggh: Yes, that a configuration parameters as well, when it is not present the default version is used, which is 30 as all of our production chains.

https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/node/subsystem-bench/src/lib/configuration.rs#L75

AndreiEres (approved):

.gitlab/pipeline/test.yml
```diff
@@ -619,3 +619,10 @@ subsystem-benchmark-approval-voting:
   script:
     - cargo bench -p polkadot-node-core-approval-voting --bench approval-voting-regression-bench --features subsystem-benchmarks
   allow_failure: true
+
+subsystem-benchmark-estimate-cpu-requirements:
+  extends:
+    - .subsystem-benchmark-template
+  script:
+    - cargo run -p polkadot-subsystem-bench  --release    -- polkadot/node/subsystem-bench/examples/estimate_cpu_requirements.yaml
```
sandreim, line 627: I'd remove these so we can merge.

AndreiEres: We don't need it in the CI, right? Is it here just to run on a reference hw?

bkchr:
@alexggh do you still want to merge this?

alexggh:
No, this is stale, I'm closing it.

Issue #1621, opened 2023-09-19
restore disable-runtime-api


xlc:
Removed at https://github.com/paritytech/polkadot-sdk/pull/1328 but we still need it

              A proper solution will be changes `disable-runtime-api` to `enable-runtime-api` and make it enabled by default so that new feature means add functionality, not the other way around.

_Originally posted by @xlc in https://github.com/paritytech/polkadot-sdk/issues/1328#issuecomment-1724808904_

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/wasm-view-functions/1045/7

Ank4n:
cc: @NachoPal @bkchr

bolajahmad:
There's a [commit here](https://github.com/paritytech/polkadot-sdk/commit/730f955b2c443d331065f88d67e8963efc9af9b4#diff-e31fc2788dafe7c3c260538ddc29864bdeb56f8b8955ee67fb4b21ce1f83f1df) that removes the `disable-runtime-api`. 
The solution should involve sort of a reversal of the changes here, am I right @xlc ?

xlc:
yes

xlc:
not needed anymore

PR #4325, opened 2024-04-29
Restoring disable-runtime-api for test chains

bolajahmad:
Adds disable-runtime-api back to rococo and westend. Previously removed.

Partly closes, #1621 . Kusams and Polkadot will be updated on the [runtimes codebase](https://github.com/bolajahmad/runtimes) in a seperate PR.


polkadot/runtime/westend/src/lib.rs
```diff
@@ -191,15 +192,30 @@ parameter_types! {
...
+	type RuntimeTask = RuntimeTask;
+	type PalletInfo = PalletInfo;
+	type OnNewAccount = ();
+	type OnKilledAccount = ();
+	type OnSetCode = ();
+	type SingleBlockMigrations = ();
+	type MultiBlockMigrator = ();
+	type PreInherents = ();
+	type PostInherents = ();
+	type PostTransactions = ();
```
bkchr, line 209:
```suggestion
```

polkadot/runtime/rococo/src/lib.rs
```diff
@@ -199,6 +201,21 @@ parameter_types! {
...
+	type RuntimeTask = RuntimeTask;
+	type PalletInfo = PalletInfo;
+	type OnNewAccount = ();
+	type OnKilledAccount = ();
+	type OnSetCode = ();
+	type SingleBlockMigrations = ();
+	type MultiBlockMigrator = ();
+	type PreInherents = ();
+	type PostInherents = ();
+	type PostTransactions = ();
```
bkchr, line 218:
```suggestion
```

polkadot/runtime/rococo/Cargo.toml
```diff
@@ -323,6 +330,8 @@ try-runtime = [
 	"sp-runtime/try-runtime",
 ]
 
+disable-runtime-api = []
```
bkchr, line 333: Some comment would be nice.

bolajahmad: That makes sense and I will add descriptive comments.
One thing I'd like to confirm though, is this still required by Cumulus to access certain runtime(s)?

polkadot/runtime/westend/src/lib.rs
bolajahmad: Screenshot 2024-05-03 at 19 20 49

The issue with removing these changes is that the runtime demands that they are implemented. These types should not be needed but I am not sure why the compiler shows this error. Are there any recommendations from your end @bkchr

polkadot/runtime/rococo/src/lib.rs
bolajahmad: Screenshot 2024-05-03 at 19 24 43

Just like the previous comment, the types should not be needed but I am not sure why the compiler shows this error. Are there any recommendations from your end @bkchr

bkchr: Did you try `cargo build`? `rust-analyzer` is may just confused.

bolajahmad: It does work, rust-analyzer was truly the problem here. I got it to build without the additional types. Thanks @bkchr

polkadot/runtime/rococo/Cargo.toml
bolajahmad: Added some comments, let me know if they're descriptive enough.

polkadot/runtime/rococo/Cargo.toml
```diff
@@ -324,6 +331,9 @@ try-runtime = [
 	"sp-runtime/try-runtime",
 ]
 
+# When enabled, the runtime API will not be built.
```
bkchr, line 334:
```suggestion
# If enabled, the runtime apis will not be exposed by the runtime. 
```

polkadot/runtime/westend/Cargo.toml
```diff
@@ -338,6 +346,9 @@ try-runtime = [
 	"sp-runtime/try-runtime",
 ]
 
+# When enabled, the runtime API will not be built.
```
bkchr, line 349:
```suggestion
# If enabled, the runtime apis will not be exposed by the runtime.
```

bkchr (commented):
Can you please revert the unrelated changes on your pr? Otherwise it looks good.

polkadot/runtime/westend/src/lib.rs
bolajahmad: These have been removed, the errors were from rust-analyzer and project still builds

bkchr:
@bolajahmad could you please revert the changes that are not related to your pr?

bolajahmad:
> @bolajahmad could you please revert the changes that are not related to your pr?

Hi @bkchr, i tried to cleanup the branch and revert unrelated changes to my PR. I've removed the changes that seemed weird and it should be good. I'm sorry to bother you when you're reviewing my changes but I can't find anymore unrelated changes in this PR. I'll appreciate if you mention something specific you think I need to remove/revert too🙏🏼

bkchr:
https://github.com/paritytech/polkadot-sdk/pull/4325/files#diff-5264d812f7aadfc486efffd3322b11c8dee5e8ddf2e77e0a9e70b573f3d2afddR91 here for example where you import `AccountIdLookup` which was not there before.

Or the formatting changes [here](https://github.com/paritytech/polkadot-sdk/pull/4325/files#diff-da220c6c064cc047fef5d3b4a594ae30b8bbf48a29b740742bc9858285c7ba57R902)

bkchr:
The issue was closed as not needed anymore, so we can do the same with this pr.

PR #4326, opened 2024-04-29
`CheckWeight` SE: Check for extrinsic length + proof size combined

skunert:
Currently the `CheckWeight` `SignedExtension` was tracking the size of the proof and the extrinsic length separately. But in reality we need one more check that ensures we don't hit the PoV limit with both combined.

The rest of the logic remains unchanged. One  scenario where the changes make a difference is when we enter this branch:
https://github.com/paritytech/polkadot-sdk/blob/f34d8e3cf033e2a22a41b505c437972a5dc83d78/substrate/frame/system/src/extensions/check_weight.rs#L185-L198

This was previously allowing to some extrinsics that is exceeding the block limit but are withing the reserved area of `BlockWeights`. This will now  be caught by the later check I introduced. I think the new behaviour makes sense, since the proof size dimension is designed for parachains and they don't want to go over the limit and get rejected.

In the long run we should maybe get rid of `RuntimeBlockLength` alltogether, however that would require a deprecation process and can come at a later point.


bkchr (approved):
Thank you!

substrate/frame/system/src/extensions/check_weight.rs
```diff
@@ -136,8 +130,27 @@ where
...
+pub fn check_combined_proof_size(
+	maximum_weight: &BlockWeights,
+	next_len: u32,
+	next_weight: &crate::ConsumedWeight,
+) -> Result<(), TransactionValidityError> {
+	// This extra check ensures that the extrinsic length does not push the
+	// PoV over the limit.
+	let total_pov_size = next_weight.total().proof_size().saturating_add(next_len as u64);
+	if total_pov_size > maximum_weight.max_block.proof_size() {
+		log::debug!(target: LOG_TARGET, "Extrinsic exceeds total pov size: {}kb, limit: {}kb", total_pov_size as f64/1024.0, maximum_weight.max_block.proof_size() as f64/1024.0);
```
acatangiu, line 143: nit:
```suggestion
		log::debug!(
			target: LOG_TARGET,
			"Extrinsic exceeds total pov size: {}kb, limit: {}kb",
			total_pov_size as f64/1024.0,
			maximum_weight.max_block.proof_size() as f64/1024.0
		);
```

acatangiu (approved):

substrate/frame/system/src/extensions/check_weight.rs
```diff
@@ -136,8 +130,27 @@ where
...
+/// Check that the combined extrinsic length and proof size together do not exceed the PoV limit.
+pub fn check_combined_proof_size(
+	maximum_weight: &BlockWeights,
+	next_len: u32,
+	next_weight: &crate::ConsumedWeight,
+) -> Result<(), TransactionValidityError> {
+	// This extra check ensures that the extrinsic length does not push the
+	// PoV over the limit.
+	let total_pov_size = next_weight.total().proof_size().saturating_add(next_len as u64);
+	if total_pov_size > maximum_weight.max_block.proof_size() {
```
ggwpez, line 142: It is just checking against `max_block` here instead of `per_class` for the class of the extrinsic?

skunert: Yes, that is intentional. We are still checking the proof and ref_time weight against the extrinsic class limits. In addition the block_length total, which has no classes. This additional check should just make sure that the total limit is respected.

ggwpez (approved):

PR #4328, opened 2024-04-29
BEEFY: Define basic fisherman

serban300:
Related to https://github.com/paritytech/polkadot-sdk/pull/1903

For #1903 we will need to add a Fisherman struct. This PR:
- defines a basic version of `Fisherman` and moves into it the logic that we have now for reporting double voting equivocations
- splits the logic for generating the key ownership proofs into a more generic separate method
- renames `EquivocationProof` to `DoubleVotingProof` since later we will introduce a new type of equivocation

The PR doesn't contain any functional changes

### DIFF
--- a/polkadot/node/service/src/fake_runtime_api.rs
+++ b/polkadot/node/service/src/fake_runtime_api.rs
@@ -242,7 +242,7 @@ sp_api::impl_runtime_apis! {
 		}
 
 		fn submit_report_equivocation_unsigned_extrinsic(
-			_: beefy_primitives::EquivocationProof<
+			_: beefy_primitives::DoubleVotingProof<
 				BlockNumber,
 				BeefyId,
 				BeefySignature,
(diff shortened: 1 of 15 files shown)


Lederstrumpf (approved):
lgtm, fwiw

acatangiu (approved):

ggwpez:
This is not `R0-silent`. It breaks compilation by renaming `EquivocationProof` and must appear in the CHANGELOG for 1.12 by declaring a `major` version bump of `beefy-primitives`.

serban300:
> This is not `R0-silent`. It breaks compilation by renaming `EquivocationProof` and must appear in the CHANGELOG for 1.12 by declaring a `major` version bump of `beefy-primitives`.

Solrry, I didn't know that renamings need to be mentioned in the CHANGELOR/PR doc. I see that 1.12 is already released. So is it worth doing the version bump no, and adding a PR, or is it just something to know in the future ?

PR #3721, opened 2024-03-17
establish_channel_with_system

xlc:
Implements https://github.com/polkadot-fellows/RFCs/issues/82
Allow any parachain to have bidirectional channel with any system parachains

Looking for initial feedbacks before continue finish it

TODOs:
- [x] docs
- [x] benchmarks
- [x] tests

PR #4332, opened 2024-04-30
HRMP - set `DefaultChannelSizeAndCapacityWithSystem` with dynamic values according to the `ActiveConfig`

bkontur:
## Summary
This PR enhances the capability to set `DefaultChannelSizeAndCapacityWithSystem` for HRMP. Currently, all testnets (Rococo, Westend) have a hard-coded value set as 'half of the maximum' determined by the live `ActiveConfig`. While this approach appears satisfactory, potential issues could arise if the live `ActiveConfig` are adjusted below these hard-coded values, necessitating a new runtime release with updated values. Additionally, hard-coded values have consequences, such as Rococo's benchmarks not functioning: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6082656.

## Solution
The proposed solution here is to utilize `ActiveConfigHrmpChannelSizeAndCapacityRatio`, which reads the current `ActiveConfig` and calculates `DefaultChannelSizeAndCapacityWithSystem`, for example, "half of the maximum" based on live data. This way, whenever `ActiveConfig` is modified, `ActiveConfigHrmpChannelSizeAndCapacityRatio` automatically returns adjusted values with the appropriate ratio. Thus, manual adjustments and new runtime releases become unnecessary.

Relates to a comment/discussion: https://github.com/paritytech/polkadot-sdk/pull/3721/files#r1541001420
Relates to a comment/discussion: https://github.com/paritytech/polkadot-sdk/pull/3721/files#r1549291588


bkontur:
bot bench polkadot-pallet --runtime=rococo --pallet=runtime_parachains::hrmp
bot bench polkadot-pallet --runtime=westend --pallet=runtime_parachains::hrmp

xlc:
lgtm

polkadot/runtime/westend/src/lib.rs
```diff
@@ -1163,15 +1164,18 @@ impl pallet_message_queue::Config for Runtime {
 impl parachains_dmp::Config for Runtime {}
 
 parameter_types! {
-	pub const DefaultChannelSizeAndCapacityWithSystem: (u32, u32) = (4096, 4);
+	pub const HrmpChannelSizeAndCapacityWithSystemRatio: Percent = Percent::from_percent(100);
```
franciscoaguirre, line 1167: I thought we wanted half of the max? Or here it's 100 because it's a testnet?

franciscoaguirre (approved):

polkadot/runtime/westend/src/lib.rs
bkontur: Well, judging based on the linked [comment from the description](https://github.com/paritytech/polkadot-sdk/pull/3721/files#r1541001420),  "half of the max value" it was somewhat of a blind guess (no offence).

If you check this (not sure if still actual): https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/roadmap/implementers-guide/src/runtime/hrmp.md
```
 ii. Check that proposed_max_capacity is less or equal to config.hrmp_channel_max_capacity and greater than zero.
iii. Check that proposed_max_message_size is less or equal to config.hrmp_channel_max_message_size and greater than zero.
```
and also the implementation [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L1448-L1456), I think, there shouldn't be a problem with setting those values to 100% (e.g. `config.hrmp_channel_max_capacity`).

Also, maybe we don't even need [`type DefaultChannelSizeAndCapacityWithSystem`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/hrmp.rs#L279) and we could just use a max  from `ActiveConfig` e.g. `config.hrmp_channel_max_capacity`.

xlc: that was just an arbitrary number I made and it is for testnet so shouldn't really matter much

acatangiu (approved):

PR #4337, opened 2024-04-30
Setup `storage_version` for bridging pallets

bkontur:
Closes: https://github.com/paritytech/parity-bridges-common/issues/2947

### DIFF
--- a/bridges/modules/beefy/src/lib.rs
+++ b/bridges/modules/beefy/src/lib.rs
@@ -40,6 +40,8 @@ pub use pallet::*;
 
 mod utils;
 
+pub mod migration;
+
 #[cfg(test)]
 mod mock;
 #[cfg(test)]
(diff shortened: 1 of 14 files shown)


franciscoaguirre (approved):

serban300 (approved):

ggwpez (changes requested):
I dont see how this is needed. cc @liamaharon  

If there is an issue with pallets that do not have storage versions, then we should fix that and not work around it by setting versions. Not having a version for a pallet should be perfectly fine.

bkchr:
Yes, this should not be needed.

liamaharon (commented):
Not needed, storage version is zero if omitted. 

That said, there's no harm in specifying it explicitly if people prefer.

bkontur:
> I dont see how this is needed. cc @liamaharon
> 
> If there is an issue with pallets that do not have storage versions, then we should fix that and not work around it by setting versions. Not having a version for a pallet should be perfectly fine.

No, no, there is no problem :)

My intentions with this PR are just to prepare the bridging pallets for future migrations, which will be introduced with BridgesV2. Additionally, we have other issues that will require migrations. I just thought it would be a good idea to add `#[pallet::storage_version]` and prepare pallets for migrations.

But let's not complicate, we will add this when needed.

PR #4339, opened 2024-04-30
Improve On_demand_assigner events

### LINKED ISSUE #4254: on-demand: Improve/fix events
I just checked on events in on-demand and noticed two things:

1.  [OnDemandOrderPlaced](https://github.com/paritytech/polkadot-sdk/blob/54e0326967fb00b64a4e33694e38ca693ae2ede3/polkadot/runtime/parachains/src/assigner_on_demand/mod.rs#L409) seems to be never triggered.
2. [SpotTrafficSet](https://github.com/paritytech/polkadot-sdk/blob/54e0326967fb00b64a4e33694e38ca693ae2ede3/polkadot/runtime/parachains/src/assigner_on_demand/mod.rs#L411). This value will be largely useless for people, maybe we should have a `SpotPriceSet`?

bolajahmad:
Closes #4254 

OnDemandOrderPlaced event that is useful for indexers to save data related to on demand orders. Check [discussion here](https://substrate.stackexchange.com/questions/11366/ondemandassignmentprovider-ondemandorderplaced-event-was-removed/11389#11389).


eskimor (commented):
Thanks! Left a few nits and triggered CI for you. Please remove unrelated formatting changes as well.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -599,6 +605,8 @@ where
 				// Only update storage on change
 				if new_traffic != old_traffic {
 					queue_status.traffic = new_traffic;
+
+					// emit the event for updating new price
```
Overkillus, line 609: See main body of the review for pointers but this event needs changes

Overkillus (commented):
Hey @bolajahmad o/ 

Always happy to have some external contributors so if you need any help please feel free to ping me.

- [OnDemandOrderPlaced](https://github.com/paritytech/polkadot-sdk/blob/54e0326967fb00b64a4e33694e38ca693ae2ede3/polkadot/runtime/parachains/src/assigner_on_demand/mod.rs#L409) looking good
- [SpotTrafficSet](https://github.com/paritytech/polkadot-sdk/blob/54e0326967fb00b64a4e33694e38ca693ae2ede3/polkadot/runtime/parachains/src/assigner_on_demand/mod.rs#L411) needs changes

I think there is a small misunderstanding so let me clear it up:

Eskimor:

> [SpotTrafficSet](https://github.com/paritytech/polkadot-sdk/blob/54e0326967fb00b64a4e33694e38ca693ae2ede3/polkadot/runtime/parachains/src/assigner_on_demand/mod.rs#L411) This value will be largely useless for people, maybe we should have a SpotPriceSet?

@bolajahmad :

> SpotTrafficSet is used in the update_spot_traffic call. What is the intended effect of changing the EventName to SpotPriceSet, especilly if the resulting event has the same arguments?

I assume that Eskimor meant **not only renaming** that event but also changing what it actually represents (to be in-line with the new name). The new event should represent the actual SpotPrice instead of the Traffic component. It seems that this might be much more useful for users. 

Notice that if both `SpotPriceSet` as proposed by Eskimor and `OnDemandOrderPlaced` have the spot price then in case of an order being made we would get a double event with the same thing. This might not be undesirable IMO. 

TODO:

- [x] rename `SpotTrafficSet` to `SpotPriceSet`
- [x] After updating traffic calculate spot price and deposit it in that new `SpotPriceSet` event

---

@eskimor I'm not sure if I agree with:
> Issue SpotTrafficSet ... only if price changed without an order getting placed.

First of all I assume you meant `SpotPriceSet` otherwise it seems inconsistent with what you said earlier.

Issue here is that SpotPrice can also change when orders are placed so someone could easily follow this event thinking he has the most up to date spot price which would be false if we emit it only when price changed without an order getting placed.

I suggest always emitting this spot price update (traffic goes down due to no orders or orders boost the price up). Then we can either drop the spot price from the `OnDemandOrderPlaced` event or simply have it in both places for absolute clarity.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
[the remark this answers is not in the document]
bolajahmad: Running the `cargo +nightly fmt --all` worked but some of the formatting remained inconsistent with the styleguide example. 

I'll try to run it while paying more attention if there's an issue

Cargo.lock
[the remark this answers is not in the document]
bolajahmad: Thank you a lot for reviewing my PR🫡. Got some conflicts while trying to merge then but I'll remove the changes in here.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
bolajahmad: Thank you very much for your willingness to help and make me understand the requirement of the issue. 

I have implemented the changes as you explained and pushed another commit to cater for them.

Overkillus:
Hey @bolajahmad 

Thanks for fixing it! Seems we are nearly good to go and the only item missing is the PRdoc. I see that you've added it into the description of that PR but what you need to do is put it in a file in this folder: https://github.com/paritytech/polkadot-sdk/tree/master/prdoc

Name it `pr_4339.prdoc` and use the same description items as the ones you have above. I added an extra `SpotPriceSet` clarification and removed the link:

```
# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json

title: Improving on_demand_assigner emitted events

doc:
  - audience: Runtime User
    description: |
      Registering OnDemandOrderPlaced event that is useful for indexers to save data related to on demand orders. Adds SpotPriceSet as a new event to monitor on-demand spot prices. It updates whenever the price changes due to traffic.

crates:
  - name: polkadot-runtime-parachains
    bump: minor

```

Once you get that done we can do a final review and merge :rocket:

eskimor:
bot fmt

Overkillus (approved):

eskimor:
bot fmt

eskimor (approved):
Would be nice to get the remaining formatting inconsistencies fixed. Other than that looks good.

polkadot/runtime/parachains/src/assigner_on_demand/mod.rs
```diff
@@ -368,10 +368,10 @@ pub mod pallet {
...
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
-		/// An order was placed at some spot price amount.
-		OnDemandOrderPlaced { para_id: ParaId, spot_price: BalanceOf<T> },
-		/// The value of the spot traffic multiplier changed.
-		SpotTrafficSet { traffic: FixedU128 },
+		/// An order was placed at some spot price amount by orderer ordered_by
+		OnDemandOrderPlaced { para_id: ParaId, spot_price: BalanceOf<T>, ordered_by: T::AccountId },
+		/// The value of the spot price has likely changed
+		SpotPriceSet { spot_price: BalanceOf<T> },
```
s0me0ne-unkn0wn, line 374: This *may* be problematic as you're changing the formats of the existing events here. As events from the previous block are reset after the runtime upgrade but before running any hooks, [event storage must always be compatible](https://github.com/paritytech/polkadot-sdk/blob/2b1c606a338c80c5220c502c56a4b489f6d51488/substrate/frame/executive/src/lib.rs#L577-L579). So, technically, this part needs more care.

Buuuut, given that, IIUC,
1) `OnDemandOrderPlaced` was never deposited by the previous version of the code, and
2) `SpotPriceSet` internal representation is the same, as `BalanceOf<T>` coincides with `FixedU128` in all the current use cases,

it may be safe to leave it as is.

That said, some more research is needed on how safe it is.

CC @liamaharon @bkchr

PR #4342, opened 2024-05-01
Add Async Backing Parachain Guide to Polkadot SDK docs

DrW3RK:
Reference Issue: https://github.com/w3f/polkadot-wiki/issues/5835

### DIFF
--- a/docs/sdk/src/guides/parachains/async_backing_guide.rs
+++ b/docs/sdk/src/guides/parachains/async_backing_guide.rs
@@ -1,0 +1,2 @@
+//! # Upgrade Parachain for Asynchronous Backing Compatibility
+//!
--- a/docs/sdk/src/guides/parachains/async_backing_guide.rs
+++ b/docs/sdk/src/guides/parachains/async_backing_guide.rs
@@ -3,7 +3,7 @@
 //! This guide is relevant for cumulus based parachain projects started in 2023 or before. Later
 //! projects should already be async backing compatible. If starting a new parachain project, please use
 //! an async backing compatible template such as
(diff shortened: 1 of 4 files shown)


DrW3RK:
The Wiki doc https://wiki.polkadot.network/docs/maintain-guides-async-backing uses async backing compatible parachain template as a reference which has been moved to https://github.com/paritytech/polkadot-sdk/tree/master/templates/parachain 

Currently, this PR https://github.com/paritytech/polkadot-sdk/pull/4295 is working on making it Async backing compatible.

The content on this Async Backing PR needs to be revisited after the above PR is mergedl

kianenigma (commented):
But this is still using harcoded code snippets right? the goal of my suggestion of moving it here was to make the code snippets be backed by `docify`.

docs/sdk/src/guides/parachains/async_backing_guide.rs
```diff
@@ -0,0 +1,452 @@
+//! # Upgrade Parachain for Asynchronous Backing Compatibility
```
alexggh, line 1: What is the plan here will we make https://wiki.polkadot.network/docs/maintain-guides-async-backing point to here ? Because otherwise we will have to maintain this information in two places.

I don't think at this point we can remove https://wiki.polkadot.network/docs/maintain-guides-async-backing, because it has references in other articles and forum posts and it would be a bad experience if it becomes a broken link.

DrW3RK: Hey @alexggh That is the plan. We will retain preliminary content on that page and add a prominent link to the Polkadot SDK doc. The Wiki doc link will not be broken.

DrW3RK:
https://github.com/paritytech/polkadot-sdk/pull/4363/

PR #4343, opened 2024-05-01
Fix: change litep2p dependency from main to working tag

metricaez:
## Description:

`litep2p` dependency on `main` repository branch led to dependency breaking build because of type change.

## Error log:
```shell
error[E0308]: mismatched types
   --> /Users/metricaez/.cargo/git/checkouts/polkadot-sdk-cff69157b985ed76/0bb6249/substrate/client/network/src/litep2p/discovery.rs:465:74
    |
465 |                 return Poll::Ready(Some(DiscoveryEvent::GetRecordSuccess { query_id, record }));
    |                                                                                      ^^^^^^ expected `Record`, found `PeerRecord`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `sc-network` (lib) due to 1 previous error
```

## Fix:

Move dependency to the stable tag `v0.3.0` 

cc @Moliholy

### DIFF
--- a/substrate/client/network/Cargo.toml
+++ b/substrate/client/network/Cargo.toml
@@ -59,7 +59,7 @@ sp-blockchain = { path = "../../primitives/blockchain" }
 sp-core = { path = "../../primitives/core" }
 sp-runtime = { path = "../../primitives/runtime" }
 wasm-timer = "0.2"
-litep2p = { git = "https://github.com/paritytech/litep2p", branch = "master" }
+litep2p = { git = "https://github.com/paritytech/litep2p", tag = "v0.3.0" }
 once_cell = "1.18.0"
 void = "1.0.2"
 schnellru = "0.2.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,7 +27,7 @@ version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
- "gimli 0.28.0",
+ "gimli 0.28.1",
 ]
 
 [[package]]
@@ -44,15 +44,6 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "aead"
-version = "0.4.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
-dependencies = [
- "generic-array 0.14.7",
-]
-
-[[package]]
-name = "aead"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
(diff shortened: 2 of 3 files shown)


metricaez:
This would only fix `polkadot-sdk` release branch, fix should be further propagated.

substrate/client/network/Cargo.toml
```diff
@@ -59,7 +59,7 @@ sp-blockchain = { path = "../../primitives/blockchain" }
 sp-core = { path = "../../primitives/core" }
 sp-runtime = { path = "../../primitives/runtime" }
 wasm-timer = "0.2"
-litep2p = { git = "https://github.com/paritytech/litep2p", branch = "master" }
+litep2p = { git = "https://github.com/paritytech/litep2p", tag = "v0.3.0" }
```
alexggh, line 62: I'm not sure I understand how this break the release branch, we do have a Cargo.lock, so the release branch should already build against a compatible `litep2p`, what am I missing ?

metricaez: Hi!

I encountered this issue while updating a parachain to `v1.11`.

`source = "git+https://github.com/paritytech/litep2p?branch=master#e03a6023882db111beeb24d8c0ceaac0721d3f0f"
`
This was the commit hash our `Cargo.lock` pointed after the upgrade and it was breaking with the build with the shared log, to fix it we had to manually modify to:
`source = "git+https://github.com/paritytech/litep2p?branch=master#b142c9eb611fb2fe78d2830266a3675b37299ceb"
`
Which is the release commit of the proposed tag. 

Could you please verify?

Moliholy: Furthermore, pointing dependencies to non-stable branches might break parachains at some point, even when Polkadot itself does not modify its own source code, but the dependency does. **This is a bad practice that must be avoided at all cost.**

For instance, a parachain running `cargo update` would break after new commits have been added to the master branch, as the dependency (or sub-dependency in this case) is not pinned to a stable commit hash.

In this scenario, any parachain trying to migrate to `v1.11.0` will break, and the fix is not straightforward. This also means that the `polkadot-v1.11.0` is not usable. Check [here](zhttps://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.11.0/substrate/client/network/Cargo.toml#L62). Same for the `release-polkadot-v1.11.0` branch as it is right now.

alexggh: Ok, I see your usecase, I thought you were just compiling polkadot-sdk from release branch and that should not be broken.

The fix seems about right, but  it seems you updated a lot more dependencies in the Cargo.lock, you need to make that litep2p is the only one changing.

FYI @lexnv @dmitry-markin @EgorPopelyaev.

alexggh: Until, this gets fix in the release branch the fastest way to unblock you would be to add a cargo patch in your main Cargo.toml that pins litep2p to a compatible version, something like this should fix your build.
```
[patch."https://github.com/paritytech/litep2p"]
litep2p = { git = "https://github.com/paritytech//litep2p", tag = "COMPATIBLE_TAG" }
```

metricaez: Understood, yes a `Cargo.lock` update is implicit since it was rebuilt on testing, I can clean it :)

lexnv:
I've created a PR to unblock this use-case by bringing the latest version: https://github.com/paritytech/polkadot-sdk/pull/4344 to substrate.

The breaking changes introduced by the PeerRecord where needed to have parity between libp2p and litep2p, as well as to unblock https://github.com/paritytech/polkadot-sdk/pull/3786 (which will be the main user of the peer records). The lastest version contains other fixes would be beneficial for litep2p users (mainly a panic on different than expected keys).

I've created a new PR since this one includes lots of dependency updates, which leads me to believe a `cargo update` was executed. We normally have to review and vet all crate version bumps to not depend on unwanted code.

Ideally, we'll switch soon the litep2p to a crate release schedule instead of depending on the master branch / having git dependencies which will make things more stable for end-users.

In the meanwhile, if you are interested in using litep2p you could specify it in the command line like `--network-backend litep2p`, your feedback would be greatly appreciate and would help move things faster 🙏 

Again, apologies for the inconvenience and hope that PR fixes the issue, let us know if you need further assistance

PR #4345, opened 2024-05-01
fix(chain-spec): ChainSpecBuilder with object as default genesis

saiintbrisson:
The current `json_patch::merge` function will replace the default configuration with the builder action if the action is not a JSON Object, which means, if not configured, the current code will try to use `Value::Null` as the config, which (almost?) always fails, making it mandatory to configure `with_genesis_config...`.

This change uses a JSON Object as the default values, which means the code will continue using the default config if none is given.


bkchr (approved):
Could you please add some test and some comment?

lexnv (approved):

michalkucharczyk (approved):

PR #4349, opened 2024-05-02
`remote-externalities`: store block header in snapshot

liamaharon:
The block header is required to derive inherents for a relay chain next block, this is useful in testing environments.


substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -58,37 +58,37 @@ type ChildKeyValues = Vec<(ChildInfo, Vec<KeyValue>)>;
 type SnapshotVersion = Compact<u16>;
 
 const LOG_TARGET: &str = "remote-ext";
-const DEFAULT_HTTP_ENDPOINT: &str = "https://rpc.polkadot.io:443";
-const SNAPSHOT_VERSION: SnapshotVersion = Compact(3);
+const DEFAULT_HTTP_ENDPOINT: &str = "https://polkadot-try-runtime-node.parity-chains.parity.io:443";
```
liamaharon, line 61: `https://rpc.polkadot.io:443` is not usable anymore due to increased rate-limiting.

ggwpez: Not sure it helps much to have a default when this can be used on any chain?

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -1255,7 +1265,7 @@ mod tests {
 
 		Builder::<Block>::new()
 			.mode(Mode::Offline(OfflineConfig {
-				state_snapshot: SnapshotConfig::new("test_data/proxy_test"),
+				state_snapshot: SnapshotConfig::new("test_data/test.snap"),
```
ggwpez, line 1268: Screenshot 2024-05-06 at 17 00 21

Dunno if something like this could be worth it: https://docs.rs/include-bytes-zstd/latest/include_bytes_zstd/

substrate/utils/frame/remote-externalities/src/lib.rs
liamaharon: It's nice to not require specifying it when you just want to run the tests and don't care about which chain

substrate/utils/frame/remote-externalities/src/lib.rs
```diff
@@ -58,37 +58,37 @@ type ChildKeyValues = Vec<(ChildInfo, Vec<KeyValue>)>;
...
-struct Snapshot<H> {
+struct Snapshot<B: BlockT> {
 	snapshot_version: SnapshotVersion,
 	state_version: StateVersion,
-	block_hash: H,
 	// <Vec<Key, (Value, MemoryDbRefCount)>>
 	raw_storage: Vec<(Vec<u8>, (Vec<u8>, i32))>,
-	storage_root: H,
+	storage_root: B::Hash,
+	header: B::Header,
```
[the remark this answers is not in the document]
liamaharon: If storage_root is the same as the state_root, then yes. I wasn't sure and wanted to keep this as backwards compatible as possible, but can remove it if you think worth it.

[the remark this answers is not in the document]
liamaharon: I tried this today, it *almost* works but breaks the ability to use snapshots containing only a subset of storage, e.g. only downloading the storage of select pallets.

substrate/utils/frame/remote-externalities/src/lib.rs
liamaharon: Git is pretty efficient, so I'd say not in this case :)

ggwpez (approved):

bkchr (approved):

Issue #4297, opened 2024-04-25
Investigate usage of `xcm::v4`


bkontur:
I found two cases for investigation:

for coretime stuff
- I think we should change to `v4` to `lastest` to avoid unnecessary problems when adding v5
```
use xcm::v4::{send_xcm, Instruction, Junction, Location, OriginKind, SendXcm, WeightLimit, Xcm};
``` 
for `query_acceptable_payment_assets`
- `if !matches!(xcm_version, 3 | 4)` or `Ok([VersionedAssetId::V4(`, I think we should use also `VersionedAssetId::from` for latest

seadanda:
There should be no issues replacing this on the coretime side

PR #3872, opened 2024-03-27
XcmDryRunApi - Dry-running extrinsics to get their XCM effects

franciscoaguirre:
# Context

Estimating fees for XCM execution and sending has been an area with bad UX.
The addition of the [XcmPaymentApi](https://github.com/paritytech/polkadot-sdk/pull/3607) exposed the necessary components to be able to estimate XCM fees correctly, however, that was not the full story.
The `XcmPaymentApi` works for estimating fees only if you know the specific XCM you want to execute or send.
This is necessary but most UIs want to estimate the fees for extrinsics, they don't necessarily know the XCM program that's executed by them.

# Main addition

A new runtime API is introduced, the `XcmDryRunApi`, that given an extrinsic, or an XCM program, returns its effects:
- Execution result
- Local XCM (in the case of an extrinsic)
- Forwarded XCMs
- List of events

This API can be used on its own for dry-running purposes, for double-checking or testing, but it mainly shines when used in conjunction with the `XcmPaymentApi`.
UIs can use these two APIs to estimate transfers.

# How it works

New tests are added to exemplify how to incorporate both APIs.
There's a mock test just to make sure everything works under `xcm-fee-payment-runtime-api`.
There's a real-world test using Westend and AssetHubWestend under `cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/xcm_fee_estimation.rs`.
Added both a test for a simple teleport between chains and a reserve transfer asset between two parachains going through a reserve.

The steps to follow:
- Use `XcmDryRunApi::dry_run_extrinsic` to get local XCM program and forwarded messages
- For each forwarded message
  - Use `XcmPaymentApi::query_delivery_fee` LOCALLY to get the delivery fees
  - Use `XcmPaymentApi::query_xcm_weight` ON THE DESTINATION to get the remote execution weight
  - (optional) Use `XcmPaymentApi::query_acceptable_payment_assets` ON THE DESTINATION to know on which assets the execution fees can be paid
  - Use `XcmPaymentApi::query_weight_to_asset_fee` ON THE DESTINATION to convert weight to the actual remote execution fees
  - Use `XcmDryRunApi::dry_run_xcm` ON THE DESTINATION to know if a new message will be forwarded, if so, continue

# Dear reviewer

The changes in this PR are grouped as follows, and in order of importance:
- Addition of new runtime API
  - Definition, mock and simple tests: polkadot/xcm/xcm-fee-payment-runtime-api/*
[...]

PR #4355, opened 2024-05-02
More `xcm::v4` cleanup and `xcm_fee_payment_runtime_api::XcmPaymentApi` nits

bkontur:
This PR:
- changes `xcm::v4` to `xcm::prelude` imports for coretime stuff
- changes `query_acceptable_payment_assets` / `query_weight_to_asset_fee` implementations to be more resilient to the XCM version change
- adds `xcm_fee_payment_runtime_api::XcmPaymentApi` to the AssetHubRococo/Westend  exposing a native token as acceptable payment asset

Continuation of: https://github.com/paritytech/polkadot-sdk/pull/3607

Closes: https://github.com/paritytech/polkadot-sdk/issues/4297

## Possible follow-ups

- [ ] add all sufficient assets (`Assets`, `ForeignAssets`) as acceptable payment assets ?

### DIFF
--- a/polkadot/runtime/westend/src/lib.rs
+++ b/polkadot/runtime/westend/src/lib.rs
@@ -2199,24 +2199,32 @@ sp_api::impl_runtime_apis! {
 
 	impl xcm_fee_payment_runtime_api::XcmPaymentApi<Block> for Runtime {
 		fn query_acceptable_payment_assets(xcm_version: xcm::Version) -> Result<Vec<VersionedAssetId>, XcmPaymentApiError> {
-			if !matches!(xcm_version, 3 | 4) {
-				return Err(XcmPaymentApiError::UnhandledXcmVersion);
-			}
-			Ok([VersionedAssetId::V4(xcm_config::TokenLocation::get().into())]
+			let acceptable = vec![
+				// native token
+				VersionedAssetId::from(AssetId(xcm_config::TokenLocation::get()))
+			];
+
+			Ok(acceptable
 				.into_iter()
 				.filter_map(|asset| asset.into_version(xcm_version).ok())
 				.collect())
 		}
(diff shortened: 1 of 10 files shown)


polkadot/runtime/westend/src/lib.rs
```diff
@@ -2199,24 +2199,32 @@ sp_api::impl_runtime_apis! {
...
+				},
+				Ok(asset_id) => {
+					log::trace!(target: "xcm::xcm_fee_payment_runtime_api", "query_weight_to_asset_fee - unhandled asset_id: {asset_id:?}!");
+					Err(XcmPaymentApiError::AssetNotFound)
+				},
+				Err(_) => {
+					log::trace!(target: "xcm::xcm_fee_payment_runtime_api", "query_weight_to_asset_fee - failed to convert asset: {asset:?}!");
+					Err(XcmPaymentApiError::VersionedConversionFailed)
+				}
+			}
```
serban300, line 2227: Looks like this code is duplicated for Rococo and Westend. Could we move it into a method and reuse it ?

Seems to also be the case for `query_acceptable_payment_assets`.

franciscoaguirre (approved):
Looks good. We should do the follow-up later. I still need to merge https://github.com/paritytech/polkadot-sdk/pull/3872 and create a follow-up where I implement this and the payment api on all runtimes. So it could probably go there.

polkadot/runtime/westend/src/lib.rs
serban300: Oh, I see. Then I think we can see how we can make this more generic later when we add this for multiple runtimes.

serban300 (approved):

seadanda (approved):

PR #4358, opened 2024-05-02
wip/PoC: Enrich metadata custom types with essential types for chain communications

lexnv:
This PR enriches the metadata's custom types via captured types needed by users to communicate with the chain.
Ideally, this will naturally evolve into a strongly typed metadata V16.

Subxt relies on a preset called `Config` to interact with the chain.
This `Config` present contains types such as `AccountId`, `Hasher` and more needed to craft transactions.
Subxt has implemented a `SubstrateConfig` and `PolkadotConfig`.

This is the subxt [config](https://github.com/paritytech/subxt/pull/1566/files#diff-88a79c5a0c04b7ac18e28856b68cb3be44b4bae69e832344415a3fc50941ef5bR14-R35) that can be almost entirely replaced by metadata types.

The end goal of this PoC is to remove as many types needed by those configs and rely on the metadata provided information instead.

For more details, see https://github.com/paritytech/subxt/pull/1566.

### DIFF
--- a/substrate/frame/support/procedural/src/construct_runtime/expand/metadata.rs
+++ b/substrate/frame/support/procedural/src/construct_runtime/expand/metadata.rs
@@ -181,6 +181,18 @@ pub fn expand_runtime_metadata(
 								"Signature".into(),
 								signature_ty,
 							),
+							(
+								"Hash".into(),
+								#scrate::__private::scale_info::meta_type::<
+										<#runtime as #system_path::Config>::Hash
+									>(),
+							),
+							(
+								"Hashing".into(),
+								#scrate::__private::scale_info::meta_type::<
+										<#runtime as #system_path::Config>::Hashing
+									>(),
+							),
 							#maybe_asset_id
 						]
 						.into_iter()
--- a/substrate/frame/support/procedural/src/construct_runtime/expand/metadata.rs
+++ b/substrate/frame/support/procedural/src/construct_runtime/expand/metadata.rs
@@ -27,6 +27,7 @@ pub fn expand_runtime_metadata(
 	scrate: &TokenStream,
 	extrinsic: &TokenStream,
 	system_path: &PalletPath,
+	assets_path: Option<PalletPath>,
 ) -> TokenStream {
 	let pallets = pallet_declarations
 		.iter()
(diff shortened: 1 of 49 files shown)


jsdw:
I'd leave this as a draft for now since we'll prob want to think about and iterate on it a bit etc :)

substrate/frame/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -226,6 +226,10 @@ use syn::{spanned::Spanned, Ident, Result};
 
 /// The fixed name of the system pallet.
 const SYSTEM_PALLET_NAME: &str = "System";
+/// The fixed name of the ForeignAssets pallet.
+const FOREIGN_ASSETS_PALLET_NAME: &str = "ForeignAssets";
+/// The fixed name of the Assets pallet.
+const ASSETS_PALLET_NAME: &str = "Assets";
```
bkchr, line 232: This is nothing we want to integrate. Users will have no fucking idea that they should call the pallets this way. These pallets can also have multiple instances. This doesn't makes any sense and we should not start doing this.

bkchr (changes requested):

substrate/frame/support/procedural/src/construct_runtime/expand/metadata.rs
```diff
@@ -140,6 +159,43 @@ pub fn expand_runtime_metadata(
...
+									>(),
+							),
+							(
+								"Address".into(),
+								address_ty,
+							),
+							(
+								"Signature".into(),
+								signature_ty,
+							),
```
bkchr, line 178: They already exist in the extrinsic metadata?

substrate/frame/support/procedural/src/construct_runtime/mod.rs
lexnv: Yep that makes sense! Its a bit of a hack to see how far I can get in subxt to remove those config types :D 

I think a better approach may be to collect the associated types from pallet's configs. However, I'll have to look how much that will increase the metadata size and if we run into limitations with it (maybe not all types impl TypeInfo for us to expose these), will think about it! Thanks for the review! 🙏

lexnv:
Superseded by: https://github.com/paritytech/polkadot-sdk/pull/5274

PR #4361, opened 2024-05-02
Change sorted list types and simplify type conversions in stake-tracker

gpestana:
Changes the voter and sorted lists score types. 

- Target list score from `Balance` to `ExtendedBalance`: for safety.
- Voter list score from `VoteWeight` to `Balance`: to avoid conversions to vote weight. (can we do this though? what if the vote weight is > u64)

This way, the conversion to `VoteWeight` (which requires the a call to `T::Currency::total_issuance`) is reduced to a minimum and never required at the stake-tracker level. The target list score is `ExtendedBalance` for safety, as the approvals stakes with unbounded nominations may overflow u64.

Note that the end of the day, the voter and target lists are used to keep stakers sorted by stake/approval stakes.

Converting the voter score from `VoteWeight` to `Balance` requires a migration since vote_weight(account) != active_stake(account)`

### DIFF
--- a/substrate/frame/staking/src/migrations/single_block.rs
+++ b/substrate/frame/staking/src/migrations/single_block.rs
@@ -406,7 +406,7 @@ pub mod v9 {
 				let weight_of_cached = Pallet::<T>::weight_of_fn();
 				for (v, _) in Validators::<T>::iter() {
 					let weight = weight_of_cached(&v);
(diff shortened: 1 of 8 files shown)


gpestana:
closing as this change will not be introduced in stake tracker. Opened a tracking issue to tackle this refactor later on https://github.com/paritytech/polkadot-sdk/issues/4725

PR #4363, opened 2024-05-02
Add Async Backing guide

DrW3RK:
Update the instructions to work with the latest parachain template on Polkadot SDK


DrW3RK:
I have embedded the code from Parachain template wherever applicable using docify.

@kianenigma Perhaps the Async Guide upgrade doc is not a great candidate to showcase docify as most of the hardcoded example snippets serve as guidance for parachain devs to edit the code on their repositories which may be built from older versions of Substrate SDK. Here are a few pitfalls of using docify on this document:

- Can't highlight specific lines like the example in [step 1.4](https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-1---update-parachain-runtime)
- Can't make edits to the exported code. This is problematic for guides like these, as it is very common that the readers will copy paste the snippet. Reference example - [step 1.5](https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-1---update-parachain-runtime) where the exported code shows "TRUE" and we want the reader to have it set to "FALSE"
- I was unable to export "spawn-tasks" in  [step 2.2](https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-2---update-parachain-nodes) and "AuraParams" in step 2.5
- Can't show partial code snippet like in [step 2.3](https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-2---update-parachain-nodes). The function `start_consensus` is huge and we just want to instruct the reader to add a single line of extra code to it.
- How to export the comments/documentation from the source code as well? docify just captures the actual code and ignores the documentation.

Any input on the above items is appreciated. Will be helpful for future docs.

@alexggh If you are able to do a walk through of the rendered doc and ensure its integrity, that would be of huge help. Thank you!

kianenigma:
Your observation is a bit unfortunate, but correct in that I might have been wrong in thinking that this doc is better demonstrated in docify. 

Some of the points you mentioned can be improved by enhancing docify. For example: 

- `#[docify::hidden]` to hide unneeded stuff
- `#[docify::diff]` to showcase just the diff of two snippets#4365 

I had the intention to ask for a small treasury tip to incentivize @sam0x17 to implement some of these. But this is a midterm effort and I don't have a better solution in the short term. 

My only suggestion in the short term is to not expect to re-use production code as-is in docify, and sometimes you might need to first relocate a subset of a working code in a new module/crate, and then use that. 

Whether this is worth the effort or not, we will have to evaluate. Thank you for giving this a try in the meantime!

DrW3RK:
Sounds good! Docify is for sure very resourceful in embedding code, especially within the tutorial docs. If we are making a tutorial on Parachain template itself, I do see the potential of using docify.

I see you have created issues on the docify repo https://github.com/sam0x17/docify/issues Let's ask for a treasury tip to have them addressed!

docs/sdk/src/guides/parachains/async_backing_guide.rs
```diff
@@ -0,0 +1,368 @@
...
+//!    `/runtime/src/lib.rs`.
+//!    
+//!    ```text
+//!    //!/ Maximum number of blocks simultaneously accepted by the Runtime, not yet included into the
+//!    //!/ relay chain.
+//!    pub const UNINCLUDED_SEGMENT_CAPACITY: u32 = 1;
+//!    //!/ How many parachain blocks are processed by the relay chain per parent. Limits the number of
+//!    //!/ blocks authored per slot.
+//!    ```
+#![doc = docify::embed!("../../templates/parachain/runtime/src/lib.rs", velocity)]
```
alexggh, line 98: Screenshot 2024-05-09 at 12 52 20

This rendering is weird and there are a few other places where this happen.

I don't think we get any benefit from referencing constants in the templates parachains, this guide is for parachains that don't use the template.

alexggh (approved):
The content is correct, but I think the rendering is a bit weird and makes it hard to pass through.

docs/sdk/src/guides/parachains/async_backing_guide.rs
DrW3RK: You're right. I tried to use docify where I could, and like I noted here https://github.com/paritytech/polkadot-sdk/pull/4363#issuecomment-2091891735 it may not be necessary. Let me work on a better rendering here and then decide if we can merge this or just keep the guide on the Wiki.

DrW3RK:
@kianenigma The doc is ready to be merged. Please review.

docs/sdk/src/guides/parachains/async_backing_guide.rs
```diff
@@ -0,0 +1,452 @@
...
+//! async backing compatibility. Latest on master branch of
+//! [Polkadot-SDK](https://github.com/paritytech/polkadot-sdk) is currently sufficient. Any 2024 release
+//! will work as well.
+//!
+//! ## Async Backing Terminology and Parameters
+//!
+//! Time for a bit of context before we get started. The following concepts will aid in demystifying the
+//! collator side of Async Backing and establish a basic understanding of the changes being made:
+//!
+//! - **Unincluded segment** - From the perspective of a parachain block under construction, the
```
kianenigma, line 26: Why do parachain developers need to know these concepts?

I would argue 99% of the possible readers of this page don't want to know this information, and rather just know: 

1. with async backing I can have faster block times
2. how shoudl I enable it. 

 I feel like this type of information is out of scope both for here and in the wiki, and should be in an in-depth writing like parachain's implementor guide.

docs/sdk/src/guides/mod.rs
```diff
@@ -26,3 +26,6 @@ pub mod xcm_enabled_parachain;
 
 /// How to enable storage weight reclaiming in a parachain node and runtime.
 pub mod enable_pov_reclaim;
+
+/// Parachain guides
+pub mod parachains;
```
kianenigma, line 31: For now I would suggest keeping the structure flat, as something like PoV reclaim is also about parachains. 

Although sadly the structure of the modules does break links, so I hope we can stick to one decision.

docs/sdk/src/guides/parachains/async_backing_guide.rs
DrW3RK: Would you happen to know where the GitHub repo for these docs is? https://paritytech.github.io/polkadot-sdk/book/index.html The top right link points to the Polkadot SDK repo. 

For now, I can remove this content from this guide and keep in Wiki, till we can find a home for them on Parachain Implementors guide.

kianenigma:
I have some further ideas about how to improve, you can take a look here: https://github.com/DrW3RK/polkadot-sdk/compare/async-backing-docs...paritytech:polkadot-sdk:kiz-feedback-kian-radha

kianenigma:
@DrW3RK in the commit of mine that you have merged, I have demonstrated how you can use docify, but I have left the rest of it to do. Please follow the same pattern, and try and make as many of the code snippets as you can use real code.

DrW3RK:
on it

DrW3RK:
@kianenigma I've refined the document based on your feedback and used `docify `where it made sense. 

On closer look, I observed that a few of the Docify imports in Phase 1 from your commit were misleading. As @alexggh mentioned here: https://github.com/paritytech/polkadot-sdk/pull/4363#discussion_r1595242205, this guide is for parachains that have not used the template and need to make modifications in multiple phases to activate the Async Backing feature.

For instance, in Phase 1, the capacity is already 3 instead of 1, and the block time is 6000 instead of 12000.

image

I've attempted to make the code snippets concise throughout.

However, I failed to get this part of your commit to compile.

> 6. Update [`sp_consensus_aura::AuraApi::slot_duration`] in [`sp_api::impl_runtime_apis`] to match the constant `SLOT_DURATION`

and got these errors. I can investigate this further, but if this is something trivial, please let me know. Thanks!

```
  = note: no item named `sp_consensus_aura` in scope
note: the lint level is defined here
   --> docs/sdk/src/guides/async_backing_guide.rs:252:9
    |
252 | #![deny(rustdoc::broken_intra_doc_links)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

kianenigma (approved):

DrW3RK:
@kianenigma This was a good learning exercise for me on how Docify works. Happy to put this to use in Polkadot SDK docs.

kianenigma (approved):

DrW3RK:
Could we add R0-silent and T11-documentation labels to this PR similar to https://github.com/paritytech/polkadot-sdk/pull/4244 ? I think that should take care of the failing CI jobs.

alindima:
> Could we add R0-silent and T11-documentation labels to this PR similar to #4244 ? I think that should take care of the failing CI jobs.

Done. This can be queued for merge now

Issue #4340, opened 2024-04-30
Persistent Bonding State Issue with Zero Existential Deposit


benefacto:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

We are building a new layer 1 using Frontier EVM alongside pallet_staking. We are using a custom Substrate token as its native token. We have set our existential deposit to zero to prevent dusting. However, we've identified an issue related to storage writes in the staking process, specifically with the `Self::bonded` entry. According to the [documentation](https://paritytech.github.io/polkadot-sdk/master/pallet_staking/struct.Pallet.html#method.bonded), this entry is not cleared unless the balance falls below the existential deposit and is removed as dust. With an existential deposit of zero, these entries are never cleaned, which leads to a situation where stakers remain indefinitely bonded with a stake of zero after they chill and unbond. This results in unnecessary storage overhead and potential unknown issues.

**Expected behavior:**
After an account is fully unbonded, it should not remain in a bonded state if it has a zero stake.

**Actual behavior:**
Accounts remain bonded with zero stake due to the non-removal of `Self::bonded` because the existential deposit is zero.

**Potential impact:**
This could lead to increased storage overhead and might cause other unknown issues in network behavior or staking mechanics.

### Steps to reproduce

1. Extend the [frontier node template](https://github.com/polkadot-evm/frontier/tree/master/template) by configuring `pallet_staking`.
2. Follow the staking, validating, chilling, and unbonding process with any account. Example steps:
   - Bond an account with a minimum stake value and set it as a validator.
   - After some operations and era transitions, chill and unbond the account.
   - Check the bonded state of the account after unbonding is completed.
   - The account remains bonded with a stake of zero which is not the expected behavior.

kianenigma:
cc @Ank4n @gpestana

Ank4n:
Yes, seems staking pallet has not been written with expectation that ED can be zero. 
Fix is to do a `<=` check with ED or an explicit zero check (safer as it is noop for runtimes with `ed > 0`).

PR #4364, opened 2024-05-03
Fix: dust unbonded for zero existential deposit

krisbitney:
When a staker unbonds and withdraws, it is possible that their stash will contain less currency than the existential deposit. If that happens, their stash is reaped. But if the existential deposit is zero, the reap is not triggered. This PR adjusts `pallet_staking` to reap a stash in the special case that the stash value is zero and the existential deposit is zero.

This change is important for blockchains built on substrate that require an existential deposit of zero, becuase it conserves valued storage space.

There are two places in which ledgers are checked to determine if their value is less than the existential deposit and they should be reaped: in the methods `do_withdraw_unbonded` and `reap_stash`. When the check is made, the condition `ledger_total == 0` is also checked. If `ledger_total` is zero, then it must be below any existential deposit greater than zero and equal to an existential deposit of 0.

I added a new test for each method to confirm the change behaves as expected.

Closes https://github.com/paritytech/polkadot-sdk/issues/4340

### DIFF
--- a/prdoc/pr_4364.prdoc
+++ b/prdoc/pr_4364.prdoc
@@ -1,0 +1,10 @@
+title: Fix: dust unbonded for zero existential deposit
+
+doc:
+  - audience: Runtime Dev
+    description: |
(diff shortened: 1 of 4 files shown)


krisbitney:
I see that a rustfmt check is failing, but running `cargo fmt` updates many files that aren't relevant to the change introduced by this PR. I think rustfmt for the entire repo should be done in a separate PR.

Ank4n (approved):

bkchr:
bot fmt

bkchr (approved):

PR #4365, opened 2024-05-03
Bridge: added subcommand to relay single parachain header

svyatonik:
Related to https://github.com/paritytech/parity-bridges-common/issues/2962
Relay companion: https://github.com/paritytech/parity-bridges-common/pull/2978

Example usage:
```
./target/release/substrate-relay relay-parachain-head rococo-to-bridge-hub-westend \
    --source-host localhost --source-port 9942 \
    --target-host localhost --target-port 8945 --target-signer //Alice \
    --at-relay-block 61
```

### DIFF
--- a/bridges/relays/lib-substrate-relay/src/parachains/mod.rs
+++ b/bridges/relays/lib-substrate-relay/src/parachains/mod.rs
@@ -19,8 +19,14 @@
 
 use async_trait::async_trait;
 use bp_polkadot_core::parachains::{ParaHash, ParaHeadsProof, ParaId};
+use pallet_bridge_parachains::{
+	Call as BridgeParachainsCall, Config as BridgeParachainsConfig, RelayBlockHash,
+	RelayBlockHasher, RelayBlockNumber,
+};
 use parachains_relay::ParachainsPipeline;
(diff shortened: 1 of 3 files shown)


acatangiu (approved):

bridges/relays/lib-substrate-relay/src/parachains/mod.rs
```diff
@@ -74,37 +68,3 @@ pub trait SubmitParachainHeadsCallBuilder<P: SubstrateParachainsPipeline>:
...
-		_is_free_execution_expected: bool,
-	) -> CallOf<P::TargetChain> {
-		BridgeParachainsCall::<R, I>::submit_parachain_heads {
-			at_relay_block: (at_relay_block.0, at_relay_block.1),
-			parachains,
-			parachain_heads_proof,
-		}
-		.into()
-	}
-}
```
serban300, line 110: Can we keep this ? I think it will be needed for https://github.com/paritytech/parity-bridges-common/issues/2952

svyatonik: Makes sense - I've reverted removal. I've also recently removed similar struct for GRANDPA submissions - you may need it as well.

serban300 (approved):

CharlesCowan (commented):
🙂‍↕️

PR #3366, opened 2024-02-16
Lift dependencies to the workspace (Part 2/x)

ggwpez:
Lifting some more dependencies to the workspace. Just using the most-often updated ones for now.  
It can be reproduced locally.

```sh
# First you can check if there would be semver incompatible bumps (looks good in this case):
$ zepter transpose dependency lift-to-workspace --ignore-errors syn quote thiserror "regex:^serde.*"

# Then apply the changes:
$ zepter transpose dependency lift-to-workspace --version-resolver=highest syn quote thiserror "regex:^serde.*" --fix

# And format the changes:
$ taplo format --config .config/taplo.toml
```

PR #4374, opened 2024-05-03
Align few crates versions

jasl:
Closes #3414 

@bkchr Here it is, I miss that DependentBot breaks compiling.

I also found there are a few dependencies in the codebase that use different versions.
This PR aims to align them to the highest version (in the code base)

I separate them into individual commits, so I can remove the changes you don't like


bkchr (approved):

jasl:
Sorry, add a few more. This is enough for now

liamaharon (commented):
Any reason to not just switch to workspace deps?

jasl:
> Any reason to not just switch to workspace deps?

if you guys want, I can help to switch to workspace deps

ggwpez:
Yea it would probably make more sense to pull them up first. Hope it can be automated like here: https://github.com/paritytech/polkadot-sdk/pull/3366

jasl:
> Yea it would probably make more sense to pull them up first. Hope it can be automated like here: #3366

Do you have any rules about which crates should be kept in the project's own `Cargo.toml`?

liamaharon:
> > Yea it would probably make more sense to pull them up first. Hope it can be automated like here: #3366
> 
> Do you have any rules about which crates should be kept in the project's own `Cargo.toml`?

Prob only if they would never be used in any other crate, I can't think of any examples.

jasl:
> > > Yea it would probably make more sense to pull them up first. Hope it can be automated like here: #3366
> > 
> > 
> > Do you have any rules about which crates should be kept in the project's own `Cargo.toml`?
> 
> Prob only if they would never be used in any other crate, I can't think of any examples.

Understood, `tokio-tungstenite` is just an example, it's only used in `zombienet-backchannel`

I'll do it later.

ggwpez:
One thing to keep in mind that larger merge-requests normally take longer to review. We had a few raw-hammer merge requests that ended up abandoned since the tried to do too much at once.  
Its probably best to do it for a few deps per merge-request.

jasl:
This should cover most, I'll split them tomorrow, and close this one

liamaharon:
> Understood, tokio-tungstenite is just an example, it's only used in zombienet-backchannel

I'd still make this a workspace dep, the dep is general and just because it isn't used by any other package today doesn't mean that won't change in the future

PR #4375, opened 2024-05-03
Support different assets for delivery fees

franciscoaguirre:
# DRAFT

This PR is still under progress.


cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -349,6 +351,33 @@ pub type TrustedTeleporters = (
...
+	pallet_assets::BalanceToAssetBalance<Balances, Runtime, ConvertInto, ForeignAssetsInstance>,
+	ForeignAssetsInstance,
+>;
+
+/// Asset converter for pool assets.
+/// Used to convert assets in pools to the asset required for fee payment.
+/// The pool must be between the first asset and the one required for fee payment.
+/// This type allows paying fees with any asset in a pool with the asset required for fee payment.
+// TODO: Finish.
+pub type PoolAssetsConverter = ();
```
joepetrowski, line 379: Looks like you are on the right track doing this. With respect to `*SufficientAssetsConverter`, the idea in the [Trader](https://github.com/paritytech/polkadot-sdk/blob/da83c684e9c04dc3a0dfbd36bb7d50b15fe7c75f/cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs#L400) was to remove `TakeFirstAssetTrader` and keep `SwapFirstAssetTrader`. They were both kept in for one release cycle because there is some time required for the pools to form and we didn't want sufficient asset support to break.

Basically, the intent is that asset sufficiency be related only to account existence and be decoupled from fee payment. *Any* asset should be equal for fee payment and use the "real" value through the pool, not the "fixed" value we use for sufficient ones.

joepetrowski: Well, one release cycle of Polkadot/Kusama, obviously many of Westend 😬 . We should probably remove them in another PR when we do for Polkadot.

franciscoaguirre: Sounds good, I'll make a test working with pools. We can easily have these `*SufficientAssetsConverter` in the fellowship runtimes or just go with the pools one.

Polkadot-Forum:

https://forum.polkadot.network/t/xcm-user-and-developer-experience-improvements/4511/21

Bullrich:
bot help

muharem (commented):
What if we disable the `FeeManager` for our runtime setups and try to account a transport weights in `Weigher`. Even if we cannot do it properly, may be some reasonable constants would be enough.
For local execution the transport fee will be included in extrinsics' weights, and for the remote execution a user will pay with `BuyExecution` for transport fee as well.
To avoid a double payment with `jit_withdraw`  fee mode, we will need to disable it via config (here we also can provide some info for dependent teams).

After reading the weights payment model in the xcm pallet and the executor, I do not see a good way to make it work with various assets through the FeeManager without breaking the APIs. With the new xcm version we can have a better solution.

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +467,88 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+			// Is a breaking change.
+			// Option 4: Create a new entrypoint `prepare_and_execute_with_fee_asset`.
+			// Should work and is not a breaking change.
+			// TODO: Might increase benchmarks too much. Should optimize for the average case.
+			let mut asset_for_fees = first.clone();
+			// Holding is empty when using jit withdrawal. Could remove it.
+			// It's not empty because of jit withdrawal, it's empty because it's taken from holding before sending.
+			// That's the problem jit withdrawal was solving, we need to "park" delivery fees in the same way we did
+			// for `DepositReserveAsset`.
+			for asset in self.holding.fungible_assets_iter() {
```
muharem, line 518: This way we basically deciding for a user which asset to use for the payment, this can be a surprise.

franciscoaguirre: Yeah, but some API needs to change or I have to add a new one to not have to do this.
I was thinking of creating a new entrypoint `prepare_and_execute_with_fee_asset` that `pallet-xcm` can use.
That'd keep the API the same for all other teams.
What do you think?

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -852,8 +934,19 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 					let (_, fee) =
 						validate_send::<Config::XcmSender>(dest.clone(), Xcm(message_to_weigh))?;
+					let first = fee.get(0).ok_or(XcmError::AssetNotFound)?;
+					let asset_id = self.asset_for_fees.as_ref().unwrap_or(&first.id);
+					// TODO: Deal with this case.
+					// Need to make a test specifically for this.
+					let actual_asset_to_use_for_fees = match Config::AssetConverter::convert_asset(&first, asset_id) {
+						Ok(new_asset) => new_asset,
+						Err(error) => {
+							log::error!(target: "xcm::DepositReserveAsset", "What happened?");
```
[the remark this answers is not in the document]
franciscoaguirre: Yeah this case hasn't been done yet

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -945,6 +1070,9 @@ impl<Config: config::Config> XcmExecutor<Config> {
 				// should be executed.
 				let Some(weight) = Option::<Weight>::from(weight_limit) else { return Ok(()) };
 				let old_holding = self.holding.clone();
+				// Store the asset being used for fees, so delivery fees can access it.
+				self.asset_for_fees = Some(fees.id.clone());
```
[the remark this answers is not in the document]
franciscoaguirre: That's my main concern, and the reason why I did the "looping through all assets in holding" trick, which I know is bad.
If local executions had a `BuyExecution` instruction then this would be great.
My current thought is to create a new entrypoint `prepare_and_execute_with_fee_asset`.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2567,8 +2568,8 @@ impl<T: Config> Pallet<T> {
 	/// Fails if:
 	/// - the `assets` are not known on this chain;
 	/// - the `assets` cannot be withdrawn with that location as the Origin.
-	fn charge_fees(location: Location, assets: Assets) -> DispatchResult {
-		T::XcmExecutor::charge_fees(location.clone(), assets.clone())
+	fn charge_fees(location: Location, assets: Assets, asset_for_fees: &AssetId) -> DispatchResult {
+		T::XcmExecutor::charge_fees(location.clone(), assets.clone(), asset_for_fees)
```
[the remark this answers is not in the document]
muharem: here I mean on the pallet

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/hybrid_transfers.rs
```diff
@@ -275,6 +275,155 @@ fn transfer_foreign_assets_from_asset_hub_to_para() {
 	assert_eq!(receiver_rocs_after, receiver_rocs_before + foreign_amount_to_send);
 }
 
+// =================================================================================
+// ======= Reserve Transfers - Sufficient Foreign Asset - AssetHub<>Parachain ======
+// =================================================================================
+/// Reserve Transfers of a sufficient foreign asset from System Parachain to Parachain should
+/// work without using any other token.
```
acatangiu, line 282: This test transfers WND (native token) from AH to Parachain.

In practice the ecosystem (currently) has the native token registered as a "trust-backed" asset with the "trust" coming from the reserve Relay using either pallet-assets or pallet-tokens. I haven't seen anyone implement it using the new "foreign assets" model.

My point is that this particular emulated test is not practically relevant to the community - and I would change the tests to cover the following real-world usecases:

## Teleport (both ways) Parachain-native (PEN) between Parachain and AH (pay with PEN)

Teleport an asset - registered on AssetHub as foreignAsset - between AH and Parachains (do it both ways):
- use the parachain's native asset (let's say PEN)
- is also registered as foreign on AH
- there is a PEN/WND pool set up on AH

This ^ would be a very realistic use-case of what we want in practice: all parachains register their native tokens as foreign on AH, and using asset-conversion can do cross-chain transfers using their native token alone.

## Reserve-transfer (both ways) ecosystem-native (WND) between Parachain and AH (pay with WND)

This is already done in existing tests, but it is achieved by allowing buying weight with WND directly on Penpal (`TakeFirstAssetTrader`) - you could remove that and set up WND/PEN pool instead on Penpal (so it is more representative of real world). This would showcase how parachains could easily support fees in WND/DOT/KSM while still practically using their native token for buying weight (execution fees).

(basically what you did with `reserve_transfer_pool_assets_from_system_para_to_para()` test, but instead of adding new test using some random theoretical asset for the pool, create WND/PEN pool and make existing tests work without `TakeFirstAssetTrader`)

## Reserve-transfer (both ways) AH trust-backed USDT between two parachains going through AH (pay with USDT)

Again, real-world usecase - USDT used for fees using USDT/??? pools available on each hop.

You already covered this in `reserve_transfer_pool_assets_from_para_to_para_through_asset_hub()` 👍

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+		asset_id,
+		asset_owner,
+		true, // Mark it as sufficient.
+		70_000,
+		vec![(sender.clone(), ASSET_HUB_WESTEND_ED * 10000)],
+	);
+	let system_para_foreign_asset_location = Location::new(
+		1,
+		[Parachain(ASSET_HUB_ID), PalletInstance(ASSETS_PALLET_ID), GeneralIndex(asset_id.into())],
+	);
```
acatangiu, line 1184: You should just register USDT in genesis [like it is done for AH Rococo](https://github.com/paritytech/polkadot-sdk/blob/18ed309a37036db8429665f1e91fb24ab312e646/cumulus/parachains/integration-tests/emulated/chains/parachains/assets/asset-hub-rococo/src/genesis.rs#L73).

Then ready for use in tests

```suggestion
	let usdt_location = Location::new(
		1,
		[Parachain(ASSET_HUB_ID), PalletInstance(ASSETS_PALLET_ID), GeneralIndex(USDT_ID)],
	);
```

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+
+	// Create sufficient foreign asset.
+	PenpalA::force_create_foreign_asset(
+		system_para_foreign_asset_location.clone(),
+		PenpalAssetOwner::get(),
+		true, // Mark it as sufficient.
+		70_000,
+		vec![],
+	);
```
acatangiu, line 1193: I would say we should register USDT as foreign in Penpal in genesis too (and not have to do it manually in so many tests).

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+	assert!(receiver_assets_after < receiver_assets_before + amount_to_send);
+}
+
+// =========================================================================
+// ======= Reserve Transfers - Sufficient Asset - AssetHub->Parachain ======
+// =========================================================================
+/// Reserve Transfers of a sufficient asset from System Parachain to Parachain should
+/// work without using any other token.
+#[test]
+fn reserve_transfer_sufficient_assets_from_system_para_to_para() {
```
acatangiu, line 1161: This is already kind of covered by `reserve_transfer_sufficient_asset_from_para_to_para_through_asset_hub` - not sure if we need to maintain it

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+	// Give the sender enough Relay tokens to pay for local delivery fees.
+	// TODO: When we support local delivery fee payment in other assets, we don't need this.
+	PenpalA::execute_with(|| {
+		assert_ok!(<PenpalA as PenpalAPallet>::ForeignAssets::mint(
+			<PenpalA as Chain>::RuntimeOrigin::signed(PenpalAssetOwner::get()),
+			Location::new(1, []),
+			sender.clone().into(),
+			10_000_000_000_000, // Big estimate to make it work.
+		));
+	});
```
acatangiu, line 1355: You can just fund pallet-balances (PENs) here for local fees (more realistic)

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+		],
+	);
+	let penpal_location = AssetHubWestend::sibling_location_of(PenpalA::para_id());
+	let penpal_sov_account = AssetHubWestend::sovereign_account_id_of(penpal_location.clone());
+
+	// Create SA-of-Penpal-on-AHR with ED.
+	// This ED isn't reflected in any derivative in a PenpalA account.
+	AssetHubWestend::fund_accounts(vec![(penpal_sov_account.clone().into(), ASSET_HUB_WESTEND_ED)]);
+
+	// Setup the pool between `native_asset` and `pool_asset` on AssetHub.
```
acatangiu, line 1436: I added some "create-pool" helper fns in https://github.com/paritytech/polkadot-sdk/pull/4870 - you can reuse across tests

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+	});
+
+	// Sender's balance is reduced by amount sent plus delivery fees
+	assert!(sender_assets_after < sender_assets_before - asset_amount_to_send);
+	// Receiver's balance is increased
+	assert!(receiver_assets_after > receiver_assets_before);
+}
+
+// ===============================================================
+// ===== Reserve Transfers - Pool Asset - AssetHub->Parachain ====
```
acatangiu, line 1416: Also nice, but in practice, this random pool asset used to pay for fees on Parachains is either DOT or USDT, so I would use DOT/WND to cover this scenario.

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+	assert!(receiver_foreign_assets_after > receiver_foreign_assets_before);
+}
+
+// ================================================================================
+// ===== Reserve Transfers - Sufficient Asset - Parachain->AssetHub->Parachain ====
+// ================================================================================
+/// Reserve Transfers of a sufficient trust-backed asset from Parachain to Parachain (through Relay
+/// reserve) should work
+#[test]
+fn reserve_transfer_sufficient_asset_from_para_to_para_through_asset_hub() {
```
acatangiu, line 1282: It's a bit confusing to use USDT in multiple tests, sometimes buy through pool, sometimes through direct weight-trader.

I would remove this test and keep only `reserve_transfer_pool_assets_from_para_to_para_through_asset_hub` - that is the "real world use case" one.

Also, "sufficient" (now) only means that asset can cover ED - not that it can be used to buy weight directly - we should be careful to not propagate this misconception that sufficient means it can pay for weight directly 🙈

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,666 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+	assert!(receiver_after_balance > receiver_initial_balance);
+	assert!(receiver_after_balance < receiver_initial_balance + asset_amount_to_send);
+}
+
+// ==========================================================================
+// ===== Reserve Transfers - Pool Asset - Parachain->AssetHub->Parachain ====
+// ==========================================================================
+// From Penpal A to Penpal B with AssetHub as the reserve.
+#[test]
+fn reserve_transfer_pool_assets_from_para_to_para_through_asset_hub() {
```
acatangiu, line 1621: Very nice! 💪 

docs suggestion:
```suggestion
// ==========================================================================
// == Reserve Transfers USDT - Parachain->AssetHub->Parachain - pay fees with USDT (asset conversion) ==
// ==========================================================================
// Transfer USDT From Penpal A to Penpal B with AssetHub as the reserve, while paying fees using
// USDT by making use of existing USDT pools on AssetHub and destination.
#[test]
fn reserve_transfer_usdt_from_para_to_para_through_asset_hub() {
```

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: I originally made this test to make sure you could pay delivery fees locally with another asset, but now that's not the case. I left the test but put a TODO to make sure this is addressed in the future but I can just remove the tests.

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: I tried but it didn't work. Penpal's XCM config is set up to pay fees in the relay token (WND)

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: Sufficient means that it only covers ED but right now it also means we swap it. That's the idea that I had. In case we no longer need to support that, I can just drop all the support for paying fees with sufficient assets from this PR.

cumulus/parachains/runtimes/assets/common/src/lib.rs
```diff
@@ -132,6 +138,185 @@ pub type PoolAssetsConvertedConcreteId<PoolAssetsPalletLocation, Balance> =
 		TryConvertInto,
 	>;
 
+pub struct SufficientAssetConverter<Runtime, Matcher, BalanceConverter, AssetsInstance>(
+	PhantomData<(Runtime, Matcher, BalanceConverter, AssetsInstance)>,
+);
```
acatangiu, line 143: I don't think we need this

```suggestion
```

cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs
```diff
@@ -426,6 +426,7 @@ impl xcm_executor::Config for XcmConfig {
 	type HrmpChannelAcceptedHandler = ();
 	type HrmpChannelClosingHandler = ();
 	type XcmRecorder = PolkadotXcm;
+	type AssetConverter = ();
```
acatangiu, line 429: Rococo Asset Hub should support this too.

```suggestion
	type AssetConverter = PoolAssetsConverter;
```

cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -419,13 +455,19 @@ impl xcm_executor::Config for XcmConfig {
 			>,
 		>,
 	);
+	type AssetConverter = (
+		PoolAssetsConverter,
+		TrustBackedSufficientAssetsConverter,
+		ForeignSufficientAssetsConverter,
+	);
```
acatangiu, line 462: I don't believe we need the "sufficient" ones, everything should go through pool, no?

```suggestion
	type AssetConverter = PoolAssetsConverter;
```

cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs
```diff
@@ -362,6 +430,11 @@ impl xcm_executor::Config for XcmConfig {
 	type HrmpNewChannelOpenRequestHandler = ();
 	type HrmpChannelAcceptedHandler = ();
 	type HrmpChannelClosingHandler = ();
+	type AssetConverter = (
+		PoolAssetsConverter,
+		TrustBackedSufficientAssetsConverter,
+		ForeignSufficientAssetsConverter,
+	);
```
acatangiu, line 437: same here, only use `PoolAssetsConverter`

acatangiu (commented):
very nice! 🚀 

left some comments

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
acatangiu: ah, interesting.. that's unfortunate because it's not representative of ecosystem parachains

can you change that to use PENs in a separate PR (to not blow up this one)? we will need PEN/WND pool on Penpal for execution fees for existing tests, but afaik we don't need this PR too - it can come on top with the new scenarios where delivery fees are also paid using other stuff

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
acatangiu: yes, I would drop it completely

at most keep it on some archive branch if we ever find a usecase for it in the future, but based on our current knowledge pools everywhere is the plan :))

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: Sounds good, I thought about changing it here, but it changed a lot of tests 😅

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: That will greatly simplify this PR then, will remove

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +446,62 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+			"taking fees: {:?} from origin_ref: {:?} in fees_mode: {:?} for a reason: {:?}",
+			fees,
 			self.origin_ref(),
 			self.fees_mode,
 			reason,
 		);
+		// TODO: Is it okay to just use the first asset?
+		// Is there a downstream team that has their own router and they use
+		// more than one asset for delivery fees.
+		let asset_needed_for_fees = fees.get(0).ok_or(XcmError::AssetNotFound)?;
```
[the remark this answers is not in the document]
franciscoaguirre: I'll just add a comment in the PRDoc in case anyone actually has their own XcmRouter returning more than one asset for fees

cumulus/parachains/runtimes/assets/common/src/lib.rs
```diff
@@ -132,6 +138,185 @@ pub type PoolAssetsConvertedConcreteId<PoolAssetsPalletLocation, Balance> =
...
+
+/// Implementation of `AssetConverter` that pays delivery fees by swapping the given asset to the
+/// `Target` asset accepted for fees.
+pub struct SwapAssetConverter<Target, Runtime, Fungibles, Matcher, SwapCredit, AccountId>(
+	PhantomData<(Target, Runtime, Fungibles, Matcher, SwapCredit, AccountId)>,
+);
+impl<Target, Runtime, Fungibles, Matcher, SwapCredit, AccountId> AssetConversion
+	for SwapAssetConverter<Target, Runtime, Fungibles, Matcher, SwapCredit, AccountId>
+where
+	Target: Get<Fungibles::AssetId>,
```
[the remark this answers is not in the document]
franciscoaguirre: Don't we achieve this by having the pools and swapping to it? I don't see the practical use for this and it would complicate things quite a bit. Which asset should you swap to?

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
franciscoaguirre: I didn't see the "create-pool" helper fns?

cumulus/parachains/runtimes/assets/common/src/lib.rs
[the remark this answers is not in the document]
acatangiu: actually, we do we even need `Target` here? why limit these swaps to just one (or few) asset types?

It could always try to go from `give` to `want` and either succeed or not based on pool existence/liquidity.

cumulus/parachains/runtimes/assets/common/src/lib.rs
```diff
@@ -132,6 +138,104 @@ pub type PoolAssetsConvertedConcreteId<PoolAssetsPalletLocation, Balance> =
...
+				"Fee asset {:?} is not fungible",
+				asset,
+			);
+			return Err(XcmError::AssetNotFound);
+		};
+
+		let new_asset_amount =
+			pallet_asset_conversion::Pallet::<Runtime>::quote_price_tokens_for_exact_tokens(
+				fungibles_asset,
+				Target::get(),
```
acatangiu, line 188:
```suggestion
				asset.id,
```

cumulus/parachains/runtimes/assets/common/src/lib.rs
```diff
@@ -132,6 +138,104 @@ pub type PoolAssetsConvertedConcreteId<PoolAssetsPalletLocation, Balance> =
...
+			);
+			return Err(XcmError::AssetNotFound);
+		};
+
+		let swap_asset = fungibles_asset.clone().into();
+		let credit_in = Fungibles::issue(fungibles_asset, balance);
+
+		// Swap the user's asset for `asset`.
+		let (credit_out, credit_change) = SwapCredit::swap_tokens_for_exact_tokens(
+			vec![swap_asset, Target::get()],
```
acatangiu, line 224:
```suggestion
			vec![swap_asset, want.id.clone()],
```

cumulus/parachains/runtimes/assets/common/src/lib.rs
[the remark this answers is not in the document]
franciscoaguirre: We want it to swap for the actual asset used for fees. But I guess that could fall to the executor and not to this struct.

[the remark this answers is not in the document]
acatangiu: Yeah, the executor already interrogates the transport protocol about the fees and gets both required id and amount, so when it calls into this it already expects a specific asset back.

The `Target` only "helps" to hard-code the transport fee asset id at runtime config, but I think we can get rid of it.

cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs
```diff
@@ -419,13 +455,19 @@ impl xcm_executor::Config for XcmConfig {
...
 	type ResponseHandler = PolkadotXcm;
 	type AssetTrap = PolkadotXcm;
 	type AssetClaims = PolkadotXcm;
 	type SubscriptionService = PolkadotXcm;
 	type PalletInstancesInfo = AllPalletsWithSystem;
 	type MaxAssetsIntoHolding = MaxAssetsIntoHolding;
 	type AssetLocker = ();
+	// TODO: Might also use `AssetExchanger` instead of `AssetConverter` if API fits.
```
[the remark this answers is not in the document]
franciscoaguirre: Actually now that we're only doing swaps and nothing weird with sufficient assets, I'm looking into this again. I think it would be a win to not have to add one more config item.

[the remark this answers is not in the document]
franciscoaguirre: It does fit! 🎉 Updated PR to use the new config item

cumulus/parachains/runtimes/assets/common/src/lib.rs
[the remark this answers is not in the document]
franciscoaguirre: Removed `Target` in the latest version

cumulus/parachains/integration-tests/emulated/tests/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -1132,8 +1145,445 @@ fn reserve_transfer_native_asset_from_para_to_para_through_relay() {
...
+
+	let relay_asset_penpal_pov = Location::parent();
+	let custom_asset_penpal_pov = Location::new(
+		1,
+		[Parachain(1000), PalletInstance(ASSETS_PALLET_ID), GeneralIndex(asset_id.into())],
+	);
+	PenpalA::force_create_foreign_asset(
+		custom_asset_penpal_pov.clone(),
+		PenpalAssetOwner::get(),
+		false, // Asset not sufficient, has a pool to pay for fees.
```
acatangiu, line 1236: please stop associating "sufficiency" with fees paying ability - this is a source of confusion in the ecosystem which we too keep perpetuating.

`Sufficient` asset means it is allowed to cover account ED. Not sufficient means ED needs to be covered by some other asset.

If the asset it allowed to be used in fee paying is orthogonal to its sufficiency and depends solely on the chain's weight trader configuration.

```suggestion
		false,
```

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,146 @@
...
+			Credit = fungibles::Credit<AccountId, Fungibles>,
+		> + QuoteExchangePrice<Balance = u128, AssetKind = Fungibles::AssetId>,
+	Fungibles: fungibles::Balanced<AccountId, Balance = u128>,
+	Matcher: MatchesFungibles<Fungibles::AssetId, Fungibles::Balance>,
+{
+	fn exchange_asset(
+		_: Option<&Location>,
+		give: AssetsInHolding,
+		want: &Assets,
+		_: bool,
```
muharem, line 53: This implementation basically with `maximal = false`. It swaps exactly as set in `want`. Also this implementation care exactly about first asset in `give` and `want`.

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,146 @@
...
+use xcm_executor::{
+	traits::{AssetExchange, MatchesFungibles},
+	AssetsInHolding,
+};
+
+/// An adapter from [`pallet_asset_conversion::SwapCredit`] and
+/// [`pallet_asset_conversion::QuoteExchangePrice`] to [`xcm_executor::traits::AssetExchange`].
+///
+/// Takes just one fungible asset in `give` and allows only one fungible asset in `want`.
+pub struct FungiblesPoolAdapter<AssetConversion, Fungibles, Matcher, AccountId>(
```
muharem, line 34: May be something like `ExchangeFirstMinimal`

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,146 @@
...
+				credit_in,
+				want_amount,
+			)
+			.map_err(|(credit_in, _)| {
+				drop(credit_in);
+				give.clone()
+			})?;
+
+		log::trace!(target: "xcm", "Credit out: {:?}, Credit change: {:?}", credit_out.peek(), credit_change.peek());
+		debug_assert!(credit_change.peek() == Zero::zero());
```
muharem, line 108: not sure if we should require a client to know the exact `give` amount. this impl looks very specific to the case we wanna handle with this PR.

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,146 @@
...
+
+		let resulting_asset: Asset = (want_asset.id.clone(), credit_out.peek()).into();
```
muharem, line 110: probably no clone needed

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,146 @@
...
+					target: "xcm::FungiblesPoolAdapter::quote_exchange_price",
+					"Could not map XCM asset {:?} to FRAME asset. Error: {:?}.",
+					asset2,
+					error,
+				);
+				()
+			})
+			.ok()?;
+		let necessary_asset1_amount =
+			<AssetConversion as QuoteExchangePrice>::quote_price_tokens_for_exact_tokens(
```
muharem, line 138: returns option, no need for tmp var

polkadot/xcm/xcm-executor/src/config.rs
```diff
@@ -122,6 +122,7 @@ pub trait Config {
 	type HrmpChannelAcceptedHandler: HandleHrmpChannelAccepted;
 	/// Allows optional logic execution for the `HrmpChannelClosing` XCM notification.
 	type HrmpChannelClosingHandler: HandleHrmpChannelClosing;
+
```
muharem, line 125: this diff can be dropped

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +449,89 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 			self.origin_ref(),
 			self.fees_mode,
 			reason,
 		);
-		let paid = if self.fees_mode.jit_withdraw {
-			let origin = self.origin_ref().ok_or(XcmError::BadOrigin)?;
-			for asset in fee.inner() {
-				Config::AssetTransactor::withdraw_asset(&asset, origin, Some(&self.context))?;
+		// We only ever use the first asset from `fees`.
+		let asset_needed_for_fees = match fees.get(0) {
```
muharem, line 465: how can we make sure no one requires multiple assets as a fee payment? if anyone does, this silently will break things.

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +449,89 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+					target: "xcm::fees",
+					"Swap was deemed necessary but couldn't be done. Given assets: {:?}.",
+					given_assets,
+				);
+				XcmError::FeesNotMet
+			})?
+			.into();
+			swapped_asset
+		} else {
+			vec![asset_to_pay_for_fees].into()
```
muharem, line 534: I think this should be `assets_to_swap`. May be test?

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +449,89 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
-			for asset in fee.inner() {
-				Config::AssetTransactor::withdraw_asset(&asset, origin, Some(&self.context))?;
+		// We only ever use the first asset from `fees`.
+		let asset_needed_for_fees = match fees.get(0) {
+			Some(fee) => fee,
+			None => return Ok(()), // No delivery fees need to be paid.
+		};
+		// If `BuyExecution` was called, we know we can try to use that asset for fees.
+		let asset_to_pay_for_fees = if let Some(asset_used_for_fees) = &self.asset_used_for_fees {
+			if asset_used_for_fees != &asset_needed_for_fees.id {
```
muharem, line 471: could be `self.fee_asset` and `chain_fee_asset`

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -444,26 +449,89 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+			} else {
+				asset_needed_for_fees.clone()
 			}
-			fee
 		} else {
-			self.holding.try_take(fee.into()).map_err(|_| XcmError::NotHoldingFees)?.into()
+			asset_needed_for_fees.clone()
+		};
+		log::trace!(target: "xcm::fees", "Actual asset for fees: {:?}", asset_to_pay_for_fees);
+		let assets_to_swap = if self.fees_mode.jit_withdraw {
```
muharem, line 497: could be `withdrawn_fee_asset`, valid for both cases.

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -80,6 +80,10 @@ pub struct XcmExecutor<Config: config::Config> {
 	appendix_weight: Weight,
 	transact_status: MaybeErrorCode,
 	fees_mode: FeesMode,
+	/// Id of asset provided in `BuyExecution` instruction (if any) in current XCM program. Same
```
muharem, line 83: in last `BuyExecution`

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -866,8 +971,10 @@ impl<Config: config::Config> XcmExecutor<Config> {
 					let assets = Self::reanchored(deposited, &dest, None);
 					let mut message = vec![ReserveAssetDeposited(assets), ClearOrigin];
 					message.extend(xcm.0.into_iter());
-					// put back transport_fee in holding register to be charged by XcmSender
-					self.holding.subsume_assets(transport_fee);
+					// put back delivery_fee in holding register to be charged by XcmSender
```
muharem, line 974: it is important that there is no possibility to use a pool between quote and actual exchange, otherwise the price might change. if this is true now, I think it still better to try to keep those operations in the single context, otherwise it easy to break that requirements in the future.

polkadot/xcm/xcm-executor/src/lib.rs
```diff
@@ -852,11 +920,48 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
-
+					let maybe_delivery_fee = if let Some(asset_needed_for_fees) = fee.get(0) {
+						log::trace!(
+							target: "xcm::DepositReserveAsset",
+							"Asset provided to pay for fees {:?}, asset required for transport fees: {:?}",
+							self.asset_used_for_fees, asset_needed_for_fees,
+						);
+						let asset_to_pay_for_fees =
+							self.asset_used_for_fees.as_ref().unwrap_or(&asset_needed_for_fees.id);
+						let actual_asset_to_use_for_fees =
```
muharem, line 931: it would be nice if we can get rid of this repetition

muharem (commented):
Left some comments to consider. 
I think it would be helpful to list the use cases that this patch addresses as well as those it does not. I see the limitation section in the description, but examples for both will be helpful.

substrate/frame/asset-conversion/src/swap.rs
```diff
@@ -112,6 +112,30 @@ pub trait SwapCredit<AccountId> {
...
+
+	/// Provides the amount of `asset2` exchangeable for an exact amount of `asset1`.
+	fn quote_price_exact_tokens_for_tokens(
+		asset1: Self::AssetKind,
+		asset2: Self::AssetKind,
+		amount: Self::Balance,
+		include_fee: bool,
+	) -> Option<Self::Balance>;
+
+	/// Provides the amount of `asset1` exchangeable for an exact amount of `asset2`.
```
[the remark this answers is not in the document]
franciscoaguirre: Had to create that trait because I needed it for this. When you merge that PR I can use that trait instead.

polkadot/xcm/xcm-executor/src/lib.rs
acatangiu: This was discussed publicly on the [XCMv5 RFC targeting new way to specify fees](https://github.com/polkadot-fellows/xcm-format/pull/53).

Consensus there was that in practice no chain will ask fees of multiple asset types at once. It'd just be bad design. They might _support_ multiple assets for paying fees, in which case any one of the supported assets can be used to cover all required fees.

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: It's not going to be possible to use the pool between the quote and exchange since it's all part of the same instruction executions.
Users might be able to put their own `ExchangeAsset` instruction in but that'll happen afterwards or before, not in between.

polkadot/xcm/xcm-executor/src/lib.rs
franciscoaguirre: Once we get rid of jit withdrawal, we'll be able to just do the swap and be done with it.
Because we have jit withdrawal, we need to first quote to later withdraw the necessary assets for delivery fees.
By v5 we'll be able to remove this.

yrong:
@franciscoaguirre Cool, thanks for the effort!

One question for the use case transfering WETH from AH back to Ethereum(AH --> BH --> Ethereum), currently the delivery fee is configured on AH [here](https://github.com/paritytech/polkadot-sdk/blob/4bcdf8166972b63a70b618c90424b9f7d64b719b/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs#L651-L654) in native token(i.e. DOT).

So with this PR is that possible to config it in `WETH` while still allow end user to pay in DOT?

> We can only pay for delivery fees in different assets in intermediate hops

IIUC does your comment above means can't?

polkadot/xcm/xcm-builder/src/asset_exchange.rs
gui1117: I think for a correct implementation it should at least error when not all `want` are supplied:
```rust
	/// - `want`: The minimum amount of assets which should be given to the caller in case any
	///   exchange happens. If more assets are provided, then they should generally be of the same
	///   asset class if at all possible.
```

and also if `maximal` is true then at least try to convert every fungible from `give` to all of `want`:
```rust
	/// - `maximal`: If `true`, then as much as possible should be exchanged.
```

franciscoaguirre:
@yrong Given how the bridge works, BridgeHub doesn't charge the delivery fee, AssetHub does. This means it wouldn't work because of the limitation unfortunately

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,144 @@
...
+use xcm::prelude::*;
+use xcm_executor::{
+	traits::{AssetExchange, MatchesFungibles},
+	AssetsInHolding,
+};
+
+/// An adapter from [`pallet_asset_conversion::SwapCredit`] and
+/// [`pallet_asset_conversion::QuoteExchangePrice`] to [`xcm_executor::traits::AssetExchange`].
+///
+/// Takes just one fungible asset in `give` and allows only one fungible asset in `want`.
```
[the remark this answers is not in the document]
franciscoaguirre: The trait might allow multiple assets, but this particular type that implements that trait works in this way. Devs are free to use another type or build their own. I should make it clear that this is not a silver bullet though

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,144 @@
...
+				give.clone()
+			})?;
+
+		debug_assert!(credit_change.peek() == Zero::zero());
+
+		let resulting_asset: Asset = (want_asset.id.clone(), credit_out.peek()).into();
+		Ok(resulting_asset.into())
+	}
+
+	fn quote_exchange_price(asset1: &Asset, asset2: &Asset, _: bool) -> Option<u128> {
```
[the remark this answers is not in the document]
franciscoaguirre: I made it work only for `maximal=false`, I'll add the other case to make it a little bit more generic

polkadot/xcm/xcm-executor/src/traits/asset_exchange.rs
```diff
@@ -37,6 +37,13 @@ pub trait AssetExchange {
 		want: &Assets,
 		maximal: bool,
 	) -> Result<AssetsInHolding, AssetsInHolding>;
+
+	/// Handler for quoting the exchange price of two assets.
+	///
+	/// - `asset1` The first asset.
+	/// - `asset2` The second asset.
+	/// - `maximal`: If `true`, then all of `asset1` should be used.
+	fn quote_exchange_price(asset1: &Asset, asset2: &Asset, maximal: bool) -> Option<u128>;
```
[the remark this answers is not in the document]
franciscoaguirre: I don't think it makes sense to accept collections of assets here.

The API was meant to mimic the `QuotePrice` trait, but using `maximal: bool` to distinguish the two cases.

polkadot/xcm/xcm-builder/src/asset_exchange.rs
franciscoaguirre: Added the `maximal = true` case but yes, this implementation uses only the first asset, I'll change the name to reflect that.

franciscoaguirre: @acatangiu Any names come to mind?

polkadot/xcm/xcm-builder/src/asset_exchange.rs
franciscoaguirre: Now I use temp var because I decide on the function to call based on `maximal`

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
gui1117: Then we should change the specification of `AssetExchange`.

It is literally written: `want: The minimum amount of assets which should be given to the caller in case any exchange happens.`

Or we can add a new method in `AssetExchange` like `exchange_one_asset`

prdoc/pr_4375.prdoc
```diff
@@ -0,0 +1,45 @@
...
+    description: |
+      In order to support different assets for delivery fees, the existing AssetExchanger config item
+      was leveraged.
+      This config item allows exchanging assets and quoting exchange prices.
+      An adapter was added to `xcm-builder` to make it compatible with `pallet-asset-conversion`:
+      `FungiblesPoolAdapter`.
+      The new functionality in the executor pays delivery fees using only one asset, which might be
+      swapped to the correct one.
+      If you have a custom XcmRouter that returns more than one asset needed for delivery fees, then know
+      that only the first one will be used.
```
gui1117, line 26: I look at the doc, it seems ambiguous about what does `Assets` means when returned as `SendResult`.

Maybe it should be specified there that `Assets` as returned in `SendResult` means only the first asset will be used as payement.

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
gui1117: Also if `asset.len()` is 0 we can return OK

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,178 @@
...
+			);
+			give.clone()
+		})?;
+		let want_asset = want.get(0).ok_or_else(|| {
+			log::trace!(
+				target: "xcm::FungiblesPoolAdapter::exchange_asset",
+				"No asset was in `want`.",
+			);
+			give.clone()
+		})?;
```
gui1117, line 70: we can actually return `OK` if there is no assets in `want` no?

yrong:
> Given how the bridge works, BridgeHub doesn't charge the delivery fee, AssetHub does. This means it wouldn't work because of the limitation unfortunately

No worry. In our new design won't need a static delivery fee config anymore, as Adrian suggested we will use the low level `pallet-xcm::execute` for a custom xcm(e.g. including a [ExchangeAsset](https://github.com/Snowfork/polkadot-sdk/blob/f7054e73a0069bd59228a79b23c2cb21df6c4e71/polkadot/xcm/xcm-executor/src/lib.rs#L1184) for the internal swap).

polkadot/xcm/xcm-builder/src/asset_exchange.rs
gui1117: Yes it looks good to me, but I still think it should return an error if there is more than one asset in `want`

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
gui1117: Great, looks good to me

polkadot/xcm/xcm-builder/src/asset_exchange.rs
```diff
@@ -0,0 +1,178 @@
...
+					want_amount,
+				)
+				.map_err(|(credit_in, _)| {
+					drop(credit_in);
+					give.clone()
+				})?;
+
+			// TODO: If we want to make this a generic adapter, this need not be 0. Handle it.
+			// Probably depositing it back to the holding.
+			debug_assert!(credit_change.peek() == Zero::zero());
```
gui1117, line 126: I guess we can put the `credit_change`, and the `give` except first fungible asset, inside the result.

If performance is a downside for this, I can understand why you would implement only a very minimal version of the trait.

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
franciscoaguirre: You're right, I should error to make sure it doesn't fail silently

polkadot/xcm/xcm-builder/src/asset_exchange.rs
franciscoaguirre: Yeah given that we return `AssetsInHolding`, it makes sense to bundle them up together

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
franciscoaguirre: I have to do tests for this type, so I'll cover this case as well

polkadot/xcm/xcm-builder/src/asset_exchange.rs
franciscoaguirre: True

polkadot/xcm/xcm-builder/src/asset_exchange.rs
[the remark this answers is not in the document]
franciscoaguirre: Same with give

polkadot/xcm/xcm-executor/src/traits/asset_exchange.rs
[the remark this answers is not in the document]
gui1117: I also think the non-symmetricity with `exchange_asset` makes it difficult to use in more general cases.

Let's say you want to quote the price for 2 assets, from 2 other assets, to later call `exchange_asset` (a more general case than the current PR). Then there is no way to use this `quote_exchange_price` to get the expected requirement for `give` to get a successful execution of `exchange_asset`.

maybe something like this could be better
 ```rust
/// Returns a vector of minimal assets in holding which would execute
/// a successful `exchange_asset(_, _, give, want, _)`
fn exchange_asset_minimal_needed(give: &[AssetId], want: &Assets) -> Vec<AssetsInHolding>
```
Note that we could also support non-fungibles with a more complex `give` parameter maybe.
I'm not sure if we should also give the `Option<Location>` information

Then the tuple implementation would just be:
```rust
{
	for_tuples!( #(
		Tuple::exchange_asset_minimal_needed(give, want)
	)* )
		.flatten()
}
```

franciscoaguirre:
Closing in favor of https://github.com/paritytech/polkadot-sdk/pull/5130 and https://github.com/paritytech/polkadot-sdk/pull/5131

PR #4380, opened 2024-05-05
Remove `parameterized-consensus-hook` feature

### LINKED ISSUE #4366: Remove `parameterized-consensus-hook` feature
Feature should be removed as per comment from @bkchr.

Not sure whom to ping here, maybe @alexggh?

s0me0ne-unkn0wn:
Closes #4366


cumulus/pallets/parachain-system/src/lib.rs
```diff
@@ -1618,10 +1612,7 @@ impl<T: Config> polkadot_runtime_common::xcm_sender::EnsureForParachain for Pall
 }
 
 /// Something that can check the inherents of a block.
-#[cfg_attr(
-	feature = "parameterized-consensus-hook",
-	deprecated = "consider switching to `cumulus-pallet-parachain-system::ConsensusHook`"
-)]
+#[deprecated(note = "consider switching to `cumulus-pallet-parachain-system::ConsensusHook`")]
```
ggwpez, line 1615: When will it be finally removed? I think it should have a deadline mentioned.

bkchr:
> @bkchr I remember we discussed briefly about the `CheckInherents` and its deprecation, do you remember the outcome?

The `hook` gets the [`StateProof`](https://github.com/paritytech/polkadot-sdk/blob/16d368bbbf5fab0dac5704c7a388435df8a9db0b/cumulus/pallets/parachain-system/src/consensus_hook.rs#L79) passed. So, essentially the same as `CheckInherents`. Even if someone would have a custom implementation (which I doubt), they can still do the same checks as before. (Okay, a little bit harder if you need the transaction, but generally not impossible).

skunert:
> (Okay, a little bit harder if you need the transaction, but generally not impossible).

I mean that was the interesting part, I know that the state proof goes into the hook :+1:  . But I agree that there are probably no custom implementations.

bkchr (approved):

skunert (approved):

bkontur (approved):

PR #4392, opened 2024-05-06
Remove pallet::getter usage from the bounties and child-bounties pallets

PolkadotDom:
As per #3326, removes pallet::getter usage from the bounties and child-bounties pallets. The syntax `StorageItem::<T, I>::get()` should be used instead.

Changes to one pallet involved changes in the other, so I figured it'd be best to combine these two.

cc @muraca


bkchr (approved):

muharem (commented):
I would prefer in such case to provide a pub getters for storage items and have a minor bump for the crate. 
These pallets are old, if someone using those public getters to read data, this will be a breaking change, and it will take some time for maintainers to get those getters back. It wont be a nice experience. 
@bkchr @ggwpez what you think?

muharem:
Also storage items are part of public API of a pallet, even if there are no getter for them. So I believe additional getters do not add much to maintenance complexity.

PolkadotDom:
> I would prefer in such case to provide a pub getters for storage items and have a minor bump for the crate. These pallets are old, if someone using those public getters to read data, this will be a breaking change, and it will take some time for maintainers to get those getters back. It wont be a nice experience. @bkchr @ggwpez what you think?

Is there some way that introducing these changes breaks systems in production? I was thinking people would just need to update their codebase if they depended on the getters, but not familiar with every part of this.

bkchr:
> I was thinking people would just need to update their codebase

Yeah that is the case. 

Generally I don't know if anyone build something on top of these crates and used these getters, but also the release notes are quite clear. Still a better way would have been to first deprecate these getters and then to remove them.

PolkadotDom:
Happy to deprecate first moving forward if that is preferred. Just let me know :+1:

bkchr:
For me it is fine to remove them. This is not the first PR doing this, so it would be weird to stop now :see_no_evil:

muharem:
The storage items actually are public. They can be accessed from outside of the pallet without getter. Sorry for confusion.

muharem (approved):

PolkadotDom:
> The storage items actually are public. They can be accessed from outside of the pallet without getter. Sorry for confusion.

No worries! Thanks for reviewing :grin:

PR #4401, opened 2024-05-07
Update prdoc for 2226

tdimitrov:
Mention that offenders are no longer chilled and suggest node operators and nominators to monitor their nodes/nominees closely.

### DIFF
--- a/prdoc/pr_2226.prdoc
+++ b/prdoc/pr_2226.prdoc
@@ -4,7 +4,8 @@ doc:
   - audience: Node Operator
     description: |
       On each committed offence (no matter slashable or not) the offending validator will be
-      disabled for a whole era.
+      disabled for a whole era. Offenders are no longer chilled so the node operator should monitor
+      the behavior of their validators and act swiftly if they are raising disputes.
   - audience: Runtime Dev
     description: |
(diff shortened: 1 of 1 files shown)


ordian (approved):
Thanks!

prdoc/pr_2226.prdoc
```diff
@@ -14,7 +15,14 @@ doc:
       will be used on Kusama and Polkadot. In nutshell `UpToLimitDisablingStrategy`
       disables offenders up to the configured threshold. Offending validators are not disabled for
       offences in previous eras. The threshold is controlled via `DISABLING_LIMIT_FACTOR` (a generic
-      parameter of `UpToLimitDisablingStrategy`).
+      parameter of `UpToLimitDisablingStrategy`). Also offending validators are no longer chilled
+      after an offence is committed because a malicious validator can unchill itself and continue
+      committing offences.
```
Overkillus, line 20:
```suggestion
      parameter of `UpToLimitDisablingStrategy`). Also offending validators are no longer chilled
      after an offence is committed. This change is made to protect the network in case of honest nodes being slashed.
```

Overkillus (approved):
One nit around the motivation but all good otherwise, thanks!

PR #4405, opened 2024-05-07
CI short benchmarking with `frame-omni-bencher` 

bkontur:
## TODO

- [ ] setup and fix `get_preset` for SP testnets - https://github.com/paritytech/polkadot-sdk/pull/5327
  - [ ] BridgeHubRococo is not working: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6195778
  - [ ] collectives-westend: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6193730
  - [ ] contracts-rococo: https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/6193731
- [x] investigate for all `short-benchmark-*` replacing `./artifacts/polkadot` and `./artifacts/polkadot-parachain` with `./artifacts/frame-omni-bencher`
  - than we won't need `build-short-benchmark` and `build-short-benchmark-cumulus`, but we need to build all runtimes with `--features runtime-benchmarks` as for `quick-benchmarks-omni`
- [ ] remove from github:
  - `continuous-integration/gitlab-build-short-benchmark`
  - `continuous-integration/gitlab-build-short-benchmark-cumulus`

## Follow-up TODO:
- [ ] remove gitlab pipelines (where everything works)
- [ ] enable: https://github.com/paritytech/polkadot-sdk/pull/5268

## Possible follow-ups
- remove benchmarking code from `polkadot` / `polkadot-parachain` binaries
- simplify gitlab short-benchmarks with matrix-like syntax


.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -98,7 +110,7 @@ jobs:
           draft: true
 
   publish-runtimes:
-    needs: [ build-runtimes, publish-release-draft ]
+    needs: [ build-runtimes, build-frame-omni-bencher, publish-release-draft ]
```
kianenigma, line 113: @EgorPopelyaev is this the right approach to add the binary to the release artifacts? is there an internal guide/wiki around this somewhere? If this is broadly correct, I will replicate it in the chain-spec-builder PR.

bkontur:
@ggwpez 
Currently, we can use `frame-omni-bencher` either with pre-generated chain-spec with runtime code included or with pre-compiled runtime wasm blob which has `GenesisBuilder::get_preset` defined.

I am thinking about another possibility that could be maybe very useful,
what if we add support for `frame-omni-bencher` to run with both combined:
- pre-compiled runtime wasm blob which has `GenesisBuilder::get_preset`, which could generate default genesis spec
- and then merge/override default genesis spec with provided json chain-spec as argument (without runtime code attribute)

```
./target/release/frame-omni-bencher v1 benchmark pallet \ 
    --runtime ./target/release/wbuild/bridge-hub-rococo-runtime/bridge_hub_rococo_runtime.compact.compressed.wasm
    --chain some-chain-spec-with-overrides.json
    --all --steps 2 --repeat 1
```

This would add flexibility, e.g. when you want to tune benchmarking with different genesis, you just change json and not need to compile wasm again and so on.

I checked the code, and it should be that hard, maybe just to add new [`GenesisBuilder`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/utils/frame/benchmarking-cli/src/pallet/types.rs#L27-L37) type e.g. `RuntimeWithSpecOverride` or something like that.

.github/workflows/release-30_publish_release_draft.yml
EgorPopelyaev: Hey Kian, unfortunately there is no guide around it as so far we had only two binaries that we release from release team. As we discussed in the coordination chat. If we go the same way for the new bins (with build on the cleanroom and signatures with the SEC key, that we use for `polkadot` and` polkadot-parachain`), then shuld be done not here but in one of our scripts [like this one](https://github.com/paritytech/cleanroom/blob/master/ansible/roles/polkadot/files/scripts/polkadot-release-strategy). The main point of these process is so that people could verify that those binaries can be trusted. If it is an overengineering for `omni-bencher` and similar bins. Then they could be build and attached theway it is done here

ggwpez:
> This would add flexibility, e.g. when you want to tune benchmarking with different genesis, you just change json and not need to compile wasm again and so on.

Yea it could be useful. But should probably rename the argument from `--chain` to `--genesis-spec` since i want to deprecate that. Its just not a good name in general 🙈

.github/workflows/release-30_publish_release_draft.yml
kianenigma: I think omni-bencher and chain-spec-builder can be built without signing, so I will use this approach. 

Thank you for the explanation @EgorPopelyaev!

EgorPopelyaev: Oki, works for me

bkontur:
closing in favor of:
https://github.com/paritytech/polkadot-sdk/pull/5327
https://github.com/paritytech/polkadot-sdk/pull/5655
https://github.com/paritytech/polkadot-sdk/pull/5706

PR #4406, opened 2024-05-08
Adds benchmarking and try-runtime support in frame crate

gupnik:


liamaharon (approved):

kianenigma (approved):
It is good, but to be better an example create built with `frame` should actually have a benchmarking setup etc.

gupnik:
> It is good, but to be better an example create built with `frame` should actually have a benchmarking setup etc.

Yeah, integrating in the minimal fork for PBA currently, but will add an example as well as a follow-up here.

PR #4410, opened 2024-05-08
`sc-chain-spec`: deprecated code removed

michalkucharczyk:
This PR removes deprecated code:
- The `RuntimeGenesisConfig` generic type parameter in `GenericChainSpec` struct.
- `ChainSpec::from_genesis` method allowing to create chain-spec using closure providing runtime genesis struct 
- `GenesisSource::Factory` variant together with no longer needed `GenesisSource`'s generic parameter `G` (which was intended to be a runtime genesis struct).

https://github.com/paritytech/polkadot-sdk/blob/17b56fae2d976a3df87f34076875de8c26da0355/substrate/client/chain-spec/src/chain_spec.rs#L559-L563


cumulus/polkadot-parachain/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "polkadot-parachain-bin"
-version = "4.0.0"
+version = "5.0.0"
```
ggwpez, line 3: I think mentioning it in the prdoc is enough. These numbers here are then automatically bumped as aggregation of all the prdocs.

lexnv (approved):

bkchr (approved):

Issue #4174, opened 2024-04-17
State Trie Migration


ggwpez:
Need to track where we are and what is missing cc @cheme @PierreBesson.  
Three check marks: migration deployed, RPC reports done, migration removed from the runtime.

- [ ] Rococo [x], [x], [ ]
  - [ ] Asset-Hub [x], [x], [ ]
  - [x] ~~Bridge-Hub [ ], [ ], [ ], [ ]~~
  - [x] ~~Contracts [ ], [ ], [ ], [ ]~~
  - [x] ~~Coretime [ ], [ ], [ ], [ ]~~
  - [x] ~~People [ ], [ ], [ ], [ ]~~
- [ ] Westend [x], [x], [ ]
  - [ ] Asset-Hub [x], [ ], [ ]
  - [x] ~~Bridge-Hub [ ], [ ], [ ], [ ]~~
  - [ ] Collectives [x], [ ], [ ]
  - [x] ~~Coretime [ ], [ ], [ ], [ ]~~
  - [x] ~~Glutton [ ], [ ], [ ], [ ]~~ 
  - [x] ~~People [ ], [ ], [ ], [ ]~~

ggwpez:
@cheme i checked all the runtimes and i think i got all the missing ones listed here.  
Could you please add the migration to the missing ones? Seems like most new chains have V1 since genesis, so its not that many.

cheme:
Thanks, 
> Could you please add the migration to the missing ones?

Will do tomorrow (most likely will be the same change as the one you adapt in https://github.com/paritytech/polkadot-sdk/commit/cecb8f4e343d6fd3f9d2bde53969fff920cf0b33 , but I will reread the pallet logic first , just in case.

> Seems like most new chains have V1 since genesis, so its not that many.

yes new chain should have started v1 (I am a bit surprise by asset_hub (we migrate at least two statemine/statemint envt), but it can be done any number of time anyway).

Did quickly check and see same thing.
Note that the three first check marks make sense to me, but not the fourth one: the state_version is switch in the runtime at the same time as the runtime with the migration is deployed (the migration only works if the state_version is 1 in the runtime).

ggwpez:
> Did quickly check and see same thing.
> Note that the three first check marks make sense to me, but not the fourth one: the state_version is switch in the runtime at the same time as the runtime with the migration is deployed (the migration only works if the state_version is 1 in the runtime).

Okay, will remove 👍

ggwpez:
[...]

PR #4414, opened 2024-05-08
Rococo AH: undeploy trie migration

ggwpez:
The state-trie migration is completed on Rococo Asset-Hub as double-checked [here](https://github.com/paritytech/polkadot-sdk/issues/4174#issuecomment-2097895275).  
Undeploying now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -830,7 +830,6 @@ dependencies = [
  "pallet-nfts-runtime-api",
  "pallet-proxy",
  "pallet-session",
- "pallet-state-trie-migration",
  "pallet-timestamp",
  "pallet-transaction-payment",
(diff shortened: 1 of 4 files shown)


acatangiu (approved):

cheme (approved):
could use :
frame_support::migrations::RemovePallet<StateTrieMigrationName, RocksDbWeight>,
and
		pub const StateTrieMigrationName: &'static str = "StateTrieMigration";
as first step (as kusuma https://github.com/polkadot-fellows/runtimes/pull/170/files#diff-efa4caeb17487ecb13d8f5eb7863c3241d84afa2e73fbf25909a2ca89df0f362R1948), but it is just removing two key value from the state.

bkchr (approved):

liamaharon (approved):

PR #4417, opened 2024-05-08
Remove pallet::getter usage from pallet-contracts-mock-network

PolkadotDom:
A part of #3326 

Removes all #[pallet::getter] usage from the contracts mock network pallet. As the storage values were pub(super), read-only visibility was lost external to the crate upon the removal of the macros. I have implemented custom getters as a replacement, keeping the api the same.

If we care very much about consistency of the storagevalue::&lt;T&gt;::get() syntax, the other option would be to set the storage values to pub. Though I find preserving data authority better myself.

@muraca


bkchr (approved):

bkchr:
From the naming this is only a mock, this means internal code and we don't need a PRDOC. However, best to wait for @athei / @pgherveou to comment.

athei (approved):

athei:
Yes we don't need the PRDoc but it also doesn't hurt? The network is used to run tests which need more than one node.

bkchr:
> The network is used to run tests which need more than one node.

Not really sure what you mean by this.

athei:
We are using it for testing XCM.

PR #4418, opened 2024-05-08
Add READ_ONLY flag to contract call function 

smiasojed:
This PR implements the `READ_ONLY` flag to be used as a `Callflag` in the `call` function.
The flag indicates that the callee is restricted from modifying the state during call execution.
It is equivalent to Ethereum's [STATICCALL](https://eips.ethereum.org/EIPS/eip-214).


smiasojed:

smiasojed:

smiasojed:
bot cancel 5-a1be7227-1f82-4a82-9891-dfc753222de3

smiasojed:
@athei I assumed that `call_chain_extension` calls the API over the `Ext` trait, so read-only protection in the `Ext` implementation is good enough, please comment

substrate/frame/contracts/src/exec.rs
```diff
@@ -1001,11 +1004,13 @@ where
 					let contract = frame.contract_info.as_contract();
 					frame.nested_storage.enforce_subcall_limit(contract)?;
 
-					let caller = self.caller();
-					Contracts::<T>::deposit_event(
-						vec![T::Hashing::hash_of(&caller), T::Hashing::hash_of(&account_id)],
-						Event::Called { caller: caller.clone(), contract: account_id.clone() },
-					);
+					if !frame.read_only {
```
smiasojed, line 1007: Do we want to `deposit_event` for read_only call?

smiasojed:

substrate/frame/contracts/src/exec.rs
```diff
@@ -1252,6 +1280,11 @@ where
 		// Protection is on a per call basis.
 		self.top_frame_mut().allows_reentry = true;
 
+		if set_frame_read_only {
+			// Revert to the previous setting, if it was changed.
+			self.top_frame_mut().read_only = false;
+		}
```
athei, line 1286: I think setting the `read_only` in the parent frame makes it overly complex. We should just have `read_only` as an argument to `push_frame` and do the inheritance here.

Then we don't need to care about resetting it.

athei (commented):
Some further ideas:

1) Maybe move the checks from `Ext` to `Runtime`. I think it is easier to understand because the checks are per host function and that way we error out as early as possible. You could even implement this as attribute on the macro making it declarative. Having checks sprinkled around `Ext` seems not the best approach.

2) I think there is no difference in how `reentrancy` and `read_only` is inherited. Can you unify them in one bit field? You may need to invert `allow_reentrancy` to make this happen.

athei:
> @athei I assumed that `chain_extension` calls the API over the `Ext` trait, so read-only protection in the `Ext` implementation is good enough, please comment Another option is to extend `Ext` trait with `fn is_read_only(&self) -> bool` so the `chain_extension` is aware of the `read_only` context.

Yes it needs to be exposed on the `Ext` and chain extensions need to check it. With my comment above you need to expose it anyways if we move checks to the `Runtime`.

substrate/frame/contracts/src/exec.rs
athei: `deposit_event` should be forbidden in read only mode. At least this is how it is on Ethereum.

substrate/frame/contracts/src/exec.rs
smiasojed: done

smiasojed:
> 1. Maybe move the checks from `Ext` to `Runtime`. I think it is easier to understand because the checks are per host function and that way we error out as early as possible. You could even implement this as attribute on the macro making it declarative. Having checks sprinkled around `Ext` seems not the best approach.

Added `[mutable] attribute.

substrate/frame/contracts/src/exec.rs
```diff
@@ -3113,6 +3163,62 @@ mod tests {
...
+			// BOB calls CHARLIE with read-only flag and non zero value
+			assert_err!(
+				MockStack::run_call(
+					contract_origin,
+					BOB,
+					&mut GasMeter::<Test>::new(GAS_LIMIT),
+					&mut storage_meter,
+					&schedule,
+					0,
+					vec![1],
```
ascjones, line 3212: What's happening here? `value` is `0` but `input_data` is set?

substrate/frame/contracts/src/tests.rs
```diff
@@ -4249,3 +4249,113 @@ fn gas_consumed_is_linear_for_nested_calls() {
...
+		// Create both contracts: Constructors do nothing.
+		let addr_caller =
+			builder::bare_instantiate(Code::Upload(wasm_caller)).build_and_unwrap_account_id();
+		let addr_callee =
+			builder::bare_instantiate(Code::Upload(wasm_callee)).build_and_unwrap_account_id();
+
+		// Read-only call fails when modifying storage.
+		assert_err_ignore_postinfo!(
+			builder::call(addr_caller).data((&addr_callee, 100u32).encode()).build(),
+			<Error<Test>>::ContractTrapped
```
ascjones, line 4269: Why in this case is it `ContractTrapped` instead of `StateChangeDenied`?

ascjones (commented):
Just a couple of (possibly dumb) questions

substrate/frame/contracts/src/exec.rs
smiasojed: BOB uses the input provided here as a value when calling CHARLIE.

substrate/frame/contracts/src/exec.rs
```diff
@@ -988,24 +993,27 @@ where
...
-						vec![T::Hashing::hash_of(&caller), T::Hashing::hash_of(&account_id)],
-						Event::Called { caller: caller.clone(), contract: account_id.clone() },
-					);
+					if !self.is_read_only() {
+						let caller = self.caller();
+						Contracts::<T>::deposit_event(
+							vec![T::Hashing::hash_of(&caller), T::Hashing::hash_of(&account_id)],
+							Event::Called { caller: caller.clone(), contract: account_id.clone() },
+						);
+					}
```
athei, line 1016: Let's create a new event for static call instead of suppressing the event here. It's only custom events by the contract that we want to suppress.

substrate/frame/contracts/src/exec.rs
```diff
@@ -988,24 +993,27 @@ where
...
-						Event::DelegateCalled { contract: account_id.clone(), code_hash },
-					);
-				},
+				(ExportedFunction::Call, Some(code_hash)) =>
+					if !self.is_read_only() {
+						Contracts::<T>::deposit_event(
+							vec![T::Hashing::hash_of(account_id), T::Hashing::hash_of(&code_hash)],
+							Event::DelegateCalled { contract: account_id.clone(), code_hash },
+						);
+					},
```
athei, line 1002: Delegate call can't be read only. So this check is not necessary.

substrate/frame/contracts/src/exec.rs
```diff
@@ -1217,16 +1225,25 @@ where
...
 		allows_reentry: bool,
+		read_only: bool,
 	) -> Result<ExecReturnValue, ExecError> {
 		// Before pushing the new frame: Protect the caller contract against reentrancy attacks.
 		// It is important to do this before calling `allows_reentry` so that a direct recursion
 		// is caught by it.
 		self.top_frame_mut().allows_reentry = allows_reentry;
 
+		// Enable read-only access if requested; cannot disable it if already set.
+		let frame_read_only = read_only || self.top_frame().read_only;
```
athei, line 1236:
```suggestion
		let read_only = read_only || self.top_frame().read_only;
```

We can just shadow here to avoid the mental load of another binding.

substrate/frame/contracts/src/lib.rs
```diff
@@ -1202,6 +1202,9 @@ pub mod pallet {
 		/// into `pallet-contracts`. This would make the whole pallet reentrant with regard to
 		/// contract code execution which is not supported.
 		ReentranceDenied,
+		/// A contract attempted to invoke a call that is flagged as read-only, and inside that
+		/// call, a state-changing call happened.
```
athei, line 1206:
```suggestion
		/// A contract attempted to invoke a state modifying API while being in read-only mode.
```

substrate/frame/contracts/proc-macro/src/lib.rs
```diff
@@ -206,9 +207,25 @@ impl HostFn {
...
 		}
+
+		if mutable {
+			let stmt = syn::parse_quote! {
+				if ctx.ext().is_read_only() {
+					return Err(Error::<E::T>::StateChangeDenied.into());
+				}
+			};
+			item.block.stmts.insert(0, stmt);
+		}
```
athei, line 227: This is dangerous because we error out without charging any gas. Even when doing no work there is an overhead of calling a host function which needs to be charged.  We need to merge https://github.com/paritytech/polkadot-sdk/pull/4233 first. And then make sure that we do things in this order:

1. Charge the general host function overhead.
2. Do the check added in this PR.
3. Charge the host function specific gas.

substrate/frame/contracts/src/tests.rs
athei: Yes it should be `StateChangeDenied` when looking at the code.

substrate/frame/contracts/src/tests.rs
```diff
@@ -4249,3 +4249,113 @@ fn gas_consumed_is_linear_for_nested_calls() {
...
+
+		// Subsequent call input.
+		let input = (&addr_callee, pallet_contracts_uapi::CallFlags::empty().bits(), 0u64, 100u32);
+
+		// Read-only call fails when modifying storage.
+		assert_err_ignore_postinfo!(
+			builder::call(addr_caller)
+				.data((&addr_subsequent_caller, input).encode())
+				.build(),
+			<Error<Test>>::ContractTrapped
```
athei, line 4327: `StateChangeDenied`?

athei (changes requested):

substrate/frame/contracts/src/exec.rs
```diff
@@ -1217,16 +1225,25 @@ where
...
+
 		let try_call = || {
 			if !self.allows_reentry(&to) {
 				return Err(<Error<T>>::ReentranceDenied.into())
 			}
+
+			// If the call value is non-zero and state change is not allowed, issue an error.
```
pgherveou, line 1243: let's explain  why here, instead of translating the if statement

substrate/frame/contracts/proc-macro/src/lib.rs
pgherveou: 1. should be done in #4233
https://github.com/paritytech/polkadot-sdk/blob/pg/rework-host-benchs/substrate/frame/contracts/proc-macro/src/lib.rs?plain=1#L708-L711

> 3. Charge the host function specific gas.

What else should you charge if you exited early ?

substrate/frame/contracts/proc-macro/src/lib.rs
```diff
@@ -206,9 +207,25 @@ impl HostFn {
...
+					if mutable {
+						return Err(err(span, "#[mutable] can only be specified once"))
+					}
+					mutable = true;
+				},
 				_ => return Err(err(span, msg)),
 			}
 		}
+
+		if mutable {
```
athei, line 220: Seems inverted. Shouldn't it be caught by your tests.

```suggestion
		if !mutable {
```

substrate/frame/contracts/proc-macro/src/lib.rs
```diff
@@ -206,9 +207,25 @@ impl HostFn {
 					}
 					not_deprecated = false;
 				},
+				"mutable" => {
```
athei, line 210: Functions shout be `mutating` by default. It is safer to assume that and only flag functions where we know they are read only. So the attribute should be `read_only`. What do you think?

substrate/frame/contracts/proc-macro/src/lib.rs
athei: I don't understand your question regarding 3. But generally, we don't refund on error. But we can make an exception for the read_only check.

substrate/frame/contracts/src/exec.rs
smiasojed: I was checking it in the [evm](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/instructions.go#L722)
I have not seen there such restriction. It is implemented in similar way to `opCallCode`, which is an exception in [EIP-214](https://eips.ethereum.org/EIPS/eip-214): `As an exception, CALLCODE is not considered state-changing, even with a non-zero value.`. Please let me know WDYT

substrate/frame/contracts/proc-macro/src/lib.rs
smiasojed: I would need to add `#[read_only]` to `call_chain_extension` and `call` to avoid generating the `read-only` check. I believe this would be misleading, as these functions may or may not be `read-only`, depending on the implementation.

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -851,10 +851,11 @@ impl<'a, E: Ext + 'a> Runtime<'a, E> {
 					value,
 					input_data,
 					flags.contains(CallFlags::ALLOW_REENTRY),
+					flags.contains(CallFlags::READ_ONLY),
 				)
 			},
 			CallType::DelegateCall { code_hash_ptr } => {
-				if flags.contains(CallFlags::ALLOW_REENTRY) {
+				if flags.contains(CallFlags::ALLOW_REENTRY | CallFlags::READ_ONLY) {
```
athei, line 858: Why this change? `READ_ONLY` is allowed to bet set. This checks if both of them are set.

substrate/frame/contracts/src/exec.rs
```diff
@@ -988,24 +993,27 @@ where
...
-						Event::Called { caller: caller.clone(), contract: account_id.clone() },
-					);
+					if !self.is_read_only() {
+						let caller = self.caller();
+						Contracts::<T>::deposit_event(
+							vec![T::Hashing::hash_of(&caller), T::Hashing::hash_of(&account_id)],
+							Event::Called { caller: caller.clone(), contract: account_id.clone() },
+						);
+					}
 				},
```
athei, line 1017: We should still emit them I think. Before I was just talking about custom events by the contract which we should not allow. But these events should still be emitted.

substrate/frame/contracts/proc-macro/src/lib.rs
athei: That is a good point. Didn't think of that. Can you rename to `mutating`?  That sounds better to me as the function itself is not `mutable` like a variable. It `mutates`.

substrate/frame/contracts/src/wasm/runtime.rs
smiasojed: In Ethereum, we have `DelegateCall`, `Call`, or `StaticCall`. There is no combination where `DelegateCall` can change the context to `read-only`. Therefore, I decided to have `delegate_call`, `call (without read_only)`, and `call (with read_only)` as the analogy.

substrate/frame/contracts/src/tests.rs
smiasojed: We have a flow call_contract_A->A.call(contract B, read_only)->B.set_storage. We encountered a trap in B. Should the error be propagated?

substrate/frame/contracts/proc-macro/src/lib.rs
smiasojed: Sure, `mutating` is fine

substrate/frame/contracts/src/tests.rs
athei: Yes you are right. The calling contract does not really learn the reason why a callee had an error. The error gets thrown away. This is the case almost any error unfortunately. So yes this is correct here.

substrate/frame/contracts/src/wasm/runtime.rs
smiasojed: You are right, thank you

substrate/frame/contracts/proc-macro/src/lib.rs
pgherveou: I was just wondering what you meant by " charge the host function specific gas ", we would just exit early here so I don't see how you could charge anything else.

substrate/frame/contracts/src/exec.rs
smiasojed: OK, done

substrate/frame/contracts/proc-macro/src/lib.rs
smiasojed: done

substrate/frame/contracts/src/exec.rs
smiasojed: Ideally, it would be placed in `mod env`, but I would need to move most of the `Runtime::call` logic there. I already had another `read-only` checks in Stack (exec) for `call`, so I decided to keep all of them in one place.
I think now that we could move this logic to `Runtime::call` as checks for events disappeared from Stack.
WDYT?

substrate/frame/contracts/src/exec.rs
athei: Sorry I misunderstood. The event should still be emitted.

substrate/frame/contracts/proc-macro/src/lib.rs
athei: Point 1. + 2. will be done in the macro. Point 3.  has to be done manually as we already do.

substrate/frame/contracts/src/exec.rs
athei: I think it should move to `mod env` so we have all the read only checks in one file. 

> but I would need to move most of the Runtime::call logic there.

I don't get this point. Can't you just move the check here and be done with it:
https://github.com/paritytech/polkadot-sdk/blob/4adfa37d14c0d81f09071687afb270ecdd5c2076/substrate/frame/contracts/src/wasm/runtime.rs#L844-L846

smiasojed: I meant to move it to the `Runtime::call` function as you pointed out.

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -839,9 +839,15 @@ impl<'a, E: Ext + 'a> Runtime<'a, E> {
 				} else {
 					self.read_sandbox_memory_as(memory, deposit_ptr)?
 				};
+				let read_only = flags.contains(CallFlags::READ_ONLY);
```
athei, line 842: If you do that here you don't have to repeat the logic twice.

```suggestion
				let read_only = flags.contains(CallFlags::READ_ONLY) || self.ext.is_read_only();
```

smiasojed: I understand, but here are my two arguments: 
1. If `ext.call()` in `chain_extension` is used, the developer will need to understand these requirements and apply this condition to the call, otherwise he will corrupt our stack.
2. This is a call that can be either readonly (staticcall) or a standard call. I think that it is the stack's responsibility to put the frame with the correct permissions.
WDYT?

athei: You are right. This is a good point. Let's keep it like this then.

xermicus (commented):
Another detail if we want to mimic EVM `STATICCALL` semantics in full: `STATICCALL`s to accounts without code (e.g. call with the static flag set and no value) should succeed.

[Note that an account with no code will return success as true (1).](https://www.evm.codes/#fa?fork=cancun)

athei:
This is a general difference we have to Ethereum that is independent of `STATICCCAL`. We don't allow balance transfers to plain account via `seal_call`. We need to add this behaviour in order to support revive. But not on this PR and probably only after the fork since it is not needed for ink!.

pgherveou:
@smiasojed sorry for the new merge conflicts 😅
Let me know when this is ready for a final review

smiasojed:
> @smiasojed sorry for the new merge conflicts 😅 Let me know when this is ready for a final review

@pgherveou conflicts resolved

pgherveou (approved):
lgtm

smiasojed:

substrate/frame/contracts/proc-macro/src/lib.rs
```diff
@@ -217,6 +224,16 @@ impl HostFn {
...
 		}
+
+		if mutating {
+			let stmt = syn::parse_quote! {
+				if ctx.ext().is_read_only() {
+					return Err(Error::<E::T>::StateChangeDenied.into());
+				}
+			};
+			item.block.stmts.insert(0, stmt);
+		}
```
athei, line 235: As stated before: We can't return that early here. This codes runs before we charge the gas for the host function call invocation itself. This has to be charged even if no worked has to be done. You have to move down this statement after:
https://github.com/paritytech/polkadot-sdk/blob/89604daa0f4244bc83782bd489918cfecb81a7d0/substrate/frame/contracts/proc-macro/src/lib.rs#L708-L711

athei: Ahh yes. Can you add a test that makes sure this weight is charged when erroring out early?

smiasojed: I tried to test it using fixture:
```
pub extern "C" fn call() {
	input!(n: u32, );

	match n {
		0 => unsafe {core::ptr::read_volatile(core::ptr::null())},
		_ => api::set_storage(&[1u8; 32], &[1u8; 1]),
	}
}
```
The fixture is called with n equal to 0 and 1. Both calls result in a trap, but the difference is more significant than just the host function call cost because the error handling differs for both cases (WASMI execution time). The difference is the host function call cost plus 10%.

athei: You can call the `tokens` function on the gas meter to check which tokens were charged.

smiasojed: As discussed, we do not have access to gas_meter in our tests. We could add such functionality to our tests to trace gas metering more deeply during testing. This could be implemented as a separate PR.

athei (commented):
You have to bump the `ApiVersion` and then we are good to go.

athei (approved):

Issue #4351, opened 2024-05-02
Coretime auto renew


bkchr:
Currently parachains that want to renew their slot need to do this every sale in the first week. They can also do this afterwards, but then all cores could already have been sold. It would be nice to support an auto renew. Parachains could opt-in to this auto renew and then only would need to ensure that the account for doing the auto renew has enough funds.

Basically the following transaction should be added:

```
fn enable_auto_renew(core, target_account)
```

First time this is called, the core should be renewed automatically (maybe it was already and we need to check "the future") and then the core & account is to some list. When rotating the sales this list should be iterated and all cores should be automatically renewed.

bkchr:
CC @seadanda

itsbirdo:
Would be a welcome improvement based on feedback from several teams

Szegoo:
Would like to work on this over the weekend if available.

Szegoo:
> fn enable_auto_renew(core, target_account)

This should be a two step process, otherwise a parachain could set any account as target without requiring the `target_account`'s approval for this. Another option would be, assuming this is called with the parachain's sovereing account is to set the target to the sovereign account. 

@bkchr did you have something else in mind maybe?

bkchr:
> @bkchr did you have something else in mind maybe?

No. Just didn't thought enough :D Probably the best is to not have `target_account` at all. Just take the account that signed the transaction as payment account.

Szegoo:
> Just take the account that signed the transaction as payment account.

But this can be just any account? For example, I could set myself as the payment account, drain my account, and then upon the next renewal, it would fail. 

Given that the broker pallet is quite abstract, it doesn't assume that cores are assigned to parachains. But would it be acceptable to assume each 'task' has an account associated with it? (basically each parachain has its sovereign account)

I would probably add a new type to the broker pallet config which would implement conversion from `TaskId` to `AccountId`. We could reuse the `LocationToAccountId` implementation such that we convert `Parachain(TaskId)`. And then this account would be used for payment.

PR #4424, opened 2024-05-09
Coretime auto-renew

Szegoo:
This PR adds functionality that allows tasks to enable auto-renewal. Each task eligible for renewal can enable auto-renewal.

A new storage value is added to track all the cores with auto-renewal enabled and the associated task running on the core. The `BoundedVec` is sorted by `CoreIndex` to make disabling auto-renewal more efficient.

Cores are renewed at the start of a new bulk sale. If auto-renewal fails(e.g. due to the sovereign account of the task not holding sufficient balance), an event will be emitted, and the renewal will continue for the other cores.

The two added extrinsics are:
- `enable_auto_renew`: Extrinsic for enabling auto renewal.
- `disable_auto_renew`: Extrinsic for disabling auto renewal.

TODOs:
- [x] Write benchmarks for the newly added extrinsics.

Closes: #4351


substrate/frame/broker/src/lib.rs
```diff
@@ -172,6 +181,11 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type CoreCountInbox<T> = StorageValue<_, CoreIndex, OptionQuery>;
 
+	/// Keeping track of cores which have auto-renewal enabled.
+	#[pallet::storage]
```
eskimor, line 185: You have an invariant that this list is supposed to be sorted. This should the very least be documented here.

eskimor (commented):
If it is the sovereign account of the para, why don't we implement auto renew on the para itself? Major reason, I guess: It would be impossible to mess up the timing. Otherwise each para would need to keep track of the sale cycle on Coretime.

Szegoo:
> If it is the sovereign account of the para, why don't we implement auto renew on the para itself?

But what if the parachain has multiple cores assigned to it? For example, perhaps the parachain wants to keep renewing only one core but has acquired an additional core for this period due to higher demand.

The reason I am storing the `TaskId` here as well is to avoid having to look it up for each core when performing auto-renewals.

substrate/frame/broker/src/lib.rs
Szegoo: Yes, keep in mind that this PR is still very much WIP.

Szegoo:
@eskimor I would appreciate a review to see if what I implemented here is sensible before writing benchmarks.

Polkadot-Forum:

https://forum.polkadot.network/t/kusama-coretime-renewal/8219/1

substrate/frame/broker/src/tick_impls.rs
```diff
@@ -329,4 +332,18 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Renews all the cores which have auto-renewal enabled.
+	pub(crate) fn renew_cores() {
+		let renewals = AutoRenewals::<T>::get();
+		for (core, task) in renewals.into_iter() {
+			let Some(payer) = T::SovereignAccountOf::sovereign_account(task) else {
+				Self::deposit_event(Event::<T>::AutoRenewalFailed { core, payer: None });
+				continue;
+			};
+			if let Err(_) = Self::do_renew(payer.clone(), core) {
```
seadanda, line 344: The core that a given task is on changes from sale to sale in general, so this will not work as implemented.
There are a few things that can cause this off the top of my head:
- a new reservation (the n system workloads are assigned a core first, so are always the first n cores
- a new lease
- a core on a lower core index which does not renew

Szegoo: Good point. The `do_renew` function returns the new `CoreIndex`, so this could work correctly if we update the auto-renewal vector with the new core indices upon the end of core renewals.

seadanda: Sounds good
I think an expansion of the tests you're adding to cover some more complex sequences of actions would also be great, i.e. including reservations and leases in the tests, one of which ends or is added.
I have it on my TODO list to expand the rest of the test cases in the broker, because they are in general quite gappy or in some cases overly simple, so any additional test cases for the feature you're adding will be very welcome!

poppyseedDev:
Could you please explain why this PR wasn't addressed with the comment, "This is out of scope for the broker pallet," similar to how it was done in this PR: [link to PR](https://github.com/paritytech/polkadot-sdk/pull/3693)?

seadanda:
I think this PR is totally in scope for the broker pallet, renewals are a first-class operation supported directly on the coretime chain and this is a UX improvement for teams who want to renew regularly.
The PR you've linked introduces secondary market functionality directly in the broker pallet, which is not in scope for the coretime chain as defined in RFC-1. It's not that we're making arbitrary decisions about scope, if an RFC were to be accepted that defines the functionality in that PR then it would be a welcome addition

substrate/frame/broker/src/coretime_interface.rs
```diff
@@ -146,3 +146,25 @@ impl CoretimeInterface for () {
...
+	/// The custom task origin. Given that all tasks on Polkadot are parachains this will most
+	/// likely be the `Parachain` origin.
+	type TaskOrigin;
+
+	/// Ensures that the origin is a task origin and returns the associated `TaskId`.
+	fn ensure_task_sovereign_account(o: Self::OuterOrigin) -> Result<TaskId, BadOrigin>;
+
+	/// Returns the associated account of a task.
+	fn sovereign_account(task: TaskId) -> Option<Self::AccountId>;
+}
```
Szegoo, line 170: I just realized we don't need this at all. Sovereign accounts are represented as signed origin: https://github.com/paritytech/polkadot-sdk/blob/d539778c3cc4e0376769472fdad37856f4051dc5/polkadot/xcm/xcm-builder/src/origin_conversion.rs#L27-L29

This will simplify a lot of the code.

seadanda: OK cool, I'll abandon my review for now then - I'm not all the way through but just have some comments

substrate/frame/broker/src/migration.rs
```diff
@@ -85,7 +85,7 @@ mod v2 {
 	};
 
 	#[storage_alias]
-	pub type AllowedRenewals<T: Config> = StorageMap<
+	pub type PotentialRenewals<T: Config> = StorageMap<
```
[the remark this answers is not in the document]
Szegoo: Yes, these were accidental.

Polkadot-Forum:

https://forum.polkadot.network/t/kusama-coretime-renewal/8219/12

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -476,6 +476,69 @@ impl<T: Config> Pallet<T> {
...
+				Error::<T>::NotAllowed
+			);
+		} else {
+			return Err(Error::<T>::NotAllowed.into())
+		}
+
+		// We are keeping the auto-renewals sorted by `CoreIndex`.
+		AutoRenewals::<T>::try_mutate(|renewals| {
+			let pos = renewals
+				.binary_search_by(|r: &AutoRenewalRecord| r.core.cmp(&core))
```
seadanda, line 505: This goes as logN, which affects your benchmarking worst case

substrate/frame/broker/src/dispatchable_impls.rs
```diff
@@ -476,6 +476,69 @@ impl<T: Config> Pallet<T> {
...
+		.map_err(|_| Error::<T>::TooManyAutoRenewals)?;
+
+		Self::deposit_event(Event::AutoRenewalEnabled { core, task });
+		Ok(())
+	}
+
+	pub(crate) fn do_disable_auto_renew(core: CoreIndex, task: TaskId) -> DispatchResult {
+		AutoRenewals::<T>::try_mutate(|renewals| -> DispatchResult {
+			let pos = renewals
+				.binary_search_by(|r: &AutoRenewalRecord| r.core.cmp(&core))
```
seadanda, line 525: Likewise

substrate/frame/broker/src/benchmarking.rs
```diff
@@ -939,6 +991,73 @@ mod benches {
...
+
+		let region = Broker::<T>::do_purchase(caller.clone(), 10u32.into())
+			.map_err(|_| BenchmarkError::Weightless)?;
+
+		Broker::<T>::do_assign(region, None, 2001, Final)
+			.map_err(|_| BenchmarkError::Weightless)?;
+		// advance to next bulk sale:
+		advance_to::<T>(6);
+
+		// The most 'intensive' path is when we renew the core upon enabling auto-renewal.
```
seadanda, line 1015: With max auto renewals. You've done it for rotate_sale, but not for enable/disable autorenew extrinsics

seadanda: Otherwise, the PR looks complete. nice

substrate/frame/broker/src/dispatchable_impls.rs
Szegoo: Fixed in the new commits.

franciscoaguirre (approved):
Looks good to me. @seadanda ?

seadanda (approved):
Nice, just did the last pass over your changes since my last review. Thanks!

substrate/frame/broker/src/lib.rs
kianenigma: PR is merged without this being really addressed, should be a 1 liner `fn try_state`

substrate/frame/broker/src/lib.rs
```diff
@@ -176,6 +183,13 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type CoreCountInbox<T> = StorageValue<_, CoreIndex, OptionQuery>;
 
+	/// Keeping track of cores which have auto-renewal enabled.
+	///
+	/// Sorted by `CoreIndex` to make the removal of cores from auto-renewal more efficient.
+	#[pallet::storage]
+	pub type AutoRenewals<T: Config> =
+		StorageValue<_, BoundedVec<AutoRenewalRecord, T::MaxAutoRenewals>, ValueQuery>;
```
kianenigma, line 191: Sorry for being a doomer, but isn't this super not scalable? As in, up to N renewals can happen at most, and all of it is `on_init`, in a single block?  

Moreover, if there is congestion (e.g. more than `MaxAutoRenewals` cores that wanna auto-renew), how do we deal with congestion? The right way to solve this kind of issue is to let user compete via more tx-fees. 

I suppose this is still assuming we will only have a maximum of a few dozen cores that want to auto-renew, which is fine, but I think we could have done a bit better.

Szegoo: You are right; there is an assumption that there is a fairly limited number of cores and also a limited amount that will own an entire core and do auto-renewals. However, what you mentioned is valid.

I am not really sure what you meant by 'let users compete via more tx-fees.'

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-coretime-renewals/10536/1

PR #4430, opened 2024-05-10
Add generate and verify logic for `AncestryProof`

serban300:
Extracting the logic for generating and verifying ancestry proofs from this PR: https://github.com/paritytech/polkadot-sdk/pull/1903 with small adjustments

@Lederstrumpf I added you as a co-author to each commit. Please let me know if you want me to also associate an e-mail address with your name

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2604,6 +2604,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ckb-merkle-mountain-range"
+version = "0.6.0"
+source = "git+https://github.com/paritytech/merkle-mountain-range.git?branch=master#537f0e3f67c5adf7afff0800bbb81f02f17570a1"
+dependencies = [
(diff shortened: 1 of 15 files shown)


acatangiu (approved):

svyatonik (approved):

ggwpez:
Not `R0-silent` either. It breaks peoples runtimes when they try to update (like Polkadot and Kusama).

Im just going to assume that i can replace `Proof` with `LeafProof` as there is no info on what to do.

serban300:
> Not `R0-silent` either. It breaks peoples runtimes when they try to update (like Polkadot and Kusama).

Sorry, I didn't realize. I thought that non-functional changes are `R0-silent`

> Im just going to assume that i can replace `Proof` with `LeafProof` as there is no info on what to do.

Yes, that's all.

PR #4459, opened 2024-05-14
Disable purchasing credit

### LINKED ISSUE #4454: Disable credits on coretimechain for now
Until we have: https://github.com/paritytech/polkadot-sdk/issues/2210

s0me0ne-unkn0wn:
Closes #4454

### DIFF
--- a/substrate/frame/broker/src/benchmarking.rs
+++ b/substrate/frame/broker/src/benchmarking.rs
@@ -525,38 +525,38 @@ mod benches {
 		Ok(())
 	}
 
-	#[benchmark]
-	fn purchase_credit() -> Result<(), BenchmarkError> {
-		setup_and_start_sale::<T>()?;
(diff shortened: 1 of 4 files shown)


seadanda (changes requested):
I don't agree with the approach here - I think we should disallow the calls using the `BaseCallFilter` in the runtimes, rather than commenting stuff out in the broker pallet.

This way could lead to version issues later - we will want to enable this at different times for each network, but potentially still want features/fixes that are added at later points. This could mean that we would need to maintain two versions of the pallet - one with purchase credit active, and one without.

bkchr:
> I don't agree with the approach here - I think we should disallow the calls using the `BaseCallFilter` in the runtimes, rather than commenting stuff out in the broker pallet.

Yes, this 10000%. Also we should just get implement the credit system, as it is not that complicated.

PR #4465, opened 2024-05-15
Bridge: added force_set_pallet_state call to pallet-bridge-grandpa

svyatonik:
closes https://github.com/paritytech/parity-bridges-common/issues/2963

See issue above for rationale
I've been thinking about adding similar calls to other pallets, but:
- for parachains pallet I haven't been able to think of a case when we will need that given how long referendum takes. I.e. if storage proof format changes and we want to unstuck the bridge, it'll take a large a few weeks to sync a single parachain header, then another weeks for another and etc.
- for messages pallet I've made the similar call initially, but it just changes a storage key (`OutboundLanes` and/or `InboundLanes`), so there's no any logic here and it may be simply done using `system.set_storage`.


acatangiu (approved):

svyatonik:
@bot help

svyatonik:
bot help

svyatonik:
bot bench cumulus-bridge-hubs --pallet=pallet_bridge_grandpa

svyatonik:
bot cancel

svyatonik:
bot bench cumulus-bridge-hubs --pallet=pallet_bridge_grandpa

serban300 (approved):

svyatonik:
bot bench cumulus-bridge-hubs --runtime=bridge-hub-westend --pallet=pallet_bridge_grandpa

Issue #4167, opened 2024-04-17
Local Collator Panic - Connecting to Rococo with Validator RPC


SBalaguer:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

I'm testing out registering a Parachain on Rococo and getting some blocks validated using onDemandAssingmentProvider. The objective is to reproduce what a new builder could potentially do when registering and testing the system for the first time.

In order to achieve this, I'm running a collator leveraging an RPC connection to Rococo instead of running a Rococo validator directly within the collator. I'm doing this by passing the flag `--relay-chain-rpc-url "wss://rococo-rpc.polkadot.io"` to my collator.

At the beginning everything seems to work fine, and I even manage to produce blocks on demand with my parachain, however it gets to a moment in time where the collator panics. The message looks like this (full logs from`alice` and `bob` attached):

```
2024-04-17 10:52:22.165 ERROR tokio-runtime-worker overseer: [Relaychain] Overseer exited with error err=Generated(SubsystemStalled("collator-protocol-subsystem", "signal", "polkadot_node_subsystem_types::OverseerSignal"))
2024-04-17 10:52:22.166 ERROR tokio-runtime-worker sc_service::task_manager: [Relaychain] Essential task `overseer` failed. Shutting down service.
2024-04-17 10:52:22.165 ERROR tokio-runtime-worker polkadot_overseer: [Relaychain] subsystem exited with error subsystem="availability-recovery" err=FromOrigin { origin: "availability-recovery", source: Generated(Context("Signal channel is terminated and empty.")) }
2024-04-17 10:52:22.166 ERROR tokio-runtime-worker polkadot_overseer: [Relaychain] subsystem exited with error subsystem="network-bridge-rx" err=FromOrigin { origin: "network-bridge", source: SubsystemError(Generated(Context("Signal channel is terminated and empty."))) }
2024-04-17 10:52:22.166 ERROR tokio-runtime-worker parachain::collation-generation: [Relaychain] error receiving message from subsystem context: Generated(Context("Signal channel is terminated and empty.")) err=Generated(Context("Signal channel is terminated and empty."))
[...]

Issue #4429, opened 2024-05-10
Remove prospective-parachains subsystem from collator nodes


alindima:
This is not really needed and brings a serious performance hit for collators which connect to RPC nodes.

See this discussion: https://github.com/paritytech/polkadot-sdk/issues/4167#issuecomment-2091203820

PR #4471, opened 2024-05-15
Remove the prospective-parachains subsystem from collators

alindima:
Implements https://github.com/paritytech/polkadot-sdk/issues/4429

Collators only need to maintain the implicit view for the paraid they are collating on.
In this case, bypass prospective-parachains entirely. It's still useful to use the GetMinimumRelayParents message from prospective-parachains for validators, because the data is already present there.

This enables us to entirely remove the subsystem from collators, which consumed resources needlessly

Aims to resolve https://github.com/paritytech/polkadot-sdk/issues/4167 

TODO:
- [ ] fix unit tests

### DIFF
--- a/polkadot/node/subsystem-util/src/backing_implicit_view.rs
+++ b/polkadot/node/subsystem-util/src/backing_implicit_view.rs
@@ -17,12 +17,17 @@
 use futures::channel::oneshot;
 use polkadot_node_subsystem::{
 	errors::ChainApiError,
-	messages::{ChainApiMessage, ProspectiveParachainsMessage},
+	messages::{ChainApiMessage, ProspectiveParachainsMessage, RuntimeApiMessage},
 	SubsystemSender,
 };
 use polkadot_primitives::{BlockNumber, Hash, Id as ParaId};
 
 use std::collections::HashMap;
+
+use crate::{
+	request_session_index_for_child,
+	runtime::{self, prospective_parachains_mode, recv_runtime, ProspectiveParachainsMode},
+};
 
 // Always aim to retain 1 block before the active leaves.
 const MINIMUM_RETAIN_LENGTH: BlockNumber = 2;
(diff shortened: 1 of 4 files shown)


polkadot/node/subsystem-util/src/backing_implicit_view.rs
```diff
@@ -265,30 +291,91 @@ pub enum BlockHeaderUnavailableReason {
...
+) -> Result<Vec<(ParaId, BlockNumber)>, FetchError> {
+	let (tx, rx) = oneshot::channel();
+	sender
+		.send_message(ProspectiveParachainsMessage::GetMinimumRelayParents(leaf_hash, tx))
+		.await;
+
+	rx.await.map_err(|_| FetchError::ProspectiveParachainsUnavailable)
+}
+
+// Request the min relay parent for a single para, directly using ChainApi.
```
sandreim, line 311: This seem to return the `leaf_number` - `allowed_ancestry_len` no matter what `para_id`. Don't we have to constrain this to be the minimum between this value and the best backable parachain block relay parent ?

sandreim (commented):
It looks like we only use this for sanity checking before advertising collations, I would propose to remove this check and rely on validators rejecting it and disconnecting the collator.

There are two situations when this would happen:
- somebody writes a bad collator that authors on ancient relay parents which is a bug, and having a strong signal as getting disconnected/banned is better than relying on this sanity check as the effect is the same - parachain doesn't make progress at all
- the collator is so far behind in syncing the relay chain such that this check would not actually fail at all, and it will still be the validator rejecting the collation

skunert:
I reran the original experiments #4167 and this PR indeed fixes the issue!

alindima:
> It looks like we only use the min relay parent for sanity checking before advertising collations, I would propose to remove this check instead and rely on validators rejecting it and disconnecting the collator.
> 
> There are two situations when this would happen:
> 
> * somebody writes a bad collator that authors on ancient relay parents which is a bug, and having a strong signal as getting disconnected/banned is better than relying on this sanity check as the effect is the same - parachain doesn't make progress at all
> * the collator is so far behind in syncing the relay chain such that this check would not actually fail at all, and it will still be the validator rejecting the collation

As discussed offline, we need to keep the implicit view on the collator side regardless of whether we query it before advertising. We still need to query the allowed ancestry so that collators are able to build and advertise collations that are built on older relay parents and for pruning these collations when the relay parents go out of scope.

So we could at most remove one line of code, the one that re-checks the ancestry before advertising. But I don't think it's worth it. It's a sanity check that'll save us from advertising outdated collations

polkadot/node/subsystem-util/src/backing_implicit_view.rs
alindima: This would complicate the code for collators, since they'll no longer have the prospective-parachains subsystem to track the backed candidates. Also, they don't receive backing statements. I guess we could use the persisted_validation_data runtime API which would return the relay chain block height of the last backed or included candidate.

I wouldn't bother with this for now, as it preserves the existing behaviour (which doesn't seem to cause issues) and the collator should anyway be smart enough to not produce collations that move the relay parents backwards.

alindima: the name of the function though is not really correct

alindima: changed the function name

sandreim (approved):

alexggh (approved):

PR #4481, opened 2024-05-16
Bridge: drop subscriptions when they are no longer required

svyatonik:
The bridge relay is **not** using `tokio`, while `jsonrpsee` does. To make it work together, we are spawning a separate tokio task for every jsonrpsee subscription, which holds a subscription reference. It looks like we are not stopping those tasks when we no longer need it and when there are more than `1024` active subscriptions, `jsonrpsee` stops opening new subscriptions. This PR adds an `cancel` signal that is sent to the background task when we no longer need a subscription.


acatangiu (approved):
nice catch!

bkontur (approved):

PR #4492, opened 2024-05-17
Bridge: fixed relayer version metric value

svyatonik:
Before relayer crates have been moved + merged, the `MetricsParams` type has been created from a `substrate-relay` crate (binary) and hence it has been setting the `substrate_relay_build_info` metic value properly - to the binary version. Now it is created from the `substrate-relay-helper` crate, which has the fixed (it isn't published) version `0.1.0`, so our relay provides incorrect metric value. This 'breaks' our monitoring tools - we see that all relayers have that incorrect version, which is not cool.

The idea is to have a global static variable (shame on me) that is initialized by the binary during initialization like we do with the logger initialization already. Was considering some alternative options:
- adding a separate argument to every relayer subcommand and propagating it to the `MetricsParams::new()` causes a lot of changes and introduces even more noise to the binary code, which is supposed to be as small as possible in the new design. But I could do that if team thinks it is better;
- adding a `structopt(skip) pub relayer_version: RelayerVersion` argument to all subcommand params won't work, because it will be initialized by default and `RelayerVersion` needs to reside in some util crate (not the binary), so it'll have the wrong value again.

### DIFF
--- a/bridges/relays/utils/src/initialize.rs
+++ b/bridges/relays/utils/src/initialize.rs
@@ -16,7 +16,12 @@
 
 //! Relayer initialization functions.
 
+use parking_lot::Mutex;
 use std::{cell::RefCell, fmt::Display, io::Write};
+
+/// Relayer version that is provided as metric. Must be set by a binary
+/// (get if with `option_env!("CARGO_PKG_VERSION")` from a binary package code).
+pub static RELAYER_VERSION: Mutex<Option<String>> = Mutex::new(None);
 
 async_std::task_local! {
 	pub(crate) static LOOP_NAME: RefCell<String> = RefCell::new(String::default());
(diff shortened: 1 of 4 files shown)


bridges/relays/utils/src/initialize.rs
```diff
@@ -16,8 +16,13 @@
 
 //! Relayer initialization functions.
 
+use parking_lot::Mutex;
 use std::{cell::RefCell, fmt::Display, io::Write};
 
+/// Relayer version that is provided as metric. Must be set by a binary
+/// (get if with `option_env!("CARGO_PKG_VERSION")` from a binary package code).
```
serban300, line 23: Typo: get **it**

serban300 (approved):

bridges/relays/utils/src/initialize.rs
svyatonik, line 23:
```suggestion
/// (get it with `option_env!("CARGO_PKG_VERSION")` from a binary package code).
```

bkontur (approved):

PR #4510, opened 2024-05-17
Contracts: Remove topics for internal events

pgherveou:


athei (approved):
Topics are not really used outside of this pallet. Apps / Indexers tend to filter through all events anyways. But it incurs a lot of overhead due to all the storage accesses it adds.

Need a prdoc though as it is an observable change for clients.

ascjones (approved):

PR #4511, opened 2024-05-17
FRAME: Enable Meta Transaction for nonexistent accounts

georgepisaltu:
Implementation of a pallet which allows sponsor accounts to provide for beneficiary nonexistent accounts. Built on top of the Meta Transaction implementation in #4122 to showcase the usecase described in #266, but the functionality is not necessarily tied to meta transactions.

Example usecase in `pallet_meta_tx::tests::nonexistent_account_meta_tx`.

This is still a POC and not production ready.

Original idea: #266 
RFC: #4123

TODO:
- events
- UTs
- benchmarks
- docs for pallet helper functions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10177,6 +10177,7 @@ dependencies = [
  "frame-benchmarking",
  "frame-support",
  "frame-system",
+ "pallet-account-sponsorship",
  "pallet-balances",
  "pallet-transaction-payment",
  "parity-scale-codec",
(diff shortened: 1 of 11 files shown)


rossbulat:
This draft PR looks stale and superseded now. It has had no follow-up activity since May 2024, and the broader meta-transaction work has since landed in #6428.

I think we can close this PR. If any remaining nonexistent-account behavior is still needed, it would be better to reopen it as a focused follow-up against current `master`.

PR #4512, opened 2024-05-17
`remote-externalities`: `rpc_child_get_keys` to use paged scraping

liamaharon:
Replace usage of deprecated `substrate_rpc_client::ChildStateApi::storage_keys` with `substrate_rpc_client::ChildStateApi::storage_keys_paged`. 

Required for successful scraping of Aleph Zero state.


bkchr (approved):

ggwpez (approved):
I think in general we should make it lazy to download keys just in time, but still cache them in a snapshot file for future re-use.  

For moonbeam eg its impossible to use try-runtime-cli or similar.

PR #4513, opened 2024-05-17
Remove pallet::getter macro usage from pallet-elections-phragmen

PolkadotDom:
As per #3326, removes pallet::getter macro usage from the pallet-elections-phragmen. The syntax `StorageItem::<T, I>::get()` should be used instead.

cc @muraca


bkchr (approved):

PolkadotDom:
@bkchr Not sure if there's a way to give me access to just setting the labels, but if so I could save you guys this step.

PR #4516, opened 2024-05-18
Adds transaction payment rpc in minimal template

gupnik:


xlc:
why? we really should depreciate it instead. everyone should be using state_call

gupnik:
> why? we really should depreciate it instead. everyone should be using state_call

So, the accounts tab in PJS apps doesn't work with the minimal template. I compared it with the [solochain](https://github.com/paritytech/polkadot-sdk/blob/e7b6d7dffd6459174f02598bd8b84fe4b1cb6e72/templates/solochain/node/src/rpc.rs#L42) one and found this to be missing. Same is present in [parachain](https://github.com/paritytech/polkadot-sdk/blob/e7b6d7dffd6459174f02598bd8b84fe4b1cb6e72/templates/parachain/node/src/rpc.rs#L47) template as well.

Are there nodes that use the state_call but work with PJS apps? If so, could you share an example please? If not, does it mean that we should change the implementation there instead?

xlc:
it is pjs apps bug, lets fix it there

kianenigma (approved):
Please make an issue in PJS apps repo about this, but in the meantime do merge this PR please so the templates are use-able. 

The broad context is to always call into runtime APIs via `state_call` and not RPC. This means both for tx payment and system nonce, or anything else. This only applies to RPC that only forward the call to the runtime API and do nothing more.

kianenigma:
Ideally the `api.rpc.foo` calls will be auto marked as deprecated with a proper message on what to do instead. This can be done as a consequence of us marking the rust APIs as deprecated. 

Then, after a few months, we can actually start removing the RPCs from the node sides.

xlc:
adding this to template will only make the deprecation process more painful as new projects using the template will inherit it and then they also need to remove those in future. 
on the other hand, it shouldn’t take more than a day of effort to fix pjs apps. it was an easy task for me to fix it for sidecar https://github.com/paritytech/substrate-api-sidecar/pull/1169

kianenigma:
Okay, I was under the wrong impression that we need this template next week for PBA, but actually we have a private fork there and can "temporarily" fix things in there, so I would have less of a strong opinion here. I would still not object to merging this though, if it gains enough support.

@gupnik please proceed as discussed with a more ground up fix and see where this goes. cc @TarikGul

gupnik:
On setting up a local instance of the PJS apps, I was able to drill this further and it turned out that the issue is actually fixed by https://github.com/polkadot-js/apps/pull/10603. Not sure why it occurred only for omni node based on the minimal fork though. 

I do see that PJS apps is already using `api.call.transactionPaymentApi`. So, it seems that this change should not be needed. Hence, closing this PR. Thanks @xlc @kianenigma.

Issue #4352, opened 2024-05-02
release chain-spec-builder


xlc:
so I can cargo install it and use it

bkchr:
CC @Morganamilo

kianenigma:
This is quite important for #3597 and related, so that we can only work with chain spec files in an easy way.

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/polkadot-parachain-omni-node-gathering-ideas-and-feedback/7823/1

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/polkadot-parachain-omni-node-gathering-ideas-and-feedback/7823/5

bkontur:
@xlc you can now `cargo install`:
```
$ cargo install staging-chain-spec-builder
$ chain-spec-builder -h
A utility to easily create a chain spec definition

Usage: chain-spec-builder [OPTIONS] <COMMAND>

Commands:
  create          Create a new chain spec by interacting with the provided runtime wasm blob
  verify          Verifies the provided input chain spec
  update-code     Updates the code in the provided input chain spec
  convert-to-raw  Converts the given chain spec into the raw format
  list-presets    Lists available presets
  display-preset  Displays given preset
  help            Print this message or the help of the given subcommand(s)

Options:
  -c, --chain-spec-path <CHAIN_SPEC_PATH>
          The path where the chain spec should be saved [default: ./chain_spec.json]
  -h, --help
          Print help
```

kianenigma:
Nice, thank you @bkontur @EgorPopelyaev!

shawntabrizi:
Why does it fail to find the binary via `cargo install` using the github repo?

```
 polkadot-sdk git:(master) ✗ cargo install --git https://github.com/paritytech/polkadot-sdk/ --tag polkadot-stable2407 chain-spec-builder
    Updating git repository `https://github.com/paritytech/polkadot-sdk/`
error: could not find `chain-spec-builder` in https://github.com/paritytech/polkadot-sdk/?tag=polkadot-stable2407 with version `*`
```

PR #4518, opened 2024-05-20
Publish `chain-spec-builder`

kianenigma:
marking it as release-able, attaching the same version number that is attached to other binaries such as `polkadot` and `polkadot-parachain`. 

I have more thoughts about the version number, though. The chain-spec builder is mainly a user of the `sp-genesis-builder` api. So the versioning should be such that it helps users know give a version of `sp-genesis-builder` in their runtime, which version of `chain-spec-builder` should they use? 

With this, we can possibly alter the version number to always match `sp-genesis-builder`.

Fixes https://github.com/paritytech/polkadot-sdk/issues/4352

- [x] Add to release artifacts ~~similar to https://github.com/paritytech/polkadot-sdk/pull/4405~~  done here: https://github.com/paritytech/polkadot-sdk/pull/4557


bkchr (approved):

gupnik (approved):

kianenigma:
@michalkucharczyk do you have further ideas about the version number to be used here?

michalkucharczyk:
> @michalkucharczyk do you have further ideas about the version number to be used here?

What you proposed is fine for me.

bkontur (approved):

bkontur:
I updated the description, release artifacts are done here: https://github.com/paritytech/polkadot-sdk/pull/4557

bkontur:
I think we can merge this.

@kianenigma
And to the same/aligned versioning for `sp-genesis-builder` and `staging-chain-spec-builder`, afaiu, it does not make sense to do any version toml magic/checks in the `master` branch, because we don't care about `[package.version]` in the `master` branch,
e.g. `pallet-balances` in the master has [28.0.0](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/Cargo.toml#L3) but actual release version is [35.0.0](https://crates.io/crates/pallet-balances). Also `sp-genesis-builder` has [0.8.0](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/genesis-builder/Cargo.toml#L3) in the master and latest creates.io release has [0.13.0](https://crates.io/crates/sp-genesis-builder).

But for every published release we have e.g. [`release-crates-io-v1.12.0/Plan.toml`](https://github.com/paritytech/polkadot-sdk/blob/release-crates-io-v1.12.0/Plan.toml) which is probably generated by [parity-publish](https://github.com/paritytech/parity-publish) tool, so I think that this tool is responsible for correct versions uploaded to the crates.io.

Thus, I think that this **tool** or **release process** to crates.io should take care about keeping the same version for the modules: `sp-genesis-builder` and `staging-chain-spec-builder`.

@Morganamilo wdyt?

kianenigma:
@bkontur Thanks for the detailed explanation. 

> afaiu, it does not make sense to do any version toml magic/checks in the master branch, because we don't care about [package.version] in the master branch,

My understanding is that this is a temporary state and we will eventually move to a system where the versions in master or more or less representative of reality. But someone from the release task force should confirm this (@liamaharon @ggwpez)

In either case, I am happy to merge this even if the version sync is not perfect yet. But, if my assumption above is not correct, please make an issue such that we can the end goal. 

Also, @michalkucharczyk do you foresee any further breaking changes to the runtime apis? if not, I am over-thinking all of this :D

bkontur:
@EgorPopelyaev @Morganamilo 
guys, when this is merged, we need to:
- [ ] 1. yank or remove actual `0.0.0` version: https://crates.io/crates/staging-chain-spec-builder/versions
- [ ] 2. backport it to the `release-crates-io-v1.12.0`
- release `staging-chain-spec-builder` as a version [1.13.0](https://github.com/paritytech/polkadot-sdk/blob/release-crates-io-v1.12.0/substrate/primitives/genesis-builder/Cargo.toml#L3)
- [ ] 3. backport it to the `release-crates-io-v1.11.0`
- release `staging-chain-spec-builder` as a version [1.12.0](https://github.com/paritytech/polkadot-sdk/blob/release-crates-io-v1.11.0/substrate/primitives/genesis-builder/Cargo.toml#L3)
- [ ] 4. close the issue: https://github.com/paritytech/polkadot-sdk/issues/4352

PR #4522, opened 2024-05-20
[BEEFY] Add runtime support for reporting fork voting

serban300:
Related to https://github.com/paritytech/polkadot-sdk/issues/4523

Extracting part of https://github.com/paritytech/polkadot-sdk/pull/1903 (credits to @Lederstrumpf  for the high-level strategy), but also introducing significant adjustments both to the approach and to the code. The main adjustment is the fact that the `ForkVotingProof` accepts only one vote, compared to the original version which accepted a `vec![]`. With this approach more calls are needed in order to report multiple equivocated votes on the same commit, but it simplifies a lot the checking logic. We can add support for reporting multiple signatures at once in the future.

There are 2 things that are missing in order to consider this issue done, but I would propose to do them in a separate PR since this one is already pretty big:
- benchmarks/computing a weight for the new extrinsic (this wasn't present in https://github.com/paritytech/polkadot-sdk/pull/1903 either)
- exposing an API for generating the ancestry proof. I'm not sure if we should do this in the Mmr pallet or in the Beefy pallet

Co-authored-by: Robert Hambrock <roberthambrock@gmail.com>


substrate/frame/beefy/src/equivocation.rs
```diff
@@ -239,49 +311,37 @@ where
...
+				log::warn!(
+					target: LOG_TARGET,
+					"rejecting unsigned report equivocation transaction because it is not local/in-block."
+				);
+				return InvalidTransaction::Call.into()
+			},
 		}
+
+		let evidence = call.to_equivocation_evidence_for().ok_or(InvalidTransaction::Call)?;
+		let tag = (evidence.offender_id().clone(), evidence.set_id(), *evidence.round_number());
```
svyatonik, line 327: Do we need to accept two different types of equivocation reports (double voting and fork voting) at a single slot? If so, we maybe need to add that type to the tag

svyatonik (approved):
LGTM

substrate/frame/beefy-mmr/src/lib.rs
```diff
@@ -172,6 +173,50 @@ where
...
 
+impl<T: Config> AncestryHelper<BlockNumberFor<T>> for Pallet<T> {
+	type Proof = AncestryProof<MerkleRootOf<T>>;
+
+	fn is_non_canonical(commitment: &Commitment<BlockNumberFor<T>>, proof: Self::Proof) -> bool {
+		let commitment_leaf_count =
+			match pallet_mmr::Pallet::<T>::block_num_to_leaf_count(commitment.block_number) {
+				Ok(commitment_leaf_count) => commitment_leaf_count,
+				Err(_) => {
+					// We consider the commitment non-canonical if the `commitment.block_number`
```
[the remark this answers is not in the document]
acatangiu: Good question!

Seeing how the keys are "autogenerated" on rotation, I don't think a validator can end up with the same key on both networks.

I think it's enough for now to say the slash can be cancelled in governance in the very unlikely case some validator does manage to do it 🙈 

We can reassess if we ever see it happening.

substrate/frame/merkle-mountain-range/src/tests.rs
```diff
@@ -787,24 +787,3 @@ fn does_not_panic_when_generating_historical_proofs() {
 		);
 	});
 }
-
-#[test]
-fn generating_and_verifying_ancestry_proofs_works_correctly() {
```
acatangiu, line 792: why remove this test?

acatangiu (approved):

substrate/frame/beefy-mmr/src/lib.rs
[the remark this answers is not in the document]
serban300: Good point ! Changes the code to return false here. It's safer like this.

substrate/frame/merkle-mountain-range/src/tests.rs
serban300: It was very similar with the new test added to `beefy_mmr`. But it's good to have it. I added it back.

substrate/frame/beefy/src/equivocation.rs
serban300: I don't think we can accept 2 equivocation reports from the same offender in the same round. I think `check_evidence` will fail:

```
	fn check_evidence(
		evidence: EquivocationEvidenceFor<T>,
	) -> Result<(), TransactionValidityError> {
		let offender = evidence.checked_offender::<P>().ok_or(InvalidTransaction::BadProof)?;

		// Check if the offence has already been reported, and if so then we can discard the report.
		let time_slot = TimeSlot { set_id: evidence.set_id(), round: *evidence.round_number() };
		if R::is_known_offence(&[offender], &time_slot) {
			Err(InvalidTransaction::Stale.into())
		} else {
			Ok(())
		}
	}
```

serban300: Or we could change `check_evidence`. I don't know what would be best. I would lean towards staying consistent with the current approach where we accept one equivocation per `(offender, set_id, round)`.

acatangiu: I don't see anywhere in this PR the extrinsic used to report these ForkEquivocations..

I am thinking since you only allow single offender per report does that mean that for a slashable fork equivocation we need to report all the offenders across multiple blocks?

serban300: > I don't see anywhere in this PR the extrinsic used to report these ForkEquivocations..

The extrinsic is [here](https://github.com/serban300/polkadot-sdk/blob/4b69c20fdd5de77b043c5fc601b4a9d111a94645/substrate/frame/beefy/src/lib.rs#L297) .

> I am thinking since you only allow single offender per report does that mean that for a slashable fork equivocation we need to report all the offenders across multiple blocks?

Hmmm good point. Could be. I can try to implement the solution that was discussed [here](https://github.com/paritytech/polkadot-sdk/pull/1903#discussion_r1538896995). It's good to have it anyway.

Lederstrumpf:
> The main adjustment is the fact that the ForkVotingProof accepts only one vote, compared to the original version which accepted a vec![]. With this approach more calls are needed in order to report multiple equivocated votes on the same commit, but it simplifies a lot the checking logic. We can add support for reporting multiple signatures at once in the future.

When you state "at once in the future":
Is this an immediate follow-up PR prior to integrating the client side changes from #1903, or possibly delayed beyond then - i.e. adjusting the client side changes in #1903 to work with a single vote prior to moving to a vec (eventually)?

substrate/frame/beefy/src/equivocation.rs
acatangiu: > The extrinsic is [here](https://github.com/serban300/polkadot-sdk/blob/4b69c20fdd5de77b043c5fc601b4a9d111a94645/substrate/frame/beefy/src/lib.rs#L297) .

Right, I got confused. I had noticed we don't have runtime APIs for the "unsigned" version of the call. Should we add it?

serban300: Yes, missed that. Will add it.

serban300:
> > The main adjustment is the fact that the ForkVotingProof accepts only one vote, compared to the original version which accepted a vec![]. With this approach more calls are needed in order to report multiple equivocated votes on the same commit, but it simplifies a lot the checking logic. We can add support for reporting multiple signatures at once in the future.
> 
> When you state "at once in the future": Is this an immediate follow-up PR prior to integrating the client side changes from #1903, or possibly delayed beyond then - i.e. adjusting the client side changes in #1903 to work with a single vote prior to moving to a vec (eventually)?

Discussed offline. I am not sure if we should support a batch of votes, since it will complicate things, but if we do, we can do any of the 2 options. They both seem ok to me.

substrate/frame/beefy/src/equivocation.rs
serban300: > > I am thinking since you only allow single offender per report does that mean that for a slashable fork equivocation we need to report all the offenders across multiple blocks?
> 
> Hmmm good point. Could be. I can try to implement the solution that was discussed [here](https://github.com/paritytech/polkadot-sdk/pull/1903#discussion_r1538896995). It's good to have it anyway.

Done. PTAL !



>> Right, I got confused. I had noticed we don't have runtime APIs for the "unsigned" version of the call. Should we add it?

> Yes, missed that. Will add it.

On a second thought, I think it's better to do this in a separate PR together with the logic for generating the ancestry proofs since the type of the ancestry proof will depend on that (it could either be `AncestryProof` or `OpaqueValue`, depending on how we generate it). WDYT ?

prdoc/pr_4522.prdoc
```diff
@@ -0,0 +1,27 @@
...
+      `submit_report_double_voting_unsigned_extrinsic`and bumps the `BeefyApi` version from 3 to 4.
+
+crates: 
+  - name: pallet-beefy
+  - name: pallet-beefy-mmr
+  - name: pallet-mmr
+  - name: sc-consensus-beefy
+  - name: kitchensink-runtime
+  - name: rococo-runtime
+  - name: westend-runtime
```
acatangiu, line 27:
```suggestion
  - name: pallet-beefy
    bump: major
  - name: pallet-beefy-mmr
    bump: major
  - name: pallet-mmr
    bump: major
  - name: sc-consensus-beefy
    bump: minor
  - name: kitchensink-runtime
    bump: major
  - name: rococo-runtime
    bump: major
  - name: westend-runtime
    bump: major
```

acatangiu (approved):

prdoc/pr_4522.prdoc
serban300: Done, Thanks !

substrate/frame/beefy/src/equivocation.rs
```diff
@@ -118,18 +121,120 @@ where
...
+	{
+		let key = (BEEFY_KEY_TYPE, self.offender_id().clone());
+		P::check_proof(key, self.key_owner_proof().clone())
+	}
+
+	fn check_equivocation_proof(self) -> Result<(), Error<T>> {
+		match self {
+			EquivocationEvidenceFor::DoubleVotingProof(equivocation_proof, _) => {
+				// Validate equivocation proof (check votes are different and signatures are valid).
+				if !sp_consensus_beefy::check_equivocation_proof(&equivocation_proof) {
```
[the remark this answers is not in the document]
serban300: Done

prdoc/pr_4522.prdoc
```diff
@@ -0,0 +1,34 @@
...
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Added runtime support for reporting BEEFY fork voting
+
+doc:
+  - audience:
+    - Runtime Dev
+    - Runtime User
+    description: |
+      This PR adds the `report_fork_voting`, `report_future_voting` extrinsics to `pallet-beefy`
```
bkontur, line 11: I would maybe mention that you need to specify `AncestryHelper` for `pallet_beefy::Config` - it is breaking change, right?

substrate/frame/beefy/src/default_weights.rs
```diff
@@ -50,6 +57,11 @@ impl crate::WeightInfo for () {
 			.saturating_add(DbWeight::get().reads(2))
 	}
 
+	// TODO: Calculate
```
bkontur, line 60: @serban300 what about this TODO?

bkontur (approved):

prdoc/pr_4522.prdoc
```diff
@@ -0,0 +1,34 @@
+# Schema: Polkadot SDK PRDoc Schema (prdoc) v1.0.0
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Added runtime support for reporting BEEFY fork voting
+
+doc:
+  - audience:
+    - Runtime Dev
+    - Runtime User
```
franciscoaguirre, line 9: Is this description relevant to runtime users?

prdoc/pr_4522.prdoc
serban300: Yes. Good point. Will mention it.

franciscoaguirre (approved):

prdoc/pr_4522.prdoc
serban300: Done

prdoc/pr_4522.prdoc
serban300: I'm thinking it might be helpful for them to know of the new extrinsics.

prdoc/pr_4522.prdoc
```diff
@@ -22,14 +22,16 @@ crates:
...
   - name: kitchensink-runtime
     bump: major
   - name: rococo-runtime
-    bump: major
+    bump: minor
   - name: westend-runtime
-    bump: major
+    bump: minor
+  - name: sp-consensus-beefy
+  - name: polkadot-service
```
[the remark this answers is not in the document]
serban300: Hmmm I'm thinking maybe it's using `patch` by default if I didn't specify anything. Anyway waiting for semver to run. Will fix if needed.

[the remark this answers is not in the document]
serban300: Done

PR #4527, opened 2024-05-20
[Assets] Call implementation for `transfer_all`

### LINKED ISSUE #4517: `transfer_all` Call in Assets Pallet
With the ability to pay fees in assets, there is demand from some users to be able to "pay fees with and transfer all", completely killing the balance of the asset on the account. We should add this `transfer_all` call to the Assets pallet to make this easier without needing to calculate fees/exchange rates to determine transferrable amount.

@pandres95 ?

pandres95:
Closes #4517

Polkadot address: 12gMhxHw8QjEwLQvnqsmMVY1z5gFa54vND74aMUbhhwN6mJR


prdoc/pr_4527.prdoc
```diff
@@ -0,0 +1,27 @@
...
+  - audience: Runtime User
+    description: |
+      This PR introduces the `transfer_all` call for `pallet-assets`.
+      The parameters are analog to the same call in `pallet-balances`.
+      When using this call, consider that transferring all funds of an asset
+      account might kill it when setting `keep_alive` as false.
+
+crates:
+  - name: pallet-assets
+    bump: major
```
joepetrowski, line 23: Should this be minor? It doesn't break any interface.

substrate/frame/assets/src/lib.rs
```diff
@@ -1687,6 +1691,49 @@ pub mod pallet {
...
+		/// deposits, etc...
+		///
+		/// The dispatch origin of this call must be Signed.
+		///
+		/// - `id`: The identifier of the asset for the account holding a deposit.
+		/// - `dest`: The recipient of the transfer.
+		/// - `keep_alive`: A boolean to determine if the `transfer_all` operation should send all
+		///   of the funds the asset account has, causing the sender asset account to be killed
+		///   (false), or transfer everything except at least the existential deposit, which will
+		///   guarantee to keep the sender asset account alive (true).
```
joepetrowski, line 1710: Assets have a `minimum_balance` but no existential deposit. This needs to be reworded. Also does `keep_alive` refer to `Accounts` in the context of the Assets pallet (i.e. `(AssetId, AccountId)`) or in the context of the system?

pandres95: In this case, it refers to the context of Assets pallet. I'll make sure to reword that part.

joepetrowski (approved):
Sorry that I missed the ping @pandres95

prdoc/pr_4527.prdoc
```diff
@@ -0,0 +1,27 @@
...
+    description: |
+      This PR introduces the `transfer_all` call for `pallet-assets`.
+      The parameters are analog to the same call in `pallet-balances`.
+      This change shouldn't break the existing APIs.
+      This change requires running benchmarkings to set accurate weights for
+      the call.
+  - audience: Runtime User
+    description: |
+      This PR introduces the `transfer_all` call for `pallet-assets`.
+      The parameters are analog to the same call in `pallet-balances`.
```
muharem, line 17:
```suggestion
```

substrate/frame/assets/src/benchmarking.rs
```diff
@@ -553,5 +553,15 @@ benchmarks_instance_pallet! {
 		assert_last_event::<T, I>(Event::Blocked { asset_id: asset_id.into(), who: caller }.into());
 	}
 
+	transfer_all {
+		let amount = T::Balance::from(100u32);
```
muharem, line 557: better use min_balance of the asset id here instead, multiply it by two for example.

substrate/frame/assets/src/tests.rs
```diff
@@ -795,6 +795,50 @@ fn transferring_to_blocked_account_should_not_work() {
...
+		assert_ok!(Assets::mint(RuntimeOrigin::signed(0), 0, 1, 200));
+		assert_ok!(Assets::mint(RuntimeOrigin::signed(0), 0, 2, 100));
+		// transfer all and allow death
+		assert_ok!(Assets::transfer_all(Some(1).into(), 0, 2, true));
+		assert_eq!(Assets::balance(0, &1), 100);
+		assert_eq!(Assets::balance(0, &2), 200);
+	});
+}
+
+// /// TODO: Cover this case once #3951 is merged
```
muharem, line 826: its merged now

prdoc/pr_4527.prdoc
```diff
@@ -0,0 +1,27 @@
...
+# See doc at https://raw.githubusercontent.com/paritytech/polkadot-sdk/master/prdoc/schema_user.json
+
+title: Call implementation for `transfer_all`
+
+doc:
+  - audience: Runtime Dev
+    description: |
+      This PR introduces the `transfer_all` call for `pallet-assets`.
+      The parameters are analog to the same call in `pallet-balances`.
+      This change shouldn't break the existing APIs.
```
muharem, line 11: new call is new enum variant within Call enum. so this is a breaking change as for a rust crate.

prdoc/pr_4527.prdoc
```diff
@@ -0,0 +1,27 @@
...
+  - audience: Runtime User
+    description: |
+      This PR introduces the `transfer_all` call for `pallet-assets`.
+      The parameters are analog to the same call in `pallet-balances`.
+      When using this call, consider that transferring all funds of an asset
+      account might kill it when setting `keep_alive` as false.
+
+crates:
+  - name: pallet-assets
+    bump: minor
```
muharem, line 23:
```suggestion
    bump: major
```

prdoc/pr_4527.prdoc
joepetrowski: That's rough. I guess we have to follow crates rules but it's not breaking for transaction encoding.

muharem: Yes, but for crates versioning I think it should follow the rust semver rules. At least that's what we have not in guidelines.

substrate/frame/assets/src/benchmarking.rs
```diff
@@ -553,5 +553,15 @@ benchmarks_instance_pallet! {
 		assert_last_event::<T, I>(Event::Blocked { asset_id: asset_id.into(), who: caller }.into());
 	}
 
+	transfer_all {
+		let amount = T::Balance::from(2u32);
```
muharem, line 557: I meant something like `Assets::<T, I>::minimum_balance(asset_id) * 2`

prdoc/pr_4527.prdoc
joepetrowski: Yeah, @ggwpez also pointed out that it breaks the `WeightInfo` trait for anyone using this pallet. So I agree it is (unfortunately) major.

muharem (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/2024-07-16-technical-fellowship-opendev-call/9512/1

joepetrowski:
@pandres95 is this waiting for anything before merging?

pandres95:
@joepetrowski not that I know. I thought it was waiting for an additional review / auditing process. 😅

substrate/frame/assets/src/lib.rs
```diff
@@ -1753,6 +1757,49 @@ pub mod pallet {
...
+		///
+		/// - `id`: The identifier of the asset for the account holding a deposit.
+		/// - `dest`: The recipient of the transfer.
+		/// - `keep_alive`: A boolean to determine if the `transfer_all` operation should send all
+		///   of the funds the asset account has, causing the sender asset account to be killed
+		///   (false), or transfer everything except at least the minimum balance, which will
+		///   guarantee to keep the sender asset account alive (true).
+		#[pallet::call_index(32)]
+		#[pallet::weight(T::WeightInfo::refund_other())]
+		pub fn transfer_all(
```
gui1117, line 1779: weight is calling the wrong function.

joepetrowski: Ahh good catch

PR #4529, opened 2024-05-20
Remove usage of the pallet::getter macro from pallet-grandpa

PolkadotDom:
As per #3326, removes pallet::getter macro usage from pallet-grandpa. The syntax `StorageItem::<T, I>::get()` should be used instead.

cc @muraca


substrate/frame/grandpa/src/lib.rs
```diff
@@ -430,6 +424,54 @@ pub enum StoredState<N> {
 }
 
 impl<T: Config> Pallet<T> {
+	/// Get the tate of the current authority set
+	pub fn state() -> StoredState<BlockNumberFor<T>> {
```
bkchr, line 428: Why did you started to add them here now?

davxy: Indeed. Furthermore each of them can be simplified. E.g.
```
pub fn state() -> StoredState<BlockNumberFor<T>> {
    State::<T>::get()
}
```

PolkadotDom: Ah good call on the simplification, I had just replaced them with what the macro generated. Will fix 👍 

And I've added the explicit getters because the getter macro made the storage values read only outside the crate (since they're pub(super)), so taking the macro away reduced their visibility to other pallets. Usually this isn't an issue because most the storage values are pub, but I opted not to switch to pub because I feel preserving data authority is important, especially since I don't know the reasoning behind why they were made read-only originally.

bkchr: There is no real reasoning and anyone could write them by just writing to the storage items. So, just make the storage items public.

PolkadotDom: Sounds good. I'll fix and subsequently do this moving forward.

davxy (approved):

bkchr:
@PolkadotDom can you rework as the other prs? With the getters as functions?

substrate/frame/grandpa/src/lib.rs
```diff
@@ -432,6 +426,46 @@ pub enum StoredState<N> {
...
+	/// A mapping from grandpa set ID to the index of the *most recent* session for which its
+	/// members were responsible.
+	///
+	/// This is only used for validating equivocation proofs. An equivocation proof must
+	/// contains a key-ownership proof for a given session, therefore we need a way to tie
+	/// together sessions and GRANDPA set ids, i.e. we need to validate that a validator
+	/// was the owner of a given key on a given session, and what the active set ID was
+	/// during that session.
+	///
+	/// TWOX-NOTE: `SetId` is not under user control.
```
bkchr, line 464:
```suggestion
```

bkchr (approved):

PR #4537, opened 2024-05-22
[Pools] Refactors and runtime apis for DelegateStake

Ank4n:
## Runtime Apis
Introduces the following runtime apis to facilitate dapps and wallets in integrating with the `DelegateStake` functionalities of the pools (related: https://github.com/paritytech/polkadot-sdk/pull/3905). These apis are meant to support pool and member migration, as well as lazy application of pending slashes of pool members.

```rust
fn pool_pending_slash(pool_id: PoolId) -> Balance;
fn member_pending_slash(member: AccountId) -> Balance;
fn pool_needs_delegate_migration(pool_id: PoolId) -> bool;
fn member_needs_delegate_migration(member: AccountId) -> bool;
```

## Refactors
- Introduces newtypes for `Agent`, `Delegator`, `Pool` and `[Pool]Member`. And refactors `StakeAdapter` and `DelegationInterface` to accept the above types. This will help make these apis typesafe against using wrong account type.
- Fixing `DelegationInterface` apis to return optional (instead of default value if key does not exist).
- Rename struct `Agent` that wraps `AgentLedger` to `AgentOuterLedger` which is clearer (naming wise) and different from the newtype `Agent`.
- Cleaning up new Pool events (related to `Delegation` feature of pool).


gpestana (commented):
wdyt about making the code a bit more readable by implementing defer in the agent/delegator account struct? something along the lines of

```rust
impl<T> Deref for AgentAccount<T> {
    type Target = T;
    
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

// now we can access the account Id by defering the `AgentAccount` instance:
let a = AgentAccount(1);
assert_eq!(a.0, *a)
```

substrate/frame/nomination-pools/runtime-api/src/lib.rs
```diff
@@ -38,5 +38,30 @@ sp_api::decl_runtime_apis! {
 
 		/// Returns the equivalent points of `new_funds` for a given pool.
 		fn balance_to_points(pool_id: PoolId, new_funds: Balance) -> Balance;
+
+		/// Returns the pending slash for a given pool.
+		fn pool_pending_slash(pool_id: PoolId) -> Balance;
+
+		/// Returns the pending slash for a given pool member.
+		fn member_pending_slash(member: AccountId) -> Balance;
+
```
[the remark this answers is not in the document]
Ank4n: I don't think we have plans to allow members belonging to multiple pools. I think though what we want to enable is have a way to switching pools (which will probably require something like a dex that exchanges points from different pools).

Ank4n:
> wdyt about making the code a bit more readable by implementing defer in the agent/delegator account struct? something along the lines of
> 
> ```rust
> impl<T> Deref for AgentAccount<T> {
>     type Target = T;
>     
>     fn deref(&self) -> &Self::Target {
>         &self.0
>     }
> }
> 
> // now we can access the account Id by defering the `AgentAccount` instance:
> let a = AgentAccount(1);
> assert_eq!(a.0, *a)
> ```

I like this suggestion. But since most of the functions in our pallets take account ref as param, we will in most cases replace `&acc.0` with `&*acc` which also is a bit weird to read. Wdyt?

substrate/frame/nomination-pools/src/lib.rs
```diff
@@ -3468,31 +3474,59 @@ impl<T: Config> Pallet<T> {
...
+		pool_member: PoolMember<T>,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		// only executed in tests: ensure the member account is correct.
+		debug_assert!(
+			PoolMembers::<T>::get(member_account.0.clone()).expect("member must exist") ==
+				pool_member
+		);
+
+		let pool_account = Pallet::<T>::generate_bonded_account(pool_member.pool_id);
+		// if the pool doesn't have any pending slash, it implies the member also does not have any
```
gpestana, line 3507: Why do we perform this check before? It seems that checking if the pool doesn't have any pending slash is as expensive as retrieving the member delegation balance.

Thus we end up paying for an extra read if the member is to be slashed and not gaining anything in return.

gpestana (approved):
Added a couple suggestions/comments but overall looks good. The type safety introduced is a big win IMO.

substrate/frame/nomination-pools/src/lib.rs
Ank4n: `pool_member.total_balance()` also does couple of extra reads along with iterating over the unbonding eras. But your point is fair if we look only at `DelegateStake` adapter. 

My intention was to short circuit these calls without any extra reads for `TransferStake` adapter (its impl just returns zero). Once we remove them (if we ever remove them), we probably can remove this check.

Ank4n:
> > > wdyt about making the code a bit more readable by implementing defer in the agent/delegator account struct? something along the lines of
> > > ```rust
> > > impl<T> Deref for AgentAccount<T> {
> > >     type Target = T;
> > >     
> > >     fn deref(&self) -> &Self::Target {
> > >         &self.0
> > >     }
> > > }
> > > 
> > > // now we can access the account Id by defering the `AgentAccount` instance:
> > > let a = AgentAccount(1);
> > > assert_eq!(a.0, *a)
> > > ```
> > 
> > 
> > I like this suggestion. But since most of the functions in our pallets take account ref as param, we will in most cases replace `&acc.0` with `&*acc` which also is a bit weird to read. Wdyt?
> 
> Yes, I agree that the deref may not help on the readability after all. One last suggestion would be to rename `AgentAccount<AccountId>` etc to `Agent<AccountId>` and then implement a method that would return the account ID, something along the lines of:
> 
> ```rust
> pub struct Agent<T>(pub T);
> impl<T: Copy> Agent<T> {
>     fn account(&self) -> T {
>         self.0
>     }
> }
> ```
> 
> In any case, the current code is good enough, just trying to come up with a way to avoid sprinkling `*.0` all over the code.

Made the inner field private and added a `fn get()` hopefully which makes it lot nicer to look at.

substrate/frame/delegated-staking/src/impls.rs
```diff
@@ -19,80 +19,78 @@
...
-			.unwrap_or(false)
+	/// Returns pending slash of the `agent`.
+	fn pending_slash(agent: Agent<Self::AccountId>) -> Option<Self::Balance> {
+		AgentLedgerOuter::<T>::get(&agent.get()).map(|d| d.ledger.pending_slash).ok()
 	}
 
 	fn delegator_slash(
-		agent: &Self::AccountId,
-		delegator: &Self::AccountId,
+		agent: Agent<Self::AccountId>,
```
kianenigma, line 88: This is a great refactor, the base `nomination-pools` pallet also has similar foot guns, maybe a good `mentor` issue if you describe it in one.

kianenigma (approved):
Pretty cool!

Polkadot-Forum:

https://forum.polkadot.network/t/changes-for-nomination-pool-members-and-opengov-participation/9651/3

PR #4542, opened 2024-05-22
Adds ability to specify chain type in chain-spec-builder

gupnik:
Currently, `chain-spec-builder` only creates a spec with `Live` chain type. This PR adds the ability to specify it while keeping the same default.


substrate/bin/utils/chain-spec-builder/src/lib.rs
```diff
@@ -154,6 +154,9 @@ pub struct CreateCmd {
 	/// The chain id.
 	#[arg(long, short = 'i', default_value = "custom")]
 	chain_id: String,
+	/// The chain type.
+	#[arg(long, short = 't', default_value = "Live")]
+	chain_type: String,
```
bkchr, line 159: Please use an `ArgEnum` for this.

kianenigma:
I wonder though, does this make any tangible difference other than changing the `ChainType` in the json file? 

What would be cool if the `sp_keyring` accounts are all injected into the JSON file.

xlc:
I feel it is easier to just have another tool to modify the JSON file

gupnik:
> I wonder though, does this make any tangible difference other than changing the `ChainType` in the json file?

PJS is able to show development accounts for types other than live. They do not carry any balance though at the moment. Happy to look into including that as well.

kianenigma:
Yeah I am also leaning toward the fact that if this is just modifying the JSON file, it can be done with hand.

michalkucharczyk (approved):

michalkucharczyk:
> What would be cool if the `sp_keyring` accounts are all injected into the JSON file.

Do we know how to serialize keys to json (which is domain of runtime)?
 
> cc @michalkucharczyk should the runtime be aware of the `chain_type` and make an informed decision here?

Well I like to ask question does this property belong to runtime domain?
If this is a kind of flag  that enables some debug features in app, I would not say. It is a property of network/nodes, so probably should stay outside runtime. You probably would not like to recompile runtime to enable it.

michalkucharczyk:
Adding flags like this is fine for me. No need to run additional tool.

substrate/client/chain-spec/Cargo.toml
```diff
@@ -16,6 +16,7 @@ workspace = true
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
+clap = { version = "4.5.3", features = ["derive"] }
```
michalkucharczyk, line 19: one more thought: maybe clap shall be optional?
That would allow to use this module w/o all clap deps. If anyone needs to expose types to cli, the `clap` feature could be enabled.
 @gupnik wdyt?

gupnik: Yeah, makes sense. Will update it accordingly. Thanks.

substrate/client/chain-spec/Cargo.toml
```diff
@@ -43,3 +43,8 @@ substrate-test-runtime = { path = "../../test-utils/runtime" }
 sp-keyring = { path = "../../primitives/keyring" }
 sp-application-crypto = { default-features = false, path = "../../primitives/application-crypto", features = ["serde"] }
 sp-consensus-babe = { default-features = false, path = "../../primitives/consensus/babe", features = ["serde"] }
+
+[features]
```
michalkucharczyk, line 48: If I recall correctly this may not be needed (should be added implicitly).

substrate/client/chain-spec/Cargo.toml
```diff
@@ -43,3 +43,8 @@ substrate-test-runtime = { path = "../../test-utils/runtime" }
...
+clap = [
+    "dep:clap"
+]
```
michalkucharczyk, line 51: If I recall correctly this may not be needed (should be added implicitly).

gupnik: You are right. Thanks!

Issue #1657, opened 2023-09-21
Move all tests to use u32 as BlockNumber


gupnik:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

As we expand the usage of `derive-impl` being tracked in https://github.com/paritytech/polkadot-sdk/issues/171, it seems that we would need to maintain two default configs to support both u32 and u64 BlockNumber types. 

Since there's no apparent reason to use different types for tests, it might be better to move them to use u32.

philoniare:
@gupnik I'd like to work on this one, could you please assign it to me?

ggwpez:
@gupnik looking at this issue again: Why `u32` and not `u64`?  
It looks like our `TestDefaultConfig` uses `u64` for `BlockHashCount`.

kianenigma:
> @gupnik looking at this issue again: Why `u32` and not `u64`? It looks like our `TestDefaultConfig` uses `u64` for `BlockHashCount`.

I don't think there is any reasonable advantage to either. I suggest @philoniare to do whichever produces the smaller diff.

philoniare:
> > @gupnik looking at this issue again: Why `u32` and not `u64`? It looks like our `TestDefaultConfig` uses `u64` for `BlockHashCount`.
> 
> 
> 
> I don't think there is any reasonable advantage to either. I suggest @philoniare to do whichever produces the smaller diff. 

Got it, will do. Appreciate the guidance 🙏

gupnik:
> @gupnik looking at this issue again: Why `u32` and not `u64`? It looks like our `TestDefaultConfig` uses `u64` for `BlockHashCount`.

@ggwpez I think the discussion at that time was to make it use the same type as the relay, but yes fine to use either.

philoniare:
Ended up choosing the u64 as it had smaller diff

xlc:
we can make it u32 😮‍💨 all relay and system parachains are using u32 and there is zero reason to use u64

gupnik:
> can we make it u32 😮‍💨 all relay and system parachains are using u32 and there is zero reason to use u64 let's correct the mistake, not make it even worse just because it had smaller diff

Yeah, I initially thought the same but does it really matter? The framework accepts any type with certain constraints and it should be acceptable to add any type for tests that satisfies those constraints?

xlc:
[...]

PR #4543, opened 2024-05-22
Attempt to avoid specifying `BlockHashCount` for different `mocking::{MockBlock, MockBlockU32, MockBlockU128}`

bkontur:
While doing some migration/rebase I came in to the situation, where I needed to change `mocking::MockBlock` to `mocking::MockBlockU32`:
```
#[derive_impl(frame_system::config_preludes::TestDefaultConfig)]
impl frame_system::Config for TestRuntime {
	type Block = frame_system::mocking::MockBlockU32<TestRuntime>;
	type AccountData = pallet_balances::AccountData<ThisChainBalance>;
}
```
But actual `TestDefaultConfig` for `frame_system` is using `ConstU64` for  `type BlockHashCount = frame_support::traits::ConstU64<10>;` [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/system/src/lib.rs#L303). Because of this, it force me to specify and add override for `type BlockHashCount = ConstU32<10>`.

This PR tries to fix this with `TestBlockHashCount` implementation for `TestDefaultConfig` which supports `u32`, `u64` and `u128` as a `BlockNumber`. 

### How to simulate error
Just by removing `type BlockHashCount = ConstU32<250>;` [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/multisig/src/tests.rs#L44)
```
:~/parity/olkadot-sdk$ cargo test -p pallet-multisig
   Compiling pallet-multisig v28.0.0 (/home/bparity/parity/aaa/polkadot-sdk/substrate/frame/multisig)
error[E0277]: the trait bound `ConstU64<10>: frame_support::traits::Get<u32>` is not satisfied
   --> substrate/frame/multisig/src/tests.rs:41:1
    |
41  | #[derive_impl(frame_system::config_preludes::TestDefaultConfig)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `frame_support::traits::Get<u32>` is not implemented for `ConstU64<10>`
    |
    = help: the following other types implement trait `frame_support::traits::Get<T>`:
              <ConstU64<T> as frame_support::traits::Get<u64>>
              <ConstU64<T> as frame_support::traits::Get<std::option::Option<u64>>>
note: required by a bound in `frame_system::Config::BlockHashCount`
   --> /home/bparity/parity/aaa/polkadot-sdk/substrate/frame/system/src/lib.rs:535:24
    |
535 |         type BlockHashCount: Get<BlockNumberFor<Self>>;
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Config::BlockHashCount`
    = note: this error originates in the attribute macro `derive_impl` which comes from the expansion of the macro `frame_support::macro_magic::forward_tokens_verbatim` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `pallet-multisig` (lib test) due to 1 previous error 
```

## For reviewers:

(If there is a better solution, please let me know!)

The first commit contains actual attempt to fix the problem: https://github.com/paritytech/polkadot-sdk/commit/3c5499e539f2218503fbd6ce9be085b03c31ee13.
The second commit is just removal of `BlockHashCount` from all other places where not needed by default.

Closes: https://github.com/paritytech/polkadot-sdk/issues/1657

### DIFF
--- a/substrate/frame/system/src/lib.rs
+++ b/substrate/frame/system/src/lib.rs
@@ -270,18 +270,8 @@ pub mod pallet {
 		/// NOTE: Avoids overriding `BlockHashCount` when using `mocking::{MockBlock, MockBlockU32,
 		/// MockBlockU128}`.
 		pub struct TestBlockHashCount<C: Get<u32>>(sp_std::marker::PhantomData<C>);
-		impl<C: Get<u32>> Get<u32> for TestBlockHashCount<C> {
-			fn get() -> u32 {
-				C::get()
-			}
-		}
-		impl<C: Get<u32>> Get<u64> for TestBlockHashCount<C> {
-			fn get() -> u64 {
-				C::get().into()
-			}
-		}
-		impl<C: Get<u32>> Get<u128> for TestBlockHashCount<C> {
-			fn get() -> u128 {
+		impl<I: From<u32>, C: Get<u32>> Get<I> for TestBlockHashCount<C> {
+			fn get() -> I {
 				C::get().into()
 			}
 		}
--- a/substrate/frame/system/src/lib.rs
+++ b/substrate/frame/system/src/lib.rs
@@ -262,7 +262,29 @@ pub mod pallet {
 	/// Default implementations of [`DefaultConfig`], which can be used to implement [`Config`].
 	pub mod config_preludes {
 		use super::{inject_runtime_type, DefaultConfig};
-		use frame_support::derive_impl;
+		use frame_support::{derive_impl, traits::Get};
+
+		/// A predefined adapter that covers `BlockNumberFor<T>` for `Config::Block::BlockNumber` of
+		/// the types `u32`, `u64`, and `u128`.
+		///
+		/// NOTE: Avoids overriding `BlockHashCount` when using `mocking::{MockBlock, MockBlockU32,
+		/// MockBlockU128}`.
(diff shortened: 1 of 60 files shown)


substrate/frame/system/src/lib.rs
```diff
@@ -300,7 +312,7 @@ pub mod pallet {
 			#[inject_runtime_type]
 			type RuntimeTask = ();
 			type BaseCallFilter = frame_support::traits::Everything;
-			type BlockHashCount = frame_support::traits::ConstU64<10>;
+			type BlockHashCount = TestBlockHashCount<frame_support::traits::ConstU32<10>>;
```
franciscoaguirre, line 315: Most mocks had `250` for this. I assume it's good anyway, right?

franciscoaguirre (approved):

substrate/frame/system/src/lib.rs
bkontur: yes, I think 250 was just random copy&pasted value, looks like there is no test depending on this value, so all test passed and should be ok

acatangiu (approved):

PR #4550, opened 2024-05-23
Define `OpaqueValue`

serban300:
Define `OpaqueValue` and use it instead of `grandpa::OpaqueKeyOwnershipProof` and `beefy:OpaqueKeyOwnershipProof`

Related to https://github.com/paritytech/polkadot-sdk/pull/4522#discussion_r1608278279

We'll need to introduce a runtime API method that calls the `report_fork_voting_unsigned()` extrinsic. This method will need to receive the ancestry proof as a paramater. I'm still not sure, but there is a chance that we'll send the ancestry proof as an opaque type. 

So let's introduce this `OpaqueValue`. We can already use it to replace `grandpa::OpaqueKeyOwnershipProof` and `beefy:OpaqueKeyOwnershipProof` and maybe we'll need it for the ancestry proof as well.

### DIFF
--- a/substrate/primitives/runtime/src/lib.rs
+++ b/substrate/primitives/runtime/src/lib.rs
@@ -1019,7 +1019,7 @@ impl OpaqueValue {
 	}
 
 	/// Try to decode this `OpaqueValue` into the given concrete type.
-	pub fn decode<T: Decode>(self) -> Option<T> {
+	pub fn decode<T: Decode>(&self) -> Option<T> {
 		Decode::decode(&mut &self.0[..]).ok()
 	}
 }
--- a/substrate/primitives/runtime/src/lib.rs
+++ b/substrate/primitives/runtime/src/lib.rs
@@ -1009,6 +1009,21 @@ pub enum ExtrinsicInclusionMode {
 	OnlyInherents,
 }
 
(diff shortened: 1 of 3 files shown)


svyatonik (approved):

bkchr (approved):

substrate/primitives/runtime/src/lib.rs
```diff
@@ -1009,6 +1009,21 @@ pub enum ExtrinsicInclusionMode {
...
+#[derive(Decode, Encode, PartialEq, TypeInfo)]
+pub struct OpaqueValue(Vec<u8>);
+impl OpaqueValue {
+	/// Create a new `OpaqueValue` using the given encoded representation.
+	pub fn new(inner: Vec<u8>) -> OpaqueValue {
+		OpaqueValue(inner)
+	}
+
+	/// Try to decode this `OpaqueValue` into the given concrete type.
+	pub fn decode<T: Decode>(self) -> Option<T> {
```
bkchr, line 1022:
```suggestion
	pub fn decode<T: Decode>(&self) -> Option<T> {
```

Not required to consume

serban300: Thanks ! Done

Issue #4354, opened 2024-05-02
Release `frame-omni-bencher`


bkontur:
I have a PoC for using `frame-omni-bencher` in the fellows repo: https://github.com/polkadot-fellows/runtimes/pull/298. However, since there is no released version, I have to build it from the sources each time for every runtime pipeline, which is a waste of resources.

#### How to re-use `frame-omni-bencher` for fellows or elsewhere:
- [ ] create a github action for that, as we have for [try-runtime-gha](https://github.com/paritytech/try-runtime-gha) e.g. used [here](https://github.com/polkadot-fellows/runtimes/blob/main/.github/workflows/check-migrations.yml#L74)
- [ ] release for `cargo install`, e.g. as we have for zepter, [`cargo install --locked -q zepter`](https://github.com/polkadot-fellows/runtimes/blob/main/.github/workflows/check-features.yaml#L27)
- [ ] release just as binary to `polkadot-sdk` repo/release, so we just setup downloading from polkadot-sdk in fellows CI?
- [ ] other?

cc: @ggwpez

ggwpez:
How long is the build time?  
Binary release is probably the fastest for CI. AFAIK the GHA for try-runtime is also doing that.  

@EgorPopelyaev do you know what would be the steps to attach another binary to the release?

bkontur:
> How long is the build time? 

@ggwpez https://github.com/polkadot-fellows/runtimes/actions/runs/8923074053/job/24506630345?pr=298#step:11:1001
```
 Compiling frame-omni-bencher v0.1.0 (/home/runner/work/runtimes/runtimes/polkadot-sdk/substrate/utils/frame/omni-bencher)
    Finished release [optimized] target(s) in 14m 16s
```

EgorPopelyaev:
Hey @bkontur @ggwpez, if there will be a github flow or job which builds the binary. I could integrate it in this [flow](https://github.com/paritytech/polkadot-sdk/blob/master/.github/workflows/release-30_publish_release_draft.yml), which creates release draft, the same way how it is done with runtimes

bkontur:
> Hey @bkontur @ggwpez, if there will be a github flow or job which builds the binary. I could integrate it in this [flow](https://github.com/paritytech/polkadot-sdk/blob/master/.github/workflows/release-30_publish_release_draft.yml), which creates release draft, the same way how it is done with runtimes

ok, thank you, I will check later,
[...]

PR #4557, opened 2024-05-23
Add omni bencher & chain-spec-builder bins to release

EgorPopelyaev:
Closes: https://github.com/paritytech/polkadot-sdk/issues/4354

This PR adds the steps to build and attach `frame-omni-bencher` and `chain-spec-builder` binaries to the release draft

## TODO
- [x] add also chain-spec-builder binary
- [ ] ~~check/investigate Kian's comment: `chain spec builder. Ideally I want it to match the version of the sp-genesis-builder crate`~~ see [comment](https://github.com/paritytech/polkadot-sdk/pull/4518#issuecomment-2134731355)
- [ ] Backport to `polkadot-sdk@1.11` release, so we can use it for next fellows release: https://github.com/polkadot-fellows/runtimes/pull/324
- [ ] Backport to `polkadot-sdk@1.12` release


.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -23,10 +23,31 @@ jobs:
...
+      - name: Checkout sources
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
+      - name: Build frame-omni-bencher binary
+        run: |
+          sudo apt update
+          sudo apt install -y protobuf-compiler
+
+          cargo build --locked --release -p frame-omni-bencher
+          target/release/frame-omni-bencher --version
```
bkontur, line 42: Let's use production profile

```suggestion
          cargo build --locked --profile=production -p frame-omni-bencher
          target/production/frame-omni-bencher --version
```

.github/workflows/release-30_publish_release_draft.yml
```diff
@@ -23,10 +23,31 @@ jobs:
...
+
+      - name: Upload frame-omni-bencher binary
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
+        with:
+          name: frame-omni-bencher
+          path: target/release/frame-omni-bencher
```
bkontur, line 48:
```suggestion
          path: target/production/frame-omni-bencher
```

bkontur:
@kianenigma 
Do we need [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/bin/utils/chain-spec-builder/Cargo.toml#L23) this `rlib`?
```
[lib]
crate-type = ["rlib"]
```

EgorPopelyaev:
@kianenigma @bkontur guys, the flow is working now and both bins can be built and attached to the release. 
Regarding the check that the version of the `chain-spec-builder` should match the `sp-genesis-builder`, not sure what to do with that, should be done here as part of the pipeline?

kianenigma:
> @kianenigma Do we need [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/bin/utils/chain-spec-builder/Cargo.toml#L23) this `rlib`?
> 
> ```
> [lib]
> crate-type = ["rlib"]
> ```

I don't think so, but I am also not the original author of the crate to know why it was there to begin with. From what I read about the definition of `rlib`, it can be removed. Is it causing issues with the release?

bkontur (approved):

bkontur:
> > @kianenigma Do we need [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/bin/utils/chain-spec-builder/Cargo.toml#L23) this `rlib`?
> > ```
> > [lib]
> > crate-type = ["rlib"]
> > ```
> 
> I don't think so, but I am also not the original author of the crate to know why it was there to begin with. From what I read about the definition of `rlib`, it can be removed. Is it causing issues with the release?

@kianenigma no, there is no problem, it just caught my attention, and we don' use it elsewhere except two binaries:  [`polkadot-cli`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/cli/Cargo.toml#L18) and [`staging-node-cli`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/bin/node/cli/Cargo.toml#L36)

and to the same versioning for those two crates, please continue discussions here: https://github.com/paritytech/polkadot-sdk/pull/4518#issuecomment-2134731355

AndWeHaveAPlan (approved):

alvicsam (approved):

pavelsupr (approved):
Everything is OK in this specific commit for me, but the initial release file contained a binary dependency (L. 76) from a single developer without pinning/checks of the integrity of it. IMO it should be localised or replaced.

kianenigma (approved):

Issue #1202, opened 2023-08-28
Improve transaction handling for Parachains


bkchr:
The transaction pool is designed to send out events about `inclusion` of a transaction in a block when the transaction got included in a block that is seen as best block of the local node. If this block then for example is removed from the best chain (because we forked), a `retracted` event is sent. On parachains a block at the tip of the chain isn't imported as best block. Setting the best block happens asynchronously as it depends on what is the best relay chain block and which parachain block got included in this relay chain block. This means that instead of waiting ~6 seconds on the relay chain we have to wait at least 12 seconds on a Parachain. 

To improve the situation we should directly send the `inclusion` event when the transaction appears in any block. It should not be important if the block is a best block or not. I also don't see any real benefit in sending a `retracted` event (we should check the new RPC spec, but I doubt that it is still send there) and we should be able to omit it.

bkchr:
This change is also important to have [`ready_at`](https://github.com/paritytech/polkadot-sdk/blob/5da00e68780dc28bdf0f08a0971ee22e5b223079/substrate/client/transaction-pool/src/lib.rs#L330) return directly when we build a new block. Currently we don't start the maintenance as we don't have seen the block import yet. A consequence of running maintenance for every block could be that we invalidate transactions on one fork, but they are not yet included in another fork. I don't think that the transaction pool currently supports "fork handling". We should look into this when working on this feature.

rphmeier:
It's also worth noting that collators will build forks deliberately, if there are relay chain forks.

And now, with asynchronous backing, may build on top of blocks which themselves aren't the best (and sometimes multiple blocks immediately in a row).

I'm not really familiar with the details of the transaction pool but it sounds like the solution here should work for those requirements as well.

michalkucharczyk:
> (we should check the new RPC spec, but I doubt that it is still send there)

Forgive dumb question, where is new spec? Is it [here](https://github.com/paritytech/polkadot-sdk/blob/2c8021f998a4b3f3e9c0416c7886ef05803aba9d/substrate/client/rpc-spec-v2/) and [here](https://github.com/paritytech/json-rpc-interface-spec)?

bkchr:
Yes this is the new spec. I already opened a proper issue here: https://github.com/paritytech/json-rpc-interface-spec/issues/87

bkchr:
[...]

PR #4561, opened 2024-05-23
txpool refactor: support multiple implementations via trait object

michalkucharczyk:
This pull request slightly refactors the transaction pool crate, laying the groundwork for a fork-aware implementation (which will be provided as separate PR). The main goal of this refactor is to allow  multiple implementations of transaction-pool. This PR does not change any logic yet, it is intended to make the review of the fork-aware transaction pool PR easier.

Key changes include:

1. **Transition to `TransactionPoolImpl` Trait Object**:
   - Replaced `sc_transaction_pool::FullPool` with `sc_transaction_pool::TransactionPoolImpl`.
   - Updated references and type definitions across multiple files.

2. **Builder Pattern**:
   - Introduced `TransactionPoolType` enum for different implementations.
   - Created a `Builder` struct for constructing transaction pool trait object.
   - Enhanced `TransactionPoolOptions` for parameter-based initialization.

3. **Modularization and Refactoring**:
   - Moved components that will be re-used by fork-aware implementation (e.g., error handling, enactment state, metrics) to a `common` module.
   - Created a `single_state_txpool` module for the existing implementation of transaction pool.
   - Renamed and moved several files for clarity, adjusting import paths accordingly.

4. **Support for `?Sized` Trait**:
   - Updated struct definitions and methods to support the `?Sized` trait for `TransactionPool`.
   - Applied these updates across various files.

Related to: #1202


davxy (commented):
I see that `?Sized` bound is spread all over the place.
What about adding the bound directly to the definition of the `TransactionPool` and `MaintainedTransactionPool` traits?

polkadot/rpc/src/lib.rs
```diff
@@ -112,7 +112,7 @@ where
 	C::Api: pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>,
 	C::Api: BabeApi<Block>,
 	C::Api: BlockBuilder<Block>,
-	P: TransactionPool + Sync + Send + 'static,
+	P: TransactionPool + Sync + Send + 'static + ?Sized,
```
davxy, line 115:
```suggestion
	P: TransactionPool + 'static + ?Sized,
```
Nit. Already bound by Send+Sync. Probably there are other instance around.

substrate/client/transaction-pool/src/builder.rs
```diff
@@ -0,0 +1,218 @@
...
+			Error = <FullChainApi<Client, Block> as ChainApi>::Error,
+		>,
+{
+}
+
+/// The type alias for a dynamic trait object implementing
+/// `FullClientTransactionPool` with the given `Client` and `Block` types.
+///
+/// This trait object abstracts away the specific implementations of the transaction pool.
+pub type TransactionPoolImpl<Block, Client> = dyn FullClientTransactionPool<Block, Client>;
```
davxy, line 165: Nit. I'm not sure about this name, I'm referring to the `Impl` suffix. But I can't suggest a better name :-)

templates/parachain/node/src/rpc.rs
```diff
@@ -42,7 +42,7 @@ where
 	C::Api: pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>,
 	C::Api: substrate_frame_rpc_system::AccountNonceApi<Block, AccountId, Nonce>,
 	C::Api: BlockBuilder<Block>,
-	P: TransactionPool + Sync + Send + 'static,
+	P: TransactionPool + Sync + Send + 'static + ?Sized,
```
davxy, line 45:
```suggestion
	P: TransactionPool +  'static + ?Sized,
```

substrate/client/transaction-pool/src/builder.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Builder allowing to create specific instance of transaction pool.
+pub struct Builder<Block, Client> {
+	options: TransactionPoolOptions,
+	_phantom: PhantomData<(Client, Block)>,
+}
```
bkchr, line 171: This is not really a builder pattern. Not really sure we need this.

substrate/client/service/src/builder.rs
```diff
@@ -349,7 +349,7 @@ where
 }
 
 /// Parameters to pass into `build`.
-pub struct SpawnTasksParams<'a, TBl: BlockT, TCl, TExPool, TRpc, Backend> {
+pub struct SpawnTasksParams<'a, TBl: BlockT, TCl, TExPool: ?Sized, TRpc, Backend> {
```
bkchr, line 352: All these `?Sized` are "confusing". IMO we should just make `transaction_pool: Arc<dyn TransactionPool>` and thats it.

substrate/client/transaction-pool/src/builder.rs
```diff
@@ -0,0 +1,218 @@
...
+};
+use prometheus_endpoint::Registry as PrometheusRegistry;
+use sc_transaction_pool_api::{LocalTransactionPool, MaintainedTransactionPool};
+use sp_core::traits::SpawnEssentialNamed;
+
+/// The type of transaction pool.
+#[derive(Debug, Clone)]
+pub enum TransactionPoolType {
+	/// Single-state transaction pool
+	SingleState,
```
bkchr, line 35:
```suggestion
	NonForkAware,
```

Maybe this way? Not sure, but `SingleState` is weird. If you keep the name, at least be a little bit more descriptive.

substrate/client/cli/src/params/transaction_pool_params.rs
```diff
@@ -16,8 +16,25 @@
...
-use sc_service::config::TransactionPoolOptions;
+use clap::{Args, ValueEnum};
+use sc_transaction_pool::TransactionPoolOptions;
+
+/// Type of transaction pool to be used
+#[derive(Debug, Clone, Copy, ValueEnum)]
+#[value(rename_all = "kebab-case")]
+pub enum TransactionPoolType {
+	/// Uses a legacy, single-state transaction pool.
+	SingleState,
```
bkchr, line 27: Yeah or just call it `LegacyTxPool`.

bkchr (commented):
I think the approach with making every unsized is not really sexy. If we touch all of this stuff any way, we should directly put transaction pool everywhere as `Arc<dyn TransactionPool>` or whatever.

substrate/client/transaction-pool/src/builder.rs
michalkucharczyk: We need to translate options to instance. I can rename it, but we need some entry point to instantiate.

substrate/client/service/src/builder.rs
michalkucharczyk: Do I understand correctly that you want to remove `TransactionPool` generic type parameter from all generic types?

bkchr: Yes, would probably make it easier.

PR #4564, opened 2024-05-24
Make `on_unbalanceds` work with `fungibles` `imbalances`

muharem:
Make `on_unbalanceds` work with `fungibles` `imbalances`.

The `fungibles` `imbalances` cannot be handled by the default implementation of `on_unbalanceds` from the `OnUnbalanced` trait. This is because the `fungibles` `imbalances` types do not implement the `Imbalance` trait (and cannot with its current semantics). The `on_unbalanceds` function requires only the `merge` function for the imbalance type. In this PR, we provide the `TryMerge` trait, which can be implemented by all imbalance types and make `OnUnbalanced` require it instead `Imbalance`.

### Migration for `OnUnbalanced` trait implementations:
In case if you have a custom implementation of `on_unbalanceds` trait function, remove it's `<B>` type argument.

### Migration for custom imbalance types:
If you have your own imbalance types implementations, implement the `TryMerge` trait for it introduced with this update.
        
The applicability of the `on_unbalanceds` function to fungibles imbalances is useful in cases like - [link](https://github.com/paritytech/polkadot-sdk/blob/3a8e675e9f6f283514c00c14d3d1d90ed5bf59c0/substrate/frame/transaction-payment/asset-conversion-tx-payment/src/payment.rs#L267) from https://github.com/paritytech/polkadot-sdk/pull/4488.


substrate/frame/balances/src/impl_currency.rs
```diff
@@ -132,6 +132,12 @@ mod imbalances {
 		}
 	}
 
+	impl<T: Config<I>, I: 'static> TryMerge for PositiveImbalance<T, I> {
+		fn try_merge(self, other: Self) -> Result<Self, (Self, Self)> {
+			Ok(self.merge(other))
```
ggwpez, line 137: I think conceptually it makes more sense to use `try_merge` from `merge` and just ignore the return value. Then try-merge would fail on overflow instead of saturating.

substrate/frame/support/src/traits/tokens/imbalance.rs
```diff
@@ -182,6 +182,13 @@ pub trait Imbalance<Balance>: Sized + TryDrop + Default {
 	fn peek(&self) -> Balance;
 }
 
+/// Try to merge two imbalances.
+pub trait TryMerge: Sized {
+	/// Consume `self` and an `other` to return a new instance that combines both. Errors with
+	/// Err(self, other) if the imbalances are not identical (e.g. imbalances of different assets).
```
ggwpez, line 188:
```suggestion
	/// Err(self, other) if the imbalances cannot be merged (e.g. imbalances of different assets).
```

substrate/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs
```diff
@@ -35,11 +35,26 @@ pub trait OnUnbalanced<Imbalance: TryDrop> {
...
+					Err((sum, next)) => {
+						Self::on_unbalanced(next);
+						Some(sum)
+					},
+				},
+				None => Some(next),
+			}
+		}
+		if let Some(sum) = sum {
+			Self::on_unbalanced(sum)
```
ggwpez, line 56: Where is the advantage of calling `on_unbalanced` on the merged result vs calling it on every balance individually?

ggwpez (approved):

substrate/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs
muharem: We were merging `tip` and `fee`, to handle both with a single `on_unbalanceds` handler. This call would result a single `deposit` instead of two if the imbalances are identical.

Current open PR with impl of payment handler for the fees, requires two on unbalanced impl, because `on_unbalanceds` cannot handle `fungibles` iterator - https://github.com/paritytech/polkadot-sdk/blob/3a8e675e9f6f283514c00c14d3d1d90ed5bf59c0/substrate/frame/transaction-payment/asset-conversion-tx-payment/src/payment.rs#L267 
pr https://github.com/paritytech/polkadot-sdk/pull/4488

I do not like the breaking change, but also feel like this change needed, since we moving away from `currency` impls and using `fungible` and `fungibles`.

I also wanna have this decided, before I merge that PR to not break it again later.

substrate/frame/balances/src/impl_currency.rs
muharem: Can you explain why? It feels less correct to me to ignore an error from `try_merge` within `merge`. In the scope of `merge` we wont know for sure, that the `try_merge` wont return an error.
We probably fine with saturation since we do not expect it to overflow the max supply type. It may happened only if you create imbalances in some hacky way, and do not use fungibles api.

> Then try-merge would fail on overflow instead of saturating. 

To handle this we would need a separate impl for `try_merge`, since `merge` is infallible

substrate/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs
ggwpez: Okay seems fine to me.

substrate/frame/balances/src/impl_currency.rs
ggwpez: The difference is that we explicitly ignore the error - instead of `merge` pretending to be infallible. We already have messed up TI on production, so i think saturation could happen.  
Anyway, i dont care so much in this case.

prdoc/pr_4564.prdoc
```diff
@@ -0,0 +1,23 @@
...
+    description: |
+        The `fungibles` `imbalances` cannot be handled by the default implementation of `on_unbalanceds` from the `OnUnbalanced` trait. This is because the `fungibles` `imbalances` types do not implement the `Imbalance` trait (and cannot with its current semantics). The `on_unbalanceds` function requires only the `merge` function for the imbalance type. In this PR, we provide the `TryMerge` trait, which can be implemented by all imbalance types and make `OnUnbalanced` require it instead `Imbalance`.
+
+        This introduces a breaking change to the `OnUnbalanced` trait, but it only requires the removal of an unnecessary type parameter.
+
+crates:
+  - name: frame-support
+    bump: major
+  - name: pallet-balances
+    bump: patch
```
ggwpez, line 14:
```suggestion
    bump: minor
```

prdoc/pr_4564.prdoc
```diff
@@ -0,0 +1,23 @@
...
+doc:
+  - audience: Node Operator
+    description: |
+        The `fungibles` `imbalances` cannot be handled by the default implementation of `on_unbalanceds` from the `OnUnbalanced` trait. This is because the `fungibles` `imbalances` types do not implement the `Imbalance` trait (and cannot with its current semantics). The `on_unbalanceds` function requires only the `merge` function for the imbalance type. In this PR, we provide the `TryMerge` trait, which can be implemented by all imbalance types and make `OnUnbalanced` require it instead `Imbalance`.
+
+        This introduces a breaking change to the `OnUnbalanced` trait, but it only requires the removal of an unnecessary type parameter.
+
+crates:
+  - name: frame-support
+    bump: major
```
ggwpez, line 12: Hm the CI suggest minor here, but a public trait was change, so should be major IMHO.

prdoc/pr_4564.prdoc
```diff
@@ -0,0 +1,23 @@
...
+    bump: patch
+  - name: pallet-asset-conversion-tx-payment
+    bump: patch
+  - name: pallet-transaction-payment
+    bump: patch
+  - name: kitchensink-runtime
+    bump: patch
+  - name: polkadot-runtime-common
+    bump: patch
+  - name: parachains-common
```
ggwpez, line 23:
```suggestion
  - name: parachains-common
     bump: minor
```
Looks like we dont require this field yet.

PR #4566, opened 2024-05-24
[pallet_contracts] Add support for transient storage in contracts host functions

smiasojed:
Introduce transient storage, which behaves identically to regular storage but is kept only in memory and discarded after every transaction. This functionality is similar to the `TSTORE` and `TLOAD` operations used in Ethereum.

The following new host functions have been introduced:
`get_transient_storage`
`set_transient_storage`
`take_transient_storage`
`clear_transient_storage`
`contains_transient_storage`
 Note: These functions are declared as `unstable` and thus are not activated.


substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,600 @@
...
+			(false, Some(prev_value)) => WriteOutcome::Overwritten(prev_value.len() as _),
+			(true, Some(prev_value)) => WriteOutcome::Taken(prev_value),
+		})
+	}
+
+	/// Remove a contract, clearing all its storage entries.
+	pub fn remove(&mut self, account: &AccountIdOf<T>) {
+		let account = account.encode();
+		// Remove all account entries.
+		let keys = self.current.keys(&account).collect::<Vec<_>>();
```
smiasojed, line 289: we have to copy all of the keys for the contract. In a corner case, this can take up 70% of MaxTransientStorageLen. We have a margin of 2x, so it should still be safe.
There is also an option not to remove the storage on contract termination, as it will be deleted at the end of the transaction anyway

substrate/frame/contracts/src/lib.rs
```diff
@@ -1235,6 +1248,8 @@ pub mod pallet {
 		DelegateDependencyAlreadyExists,
 		/// Can not add a delegate dependency to the code hash of the contract itself.
 		CannotAddSelfAsDelegateDependency,
+		/// Can not add more data to transient storage.
+		OutOfStorage,
```
athei, line 1252:
```suggestion
		OutOfTransientStorage,
```

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,626 @@
...
+	current: Storage,
+	journal: Journal,
+	// The size of the StorageMeter is limited by the stack depth.
+	meter: StorageMeter<T>,
+	// The size of the checkpoints is limited by the stack depth.
+	checkpoints: Checkpoints,
+}
+
+impl<T: Config> TransientStorage<T> {
+	pub fn new(total_limit: u32, transaction_limit: u32) -> Self {
```
athei, line 223: I suggest passing in the max call depth size instead of the transaction limit. It makes sense to encapsulate the logic here. That also means removing the default derive.

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,626 @@
...
+type Checkpoints = Vec<usize>;
+
+/// Meter entry tracks transaction allocations.
+#[derive(Default, Debug)]
+struct MeterEntry {
+	/// Allocation commited from subsequent transactions.
+	pub nested: u32,
+	/// Allocation made in the current transaction.
+	pub current: u32,
+}
```
athei, line 46: I just realised that with this structure the global memory limit can run out before the limit per frame. This is bad because frames (even from different) contracts can influence the available limit they have. This looks like an attack vector to me.

I think we might need to stick to the per contract (instead of per frame) limit that ethereum uses.

substrate/frame/contracts/src/lib.rs
```diff
@@ -409,6 +410,10 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxDebugBufferLen: Get<u32>;
 
+		/// The maximum length of the transient storage in bytes.
```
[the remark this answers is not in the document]
smiasojed: This one is a global limit for transient storage including keys, values and journal entries. I agree, description shall be extended.

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -1098,6 +1181,88 @@ pub mod env {
...
+			None,
+			true,
+		)? {
+			ctx.adjust_gas(charged, RuntimeCosts::TakeStorage(value.len() as u32));
+			ctx.write_sandbox_output(memory, out_ptr, out_len_ptr, &value, false, already_charged)?;
+			Ok(ReturnErrorCode::Success)
+		} else {
+			ctx.adjust_gas(charged, RuntimeCosts::TakeStorage(0));
+			Ok(ReturnErrorCode::KeyNotFound)
+		}
```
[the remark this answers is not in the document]
smiasojed: It is just a copy-paste of the storage functions we had, with the word 'transient' added to their names.
Benchmarks are also not done yet. Here is still work in progress.

substrate/frame/contracts/src/transient_storage.rs
smiasojed: I was thinking about it, but limit per contract does not help much. Attackers might exploit per-contract limits by deploying multiple contracts, each consuming memory up to their limit. This can still lead to similar attack, just distributed across multiple contracts.
In Ethereum they are limiting the storage writes by gas: 8 gas + 1 gas * len(transient storage mapping)

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,626 @@
...
+use frame_support::DefaultNoBound;
+use sp_runtime::{DispatchError, DispatchResult};
+use sp_std::{
+	collections::btree_map::BTreeMap,
+	mem,
+	ops::Bound::{Included, Unbounded},
+	vec::Vec,
+};
+type MeterEntries = Vec<MeterEntry>;
+type Checkpoints = Vec<usize>;
```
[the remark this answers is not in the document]
smiasojed: Removed

substrate/frame/contracts/src/lib.rs
smiasojed: Updated

smiasojed:

smiasojed:

smiasojed:

substrate/frame/contracts/src/transient_storage.rs
smiasojed: The benchmarks for set_transient_storage show values between 11.3 µs and 10.8 µs, depending on the number of elements in the BTreeMap. For journal_rollback, the values range from 1.8 µs to 1.3 µs. These times are significantly smaller compared to set_storage, where the dbread and dbwrite operations alone cost 125 µs. This indicates that we could charge the rollback cost together with set_transient_storage.

However, the weight function used in Ethereum does not work well in our case. For gas calculations, Ethereum uses a formula of some_const + storage_items * slope. In our case, we could use 131000000 + storage_items * 1000000. But in this scenario (assuming a block weight of 2 seconds), we can allocate 2000 items (https://www.wolframalpha.com/input?i=x%28x-1%29%2F2*1000000+%2B+13100000*x+%3D+2000000000000), which would cause a memory overflow when allocating maximum values.

In a scenario where we allocate maximum values (with a key size of 256 bytes and a value size of 16 KB), we can have around 60 items (including storage data overhead). To approximate this limit, we need to change the slope (https://www.wolframalpha.com/input?i=x%28x-1%29%2F2*1000000000+%2B+13100000*x+%3D+2000000000000). However, in this case, only first set_transient_storage is cheaper than `set_storage` (https://www.wolframalpha.com/input?i=x%28x-1%29%2F2*1000000000+%2B+13100000*x+%3D+140000000).

smiasojed: Another option is to implement a solution where each new frame receives a portion of the remaining transient storage, for example, (15/16). In this case, in the nth subsequent call, we get `free_storage + (15/16)^n`.
https://www.wolframalpha.com/input?i=sum+%28%2815%2F16%29%5En+%29%2C+n%3D1+to+infinity

smiasojed:

substrate/frame/contracts/src/lib.rs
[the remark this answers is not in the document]
smiasojed: Ideally, the limit would only account for keys and values, but there's a risk that someone might allocate only small values. The minimal key size is 16 bytes with a 0-byte value. The data overhead for each transient storage entry is 48 bytes, meaning the overhead is larger than the content itself. Therefore, we should ensure a safe margin of at least 3 times the content size (Btrremap also has overhead, so 2 times the content size might be sufficient there). Overall, a margin of 6 times the content size is necessary, which poses a challenge because we have limited free memory at runtime.

smiasojed:

smiasojed:

smiasojed:

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -791,10 +821,129 @@ impl<'a, E: Ext + 'a> Runtime<'a, E> {
 		let key = self.decode_key(memory, key_type, key_ptr)?;
 		let outcome = self.ext.get_storage_size(&key);
 
-		self.adjust_gas(charged, RuntimeCosts::ClearStorage(outcome.unwrap_or(0)));
```
smiasojed, line 794: Seems to be a bug. Can it be fixed in this commit, or should it be addressed in a separate one

substrate/frame/contracts/uapi/src/host/wasm32.rs
```diff
@@ -650,6 +709,23 @@ impl HostFn for HostFnImpl {
 		ret_code.into()
 	}
 
+	#[inline(always)]
```
smiasojed, line 712: Is there any rule for when we should annotate a function as inline here?

smiasojed:

substrate/frame/contracts/src/exec.rs
```diff
@@ -209,6 +210,27 @@ pub trait Ext: sealing::Sealed {
 		take_old: bool,
 	) -> Result<WriteOutcome, DispatchError>;
 
+	/// Returns the transient storage entry of the executing account by the given `key`.
```
pgherveou, line 213:
```suggestion
	/// Returns the transient storage entry of the executing account for the given `key`.
```

substrate/frame/contracts/src/exec.rs
```diff
@@ -209,6 +210,27 @@ pub trait Ext: sealing::Sealed {
...
+	/// was deleted.
+	fn get_transient_storage(&self, key: &Key<Self::T>) -> Option<Vec<u8>>;
+
+	/// Returns `Some(len)` (in bytes) if a transient storage item exists at `key`.
+	///
+	/// Returns `None` if the `key` wasn't previously set by `set_transient_storage` or
+	/// was deleted.
+	fn get_transient_storage_size(&self, key: &Key<Self::T>) -> Option<u32>;
+
+	/// Sets the transient storage entry by the given key to the specified value. If `value` is
```
pgherveou, line 225:
```suggestion
	/// Sets the transient storage entry for the given key to the specified value. If `value` is
```

substrate/frame/contracts/src/exec.rs
```diff
@@ -1046,6 +1080,12 @@ where
 			Err(error) => (false, Err(error.into())),
 		};
 
+		if success {
+			self.transient_storage.commit_transaction();
+		} else {
+			self.transient_storage.rollback_transaction();
+		}
+
```
pgherveou, line 1088: can we adopt an API similar to the transaction API here, wrapping the code in a closure, that handle start_transaction & comit or rollback, this should make it easier not to introduce bug when we refactor this code.

substrate/frame/contracts/src/exec.rs
```diff
@@ -3826,6 +3889,262 @@ mod tests {
...
+			// Overwrite
+			assert_eq!(
+				ctx.ext.set_transient_storage(&Key::Fix([1; 32]), Some(vec![42]), false),
+				Ok(WriteOutcome::Overwritten(3))
+			);
+			assert_eq!(
+				ctx.ext.set_transient_storage(&Key::Fix([2; 32]), Some(vec![48]), true),
+				Ok(WriteOutcome::Taken(vec![4, 5, 6]))
+			);
```
pgherveou, line 3929: I think it would read better if you group write and the overwrite, instead of doing all the write then the overwrite, it helps make sure we don't have duplicate or missing scenario as well.

substrate/frame/contracts/src/wasm/mod.rs
```diff
@@ -3046,6 +3068,337 @@ mod tests {
 		assert_eq!(&result.data[4..], &[0u8; 0]);
 	}
 
+	#[test]
```
pgherveou, line 3071: 🤔  @athei at some point,  shouldn't we translate these code examples to Rust like we did with the other ones

substrate/frame/contracts/uapi/src/host/wasm32.rs
pgherveou: my approach is to only add it if this make a significant diff in the benchmarks

pgherveou (commented):
Looks good, haven't reviewed the tests yet

substrate/frame/contracts/src/exec.rs
smiasojed: @pgherveou do you mean something like this:
```
		self.transient_storage.with_transaction(|| ->TransactionOutcome<Result<_, DispatchError>> {
			storage::with_transaction(|| -> TransactionOutcome<Result<_, DispatchError>> {
				let output = do_transaction();
```
With this, I encountered issues with the borrow checker. The issue is with `do_transaction()`closure which borrows `self` .

substrate/frame/contracts/src/tests.rs
```diff
@@ -962,6 +962,59 @@ fn storage_max_value_limit() {
...
+		let addr_caller = builder::bare_instantiate(Code::Upload(wasm_caller))
+			.value(min_balance * 100)
+			.build_and_unwrap_account_id();
+		let addr_callee = builder::bare_instantiate(Code::Upload(wasm_callee))
+			.value(min_balance * 100)
+			.build_and_unwrap_account_id();
+
+		// The transient storage limit is set to 4KB.
+		assert_ok!(builder::call(addr_caller.clone())
+			.data((1_000u32, 2_000u32, &addr_callee).encode())
```
pgherveou, line 999: maybe put 1_000 and 2_000 behind variable name, is not really straight forward what they mean reading this. 
You could set the value once and just change the MaxTransientStorageSize setting as well

substrate/frame/contracts/src/tests.rs
```diff
@@ -962,6 +962,59 @@ fn storage_max_value_limit() {
...
+
+		// Create both contracts: Constructors do nothing.
+		let addr_caller = builder::bare_instantiate(Code::Upload(wasm_caller))
+			.value(min_balance * 100)
+			.build_and_unwrap_account_id();
+		let addr_callee = builder::bare_instantiate(Code::Upload(wasm_callee))
+			.value(min_balance * 100)
+			.build_and_unwrap_account_id();
+
+		// The transient storage limit is set to 4KB.
```
pgherveou, line 997: Maybe mutate MaxTransientStorageSize setting here so that this comment does not become invalid if that is refactored in the default config later on

substrate/frame/contracts/src/exec.rs
smiasojed: I used the same approach as is used for the `set_storage` call. If we want to change it, I need to modify it in three places.

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -1162,6 +1163,300 @@ mod benchmarks {
...
+	#[benchmark(pov_mode = Ignored)]
+	fn set_transient_storage_empty() -> Result<(), BenchmarkError> {
+		let max_value_len = T::Schedule::get().limits.payload_len;
+		let max_key_len = T::MaxStorageKeyLen::get();
+		let key = Key::<T>::try_from_var(vec![0u8; max_key_len as usize])
+			.map_err(|_| "Key has wrong length")?;
+
+		let mut setup = CallSetup::<T>::default();
+		let (mut ext, _) = setup.ext();
+		let mut runtime = crate::wasm::Runtime::new(&mut ext, vec![]);
```
pgherveou, line 1175: nit can't we just use the macro here `build_runtime!(runtime, memory: [ ]);` same apply below I believe

substrate/frame/contracts/src/wasm/mod.rs
athei: Yeah we can just remove those tests. It is just boilerplate and everything is testet in the fixtures anyways.

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,693 @@
...
+			if let Some(value) = &value {
+				// Charge the key, value and journal entry.
+				// If a new value is written, a new journal entry is created. The previous value is
+				// moved to the journal along with its key, and the new value is written to
+				// storage.
+				let key_len = key.capacity();
+				let mut amount = value
+					.capacity()
+					.saturating_add(key_len)
+					.saturating_add(mem::size_of::<JournalEntry>());
```
athei, line 258: Why are you using `capacity` and not `len`?

smiasojed: I used it to track reserved memory to ensure safety, but assuming the lazy memory allocation model in WASM —where physical memory is not committed until it is actually used—using len might be sufficient.

substrate/frame/contracts/src/benchmarking/mod.rs
smiasojed: Yes, we can do that. There will be some additional, unnecessary calls, but the construction will be simpler.

prdoc/pr_4566.prdoc
```diff
@@ -0,0 +1,23 @@
...
+      but is kept only in memory and discarded after every transaction.
+      This functionality is similar to the `TSTORE` and `TLOAD` operations used in Ethereum.
+      The following new host functions have been introduced: `get_transient_storage`,
+      `set_transient_storage`, `take_transient_storage`, `clear_transient_storage` and
+      `contains_transient_storage`.
+      These functions are declared as unstable and thus are not activated.
+
+crates:
+  - name: pallet-contracts
+    bump: minor
```
athei, line 19: Shouldn't this be major as we are adding something to the `Config` type? It will break downstream crates.

substrate/frame/contracts/src/transient_storage.rs
athei: We are diving deep into operating system specifics here to know when this memory is actually faulted in. I guess we leave it with capacity just to be sure.

prdoc/pr_4566.prdoc
smiasojed: Yes, it should. Thanks

substrate/frame/contracts/src/tests.rs
smiasojed: Done

substrate/frame/contracts/src/tests.rs
smiasojed: Done

prdoc/pr_4566.prdoc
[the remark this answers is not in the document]
smiasojed: Done

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,693 @@
...
+		&mut self,
+		account: &AccountIdOf<T>,
+		key: &Key<T>,
+		value: Option<Vec<u8>>,
+		take: bool,
+	) -> Result<WriteOutcome, DispatchError> {
+		let prev_value = self.read(account, key);
+		// Skip if the same value is being set.
+		if prev_value != value {
+			let key = Self::storage_key(&account.encode(), &key.hash());
```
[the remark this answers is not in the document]
smiasojed: Fixed

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,693 @@
...
+
+		self.nested_meters.push(MeterEntry::new(transaction_limit));
+	}
+
+	/// Commit a transaction meter.
+	pub fn commit(&mut self) {
+		let transaction_meter = self
+			.nested_meters
+			.pop()
+			.expect("There is no nested meter that can be committed.");
```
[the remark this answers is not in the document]
smiasojed: Updated

substrate/frame/contracts/src/lib.rs
```diff
@@ -223,7 +224,7 @@ pub struct Environment<T: Config> {
 pub struct ApiVersion(u16);
 impl Default for ApiVersion {
 	fn default() -> Self {
-		Self(4)
+		Self(5)
```
athei, line 227: Only bumped when the stable API gets extended.

```suggestion
		Self(4)
```

substrate/frame/contracts/src/wasm/runtime.rs
athei: Let's leave it in here.

substrate/frame/contracts/src/transient_storage.rs
```diff
@@ -0,0 +1,690 @@
...
+			return Err(Error::<T>::OutOfTransientStorage.into());
+		}
+		meter.amount.saturating_accrue(amount);
+		Ok(())
+	}
+
+	/// Revert a transaction meter.
+	fn revert(&mut self) {
+		self.nested_meters.pop().expect(
+			"A call to revert a meter must be preceded by a corresponding call to start a meter.",
```
athei, line 86: Almost

```suggestion
			"A call to revert a meter must be preceded by a corresponding call to start a meter. The code within this crate makes sure that this is always the case; qed",
```

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -1162,6 +1163,300 @@ mod benchmarks {
...
+			);
+		}
+
+		assert_eq!(result, Ok(WriteOutcome::New));
+		assert_eq!(runtime.ext().get_transient_storage(&key), Some(vec![42u8; max_value_len as _]));
+		Ok(())
+	}
+
+	#[benchmark(pov_mode = Ignored)]
+	fn set_transient_storage_full() -> Result<(), BenchmarkError> {
```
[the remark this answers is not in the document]
smiasojed: Added

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -1162,6 +1163,300 @@ mod benchmarks {
...
+		let mut setup = CallSetup::<T>::default();
+		let (mut ext, _) = setup.ext();
+		let mut runtime = crate::wasm::Runtime::new(&mut ext, vec![]);
+		runtime.ext().transient_storage().meter().current_mut().limit = u32::MAX;
+		let result;
+		#[block]
+		{
+			result = runtime.ext().set_transient_storage(
+				&key,
+				Some(vec![42u8; max_value_len as _]),
```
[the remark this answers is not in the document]
smiasojed: Moved

substrate/frame/contracts/src/transient_storage.rs
smiasojed: Done

substrate/frame/contracts/src/benchmarking/mod.rs
```diff
@@ -1162,6 +1163,296 @@ mod benchmarks {
...
+	#[benchmark(pov_mode = Ignored)]
+	fn set_transient_storage_full() -> Result<(), BenchmarkError> {
+		let max_value_len = T::Schedule::get().limits.payload_len;
+		let max_key_len = T::MaxStorageKeyLen::get();
+		let key = Key::<T>::try_from_var(vec![0u8; max_key_len as usize])
+			.map_err(|_| "Key has wrong length")?;
+		let value = Some(vec![42u8; max_value_len as _]);
+		let mut setup = CallSetup::<T>::default();
+		let (mut ext, _) = setup.ext();
+		CallSetup::<T>::with_transient_storage(&mut ext, T::MaxTransientStorageSize::get())?;
```
pgherveou, line 1201: would be a bit nicer to have `with_transient_storage` be a method of the builder

```rs
let mut setup = CallSetup::<T>::default();
setup.transient_storage_size(T::MaxTransientStorageSize::get())
let (mut ext, _) = setup.ext();
```

substrate/frame/contracts/src/benchmarking/call_builder.rs
```diff
@@ -160,6 +160,31 @@ where
 		)
 	}
 
+	/// Add transient_storage
+	pub fn with_transient_storage(ext: &mut StackExt<T>, size: u32) -> Result<(), &'static str> {
+		let amount = ext.transient_storage().meter().current().amount;
+		let limit = ext.transient_storage().meter().current().limit;
```
pgherveou, line 166: maybe
```rust
let MeterEntry { amount, limit } = ...
```

substrate/frame/contracts/src/wasm/runtime.rs
```diff
@@ -244,6 +254,30 @@ pub enum RuntimeCosts {
 	UnlockDelegateDependency,
 }
 
+macro_rules! cost_write {
+	// cost_write!(name, a, b, c) -> T::WeightInfo::name(a, b, c).saturating_add(T::WeightInfo::rollback_transient_storage())
+	// .saturating_add(T::WeightInfo::set_transient_storage_full().saturating_sub(T::WeightInfo::set_transient_storage_empty())
```
pgherveou, line 259: Not clear to me why the cost is full - empty, can we explain the reasoning here

pgherveou (approved):
lgtm as well, added a few more nits and a question

substrate/frame/contracts/uapi/src/host.rs
```diff
@@ -292,6 +292,17 @@ pub trait HostFn {
...
+	/// Clear the value at the given key in the contract transient storage.
+	///
+	/// # Parameters
+	///
+	/// - `key`: The storage key.
+	///
+	/// # Return
+	///
+	/// Returns the size of the pre-existing value at the specified key if any.
+	fn clear_transient_storage(key: &[u8]) -> Option<u32>;
```
xermicus, line 304: Should those be marked as unstable here?

xermicus (approved):
LGTM

substrate/frame/contracts/uapi/src/host.rs
athei: Yes. I missed that. They should use the `deprecated` attribute to warn users about their instability.

athei:
Can you do one additional thing in this PR: Since we are bumping the major version of the uapi crate it would be a good time to also seal the `HostFn` trait. The only reason we have to do a major bump is because we forgot to prevent downstream crates to implement it.

substrate/frame/contracts/uapi/src/host.rs
smiasojed: done

substrate/frame/contracts/src/benchmarking/mod.rs
smiasojed: done

smiasojed:

substrate/frame/contracts/src/wasm/runtime.rs
smiasojed: `full - empty` represents the weight of adding an item to a full BTreeMap. All transient storage host function benchmarks are done with an empty transient storage (BTreeMap), so this represents the worst-case scenario.

pgherveou: follow up question, should we refund the cost of rollback_transient_storage in the happy path since you pre-charge it

smiasojed: The clear_transient_storage function modifies the storage. The benchmark is currently done for one item in the BTreeMap, but we would like to consider the worst-case scenario, where the modification is done on a fully filled BTreeMap.

I am treating it as a constant cost, similar to DBWrite, which is 100µs.

pgherveou: Make sense then, we should add that to the comment to make it clearer

smiasojed: Added

Issue #4565, opened 2024-05-24
Para Owner cannot unlock and deregister a Parachain when downgraded to Parathread


chrisdcosta:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

After a parachain has been downgraded to a parathread, the para owner (manager) should be able to deregister the parathread if so desired. The code hints that this should be possible however in practice there is a logic trap:

* A lock is applied to the parachain when it starts to produce blocks that is not lifted when the chain is downgraded.
* A parathread cannot unlock itself because it no longer produces blocks.
* The para owner is not allowed to remove the lock.
* A lock can therefore only be removed by Root.
* There is no appropriate governance track to unlock a parathread.

This appears to be a bug following [this question](https://substrate.stackexchange.com/questions/11485/what-is-the-opengov-track-required-to-deregister-a-parathread/11488#11488) on stack exchange.

The affected file is found at [polkadot-sdk/polkadot/runtime/common/src/paras_registrar/mod.rs](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/common/src/paras_registrar/mod.rs)

The context is that the para owner should not be able to remove a lock from a running parachain, because that is not decentralised. I agree with this.

However I would also argue that when a parachain is down-graded it may need to be unlocked to either perform a parachain swap or to de-register - either of those case involve some centralised decision making and therefore the para owner / manager must be given the power to unlock the parathread.

### Steps to reproduce

This can be performed on any non-production relaychain:

1. Register a parachain ID with an owner.

2. Add the PVF.

3. Upgrade to parachain for 1 lease period.

4. When the parachain automatically downgrades to a parathread following the end of the lease, attempt to de-register.

The Parathread cannot be de-registered due to a lock

5. Attempt to remove the lock by the para owner

The Para owner will get a BadOrigin error.

xlc:
It will be a centralization risk if parathread/on-demand parachain owner can just remove it without any restrictions. I will say this is inteded behaviour and there isn't much we can do to improve without introducing new risks.

bkchr:
@chrisdcosta you beat me to create an issue for this ;) It was on my list!

[...]

PR #4568, opened 2024-05-24
Proposed fix to issue where a para owner is unable to de-register a parathread #4565

chrisdcosta:
This is a proposal for issue #4565  with some additional minor improvements.

The logic of this proposal is that it provides some additional checks and balances as to the circumstances of the removal of a lock so that it only occurs when a parachain has become downgraded and effectively cannot be upgraded in any way or deregistered. Additionally a more robust set of tests are performed to support this proposal.

This proposal exists in the absence of any reasonable governance track to allow the unlocking of a parathread for maintenance or de-registration.

* a new function `ensure_root_or_para_or_owner()` to determine only three allowed callers, returning a caller type or error

* renamed `ensure_root_para_or_owner()` to `ensure_unlocked_root_para_or_owner()` 

* implemented more robust test in `para_lock_works()`

* added getter functions to storage

* a new event `Unlocked`

* new error type `ParaNotOwner`


polkadot/runtime/common/src/paras_registrar/mod.rs
```diff
@@ -357,12 +373,29 @@ pub mod pallet {
...
+			match Self::ensure_root_or_para_or_owner(origin.clone(), id) {
+				Ok(ParaOriginType::Root) | Ok(ParaOriginType::Para) => {
+					// The origin is root or para, no action needed
+				}
+				Ok(ParaOriginType::Manager) => {
+					// The origin is the para manager, check if the chain is a parathread
+					let id_lifecycle = paras::Pallet::<T>::lifecycle(id).ok_or(Error::<T>::NotRegistered)?;
+					if id_lifecycle != ParaLifecycle::Parathread {
+						return Err(Error::<T>::NotParathread.into());
+					}
```
bkchr, line 390: In the near future, with Coretime all chains will be "LifeCycle::Parathread". Basically this lifecycle stuff gets removed any way and this is not a good idea to introduce this change here.

bkchr (changes requested):

polkadot/runtime/common/src/paras_registrar/mod.rs
chrisdcosta: Then this should be part of a greater review of the `para_registrar` functionality at that time - because there are some complications implied by having only one lifecycle type not just for the lock, but for swapping also. 

As for existing parachains, as far as I know, have no way to force a downgrade to become a parathread during their lease periods.  In that case they have to live out their lifecycles which could still be nearly two years from now and what would happen to the locked funds in that case?

Ok though, assuming that "all chains will be 'LifeCycle::Parathread'" then the swap mechanism becomes irrelevant, and the para manager role has little sense either since it will have no meaningful utility - which is at odds with putting up funds. Which is why I think this at least has some utility in the interim - simply because there is a lot to refactor and it will become technical debt in the end.

chrisdcosta: @bkchr Are you suggesting I remove the code you've selected from the PR? I can of course do that and adjust the unit test, but if I do - then this will permit the para owner to unlock without restriction. Is this the intention?

Captain1653:
Hi. 

I am reviewing old pull requests and noticed that this one is no longer relevant.

**Architectural Shift**: With the introduction of on-demand coretime, the underlying issue is resolved because a chain can  purchase blockspace to execute necessary operations—such as deregistration.

Please, see [this comment](https://github.com/paritytech/polkadot-sdk/issues/4565#issuecomment-4080473965).


It makes sense to close both this pull request and the original issue as irrelevant.

PR #3151, opened 2024-01-31
Collective: dynamic deposit based on number of proposals

muharem:
Introduce a dynamic proposal deposit mechanism influenced by the total number of active proposals, with the option to set the deposit to none.

The potential cost (e.g., balance hold) for proposal submission and storage is determined by the implementation of the `Consideration` trait. The footprint is defined as `proposal_count`, representing the total number of active proposals in the system, excluding the one currently being proposed. This cost may vary based on the proposal count. The pallet also offers various types to define a cost strategy based on the number of proposals.

Two new calls are introduced:
- kill(origin, proposal_hash): the cancellation of a proposal, accompanied by the burning of the associated cost/consideration ticket.
- release_proposal_cost(origin, proposal_hash): the release of the cost for a non-active proposal.

Additionally change: 
- benchmarks have been upgraded to benchmarks::v2 for collective pallet;
- `ensure_successful` function added to the `Consideration` under `runtime-benchmarks` feature.

PR #4596, opened 2024-05-27
Frame: `Consideration` trait generic over `Footprint` and indicates zero cost

muharem:
`Consideration` trait generic over `Footprint` and indicates zero cost for a give footprint.

`Consideration` trait is generic over `Footprint` (currently defined over the type with the same name). This makes it possible to setup a custom footprint (e.g. current number of proposals in the storage).

`Consideration::new` and `Consideration::update` return an `Option<Self>` instead `Self`, this make it possible to indicate a no cost for a specific footprint (e.g. if current number of proposals in the storage < max_proposal_count / 2 then no cost).

These cases need to be handled for https://github.com/paritytech/polkadot-sdk/pull/3151


substrate/frame/support/src/traits/storage.rs
```diff
@@ -206,16 +206,20 @@ where
...
-pub trait Consideration<AccountId>: Member + FullCodec + TypeInfo + MaxEncodedLen {
+pub trait Consideration<AccountId, Footprint>:
+	Member + FullCodec + TypeInfo + MaxEncodedLen
+{
 	/// Create a ticket for the `new` footprint attributable to `who`. This ticket *must* ultimately
-	/// be consumed through `update` or `drop` once the footprint changes or is removed.
-	fn new(who: &AccountId, new: Footprint) -> Result<Self, DispatchError>;
+	/// be consumed through `update` or `drop` once the footprint changes or is removed. `None`
+	/// implies no cost for a given footprint.
+	fn new(who: &AccountId, new: Footprint) -> Result<Option<Self>, DispatchError>;
```
muharem, line 215: `Self` can also say for itself that there is no cost (with 0 balance for example), but this wont let a pallet to have no storage entry for such case.

gupnik, line 215: This seems to be bleeding the implementation details outside. Basically, the caller should know that the consideration being None means that the storage cost was 0. 

Instead, shouldn't the consideration expose an API to check whether it has some underlying storage or not?

muharem: > This seems to be bleeding the implementation details outside. Basically, the caller should know that the consideration being None means that the storage cost was 0.

Does not the client already know that the ticket is about the storage cost? If you check the doc for this trait, it does expect a ticket to carry an information about a cost. Or I misunderstand what you mean. 

> Instead, shouldn't the consideration expose an API to check whether it has some underlying storage or not?

Not sure I understand this part. Does not the client of the trait the one how knows about a storage? What kind of API you mean, can you please give an example.

My example is:

``` rust

type Proposals = StorageMap<...>;
type ProposalDeposit = StorageMap<.., ProposalIndex, T::Consideration::Ticket...>;
...
pub fn submit_proposal(...) {
  ..
  // for example if proposal_count < 10, no cost, if greater some cost.
  let ticket = T::Consideration::new(proposal_count);
  // if ticket cannot be none, we have to always save a ticket with zero deposit.
  ProposalDeposit::insert(proposal_index, ticket);
}
```

Not bid deal if the ticket itself can mean a zero cost (with zero balance for example), but a none option looks better to me.

gupnik: > Does not the client already know that the ticket is about the storage cost? If you check the doc for this trait, it does expect a ticket to carry an information about a cost. Or I misunderstand what you mean.

So, the caller already has access to the footprint and hence, can already know that it doesn't require a ticket. However, we still try to obtain a consideration for it. In this case, it returns None which is then used to signify that the ticket doesn't need to be saved. May be I misunderstood something, but this sounds like a round about way to achieve something that could be done at the first step itself? I mean what's the point of trying to obtain a Consideration here?

muharem: @gupnik the client (e.g. a pallet) of the `Consideration` does not know if there is a cost for a certain `footprint`. The `Consideration` provider (e.g. a pallet's client) defines the rules.

gupnik: > @gupnik the client (e.g. a pallet) of the `Consideration` does not know if there is a cost for a certain `footprint`. The `Consideration` provider (e.g. a pallet's client) defines the rules.

Ahh, makes sense! Thank you.

substrate/frame/support/src/traits/tokens/fungible/mod.rs
```diff
@@ -330,22 +363,34 @@ impl<
...
-	fn update(self, who: &A, footprint: Footprint) -> Result<Self, DispatchError> {
-		F::set_on_hold(&R::get(), who, D::convert(footprint)).map(|_| Self(PhantomData))
+	fn update(self, who: &A, footprint: Fp) -> Result<Option<Self>, DispatchError> {
+		let new = D::convert(footprint);
+		let _ = F::set_on_hold(&R::get(), who, new)?;
+		if new.is_zero() {
+			Ok(None)
+		} else {
+			Ok(Some(Self(PhantomData)))
+		}
```
gupnik, line 393:
```suggestion
		if new.is_zero() {
			Ok(None)
		} else {
			let _ = F::set_on_hold(&R::get(), who, new)?;
			Ok(Some(Self(PhantomData)))
		}
```

substrate/frame/support/src/traits/tokens/fungible/mod.rs
```diff
@@ -291,22 +312,34 @@ impl<
...
-	fn update(self, who: &A, footprint: Footprint) -> Result<Self, DispatchError> {
-		Fx::set_frozen(&Rx::get(), who, D::convert(footprint), Polite).map(|_| Self(PhantomData))
+	fn update(self, who: &A, footprint: Fp) -> Result<Option<Self>, DispatchError> {
+		let new = D::convert(footprint);
+		let _ = Fx::set_frozen(&Rx::get(), who, new, Polite)?;
+		if new.is_zero() {
+			Ok(None)
+		} else {
+			Ok(Some(Self(PhantomData)))
+		}
```
gupnik, line 342:
```suggestion
		if new.is_zero() {
			Ok(None)
		} else {
			let _ = Fx::set_frozen(&Rx::get(), who, new, Polite)?;
			Ok(Some(Self(PhantomData)))
		}
```

gupnik (approved):

substrate/frame/support/src/traits/tokens/fungible/mod.rs
gupnik: Got it. Could we please add a test for this?

muharem: I have added the tests. But again, I have not changed the behaviour here, and `set_frozen` is supposed to be invoked for `0` cost. 
From doc
```
/// NOTE: This is an optimized implementation, which can only be used for systems where each
/// account has only a single active ticket associated with it since individual tickets do not
/// track the specific balance which is frozen. If you are uncertain then use `FreezeConsideration`
/// instead, since this works in all circumstances.
```

substrate/frame/preimage/src/lib.rs
```diff
@@ -122,7 +122,9 @@ pub mod pallet {
 		type ManagerOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 
 		/// A means of providing some cost while data is stored on-chain.
-		type Consideration: Consideration<Self::AccountId>;
+		///
+		/// Should never return a `None`, implying no cost for a non-empty preimage.
```
ggwpez, line 126: Why should it not do that? The implementor of the `Consideration` should be able to return as they see fit.  
Otherwise it would probably need a `Consideration` and `MaybeConsideration` trait to disambiguate.

muharem: Yes, I agree. But I did now want to change the behaviour of this pallet, and make only a patch. Otherwise we need a migration for the preimage pallet.
Maybe I should rephrase it, that we do expect some cost for any non zero footprint, so it should not return none.

ggwpez (approved):

muharem:
@ggwpez we can also drop the option, and make it always return some ticket. but in this case I think we will eventually have to bound the ticket with some additional trait, so that the caller can check if it implies a zero cost.
less likely we will never care, and persist zero balance or none ticket.

PR #4598, opened 2024-05-27
Remove workspace lints from templates

rzadp:
This detaches the templates from monorepo's workspace lints, so the lints for the templates can evolve separately as needed.

Currently the templates [re-use the monorepo's lints](https://github.com/paritytech/polkadot-sdk-minimal-template/blob/bd8afe66ec566d61f36b0e3d731145741a9e9e19/Cargo.toml#L16-L43) which looks weird.

cc @kianenigma @gupnik

### DIFF
--- a/.github/workflows/misc-sync-templates.yml
+++ b/.github/workflows/misc-sync-templates.yml
@@ -105,8 +105,6 @@ jobs:
           toml set templates/${{ matrix.template }}/Cargo.toml 'workspace.package.edition' "$(toml get --raw Cargo.toml 'workspace.package.edition')" > Cargo.temp
           mv Cargo.temp ./templates/${{ matrix.template }}/Cargo.toml
 
-          toml get Cargo.toml 'workspace.lints' --output-toml >> ./templates/${{ matrix.template }}/Cargo.toml
-
           toml get Cargo.toml 'workspace.dependencies' --output-toml >> ./templates/${{ matrix.template }}/Cargo.toml
         working-directory: polkadot-sdk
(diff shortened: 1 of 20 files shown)


bkchr:
> Currently the templates [re-use the monorepo's lints](https://github.com/paritytech/polkadot-sdk-minimal-template/blob/bd8afe66ec566d61f36b0e3d731145741a9e9e19/Cargo.toml#L16-L43) which looks weird.

Why does it looks weird? IMO the problem is that this then maybe breaks clippy for them because there is maybe code that we accept, but default clippy rejects.

rzadp:
> Why does it looks weird?

Actually it's the comments that look weird and out of place (to me).
Comments like `One case where we do 0 +` - it becomes untrue in the individual template repositories, because it does not happen in the templates.

Perhaps we should keep the lints, but strip the comments?

---

> there is maybe code that we accept, but default clippy rejects.

That's not the case, because with this change, the templates are processed with default clippy inside the monorepo as well - as evidenced by [this red check above](https://github.com/paritytech/polkadot-sdk/pull/4598#issuecomment-2133353515).

So if the clippy job is green in the monorepo, it will be green in individual repos as well - unless I'm missing something.

bkchr:
> So if the clippy job is green in the monorepo, it will be green in individual repos as well - unless I'm missing something.

Yeah good point, didn't thought about this.

bkchr (approved):

gupnik (approved):

kianenigma (approved):

PR #4599, opened 2024-05-27
Filter workspace dependencies in the templates

rzadp:
This detaches the templates from monorepo's workspace dependencies.

Currently the templates [re-use the monorepo's dependencies](https://github.com/paritytech/polkadot-sdk-minimal-template/blob/bd8afe66ec566d61f36b0e3d731145741a9e9e19/Cargo.toml#L45-L58), most of which are not needed.

The simplest approach is to specify versions directly and not use workspace dependencies in the templates.

Another approach would be to programmatically filter dependencies that are actually needed - but not sure if it's worth it, given that it would complicate the synchronization job.

cc @kianenigma @gupnik


alvicsam (approved):

bkchr:
> Another approach would be to programmatically filter dependencies that are actually needed - but not sure if it's worth it, given that it would complicate the synchronization job.

Yes please do this. Using workspace dependencies is that much easier for the user etc.

EgorPopelyaev (approved):

rzadp:
> Yes please do this. Using workspace dependencies is that much easier for the user etc.

Ok, turned the PR into a draft and will look into it.

kianenigma (approved):

ggwpez (approved):

PR #4604, opened 2024-05-27
Port paged-nmap changes

ggwpez:
Port of https://github.com/paritytech/substrate/pull/14747

This MR implements `StorageList` and `StoragePagedNMap`. Only the latter contains code, the former is an adapter type of it. Alike it should be possible to use it as `StoragePagedDoubleMap` and `StoragePagedMap`. All variants are counted, so the Counted prefix is omitted.

## Metadata

It uses fudged metadata by setting the metadata of a `StoragePagedList<Value>` to the metadata of a `StorageMap<u32, Vec<Value>>`. This is correct on the storage level, since it maps the page-index (`u32`) to a page (`Vec<Value>`). But when reading the page, it is unclear to the reader what offset to use within the page to avoid reading removed values.  
FRAME metadata V17 would be needed to fix this properly so that the reader is aware of the metadata field in storage.


ggwpez:
Closing as ultra low prio.

Issue #2997, opened 2024-01-19
[pallet_balances] Align `FreezeIdentifier`, `RuntimeFreezeReason` and `MaxFreezes`


bkontur:
Continuation of https://github.com/paritytech/polkadot-sdk/pull/2657 for `RuntimeHoldReason` where `MaxHolds` are removed.

## Summary

The `pallet_balances` has storage item [`Freezes`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/lib.rs#L489-L495). They use  [`T::FreezeIdentifier`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/lib.rs#L307) as identifier, which **is not** meant to be a **composite_enum**. There is one usage of `FreezeIdentifier` for implementing `InspectFreeze` [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/impl_fungible.rs#L294-L295). However, we also have [`RuntimeFreezeReason`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/lib.rs#L263) here, which **is not used anywhere**.

For example, I see one usage of freezes in `pallet_nomination_pools` [here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/nomination-pools/src/lib.rs#L1611-L1612), where the composite\_enum `FreezeReason` (evaluated as `RuntimeFreezeReason`) is used as a `FreezeIdentifier`.

In the [Westend runtime](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/westend/src/lib.rs#L304-L305), it is configured as:

```
type RuntimeFreezeReason = RuntimeFreezeReason`
type FreezeIdentifier = RuntimeFreezeReason`
```

There is also an [`integrity_test`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/lib.rs#L559-L563) that verifies `RuntimeFreezeReason::VARIANT_COUNT` against `T::MaxFreezes`. 

And it seems that `RuntimeFreezeReason::VARIANT_COUNT` is kind of disconnected from `FreezeIdentifier`.

According to the [original implementation](https://github.com/paritytech/substrate/pull/12951): 
> Both Holds and Freezes require an identifier which is configurable and is expected to be an enum aggregated across all pallet instances of the runtime.

## Questions:

[...]

PR #4615, opened 2024-05-28
Use aggregated types for `RuntimeFreezeReason` and better examples of `MaxFreezes`

bkontur:
This PR aligns the settings for `MaxFreezes`, `RuntimeFreezeReason`, and `FreezeIdentifier`.

#### Future work and improvements
https://github.com/paritytech/polkadot-sdk/issues/2997 (remove `MaxFreezes` and `FreezeIdentifier`)


substrate/frame/balances/src/lib.rs
```diff
@@ -222,13 +222,13 @@ pub mod pallet {
 			type ExistentialDeposit = ConstU64<1>;
 
 			type ReserveIdentifier = ();
-			type FreezeIdentifier = ();
+			type FreezeIdentifier = Self::RuntimeFreezeReason;
```
kianenigma, line 225: Seems like there is still a few pallets that don't use `TestDefaultConfig`, reasonable to update them as a part of this PR?

substrate/frame/examples/basic/src/tests.rs
```diff
@@ -81,10 +81,10 @@ impl pallet_balances::Config for Test {
 	type ExistentialDeposit = ConstU64<1>;
 	type AccountStore = System;
 	type WeightInfo = ();
-	type FreezeIdentifier = ();
-	type MaxFreezes = ();
-	type RuntimeHoldReason = ();
-	type RuntimeFreezeReason = ();
+	type FreezeIdentifier = RuntimeFreezeReason;
+	type MaxFreezes = VariantCountOf<RuntimeFreezeReason>;
```
gupnik, line 85: Can't we compute it internally using the already provided `RuntimeFreezeReason`? Do we still need to keep a separate config parameter?

gupnik: > https://github.com/paritytech/polkadot-sdk/issues/2997 (remove MaxFreezes and FreezeIdentifier)

I just saw this 🙈 

Shouldn't we just to go to that one directly?

bkontur: yes, exactly, that is the point of https://github.com/paritytech/polkadot-sdk/issues/2997 and will be done as separate PR

substrate/frame/balances/src/lib.rs
bkontur: > Seems like there is still a few pallets that don't use `TestDefaultConfig`, reasonable to update them as a part of this PR?

yes, good point, I will do that here

substrate/frame/examples/basic/src/tests.rs
bkontur: > > #2997 (remove MaxFreezes and FreezeIdentifier)
> 
> I just saw this 🙈
> 
> Shouldn't we just to go to that one directly?

Well, I would better do it as a separate PR (needs migration from FreezeIdentifier to RuntimeFreezeReason or whatever , plus some checks and so on...).
And I am also going to address Kian's comment here: https://github.com/paritytech/polkadot-sdk/pull/4615#discussion_r1618574118, so it will be a big PR by itself.

This PR fixes correct usage of these freeze stuff, so it will be a better ground for that #2997 refactor.

gupnik (approved):

substrate/frame/balances/src/lib.rs
bkontur: @kianenigma I updated all (hopefully) pallets for `pallet_balances::Config and `TestDefaultConfig`, please take a look

serban300 (approved):

ggwpez (approved):

substrate/frame/whitelist/src/mock.rs
```diff
@@ -43,20 +43,9 @@ impl frame_system::Config for Test {
 	type AccountData = pallet_balances::AccountData<u64>;
 }
 
+#[derive_impl(pallet_balances::config_preludes::TestDefaultConfig)]
 impl pallet_balances::Config for Test {
-	type MaxLocks = ();
```
kianenigma, line 47: HELL YEAH, thank you for moving all of this to `derive_impl`!

kianenigma (approved):

PR #4618, opened 2024-05-28
Unify `code_at` logic between `CallExecutor` & `Client`

bkchr:
This unifies the logic between `CallExecutor` and `Client` when it comes to fetching the `code` for a given block. The actual `code` depends on potential overrides/substitutes.

Besides that it changes the logic in the lookahead collator on which `ValidationCodeHash` it sends to the validator alongside the `POV`. We are now sending the code hash as found on the relay chain. This is done as the local node could run with an override which is compatible to the validation code on the relay chain, but has a different hash.


cumulus/client/consensus/aura/src/collators/mod.rs
```diff
@@ -51,29 +53,34 @@ async fn check_validation_code_or_log(
...
+		Some(state) => {
 			if state != *local_validation_code_hash {
 				tracing::warn!(
 					target: super::LOG_TARGET,
 					%para_id,
 					?relay_parent,
 					?local_validation_code_hash,
 					relay_validation_code_hash = ?state,
-					"Parachain code doesn't match validation code stored in the relay chain state",
+					"Parachain code doesn't match validation code stored in the relay chain state. This is expected if you are using an override for example.",
```
skunert, line 69: Is it expected actually? If we look at how `lookahead` collator is typically instantiated, we pass `client.code_at` as code_hash_provider like here:
https://github.com/paritytech/polkadot-sdk/blob/b9c81e1fafd2e03a1d452302e595d4e470c0d9fc/cumulus/polkadot-parachain/src/service.rs#L830-L832

With your changes in `client.rs` this will already take substitutes and overrides into account. The hash should match then and we would not even need to return the code hash from the relay chain, right?

prdoc/pr_4618.prdoc
```diff
@@ -0,0 +1,19 @@
...
+      This is now fixed by always sending the code hash as found in the state of the relay chain. Still a warning is printed when
+      the code hash doesn't match to inform the operator about the mismatch and its potential impact on the validation.
+  - audience: Node Dev
+    description: |
+      `Client::code_at` will now use the same `overrides`/`substitute` to determine the code for a given block as it is
+      done when executing any runtime call.
+
+crates:
+  - name: cumulus-client-consensus-aura
+    bump: patch
```
skunert, line 17: Since this changes behaviour I am in favor of `minor`.

cumulus/client/consensus/aura/src/collators/mod.rs
bkchr: If you are using a `override`, you will never find the hash of the override in the relay chain state. However, the relay chain would reject your pov if you send it with the hash of you override, as it doesn't know this hash.

skunert: So what is not super clear to me is what is the point of this check here except the logging message.

We fetch the code hash from the relay chain here and then submit it to the relay chain along with the collation. Previously if there is some other reason for a code hash mismatch, it would be caught. Can we not check if there is an override or substitute in place and only in that case return the relay chain code hash here?

bkchr: > So what is not super clear to me is what is the point of this check here except the logging message.

The entire purpose of this function was always to just log. There is nothing we can do, besides informing the human that something is wrong.

> Previously if there is some other reason for a code hash mismatch, it would be caught.

Not sure what other reason you mean, but we would still catch it. I mean even if the parachain is using a completely different code in its state, it will be printed. There is no real difference to before. The only difference is that we tell the relay chain what it expects from us (a little bit like cheating). However, the worst case would be that the relay chain rejects the Pov.

skunert (approved):

koute (approved):

PR #4620, opened 2024-05-28
pallet-staking: Put tests behind `cfg(debug_assertions)`

bkchr:
Otherwise these tests are failing if you don't run with `debug_assertions` enabled, which happens if you run tests locally in release mode.

### DIFF
--- a/substrate/frame/staking/src/tests.rs
+++ b/substrate/frame/staking/src/tests.rs
@@ -5251,6 +5251,7 @@ mod election_data_provider {
 	// maybe_max_len`.
 	#[test]
 	#[should_panic]
+	#[cfg(debug_assertions)]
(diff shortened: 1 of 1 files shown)


ggwpez (approved):
Maybe `#[cfg_attr(debug_assertions, should_panic)]` works?

bkchr:
> Maybe `#[cfg_attr(debug_assertions, should_panic)]` works?

Yeah that would also work, but renders the test even more useless?

franciscoaguirre (approved):
Why do the tests fail on release?

ggwpez (approved):
>Yeah that would also work, but renders the test even more useless?

Only tests that it wont panic otherwise.

PR #4621, opened 2024-05-28
Change `XcmDryRunApi::dry_run_extrinsic` to take a call instead

franciscoaguirre:
Follow-up to the new `XcmDryRunApi` runtime API introduced in https://github.com/paritytech/polkadot-sdk/pull/3872.

Taking an extrinsic means the frontend has to sign first to dry-run and once again to submit.
This is bad UX which is solved by taking an `origin` and a `call`.
This also has the benefit of being able to dry-run as any account, since it needs no signature.

This is a breaking change since I changed `dry_run_extrinsic` to `dry_run_call`, however, this API is still only on testnets.
The crates are bumped accordingly.

As a part of this PR, I changed the name of the API from `XcmDryRunApi` to just `DryRunApi`, since it can be used for general dry-running :)

Step towards https://github.com/paritytech/polkadot-sdk/issues/690.

Example of calling the API with PAPI, not the best code, just testing :)

```ts
// We just build a call, the arguments make it look very big though.
const call = localApi.tx.XcmPallet.transfer_assets({
  dest: XcmVersionedLocation.V4({ parents: 0, interior: XcmV4Junctions.X1(XcmV4Junction.Parachain(1000)) }),
  beneficiary: XcmVersionedLocation.V4({ parents: 0, interior: XcmV4Junctions.X1(XcmV4Junction.AccountId32({ network: undefined, id: Binary.fromBytes(encodeAccount(account.address)) })) }),
  weight_limit: XcmV3WeightLimit.Unlimited(),
  assets: XcmVersionedAssets.V4([{
    id: { parents: 0, interior: XcmV4Junctions.Here() },
    fun: XcmV3MultiassetFungibility.Fungible(1_000_000_000_000n) }
  ]),
  fee_asset_item: 0,
});
// We call the API passing in a signed origin 
const result = await localApi.apis.XcmDryRunApi.dry_run_call(
  WestendRuntimeOriginCaller.system(DispatchRawOrigin.Signed(account.address)),
  call.decodedCall
);
if (result.success && result.value.execution_result.success) {
  // We find the forwarded XCM we want. The first one going to AssetHub in this case.
  const xcmsToAssetHub = result.value.forwarded_xcms.find(([location, _]) => (
    location.type === "V4" &&
      location.value.parents === 0 &&
      location.value.interior.type === "X1"
      && location.value.interior.value.type === "Parachain"
      && location.value.interior.value.value === 1000
  ))!;

  // We can even find the delivery fees for that forwarded XCM.
  const deliveryFeesQuery = await localApi.apis.XcmPaymentApi.query_delivery_fees(xcmsToAssetHub[0], xcmsToAssetHub[1][0]);

  if (deliveryFeesQuery.success) {
    const amount = deliveryFeesQuery.value.type === "V4" && deliveryFeesQuery.value.value[0].fun.type === "Fungible" && deliveryFeesQuery.value.value[0].fun.value.valueOf() || 0n;
    // We store them in state somewhere.
    setDeliveryFees(formatAmount(BigInt(amount)));
  }
}
```


polkadot/runtime/rococo/src/lib.rs
```diff
@@ -1809,19 +1809,12 @@ sp_api::impl_runtime_apis! {
 		}
 	}
 
-	impl xcm_fee_payment_runtime_api::dry_run::XcmDryRunApi<Block, RuntimeCall, RuntimeEvent> for Runtime {
-		fn dry_run_extrinsic(extrinsic: <Block as BlockT>::Extrinsic) -> Result<ExtrinsicDryRunEffects<RuntimeEvent>, XcmDryRunApiError> {
+	impl xcm_fee_payment_runtime_api::dry_run::XcmDryRunApi<Block, RuntimeCall, RuntimeEvent, OriginCaller> for Runtime {
+		fn dry_run_call(origin: OriginCaller, call: RuntimeCall) -> Result<ExtrinsicDryRunEffects<RuntimeEvent>, XcmDryRunApiError> {
 			use xcm_builder::InspectMessageQueues;
```
bkchr, line 1814: Would be nice to provide an abstraction over this code to not always require to replicate it

franciscoaguirre: Thankfully had all the traits so I managed to put both implementations in `pallet-xcm` 😄

polkadot/runtime/westend/src/lib.rs
```diff
@@ -2271,63 +2271,13 @@ sp_api::impl_runtime_apis! {
...
-			let local_xcm = pallet_xcm::Pallet::<Runtime>::recorded_xcm();
-			let forwarded_xcms = xcm_config::XcmRouter::get_messages();
-			let events: Vec<RuntimeEvent> = System::read_events_no_consensus().map(|record| record.event.clone()).collect();
-			Ok(ExtrinsicDryRunEffects {
-				local_xcm: local_xcm.map(VersionedXcm::<()>::from),
-				forwarded_xcms,
-				emitted_events: events,
-				execution_result: result,
-			})
+	impl xcm_fee_payment_runtime_api::dry_run::XcmDryRunApi<Block, RuntimeCall, RuntimeEvent, OriginCaller> for Runtime {
```
acatangiu, line 2274: If we're making this breaking change (to this brand new API that noone is using yet), wdyt we name it a bit more generic since it can be used for non-xcm usecases as well?

```suggestion
	impl xcm_fee_payment_runtime_api::dry_run::DryRunApi<Block, RuntimeCall, RuntimeEvent, OriginCaller> for Runtime {
```

acatangiu (approved):

polkadot/xcm/xcm-fee-payment-runtime-api/src/dry_run.rs
```diff
@@ -58,9 +57,9 @@ sp_api::decl_runtime_apis! {
 	/// Extrinsics or XCMs might fail when executed, this doesn't mean the result of these calls will be an `Err`.
 	/// In those cases, there might still be a valid result, with the execution error inside it.
 	/// The only reasons why these calls might return an error are listed in the [`Error`] enum.
-	pub trait XcmDryRunApi<Call, Event: Decode> {
-		/// Dry run extrinsic.
-		fn dry_run_extrinsic(extrinsic: <Block as BlockT>::Extrinsic) -> Result<ExtrinsicDryRunEffects<Event>, Error>;
+	pub trait XcmDryRunApi<Call: Encode, Event: Decode, OriginCaller: Encode> {
+		/// Dry run call.
+		fn dry_run_call(origin: OriginCaller, call: Call) -> Result<ExtrinsicDryRunEffects<Event>, Error>;
```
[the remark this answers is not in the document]
franciscoaguirre: This was the first time I heard of `OriginCaller` as well 😂 
[`RuntimeOrigin` doesn't implement `Encode`](https://github.com/paritytech/polkadot-sdk/blob/aa32faaebf64426becb2feeede347740eb7a3908/substrate/frame/support/procedural/src/construct_runtime/expand/origin.rs#L106) while [`OriginCaller` does](https://github.com/paritytech/polkadot-sdk/blob/aa32faaebf64426becb2feeede347740eb7a3908/substrate/frame/support/procedural/src/construct_runtime/expand/origin.rs#L204)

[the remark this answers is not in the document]
franciscoaguirre: I added the example code I used with PAPI to call it from the frontend. PJS has the list of runtime APIs hard-coded so I don't know the UX there.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,74 @@ impl<T: Config> Pallet<T> {
 		AccountIdConversion::<T::AccountId>::into_account_truncating(&ID)
 	}
 
+	pub fn dry_run_call<Runtime, Router, OriginCaller, RuntimeCall>(
```
bkchr, line 2441: Missing docs :P At least it could mention that this should be used by the runtime api or similar.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,74 @@ impl<T: Config> Pallet<T> {
...
+				.collect();
+		Ok(CallDryRunEffects {
+			local_xcm: local_xcm.map(VersionedXcm::<()>::from),
+			forwarded_xcms,
+			emitted_events: events,
+			execution_result: result,
+		})
+	}
+
+	pub fn dry_run_xcm<Runtime, Router, RuntimeCall, XcmConfig>(
```
bkchr, line 2468: Same

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,74 @@ impl<T: Config> Pallet<T> {
...
+	{
+		crate::Pallet::<Runtime>::set_record_xcm(true);
+		let result = call.dispatch(origin.into());
+		crate::Pallet::<Runtime>::set_record_xcm(false);
+		let local_xcm = crate::Pallet::<Runtime>::recorded_xcm();
+		let forwarded_xcms = Router::get_messages();
+		let events: Vec<<Runtime as frame_system::Config>::RuntimeEvent> =
+			frame_system::Pallet::<Runtime>::read_events_no_consensus()
+				.map(|record| record.event.clone())
+				.collect();
```
bkchr, line 2459: You did not delete the events from the old block, this mean you will include these events here. I think you don't want this :P

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,74 @@ impl<T: Config> Pallet<T> {
...
+			xcm,
+			&mut hash,
+			Weight::MAX, // Max limit available for execution.
+			Weight::zero(),
+		);
+		let forwarded_xcms = Router::get_messages();
+		let events: Vec<<Runtime as frame_system::Config>::RuntimeEvent> =
+			frame_system::Pallet::<Runtime>::read_events_no_consensus()
+				.map(|record| record.event.clone())
+				.collect();
```
bkchr, line 2505: Same

polkadot/xcm/pallet-xcm/src/lib.rs
bkchr: `frame_system::Pallet::<T>::reset_events()` should do it. (Before calling dispathc)

bkchr (approved):
Besides the small issues, it looks good.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,83 @@ impl<T: Config> Pallet<T> {
 		AccountIdConversion::<T::AccountId>::into_account_truncating(&ID)
 	}
 
+	/// Dry-runs `call` with the given `origin`.
```
bkchr, line 2441:
```suggestion
	/// Dry-runs `call` with the given `origin`.
	///
```

You should always have an empty line between the explanation and further docs. I don't have a link at hand that describes this. However, this way only the first line is rendered in a "coarse" overview and you see the rest when clicking on the function. Makes the docs looks better.

polkadot/xcm/pallet-xcm/src/lib.rs
```diff
@@ -2432,6 +2438,83 @@ impl<T: Config> Pallet<T> {
...
+				.collect();
+		Ok(CallDryRunEffects {
+			local_xcm: local_xcm.map(VersionedXcm::<()>::from),
+			forwarded_xcms,
+			emitted_events: events,
+			execution_result: result,
+		})
+	}
+
+	/// Dry-runs `xcm` with the given `origin_location`.
```
bkchr, line 2473:
```suggestion
	/// Dry-runs `xcm` with the given `origin_location`.
	///
```

PR #4627, opened 2024-05-29
[subsystem-bench] Emulate latency for messages and requests from node 

### LINKED ISSUE #4611: subsystem-bench: add emulated network latency for messages and requests coming from local validator
AFAICT from the code, we're only emulating peer latency when sending a message from an emulated peer to the local node.
We should add latency for sending messages the other way around also (when sending a message from the local node to an emulated peer).

I validated that this is indeed the case by adding a 10s mean latency to the availability recovery bench and there wasn't any impact on performance

AndreiEres:
Fixes https://github.com/paritytech/polkadot-sdk/issues/4611
- Adds latency for messages from node
- The real delay is split in half because the original one is meant to be RTT.


polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -561,6 +583,7 @@ async fn emulated_peer_loop(
 						}
 					}
 					if let Some(message) = message {
+						// TODO: message.peer() is always None because it's a message from node
```
alindima, line 586: this needs to be fixed?

polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -476,25 +483,36 @@ impl EmulatedPeer {
 	pub async fn send_message(&mut self, message: NetworkMessage) {
 		self.tx_limiter.reap(message.size()).await;
 
-		if self.latency_ms == 0 {
-			self.to_node.unbounded_send(message).expect("Sending to the node never fails");
-		} else {
+		if self.has_latency() {
+			let latency = self.emulate_latency();
```
alindima, line 487: nit: we could have a `get_latency()` which returns a `Option<tokio::time::Sleep>` instead of these two functions and use `if let Some(latency) = peer.get_latency() {...`

alindima (approved):
Looks good and seems to work

polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -548,6 +566,10 @@ async fn emulated_peer_loop(
 					emulated_peer.rx_limiter().reap(size).await;
 					stats.inc_received(size);
 
+					if peer_message.is_from_node() && emulated_peer.has_latency() {
+						emulated_peer.emulate_latency().await;
```
sandreim, line 570: This will effectively stall the peer loop for up to whatever latency/2 is configured, meaning the peer also won't send messages to the node, even if we request it to from main thread. This latency needs to be emulated in the task that is supposed to send messages from the node to the peer, see `proxy_send_request` method.

polkadot/node/subsystem-bench/src/lib/network.rs
AndreiEres: Yes but not in this PR

polkadot/node/subsystem-bench/src/lib/network.rs
AndreiEres: Changed it. Is that what you were thinking about?

polkadot/node/subsystem-bench/src/lib/network.rs
```diff
@@ -349,8 +349,20 @@ impl NetworkInterface {
...
-							tx_network.send_message_to_peer(&peer, message),
+						NetworkMessage::MessageFromNode(peer, message) => {
+							let tx_network = tx_network.clone();
+							let send_task = async move {
+								let receiver = tx_network.peer(&peer);
+								let latency_ms = receiver.latency_ms();
+								if latency_ms > 0 && receiver.is_connected() {
+									emulate_latency(latency_ms).await;
+								}
+								tx_network.send_message_to_peer(&peer, message);
```
sandreim, line 360: Why do we still want to send to a disconnected peer ? Wouldn't the assert in `send_message_to_peer` trigger if peer is not connected ?

PR #4631, opened 2024-05-29
feat: add runtime-wide `try-state` hooks

### LINKED ISSUE #235: Allow runtime-wide `TryState` checks
Currently the `TryState` can only be utilized on a per-pallet level by implementing the `try_state` hook.   
This limits the usage tor multi-pallet `try_state` functionality. `Executive` could be extended to accept an optional `TryState` which will then be executed by the try-runtime CLI on a runtime-wide level.

cc @kianenigma

ntn-x2:
Fixes https://github.com/paritytech/polkadot-sdk/issues/235.

I tried different approaches, including the suggestion of adding a `fn id() -> String;` function on the `trait TryState<BlockNumber>` trait, but that was giving me issues when trying to derive `TryState` for tuples (what is the `id()` of a tuple of elements such as `AllPalletsWithSystem`)? Also, it feels less than ideal to me for a trait to implement both logic that can take a `Select::Only` parameter (on which you should apply filtering logic, that's why it's ignored in 100% of the cases that I found) and also returns an `id() -> String` in there.

So, I am proposing the following approach:

* `Executive` takes a new generic with a default `()` implementation for custom `TryState` hooks, as done for the `OnRuntimeUpgrade` ones. Like the `OnRuntimeUpgrade` hooks, `TryState` hooks are executed before pallet-specific hooks.
* I introduce two new traits, `trait TryStateLogic<BlockNumber>` which is responsible for the `try-state` logic itself, and the `trait IdentifiableTryStateLogic<BlockNumber>: TryStateLogic<BlockNumber>`, which returns a function that is used to match the specific test with one of the provided identifiers in `TryStateSelect::Only(names)`. The two traits could be collapsed into one, if need be, as long as they expose the two functions declared here.
* By default, the `pallet` macro also implements the two new traits, implementing `TryStateLogic<BlockNumber>` re-using the `Hooks<BlockNumber` logic for `try_state()` and `IdentifiableTryStateLogic<BlockNumber>` re-using the `PalletInfoAccess` logic for `name()`.

## Open points

* I did not find a way to implement a trait for a variable number of tuples without the requirement for the elements of the tuple to also implement the trait, so I implement the `TryState` trait for the empty tuple `()`, the `(T,)` tuple and the `(T1, T2)` tuple, which covers the `(CustomHook, AllPallets*)` case. We might want to have a macro where the **whole** `where` clause is configurable, not only additional traits. But it would happen in a different PR.
* I augmented the `Executive` unit tests to test that the `TryState` hooks are called in the right order. Because `TryState` hooks are not supposed to modify storage, I could not use the same approach used for `OnRuntimeUpgrade` hooks. Hence, I used a static variable that is reset to `false` with a new test ext, and set to `true` when the test custom `TryState` is invoked. I did not get any errors in the tests, but it could be that this leads to concurrency issues, in which case an alternative strategy is needed, perhaps including an additional boolean flag in what is returned by `new_test_ext`?

### DIFF
--- a/substrate/frame/executive/src/lib.rs
+++ b/substrate/frame/executive/src/lib.rs
@@ -86,6 +86,12 @@
 //! done by setting an optional generic parameter. The custom logic will be called before
 //! the on runtime upgrade logic of all modules is called.
 //!
+//! ### Custom `TryState` logic
+//!
+//! You can add custom logic that should be called in your runtime on a try-state test. This is
+//! done by setting an optional generic parameter after the `OnRuntimeUpgrade` type. The custom logic will be called before
+//! the `try-state` logic of all modules is called.
+//!
 //! ```
 //! # use sp_runtime::generic;
 //! # use frame_executive as executive;
@@ -115,7 +121,16 @@
 //!     }
 //! }
 //!
-//! pub type Executive = executive::Executive<Runtime, Block, Context, Runtime, AllPalletsWithSystem, CustomOnRuntimeUpgrade>;
+//! struct CustomTryState;
+//! impl frame_support::traits::TryState<BlockNumber> for CustomTryState {
+//!
+//! 	fn try_state(_: BlockNumber, _: Select) -> Result<(), sp_runtime::TryRuntimeError> {
+//!         // Do whatever you want.
+//!         Ok(())
+//! 	}
+//! }
+//!
+//! pub type Executive = executive::Executive<Runtime, Block, Context, Runtime, AllPalletsWithSystem, CustomOnRuntimeUpgrade, CustomTryState>;
 //! ```
 
 #[cfg(doc)]
@@ -205,6 +220,8 @@ pub type OriginOf<E, C> = <CallOf<E, C> as Dispatchable>::RuntimeOrigin;
 ///   used to call hooks e.g. `on_initialize`.
 /// - `OnRuntimeUpgrade`: Custom logic that should be called after a runtime upgrade. Modules are
 ///   already called by `AllPalletsWithSystem`. It will be called before all modules will be called.
+/// - `TryState`: Custom logic that should be called when running `try-state` tests. Modules are
+///   already called by `AllPalletsWithSystem`. It will be called before all modules will be called.
 pub struct Executive<
 	System,
 	Block,
@@ -212,6 +229,7 @@ pub struct Executive<
 	UnsignedValidator,
 	AllPalletsWithSystem,
 	OnRuntimeUpgrade = (),
+	TryState = (),
 >(
 	PhantomData<(
 		System,
@@ -220,6 +238,7 @@ pub struct Executive<
 		UnsignedValidator,
 		AllPalletsWithSystem,
 		OnRuntimeUpgrade,
+		TryState,
 	)>,
 );
 
@@ -280,8 +299,17 @@ impl<
 			+ TryState<BlockNumberFor<System>>
 			+ TryDecodeEntireStorage,
 		COnRuntimeUpgrade: OnRuntimeUpgrade,
-	> Executive<System, Block, Context, UnsignedValidator, AllPalletsWithSystem, COnRuntimeUpgrade>
-where
+		CTryState: frame_support::traits::TryState<BlockNumberFor<System>>,
+	>
+	Executive<
+		System,
+		Block,
+		Context,
+		UnsignedValidator,
+		AllPalletsWithSystem,
+		COnRuntimeUpgrade,
+		CTryState,
+	> where
 	Block::Extrinsic: Checkable<Context> + Codec,
 	CheckedOf<Block::Extrinsic, Context>: Applyable + GetDispatchInfo,
 	CallOf<Block::Extrinsic, Context>:
@@ -318,7 +346,7 @@ where
 
 		// Check if there are any forbidden non-inherents in the block.
 		if mode == ExtrinsicInclusionMode::OnlyInherents && extrinsics.len() > num_inherents {
-			return Err("Only inherents allowed".into())
+			return Err("Only inherents allowed".into());
 		}
 
 		let try_apply_extrinsic = |uxt: Block::Extrinsic| -> ApplyExtrinsicResult {
@@ -343,7 +371,7 @@ where
 			let r = Applyable::apply::<UnsignedValidator>(xt, &dispatch_info, encoded_len)?;
 
 			if r.is_err() && dispatch_info.class == DispatchClass::Mandatory {
-				return Err(InvalidTransaction::BadMandatory.into())
+				return Err(InvalidTransaction::BadMandatory.into());
 			}
 
 			<frame_system::Pallet<System>>::note_applied_extrinsic(&r, dispatch_info);
@@ -359,7 +387,7 @@ where
 					e,
 					err,
 				);
-				break
+				break;
 			}
 		}
 
@@ -377,7 +405,7 @@ where
 
 		// run the try-state checks of all pallets, ensuring they don't alter any state.
 		let _guard = frame_support::StorageNoopGuard::default();
-		<AllPalletsWithSystem as frame_support::traits::TryState<
+		<(CTryState, AllPalletsWithSystem) as frame_support::traits::TryState<
 			BlockNumberFor<System>,
 		>>::try_state(*header.number(), select.clone())
 		.map_err(|e| {
@@ -456,7 +484,7 @@ where
 
 		// Check all storage invariants:
 		if checks.try_state() {
-			AllPalletsWithSystem::try_state(
+			<(CTryState, AllPalletsWithSystem)>::try_state(
 				frame_system::Pallet::<System>::block_number(),
 				TryStateSelect::All,
 			)?;
@@ -621,9 +649,9 @@ where
 		// Check that `parent_hash` is correct.
 		let n = *header.number();
 		assert!(
-			n > BlockNumberFor::<System>::zero() &&
-				<frame_system::Pallet<System>>::block_hash(n - BlockNumberFor::<System>::one()) ==
-					*header.parent_hash(),
+			n > BlockNumberFor::<System>::zero()
+				&& <frame_system::Pallet<System>>::block_hash(n - BlockNumberFor::<System>::one())
+					== *header.parent_hash(),
 			"Parent hash should be valid.",
 		);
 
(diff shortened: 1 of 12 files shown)


ntn-x2:
I can fix the other CI issues once I get a thumbs-up on the current approach.

liamaharon:
bot fmt

ntn-x2:
@liamaharon can I get a review from anyone? The PR has now been hanging for quite a while with 0 interactions with anyone from Parity.

ntn-x2:
@liamaharon can I get anyone to review this PR? In particular, I am faced with a problem: the `try-runtime` hooks are not supposed to make any storage changes. Hence, I did not find out any other way than using a `static mut` global variable, which is by itself very bad. Do you have any other way in mind to test the right of hooks being called?

ggwpez:
I will try to take a look today. Liam is currently offboarding 😢

substrate/frame/support/src/traits/try_runtime/mod.rs
```diff
@@ -145,29 +145,90 @@ impl core::str::FromStr for UpgradeCheckSelect {
...
+
+/// Execute some checks to ensure the internal state of a pallet is consistent.
+///
+/// Usually, these checks should check all of the invariants that are expected to be held on all of
+/// the storage items of your pallet.
+///
+/// This hook should not alter any storage.
+pub trait TryState<BlockNumber> {
+	/// Execute the state checks.
+	fn try_state(_: BlockNumber, _: Select) -> Result<(), TryRuntimeError>;
```
ggwpez, line 202: Why is the block number passed in? It can just be read from the System pallet?

ggwpez (commented):
Im not sure if i completely get the architecture here.  
Can you please briefly put the new traits into the MR description maybe and what they should do? A graphic can also help, but maybe not worth it if we have to change it.

substrate/frame/executive/src/lib.rs
```diff
@@ -115,7 +122,23 @@
 //!     }
 //! }
 //!
-//! pub type Executive = executive::Executive<Runtime, Block, Context, Runtime, AllPalletsWithSystem, CustomOnRuntimeUpgrade>;
+//! struct CustomTryState;
+//!
+//! #[cfg(feature = "try-runtime")]
+//! impl frame_support::traits::IdentifiableTryStateLogic<BlockNumber> for CustomTryState {
+//!
+//! 	fn try_state(_: BlockNumber) -> Result<(), sp_runtime::TryRuntimeError> {
```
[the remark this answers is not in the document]
ntn-x2: Mhh but the trait `IdentifiableTryStateLogic` is also implemented for `AllPalletsWithSystem`, so having a `before_all_pallets` method would not be ideal naming, I think. Also, this would be implemented by each pallet, so for a pallet to implement a function `before_all_pallets` can get quite confusing. 

I'll shortly update the PR to add more context, but I also tried to keep the changeset to the minimum without changing too many things from how things work today.

substrate/frame/support/src/traits/try_runtime/mod.rs
ntn-x2: I mean, this is just the `TryState` trait as it is today. I would not want to change this in this PR as well. Of course, if you think that we should change that here, I am open to do so. But not having the `BlockNumber` argument implicitly assumes all implementors always have a way to retrieve it. This is currently true for both pallet and runtime logic, but it's nevertheless one more assumption we would bake into this trait.

substrate/frame/executive/src/lib.rs
```diff
@@ -115,7 +122,23 @@
...
+//!         // Do whatever you want.
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/paritytech/polkadot-sdk/pull/4631/commits/56e91c7b3915a0d538a72a9ccc3ccc4f663f0106

ntn-x2:
@ggwpez I actually re-checked the description I gave to the PR, and I am not sure what is unclear about it. Could you please tell me what's the part that you think should be improved?

substrate/frame/executive/src/lib.rs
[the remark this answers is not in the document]
ggwpez: Ah, i mixed it up with runtime-wide post migration check. nvm

substrate/frame/executive/src/tests.rs
```diff
@@ -44,6 +44,40 @@ use pallet_balances::Call as BalancesCall;
...
+
+	pub(super) struct CustomTryState;
+
+	impl<BlockNumber> TryStateLogic<BlockNumber> for CustomTryState
+	where
+		BlockNumber: Clone + sp_std::fmt::Debug + AtLeast32BitUnsigned,
+	{
+		fn try_state(_: BlockNumber) -> Result<(), sp_runtime::TryRuntimeError> {
+			unsafe {
+				CANARY_FLAG = true;
```
ggwpez, line 65: `CANARY_FLAG::set(true);`

in combination with `parameter_types!` pub static please. Cell would also work, but normally we use parameter types.

substrate/frame/support/src/traits/try_runtime/mod.rs
```diff
@@ -145,29 +145,90 @@ impl core::str::FromStr for UpgradeCheckSelect {
...
+#[cfg_attr(all(feature = "tuples-96", not(feature = "tuples-128")), impl_for_tuples(96))]
+#[cfg_attr(all(feature = "tuples-128"), impl_for_tuples(128))]
+impl<BlockNumber> IdentifiableTryStateLogic<BlockNumber> for Tuple
+where
+	BlockNumber: Clone + sp_std::fmt::Debug + AtLeast32BitUnsigned,
+{
+	for_tuples!( where #( Tuple: TryStateLogic<BlockNumber> )* );
+	fn matches_id(id: &[u8]) -> bool {
+		for_tuples!( #( if Tuple::matches_id(id) { return true; } )* );
+		return false;
```
ggwpez, line 190: That means that it returns true when one of the tuple elements returns true? But what if we only want to run that single tuple element instead of the whole tuple?  

I think i am missing some critical info about this merge request, which is why i dont get what it was done this way.  
Is it not enough to just add a `TryState` to the end of `Executive`?

ntn-x2: Perhaps the context in https://github.com/paritytech/polkadot-sdk/issues/235 could help you understand better?

The goal of this PR is to achieve exactly what you are suggesting here. Having a generic to run `try-state` checks at the runtime level. The problem with the current structure (as also mentioned in the linked issue), is that the implementation on tuples is coupled with the `PalletInfoAccess` trait, as you can see here: https://github.com/paritytech/polkadot-sdk/blob/4aa29a41cf731b8181f03168240e8dedb2adfa7a/substrate/frame/support/src/traits/try_runtime/mod.rs#L156.

This means that in order to use tuple elements that are not pallet, such as runtime-wide logic, the coupling to such a trait must be removed. And because of some requirements the `impl_for_tuples` has about knowing the type and size of the types involved, it was not possible to use anything else than a function that returns `true` if the specified identifier matches what was provided by the user in the `try-state` command.

So now if you have a tuple of, for instance, `(CustomLogic, AllPalletsWithSystem)`, the matching function will return `true` if `CustomLogic` matches, or if any pallet has the specified ID. It is assumed that no two components, be they pallets or else, have the same ID, as that would mean the latter of them would never be called.

ntn-x2: And this goes beyond the `AllPalletsWithSystem` usage. If even just `CustomLogic` is itself a tuple, then with the current design it is required to implement `PalletInfoAccess`, which of course makes no sense for anything else than a pallet.

substrate/frame/executive/src/tests.rs
ntn-x2: Addressed in https://github.com/paritytech/polkadot-sdk/pull/4631/commits/a87c5947c8c025981475c52121e5ce48c42b0eb2.

substrate/frame/support/src/traits/try_runtime/mod.rs
ggwpez: > The goal of this PR is to achieve exactly what you are suggesting here. Having a generic to run try-state checks at the runtime level

Then the easiest way is to just make a new trait, since the `TryState` trait is aimed at pallets (should probably have been named `PalletTryState` but now we are stuck with this for compatibility reasons).  
```rust
pub trait RuntimeTryState {
    fn try_state(whatever...) -> Result<(), TryRuntimeError>;
}
```

Anyway, i dont really want to block this approach if it has clear advantages over ^, so going to ask @kianenigma what he thinkgs.

kianenigma: We've had a lot of iterations here, thank you both, esp @ntn-x2 for exploring solutions with us. 

The systemic mistake I am realizing (noted for future) is that we should have first discussed the "ideal outcome" vs the approach. So let's do that. 

what should we expect of `TryStateSelect` in the context of this PR? Should we be able to tweak `RuntimeTryState` with it? 

If yes, the current approach is the right way. If not, Oliver's approach is much simpler and `RuntimeTryState` will always be executed.

I am leaning toward not allowing runtime level try-state logic to be configurable with `TryStateSelect`. 

**First reason**: This API, as pointed out here, is already somewhat broken, and we are not even sure if it is "enough". In that, the point of it was to _regulate_ how much execution time is given to try-state hooks. But, through experience, we have seen that the issue is that a single pallet has a try-state-bomb that takes forever. In an ideal sense, instead of `TryStateSelect` you would pass some `Weight`-equivalent amount, and this much weight was used in try-state. With PVM we can do this.

**Second reason**: If we allow `TryStateSelect` to _properly_ configure runtime-try-state, the variants would be a lot. If you want "round-robin pallets + runtime", or "pallets x, y, z and runtime". These are hard to express.  

**Third Reason**: Let's do the simpler approach. IFF someone needs to disable runtime level try-state, we have a blueprint of how to solve it.

bkchr: > But, through experience, we have seen that the issue is that a single pallet has a try-state-bomb that takes forever. In an ideal sense, instead of `TryStateSelect` you would pass some `Weight`-equivalent amount, and this much weight was used in try-state. With PVM we can do this.

PVM will not come tomorrow and I also don't see why we should need this here. These `try-state` hooks are not being run as part of the on chain logic. How long they take is really not important at all. We need to ensure that they finish on time. If there is some kind of "try-state bomb", it probably means that we read too much state. Still the amount of state is not in the terabytes or similar, thus I would say that our current read performance needs to be optimized.

ntn-x2: @kianenigma sorry I think I did not fully understand what your suggestion here is. Are you suggesting to have a new trait which takes only a block number and that is required to be implemented by the new generic? That would mean that the `TryStateSelect` logic would only be applied to pallets, while all other runtime try-states would all always be executed. Is that the proposed way forward? Would the runtime logic still be called before the `AllPallets` logic?

PR #4634, opened 2024-05-29
Implement `XcmPaymentApi` and `DryRunApi` on all system parachains

franciscoaguirre:
Depends on https://github.com/paritytech/polkadot-sdk/pull/4621.

Implemented the [`XcmPaymentApi`](https://github.com/paritytech/polkadot-sdk/pull/3607) and [`DryRunApi`](https://github.com/paritytech/polkadot-sdk/pull/3872) on all system parachains.

More scenarios can be tested on both rococo and westend if all system parachains implement this APIs.
The objective is for all XCM-enabled runtimes to implement them.
After demonstrating fee estimation in a UI on the testnets, come the fellowship runtimes.

Step towards https://github.com/paritytech/polkadot-sdk/issues/690.


acatangiu (approved):
🚀

svyatonik (approved):
One nit: `query_acceptable_payment_assets` and `query_weight_to_asset_fee` have some logic (converting, matching and logging) and since it'll propagate to another 8+ chains in the future (meaning DotSama), maybe it worth to add that implementation to some crate instead of copying the same code over and over. Ideally it should be just one liner :)

franciscoaguirre:
> One nit: `query_acceptable_payment_assets` and `query_weight_to_asset_fee` have some logic (converting, matching and logging) and since it'll propagate to another 8+ chains in the future (meaning DotSama), maybe it worth to add that implementation to some crate instead of copying the same code over and over. Ideally it should be just one liner :)

Good point, if we had a good config item on pallet-xcm that gave us what we need I could de-duplicate very easily. However, that's not the case. I can make one in the future. For now, I managed to get `query_acceptable_payment_assets` to be shorter. For `query_weight_to_asset_fee` I'll make one in a future PR so this can be merged.

PR #4636, opened 2024-05-29
Fix broken windows build

### LINKED ISSUE #4625: Building on Windows is broken due to symbolic link

### Description of bug

#3087 added the symbolic link pointing to `docs` to`substrate/primitives/runtime`, but on Windows (tested on Windows 10), symbolic link cannot be recognized, so the following line causes file not found error during compiling Substrate.

https://github.com/paritytech/polkadot-sdk/blob/89604daa0f4244bc83782bd489918cfecb81a7d0/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L60

### Steps to reproduce

Compile `sp-runtime` on Windows.

conr2d:
Fixes #4625.

Specifically, the `cfg` attribute `windows` refers to the compile target and not the build environment, and in the case of cross-compilation, the build environment and target can differ. However, the line modified is related to documentation generation, so there should be no critical issue with this change.


bkchr (approved):

bkchr:
Ty!

ggwpez (approved):

PR #4638, opened 2024-05-29
Update the `polkadot_builder` Dockerfile

rzadp:
This Dockerfile seems outdated - it currently fails to build (on my machine).
I don't see it being built anywhere on CI.

I did a couple of tweaks to make it build.

### DIFF
--- a/docker/dockerfiles/polkadot/polkadot_builder.Dockerfile
+++ b/docker/dockerfiles/polkadot/polkadot_builder.Dockerfile
@@ -1,5 +1,5 @@
 # This is the build stage for Polkadot. Here we create the binary in a temporary image.
-FROM docker.io/paritytech/ci-linux:production as builder
+FROM docker.io/paritytech/ci-unified:latest as builder
 
 WORKDIR /polkadot
 COPY . /polkadot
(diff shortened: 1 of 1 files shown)


docker/dockerfiles/polkadot/polkadot_builder.Dockerfile
```diff
@@ -19,7 +19,8 @@ LABEL description="Multistage Docker image for Polkadot: a platform for web3" \
 
 COPY --from=builder /polkadot/target/release/polkadot /usr/local/bin
 
-RUN useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
+USER root
```
rzadp, line 22: Changed back to an unprivileged user below, need root to execute those commands.

AndWeHaveAPlan (approved):

docker/dockerfiles/polkadot/polkadot_builder.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This is the build stage for Polkadot. Here we create the binary in a temporary image.
-FROM docker.io/paritytech/ci-linux:production as builder
+FROM docker.io/paritytech/ci-unified:latest as builder
```
alvicsam, line 2:
```suggestion
FROM docker.io/paritytech/ci-unified:bullseye-1.77.0-2024-04-10-v20240408 as builder
```
I'd use current image (can be found [here](https://github.com/paritytech/polkadot-sdk/blob/master/.github/env)) because latest can be newer

alvicsam (approved):

EgorPopelyaev (approved):

Issue #4552, opened 2024-05-23
CRV2: Missing pallet index causes misleading error message


kianenigma:
When I removed all pallet indices in `mod runtime` to see if it is supported or not, I get:

```
error: `System` pallet declaration is missing. Please add this line: `pub type System = frame_system;`
    --> /home/kian/remote-builds/polkadot-sdk_minimal-template/runtime/src/lib.rs:96:5
     |
  96 | mod runtime {
     |     ^^^^^^^
```

I think it is a good choice to no longer support implicit indices, but the error message should be improve.

PR #4643, opened 2024-05-30
Better error for missing index in CRV2

gupnik:
Fixes https://github.com/paritytech/polkadot-sdk/issues/4552


gupnik:
bot fmt

substrate/frame/support/procedural/src/runtime/parse/mod.rs
```diff
@@ -187,6 +187,13 @@ impl Def {
 				}
 			}
 
+			if pallet_index.is_none() {
```
bkchr, line 190:
```suggestion
			let Some(pallet_index) = pallet_index else {
```

substrate/frame/support/procedural/src/runtime/parse/mod.rs
```diff
@@ -187,6 +187,13 @@ impl Def {
...
+				if let syn::Item::Type(item) = item {
```
bkchr, line 191: This if doesn't make sense? I mean the index can only be set when the item is a type, at least judging based on the code above.

bkchr (approved):

substrate/frame/support/procedural/src/runtime/parse/mod.rs
gupnik: > This if doesn't make sense? I mean the index can only be set when the item is a type, at least judging based on the code above.

So, the module marked with #[frame_support::runtime] can have arbitrary items. For example,
```rust
#[frame_support::runtime]
mod runtime {
    use super::*;
    
    pub type System = frame_system;
}
```

In this case, you only want this check for `pub type System = frame_system;` and not for `use super::*;`. 

Basically, this change requires all type aliases within the runtime macro to be annotated with `#[runtime::pallet_index]` while the other items are preserved.

substrate/frame/support/procedural/src/runtime/parse/mod.rs
gupnik: `pallet_index` can be None for all items other than type aliases.

PR #4646, opened 2024-05-30
[Identity] Remove double encoding username signature payload

georgepisaltu:
In order to receive a username in `pallet-identity`, users have to, among other things, provide a signature of the desired username. Right now, there is an [extra encoding step](https://github.com/paritytech/polkadot-sdk/blob/4ab078d6754147ce731523292dd1882f8a7b5775/substrate/frame/identity/src/lib.rs#L1119) when generating the payload to sign.

Encoding a `Vec` adds extra bytes related to the length, which changes the payload. This is unnecessary and confusing as users expect the payload to sign to be just the username bytes. This PR fixes this issue by validating the signature directly against the username bytes.


joepetrowski (approved):

seadanda (approved):

prdoc/pr_4646.prdoc
```diff
@@ -0,0 +1,20 @@
...
+      encode these bytes before signing.
+  - audience: Runtime User
+    description: |
+      The signature payload for setting a username for an account in `pallet-identity` is now just
+      the raw bytes of said username (still including the suffix), removing the need to first
+      encode these bytes before signing.
+
+crates:
+  - name: pallet-identity
+    bump: minor
```
acatangiu, line 20: Isn't this breaking the `Identity::set_username_for()` call? The API is not changed per-se, but IIUC the caller has to provide a "different" signature now.

```suggestion
    bump: major
```

georgepisaltu: The API is the same and the caller has to construct the payload to be signed differently, but at the same time:
- this is a bug and the payload to be signed should have never had the leading length bytes
- this username feature is relatively new and unused (hence why nobody complained about the bug)

Given that the actual code is backwards compatible, I didn't mark it as a major change, but because of the behavior change it's not a patch either, so I went with minor. Do you still think it should be marked as a major change?

acatangiu: > this is a bug and the payload to be signed should have never had the leading length bytes

Was it working before? If yes, then I see it as a breaking/major change.
Like you correctly say, at the API definition code contract it is "backward compatible", but in reality it isn't (any existing tool/lib/UI will continue to "work" but provide incorrect signatures).

If it was not working before because of this bug, and now it does, it can even be a patch _fix_.

> this username feature is relatively new and unused (hence why nobody complained about the bug)

I'm glad we caught this early, that means nobody will complain about the potential `major` change.

acatangiu (approved):

PR #4637, opened 2024-05-29
Add Dockerfiles to the templates

rzadp:
As requested [here](https://github.com/paritytech/polkadot-sdk/issues/3155#issuecomment-2126934939).

The Dockerfiles are inspired by [this one](https://github.com/paritytech/polkadot-sdk/blob/aa32faaebf64426becb2feeede347740eb7a3908/docker/dockerfiles/polkadot/polkadot_builder.Dockerfile).

PR #4649, opened 2024-05-30
Update the Readme of the minimal template

rzadp:
- Addresses [this](https://github.com/paritytech/polkadot-sdk/issues/3155#issuecomment-2126934939).
- Technical content got adopted from the existing [solochain readme](https://github.com/paritytech/polkadot-sdk/tree/master/templates/solochain).
- The docker instructions will work after https://github.com/paritytech/polkadot-sdk/pull/4637.
- See the [rendered version](https://github.com/paritytech/polkadot-sdk/tree/rzadp/minimal-template-readme/templates/minimal).


templates/minimal/README.md
```diff
@@ -1,13 +1,93 @@
...
+# Polkadot SDK's Minimal Template
 
-# Docs
+<img height="70px" alt="Polkadot SDK Logo" src="https://github.com/paritytech/polkadot-sdk/raw/master/docs/images/Polkadot_Logo_Horizontal_Pink_White.png#gh-dark-mode-only"/>
+<img height="70px" alt="Polkadot SDK Logo" src="https://github.com/paritytech/polkadot-sdk/raw/master/docs/images/Polkadot_Logo_Horizontal_Pink_Black.png#gh-light-mode-only"/>
 
-You can generate and view the [Rust
-Docs](https://doc.rust-lang.org/cargo/commands/cargo-doc.html) for this template
-with this command:
+> This is a minimal template for creating a [Substrate](https://substrate.io/) blockchain.
```
kianenigma, line 8: Can we already avoid linking to substrate.io overall? it is all future refactor if we leave it like this. 

I suppose for all of this we can either find a replacement, or bring it to `polkadot-sdk-docs` crate.

templates/minimal/README.md
```diff
@@ -1,13 +1,93 @@
-# Minimal Template
+<div align="center">
 
-This is a minimal template for creating a blockchain using the Polkadot SDK.
+# Polkadot SDK's Minimal Template
 
-# Docs
```
kianenigma, line 5: So the question is what is the minimal template? 

The main difference between the minimal template and others is that it has the least number of components to run. 

In the runtime side, this template only has a handful of pallets, which you can list. 

On the node side, this template has no consensus and can run both with custom block time via CLI. see `enum Consensus` in the `node` folder.

In general, for palying around with a single node, this template is best.

templates/minimal/README.md
```diff
@@ -1,13 +1,93 @@
...
+
+# docker version:
+docker run --rm polkadot-sdk-minimal-template --dev
+```
+
+Development chains:
+
+- 🧹 Do not persist the state.
+- 💰 Are preconfigured with a genesis state that includes several prefunded development accounts.
+- 🧑‍⚖️ Development accounts are used as default validator authorities and a `sudo` account.
```
kianenigma, line 52: Does the template actually have this? needs to be double checked

templates/minimal/README.md
```diff
@@ -1,13 +1,93 @@
...
+
+## Contributing
+
+🔄 This template is automatically updated after releases in the main [Polkadot SDK monorepo](https://github.com/paritytech/polkadot-sdk).
+
+➡️ Any pull requests should be directed to this [source](https://github.com/paritytech/polkadot-sdk/tree/master/templates/minimal).
+
+😇 Please refer to the monorepo's [contribution guidelines](https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/CONTRIBUTING.md) and [Code of Conduct](https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/CODE_OF_CONDUCT.md).
+
+## Getting Help
```
kianenigma, line 87: This is great.

templates/minimal/node/README.md
```diff
@@ -0,0 +1,38 @@
+# Node
+
+A blockchain node is an application that allows users to participate in a
```
kianenigma, line 3: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/reference_docs/wasm_meta_protocol/index.html is a good resource here.

kianenigma (approved):

templates/minimal/README.md
```diff
@@ -1,13 +1,93 @@
...
+
+</div>
+
+## Getting Started
+
+🛠️ Depending on your operating system and Rust version, there might be additional
+packages required to compile this template.
+
+👉 Check the
+[Substrate Install](https://docs.substrate.io/install/) instructions for your platform for
```
kianenigma, line 20: For this, we have: 

rust-setup.md in solochain template
and `https://getsubstrate.io/` is a legacy project that needs to be archived.

For now I suggest is to remove both of these, add one single bash file to `polkadot-sdk` that installs latest rust and whatever else you need and reference to it.

templates/minimal/README.md
rzadp: The accounts are prefunded, and I sent a `sudo` extrinsic from Alice and it worked.
Not sure what to check in terms of `default validator authorities`.

Screenshot 2024-05-31 at 10 11 31

templates/minimal/README.md
rzadp: There used to be [this `init.sh` script](https://github.com/paritytech/polkadot/blob/master/scripts/init.sh) in the old repo that did this.

Do you know why it was abandoned and eventually removed?

templates/minimal/README.md
rzadp: OK, I'll try to find replacement for those substrate.io links.

rzadp: I have removed all `substrate.io` links. I also revamped the sub-readmes for Node, Pallets and Runtime, please take another look.

templates/minimal/README.md
rzadp: I felt like some more introduction of what you're looking at is missing.

I added something based on this comment.

templates/minimal/README.md
rzadp: For now I wrote a minimal instruction, until we have a script or a doc that could be linked.

rzadp:
I have made some minor changes in the main readme, and a major change in [Node](https://github.com/paritytech/polkadot-sdk/blob/rzadp/minimal-template-readme/templates/minimal/node/README.md), [Runtime](https://github.com/paritytech/polkadot-sdk/blob/rzadp/minimal-template-readme/templates/minimal/runtime/README.md) and [Pallets](https://github.com/paritytech/polkadot-sdk/blob/rzadp/minimal-template-readme/templates/minimal/pallets/README.md) sub-readmes.

I believe we shouldn't have deep technical info there (links instead), otherwise it will be an additional liability to keep up to date.

@kianenigma Please take another look, in my eye it all looks neat now.

templates/minimal/node/README.md
rzadp: That was helpful.

templates/minimal/README.md
kianenigma: there are no validators in this setup, you can remove it.

rzadp: Removed.

rzadp:
Looking for a second approval please.
@gupnik ?

templates/minimal/node/README.md
```diff
@@ -0,0 +1,14 @@
+# Node
+
+ℹ️ A node -  in Polkadot - is a binary executable, whose primary purpose is to execute the [runtime](../runtime/README.md).
+
+🔗 It communicated with other nodes in the network, and aims for [consensus](https://wiki.polkadot.network/docs/learn-consensus) among them.
```
gupnik, line 5:
```suggestion
🔗 It communicates with other nodes in the network, and aims for [consensus](https://wiki.polkadot.network/docs/learn-consensus) among them.
```

gupnik (approved):

Issue #4324, opened 2024-04-29
Validators entering the active set are slow on validation because PVF artifacts are not compiled


alexggh:
PVF artifacts are cleanup every 24h if unused: https://github.com/paritytech/polkadot-sdk/blob/f34d8e3cf033e2a22a41b505c437972a5dc83d78/polkadot/node/core/pvf/src/host.rs#L296

So, when nodes join the active set for the first time or they have been out of the active set for 24 hours, they won't the PVF artifacts ready for approving or backing blocks so they will have to compile all PVFs the first time they need to execute a block from each parachain. 

Each PVF compilation will take around  3 seconds or more and compiling all polkadot PVFs will take around 3 minutes, hence why the validator will cause no-shows on the approval-voting and will probably fail to get some of the backing points if the PVF of the parachain is not compiled yet.

This is a transient problems since all the PVF should be compiled after around 3-4 minutes, however this problem could probably be relatively easy to fix by proactively compiling the PVFs before the node become actives.

sandreim:
Good idea, nodes should be able to know that already by checking the next session keys.

s0me0ne-unkn0wn:
I had another idea lost somewhere in discussions: instead of invalidating the cache after 24h, make it size-bounded and only remove the stalest artifact if the cache size overflows. Not 100% sure but sounds like it's somewhat easier to implement than the lookahead compilation.

alexggh:
> I had another idea lost somewhere in discussions: instead of invalidating the cache after 24h, make it size-bounded and only remove the stalest artifact if the cache size overflows

That would work for nodes that enter and live the active set, however it won't work for the situation where the node is simply a fresh node that just joins the active set.

s0me0ne-unkn0wn:
> > 
> 
> That would work for nodes that enter and live the active set, however it won't work for the situation where the node is simply a fresh node that just joins the active set.

Fair enough, but that's just a single no-show. Running a validator from scratch is not something that happens very often, I believe.
Maybe we could implement both? Mine would save them some CPU for the price of some storage, and yours would address all the remaining corner cases.

alexggh:
> Maybe we could implement both? 

Yes, implementing both makes sense.

PR #4662, opened 2024-05-31
Cleanup PVF artifact by cache limit and stale time

AndreiEres:
Part of https://github.com/paritytech/polkadot-sdk/issues/4324
We don't change but extend the existing cleanup strategy. 
- We still don't touch artifacts being stale less than 24h
- First time we attempt pruning only when we hit cache limit (10 GB)
- If somehow happened that after we hit 10 GB and least used artifact is stale less than 24h we don't remove it.

### DIFF
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -177,6 +177,7 @@ pub enum CleanupBy {
 	// Inactive time after which artefact is deleted
 	Time(Duration),
 	// Max size in bytes. Reaching it the least used artefacts are deleted
+	#[allow(dead_code)]
 	Size(u64),
 }
 
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -289,8 +289,8 @@ impl Artifacts {
 		})
 	}
 
-	/// Remove artifacts older than the given TTL or the total artifacts size limit and return id
-	/// and path of the removed ones.
+	/// Remove artifacts older than the given TTL when the total artifact size reaches the limit
+	/// and return id and path of the removed ones
 	pub fn prune(&mut self, cleanup_config: &ArtifactsCleanupConfig) -> Vec<(ArtifactId, PathBuf)> {
 		let mut to_remove = vec![];
 		let now = SystemTime::now();
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -172,7 +172,7 @@ pub struct Artifacts {
 }
 
 /// Parameters we use to cleanup artifacts
-/// After we hit the cache limit we remove the least used artefacts
+/// After we hit the cache limit we remove the least used artifacts
 /// but only if they are stale more than minimum stale time
 #[derive(Debug)]
 pub struct ArtifactsCleanupConfig {
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -311,19 +311,18 @@ impl Artifacts {
 			let Some((artifact_id, path, size, last_time_needed)) = artifact_sizes.pop() else {
 				break
 			};
-			if now
+
+			let used_recently = now
 				.duration_since(last_time_needed)
 				.map(|stale_time| stale_time < cleanup_config.min_stale_time)
-				.unwrap_or(true)
-			{
+				.unwrap_or(true);
+			if used_recently {
 				break;
 			}
-			to_remove.push((artifact_id, path));
-			total_size -= size;
-		}
 
-		for artifact in &to_remove {
+			to_remove.push((artifact_id, path));
 			self.inner.remove(&artifact.0);
+			total_size -= size;
 		}
 
 		to_remove
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -314,7 +314,7 @@ impl Artifacts {
 			if now
 				.duration_since(last_time_needed)
 				.map(|stale_time| stale_time < cleanup_config.min_stale_time)
-				.unwrap_or(false)
+				.unwrap_or(true)
 			{
 				break;
 			}
--- a/polkadot/node/core/pvf/src/artifacts.rs
+++ b/polkadot/node/core/pvf/src/artifacts.rs
@@ -171,14 +171,31 @@ pub struct Artifacts {
 	inner: HashMap<ArtifactId, ArtifactState>,
 }
 
-/// A condition which we use to cleanup artifacts
+/// Parameters we use to cleanup artifacts
+/// After we hit the cache limit we remove the least used artefacts
+/// but only if they are stale more than minimum stale time
 #[derive(Debug)]
-pub enum CleanupBy {
-	// Inactive time after which artefact is deleted
-	Time(Duration),
+pub struct ArtifactsCleanupConfig {
 	// Max size in bytes. Reaching it the least used artefacts are deleted
-	#[allow(dead_code)]
-	Size(u64),
+	cache_limit: u64,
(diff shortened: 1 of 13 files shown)


polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -169,6 +169,15 @@ pub struct Artifacts {
 	inner: HashMap<ArtifactId, ArtifactState>,
 }
 
+/// A condition which we use to cleanup artifacts
+#[derive(Debug)]
+pub enum CleanupBy {
+	// Inactive time after which artefact is deleted
+	Time(Duration),
+	// Max size in bytes. Reaching it older artefacts are deleted
```
sandreim, line 177: The comment seems wrong, we delete least used ones. At least for me `older` usually means in relation to creation time.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -251,21 +260,47 @@ impl Artifacts {
...
 				}
-			}
+			},
+			CleanupBy::Size(size_limit) => {
+				let mut total_size = 0;
+				let mut artifact_sizes = vec![];
+
+				for (k, v) in self.inner.iter() {
+					if let ArtifactState::Prepared { ref path, last_time_needed, .. } = *v {
+						if let Ok(metadata) = fs::metadata(path) {
```
s0me0ne-unkn0wn, line 289: What bothers me here is that we're running a (possibly large) number of synchronous blocking filesystem operations in `Artifacts::prune()`, which is called from the PVF host's main loop running on a non-blocking threadpool. In case of I/O problems the whole PVF host will stall. I think we should either make use of `tokio::fs::metadata()`, or, even better, store the artifact's size as a property of the artifact itself along with the other data, and then no filesystem access is required inside `prune()`.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -251,21 +260,47 @@ impl Artifacts {
...
+						}
+					}
+				}
+				artifact_sizes.sort_by_key(|&(_, _, _, last_time_needed)| last_time_needed);
+
+				while total_size > *size_limit {
+					let Some((artifact_id, path, size, _)) = artifact_sizes.pop() else { break };
+					to_remove.push((artifact_id, path));
+					total_size -= size;
+				}
```
s0me0ne-unkn0wn, line 302: A unit test to check the correctness of this behavior would be definitely appreciated :)

AndreiEres: Added (and it saved me because the implementation was not correct)

polkadot/node/core/pvf/src/artifacts.rs
AndreiEres: Yeah, added the size to artifact's state

polkadot/node/core/pvf/src/artifacts.rs
AndreiEres: Fixed

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -251,22 +289,38 @@ impl Artifacts {
...
+			.sort_by_key(|&(_, _, _, last_time_needed)| std::cmp::Reverse(last_time_needed));
+
+		while total_size > cleanup_config.cache_limit {
+			let Some((artifact_id, path, size, last_time_needed)) = artifact_sizes.pop() else {
+				break
+			};
+			if now
+				.duration_since(last_time_needed)
+				.map(|stale_time| stale_time < cleanup_config.min_stale_time)
+				.unwrap_or(false)
```
alexggh, line 317: > Returns an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) if earlier is later than self, and the error contains how far from self the time is.

In case of error we should return true to break here, we don't want to delete things if time move backwards for whatever reason.

AndreiEres: good catch!

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -169,6 +171,33 @@ pub struct Artifacts {
...
+	// Max size in bytes. Reaching it the least used artefacts are deleted
+	cache_limit: u64,
+	// Inactive time after which artefact is allowed to be deleted
+	min_stale_time: Duration,
+}
+
+impl Default for ArtifactsCleanupConfig {
+	fn default() -> Self {
+		Self {
+			cache_limit: 10 * 1024 * 1024 * 1024,              // 1 GB
```
s0me0ne-unkn0wn, line 188:
```suggestion
			cache_limit: 10 * 1024 * 1024 * 1024,              // 10 GB
```

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -251,25 +289,40 @@ impl Artifacts {
...
+				break
+			};
+
+			let used_recently = now
+				.duration_since(last_time_needed)
+				.map(|stale_time| stale_time < cleanup_config.min_stale_time)
+				.unwrap_or(true);
+			if used_recently {
+				break;
+			}
```
s0me0ne-unkn0wn, line 321: Hmmm :thinking: So, the cache may grow unbounded as long as all the artifacts are fresh? I mean, this mimics the current behavior, so no new attack vectors are introduced here for sure, but probably we should review the old ones... I don't call for any change here right now, it's just for discussion.

s0me0ne-unkn0wn (approved):
Nice, thank you!

polkadot/node/core/pvf/src/artifacts.rs
AndreiEres: > the cache may grow unbounded as long as all the artifacts are fresh

I believe it growths no more than now because we use same 24h limit

s0me0ne-unkn0wn: Yes, exactly. I'm just wondering if there's some scenario where someone buys some cheap coretime leftovers and starts pushing tons of PVFs around within 24 hours to overflow validators' disk space. Sounds unlikely, just paranoia, probably.

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -169,6 +171,33 @@ pub struct Artifacts {
 	inner: HashMap<ArtifactId, ArtifactState>,
 }
 
+/// Parameters we use to cleanup artifacts
+/// After we hit the cache limit we remove the least used artefacts
```
sandreim, line 175: nit: Consistency

```suggestion
/// After we hit the cache limit we remove the least used artifacts
```

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -169,6 +171,33 @@ pub struct Artifacts {
...
+	// Max size in bytes. Reaching it the least used artefacts are deleted
+	cache_limit: u64,
+	// Inactive time after which artefact is allowed to be deleted
+	min_stale_time: Duration,
+}
+
+impl Default for ArtifactsCleanupConfig {
+	fn default() -> Self {
+		Self {
+			cache_limit: 10 * 1024 * 1024 * 1024,              // 10 GB
```
sandreim, line 188:
```suggestion
			cache_limit: 10 * 1024 * 1024 * 1024,              // 10 GiB
```

polkadot/node/core/pvf/src/artifacts.rs
```diff
@@ -251,25 +289,40 @@ impl Artifacts {
 		})
 	}
 
-	/// Remove artifacts older than the given TTL and return id and path of the removed ones.
-	pub fn prune(&mut self, artifact_ttl: Duration) -> Vec<(ArtifactId, PathBuf)> {
+	/// Remove artifacts older than the given TTL or the total artifacts size limit and return id
+	/// and path of the removed ones.
```
sandreim, line 293: This  doc comment needs fixing: we evict LRU artifacts only if we go over cache limit.

sandreim (approved):

